Issue #80, opened 2020-05-29
Lists and list comprehensions


yannham:
Currently, Nickel does not have lists nor arrays. We propose to add lists and a set of tools for manipulating them to the language, namely list operations and list comprehensions. The set of operations must be complete enough to allow users to avoid general recursion except for very few particular cases. This proposition, including syntax, is not final but rather a concrete basis for sparking discussion.

## Definition
A list is an immutable sequence of comma-separated expressions. Each element is stored inside a thunk, such that it is only evaluated if needed, and at most once. Elements can have different types.

Example: `["a", 2, "/a/b/c", false, {a={b=2;};}]`

## Operations

| Operation     | Syntax        | Description                                                                                              | Example                                                  |
|---------------|---------------|----------------------------------------------------------------------------------------------------------|--------------------------------------------------------- |
| concatenation | `l ++ l'`     | Concatenate two lists                                                                                    | `["a", 2] ++ ["3", {a=1;}] = ["a", 2, "3", {a=1;}]`     |
| access        | `elemAt l n`  | Access the nth element of l                                                                              | `elemAt ["a", 2, "3] 1 = 2`                             |
| map           | `map f l`     | Apply f to each element of l                                                                             | `map (fun x => x + 1) [1,2,3] = [2,3,4]`                |
| filter        | `filter f l`  | Return a list consisting of the elements of l for which the function f returns true                     | `filter (fun x => isNum x) [1, "2", 3, {a=4;}] = [1, 3]`|
| flatten       | `flatten f l` | Concatenate a list of lists                                                                              | `flatten [[1, 2], [3, 4]] = [1, 2, 3, 4]`         |
| fold left     | `foldl f x l` | Iteratively apply (strictly) the binary operator f on elements of l, from left to right, starting with x | `foldl (fun x y => x + y) 0 [1, 2, 3] = 6`              |
| membership    | `elem x l`    | Return true if x is an element of l, and false otherwise                                                 | `elem 3 [1, 2, 3] = true`                               |
| all           | `all p l`     | Return true if the predicate p is true for all elements of l, and false otherwise                        | `all (fun x => x % 2 == 0) [1, 2, 3] = false`           |
| any           | `any p l`     | Return true if the predicate p is true for at least one element of l, and false otherwise                | `any (fun x => x % 2 == 0) [1, 2, 3] = true`            |

## Comprehensions
A syntax for easing the construction of lists. We basically propose Haskell list comprehensions.

### Examples
```
[...]

PR #156, opened 2020-10-07
Add basic list functions in the stdlib

yannham:
Partly address #80. Depends on #153 and #155. Add the functions listed there to the Nickel standard library.

Some code is not ideal (such as `if isZero (length l)`), because there is a bootstrapping problem: it would need a polymorphic equality to become the more idiomatic `if length l == 0` (or pattern matching), but polymorphic equality on list needs `lists.all` first. This is a first working set of helpers: when new features are implemented, we can improve the standard lib accordingly.

**what it does**
 - add a bunch of list functions in `stdlib/lists.ncl` and load them in the global environment (in a `lists` record).
 - typecheck and apply program transformations on the stdlib parts, which was not done before.

### DIFF
--- a/stdlib/lists.ncl
+++ b/stdlib/lists.ncl
@@ -1,32 +1,37 @@
 {
   lists = {
-    concat = fun l1 l2 => l1 @ l2;
+    concat = Promise(List -> List -> List, fun l1 l2 => l1 @ l2);
 
-    foldl = fun f fst l =>
-      if isZero (length l) then
-        fst
-      else
-        let rest = foldl f fst (tail l) in
-        seq rest (f rest (head l));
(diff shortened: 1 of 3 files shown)


stdlib/lists.ncl
```diff
@@ -0,0 +1,37 @@
+{
+  lists = {
+    concat = Promise(List -> List -> List, fun l1 l2 => l1 @ l2);
+
+    foldl = Promise(forall a. (a -> Dyn -> a) -> a -> List -> a,
```
aspiwack, line 5: I'm rather surprised that we don't have a type `List b`. But I suppose this can be solved later.

aspiwack (approved):

stdlib/lists.ncl
yannham: True. Saved in #171.

Issue #157, opened 2020-10-08
Implement contracts for records and enums


yannham:
**Describe the bug**
Currently, the builtin contracts for row types are not implemented yet. This means that row types are unusable at runtime, since any term which is a `Promise` or `Assume` with a row type in the annotation leads to a panic if evaluated.

**To Reproduce**
Write an `Assume` or a `Promise` that is consumed during evaluation:

```
nickel <<< 'Assume({ {| a: Num |} }, {a = 1})'
thread 'main' panicked at 'TODO implement', src/types.rs:241:41
```

**Expected behavior**
We expect contracts to be implemented for enums and records, making the previous code to evaluate without errors.

PR #162, opened 2020-10-08
Add a fieldsOf operator to get record field names

yannham:
Required by #157 to implement records contracts in pure Nickel. Also useful in general. Add a builtin operator `fieldsOf` which takes a record and return the name of its fields a sorted string list. Same as [`attrNames`](https://nixos.org/manual/nix/stable/#ssec-builtins) in Nix.

### DIFF
--- a/src/operation.rs
+++ b/src/operation.rs
@@ -328,6 +328,21 @@ fn process_unary_operation(
                 ))
             }
         }
+        UnaryOp::FieldsOf() => {
+            if let Term::Record(map) = *t {
+                let mut fields: Vec<String> = map.keys().map(|Ident(id)| id.clone()).collect();
+                fields.sort();
+                let terms = fields.into_iter().map(|id| Term::Str(id).into()).collect();
+                Ok(Closure::atomic_closure(Term::List(terms).into()))
+            } else {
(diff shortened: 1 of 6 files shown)


src/operation.rs
```diff
@@ -328,6 +328,21 @@ fn process_unary_operation(
                 ))
             }
         }
+        UnaryOp::FieldsOf() => {
+            if let Term::Record(map) = *t {
+                let mut fields: Vec<String> = map.keys().map(|Ident(id)| id.clone()).collect();
+                fields.sort();
```
aspiwack, line 334: Is this necessary? Does it mean that our internal representation doesn't have a canonical field order?

aspiwack (approved):

src/operation.rs
yannham: Our internal representation is Rust's `HashMap`, and the documentation says that `keys()` visits the keys in arbitrary order. Nix does the same thing: `attrNames` returns a sorted list, probably for the same reason (grep for `attrNames` [here](https://nixos.org/manual/nix/stable/)).

aspiwack: Ok, this is quite fair.

PR #165, opened 2020-10-12
Do not allow hyphen in identifiers

### LINKED ISSUE #164: Do not allow the hyphen character in identifiers
**Describe the bug**
Currently, the lexer allows to use hyphens `-` in identifiers. This clashes with the use of, say, the dynamic field removing operator `-$`, or any binary operator starting with an hyphen.

**To Reproduce**
Using the field removing operator without spaces after an identifier causes an error:
```
nickel <<< 'let field = "a" in let rec = {a=1} in rec-$field'
error: Unexpected token
  ┌─ <stdin>:1:43
  │
1 │ let field = "a" in let rec = {a=1} in rec-$field
  │                                           ^

```

while adding a space makes parsing succeed:
```
nickel <<< 'let field = "a" in let rec = {a=1} in rec -$ field'
[...]
```

**Expected behavior**
Do not consider the hyphen `-` character as valid identifier character. This is not standard and clashes with potentially any binary operator starting with a `-`. Parsing `exp binop exp` should not be whitespace sensitive.

yannham:
Close #164.

### DIFF
--- a/src/parser/lexer.rs
+++ b/src/parser/lexer.rs
@@ -276,7 +276,7 @@ fn is_ident_start(chr: char, look_ahead: Option<char>) -> bool {
 
 fn is_ident_char(chr: char) -> bool {
     match chr {
-        '0'..='9' | '-' | '_' => true,
+        '0'..='9' | '_' => true,
         chr => is_ident_start(chr, None),
     }
 }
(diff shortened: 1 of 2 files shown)


aspiwack (approved):
I like how the choice was absolutely conscious when it was made.

PR #172, opened 2020-10-19
Correctly generate and check row constraints on polymorphic types

### LINKED ISSUE #144: The typechecker infers invalid row types (polymorphic row types are not correctly constrained).
**Describe the bug**
The typechecker accepts invalid programs when using polymorphic row types. In particular, it can infer a row type with multiple conflicting types for the same field identifier, such as `{ {| a: Num, a: Str |} }`.

**To Reproduce**
Here is an example of a bogus program being accepted:
```
$nickel <<< '
let extend = Assume(forall c. ({{| | c} }) -> ({ {| a: Str, | c } }), fun rec => rec$["a"="b"]) in
let bad = Promise(Str, (extend ({a = 1;})).a) in 0'
Typechecked: Ok(Types(Dyn))
```
Here, we have an extend function, which adds a field named `a` to a record and assigns it to a string. Then, this function is applied to the record `{a = 1}` which leads `c` to be unified with the type `{ {| a: Num |} }`. Consequently, the return type of `extend` inside `bad` ends up being `{| a: Str, a: Num |}` (I checked by printing the type inferred by the typechecker) which is not a valid row type.

**Expected behavior**
The typechecker should constrain row unification variables introduced by `forall`s a to not contain fields already defined in the rows they appear in. Here, it should constrain the unification variable introduced by `c` to not contain the field `a`, and fail accordingly when we try to unify it with `{ {| a: Num |} }`.

yannham:
Close #144. The problem was twofold:

1. Row constraints were not correctly generated.
2. Row constraints were not correctly checked. Depending on the shape the types being unified, the unification function did not always check constraints or update them accordingly.

**what it does**
- after instantiating a polymorphic type, the new `constrain_var` function is called. It takes the instantiated type and the freshly created unification variable as argument, visits the type recursively and look for every row type that contains the variable in tail position, of the form `{id1: Type1, .., idn: TypeN | var}`. For each such type, `id1, .., idn` are added to the row constraints of the variable.
- fix the checking of row constraints. Row constraints were enforced when unifying two populated row type of the form `{id1, .., idn | tail}` and `{id1', .., idn' | tail'}`, or when unifying two variables, but not when unifying a variable with a row type.
- add a regression test

### DIFF
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -528,7 +528,7 @@ fn type_check_(
                 .get(&x)
                 .ok_or_else(|| TypecheckError::UnboundIdentifier(x.clone(), pos.clone()))?;
 
-            let instantiated = instantiate_foralls_with(state, x_ty.clone(), TypeWrapper::Ptr);
+            let instantiated = instantiate_foralls(state, x_ty.clone(), ForallInst::Ptr);
             unify(state, strict, ty, instantiated)
                 .map_err(|err| err.to_typecheck_err(state, &rt.pos))
         }
@@ -629,7 +629,7 @@ fn type_check_(
         Term::Promise(ty2, _, t) => {
             let tyw2 = to_typewrapper(ty2.clone());
 
-            let instantiated = instantiate_foralls_with(state, tyw2, TypeWrapper::Constant);
+            let instantiated = instantiate_foralls(state, tyw2, ForallInst::Constant);
 
             unify(state, strict, ty.clone(), to_typewrapper(ty2.clone()))
                 .map_err(|err| err.to_typecheck_err(state, &rt.pos))?;
@@ -694,6 +694,7 @@ pub enum TypeWrapper {
 }
 
 impl TypeWrapper {
+    /// Substitute all the occurrences of a type variable for a typewrapper.
     pub fn subst(self, id: Ident, to: TypeWrapper) -> TypeWrapper {
         use self::TypeWrapper::*;
         match self {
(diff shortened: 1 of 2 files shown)


src/typecheck.rs
```diff
@@ -1577,6 +1597,113 @@ fn constraint(state: &mut State, x: TypeWrapper, id: Ident) -> Result<(), RowUni
...
+/// unified with a row type containing another declaration for the field `x`.
+///
+/// # Preconditions
+///
+/// Because `constraint_var` should be called on a fresh unification variable `p`, the following is
+/// assumed:
+/// - `get_root(state.table, p) == p`
+/// - `state.constr.get(&p) == None`
+fn constrain_var(state: &mut State, tyw: &TypeWrapper, p: usize) {
+    fn constrain_var_(state: &mut State, mut constr: HashSet<Ident>, tyw: &TypeWrapper, p: usize) {
```
aspiwack, line 1614: I'm not fully sure what is happening here. But I think that what you are doing is, when instantiating a `forall a. t` with a unification variable, you traverse `t` in search of row constraints. I fear that this may be a bit costly (there is some quadraticness to it).

An alternative strategy is to simply generate a unification variable, which doesn't have constraints. Then, when unifying two types, one of them being a record type, you can
1. generate a new type variable for this row, with the appropriate row constraints
2. unify the row variable with the new variable

aspiwack (approved):
It's a fine PR. Though I want us to consider the alternative below, probably in a separate issue/PR.

src/typecheck.rs
yannham: >I'm not fully sure what is happening here. But I think that what you are doing is, when instantiating a forall a. t with a unification variable, you traverse t in search of row constraints. I fear that this may be a bit costly (there is some quadraticness to it).

Yes, indeed.  For a type `forall a1 ... an . type`, it's `O(n*size(type))`.

If I understand correctly your suggestion, you propose to do this (constraint generation) at unification time, rather than instantiation time, so that:
1. Non-row parts of the type are not uselessly visited. Currently (this PR), instantiating `forall a b c d. Num -> Num  -> Num -> a -> b -> c -> d` does cause some useless work.
2. Rows that are not unified are not treated, while they currently are.

On the other hand, would that be possible that it duplicates some work in other instances ? First guess is in a situation where a unification would be immediate otherwise (say, unify a unification variable with a row type) and repeated on a same row type (several consecutive unification of different variables with the same row type). Then the row type would be visited many times in vain (the first visit is the. Not sure it can happen though (other than cheating with `Assume` to create otherwise uninhabited polymorphic types, like `Assume(forall a. {field: Num | a}, "I'm lying")`).

Another solution is to do type variable substitution and constraint generation in one pass, as substitution already traverses the type anyway.

aspiwack: Ah yes… substitution. In an ideal world, substitution would be performed lazily. If we can't have this, we would like to have a single substitution for all the variables. But, yes, I suppose that's an unavoidable cost, so maybe it works just as well to infer row constraints during substitution.

PR #174, opened 2020-10-20
Fix incorrect labels of merged contracts

### LINKED ISSUE #114: Incorrect labels for merged contracts
**Describe the bug**
When merging two contracts, a contract corresponding to their composition is generated. However, there is currently no way to merge or combine labels. Nickel currently makes the arbitrary and unsound choice of using the label of the first contract for the composite in order to appease the rust compiler, but this leads to error messages pointing to a wrong location when such composed contract fails.

**To Reproduce**
```
$ nickel <<< '(merge (merge {a=Contract(Num);} {a=Contract(Str);}) {a=1;}).a'
[..]
error: Blame error: [Contract
str].  The blame is on the value (positive blame)

  ┌─ <stdin>:1:18
  │
1 │ (merge (merge {a=Contract(Num);} {a=Contract(Str);}) {a=1;}).a
  │                  ^^^^^^^^^^^^^ bound here
```

**Expected behavior**
Have the error pointing to the actual failed contract if this is possible, or to another location (such as the whole merge) which may be less precise but is at least not wrong.
```
error: Blame error: [Contract
str].  The blame is on the value (positive blame)

  ┌─ <stdin>:1:18
  │
1 │ (merge (merge {a=Contract(Num);} {a=Contract(Str);}) {a=1;}).a
  │                                     ^^^^^^^^^^^^^ bound here
```

yannham:
Close #114. To compose two contracts, we create a new composite one, which requires its own label. Before, an arbitrary label was provided, and passed to each subcontract: in case of a blame error, it could point at an irrelevant location. Since the underlying contracts already possess their own label with the relevant information, we now pass a dummy label to the composite contract. The composite contract ignores it,  and each subcontract is called with its own original label. This correctly points at the failing subcontract if the composite contract fails.

**what it does**
 - to compose contracts `c1` and `c2`, instead of generating the contract `fun l t => c1 l (c2 l t)`, generate `fun _l t => c1 l1 (c2 l2 t)` where `l1` and `l2` are the respective original labels of `c1` and `c2`.

### DIFF
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -53,6 +53,7 @@
 //! evaluates to a contract check, that is an `Assume(..., t)`
 use crate::error::EvalError;
 use crate::eval::{Closure, Environment};
+use crate::label::Label;
 use crate::position::RawSpan;
 use crate::term::{BinaryOp, RichTerm, Term};
 use crate::transformations::Closurizable;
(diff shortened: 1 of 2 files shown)


src/merge.rs
```diff
@@ -186,13 +187,17 @@ pub fn merge(
...
+        // Composing two contracts creates a new one which applies these contracts successively.
+        // This composite contract requires its own label, but its components already had their
+        // own, hence they will ignore this label. We make the arbitrary choice of passing the
+        // first label `lbl1` in each such situation: because it will be ignored, it doesn't
+        // matter.
+        (Term::ContractWithDefault(ty1, lbl1, t1), Term::ContractWithDefault(ty2, lbl2, t2)) => {
             let Closure { body, mut env } = mk_merge_closure(t1, env1.clone(), t2, env2.clone());
             let body = Term::ContractWithDefault(
-                merge_types_closure(&mut env, ty1, env1, ty2, env2),
+                merge_types_closure(&mut env, ty1, lbl1.clone(), env1, ty2, lbl2, env2),
```
aspiwack, line 198: If you are going to not use the other copy of `lbl1`, maybe it's better to avoid a potentially non-trivial call to `clone` and instead, make a truly dummy label with no content.

aspiwack (approved):

PR #184, opened 2020-10-26
Note CUE's lazy evaluation

wbadart:
I could be misinterpreting the docs, since it's an off-handed comment in a sidebar, but the "[Logic of CUE][cue]" page alludes to lazy evaluation:

> This highly abstract definition determines almost everything about CUE. For instance, **lazy binding was not a design decision, but a direct consequence of following this definition.** It determines the possible evaluation strategies and what cycles mean, if allowed.

[cue]: https://cuelang.org/docs/concepts/logic/


yannham:
Even if the graph semantics of CUE sometimes makes you wonder if one should even use the same classification as for traditional languages, it seems that for all intent of purpose, you are right (see also https://github.com/cuelang/cue/issues/146#issuecomment-539088750). Thanks for your contribution !

Issue #191, opened 2020-10-30
Implement basic arithmetic, boolean and comparison operators


yannham:
**Is your feature request related to a problem? Please describe.**
Several basic operators are currently missing:
- arithmetic: `-`, `*`, `%`, `/`
- comparison: `<`, `<=`, `>`, `>=`
- boolean: `&&`, `||`, `!`

**Describe the solution you'd like**
Have these operators available in Nickel.

PR #192, opened 2020-10-30
[Refactor]Do not use one catch-all token for binops in the lexer

yannham:
Preparation for #191. The lexer was using one parametrized token for all infix binary operators (anything composed of one or more operator characters was matched). This makes sense in languages that allow user-defined operators, but otherwise, it just makes everything harder. This PRs removes this catch-all token and replace it by specific tokens for every binary operator.


Infinisil (approved):
Nice, this also improves the error message for unknown operators. Looking good!

PR #197, opened 2020-11-02
Multi-line indented strings

### LINKED ISSUE #82: String interpolation
The fate of a Nickel program is eventually to produce text in the form of configuration files. A fair amount of time is spent gathering and gluing together various pieces of strings in order to produce this configuration. Hence [string interpolation](https://en.wikipedia.org/wiki/String_interpolation), making text substitution easier, is a critical feature.

We propose to unashamedly reuse [Nix string interpolation](https://nixos.org/nix/manual/#idm140737322106128), which is well appreciated in the ecosystem and proved itself efficient for writing self-contained expressions, which mix strings containing shell commands, shell scripts, text descriptions, URLs, and so on. The syntax using `$` is common (Shell, JavaScript, Perl, Scala, Kotlin, Julia, etc.) while the required additional `{` avoid clashing with the simple shell interpolation `$var` when embedding scripts or commands inside strings.

In the following, we provide a concrete proposal, as a starting point for discussion.

## Examples
```
"here is a simple string"

/* indented multi-line string */
...
    let script = ''
        cd ~/.vim/
        if [ -f "vimrc" ]; then
            mv vimrc vimrc.bk
        fi
        touch vimrc
        '' in foo
/* script = 
"cd ~/.vim/
if [ -f "vimrc" ]; then
    cp vimrc vimrc.bk
fi
touch vimrc" */

/* interpolation */
let gnu = "GNU" in "${gnu} = ${gnu}'s Not Unix"
/* "GNU = GNU's Not Unix" */

/* nested interpolation */
let cmdFlag = "foo" in
let optionsFlag = "-foo" in
let cmdDefault = "bar" in
let options = "-bar" in
"${if flag then "${cmdFlag} ${optionsFlag}" else "${cmdDefault}"} ${options} $out"
/* if flag is true, gives "foo -foo -bar $out"
 * otherwise, "bar -bar $out"
 */
```

## Definition
We propose two ways of defining strings:

- The usual, existing double quote delimited literals: `"foo"`
- Additional double single quote delimited literals: `''foo''`

The alternative delimiter `''` is intended for enclosing strings that contain numerous `"`, such as shell commands, which would be then tiresome to escape. For indented strings enclosed by double single quotes, as the second example of the previous section, a number of spaces corresponding to the minimal indentation of the string as a whole is stripped from the beginning of each line. Also, the first line is removed if it is empty or contains only whitespaces. This makes it possible to write multi-line strings with the same level of indentation as the surrounding code without polluting the final string with unnecessary heading spaces.

Interpolation allows to embed a Nickel expression, which must evaluate to a string, inside another string using the `${expr}` syntax. Such expressions can be arbitrary Nickel expressions and may contain themselves nested interpolated strings.

## Syntax
**str ::=** `"`*strprts*`"` | `''`*strprts*`''`\
**strprts ::=** ε | *strprts* *STR* | *strprts* ${*e*}

where *e* is any Nickel expression, *STR* is a sequence of string characters (without the special sequence `${`, unless escaped), ε denotes an empty string and spaces between tokens denote concatenation. 

### Escaping
Inside `"`, `\` is used as usual to represent special sequences (`"\n"`, `"\t"`, and so on), to escape itself or to escape the interpolation sequence `${`. On the other hand, it has no special meaning alone inside `''`, and must be prefixed with `''` to recover it. For example, a tab is written `''''\t''`.

| Sequence | Escape/`"` | Escape/`''` |
|----------|------------|-------------|
| `${`     | `''${`     | `\${`       |
| `'`      | `'''`      | `'`         |
| `\`      | `\`        | `\\`        |

## Semantics
| Source       | Evaluation                     | Condition |
|--------------|--------------------------------|-----------|
| ε            |                                |           |
| strprts STR  | (eval strprts) STR             |           |
| strprts ${e} | (eval strprts) (eval e)        | isStr e   |
| `"`strprts`"`    | "eval(strprts)"                |           |
| `''`strprts`''`  | ''stripIndent (eval strprts)'' |           |

where stripIndent implements the behavior described above about smart indentation.

## Remarks
We may want to improve the behavior of Nix when substituting multi-line strings inside multi-line, indented strings (See [#543](https://github.com/NixOS/nix/issues/543) for Nix, or [#200](https://github.com/dhall-lang/dhall-lang/issues/200) for a similar discussion about Dhall).

yannham:
Close #82. Implement multi-line indented string *à la Nix*, with a few differences.

## Delimiters
As suggested in https://github.com/tweag/nickel/issues/82#issuecomment-646607817, this PR introduces asymmetric and variable-length delimiter, so that it is easier to not clash with external language syntax. It is inspired from Rust's [raw strings](https://doc.rust-lang.org/stable/rust-by-example/std/str.html), using the opening delimiters `m#"` (`m` stands for multiline) and closing `"#m`. As in Rust, one can use an arbitrary number of `#` in the opening delimiter, and the closing delimiter then have to match this number, which allows to write strings containing `m#"` or `"#m` without escaping:

```
let stringception = m##"
  let inner = m#"
    what, Nickel code
    inside Nickel String ?
  "#m in
  inner ++ " Wow !"
"##m
//Result: "let inner = m#\"what, Nickel code\ninside Nickel String ? Wow !\"#m in\ninner ++ \" Wow !\""
```

The choice of the precise Rust-like delimiters is quite arbitrary, as there doesn't seem to be a consensus among languages for raw string delimiters ([C++](https://en.cppreference.com/w/cpp/language/string_literal), [Java](https://openjdk.java.net/jeps/355), [C#](https://docs.microsoft.com/fr-fr/dotnet/csharp/language-reference/tokens/verbatim), [OCaml](https://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sss:stringliterals), [Python](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)) but it can be amended.

### General
Multiline strings strip leading indentation in the same way as Nix in the absence of interpolation:
- the minimal indentation among all lines is computed in a first pass. as the number of space or tabulation between each start of line and the first non-whitespace character
- an "empty" (containing only whitespace characters, or not character at all) is not taken into account for this calculation
- this indentation is stripped from the beginning of every line
- the first line and the last line are ignored if they are "empty"

### Interpolation
As requested in #82 (see the original issues https://github.com/NixOS/nix/issues/543 in Nix and https://github.com/dhall-lang/dhall-lang/issues/200 in Dhall), Nickel adds a specific case for the indentation of an interpolated expression which result itself contains newline characters.
The motivating case is the following: 
```
let inner = m#"
  BEGIN
    some code
  END
"#m in
let outer = m#"
  SOME
    INDENT
      ${inner}
    ENDINDENT
  ENDSOME
"#m
```
In this case, Nix does the following:
```
SOME
  INDENT
    BEGIN
  some code
END
  ENDINDENT
ENDSOME
```
One would arguably expect and want `${inner}` to act like a **block**, rather than just a string literal to insert, that is to have:
```
SOME
  INDENT
    BEGIN
      some code
    END
  ENDINDENT
ENDSOME
```
This is the behavior implemented in this PR. The indentation level of the interpolated expression is attached, and when it is evaluated, an indentation is inserted at each beginning of line. Note that this happens only if an interpolated expression is alone on its line (with the usual gotcha that whitespaces are allowed). That is, in the following case:
```
let inner = "a\nb" in
m#"
begin
  ${inner} ${inner}
end
"#m
```
`${inner}` is not reindented, because it is not alone on its line, giving:
```
m#"
begin
  a
b a
b
end
```
The rationale is that it becomes tedious to define a reasonable behavior with a simple implementation for the general case, and it does not really make sense to have two  interpolated expressions with new lines inside on the same line. In this case, one probably doesn't expect a "block" behavior. We thus restrict this re-indentation to the specific case above. Note that the previous example is easily transformed to trigger a re-indentation if needed:
```
let inner = "a\nb" in
m#"
begin
  ${inner ++ " " ++ inner}
end
"#m
```
```
begin
  a
  b a
  b
end
```

## Shortcomings / coming next
- as opposed to Nix, multi-line strings use the same `\`-based escape mechanism as standard strings. There is no hidden reasons, I just couldn't come up with a good re-escape sequence (`''` in Nix, because these are the delimiters of multi-line strings). Any ideas ?
- as opposed to Nix, but as in Dhall, tabulations are also counted as whitespace characters. I think this is desirable in general, as Nickel could very much needs to represent chunks from languages that favor them. On the other hand, it makes the re-indentation of interpolated string a bit more subtle: currently, what is saved is just a number of characters to re-add to the beginning of each line. But if the input used, say, only tabulations, the re-identation would turn those to spaces, which is maybe not desirable. Should we rather store a whole indentation prefix with interpolated expressions to fix this ? Or just forget about tabs ?
 - raw strings are a low-hanging fruit, just changing `m` to `r` and the special meaning of `${ }` and stuff.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -261,15 +261,15 @@ where
                     env: HashMap::new(),
                 },
                 Some(chunk) => {
-                    let arg = match chunk {
-                        StrChunk::Literal(s) => Term::Str(s).into(),
-                        StrChunk::Expr(e) => e,
+                    let (arg, indent) = match chunk {
+                        StrChunk::Literal(s) => (Term::Str(s).into(), 0),
+                        StrChunk::Expr(e, indent) => (e, indent),
                     };
 
                     Closure {
                         body: RichTerm {
                             term: Box::new(Term::Op1(
-                                UnaryOp::ChunksConcat(String::new(), chunks),
+                                UnaryOp::ChunksConcat(indent, String::new(), chunks),
                                 arg,
                             )),
                             pos,
@@ -712,7 +712,7 @@ mod tests {
     fn interpolation_simple() {
         let mut chunks = vec![
             StrChunk::Literal(String::from("Hello")),
-            StrChunk::Expr(
+            StrChunk::expr(
                 mk_term::op2(
                     BinaryOp::PlusStr(),
                     mk_term::string(", "),
@@ -721,7 +721,7 @@ mod tests {
                 .into(),
             ),
             StrChunk::Literal(String::from(" How")),
-            StrChunk::Expr(mk_term::if_then_else(
+            StrChunk::expr(mk_term::if_then_else(
                 Term::Bool(true),
                 mk_term::string(" are"),
                 mk_term::string(" is"),
@@ -741,7 +741,7 @@ mod tests {
     fn interpolation_nested() {
         let mut inner_chunks = vec![
             StrChunk::Literal(String::from(" How")),
-            StrChunk::Expr(
+            StrChunk::expr(
                 Term::Op2(
                     BinaryOp::PlusStr(),
                     mk_term::string(" ar"),
@@ -749,7 +749,7 @@ mod tests {
                 )
                 .into(),
             ),
-            StrChunk::Expr(mk_term::if_then_else(
+            StrChunk::expr(mk_term::if_then_else(
                 Term::Bool(true),
                 mk_term::string(" you"),
                 mk_term::string(" me"),
@@ -759,7 +759,7 @@ mod tests {
 
         let mut chunks = vec![
             StrChunk::Literal(String::from("Hello, World!")),
-            StrChunk::Expr(Term::StrChunks(inner_chunks).into()),
+            StrChunk::expr(Term::StrChunks(inner_chunks).into()),
             StrChunk::Literal(String::from("?")),
         ];
         chunks.reverse();
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -181,7 +181,7 @@ ChunkLiteral : String =
         })
     };
 
-ChunkExpr: StrChunk<RichTerm> = DollarBrace <t: SpTerm<RichTerm>> "}" => StrChunk::Expr(t);
+ChunkExpr: StrChunk<RichTerm> = DollarBrace <t: SpTerm<RichTerm>> "}" => StrChunk::Expr(t, 0);
 
 DollarBrace = { "${", "${ bis" };
 
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -183,17 +183,19 @@ ChunkLiteral : String =
 
 ChunkExpr: StrChunk<RichTerm> = DollarBrace <t: SpTerm<RichTerm>> "}" => StrChunk::Expr(t, 0);
 
-DollarBrace = { "${", "${ bis" };
+DollarBrace = { "${", "multstr ${" };
 
 Str: String = "\"" <s: ChunkLiteral> "\"" => s;
 
 ChunkLiteralPart: Either<&'input str, char> = {
     "str literal" => Either::Left(<>),
     "multstr literal" => Either::Left(<>),
+    "multstr $" => Either::Left(<>),
+    "multstr \\" => Either::Left(<>),
+    "multstr \\${" => Either::Left(<>),
     "false end" => Either::Left(<>),
     "str esc char" => Either::Right(<>),
-    "multstr esc char" => Either::Right(<>),
-};
+    };
 
 UOp: UnaryOp<RichTerm> = {
     "isZero" => UnaryOp::IsZero(),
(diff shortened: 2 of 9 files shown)


edolstra:
> as opposed to Nix, multi-line strings use the same \-based escape mechanism as standard strings. There is no hidden reasons, I just couldn't come up with a good re-escape sequence ('' in Nix, because these are the delimiters of multi-line strings). Any ideas ?

I think we don't need an escape mechanism, because you can always pick a delimiter that doesn't appear in the string. Using `\` as an escape character is undesirable because it's already special in many languages that we'd want to generate.

yannham:
> I think we don't need an escape mechanism, because you can always pick a delimiter that doesn't appear in the string. Using `\` as an escape character is undesirable because it's already special in many languages that we'd want to generate.

Sorry, I think I was not really clear, mixing two things:
 - one is re-escaping. We do want to disable the special meaning of `\`, but then what if we need to insert a special escape sequence ? I think Nix does it with `''\`, right ? where `''\n` becomes a new line again in a multi-line string. One solution is to abandon this totally and just ask people to use an interpolation for this: `${"\n"}`, and keep the interpolation as the only sequence with a special meaning.
- one is escaping. There is one thing we need to escape, which is said `${` sequence. That said the solution for the previous point also works for this one, even if a bit ugly: use `${"\${"}`.

edolstra:
We had a quick discussion and decided (for now) to only have `${`, `\${` and the close delimiter be special in multi-line strings.

In the future it might be worth considering to replace `${` with something derived from the delimiter (e.g. `#{`, `##{`, ..., depending on how many `#` there are in the delimiter).

yannham:
Note: the last commit allows dollars `$` which are not followed by a brace `{` to remain unescaped, as in Nix: `$this$is$fine$` (same in multiline strings). Previously, all dollars characters needed to be escaped.

yannham:
@edolstra , I don't know if you already had a look on the previous version, but it's ready for review.

mboes:
> In the future it might be worth considering to replace ${ with something derived from the delimiter (e.g. #{, ##{, ..., depending on how many # there are in the delimiter).

I think this would be very useful. In early design discussions for Nickel, we had opted for `#{}`, which Ruby uses, and which doesn't clash with Bash variables. I created a ticket to track this idea: https://github.com/tweag/nickel/issues/224.

PR #217, opened 2020-11-20
Open record contracts

### LINKED ISSUE #158: Dynamic row type for open records

yannham:
Close #158. Allow the dynamic type in row tail position to represent a record open contract. An open contract is a contract that checks the presence and the conformity of some fields, but also allow additional ones. Ex:

```
Assume({a: Num}, {a = 1; b = 2}) //error: extra field `b`
Assume({a: Num | Dyn}, {a = 1; b = 2}) //ok
```

## Syntax

The chosen syntax is similar to the one of [this paper](https://wgt20.irif.fr/wgt20-final35-acmpaginated.pdf), where the unitype `*` is used directly instead of a type variable. Another possible syntax would be `{a : Num | _: Dyn}`, which is slightly more verbose, but more consistent with the already existing dynamic record type `{_: Type}`. It would also make it possible to use a different type for the tail, like `{name: Str | _: Num}`. In this case we would probably get rid of the dynamic record type `{_: Type}` (which is currently a separate type), that could be written as `{ | _: Type}`.

## Typechecking

This PR extends the typechecker in a conservative way: a dynamic tail doesn't unify with anything but itself, and must explicitly be cast to/from, much as the standard dynamic type. It's probably gonna be used mainly for contracts, and a proper treatment of this kind of types doesn't look trivial, likely requiring something like subtyping/existential.

### DIFF
--- a/stdlib/contracts.ncl
+++ b/stdlib/contracts.ncl
@@ -65,6 +65,8 @@
         else
             acc$[magic_fld = wrap sy t];
 
+    dyn_tail = fun acc l t => acc & t;
+
     empty_tail = fun acc l t =>
         if t == {} then acc
         else blame (tag "extra field" l);
--- a/src/error.rs
+++ b/src/error.rs
@@ -87,6 +87,12 @@ pub enum TypecheckError {
         /* the inferred/annotated type */ Types,
         Option<RawSpan>,
     ),
+    /// A dynamic tail was expected to be in the type of an expression, but was not.
+    MissingDynTail(
+        /* the expected type */ Types,
+        /* the inferred/annotated type */ Types,
+        Option<RawSpan>,
+    ),
     /// A specific row was not expected to be in the type of an expression.
     ExtraRow(
         Ident,
@@ -94,6 +100,13 @@ pub enum TypecheckError {
         /* the inferred/annotated type */ Types,
         Option<RawSpan>,
     ),
+    /// A additional dynamic tail was not expected to be in the type of an expression.
+    ExtraDynTail(
+        /* the expected type */ Types,
+        /* the inferred/annotated type */ Types,
+        Option<RawSpan>,
+    ),
+
     /// An unbound type variable was referenced.
     UnboundTypeVariable(Ident, Option<RawSpan>),
     /// The actual (inferred or annotated) type of an expression is incompatible with its expected
@@ -956,6 +969,16 @@ impl ToDiagnostic<FileId> for TypecheckError {
                         format!("The type of the expression was inferred to be `{}`, which does not contain the field `{}`", actual,  ident),
                     ])]
             ,
+            TypecheckError::MissingDynTail(expd, actual, span_opt) =>
+                vec![Diagnostic::error()
+                    .with_message(String::from("Type error: missing dynamic tail `| Dyn`"))
+                    .with_labels(mk_expr_label(span_opt))
+                    .with_notes(vec![
+                        format!("The type of the expression was expected to be `{}` which contains the tail `| Dyn`", expd),
+                        format!("The type of the expression was inferred to be `{}`, which does not contain the tail `| Dyn`", actual),
+                    ])]
+            ,
+
             TypecheckError::ExtraRow(Ident(ident), expd, actual, span_opt) =>
                 vec![Diagnostic::error()
                     .with_message(format!("Type error: extra row `{}`", ident))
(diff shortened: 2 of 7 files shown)


regnat:
A few points regarding the syntax (as I know you love bikeshedding):
1. The use of `|` is a bit confusing as it's also the symbol used to declare meta-attributes − which is a totally different use-case
2. The way it was done in ptyx was that there was a `low-level` syntax (rougly) like `{a : Num; _: Foo}`, and there was a syntactic sugar `{ a: Num; ... }` for `{a : Num; _: Dyn }`. I guess that requires you to forbid `_` as an unquoted field name (or you could replace it with something already forbidden like `*`) but this has the advantage of unifying with the `{ _ : Foo }` notation (which would just be a special case of the above). Also I find it cool ;)

yannham:
> 1. The use of `|` is a bit confusing as it's also the symbol used to declare meta-attributes − which is a totally different use-case

True. By the way this is not specific to this PR, because it's a common symbol used for row types (in [Purescript](https://github.com/purescript/documentation/blob/master/language/Types.md#rows), e.g), and it's used in Nickel as such.

> 2\. The way it was done in ptyx was that there was a `low-level` syntax (rougly) like `{a : Num; _: Foo}`, and there was a syntactic sugar `{ a: Num; ... }` for `{a : Num; _: Dyn }`. I guess that requires you to forbid `_` as an unquoted field name (or you could replace it with something already forbidden like `*`) but this has the advantage of unifying with the `{ _ : Foo }` notation (which would just be a special case of the above). Also I find it cool ;)

This looks good, and `_` alone is already not accepted as an identifier.

stdlib/contracts.ncl
```diff
@@ -65,6 +65,8 @@
         else
             acc$[magic_fld = wrap sy t];
 
+    dyn_tail = fun acc l t => acc & t;
```
aspiwack, line 68: This is the heart of the change? Remarkable.

aspiwack (approved):

PR #241, opened 2020-12-04
Query sub-command

### LINKED ISSUE #232: `query` subcommand to retrieve metadata
Currently, there is no way to access the data contained in metavalues, in particular the documentation, which limits their usefulness.

Add a `query` sub-command to the interpreter which shows the metadata associated to a value. One could provide a field path: `nickel -f input.ncl query foo.bar` which would be equivalent to evaluating `let x = import "input.ncl" in x.foo.bar` and look at the metadata. If `x.foo.bar` is a record without contract, it could also list all fields and associated metadata of this record. Long flag may specify which part of the meta value once is interested in, e.g. `--doc`, `--contract`, `--default`, etc.

The documentation could also be formatted following Markdown (using [termad]()).

Example (this is illustrative, not necessarily the right output formatting):
```
$nickel query foo.bar <<< '{ foo = {bar | Str | doc "hello" | default 2}}'
foo.bar:
  - contract: Str
  - doc: hello
  - default value: Num(2)

$nickel query foo <<< '{ foo = {bar | Str | doc "hello" | default 2}}'
foo:
  * bar:
    - contract: Str
    - doc: hello
    - default value: Num(2)

$nickel query baz <<< '{ foo = {bar | Str | doc "hello" | default 2}; baz = "a"}'
No metadata found for field `baz`.
```

The syntax of the field path may also be more general, such as a term with a hole, which would allow stuff like `_.$"^$foo%".baz`. The future REPL should also have a similar command, but this is a separate issue.

yannham:
Depend on #234. Close #232.

**what it does**
- add required functions in `eval` and `program` to be able to evaluate a term, but to stop when a meta-value is reached (the default behavior of the interpreter is to silently force meta-values). This was already almost supported via the `enriched_strict` flag but needed small adaptations.
- add support for rendering markdown in the terminal. This feature can be disabled at compile time (an embedded nickel interpreter is probably better off without too much dependencies)
- add a `query` sub-command taking an optional path parameter. Evaluate the term (or `term.path` if `path` option is set) in non-strict mode (wrt metavalues) and print corresponding metadata. If the result is a record without metadata, or without the requested metadata, the list of its fields is printed. Documentation is interpreted as markdown if markdown support is enabled.

It uses the default markdown skin but this is configurable if we don't like the look & feel.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,6 +48,11 @@ version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "autocfg"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "beef"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -105,6 +110,11 @@ version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "clap"
 version = "2.33.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


edolstra:
Trying to use this I get an error about `seq`:
```
# cargo run -- query foo.bar <<< '{ foo = {bar = 2}}'
error: Unbound identifier
  ┌─ <query path>:1:24
  │
1 │ let x = (y.foo.bar) in seq x x
  │                        ^^^ this identifier is unbound
```

yannham:
@edolstra, I'm gonna rebase this branch on master and see what happens, but I suspect this PRs was made before the rework of primitive operators, so now they are merged I need to update this one to use the new syntax `%seq%` or `builtins.seq` most probably.

yannham:
@edolstra, it's good to go.

Issue #226, opened 2020-11-30
Improve the flow of typing information


yannham:
**Is your feature request related to a problem? Please describe.**
Currently, is not possible to use a `lists.*` function from the standard library in a typed context, although these functions are typed. This rather unfortunate situation is created by the following ingredients:
- libraries are currently just records. In particular, the top-level expression of the lists standard lib is of the form `{ lists = {..} }`
- unannotated bindings are given the `Dyn` type by default (see #139).

While lists functions are typed at the level of fields, the enclosing records are not. This type information doesn't propagate, and `lists` is just given the type `Dyn` in the typing environment, resulting in:

```
% nickel <<< 'let l : List = lists.concat [1,2] [3,4] in l'
error: Incompatible types
  ┌─ <stdin>:1:16
  │
1 │ let l : List = lists.concat [1,2] [3,4] in l
  │                ^^^^^ this expression
  │
  = The type of the expression was expected to be `{concat: _b -> _c -> List | _a}`
  = The type of the expression was inferred to be `Dyn`
  = These types are not compatible
```

**Describe the solution you'd like**
Several directions:
1. Make the typing information "bubble up" to the top-level record
    - Automatically. But this means inspecting every record literal searching for typed subfields. This would go against [the idea of not recursing to infer a type for untyped terms](https://github.com/tweag/nickel/issues/139#issuecomment-697249005).
    - Manually: having a construct `term : _` or `term : ?` to trigger typechecking, but ask the typechecker to infer the type for us. Then, just append ` : _` to library source files. This has a broader usage than just solving this issue. There is also the question of what to do with unification variables: if we don't want to generalize them, should we just set them to `Dyn` ? Or make it an error ?
2. Do not change typechecking, but argue that libraries should rather have a top level signature (see example [1]). This is how ML module are handled, with a signature separated from an implementation, the latter having possibly has few or no type annotations.
3. Do not change typechecking, but argue that libraries should not be treated as records. We could have, say, top-level let declarations, and `lists.foo` would not be a record operation but a namespace operation which would behave differently with respect to typechecking.

[1]:
```
//Going from
{
  lists = {
    concat: List -> List -> List = ...;
    ...
  }
}
// To
{
  lists = {
    concat = ...;
    ...
  }
} : {
  lists : {
    concat: List -> List -> List,
    ...
  }
}

yannham:
Superseded by #495 and partly solved by #498.

PR #243, opened 2020-12-08
Special syntax for fully-applied primitive operations

### LINKED ISSUE #228: Better handling of primitive functions and operators
Currently, primitive operators (either infix operators like `+` or builtin functions such as `length`) can only appear fully applied in the AST, inside the nodes `Op1(UnaryOp, Term)` and `Op2(BinaryOp, Term, Term)`. This has the following consequences:
 1. They must be parsed as specific keywords, that can't be used as identifiers elsewhere. As the list of operators will grow bigger in tandem with the standard lib, this will be polluting the global environment. Also, naming is not consistent: you write `lists.fold f l empty` because `fold` is written in pure Nickel in the lists standard library but write directly `map f l` or `length l` because they are primitive operators. It would be nice to have them in the corresponding libraries/namespaces only.
 2. It makes the process of adding an operator clumsier: first add the corresponding token in the lexer, then add some boilerplate to interface the lexer and the parser, then write the implementation in the abstract machine (although to be fair it's not that terrible either)
 3. They can't be partially applied, as in `let mapF = map f in ...`

## Proposals
To circumvent part of the above difficulties, one can either directly change the way operators are handled in the AST, or keep it as it is now but use a specific syntax for primitive operators.

### Without changing the AST
Use a special syntax, different from identifiers, for prefix primitive operators, like `%map%`. This wouldn't clash with identifiers anymore, and can be wrapped in libraries by defining e.g. `map = fun f l => %map% f l`. One could even allow this syntax only when parsing the standard library, such that in normal programs the only way to access `%map%` is to use the wrapper `lists.map`.

This solves 1. and 3., but 2. is harder: because operators must still be fully applied, the parser has to be aware of the arity of the operator, making it difficult to parse uniformly an application of `%op%`. Unless we have a different syntax for unary and binary operators. This also induces some overhead, namely calling the intermediate function.

### By changing the AST
Another direction is to make primitive operators first class terms, that can be applied to via the standard application. This raises the following questions:
 - Strictness: the usual function application is lazy, while primitive application is eager. Let's keep the `Op1` and `Op2` constructs. A fully applied operator (using standard application) then reduces to the corresponding eager operator application. Infix operators  could be parsed directly as operator application like before, to avoid the indirection.
 - Partial application: one needs to be able to represent the new WHNF corresponding to a partially applied binary operator (such as `map f`), either encoding it as a function `fun x => map f x`, or using a dedicated construct `PartialOp`.

## Conclusion
Both approaches amount roughly to the same thing: wrap primitive eager operators as lazy functions to make them compatible with standard application. The first approach makes prefix operators syntactically distinct, and wrap them in libraries as standard functions. It has the advantage of not touching the AST or the abstract machine, but makes 2. a bit clumsier to solve. The second approach avoids changing the syntax by supporting this wrapping directly in the AST. It gives a bit more flexibility, incurs a bit less execution overhead, and makes 2. simpler, because prefix operators are now just usual identifiers. The price to pay is modifying the AST and the abstract machine.

yannham:
Close #228. Use a specific syntax for fully applied built-in operators, `%operator%`, and wrap this primitive version in vanilla functions of the standard library.

This has the unfortunate consequence described in #226, that is that wrapped operator does not behave as unwrapped operator with respect to typing, but this is a separate topic that needs to be addressed anyway.

This also led to #244. This PR currently assigns correct but less ergonomic types - the ones with `Dyn` - to primitive operators wrapped in the standard library, while the actual primitive ones stay polymorphic. We'll have to resolve this discrepancy together with #244.

### DIFF
--- a/src/program.rs
+++ b/src/program.rs
@@ -181,13 +181,22 @@ impl Program {
     fn mk_global_env(&mut self) -> Result<eval::Environment, ImportError> {
         let mut global_env = HashMap::new();
 
-        self.load_stdlib("<stdlib/builtins.ncl>", nickel_stdlib::BUILTINS, &mut global_env)?;
+        self.load_stdlib(
+            "<stdlib/builtins.ncl>",
+            nickel_stdlib::BUILTINS,
+            &mut global_env,
+        )?;
         self.load_stdlib(
             "<stdlib/contracts.ncl>",
             nickel_stdlib::CONTRACTS,
             &mut global_env,
         )?;
         self.load_stdlib("<stdlib/lists.ncl>", nickel_stdlib::LISTS, &mut global_env)?;
+        self.load_stdlib(
+            "<stdlib/records.ncl>",
+            nickel_stdlib::RECORDS,
+            &mut global_env,
+        )?;
         Ok(global_env)
     }
 
--- a/src/program.rs
+++ b/src/program.rs
@@ -181,6 +181,7 @@ impl Program {
     fn mk_global_env(&mut self) -> Result<eval::Environment, ImportError> {
         let mut global_env = HashMap::new();
 
+        self.load_stdlib("<stdlib/builtins.ncl>", nickel_stdlib::BUILTINS, &mut global_env)?;
         self.load_stdlib(
             "<stdlib/contracts.ncl>",
             nickel_stdlib::CONTRACTS,
--- a/src/program.rs
+++ b/src/program.rs
@@ -580,7 +580,7 @@ mod tests {
 
     #[test]
     fn simple_type_check() {
-        let res = eval_string("let x = 5 in if isNum x then true else 1");
+        let res = eval_string("let x = 5 in if %isNum% x then true else 1");
 
         assert_eq!(Ok(Term::Bool(true)), res);
     }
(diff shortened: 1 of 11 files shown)


src/program.rs
```diff
@@ -181,12 +181,22 @@ impl Program {
...
             "<stdlib/contracts.ncl>",
             nickel_stdlib::CONTRACTS,
             &mut global_env,
         )?;
         self.load_stdlib("<stdlib/lists.ncl>", nickel_stdlib::LISTS, &mut global_env)?;
+        self.load_stdlib(
+            "<stdlib/records.ncl>",
+            nickel_stdlib::RECORDS,
+            &mut global_env,
+        )?;
```
aspiwack, line 199: Maybe `stdlib.rs` should export a list of all its Nickel modules. I feel that it's rather impractical to have to add said modules in two different files.

stdlib/lists.ncl
```diff
@@ -1,21 +1,31 @@
...
+    map : (Dyn -> Dyn) -> List -> List = fun f l => %map% f l;
+
+    elemAt : List -> Num -> Dyn = fun l n => %elemAt% l n;
+
     concat : List -> List -> List = fun l1 l2 => l1 @ l2;
 
     foldl : forall a. (a -> Dyn -> a) -> a -> List -> a =
       fun f fst l =>
-        if length l == 0 then
+        if %length% l == 0 then
```
aspiwack, line 17: Unrelated to this PR, but I hope that `%length% l` is O(1). Otherwise `foldl` (and `fold`) as written here is quadratic.

aspiwack (approved):

stdlib/lists.ncl
yannham: Yes, list are currently implemented as vectors so `length` is O(1). I suspect this was written because at that time equality was implemented only for numbers though, so `l == []` is probably more future-proof.

src/program.rs
yannham: True. I'm currently reworking the program module to prepare for the REPL anyway, I will do that.

Issue #103, opened 2020-07-22
Merging recursive record: pondering the semantics


aspiwack:
I've been thinking a bit about the problem of overriding à la Nixpkgs.

My thoughts is that it boils down to the semantics of merging recursive records. Or, a bit pedantically, merging codata.

Let me illustrate. Suppose we have a bunch of pkgs:

```nix
# nixpkgs.nix
rec {
  packageA = …;
  packageB = … packageA …;
}
```

I want to override `packageA` so that `packageB` sees the new version. The current way that we think about this is that `packageA` will have some default values, and to make an override, we will do something like

```nix
let nipxkgs = import ./nixpkgs.nix; in
merge nixpkgs {
  packageA = … some new stuff …;
}
```

This returns a new `nixpkgs` where `packageA` has been changed (or, rather, made more definite, as per the merging model). The question, then, is: what version of `packageA` does `packageB` see?

So far, we've been thinking of it as if recursive records were data: unfold the record as a tree, and merge recursively. It's a simple model, it makes sense. And `packageB` sees the original version of `packageA`. But the fact is that, by virtue of the field being lazy, recursive records are, really codata. And maybe, thinking of them this way allows us to define a merging scheme where `packageB` will see the overriden `packageA`. Pretty much like in https://github.com/tweag/nickel/issues/83#issuecomment-646621649 . I haven't thought this through, so I don't know if we can carry this thought to a happy conclusion. But that would certainly be an elegant explanation for overriding.

Related issues: #74, #83 .

yannham:
I think this discussion is superseded by RFC001 (#330).

PR #245, opened 2020-12-08
Lazy `map` operator

yannham:
Depend on #243. For some reason the `map` primitive operation on list was not lazy in its first argument, `f`, which is unjustified. This PRs fixes another small issue at the same time, which was that `map` and `mapRec` (renamed `recordMap` on the way) created lists and records respectively which content was not directly put into a thunk (closurized). This means that expression had to be recomputed multiple times in a call-by-name way, and was also responsible of serialization errors.


aspiwack:
Frankly, this whole laziness thing is a bit of a distraction. We don't have to care too much about laziness. And I assume that we will be revisiting the semantics in more details soon (such as for #103 ).

src/operation.rs
```diff
@@ -392,24 +392,57 @@ fn process_unary_operation(
...
         }
-        UnaryOp::MapRec(f) => {
+        UnaryOp::ListMap(f) => {
+            if let Term::List(ts) = *t {
+                let mut shared_env = Environment::new();
+                let f_as_var = f.body.closurize(&mut env, f.env);
+
+                // List elements are closurized to preserve lazyness of data structures. It
+                // maintains the invariant that any data structure only contain thunks (that is,
+                // currently, variables).
```
aspiwack, line 402: For instance, I think this is not a particularly useful invariant. But, again, let's revisit later.

aspiwack (approved):
Anyway, what's done is done. And it doesn't make things worse. So let's merge.

src/operation.rs
yannham: I agree that this may not very important in general, but currently, this allows for simple serialization: because this invariant is enforced, when you `deepSeq` a term, you are guaranteed to only encounter values or variables. Hence the serializer can just evaluate `let x = stuff in deepSeq x x`, walk the AST of the evaluated term and serialize basic values. Here, `let x = list.map (fun x => x + 1) [1,2,3]) in deepSeq x x` would leave thunks of the form `%somevar + 1` inside, on which the serializer choked with errors, so very concretely, you couldn't export a term with a `lists.map`, which was the actual motivation behind this PRs. Sorry, I should have made this clear.

PR #247, opened 2020-12-09
Built-in evaluation of a sequence of equalities

### LINKED ISSUE #163: Avoid (or reduce) allocations and function calls when computing equality
The current implementation of polymorphic equality #159 rewrites `{ a = expa1, b = expb1, ...} == { a' = expa2, b = expb2, ..}` to `list.all (fun x => x) [ expa1 == expa2, expb1 == expb2, ..]`, and then proceeds. We should avoid allocating too much stuff and making numerous functions calls (implicitly to `fold`, `head`, `tail`, etc.) when computing equality, which is a fundamental operation that should stay reasonably lightweight.

Implement the logic of the previous term directly in Rust land, avoiding Nickel functions calls and allocations as much as possible. One solution would be to add a builtin operator of varying arity that would basically be a native version of `list.all id`.

yannham:
Depend on #245. Close #163. Instead of reducing `{field1 = val1; ... ; fieldn = valn} == {field1 = val1'; ... ; fieldn = valn'}` to `lists.all id [val1 == val1', ..., valn == valn']`, reduce it to a specific built-in operator `MultiEq` which stores a list of pairs of terms `(ti,ui)` and computes the same value as `t1 == u1 && ... && tn == un` without allocating intermediate expressions, similarly to how `ChunksConcat` evaluate nested interpolated expressions.

For the story, this previous implementation of equality came to bite in a unexpected way. Following the typing issue described in #226, I tried to change the definition of the list standard library:

```
// From
lists = {
  concat: List -> List -> List = fun l1 l2 => l1 @ l2;
  ...
}
// To
lists : {
  concat: List -> List -> List,
  ...
} = {
  concat = fun l1 l2 => l1 @ l2;
  ...
}
```
Seems innocent. Now, if try to evaluate `{foo = 1} == {foo = 1}`, and we get an infinite loop! Here is what happens:
1. Since we compare two records, this triggers the reduction defined above, so we evaluate `lists.all id [1 == 1]`.
2. We push the arguments on the stack, and evaluate `lists.id`. But `lists` is typed, so it comes with a contract. Before changing the type of the lists library, `lists` itself wouldn't have triggered any check, and `all` would have triggered the `List -> List -> List` contract. But here, we check the record contract `{concat : ... }` on `lists`
3. Once it popped all the fields to apply individual contracts, the record contracts check that there are no extra fields in the record. For example, you want to reject `{foo =1; bar = 2}` for `{foo : Num}` because it has an extra fields. This leads to evaluating `tail == {}` at some point in the definition on the builtin record contract.
4. Cool, let's evaluate an equality of empty records! So, it's reduced to `lists.all id []`, so we need to check the contract for `lists` and...oh....

The moral of the story is that allocating and calling to stdlib function to evaluate something fundamental such as equality is already questionable, but requiring the evaluation of *contracts* for this is flat out stupid.


src/operation.rs
```diff
@@ -1,4 +1,4 @@
-//! Implementation of primitive operations.
+//! Implementat Closure::atomic_closureon of primitive operations.
```
aspiwack: Let me assume that this is a wild copy/paste

```suggestion
//! Implementation of primitive operations.
```

src/operation.rs
```diff
@@ -838,64 +852,77 @@ fn process_binary_operation(
...
-            })
+            let mut eqs: Vec<_> = eqs
+                .into_iter()
+                .map(|(c1, c2)| {
+                    (
+                        c1.body.closurize(&mut env, c1.env),
+                        c2.body.closurize(&mut env, c2.env),
+                    )
+                })
+                .collect();
```
aspiwack, line 906: I'm confused. Isn't everything properly closurized by `eq`, by assumption?

src/operation.rs
```diff
@@ -838,64 +852,77 @@ fn process_binary_operation(
...
-                    let (left, center, right) = merge::hashmap::split(m1, m2);
+            match eq(&mut env, c1, c2) {
+                EqResult::Bool(b) => Ok(Closure::atomic_closure(Term::Bool(b).into())),
+                EqResult::Eqs(fst, snd, eqs) => Ok(Closure {
+                    body: RichTerm::new(Term::Op2(BinaryOp::MultiEq(eqs), fst, snd), pos_op),
+                    env,
+                }),
+            }
+        }
+        BinaryOp::MultiEq(eqs) => {
```
aspiwack, line 880: I'm wondering, though. Instead of making an operation which stores a bunch of values, how about making a binary operation which returns a function:

- It forces the two values on the stack, and test them for equality
  - If not equal, pop the top of the stack, push `false` (that's the function `const false`)
  - If equal, then evaluate the value on the top of the stack (that's `id`)

Then
- The standard equality test `a == b` is `%eqThen% a b True`
- You can push stuff onto the stack itself to store computation (some care has to be taken, as you can only use the top of the stack, but the point is that we are not using an extraneous storage that the virtual machine doesn't control)

An alternative may be, instead of returning a boolean function, to take a continuation as the last argument

While we are fiddling with this semantics, it's worth considering.

src/operation.rs
yannham: Indeed, `eq` already closurizes everything.

This confusing trickery is necessary because of the (old) way operators are closurized. Operator are parametrized by a `CapturedTerm` type which represents the type of captured terms, that is stored terms other than their argument (typically, for `MultiEq`, the list of equalities to process). This parameter is initially instantiated with `RichTerm`, and when the abstract machine evaluates an application of an operator to some arguments, it maps these potential captured terms  to `Closure`s with an appropriate environment. So at this point, `eqs` is of type `Vec<(Closure, Closure)>`[1].

This is really just a representation/typing problem: at the beginning of this snippet, eqs is `Vec<(Closure, Closure)>`, while what you need are `RichTerm`, so you have to closurize `ci`s back to `RichTerm`s. This adds useless overhead, and this ought to be fixed, by making operators always hold `RichTerm` and use the standard `closurize()`[2] method instead.

-----------------------------------
[1]: `Closure` is basically a product `(RichTerm, Environment)`\
[2]: `closurize()` acts on a `RichTerm` and returns a `RichTerm`: it generates a fresh variable and bind it to a closure with the term in some given environment, then return this variable.

src/operation.rs
yannham: Currently `MultiEq` is a binary operator (it is applied to the two next operands to force and test for equality in the sequence of equalities) which stores a continuation of the remaining equalities to test as an inner parameter `Vec<(RichTerm, RichTerm)>`. What about the following changes:

- instead of storing `eqs` inside the `MultiEq` operator, push the whole list as one element on the stack (or as a sequence of elements using a stack marker to determine the end). This is the continuation of equalities to test.
- when processing the next equality, instead of returning an application `MultiEq t1 t2` which would go back to the main eval loop, push the continuation `(MultiEq, t2)` on the stack, and goes on to evaluate `t1`, we can inline this step in `MultiEq` and directly push the continuation `(MultiEq, t2)` on the stack and go on evaluating `t1`, so that we don't create the intermediate `MultiEq t1 t2`.

Then `MultiEq` would test for equality its two evaluated argument, and proceed as before. If new equalities needs to be added, we can insert them in the `eqs` continuation on the top of the stack. If one equality is false, we discard the top of the stack (or everything up to the marker) and return false. If one equality is true, we pop the next one to compute from the top of the stack, or return true if there's no more.

This would be more or less the same process as the current implementation, but storing `eqs` and next terms to proceed on the stack rather than as a `MultiEq` application, which is what your suggestion is about, if I understand correctly. Would that do ?

aspiwack: You don't need the `MultiEq` to be an n-ary operator, because it's annoying to write. The binary-operator-returning-a-function idea avoids this: you just have a binary operator, and the third element does not need to be forced (well, I mean, you need to eventually enter it if the first equality is true, since it's basically a thunk to the boolean value that you need).

On the other hand, this approach does build up quite a bit of heap values, which may not be very nice. If you want to use purely the stack, you need either a stack marker or a count (it is worth noting that pushing everything to the stack, with a bit of care, flattens all the equalities to test, so you really have a single loop going on, which is nice). What I would do, there, is that `MultiEq ; t1 :: t2 :: stack`, if `t1 == t2` would step to `MultiEq ; stack`. Still a binary operation.

You have to be careful to clear the stack when `t1 /= t2`, though.

yannham: > On the other hand, this approach does build up quite a bit of heap values, which may not be very nice. If you want to use purely the stack, you need either a stack marker or a count (it is worth noting that pushing everything to the stack, with a bit of care, flattens all the equalities to test, so you really have a single loop going on, which is nice). What I would do, there, is that `MultiEq ; t1 :: t2 :: stack`, if `t1 == t2` would step to `MultiEq ; stack`. Still a binary operation.
> 
> You have to be careful to clear the stack when `t1 /= t2`, though.

Yes, this is what I was referring to, even if I strangely wanted to store the continuation as one big vector on the stack first, rather than directly as stack consecutive elements, I don't know why. Anyway, this is implemented in c82eed5.

aspiwack (approved):

PR #251, opened 2020-12-14
Eliminate the inner terms of primitive operators 4: DynExtend

yannham:
Depend on #250. Follow-up of #247, #248, #249, and #250. Last PR of this series. Use the stack to store the non-strict argument of `DynExtend`. `DynExtend` is the dynamic record extension operator, currently written as `someRecord $[foo=bar]` which extends `someRecord` with the field `field = bar` if the `foo` expression evaluates to the string `"field"`, where `bar` is said non-strict argument. This terminates the elimination of the dependency of `UnaryOp` and `BinaryOp` on `RichTerm`.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -295,11 +295,6 @@ where
                 Closure { body: t, env }
             }
             Term::Op2(op, fst, snd) => {
-                let op = op.map(|t| Closure {
-                    body: t,
-                    env: env.clone(),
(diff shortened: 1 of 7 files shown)


edolstra:
@yannham There is a merge conflict, can you resolve it please?

yannham:
@edolstra, it's done.

PR #265, opened 2021-01-19
ASCII escape sequences

### LINKED ISSUE #256: Full ASCII Hex escape sequence support for string literals
Current workarounds include:

- https://github.com/numtide/devshell/blob/17f46732ce299daa2977be2978f60d258c2d1b41/mkDevShell/options.nix#L19-L21
- https://github.com/nix-community/home-manager/blob/master/modules/programs/alacritty.nix#L33
  - and then: https://github.com/nix-community/home-manager/blob/3627ec4de58d7fbda13c82dfec94eace10198f23/modules/programs/alacritty.nix#L53-L60

Add escape sequences for literals addressing them by their ASCII Hex values eg. `\x1B` (`1B` = `<ESC>`)

Continue to work around limited ASCII support in literals. See above.

**Additional context**

- https://notes.burke.libbey.me/ansi-escape-codes/
- http://www.asciitable.com

This is especially relevant in the context of configuring input methods &mdash; a supposed first class use case for nickel via nix.

yannham:
Close #256.

### DIFF
--- a/src/parser/tests.rs
+++ b/src/parser/tests.rs
@@ -1,18 +1,19 @@
 use super::lexer::{Lexer, LexicalError, NormalToken, StringToken, Token};
+use crate::error::ParseError;
 use crate::identifier::Ident;
 use crate::term::make as mk_term;
 use crate::term::Term::*;
 use crate::term::{BinaryOp, RichTerm, StrChunk, UnaryOp};
 use crate::{mk_app, mk_switch};
+use assert_matches::assert_matches;
 use codespan::Files;
 
-fn parse(s: &str) -> Option<RichTerm> {
+fn parse(s: &str) -> Result<RichTerm, ParseError> {
     let id = Files::new().add("<test>", String::from(s));
 
     super::grammar::TermParser::new()
         .parse(id, Lexer::new(&s))
-        .map_err(|err| println!("{:?}", err))
-        .ok()
+        .map_err(|err| ParseError::from_lalrpop(err, id))
 }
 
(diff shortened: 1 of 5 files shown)


src/parser/tests.rs
```diff
@@ -130,8 +131,8 @@ fn functions() {
 
 #[test]
 fn lets() {
-    assert!(parse("let x1 = x2 in x3").is_some());
-    assert!(parse("x (let x1 = x2 in x3) y").is_some());
+    assert!(parse("let x1 = x2 in x3").is_ok());
```
regnat, line 134: Mostly nitpicking, but I guess this won't print anything helpful if the parsing fails. Maybe you could have something more verbose with

```suggestion
    assert_matches!(parse("let x1 = x2 in x3"), Ok(..));
```

src/parser/tests.rs
```diff
@@ -270,6 +274,39 @@ fn str_escape() {
...
+        Err(ParseError::InvalidAsciiEscapeCode(..))
+    );
+    assert_matches!(
+        parse("\"\\xab\""),
+        Err(ParseError::InvalidAsciiEscapeCode(..))
+    );
+    assert_matches!(
+        parse("\"\\xFF\""),
+        Err(ParseError::InvalidAsciiEscapeCode(..))
+    );
```
regnat, line 303: Might be worth to also add some tests ensuring that the escape codes are ignored in multi-line strings

regnat (approved):

PR #268, opened 2021-01-20
Remove extended parser and use directly LALRPOP

yannham:
#264 introduced an `extended` module for the REPL that wraps two different parsers and try them in order, one for usual terms and one for toplevel let. It turns out it's not required as we can directly make a dedicated rule that accepts both constructions in the grammar, as implemented by this PR, which is simpler.


regnat (approved):
Nice :) That's indeed cleaner and more concise

PR #269, opened 2021-01-20
[Fix]Swapped control sequences Ctrl+D and Ctrl+C in REPL

yannham:
Ctrl+D (which should exit) and Ctrl+C (which should abort current input) mapping were swapped in the REPL. This PR fixes it. On the side, it also removes the empty line printed after a successful toplevel let.


Infinisil:
Looking good, the only unnatural behavior now is that an Enter with an empty line doesn't print a new `nickel>` prefix.

PR #271, opened 2021-01-21
[Fix]Cyclic type inference

### LINKED ISSUE #270: Infinite loop during typechecking
**Describe the bug**
Typechecking some recursive functions, such as `{gen_ = fun acc x => if x == 0 then acc else gen_ (acc @ [x]) (x - 1)}.gen_ [] : Num -> List` may results in a stack overflow.

**To Reproduce**
```
$nickel typecheck <<< 'let gen = 
  {gen_ = fun acc x => if x == 0 then acc else gen_ (acc @ [x]) (x - 1)}.gen_ [] : Num -> List
  in gen 10'

```

**Expected behavior**
The command returns without error.

**Additional context**
Debugger shows that the looping function is `typecheck::get_root`, meaning that there is a loop between two unification variables in the unification table, probably introduced by the typing rule for recursive records.

yannham:
Fix #270. The typechecking algorithm was creating a cycle in the unification table when unifying a variable with itself.

Note that infinite loops are still possible because there's currently no occur check:
```
nickel typecheck <<< '{g = g 0}.g : Num'

```

This will be fixed in another PR.

### DIFF
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -1123,6 +1123,7 @@ pub fn unify_(
                 TypeWrapper::Concrete(ty2),
             )),
         },
+        (TypeWrapper::Ptr(p1), TypeWrapper::Ptr(p2)) if p1 == p2 => Ok(()),
         // The two following cases are not merged just to correctly distinguish between the
         // expected type (first component of the tuple) and the inferred type when reporting a row
         // unification error.
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -2267,7 +2267,7 @@ mod tests {
     }
 
     /// Regression test following [#270](https://github.com/tweag/nickel/issues/270). Check that
-    /// uniyfing a variable with itself doesn't introduce a loop. The failure of this test result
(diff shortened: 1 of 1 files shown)


aspiwack (approved):
Honestly, I'm not sure what's happening here. This looks like a suspicious way to handle recursion to be honest. But it's a very simple fix, so I don't see why not merging it.

Is `Ptr` the name of unification variables?

yannham:
> Honestly, I'm not sure what's happening here. This looks like a suspicious way to handle recursion to be honest. 

This is related to unification constraints of the form `_a' = _a'` that were mishandled. In union-find, when performing a union, you need to check that the two components are not already the same before assigning the parent pointers of one to the other (e.g. https://docs.rs/union-find/0.3.2/src/union_find/quick_union.rs.html#53). This check was not performed here, leading unification constraints `_a' = _a'` to generate a cyclic node which parent pointed to itself. Then, a find operation leads to an infinite loop. This PR fixes this by adding the missing check.

> Is `Ptr` the name of unification variables?
Yes.

aspiwack:
Got it. Thanks.

Issue #244, opened 2020-12-08
Ergonomic gradual typing


yannham:
**Is your feature request related to a problem? Please describe.**
While working on primitive operators, I had to wrap them as functions of the standard library. It unveiled a typing issue: a lot of primitive operators are typed with polymorphic arguments where they should really use `Dyn`. For example, `isNum` is currently typed as `forall a. a -> Bool` (not exactly, but this is equivalent to this), but it clearly doesn't respect parametricity.

Indeed, if you do `let f : forall a. a -> Bool = fun x => isNum x`, then `f` is well-typed, but will be blamed at runtime (or ends up on a type error because of a wrapped value, but this is morally a blame error), and rightly so, because it tries to inspect its argument `x`. This means the language is not blame safe in an obvious and fundamental way, that is, *well-typed program **can** be blamed*. The conclusion is that the actual type of `isNum` is `Dyn -> Bool`.

The rationale behind polymorphism was ergonomics of typed code: the polymorphic version lets you write `let x : Bool = isNum 2`, while you need an explicit cast on the argument `2` with the `Dyn -> Bool` version: `let x : Bool = isNum (2 | Dyn)`, although casts to `Dyn` can never fail (excepted on type variables, for exactly the same reason as explained above).

**Describe the solution you'd like**
It could be nice to automatically infer and insert (or elide) casts to `Dyn` (unless the type of the source is a rigid type variable, that is a type variable that is generalized, since this is doomed to fail at run-time), so that we pay no penalty for using the right type for primitive operators, and more generally, we can call functions with `Dyn` arguments seamlessly.

From there, we can go the subtyping route, introducing a subtyping relation generated by the rule `T <= Dyn` for any type `T`, and the usual variance rules for lists, records, and functions. I imagine this relation is decidable (~first guess is that an algorithm to check this would be to replace each occurrence of `Dyn` by a fresh unification variable and see if standard unification succeeds~ we can probably come up with a simple syntax-directed system).

**Describe alternatives you've considered**

- keep the current state of affairs with special casing for the type of primitive operators, which stay polymorphic, and accept that Nickel is not blame safe.
- change the type of primitive operators to use `Dyn` when required, but without inferring casts automatically. This would remove this source of blame unsafety but force the users to write more explicit casts to `Dyn`. 
[...]

PR #272, opened 2021-01-21
Parametrized list type

### LINKED ISSUE #171: Parametrized list type
Currently, we just have a `List` type for heterogeneous lists. A parametrized list type `List t` is more expressive, as it make the old `List` type representable as `List Dyn`, but it also makes it possible to represent lists which hold items of one fixed type, which is a common thing to use.

Have a parametrized list type `List t`, where `t` is a type.

yannham:
Close #171. The list type is now parametrized by the type of its elements. The standard `List` is kept as an alias for `List Dyn`, both in the grammar and in error reporting messages. The list contract is updated and the types of the lists module of the stdlib are improved accordingly.

This introduces a slight regression, that is the typechecker now rejects `[1, "a", false] : List` because currently nothing unifies with the dynamic type, unless an explicit case is used. We could make a special case to have this typecheck again, but it seems adhoc and is actually part of a bigger issue to think about, see #244 for example. Maybe bidirectional typechecking could also help here.


src/error.rs
```diff
@@ -750,7 +754,7 @@ impl ToDiagnostic<FileId> for EvalError {
                 let mut msg = String::from("Blame error: ");
 
                 // Writing in a string should not raise an error, whence the fearless `unwrap()`
```
aspiwack, line 756: (unrelated, but I saw the typo while going through the diff)
```suggestion
                // Writing in a string should not raise an error, hence the fearless `unwrap()`
```

src/types.rs
```diff
@@ -316,6 +318,19 @@ impl Types {
             }
         }
     }
+
+    /// Determine if a type is an atom, that is a either an atom or a type delimited by specific
```
aspiwack, line 322: This is a slightly confusing sentence.

It also feels like a bit overspecialised. More generally, you may need to know whether something has tighter priority than its surrounding or not. Not an urgent thing to fix, though.

aspiwack (approved):

Issue #139, opened 2020-09-09
Basic type inference outside of statically typed code


yannham:
**Is your feature request related to a problem? Please describe.**
At the typechecking phase, the typechecker maintains an environment mapping variables introduced by let expressions to types. Indeed, whether a variable was introduced in a statically typed `Promise` or outside, it can be used inside a `Promise` at some point and the typechecker must thus know about its type, or have at least an approximation.

The current rule to determine the type of a variable introduced *outside* of a `Promise` is:
 - if the variable is annotated, meaning that it is introduced by an expression of the form `let var = Assume(type, ...) in ...` or `let var = Promise(type, ...) in ...`, then use the user-defined `type`.
 - otherwise, use the unitype `Dyn`

This is an understandable rule in general as Nickel is untyped by default: outside of a statically checked `Promise` block, the typechecker shouldn't - and in general, can't - infer the type of an expression. But assigning the type `Dyn` to a variable forces the user to resort to explicit type casts (AKA `Assume`s) when using it inside a `Promise`:

```
$nickel <<< 'let magic_number = 2 in
Promise(Num -> Num,
  fun x => x + magic_number
)'
Typechecked: Err("The following types dont match Num -- Dyn")

$nickel <<< 'let base_path = "/home/me/nickel/" in
let full_path = Promise(Str -> Str,
    fun filename => base_path ++ filename ++ ".ncl"
) in
full_path "myfile"'
Typechecked: Err("The following types dont match Str -- Dyn")
```

**Describe the solution you'd like**
We could improve a bit the rule above, which determines the type associated to a variable, to accommodate simple common cases as long as the rule stays cheap and simple. For example:
 - If the bound term is a constant, as in `let x = 1 in` or `let x = "str" in `, deduce its type. This would make the previous examples to be accepted by the typechecker.
 - If the bound term is also a variable, as in `let x = y`, then use the type determined for `y`.
 - and others that we haven't thought of yet.

yannham:
Closed #273.

PR #273, opened 2021-01-22
Shallow type inference

yannham:
Depend on #272. Address #139. Determine the type of some unannotated let-bound expressions outside of statically typed code:

- **constants** (of the language): number literals, string literals (without interpolated expression), booleans
- **lists**: list literals are given the `List Dyn` type.
- **variables**: retrieve the type from the environment.

**what this PR doesn't do**
- infer function types (`Dyn -> Dyn`). It's not clear in what practical situation you would actually need it. If needed, we can do add this later.
- infer record types: Same as functions. Plus, there are two possible competing types `{_ : Dyn}` and `{ | Dyn}`, but this is maybe a sign that this a problem with the type system's definition, rather than with inferring the type of records.

### DIFF
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -792,13 +792,6 @@ impl Into<TypeWrapper> for ApparentType {
 /// - Otherwise, return an approximation of the type (currently `Dyn`, but could be more precise in
 ///   the future, such as `Dyn -> Dyn` for functions, `{ | Dyn}` for records, and so on).
 pub fn apparent_type(t: &Term, envs: Option<&Envs>) -> ApparentType {
-    fn is_literal(chunk: &StrChunk<RichTerm>) -> bool {
-        match chunk {
-            StrChunk::Literal(_) => true,
-            _ => false,
-        }
-    }
-
     match t {
         Term::Assume(ty, _, _) | Term::Promise(ty, _, _) => ApparentType::Annotated(ty.clone()),
         Term::MetaValue(MetaValue {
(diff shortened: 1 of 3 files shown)


src/typecheck.rs
```diff
@@ -734,36 +726,106 @@ fn type_check_(
...
-        }) => apparent_type(v.as_ref()),
-        _ => None,
+        }) => apparent_type(v.as_ref(), envs),
+        Term::Num(_) => ApparentType::Exact(Types(AbsType::Num())),
+        Term::Bool(_) => ApparentType::Exact(Types(AbsType::Bool())),
+        Term::Sym(_) => ApparentType::Exact(Types(AbsType::Sym())),
+        Term::Str(_) => ApparentType::Exact(Types(AbsType::Str())),
+        // If there are interpolated expressions inside a string, we can't guarantee that this is a
+        // well-typed string (e.g. "#{1 + 1}" is not). We can only do so if all chunks are string
+        // literals.
```
aspiwack, line 819: But isn't it guaranteed that _if_ it is well-typed (dynamically), then it is a string? In which case it's fine to return the string type in all cases.

src/typecheck.rs
```diff
@@ -734,36 +726,106 @@ fn type_check_(
...
+        Term::StrChunks(chunks) if chunks.iter().all(is_literal) => {
+            ApparentType::Exact(Types(AbsType::Str()))
+        }
+        Term::List(_) => ApparentType::Exact(Types(AbsType::List(Box::new(Types(AbsType::Dyn()))))),
```
aspiwack, line 823: It's a little bit weird to call this case `Exact` isn't it?

src/typecheck.rs
yannham: My rule of thumb was that it shouldn't infer a precise (`Exact`) type for an expression that wouldn't pass the typechecker, that is if `apparent_type(e) = exact(T)` then `e : T` must be well-typed. But `"#{1 + 1} : Str` is not well-typed.

That said, the semantics of interpolated string enforce that such expressions either fails with a type error or evaluate to a string at run-time indeed. So let's return `Str` in all cases then.

src/typecheck.rs
yannham: Yeah, it was in opposition to `Approximated`. Maybe `Precise`, or `Inferred`?

aspiwack (approved):

src/typecheck.rs
aspiwack: Note that this rule already didn't apply to the `Assume` case. The only thing that you need of a type, in this gradual type context, is that if the expression doesn't blame, then it has this type (with a fairly strong notion of “doesn't blame”).

PR #274, opened 2021-01-23
Migrate to CI to Nix and GitHub Actions

garbas:


garbas:
# TODO

- [ ] remote builder or cachix
- [ ] cross compile or build on osx/windows
- [ ] commit with @yannham email on rust channel update
- [ ] docker image (also push it to the docker hub)

yannham:
@garbas what needs to be done for the docker image? If you have some pointer I can try to do this. I think we should postpone osx/windows cross compilation for now and create a dedicated issue to track this as it seems not clear how to do it, so that it doesn't hold back this PR.

garbas:
@yannham :+1: I will create a ticket for cross/windows build.

I've added a docker derivation for nickel. I asked for credentials to publish docker image. We can actually merge this already and I will open a PR separately that will add a GH action that will push docker image to hub.docker.com.

Also don't forget to remove circleci integration.

mboes:
Agreed @garbas - best to make each PR have as small a scope as you can get away with.

Issue #240, opened 2020-12-04
Default should be the default


aspiwack:
**Is your feature request related to a problem? Please describe.**
I've been thinking about overridability in the recent weeks. On and off, really. Nothing really concrete. Anyway, there's been a problem that's been bothering me for a while.

The way we plan on handling overridability is: we have a notion of `default` value, and if someone comes with a definite value, it's going to be used instead. That's neat, this is how Cue works as well, this squares well with the documentation as well, which will print: “default: `false`”.

But the problem is that we need to design fields for being overridden. By which I mean that any field which is not explicitly marked with `default` will consider its value final. That's bad! Because really, a field which is not overridable is considered a misfeature (this has been discussed in #218 , I believe). It shouldn't be a bad practice to use the default!

**Describe the solution you'd like**

Therefore, it is my current belief that, defining the value of a field should mean `default`, and definite/final values should be explicitly specified as such (in fact, there should probably be no definite value, and we probably want to specify a priority one way or another, but it's a somewhat orthogonal concern).

yannham:
> But the problem is that we need to design fields for being overridden. By which I mean that any field which is not explicitly marked with `default` will consider its value final. That's bad! Because really, a field which is not overridable is considered a misfeature (this has been discussed in #218 , I believe). It shouldn't be a bad practice to use the default!

Yes, I think we more or less all agree that overridable should be the default. Actually I've never been entirely convinced that just overriding default values would be sufficient, for the reason you cite, and just think of "normal/default" as a very simplistic binary priority system with two values. "normal" happens to be final, but just because it's top, and we could imagine add a "force" over it for example.

I feel like the question is: what to do when two non-mergeable values have the same priority, or how to ergonomically make this case avoidable in practice ? I see two approach:
1. unbounded priorities: say, integers priorities, so that you can always set a priority higher or lower than an existing one.
2. overriding operation: add a `//` operator, as in Nix. Internally this would just be a right-biased `merge`, otherwise working in exactly the same way.

What I like about the second approach is that:
- (a) this clearly distinguishes non commutative overriding from merging. Merge is already quite overloaded (applying contracts, merging records recursively, setting default value), so keeping a "clean", commutative merge that can only override default values from schemas (keeping this simple two-value priority system), versus a monkey-patching one that can override any value of actual configurations, is, I think, not a bad idea. Even if in the end there's no real language-level distinction between an "actual configurations" and a "contract" (as record with metavalues), I imagine there is one in the user's head.
[...]

PR #275, opened 2021-01-25
What is r.b.c?

blaggacao:
I'm asking this out of discussion here: https://github.com/NixOS/rfcs/pull/42#discussion_r563193452 It has more context. Beware, that my transpilation I/O is lossy.

/cc @roberth Is this a correct interpretation of the issue you raised over there?

### DIFF
--- a/src/examples/merge3.ncl
+++ b/src/examples/merge3.ncl
@@ -1,4 +1,4 @@
 let rNest1 = {b={c=10}} in
-let rNest2 = (fun x => {a=x; b={c=x}}) 10 in
+let rNest2 = (fun x => {a=x; b={d=x}}) 20 in
 let r = rNest1 & rNest2 in
 r.b.c


roberth:
I can't really tell what the test is trying to achieve because I couldn't find any documentation for what requirements it tries to test.

The issue I raised in the mentioned comment relates to an 'unintuitive' interaction between option priorities and option nesting in the NixOS module system. I don't see a use of defaults in this example.

Regardless, I'd recommend to document which requirements are being checked in a test case.
That's assuming this example even serves as a test case.

blaggacao:
Wouldn't the requirement be: `r.b.c` is not undefined?

I understand the concept of defaults is orthogonal, since it is only relevant with regard of the ([potentially unbounded](https://github.com/tweag/nickel/issues/240#issuecomment-738752158)) priority of competing values. If I read the context correctly, what seems to matter is that a merge **does** descend down into the tree and performs an in-depth first merge.

Or maybe I's falling short and the question is:

```
let rNest1 = {b=Default({c=10})} in
let rNest2 = (fun x => {a=x; b={d=x}}) 20 in
let r = rNest1 & rNest2 in
r.b.c # what is r.b.c?
```

yannham:
The sources in `examples/xxx.ncl` are examples, not tests, and they are terribly outdated (they need to be updated soon). Note also that since #266, the examples can be rewritten as:
```
let rNest1 = {
    b | default = {c = 10}
} in
...
```

Indeed, `r.b.c` is undefined, because `default` works the same as NixOS `lib.mkDefault`. One needs to write `{b = {c | default = 10}}` to have it defined in the final result.

If I understand correctly, the issue is when you want to specify a set of "leaf" defaults  (that is, of the latter form `{b = { c | default = ... }}` but without knowing the content  `{c = 10}` of the record beforehand, and without having to trust the user providing the record to put the required `default` attributes everywhere? In that case, I guess you would need either `default` or another specific attribute (say, `recDefault`) to be automatically pushed downward the leafs during evaluation.

However it's not clear to me if it's a general issue that should be solved in the language, or just a consequence of some design choices on the NixOS module system way of defining default.

In any case, this should be an issue rather than a PR. Once the subject is clear, let's move it there.

blaggacao:
I was acting on behalf of the messanger not the interested party role with a faint 20% understanding of the issue. Hence, I figured fiddling with an example via draft PR would be the best paper on which to IO the message.

I think you understood the use case better than me, already, which is pretty rewarding for _just a messager_. 😄 

> However it's not clear to me if it's a general issue that should be solved in the language, or just a consequence of some design choices on the NixOS module system way of defining default.

10000ft flight level: a "leafy" default would constitute no more than defaulting semantic sugar for a common data structure, that is a _nested object_.

It might be a terciary concern, though, for a language feature. I'm closing as a means to give room to a proper issue, should that call be made.

roberth:
I must have been confused about who the interested party is. I was trying to support @blaggacao.

@yannham you seem to have a good understanding of the interaction between defaults and nesting.

@ me if anyone has questions for me.

blaggacao:
> I must have been confused about who the interested party is. 

I think the interested party is posterity on their behalf we serve. 😉

blaggacao:
Second guessing myself: shall I formulate that issue? I'd be happy to give it a try (disclaimer: potentially lossy IO ahead :wink:).

yannham:
If you feel like it, please do! Thanks for contributing.

PR #276, opened 2021-01-25
Thunks black-holing

### LINKED ISSUE #278: Detect (direct) infinite recursion
Since records are recursive by default, the introduction of accidental infinite recursion is not rare. Most of these simple errors can be detected without actually looping by marking thunks currently under evaluation with a special flag, such that if such a thunk is accessed again while still under evaluation, an error is raised.

Example of current behavior:
```
$nickel <<< '{x = y; y = x}.x'
[...never returns...]
```

Direct infinite recursion, that is the one introduced by laziness (and not by actual mutually recursive functions), should be detected and reported:

```
$nickel <<< '{x = y; y = x}.x'
error: infinite recursion
  ┌─ <stdin>:1:13
  │
1 │ {x = y; y = x}.x
  │             ^ recursive reference
```

yannham:
Close #278. Flag thunks to detect infinite recursion. Doing so, refactor the definitions a bit to make thunk implementation leak less in other modules.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -68,6 +68,8 @@ pub enum EvalError {
     ),
     /// An unbound identifier was referenced.
     UnboundIdentifier(Ident, Option<RawSpan>),
+    /// A thunk was entered during its own update.
+    InfiniteRecursion(CallStack, Option<RawSpan>),
     /// An unexpected internal error.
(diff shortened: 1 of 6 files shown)


edolstra (approved):
@yannham Could you fix the merge conflict?

Issue #254, opened 2020-12-15
Next batch of functions for the standard library


yannham:
The current standard library is very minimalist. Let's add a new batch of functions and modules. This scope of this issue is to gather proposals, discussions and track progress on implementation. The idea is not to have a perfect, final version of the stdlib but rather to provide all the basic functionalities required for a first release of Nickel, without crippling holes in it.

Functions are grouped by modules, which may not exist yet. ~~List functions are typed with the current monomorphic list type (with `Dyn` pretty much everywhere), but will obviously be patched after #171.~~ [Edit: now merged]

## Builtins
- [x] `id : forall a. a -> a`
- [x] `hash : Dyn -> Str -> Str` (like Nix `builtins.hash`)
- [x] ~`toJSON : Dyn -> Str`~ `serialize : <Json, Yaml, Toml> -> Dyn -> Str`, or alternatively `#Serializable -> Str` with an appropriate `Serializable` contract
- [x] ~`fromJSON : Str -> Dyn`~ `deserialize : <Json, Yaml, Toml> -> Str -> Dyn`
- [x] `typeOf : Dyn -> Str`

## Contracts
- [x] `#Int`
- [x] `#Nat`
- [x] `#NonZero`
- [x] `#NonZeroNat`
- [x] `#NonEmptyStr`
- [x] `#NonEmptyList`

## Numbers
- [x] `max/min: Num -> Num -> Num`
- [x] `floor: Num -> Num`
- [x] `isInt: Num -> Bool`

## Lists
- [x] `elem : forall a. a -> List a -> Bool` (is member of)
- [x] `partition : forall a. (a -> Bool) -> List a -> {in: List a, out: List a}` (open to other field names)
- [x] `sort : forall a. (a -> a -> Bool) -> List a -> List a`
- [x] `generate : forall a. (Num -> a) -> Num -> List a` or `forall a. (#Nat -> a) -> #Nat -> List a`.  Generate a list from a generator and a length. May be deprecated by list comprehensions.

## Records
- [x] `valuesOf : { | Dyn} -> List`

## Strings
- [x] `join : List Str -> Str -> Str`
- [x] `split : Str -> Str -> List Str`
- [x] `trim: Str -> Str`
- [x] `uppercase: Str -> Str`
- [x] `lowercase: Str -> Str`
- [x] `contains: Str -> Str -> Bool`
- [x] `replace : Str -> Str -> Str -> Str`
[...]

PR #284, opened 2021-02-02
Add sort,generate and partition to lists stdlib

yannham:
Partly address #254. Depend on #317. Tests and docs are saved for later, see #321.


stdlib/lists.ncl
```diff
@@ -46,5 +46,33 @@
     any : forall a. (a -> Bool) -> List a -> Bool =
       fun pred l =>
         fold (fun x acc => if pred x then true else acc) l false;
+
+    elem : Dyn -> List -> Bool =
```
regnat, line 50: Why not `forall a. a -> List a -> Bool` ?

stdlib/lists.ncl
```diff
@@ -46,5 +46,33 @@
...
+        foldl aux {right = []; wrong = []} l;
+
+    generate : forall a. (Num -> a) -> Num -> List a =
+      fun gen n =>
+        if n <= 0 then
+          []
+        else
+          generate gen (n-1) @ [gen n];
+
+    sort : forall a. (a -> a -> Bool) -> List a -> List a =
```
regnat, line 69: `a -> a -> Bool` isn't really cool as a type, because there's no way to infer its semantics just by looking at it :/

Afair the language has no support for tagged unions or symbols to make this nicer, but maybe wrapping it into a record like `{ isLower : a -> a -> Bool }` could help (at the cost of being more verbose)

stdlib/lists.ncl
yannham: Good question. This is the type I initially wrote, but actually the equality operator, such as other builtins like `isNum`, doesnn't respect parametricity because it does inspect their arguments. In fact, the implementation clearly depends on the type of each argument. So if you write the proposed polymorphic type, you break the `forall` contract when trying to evaluate the parameters `x` and `elt` in `x == elt`.

For example, in OCaml, the equality does have the type `forall a. a -> a`, but it's a compiler hack, which is regularly called out [[1]](https://blog.janestreet.com/the-perils-of-polymorphic-compare/). On the other hand, in Haskell, the type doesn't lie, since it is constrained by `Eq`, and is not fully polymorphic.

It's even worse currently in Nickel, since such polymorphic values are protected by the contract but don't raise blame when evaluated (they are just special constructs `Wrapped(key, stuff)` preventing further evaluation). Using the proposed polymorphic type, you end up with wrong results without errors, such as :
```
nickel> lists.elem 1 [1,2]
false
```
because everything is wrapped equality doesn't try to inspect wrapped values.

One solution is to cheat and let equality inspect wrapped values, but it feels ad-hoc. Why equality and not `isNum` for example? If every builtin operators can inspect polymorphic values, parametricity doesn't make any sense anymore. I don't see an easy way around it, unless we go through the subtyping route (#244) in which case we could keep the actual types with `Dyn` but without suffering casts in statically typed code, or we can abandon enforcing parametricity in function contracts.

In the end, I think that this is just telling us that the actual type of equality is `Dyn -> Dyn`. It's just typechecked differently (as `_a' -> _a'`) for convenience inside statically typed code, because otherwise you would have to make a type cast to `Dyn` each time you use it in statically typed code, but this is a pragmatic lie.

stdlib/lists.ncl
yannham: Well, we do have simple enums, so what about something like `a -> a -> <StrictlyLesser, GreaterOrEq>`? Also, there will be documentation at some point.

regnat: Oooh, then yes, that looks much better imho. Or maybe `<StriclyLesser, Equal, StrictlyGreater>` to be more symmetrical − I guess we'll want a standard definition of an ordering function anyways, so we might as well define it right now

stdlib/lists.ncl
regnat: Oh, I see, thanks.

I guess it would theoretically make sense to pass the equality function as a first argument (`elem : forall a. (a -> a -> bool) -> List a -> Bool`), but in that case calling `list.any` directly is as simple

yannham: > I guess it would theoretically make sense to pass the equality function as a first argument (`elem : forall a. (a -> a -> bool) -> List a -> Bool`), but in that case calling `list.any` directly is as simple

Yes, exactly.

stdlib/lists.ncl
yannham: I unshamedly copied Rust's [`Ordering`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) enum.

curiousleo: There is nothing to be ashamed of when copying Rust.

regnat (approved):

PR #287, opened 2021-02-03
[Fix]Equality operator mishandling enums

### LINKED ISSUE #286: Equality always returns false on enum
**Describe the bug**
Comparing an enum `` `Something`` with anything else, including itself, always return false.

**To Reproduce**
```
$nickel <<< '`Test == `Test`
Bool(false)
```

**Expected behavior**
Comparing two enums with the same tag should return true.

yannham:
Fix #286.


aspiwack (approved):

aspiwack:
I think that for something this small, if you intend to solve it immediately, you can simply write the PR directly, explaining the bug in the PR description.

PR #290, opened 2021-02-08
Fix arguments to the on attribute in GH actions

yannham:
Github was unhappy with some of the GH actions defined in #274. It seems the `on` attribute must have events as children, but `branches:` was used directly.


.github/workflows/specs.yml
```diff
@@ -2,9 +2,8 @@ name: "Test specs"
 
 on:
   pull_request:
-  branches:
-    - master
-
```
garbas, line 7: I think here is "push:" missing

yannham: Branches is now attached to `pull_request` so it should be fine. However we probably want these tests to also run on dependent PR (that are not necessarily based on `master`), so I imagine the thing to do is to remove the branches filter entirely.

PR #291, opened 2021-02-11
[Fix]Avoid panic when handling serde_json errors with no location

yannham:
Nickel converts serde's (de)serialization errors to its own error reporting infrastructure. It uses in particular the `serde_json::Error::line()` method, that should always be greater or equals to 1 according to the documentation, but sometime it actually doesn't, which caused an integer underflow panic when subtracting `1` to the line number.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -345,18 +345,25 @@ impl ParseError {
     ) -> Self {
         use codespan::ByteOffset;
 
-        let line_span = files.line_span(file_id, (error.line() - 1) as u32);
+        // error.line() should start at `1` according to the documentation, but in practice, it may
+        // be 0 for the error `json parse error: data did not match any variant of untagged enum
+        // Term`. Although this error should not happen, if it does, it's better to get a message
+        // than a panic message `subtract with overflow`.
+        let line_span = if error.line() == 0 {
+            None
+        } else {
+            files.line_span(file_id, (error.line() - 1) as u32).ok()
+        };
+
         let start = line_span.map(|ls| ls.start() + ByteOffset::from(error.column() as i64 - 1));
         ParseError::ExternalFormatError(
             String::from("json"),
             error.to_string(),
-            start
-                .map(|start| RawSpan {
-                    src_id: file_id,
-                    start,
-                    end: start + ByteOffset::from(1),
-                })
-                .ok(),
+            start.map(|start| RawSpan {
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -345,18 +345,25 @@ impl ParseError {
...
-        let line_span = files.line_span(file_id, (error.line() - 1) as u32);
+        // error.line() should start at `1` according to the documentation, but in practice, it may
+        // be 0 for the error `json parse error: data did not match any variant of untagged enum
+        // Term`. Although this error should not happen, if it does, it's better to get a message
+        // than a panic message `subtract with overflow`.
+        let line_span = if error.line() == 0 {
+            None
+        } else {
+            files.line_span(file_id, (error.line() - 1) as u32).ok()
+        };
```
regnat, line 356: Can't we just display the same span, but with a line of 0 if `error.line()` is zero? Or does the `file_id` doesn't make sense either in that case?

```suggestion
        let line_span files.line_span(file_id, (if error.line() > 0 {error.line() - 1} else {0}) as u32).ok();
```

(can probably be rewritten more concisely as `min (error.line(), 1) - 1`, but I'm not sure where to pull the `min` function from)

yannham: Lines are numbered from `1` to `n` in `serde`, while they are numbered from `0` to `n-1` in `codespan` (the `files` stuff). So when `serde` return line `0`, this is actually an invalid value (IMHO the API should have been `Option<NonZeroU64>` instead of `u64` to make this clear) corresponding to a error that is not located.

It sounds a bit strange to me to coerce this invalid span to a valid line number in `files`, causing the error to point at the first line (and, I imagine, first character) of the JSON/YAML/TOML file. Or maybe we can make the span to be the entire file, but I don't think this adds useful information, because the Nickel error message includes the original import statement nonetheless, which points at the problematic file anyway.

Also, I think this error won't show up again once we have `null` (#292): although undocumented, it seems that `serde` returns this invalid line error only when it finds a JSON value that it can't decide how to map to a Nickel value, which should currently only happens for `null`. This PR is "just in case", to avoid unnecessary (although unlikely) panics.

regnat: > the Nickel error message includes the original import statement nonetheless, which points at the problematic file anyway.

Ah that removes my only reason for the change. It's all good for me then

regnat (approved):

PR #292, opened 2021-02-11
Add the null value

yannham:
Some time ago, we agreed upon having `null` in the language, as we aim for easy interoperability with JSON. With the recent addition of multiple export and import formats, the time has come, especially since #282: if one tries to import a JSON or a YAML file with null values, this currently raises an error.

This PR adds the null value to the language. `null` is typechecked as `Dyn`. Serialization prevalidation ensures that formats that don't support null (such as TOML, for example) raise an appropriate error at export.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,7 @@ use crate::label::ty_path;
 use crate::parser::lexer::LexicalError;
 use crate::parser::utils::mk_span;
 use crate::position::RawSpan;
+use crate::serialize::ExportFormat;
 use crate::term::RichTerm;
 use crate::types::Types;
 use crate::{label, repl};
@@ -232,6 +233,10 @@ pub enum ImportError {
 /// An error occurred during serialization.
 #[derive(Debug, PartialEq, Clone)]
 pub enum SerializationError {
+    /// Encountered a null value for a format that doesn't support them.
+    UnsupportedNull(ExportFormat, RichTerm),
+    /// Tried exporting something else than a `Str` to raw format.
+    NotAString(RichTerm),
     /// A term contains constructs that cannot be serialized.
     NonSerializable(RichTerm),
     Other(String),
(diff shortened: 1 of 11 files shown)


edolstra:
I'm getting a compilation error if I try to merge this locally:
```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
    --> src/operation.rs:1398:17
     |
1398 |                 serialize::validate(&rt2)?;
     |                 ^^^^^^^^^^^^^^^^^^^ ---- supplied 1 argument
     |                 |
     |                 expected 2 arguments
     |
note: function defined here
    --> src/serialize.rs:70:8
     |
70   | pub fn validate(t: &RichTerm, format: ExportFormat) -> Result<(), SerializationError> {
     |        ^^^^^^^^ ------------  --------------------
```

BTW maybe we should configure the github action to try to test the PR against master if it's not already doing that.

yannham:
I rebased and fixed the compilation error.

> BTW maybe we should configure the github action to try to test the PR against master if it's not already doing that.

Indeed

yannham:
> BTW maybe we should configure the github action to try to test the PR against master if it's not already doing that.

After looking at this, GH actions should have done that. We have a github action on `pull_request` event, which includes by default `opening` and `synchronizing` (ie, pushing to the branch) events. According to [this discussion](https://github.com/actions/checkout/issues/15#issuecomment-524093065), the default behavior on `pull_request` is to checkout the merge commit. I don't understand why the build/test action doesn't even run on this PR. Bad timing between #274 and disabling the old CI?

regnat:
> I don't understand why the build/test action doesn't even run on this PR

There was an issue with GH actions this morning which apparently prevented some actions from being triggered (https://www.githubstatus.com/incidents/xn0sd2x4nd7f). Quite possibly that's the reason here and just re-pushing anything would resolve it

Issue #293, opened 2021-02-15
Better location information when a reporting a blame error


yannham:
**Is your feature request related to a problem? Please describe.**
Currently, if a contract enforced via merging is broken, the error message only points at the definition of the contract, but doesn't give much information about the location of the offending value:

```
let SomeContract = {
 foo | Num;
} in
...
// a lot of code
...
{ foo = "a" } & SomeContract
error: Blame error: contract broken by a value.
  ┌─ :1:1
  │
1 │ Num
  │ --- expected type
  │
  ┌─ repl-input-0:2:9
  │
2 │   foo | Num;
  │         ^^^ bound here
```

**Describe the solution you'd like**
The error message should provide information about the value that has failed the contract (here, `"a"`) and/or the location where the contracts have been enforced, instead of just defined (here, could be the location of the merging).

PR #300, opened 2021-02-23
Improve position information

yannham:
Preliminary PR for #293. Currently, terms either have a position if they come from source, or `None` when they are the result of evaluation. However, keeping the position of the parent expression may be better for error reporting, as usually the more position information is available, the better.

A motivating example is contract evaluation, when we would like to report both the original expression in the source as well as the evaluated expression, when a contract is broken. For example, in `let x = "a" ++ "b" in ... { stuff | Num = x }`, we would like to point to both locations of `(stuff | Num = x)` and `"a" ++ "b"`, as well as precising that the whole argument evaluated to `"ab"`. 

This PR changes the type of position of terms from `Option<RawSpan>` to a ternary variant type `Original(RawSpan), Inherited(RawSpan), None` where `Original` corresponds to an expression that is coming directly from a source (or has been transformed from a source expression), `Inherited` means that the expression is derived (evaluated) from a source expression.

This PR just introduces the new type and do the required refactoring. Correctly setting and using the position of derived expressions is left for subsequent PRs.


edolstra:
@yannham This has a merge conflict.

PR #305, opened 2021-02-25
Parsing: add missing != operator, be more liberal with list elements

yannham:
This PR fixes two small quirks in the parser:

1. Add the missing inequality operator (`!=`).
2. List elements were syntactically restricted, requiring parentheses around non atomic expressions: `[(1 + 1), (x : Num), (someFunc "someArg")]`. Now,  any kind of expression is accepted as an element: `[1 + 1, let x : Num = 1 in x + 2, "http://www.google.com/" | #Url]`

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -60,8 +60,7 @@ MetaAnnot: MetaValue =
     };
 
 LeftOp<Op, Current, Previous>: RichTerm =
-    <t1: Current> <op: Op> <t2: Previous> => mk_term::op2(op, t1,
-    t2);
+    <t1: Current> <op: Op> <t2: Previous> => mk_term::op2(op, t1, t2);
 
 LeftOpLazy<Op, Current, Previous>: RichTerm =
     <t1: Current> <op: Op> <t2: Previous> => mk_app!(Term::Op1(op, t1), t2);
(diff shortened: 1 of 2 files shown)


src/grammar.lalrpop
```diff
@@ -484,6 +482,15 @@ BinOp8: BinaryOp = {
 InfixExpr8: RichTerm = {
     InfixExpr7,
     LeftOp<BinOp8, InfixExpr8, InfixExpr7> => <>,
+    <l: @L> <t1: InfixExpr8> "!=" <t2: InfixExpr7> <r: @R> => {
```
regnat, line 485: Any reason why the `!=` is special-cased (compared to `=`

regnat, line 485: Any reason why the != is special-cased (compared to `=` for example)?

yannham: `!=` isn't a primitive operator per se, but is parsed as `! (t1 == t2)`, which doesn't fit in the `LeftOp`/`InfixExpr` rules that rely on subrules like `BinOp8` to return a binary operator, and not an arbitrary expression.

The position (`@L` and `@R`) is needed if we want to pass it to the generated subexpression `t1 == t2`, which is set to have the same position as the original one (as well as the whole `! (t1 == t2)`). Usually, the more position information, the better.

However, thinking about this specific case a bit more, I don't think this position won't ever be useful: any typechecking error or evaluation error should be either located in a subterm `t1` or `t2`, or be about the whole expression `!(t1 == t2)`. Same for evaluation, if the whole expression is evaluated, `t1 == t2` will be evaluated away and the final position will be inherited from the enclosing `!`. 

The sub-expression `t1 == t2` is probably invisible for all intent of purpose, so maybe I should get rid of the `<l : @L>` and `<r: @R>`.

Issue #161, opened 2020-10-08
Proper testing infrastructure


yannham:
As the project codebase grows, more and more tests are accumulated. They are currently all mixed (unit, integration, regression, etc.), with sometimes hardcoded Nickel code, sometimes hand builded ASTs, and what not. This is not always easy to see what is actually covered and where. This is not the current priority until the MVP, but that could be nice to have a clean test suite and to detach generic language tests into separate nickel source files at some point.

yannham:
Closing after #306 and #313. Tests for the stdlib are tracked in #321. There is of course a lot of room for improvement (property-based tests, generating lambda terms and testing against an reference implementation like Haskell/Dhall/Jsonnet, and so on), but the basic migrations have been carried out, and keeping this issue open would make its scope probably too broad and blurry.

PR #306, opened 2021-02-25
Migrate passing tests to standalone nickel files

yannham:
Depend on #305. Partly address #161.

Migrate all tests consisting in evaluating a standalone Nickel expression and expecting it to succeed to separate `.ncl` files, now run by an integration test:
- As shown by the diff (net -456 loc), this removes a lot of boilerplate related to running Nickel code from inside Rust.
- This makes tests easier to add, modify or remove, without having to touch Rust code.
- Test files are directly runnable from nickel, e.g.: `$nickel -f tests/pass/eq.ncl`. Thanks to the use of contracts for checking assertions, this gives good error messages and inspection capabilities in case of a failing test.

**What's next**
- refactor and migrate integration tests expecting failures


tests/pass.rs
```diff
@@ -0,0 +1,46 @@
...
+    d.push("tests/pass");
+
+    let children: Vec<_> = read_dir(d)
+        .unwrap()
+        .into_iter()
+        .map(|file| {
+            let file = file.unwrap();
+            let path = file.path();
+            let name = String::from(path.file_name().map(OsStr::to_string_lossy).unwrap_or(path.to_string_lossy()));
+            thread::Builder::new()
```
regnat, line 39: Does that mean that all the tests will run in parallel or is there some kind of implicit limit?

regnat: Unless this is too tricky to do, I think it would be much nicer to have one test per input file, so that the test runner can handle the parallelism and all the resource management/error reporting itself

regnat (changes requested):
LGTM, except for the fact that all the tests run as a big one which could quickly become an issue.

(I also didn't check that the new tests were semantically the same as before, but…)

tests/pass.rs
yannham: > Does that mean that all the tests will run in parallel or is there some kind of implicit limit?
Yes, It should spawn one thread per input file

> Unless this is too tricky to do, I think it would be much nicer to have one test per input file, so that the test runner can handle the parallelism and all the resource management/error reporting itself

Since we need to spawn a thread manually to use a bigger stack size anyway, I went with doing everything in one function.  But I think you're right, it's cleaner with one function per test. It just means that we have to write a new function when we add a new test file, but this sounds like a reasonable expectation.

regnat (approved):
LGTM.

I guess we'll probably want a macro at some point to generate all the boilerplate for declaring the individual tests, but let's not over-think it for now

edolstra (approved):

PR #319, opened 2021-03-09
Add support for n-ary operators

yannham:
While addressing #254, it turned out some builtin string operations require corresponding primitive operations with more than 2 arguments. This PR keeps the existing unary and binary operator nodes, as they make up for the majority of primitive operations, but add a generic n-ary operator node for arity greater than 2.


edolstra (approved):
There's a merge conflict, otherwise LGTM.

PR #325, opened 2021-03-11
[Fix]Primitive operators typing errors

### LINKED ISSUE #320: Inexact error messages when mistyping builtin operators
**Describe the bug**

When statically typechecking expressions containing builtin operations, such as addition, some error messages report "function types mismatch" on some arrow types that seem to come from nowhere. While still pointing at the problem more or less reasonably, this is confusing.

**To Reproduce**
```
let x : Str = 2 + 1
error: Function types mismatch
  ┌─ :1:15
  │
1 │ Num -> Num -> Str
  │               --- This part of the expected type
  │
  ┌─ :1:15
  │
1 │ Num -> Num -> Num
  │               --- does not match this part of the inferred type
  │
  ┌─ repl-input-2:1:15
  │
1 │ let x : Str = 2 + 1
  │               ^^^^^ this expression
  │
  = The type of the expression was expected to be `Num -> Num -> Str`
  = The type of the expression was inferred to be `Num -> Num -> Num`
  = Could not match the two function types
```

**Expected behavior**
The error message should be a standard "Type mismatch" between expected Str and actual Num, and shouldn't involve any arrow type.

yannham:
Fix #320. Depend on #319. The type of primitive operators was internally passed as an arrow `TypeArg -> TypeArg -> TypeResult`, causing unification errors on the return type to mysteriously report about a function type. It is now passed as tuples, and directly unified with the corresponding components. This also gets rid of a few unification steps and unification variable creations.

### DIFF
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -1772,10 +1772,10 @@ pub fn get_bop_type(
             mk_typewrapper::str(),
             mk_typewrapper::dynamic(),
         ),
-        BinaryOp::Pow() => mk_tyw_arrow!(
+        BinaryOp::Pow() => (
             mk_typewrapper::num(),
             mk_typewrapper::num(),
-            mk_typewrapper::num()
+            mk_typewrapper::num(),
         ),
     })
 }
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -652,26 +652,19 @@ fn type_check_(
             }
         }
         Term::Op1(op, t) => {
-            let ty_op = get_uop_type(state, op)?;
+            let (ty_arg, ty_res) = get_uop_type(state, op)?;
 
-            let src = TypeWrapper::Ptr(new_var(state.table));
-            let arr = mk_tyw_arrow!(src.clone(), ty);
-
-            unify(state, strict, arr, ty_op)
+            unify(state, strict, ty, ty_res)
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))?;
-            type_check_(state, envs.clone(), strict, t, src)
+            type_check_(state, envs.clone(), strict, t, ty_arg)
         }
-        Term::Op2(op, e, t) => {
-            let ty_op = get_bop_type(state, op)?;
+        Term::Op2(op, t1, t2) => {
+            let (ty_arg1, ty_arg2, ty_res) = get_bop_type(state, op)?;
 
-            let src1 = TypeWrapper::Ptr(new_var(state.table));
-            let src2 = TypeWrapper::Ptr(new_var(state.table));
-            let arr = mk_tyw_arrow!(src1.clone(), src2.clone(), ty);
-
-            unify(state, strict, arr, ty_op)
+            unify(state, strict, ty, ty_res)
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))?;
-            type_check_(state, envs.clone(), strict, e, src1)?;
-            type_check_(state, envs, strict, t, src2)
+            type_check_(state, envs.clone(), strict, t1, ty_arg1)?;
+            type_check_(state, envs, strict, t2, ty_arg2)
         }
         Term::OpN(op, args) => {
             let (tys_op, ty_ret) = get_nop_type(state, op)?;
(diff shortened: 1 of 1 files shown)


src/typecheck.rs
```diff
@@ -1499,17 +1492,18 @@ fn instantiate_foralls(state: &mut State, mut ty: TypeWrapper, inst: ForallInst)
...
             let branches = TypeWrapper::Ptr(new_var(state.table));
 
-            mk_tyw_arrow!(
-                AbsType::Bool(),
-                branches.clone(),
-                branches.clone(),
-                branches
+            (
+                mk_typewrapper::bool(),
+                mk_tyw_arrow!(branches.clone(), branches.clone(), branches),
```
regnat, line 1506: I'm a bit curious: I was expecting that given that you still return an arrow type here, you could still exhibit the issue given a well-crafted `if-then-else` statement (something like `let x : Num = if true then 1 else "foo"`), but I couldn't get the typechecker to issue a bad error message. Is there a reason for that?

regnat (approved):

src/typecheck.rs
yannham: You're right that in general, this kind of operator - the ones that are "lying" about their arity because they accept additional arguments lazily, and thus needs to return an arrow type - should still exhibit the issue. Here, you tried `(ifthenelse true) 1 "foo"` where `(ifthenelse true) : a -> a -> a`.

As long as the type mismatch is in the argument, the error should be local, because of the way of typechecking function application work: we first first determine the type of the function, then try to match the argument against the expected type.

However, having a wrong outer annotation while the if-then-else itself is well-typed should trigger the problem:

```
nickel> (if true then 0 else 0) : Str
error: Incompatible types
  ┌─ repl-input-4:1:15
  │
1 │ (if true then 0 else 0) : Str
  │               ^ this expression
  │
  = The type of the expression was expected to be `Str`
  = The type of the expression was inferred to be `Num`
```

Hmm, it doesn't...After more thought, this is due to the order in which the unification is done, which was a bit different in the removed code versus the code for standard application (which it shouldn't have been, really). Because of row types (see comment https://github.com/tweag/nickel/blob/9a594f460008c467d4371259ce2999621b847720/src/typecheck.rs#L564), the left hand side of an application is checked first before the argument. When typechecking `(ifthenelse true) 0 0 : Str`:
1. On the first application, it first typechecks the left hand side: generates a type `a -> Str`, typecheck `(ifthenelse true) 0` against it
2. Again, generate a type `b -> a -> Str`, typecheck `(ifthenelse true)` against it
3. By `get_uop_type`, `ifthenelse true` has type `c -> c -> c`. Unifying the two sets the type to `Str -> Str -> Str`.
4. Then, typing the argument of the inner application raise an error: `Num != Str`.

On the other hand, if the type is not polymorphic, you do get an arrow type error, because at step `3` you try to unify the whole `Num -> Num -> Num` with `b -> a -> Str`, which fails:

```
nickel> let ite : Bool -> Num -> Num -> Num = fun x y z => z  in ((ite true 0 0) : Str) 
error: Function types mismatch
  ┌─ :1:23
  │
1 │ Bool -> Num -> Num -> Str
  │                       --- This part of the expected type
  │
  ┌─ :1:23
  │
1 │ Bool -> Num -> Num -> Num
  │                       --- does not match this part of the inferred type
  │
  ┌─ repl-input-3:1:60
  │
1 │ let ite : Bool -> Num -> Num -> Num = fun x y z => z  in ((ite true 0 0) : Str) 
  │                                                            ^^^ this expression
  │
  = The type of the expression was expected to be `Bool -> Num -> Num -> Str`
  = The type of the expression was inferred to be `Bool -> Num -> Num -> Num`
  = Could not match the two function types
```

These "Arrow type mismatch" are somehow an error reporting feature. Beforehand, both errors would be reported the same, but this was confusing in the case where you explicitly define `let ite : Bool -> Num -> Num -> Num`, because this got you an `expected Str, got Num` located on the `ite` (IIRC), so more context about the "top-level" unification that was actually going on was added. But at the end of the day, the error is caused by unification on some sub-component.

Finally, error reporting aside, it doesn't really make sense to return an arrow type for primitive operators, just to generate fresh variables to unify with the components of these arrows: in some way, we were constructing an arrow that we generically destructed immediately, all on the type system side, which sounds wasteful.

Issue #279, opened 2021-01-26
"Leafy" defaults


blaggacao:
**Is your feature request related to a problem? Please describe.**
In [some situations](https://github.com/NixOS/rfcs/pull/42#discussion_r563193452), a designer want's to forward-comply a user's input in a way so that the user can submit a (syntactically simple) nested set of defaults in the form:
```nix
{
  a = "a";
  b = {
    c = "c";
    d = "d";
  };
}
```

... where all leaf values are default values, hence ...

```nickel
{
  a | default = "a";
  b = { # but not here
    c | default = "c";
    d | default = "d";
  };
}
```

This provides good UX for cases where we can think of an arbitrary shaped user supplied defaults primer that the same (or a different) user would want to have the ability override in concrete applications.

**Describe the solution you'd like**

_It wouldn't go so as far to say that **I**'d like this solution. Rather, posterity might have a use for it._

A leafy default syntactic sugar idiom `leafyDefaults` or similar, so that the designer can declare:

```nickel
defaults | leafyDefaults = {};
```

**Describe alternatives you've considered**
none, I can't perceive the full breath of the use case.

&rarr; I'm not sure of the implications to do forward compliance, especially so with regard to default modelled as unbounded priority metadata.

**Additional context**
- https://github.com/NixOS/rfcs/pull/42#discussion_r563193452
- https://github.com/tweag/nickel/pull/275

Specifically, citing @yannham from https://github.com/tweag/nickel/pull/275#issuecomment-767402336
> However it's not clear to me if it's a general issue that should be solved in the language, or just a consequence of some design choices on the NixOS module system way of defining default.

but also plans for https://github.com/NixOS/rfcs/pull/42#discussion_r563321725
```nix
mkOption {
  recursiveDefault = {
    foo.bar = 10;
  };
}
```

Issue #255, opened 2020-12-23
Associative Array Merges


blaggacao:
A part from 
https://github.com/tweag/nickel/blob/9a3f5e7f4a892e28f3137e8490bad1650a7fb01b/src/merge.rs#L26

I couldn't find any reasoning about nickel's take on associative array merges or in general array merge strategies.

This is a very common use case for k8s yaml configurations, and being able to encode the merge strategy as metadata on an array data-structure seems important to me for a config management language with merge semantics.

Maybe I wasn't just able to infer nickel's stance on this topic?

```yaml
- name: "foo"
  thing: "bar"
```
merges with 
```yaml
- name: "foo"
  aspect: "qux"
```
to 

```yaml
- name: "foo"
  thing: "bar"
  aspect: "qux"
```
when associative array merge strategy is configured as:
`mergeKey: name`

blaggacao:
This is also a pre-requisite for semantic diffing.

yannham:
Merging is indeed defined in a non trivial way only for records, the idea being that this merge strategy is the expected natural way of merging dictionnaires, while other data types (such as lists) may have several possible strategies, none of which being THE natural and expected one.

As you point out, it is however often required to combine various data structures in specific ways. The currently favored direction is to add a metadata field to specify a custom merge strategy, giving something like this in practice:

```
let myMerge : Str -> List -> List -> List = fun key l1 l2 =>
 // ... implementation of merging by key..
 in

{
  listOfStuff
    | List
    | merge (myMerge "name")
     = [
      {name = "foo"; thing = "bar"},
      {name = "other"; stuff="stuff"}
     ]
} & {
  listOfStuff = [
    {name = "foo"; thing = "bar"; aspect = "qux"},
  ]
}
// RESULT:
{
  listOfStuff = [
    {name = "foo"; thing = "bar"; aspect = "qux"},
    {name = "other"; stuff="stuff"}
  ]
}
```
This is just an illustrative example, please don't take it too literally (`myMerge` would probably go in a library, the metadata would be better defined in a separate contract than in one of the concrete configuration being merged, we could use a more specific contract than `List` such as `List {name: Str | Dyn}`, etc.).

I'll let this issue open until a concrete proposal is detailed in a fresh issue.

yannham:
Superseded by #330 (and in particular custom merge functions).

PR #330, opened 2021-03-19
[RFC]Overriding

yannham:
This document contains a proposal for an overriding mechanism.

Related issues: #103, #240, #255, #279

### DIFF
--- a/notes/overriding.md
+++ b/notes/overriding.md
@@ -277,15 +277,15 @@ etc.) and their assignment must be made separately.
       type = types.listOf types.path;
     };
   }
-  
+
   # Config
   [..]
   someModule.paths = [foo/bar];
-  
+
   # Other config
   [..]
   someModule.paths = [/bar/baz];
-  
+
   # Resulting config: [foo/bar /bar/baz]
   ```
 
@@ -340,7 +340,7 @@ too limited for Nix use cases.
 #### Limits
 
 - ~~**(EXP)**~~: One can only replace a value by another one. It is not
-     possible to add one to the previous value, for example. 
+     possible to add one to the previous value, for example.
 - ~~**(COMP)**~~: Overriding a record a second time is not possible
 
 ### Jsonnet
@@ -622,7 +622,7 @@ Example:
 
   bar | Str,
   //equivalent to `bar | Str | priority 0`
-  
+
   baz.boo.bor | priority -4 = "value",
 
   final | force = `CantOverrideMe,
@@ -642,7 +642,7 @@ new meta-values `default rec` and `force rec`, whose semantics are defined as:
 
 - `eval(expr | default rec)`: case of `eval(expr)`:
   * `{field1 = value1, .., fieldn = valuen} | annots`: `{field1 = (value1 | default rec aux), .., fieldn =
-      (valuen | default rec aux)} | annots` 
+      (valuen | default rec aux)} | annots`
   * `v | annots` if `v` is not a record: `v | defaulted(annots)` where
       `defaulted(annots)` is defined below.
 
@@ -793,17 +793,17 @@ choices previously described.
 
 **THIS IS WORK IN PROGRESS FROM HERE**
 
-As explained in the [custom merge function section](custom-merge-function),
-supporting merge functions requires to re-interpret a merge expression a
-posteriori. Thus, in one way or another, we must keep track of the original AST
-of a merge expression if we want to implement this in all generality. Let us
-define the merge AST of an expression. A merge AST is a binary tree with nodes
-`Merge(ast1, meta1, ast2, meta2)`, that are labelled by meta-values `meta1` and
-`meta2`, and leafs `Exp(e)`, labeled with an expression `e`. In this context, a
-meta-value is a record which field names corresponds to meta attributes
-`custom,priority,default,Contract,etc.` and values are the corresponding value
-of the attribute. Field are not necessarily all defined, excepted `priority`,
-which must always be defined. In Rust syntax:
+As explained in the [custom merge function section](#custom-merge-functions-2),
+supporting merge functions requires to potentially re-interpret a merge
+expression a posteriori. Thus, in one way or another, we must keep track of the
+original AST of a merge expression if we want to implement this in all
+generality. Let us define the merge AST of an expression. A merge AST is a
+binary tree with nodes `Merge(ast1, meta1, ast2, meta2)`, that are labelled by
+meta-values `meta1` and `meta2`, and leafs `Exp(e)`, labeled with an expression
+`e`. In this context, a meta-value is a (meta-)record which field names
+corresponds to meta attributes `custom,priority,default,contract` etc. Field are
+not necessarily all defined, excepted `priority`, which must always be defined.
+In Rust syntax:
 
 ```rust
 struct MetaData {
@@ -818,19 +818,19 @@ enum AbsMergeAST<T> {
     Exp(Expression)
 }
 
-// This kind of recursive type is illegal in Rust, but this is not important
+// This kind of recursive type is illegal in Rust, but this is irrelevant
 type MergeAST = AbsMergeAST<(MergeAST, MetaData)>;
 ```
 
 We define `mergeAst: Expression -> MergeAST` in the following way:
 
 ```
-// we maintain an environment
+// we maintain an environment of bindings
 env
 
-// if priority is not set, we add priority = normal in the RHS
-metaData (e | attr = val, ...) ::= { attr = val, ... }                  if priority is set
-                                   { piority = normal, attr = val, ...} otherwise
+metaData (e | attr = val, ...) ::=
+  { attr = val, ... }                    if priority is set
+  { priority = normal, attr = val, ...}  otherwise
 
 mergeAst e ::= (mergeAstAux e, metaData e)
 
@@ -845,37 +845,45 @@ mergeAstAux (primop exp1 exp2 ... @ e) = Exp(e)
 mergeAstAux (let x = val in exp) = mergeAst(exp), env ::= env,x <- val
 mergeAstAux x = mergeAstAux (env(x))
 
-// Should merge AST cross import boundaries?
+// Should mergeAst cross import boundaries?
 mergeAstAux (import path) = ?
 ```
 
 One important problem, stated in the [custom merge function
-section](custom-merge-function) (although put differently), is that the merge
+section](#custom-merge-function) (although put differently), is that the merge
 AST is not stable by reduction. In particular, evaluating the content of a
-variable may rewrite a whole subtree to a leaf. Hence, what we care about is the
-initial merge tree of an expression, before any reduction happens.
+variable may rewrite a whole subtree to a leaf, forgetting the original
+information. Hence, what is important is the initial merge tree of an
+expression, before any reduction happens.
 
-Then, we need to extract a potential merge fonction from this AST:
+We then need to extract a potential merge fonction from this AST:
 
 ```
-mergeFunctions (Merge(ast1,_),Merge(ast2,_)) = mergeFunctions ast1 @ mergeFunctions ast2
+mergeFunction (Merge(ast1,_),Merge(ast2,_)) = mergeFunctions ast1 @ mergeFunctions ast2
 mergeFunctions (Leaf(e,meta)) = [f] if meta.merge == Some(f)
                                 []  otherwise
 ```
 
 ### Semantics
 
-Let us define the evaluation of merge AST parametrized by a
-merge function:
+We define the evaluation of merge AST parametrized by a merge function:
 
 ```
-interpret (Merge(ast1,meta1),Merge(ast2,meta2)) f =
+Priority = Number | -inf | +inf
+
+type MergeFunction =
+  {value: Dyn, priority: Priority} ->
+  {value: Dyn, priority: Priority} -> Dyn
+
+interpret (Merge(ast_1,meta_1),Merge(ast_2,meta_2)) f =
     f {value = interpret(astMin), prio = metaMin.priority}
       {value = interpret(astMax), prio = metaMax.priority}
     where
-      i s.t min(meta1.priority, meta2.priority) = metai.priority 
-      astMin = asti, metaMin = metai
-      astMax = ast(3-i), metaMax = meta(3-i)
+      i s.t min(meta_1.priority, meta_2.priority) = meta_i.priority
+      j = 1  if i == 0
+          0  otherwise
+      astMin = ast_i, metaMin = meta_i
+      astMax = ast_j, metaMax = meta_j
 ```
 
 We can finally define the evaluation of a merge AST:
@@ -892,5 +900,6 @@ eval ast =
 
 **TODO**
 
-- implementation: separate merge AST/merge expression from values. Use a `force`
-    or `eval` semantics to be correct w.r.t lazy evaluation.
+- implementation: separate merge AST/merge expression from values, as is already
+  the case with meta values. Use a `force` or `eval` semantics to be correct
+  w.r.t lazy evaluation.
--- a/notes/overriding.md
+++ b/notes/overriding.md
@@ -439,14 +439,14 @@ This brings the problem of scoping: should a record be able to access a yet
 undefined field because it is expected to be provided by a subsequent merge?
 Two possible approaches:
 
- - **dynamically scoping**: records can reference fields that are not explicitly
+ - **dynamic scoping**: records can reference fields that are not explicitly
    defined locally, such as:
    ```nickel
    {a = b} & {b = 1}
    ```
    However, dynamic scoping have a number of issues, and is usually considered
    bad practice.
- - **lexically scoping**: as currently, require self-referenced fields to be
+ - **lexical scoping**: as currently, require self-referenced fields to be
    defined locally. Note that thanks to contracts, one can require the existence
    of a field without defining it. For example, we could write the previous
    example as:
--- a/notes/overriding.md
+++ b/notes/overriding.md
@@ -1,0 +1,651 @@
+---
+feature: overriding
+start-date: 2021-03-18
+author: Yann Hamdaoui
+---
+
+# Overriding
+
+This document is a proposal for an overriding mechanism in Nickel.
+
+Related issues: [#103](https://github.com/tweag/nickel/issues/103),
+[#240](https://github.com/tweag/nickel/issues/240),
+[#255](https://github.com/tweag/nickel/issues/255),
+[#279](https://github.com/tweag/nickel/issues/279).
+
+## Context
+
+### Definition of the problem
+
+*Overriding* is the ability to take an existing configuration - in our case, a
+record - and update a specific part of it, such as a field or the field of a
+nested record.
+
+A naive and unergonomic of way of doing it is to repeat all the original
+values in a new record:
+
+```
+let record = {
+  a = 1,
+  b = "str",
+  c = false
+} in
+{
+  a = record.a,
+  b = record.b,
+  c = true
+}
+```
+
+This quickly become unmanageable for large records, which is common in Nix and
+Nixpkgs. In functional languages, this is usually better done using what's
+called functional update. Functional update is an operation that takes the
+original record, one or several field to update, the corresponding new value,
+and return a new updated record. It has the same semantics as our first snippet,
+but doesn't require to rewrite all unchanged fields.
+
+It can have a builtin syntax, such as OCaml's `with`: `{record with field =
+new_value}`, Haskell's `record {field = newValue}`, Nix `//` operator `record //
+{field = newValue}`, or Rust's syntax `RecordDataType {field: new_value,
+..record}`. There are more advanced programming techniques that make updating
+deeply nested records ergonomic such as
+[Lenses](https://www.fpcomplete.com/haskell/tutorial/lens/) in Haskell, but
+these rely on advanced language and typing features to be practical.
+
+In Nickel, we can already do a functional update using merging, although the
+updated field must have been marked as `default`:
+
+```nickel
+let record = {
+  a = 1,
+  b = "str",
+  c | default = false,
+} in
+record & {c = true}
+```
+
+As explained in the next section though, this is not satisfying.
+
+### Overriding recursive records
+
+Nickel's records are different from the ones of OCaml, Haskell or Rust.  They
+are lazy and recursive by default. They are thus better understood as
+[codata](https://www.researchgate.net/publication/332248372_Codata_in_Action)
+rather than data. Take the following example:
+
+```nickel
+let record = {
+  protocol = `Http,
+  port | default = switch protocol {
+    `Http => 80,
+    `Ftp => 21,
+    _ => 8181,
+  },
+} in
+record & { protocol = `Ftp }
+```
+
+Here, our final record will be ``{protocol = `Ftp, port = 80}``. `port` is an
+expression depending on `protocol`, but it is not updated in the new version,
+which is something one could intuitively expect from a recursive record. Such
+recursive update is a pattern actively used in Nix, and is probably a common
+scenario for configurations in general: if I patch a base configuration by
+changing an option, I usually want other default values depending on this option
+to be updated as well.
+
+## Overriding in Nix
+
+Nix expressions don't have a built-in way of doing this either. Since the
+feature is still actively needed, it is provided by mechanisms implemented in
+library code.
+
+Keeping a language lean is usually a good design guideline: to provide an
+expressive yet simple set of features upon which others can be built as
+libraries. However, in the case of Nix, overriding is a fundamental pattern, and
+having it implemented in user code leads to some well-known issues:
+
+- Definition, update or field access may not be as ergonomic as with vanilla
+  native records.
+- Several competing mechanisms have been implemented in slightly different
+  contexts. Overriding in general is already difficult for beginners, but having
+  to learn several ones, all similar but still different, is hard.
+- User land implementations don't have a first-class access to information such
+  as source location and metavalues. In a dynamic language, this can make good
+  error reporting much harder.
+- It is also potentially harder to make user land implementations efficient.
+
+See [this gist](https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d)
+for more details. We continue with an overview of existing mechanisms in Nix and
+related languages.
+
+### Nixpkgs overriding
+
+The basic idea is to represent a recursive record explicitly as a function from
+a `self` record to the final record:
+
+```nix
+r = rec {
+  a = 1;
+  b = a + 1;
+}
+# r is represented as:
+rRepr = self: {
+  a = 1;
+  b = self.a + 1;
+}
+```
+
+`self` is a self-reference, in the style of `this` in object oriented languages.
+It is computed as a
+[fixpoint](https://en.wikipedia.org/wiki/Fixed-point_combinator), simply
+realized by auto-application thanks to laziness:
+
+```
+r = let fixpoint = rRepr fixpoint; in fixpoint
+```
+
+The explicit dependency to `self` gives the ability to provide a different
+definition for `self.a` in the final value. Overriding is achieved exactly by
+functionally updating the `a` field of the `self` parameter before passing it to
+the original representation:
+
+```nix
+let extension = {a = 2;}; in
+# The fixpoint of result is { a = 2; b = 3; }
+resultRepr = self: (rRepr (self // extension)) // extension
+```
+
+The second outer update ensures that the final result is also set to `a = 2`,
+and not only the `a` appearing in `b`.
+
+Some details are left out, but this is the gist of it. See also
+[this pill on overriding](https://nixos.org/guides/nix-pills/override-design-pattern.html)
+or [this article on fixpoints in Nix](http://r6.ca/blog/20140422T142911Z.html).
+
+**limits**
+
+- Use a specific representation, rather than handling good old plain records
+  (although the actual representation in Nixpkgs is more ergonomic than a plain
+  function).
+- Overriding nested attribute sets is painful. If one do the naive thing,
+  the whole subrecord is erased:
+  ```nix
+  let rRepr = self: {
+    a = {b = self.a.c;};
+  }; in
+  let extension = {a = {c = 2;};}; in
+  rExt = let fixpoint = rRepr (fixpoint // extension); in
+  fixpoint // extension
+  # Gives {a = {c = 2;};} instead of expected {a = {b = 2; c = 2;};}
+  ```
+- They don't [compose well](https://github.com/NixOS/nixpkgs/issues/16394).
+
+### Nixpkgs overlays
+
+[Overlays](https://nixos.wiki/wiki/Overlays) were defined to address the
+composability issue of the previous mechanism, that is, being able to apply
+several overrides in a modular way. Overlays can be seen as a sequence of
+transformations from a base record, each layer having access to a `super`
+reference to the previous layer and the `self` reference to the final value.
+
+Take two consecutive transformations:
+
+1. Set the `a` field to `1`
+2. Set the `b` field to `1`, add `1` to the `a` field
+
+```nix
+let baseRepr = self: {c = self.a + self.b;}; in
+let overlay1 = self: super: {a = 1;}; in
+let overlay2 = self: super: {b = 1; a = super.a + 1;}; in
+let applyOverlays = self:
+  let base = baseRepr self; in
+  let first = base // overlay1 self base; in
+  let second = first // overlay2 self first; in
+  second; in
+let fixpoint = applyOverlays fixpoint; in fixpoint
+```
+
+In practice, the `super // ..` and fixpoints parts can be factorised in
+dedicated helper functions.
+
+**limits**
+ - Modifications must be applied in the right order.
+ - Still need to manipulate representations.
+ - Overriding nested fields is still clumsy:
+   ```nix
+   self: super:
+   {
+     lib = (super.lib or {}) // {
+       firefoxVersion = ...;
+     };
+   }
+   ```
+
+### NixOs module system
+
+The NixOS module system takes a different approach. There, all the basic blocks
+-- the modules -- are merged together in an unspecified order. What's deciding
+the priority of one option over the other are attributes that are explicitly
+stated in the modules themselves.
+
+Nonetheless, modules need to combine different pieces of data together rather
+than just overriding one value with another. This is made difficult by the
+disappearance of the explicit `super`, compared to overlays.
+
+The issue is solved by custom merge functions, that can define how to combines
+different values of the field of a configuration. By default, when merging two
+list values, the module system only knows how to replace one value with the
+other, because there's no canonical and commutative way of merging two lists.
+However, the user can specify that these lists should in fact be concatenated,
+resulting in the possibility of defining a list of paths by pieces:
+
+```nix
+# Some module
+{
+  paths = mkOption {
+    type = types.listOf types.path;
+  };
+}
+
+# Config
+[..]
+someModule.paths = [foo/bar]; 
+
+# Other config
+[..]
+someModule.paths = [/bar/baz];
+
+# Resulting config: [foo/bar /bar/baz]
+```
+
+## Overriding elsewhere
+
+To my knowledge, among the other configuration languages Nickel is close to,
+Jsonnet is the only one to feature a proper overriding mechanism.
+
+### CUE
+CUE allows a form of late-binding for recursive attributes:
+
+```
+$cat test.cue
+fields: {
+ a: int
+ b: a + 1
+} & {
+ a: 2
+}
+$cue eval test.cue
+fields: {
+    a: 2
+    b: 3
+}
+```
+
+However, by the very nature of CUE's merge operation, the value of `a` can't be
+changed by merging, it can only be made more precise:
+
+```
+$cat test2.cue
+fields: {
+ a: 1
+ b: a + 1
+} & {
+ a: 2
+}
+$cue eval test2.cue
+fields.a: conflicting values 1 and 2:
+    ./test2.cue:1:9
+    ./test2.cue:2:5
+    ./test2.cue:5:5
+```
+
+It is thus not really an overriding mechanism.
+
+### Jsonnet
+
+Jsonnet refers to its overriding mechanism as *inheritance*, implemented by the
+`+` operator on objects:
+
+```jsonnet
+local obj = {
+  name: "Alice",
+  greeting: "Hello, " + self.name,
+};
+[
+  obj,
+  obj + { name: "Bob" },
+  obj + { greeting: super.greeting + "!"},
+  obj + { name: "Bob", greeting: super.greeting + "!"},
+]
+```
+
+This is similar to the Nix operator `//`, but doing recursive overriding in the
+expected way out of the box. The extension can access the previous version in
+the same way as Nixpkgs overlays, using the `super` keyword.
+
+## Taking a step back
+
+While not exactly the same, all these overriding mechanisms are based on the
+same underlying principles:
+1. Represent recursive records explicitly as a function of `self` (and `super`
+   in some cases)
+2. Compute the combination as a **fixpoint** of several recursive records,
+   giving back a standard record.
+
+As hinted by Jsonnet's terminology, the semantics of recursive records
+strikingly resemble the semantics of objects and classes in OOP.  Replace
+records with objects, fields with methods and overriding with inheritance. This
+is not so surprising: there's actually an history of encoding objects in
+functional languages as recursive records (see the introduction of
+[The Recursive Record Semantics of Objects Revisited](https://hal.inria.fr/inria-00072423)
+for a good overview) going back to 1988
+\[[Cardelli](http://lucacardelli.name/papers/inheritance.pdf)\].
+
+This is also mentioned in the
+[README](https://github.com/MuKnIO/nixpkgs/blob/devel/lib/pop.md#some-historical-context)
+of POP (an object system in Nix), where the author observes that overriding
+mechanisms in Nix (and Jsonnet for that matter) are a simplified lazy object
+system (simplified because objects lack proper state and there is no distinction
+between classes and instances). Their logical conclusion is to embrace this fact
+and design a proper object system helped by existing literature, rather than
+reinventing the wheel. Similarly, Nix [overlays](https://nixos.wiki/wiki/Overlays)
+can be seen as a single inheritance mechanism.
+
+Inheritance-based overriding imposes an order on the overrides. A single level
+inheritance is usually fine, but a complex hierarchy can become hard to maintain
+and to reason about.
+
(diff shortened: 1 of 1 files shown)


notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+   val1 & val2 & (val3 | merge func)
+   <=> (val1 | merge func) & val2 & val3
+   <=> func (func val1 val2) val3
+   // instead of the naive
+   <=/=> func (val1 & val2) val3
+   ```
+
+   But what if `val1 & val2` is in a thunk that happens to be forced in between?
+   What should `let x = val1 & val2 in x & (val3 | merge func)` evaluate to? And
+   `let x = val1 & val2 in builtins.seq x (x & (val3 | merge func))`?
```
regnat, line 654: I think forcing and “closing” a record should be too distinct operations.

The way I understand the rest of the semantics above, there's two kind of records: open and closed. Open records are the one available in the surface language and with all the shiny overriding semantics while closed records are an internal thing only.

If we represent closed records with `{| |}` (to distinguish them from open records `{ }`), we have that

1. `{ x = 1, y = x } == fun self => {| x= 1, y = self.y |}` (the `repr(r)` operator above)
2. `{| something |} == let fix = { something } fix in fix` (what we can call the `close(r)` operator)

To rephrase what you defined before, we then have that for an open record `r`, `r.foo == close(r).foo`.

All this to say that `val1 & val2` is still an open record (whether you force it or not), so `let x = val1 & val2 in builtins.seq x (x & (val3 | merge func))` is the same as `val1 & val2 & (val3 | merge func)`

yannham: I was not even thinking about records, but values in general. The problem is, morally, that using a custom merge function amounts to interpret the merge AST with a different semantics for `&`. If we want it to be non-local, as in only one term with an annotation in the merge expression is sufficient to change the merge function for all operands, you need to have access to the whole original AST. Forcing thunks may evaluate sub-expressions with the original merge semantics (and not the custom, because it is not in the scope of the sub-expression), and this changes the result.

Example:

```
let add = fun x y => x + y in
let val | merge add = 1 in
// Say we do "the right thing", so this is 3 as expected
1 & 1 & val

let x = 1 & 1 in
// This has no chance of being 3
// x is forced, without any custom merge function in sight.
// `1 & 1` is thus computed to be `1` (merge is idempotent)
// The thunk of x is updated with 1, while the original expression `1 & 1` is lost.
// so x is `1` and the final result is `2`
seq x (x & val)
```
Merging numbers directly is dubious, but you can imagine them being values for an option: `{ ..., someOption | Num | merge add, ...}`.

The next paragraph notices that, though, records will need to be instrumented to store the original expression along the computed thunk (as a way of representing open records), and in that case we can make it work. Which is I think what you meant? But, in general, `val1` and others are not necessarily records nor record fields.

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+record & {c = true}
+```
+
+As explained in the next section though, this is not satisfying.
+
+### Overriding recursive records
+
+Nickel's records are different from the ones of OCaml, Haskell or Rust.  They
+are lazy and recursive by default. They are thus better understood as
+[codata](https://www.researchgate.net/publication/332248372_Codata_in_Action)
```
aspiwack, line 75: Please no researchgate links. Let's provide the publisher's link. 
```suggestion
[codata](https://link.springer.com/chapter/10.1007/978-3-030-17184-1_5)
```

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+and return a new updated record. It has the same semantics as our first snippet,
+but doesn't require to rewrite all unchanged fields.
+
+It can have a builtin syntax, such as OCaml's `with`: `{record with field =
+new_value}`, Haskell's `record {field = newValue}`, Nix `//` operator `record //
+{field = newValue}`, or Rust's syntax `RecordDataType {field: new_value,
+..record}`. There are more advanced programming techniques that make updating
+deeply nested records ergonomic such as
+[Lenses](https://www.fpcomplete.com/haskell/tutorial/lens/) in Haskell, but
+these rely too heavily on advanced language and typing features to be practical.
```
aspiwack, line 55: It's also incorrect, in that we want some kind of “late binding” behaviour in recursive records. Maybe it should be explained.

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+expression depending on `protocol`, but it is not updated in the new version,
+which is something one could intuitively expect from a recursive record. Such
+recursive update is a pattern actively used in Nix, and is probably a common
+scenario for configurations in general: if I patch a base configuration by
+changing an option, I usually want other default values depending on this option
+to be updated as well.
+
+## Overriding in Nix
+
+Nix expressions don't have a built-in way of doing this either. Since the
```
aspiwack, line 100: “this” = ?

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+  b = self.a + 1;
+}
+```
+
+`self` is a self-reference, in the style of `this` in object oriented languages.
+It is computed as a
+[fixpoint](https://en.wikipedia.org/wiki/Fixed-point_combinator), simply
+realized by auto-application thanks to laziness:
+
+```
```
aspiwack, line 145:
```suggestion
```nix
```

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+[this pill on overriding](https://nixos.org/guides/nix-pills/override-design-pattern.html)
+or [this article on fixpoints in Nix](http://r6.ca/blog/20140422T142911Z.html).
+
+**limits**
+
+- Use a specific representation, rather than handling good old plain records
+  (although the actual representation in Nixpkgs is more ergonomic than a plain
+  function).
+- Overriding nested attribute sets is painful. If one do the naive thing,
+  the whole subrecord is erased:
```
aspiwack, line 173: You're making a good job at pointing out the limitations at the beginning. But then you don' t use them that much. I would list the various limitations in a dedicated section near the top. And _give them names_. Then you just say: this has problem Foo.

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+  ```nix
+  let rRepr = self: {
+    a = {b = self.a.c;};
+  }; in
+  let extension = {a = {c = 2;};}; in
+  rExt = let fixpoint = rRepr (fixpoint // extension); in
+  fixpoint // extension
+  # Gives {a = {c = 2;};} instead of expected {a = {b = 2; c = 2;};}
+  ```
+- They don't [compose well](https://github.com/NixOS/nixpkgs/issues/16394).
```
aspiwack, line 183: This is a bit of a fud-type sentence. It doesn't bear much meaning. Maybe you can say an extra sentence (like “Nixpkgs has been known to have several incompatible overriding mechanism”). Or maybe just “There are more known difficulties”.

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+ a: 2
+}
+$cue eval test2.cue
+fields.a: conflicting values 1 and 2:
+    ./test2.cue:1:9
+    ./test2.cue:2:5
+    ./test2.cue:5:5
+```
+
+It is thus not really an overriding mechanism.
```
aspiwack, line 304: Well, it kind of is. With the default value thing. But it is not as complete as what we are trying to achieve, probably. There lacks in the initial paragraph an explicit goal that would rule Cue's mechanism out.

notes/overriding.md
```diff
@@ -0,0 +1,675 @@
...
+  obj + { greeting: super.greeting + "!"},
+  obj + { name: "Bob", greeting: super.greeting + "!"},
+]
+```
+
+This is similar to the Nix operator `//`, but doing recursive overriding in the
+expected way out of the box. The extension can access the previous version in
+the same way as Nixpkgs overlays, using the `super` keyword.
+
+## Taking a step back
```
aspiwack, line 328: I'll read from here on later.

notes/overriding.md
yannham: My bad, I forgot that CUE also had default values. Then the limitation is the one mentioned in the section on priorities, that you probably don't want to have the config writers to annotate everything with `default` to make it overridable. But indeed, it is an overriding mechanism. I'll rewrite this bit

notes/overriding.md
yannham: This is explained in the following section.

notes/overriding.md
yannham: It is not about several mechanisms, but about applying several overrides to the same configuration. It seems the Nixpkgs basic overriding mechanism doesn't support this well, and may reset initial default values that were changed at some point as long as the last override doesn't update them (https://github.com/NixOS/nixpkgs/issues/16394#issuecomment-233032852). I should include a stand-alone example.

yannham: After trying on simple examples, it doesn't seem to be a problem actually. Let me gather more info and come back to this.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+}
+```
+
+`self` is a self-reference, in the style of `this` in object oriented languages.
+It is computed as a
+[fixpoint](https://en.wikipedia.org/wiki/Fixed-point_combinator), simply
+realized by auto-application thanks to laziness:
+
+```nix
+r = let fixpoint = rRepr fixpoint; in fixpoint
```
aspiwack, line 147: Maybe simply:

```suggestion
r = rRepr r
```

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+fields: {
+    a: 2
+    b: 3
+}
+```
+
+The basics of this overriding mechanism are close in spirit to what we want to
+achieve in this proposal. This is similar to Nickel's current `default`
+behavior, but with the desired late-bound merging. This is however not enough
+for Nix use cases: we will need to add other features to the merge system.
```
aspiwack, line 305: What other features?

Cf what I said the other day about naming stuff. You can start by naming the features that you need, and then explain which features are met or not by different solutions. The text will flow much better. And it will help frame the conversation and the documentation in the future.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+recursive (or "leafy", or "push down") annotations, as described in
+[#279](https://github.com/tweag/nickel/issues/279):
+
+```nickel
+let someNeutraleConfig = {
+  foo = 1,
+  bar.baz = "stuff",
+  bar.blorg = false,
+}
+let defaulted | default rec = someNeutralConfig;
```
aspiwack, line 494: If `default rec` is a new keyword that you are introducing, please say so.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+   This is also a better practice to explicitly state the fields whose presence
+   is required.
+
+### Priorities
+
+Merge is fundamentally commutative. The problem is, not all fields should be
+treated the same: some are default values that ought to be overridden, some are
+high-priority values that ought to override. The solution is to use a priority
+system, encoding the precedence of each value while retaining commutativity
+(only the priority annotation counts, not the order in which values are merged).
```
aspiwack, line 465: You want to be much more precise here. You really want to spell out the semantics of priorities. So basically the entire merge algorithm with priorities (you don't need to specify the set priorities for that, you just need that the set of priorities has an order (even a partial order will do)).

The question of how to merge records with different priorities may arise. If so it will then be a natural question about whether we want to just take the record with the highest priority, or take some more complex merge, which may remove the need for the leafy default thing.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+
+Priorities have drawbacks. One is non-locality: the final result depends on
+other priorities of the field's values you are being merged with, potentially
+defined elsewhere. However, this can be mitigated by good defaults and an
+adapted set of priorities. For example:
+
+1. `default` is lowest than everything else (bottom)
+2. Integer priorities in the middle, with `0` being the default
```
aspiwack, line 473: You probably want to pull out the remark “`0` being the default`” in its own paragraph, because it deserves justification.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+
+This way, an existing definition (arbitrarily complex: that could be the root of
+all Nixpkgs) can easily (and lazily) be turned into a full overriding or
+overridable configuration.
+
+### Custom merge functions
+
+Custom merge functions would be specified by a `merge` metavalue attribute. In
+order to enforce commutativity, they would receive their arguments in an
+indistinguishable order, such as `{lower: Dyn, higher: Dyn}`, just knowing which
```
aspiwack, line 518: I assume that the custom merge functions will probably need to know when they are getting two terms of equal priority.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+one has a higher priority than the other. If both have the same priority, the
+order is not specified, and may even be randomized by the interpreter.
```
aspiwack, line 520: You definitely want to specify how merge functions will work precisely.

notes/overriding.md
```diff
@@ -0,0 +1,676 @@
...
+
+```nickel
+let mergeLists : forall a. {lower: List a, higher: List a} -> List a = fun args =>
+  args.lower @ args.higher in
+
+let Contract = {
+  path | List Str
+       | merge mergeLists
```
aspiwack, line 528: Here is an example of question that needs answering: what happens if you are trying to merge two meta values with the same type but different merge function. As your syntax seems to allow it.

notes/overriding.md
```diff
@@ -0,0 +1,797 @@
...
+(only the priority annotation counts, not the order in which values are merged).
+
+Priorities have drawbacks. One is non-locality: the final result depends on
+other priorities of the field's values you are being merged with, potentially
+defined elsewhere. However, this can be mitigated by good defaults and an
+adapted set of priorities, such as a bottom `default`, a top `force`, and a
+infinite range of integers priorities in between. Doing so, it's easy to just
+use `default` or `force` to override or be overridden by anything, without
+knowing the precise integer priority. Integer priorities still gives freedom
+with an infinite supply of levels if required.
```
Radvendii, line 506: This adds a problem of "what arbitrary integer priority do I assign this? The only way I can think to figure this out is look at everything else that sets that value, and figure out exactly what you want to override and be overridden by, and if people didn't leave enough space between their integers, you have to adjust their values. Maybe this is the best we can do, but it doesn't seem very ergonomic to me.

Maybe this has already been addressed in the NixOS module system? Just reading this document it's not clear to me how to decide on an integer.

notes/overriding.md
```diff
@@ -0,0 +1,797 @@
...
+
+### Custom merge functions
+
+Custom merge functions would be specified by a `merge` metavalue attribute. In
+order to enforce commutativity, they would receive their arguments in an
+indistinguishable order, such as having the type `{lower: Dyn, higher: Dyn,
+priority: <Different, Equal> -> Dyn`.  If both have the same priority, the order
+is not specified, and may even be randomized by the interpreter. The `priority`
+field indicates when it is the case, if this case needs special handling.
```
Radvendii, line 515: This doesn't actually enforce commutativity, it just encourages it. The function can still be non-commutative, and this will introduce weird ordering dependencies that people might be surprised by. 

If the order is randomized, then something might work and break randomly half the time. That seems like a difficult bug to track down.

If the order is not randomized, then you may inadvertently create a dependency on order that only shows up when you refactor.

Radvendii (commented):
Just my thoughts as an avid Nix user with almost no experience of Nickel. Let me know if I'm way off base and need to learn more about Nickel before commenting.

---

I don't understand after reading this why the NixOS module - style updating is the preferred option. It seems that the main advantage is commutativity, but it also introduces a good deal of complexity. As a user, I find the Nix module system quite mysterious compared to the semantics of `lib.recursiveUpdate`, `//`, or even overrlays, which are all non-commutative.

Modules always "just work" for me because the merge functions were designed very carefully to do the right thing in all of those cases. If something went wrong though, I think it would be fairly confusing to work out where or why. As a user, I'm not confident in my ability to consistently design those merge functions correctly.

As such, I think this is an incredibly powerful approach in something like modules, which has a very clear develop / use divide, but I'm not sold on it for general records.

It's not clear to me what the simple / basic use case is. The use of "override this record with some other values" (`//` or `lib.recursiveUpdate`) requires extra syntax (at  least the `force rec` you introduce at the end), and isn't even guaranteed to work, if there has already been a forced value. On the other hand, setting up a generic way for attributes to be merged requires writing a function and adding a `| merge func`. This is a lot of effort for something farily basic. I'm left feeling like there is no simple / obvious way to change records.

notes/overriding.md
```diff
@@ -0,0 +1,797 @@
...
+  port | default = switch protocol {
+    `Http => 80,
+    `Ftp => 21,
+    _ => 8181,
+  },
+} in
+record & { protocol = `Ftp }
+```
+
+Here, our final record will be ``{protocol = `Ftp, port = 80}``. `port` is an
```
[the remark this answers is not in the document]
Radvendii: I think that's the point of this part of the doc. It's counter-intuitive the way it works ATM. The reason is that port is not defined via a thunk, but directly in terms of protocol, and therefore gets evaluated immediately.

If you wrote
```
let
  foo = 1;
  bar = foo + 1;
in
{
  foo = 5;
  bar = bar;
}
```

you would get `bar = 2`, not `bar = 6`. It's the same thing going on, it's just in this case our intuition tends to be different.

notes/overriding.md
Radvendii: To be clear, the way things are combined in NixOS modules is also non-commutative, it's just usually okay enough to get by. For instance, bash scripts are concatenated, which is only commutative if you're careful about what content you're putting in those bash scripts.

yannham:
First, thanks for the review, @Radvendii. Familiarity with Nickel is not required: having an external point of view is valuable.

> Just my thoughts as an avid Nix user with almost no experience of Nickel. Let me know if I'm way off base and need to learn more about Nickel before commenting.
> 
> I don't understand after reading this why the NixOS module - style updating is the preferred option. It seems that the main advantage is commutativity, but it also introduces a good deal of complexity. As a user, I find the Nix module system quite mysterious compared to the semantics of `lib.recursiveUpdate`, `//`, or even overrlays, which are all non-commutative.

The merge system is inspired in part from NixOS, but it is not exactly the same thing. It's also inspired from CUE. I would argue that the semantics of merging is more natural than `//` in Nix, that doesn't do the intuitive thing:
 - Usually you want the semantics of `lib.recursiveUpdate` for nested fields, but `//` doesn't do that
 - You would like recursive records to be updated in the expected way and not as in the port/protocol example here, but `//` doesn't do that

So you need three different notions (and actually there are even more beside overlays for overriding) defined in different places, working in different ways, to do the same thing in Nix currently. Overlays and co also don't work with basic records, as recalled in this document. Meanwhile:

- Merging two records recursively is just `r1 & r2`
- If you need to override something, you just have to write one additional annotation explicitly `r1 & (r2 | force rec)` for example. I think requiring an explicit marker for overriding is a good thing in general, not only from a technical/commutativity point of view.
 
> Modules always "just work" for me because the merge functions were designed very carefully to do the right thing in all of those cases. If something went wrong though, I think it would be fairly confusing to work out where or why. As a user, I'm not confident in my ability to consistently design those merge functions correctly.
> As such, I think this is an incredibly powerful approach in something like modules, which has a very clear develop / use divide, but I'm not sold on it for general records.
> 
> It's not clear to me what the simple / basic use case is. The use of "override this record with some other values" (`//` or `lib.recursiveUpdate`) requires extra syntax (at least the `force rec` you introduce at the end), and isn't even guaranteed to work, if there has already been a forced value. On the other hand, setting up a generic way for attributes to be merged requires writing a function and adding a `| merge func`. This is a lot of effort for something fairly basic. I'm left feeling like there is no simple / obvious way to change records.

I think the use case is modularity. We are talking about the NixOS module system a lot here, but the question is probably how to provide the language features to be able to write big and complex configurations in separate blocks that can then be combined, reused and possibly customized.

The general idea is inspired from CUE, where merging is ubiquitous, rather than it is an ad-hoc re-implementation of the NixOS system. It just turns out NixOS is a good inspiration for enhancements than CUE doesn't need nor want to provide, as it is intentionally very restricted in processing and computing power.

To answer more precisely to some of your points:

> As such, I think this is an incredibly powerful approach in something like modules, which has a very clear develop / use divide, but I'm not sold on it for general records.

I'm genuinely curious (as just a basic user of Nix/NixOS), what makes NixOS modules different from, say, writing a big Kubernetes configuration in the same modular style with respect to this develop/use ratio? 

> and isn't even guaranteed to work, if there has already been a forced value.

True. Note that in CUE things can only be overridden once, and it is by design, as multiple levels of overriding is seen as the root of all evil (overriding being in some sense similar to inheritance). I've been told Nix requires more 'monkey-patching' capabilities. This has been discussed in https://github.com/tweag/nickel/issues/240, with a proposal to be able to access the priority of values to implement an equivalent of `//`.

>  On the other hand, setting up a generic way for attributes to be merged requires writing a function and adding a `| merge func`. This is a lot of effort for something farily basic.

I think calling custom merge functions _basic_ is a bit unfair. As proposed here, merging is akin to building a free monoid with a default implementation for records (and a trivial one for other values). Custom merge functions would allow to re-interpret some part of the AST with a user-defined function, as a form of overloading. To my knowledge, neither Nix, CUE, Jsonnet nor Dhall supports something like this out of the box.

> I'm left feeling like there is no simple / obvious way to change records.

I don't think it is the case. Jsonnet, for example, features a fairly simple mechanism for overriding: it is basically `//` that does the right thing for recursive records. The question is how to handle the case where several values are defined for the same field. Jsonnet encodes this directly in the order in which records are extended: the last (right-most) value wins. The NixOS module system and CUE's merge system, on the other hand, rely on a priority system (with only a bottom and a top for CUE, so field can be overridden only once). Those are the different tradeoffs that need to be discussed here, keeping in mind the use-case of Nickel as, on one side, a "general-purpose" configuration language, but also as a possible future for Nix.

notes/overriding.md
yannham: That's currently the biggest complaint I have with the priority system. One solution is to only allow very few and natural priorities (normal, default and force e.g.), but it seems this would be too limiting for Nix. Maybe @edolstra , @Infinisil or @regnat have more to say on the practicality of priorities.

notes/overriding.md
yannham: The general problem is undecidable (not just by Rice's theorem, but in a deeper way, see e.g. [this resource](https://dl.acm.org/doi/10.1145/570886.570889)) and I don't think we can do much better (without a proof system to express and prove that a function is commutative). The problem is here for type class laws in Haskell for example, where you can use the do-notation with instances that don't actually respect monad laws, and could make expected equivalences invalid in practice.

One solution in the case of Nickel could be to use contracts to check that each application of the custom merge function respect commutativity:
```
let Commutative = fun label f x y =>
  let value = f x y in
  if value == f y x then
    value
  else
    contracts.blame label
```

That would add quite some overhead though. Although it will be possible to disable contracts with a flag in the future.

notes/overriding.md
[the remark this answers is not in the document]
yannham: Roughly yes, but to be pedantic `port` is actually a thunk, and is only evaluated if used. The thing is, the thunk captures the **initial** `protocol` variable corresponding to the moment the thunk is created, and not the second version of `protocol` that is coming from the merge. See https://github.com/tweag/nickel/issues/103 for more context about the interaction between evaluation, recursive records and merging.

notes/overriding.md
Radvendii: Ooh, I haven't looked into contracts but that looks really cool. This seems to just be wrapping the function in a way that verifies it's commutative, which makes a lot of sense to me.

If that were how things were implemented, I would be much more sold on this solution. Even if it weren't enforced, but were expected that things were actually commutative. Like how monad laws are expected to hold. The person implementing the monad laws are on the hook for making sure they hold; it's not the job of the person *using* `bind` to make sure they pass in the right arguments so that the monad laws hold.

By contrast, NixOS modules have concatenation of bash scripts as a "commutative" operation. But that clearly isn't. I have to limit myself to a subset of bash that is commutative. But it's even worse than that, For instance at one point in my configuration I set the default cursor shape in the X init script. If I were to do that same operation (with different cursor shapes) in two different places that were supposed to be merged, the behaviour is undefined. It's more like if there were tons of examples of Monads people use on a daily basis where the monad laws only hold on a subset of that type and people are just expected to only use that subset in a `do` block.

Maybe this is a way that nickel would be different from NixOS modules (I don't know how cue handles this).

Radvendii:
> The merge system is inspired in part from NixOS, but it is not exactly the same thing. It's also inspired from CUE. I would argue that the semantics of merging is more natural than `//` in Nix, that doesn't do the intuitive thing:
>
> * Usually you want the semantics of `lib.recursiveUpdate` for nested fields, but `//` doesn't do that
> * You would like recursive records to be updated in the expected way and not as in the port/protocol example here, but `//` doesn't do that

Totally agree that both of these are what I want most of the time. My only issue is with the merging / commutative semantics

> * Merging two records recursively is just `r1 & r2`
> * If you need to override something, you just have to write one additional annotation explicitly `r1 & (r2 | force rec)` for example. I think requiring an explicit marker for overriding is a good thing in general, not only from a technical/commutativity point of view.

I am generally loathe to use things like `force` or `mkForce` (NixOS) or `!important` (CSS). It feels like a hack that is bound to cause trouble down the road. How do I know that I will never have something of higher priority later on? I think it's important to be able to have something where I am locally guaranteed to be overriding. Put differently, once you start using `r1 & (r2 | force rec)`, that operation is no longer composable.

> > I'm left feeling like there is no simple / obvious way to change records.
>
> I don't think it is the case. Jsonnet, for example, features a fairly simple mechanism for overriding: it is basically `//` that does the right thing for recursive records. The question is how to handle the case where several values are defined for the same field. Jsonnet encodes this directly in the order in which records are extended: the last (right-most) value wins. The NixOS module system and CUE's merge system, on the other hand, rely on a priority system (with only a bottom and a top for CUE, so field can be overridden only once). Those are the different tradeoffs that need to be discussed here, keeping in mind the use-case of Nickel as, on one side, a "general-purpose" configuration language, but also as a possible future for Nix.

I guess a key point that I wasn't thinking about is this: most of the time, there are no conflicts (several values defined for the same field), so one can just use `&` without thinking about it too much. This is the "simple case" that I was looking for. What this means is any conflict is marked by the language as a complex case. It has to be dealt with either by explicitly forcing, or by a custom merge function.

> I'm genuinely curious (as just a basic user of Nix/NixOS), what makes NixOS modules different from, say, writing a big Kubernetes configuration in the same modular style with respect to this develop/use ratio?

I've never used kuberrnetes, but what I'm referring to is the fact that attrsets are used as the ubiquetous data structure in Nix. A good deal of the time I want to pass information around, I use attrsets. To give them a complicated semantics where individual fields "know" how to merge themselves involves keeping a lot in my head, unless things "just work". But that takes a lot of very careful design work. Having a simple semantics like `lib.recursiveUpdate` may not always do what I want, but it usually does what I *expect*. I'm worried about not knowing what to expect, because the implementation is dependent on a whole bunch of details that can be buried in various places across a code base.

If I think about attrsets not as a single datatype, but as a form that datatypes take, which I think might be more accurate in Nickel than Nix, this starts to make more sense. Any given attrset that I'm working with, I have to understand what type it is, and that means understanding how it merges.

> I think calling custom merge functions _basic_ is a bit unfair.

Right, sorry, this was me trying to understand what the basic use case was. So it sounds like you're not expecting users to implement custom merge functions very much.

I guess one question I have is: what is the default merging behaviour? This might be in the proposal and I just missed it. Does it error by default if you try to merge two values of the same priority?

notes/overriding.md
Radvendii: If we need only a few more, we could go with default < weak < normal < strong < force, but this seems like a fundamental problem of assigning these kinds of priorities. Unless there are very clear tiers of priority that naturally arise from the problem space, what you're really doing is putting an ordering on a bunch of things that are spread out across the code base.

Another example of this kind of thing is haskell's infix priorities. These work alright for a few different things `($)`, which is a "default" kind of thing, and PEMDAS operations, which have a clearly defined ordering which has to be drilled into your head as a child. I guess in general it works for multiplcation / addition type things where one operation distributes over another. In general though, if I have two random operations? I put in parentheses, because I don't have an intuition for what should be going on.

Radvendii:
Oh! I've come to another realization. While `r1 & (r2 | force rec)` isn't composable, `(r1 | default rec) & r2` *is*! You can do
```
(((r1 | default rec) & r2) | default rec) & r3
```

My issue with this solution is just that it leaves the final attrset with most things default, which might not play nicely combining it with other attrsets later.

yannham:
> I am generally loathe to use things like `force` or `mkForce` (NixOS) or `!important` (CSS). It feels like a hack that is bound to cause trouble down the road. How do I know that I will never have something of higher priority later on? I think it's important to be able to have something where I am locally guaranteed to be overriding. Put differently, once you start using `r1 & (r2 | force rec)`, that operation is no longer composable.

True. As you mentioned later in https://github.com/tweag/nickel/pull/330#issuecomment-867892294, `default rec` seems to be more composable. But not totally, depending on how you define the action of `default rec` on something with the `force` priority for example. One consolation is that in the case of two field with the force priority, the merge system wouldn't chose one arbitrarily but fail, so at least you know what's going on, as opposed to CSS for example.

> I guess a key point that I wasn't thinking about is this: most of the time, there are no conflicts (several values defined for the same field), so one can just use & without thinking about it too much. This is the "simple case" that I was looking for. What this means is any conflict is marked by the language as a complex case. It has to be dealt with either by explicitly forcing, or by a custom merge function.

Exactly. If you don't use any `priority` nor `merge` annotation, you can operate in a subset of the language that just merge disjoint attribute sets and fail otherwise. If you only use the `default` priority, you can operate in subset that behaves like CUE. However, maybe saying that you don't need to care about custom merge functions, integer priorities and `force rec` and `default rec` is wishful thinking from my part: if it's available, it may be used in libraries, and the complexity may leak anyway. Still, locally, the proposal try to sticks to the "if you need to achieve simple results, you can do the intuitive simple thing and ignore the additional complexity" philosophy.

> Right, sorry, this was me trying to understand what the basic use case was.

It's all right :) I didn't mean unfair to someone, but unfair to the expressiveness of custom merge functions.

> So it sounds like you're not expecting users to implement custom merge functions very much.

Yes, I hope. Most custom merge functions would probably be provided by libraries, but I don't expect you'll need a new one very often. A bit like monad instances, in the end.

> I guess one question I have is: what is the default merging behavior? This might be in the proposal and I just missed it. Does it error by default if you try to merge two values of the same priority?

The behavior of `a & b` is:
- if `priority(a) > priority(b)`, keep `a`, and conversely.
- if they have the same priority:
    * if they have different types (record, list, number, etc.), fail
    * if they are both records, merge them recursively (`{ fieldA1 = a.fieldA1, ..., fieldB1 = a.fieldB1, ..., fieldAB1 = a.fieldAB1 & b.fieldAB1, ...}`)
    * if they are of the same type but they are not records, fail, unless they are the same value. That is, `a & b = a` iff `a == b`, or fails otherwise. Currently merging functions always fails, even `f & f`, but this is rather a limitation of the current implementation.

notes/overriding.md
yannham: > By contrast, NixOS modules have concatenation of bash scripts as a "commutative" operation. But that clearly isn't. I have to limit myself to a subset of bash that is commutative. But it's even worse than that, For instance at one point in my configuration I set the default cursor shape in the X init script. If I were to do that same operation (with different cursor shapes) in two different places that were supposed to be merged, the behaviour is undefined. It's more like if there were tons of examples of Monads people use on a daily basis where the monad laws only hold on a subset of that type and people are just expected to only use that subset in a `do` block.
> 
> Maybe this is a way that nickel would be different from NixOS modules (I don't know how cue handles this).

CUE doesn't handle this, because it doesn't have custom merge functions. You can only override a default value with a new value once, but not combine them in a specific way.

This is indeed a problem, but it looks like it is one with the current NixOS module implementation. I see two cases:

- Your merge function is morally commutative, but may not be in practice (for example, if you represent sets with lists, or a set of bash programs as a single string). One solution is to just not care, if we don't enforce commutativity with a contract, as long as it doesn't matter in the end (e.g. the tool consuming the config doesn't care bout it either). Or, you can make it actually commutative (use an ordered list without duplicated elements, for example).
- Your merge function is fundamentally not commutative. Then the merge function can use the priority information to decide of the order: it could fail to merge two scripts with the same priority for example, and otherwise concatenate them from the higher priority to the lower one. There are still subtleties though: you need to remember the priority to keep the operation commutative (otherwise merging priorities 4,2,1 and 4,1,2 wouldn't give the same thing).

To me the bash scripts problem seems to fall in the latter case: while the concatenation is morally commutative for a subset of inputs, it's simply not in general.

Radvendii: Agreed. If there is a strong norm (or enforcement) that we don't have merge functions of the latter sort, we should be alright. clearly it can also work to have such functions, but I think it adds cognitive overhead.

As for the 4,2,1 =/= 4,1,2 issue, that's a failure of associativity I think? Is that an additional property we want to enforce / have as a norm?

notes/overriding.md
```diff
@@ -0,0 +1,942 @@
...
+Merge is fundamentally commutative. The problem is, not all fields should be
+treated the same: some are default values that ought to be overridden, some are
+high-priority values that ought to override. The solution is to use a priority
+system, encoding the precedence of each value while retaining commutativity
+(only the priority annotation counts, not the order in which values are merged).
+
+Priorities have drawbacks. One is non-locality: the final result depends on
+other priorities of the field's values you are being merged with, potentially
+defined elsewhere. However, this can be mitigated by good defaults and an
+adapted set of priorities, such as a bottom `default`, a top `force`, and a
```
aspiwack, line 505: At this point, I'd like to know what the default priority ought to be.

notes/overriding.md
```diff
@@ -0,0 +1,942 @@
...
+  bar | Str,
+  //equivalent to `bar | Str | priority 0`
+
+  baz.boo.bor | priority -4 = "value",
+
+  final | force = `CantOverrideMe,
+}
+```
+
+#### Recursive priorities
```
aspiwack, line 636: I think that this section will need refinement. Its not as well understood as it ought to.

notes/overriding.md
```diff
@@ -0,0 +1,942 @@
...
+  contracts :: Option (List Contract),
+  // ...
+}
+
+data AbsMergeTree a =
+  Merge a a
+  | Exp Expression
+
+type MergeTree = MergeTree (AbsMergeTree (MergeTree, Metadata))
+```
```
aspiwack, line 857: You says Rust. But it really doesn't look like Rust syntax :stuck_out_tongue:

notes/overriding.md
```diff
@@ -0,0 +1,942 @@
...
+### Overriding recursive records
+
+Nickel's records are different from the ones of OCaml, Haskell or Rust.  They
+are lazy and recursive by default. They are thus better understood as
+[codata](https://link.springer.com/chapter/10.1007/978-3-030-17184-1_5)
+rather than data. Take the following example:
+
+```nickel
+let record = {
+  protocol = `Http,
```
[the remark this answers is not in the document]
yannham: `protocol` is not overridden, so I don't think this example particularly needs `protocol` to be `default`

notes/overriding.md
```diff
@@ -0,0 +1,942 @@
...
+  another, instead of combining different pieces of data together. Compared to
+  overlays, the explicit reference to `super` has disappeared.
+
+  The issue is mitigated by custom merge functions, that can redefine how to
+  combine different values of the field of a configuration. By default, when
+  merging two list values, the module system only knows how to replace one value
+  with the other, because there's no canonical and commutative way of merging
+  two lists.  However, the user can specify that these lists should in fact be
+  concatenated, resulting in the possibility of defining a list of paths by
+  pieces:
```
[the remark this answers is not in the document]
yannham: Indeed, the progression is strange. The paragraph about custom merge functions should go higher and be taken into account in the advantages/limits discussion.

notes/overriding.md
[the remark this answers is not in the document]
aspiwack: Not only is `protocol` overridden, but it's the only overridden field in your example.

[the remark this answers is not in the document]
yannham: Wow, it seems I can't read, sorry.

Profpatsch:
Imo the biggest problem with the nix module system is that it’s anti-modular.

That is, in order to find the actual value of an option, you have to first strictly evaluate all modules to the point where you see whether it is overwritten. This makes for a real performance problem in modern nixpkgs, where evaluating a nixos system configuration can take dozens of seconds, even with just a small number of options set (although some of it is excarbated by option definitions being read strictly afaik).

I can’t remember much of this proposal, but in order to scale to very big e.g. kubernetes configs, that has to be a consideration.

roberth:
> you have to first strictly evaluate all modules

Not _strictly evaluate_ but you do have to parse them and evaluate a small amount to find that its contribution to `config` is e.g. `{ config = mkIf false <bottom>; }`.

The problem is that NixOS still imports everything, relying on `*.enable` instead of an accurate `imports` closure. This softened module system requirements by removing failure modes relating to ill-specified imports. If we checked `imports` hermeticity, it seems we could cover most of those. Some modules will have to be split to allow their interface to be used (to no effect) without their implementation.

In this proposal for Nickel I don't see a way to include an externally defined interface. It would be possible with the "dynamic scoping" option which wasn't chosen. The chosen "lexical scoping" seems to require that all field references are declared locally, reminding me of structural types instead of the nominal "I want to interact with the network configuration". Perhaps "network" could be an alias for or reference to the structural type. Is that possible? Is it possible to write a Nickel library that resolves imports?

yannham:
> In this proposal for Nickel I don't see a way to include an externally defined interface. It would be possible with the "dynamic scoping" option which wasn't chosen. The chosen "lexical scoping" seems to require that all field references are declared locally, reminding me of structural types instead of the nominal "I want to interact with the network configuration". Perhaps "network" could be an alias for or reference to the structural type. Is that possible?

Indeed. You can always give names/aliases to structural types, as you suggest (currently general type aliases are not supported, but it wouldn't be hard to do, and we are mostly talking about contracts here which can already be given names using let-binding).

Concerning the inclusion of a dynamic interface, one possibility is to store the external information inside a field that can be guarded by a contract that is imported dynamically, as in:

```
let moduleFoo = import "foo.ncl" in

{
  baz = if foo.bar then "enabled" else "disabled",
  foo | moduleFoo.Interface,
}
```

But indeed, you can't do that at the top-level of the record in the current form of the proposal, by design. We can imagine adding a way to do that, such as:

```
{
  baz = if bar then "enabled" else "disabled",
  ..moduleFoo.FooInterface,
}
```

This adds back dynamic scoping though, because `moduleFoo.FooInterface` is not necessarily an evaluated expression, so the interpreter can't always know in advance which fields are defined or not. This construct could require `moduleFoo.FooInterface` to be an evaluated record, but I expect that some contracts will be build using merging (like in `let Extension = Contract1 & Contract2`), so this would be limiting.

We could also just use the `..` without further ceremony to allow dynamic scoping explicitly:

```
// Error: bar is undefined
{
  foo = bar
} 

// Ok
{
  foo = bar,
  ..
}
```

> Is it possible to write a Nickel library that resolves imports?

Sorry, I'm not sure to totally understand the question.

notes/overriding.md
```diff
@@ -0,0 +1,957 @@
...
+
+### Priorities
+
+Merge is fundamentally commutative. The problem is, not all fields should be
+treated the same: some are default values that ought to be overridden, some are
+high-priority values that ought to override. The solution is to use a priority
+system, encoding the precedence of each value while retaining commutativity
+(only the priority annotation counts, not the order in which values are merged).
+
+Priorities have drawbacks. One is non-locality: the final result depends on
```
goldfirere, line 516: This reminds me of precedence of binary operators. Some languages[citation needed] support rational (as opposed to integral) precedences, so that you can always insert one more layer between two existing ones. I don't see a reason not to do so here, even if we expect most users to stick to integers.

notes/overriding.md
```diff
@@ -0,0 +1,957 @@
...
+`force` to override or be overridden by anything, without knowing the precise
+integer priority. Integer priorities still gives freedom with an infinite supply
+of levels if required.
+
+### Custom merge functions
+
+Custom merge functions would be specified by a `merge` metavalue attribute. In
+order to enforce commutativity, they would receive their arguments in an
+indistinguishable order, such as having the type `{lower: Dyn, higher: Dyn,
+priority: <Different, Equal> -> Dyn`.  If both have the same priority, the order
```
goldfirere, line 531:
```suggestion
priority: <Different, Equal>} -> Dyn`.  If both have the same priority, the order
```

?

notes/overriding.md
```diff
@@ -0,0 +1,957 @@
...
+```
+repr(r) := {
+  a = 1;
+  b = fun self => self.a + 1;
+}
+```
+
+Concretely, each field of a record may either be a thunk as usual, in the case
+of non recursive expressions like `a`, or a thunk together with the original
+expression in the recursive case. All record operations excepted merge operate
```
goldfirere, line 581:
```suggestion
expression in the recursive case. All record operations except merge operate
```

goldfirere (commented):
Just catching up here, and have run out of time to read the meatiest part of the proposal, at the end.

I do have one key question, though: why is commutativity desired? I don't think of record-update as a commutative operation, and I'm not sure why it should be. If we lose the desire for commutativity, we have new design space to work with.

notes/overriding.md
yannham: It makes even more sense as Nickel has currently only one type of numbers, and doesn't feature a separate integer type.

mboes:
#491 seems to assume this RFC is the way forward. Any reason we're keeping it from hitting `master` just yet?

blaggacao:
Out of the perceived pain with NixOS Module system commutativity, I had written a modified recursive merge that mostly does what I want / expect.

As others have noted above, I would resume that pain in that: "local reasoning about a configuration is impossible."

CUE only partially mitigates this by limiting to the priority of one, but it's still commutative, hence jumping back and forth between hops or, as others noted: just evaluate & see, trial & error - a bad strategy for a config language that should ideally build a mental map of the state of the world while _reading the source_.

My take can be scrutinized at github.com/divnix/data-merge. It mainly adds merge function annotations to deal with merge functions for non-associative arrays.

It also is monotone, a great, human-friendly simplification, which means that the RHS can never destructively modify the "data spine", where the data spine consists of the type structure + (associative or non-associative) array items. It means that the RHS cannot:

- change the type
- delete an attribute set (associative array) item
- delete a list (non-associative array) item

When reading such a configuration, and it's variants/inheritances/overlays/overrides, things can only "progress" monotonely with a felt "+" operator.

yannham:
@blaggacao thanks for your input, we'll look into it. We already had a fair amount of priority-based vs extension/override-based discussions, but it's still a valuable data point.

In the meantime, I agree with @mboes, since this PR has been open for way too long. It has received a fair number of review passes, and it's not like we have any actual RFC process yet that would make merging it irreversible or fundamentally hand-tying. I think we can move new issues, requests and discussions to new issues.

PR #333, opened 2021-03-26
Add basic benchmarks

### LINKED ISSUE #332: Benchmarks
Some aspect of the implementation of Nickel are relatively naive with respect to performance. This was fine during the experimental phase, but this needs to be addressed now that the language is consolidating. Even for obvious improvements, it's better to have at least basic benchmark to validate that upgrades actually improve performance and don't incur regressions somewhere else.

A simple benchmark suite, testing the common aspects of the evaluation of Nickel programs in order to guide performance related decisions.

yannham:
Depend on  #284 #296. Close #332.

- add various basic benchmarks, run using the [criterion](https://crates.io/crates/criterion) crate.
- add an action to automatically run them and compare them against the result of master on each PR.

[Criterion](https://crates.io/crates/criterion) is a statistical benchmarking tool. However, comparison results must be taken with prudence: I observed important, statically significant variations on my local machine between 2 consecutive runs in a relatively stable environment. It can still give us an idea about the impact of some changes.


Acaccia:
Since the benches where OOM-killed by my OS, I replaced the Bencher::iter function by Bencher::iter_batched_ref, which seems to solve the issue.

Now, there is the remaining issue of benches for records, which are OOM-killed. From my DHAT profiling, it seems the problem comes due to too many clonings of environments. I will open a new issue for it.

Avi-D-coder:
The branch https://github.com/tweag/nickel/tree/avi/benches-and-flamegraphs-master-2021-12-06 has a version of this PR up to date with master.

mboes:
Merge or close? Open since March, and the associated #332 is part of the first release milestone.

Avi-D-coder:
> Merge or close? Open since March, and the associated #332 is part of the first release milestone.

The benchmark suite (on the branch linked above) peaks at 24GB, and is not yet ready to merge. I'm currently working on locating and removing memory leaks/cycles.

mboes:
@Avi-D-coder ok nice.

> I'm currently working on locating and removing memory leaks/cycles.

ETA? Having at least *a* benchmark suite that most people can run on their rigs is already better than having none. Consider splitting out the offending benchmarks.

Avi-D-coder:
>  Consider splitting out the offending benchmarks.

Currently almost all benchmarks leak. Almost all programs leak too, but they are not run in a loop or for long enough for us to notice. Most, maybe all of the leaks are due to evaluating recursive functions.

I am hoping to have a branch without leaks early next week, but it will not be ready to merge. My solution is a cycle collector, it's mostly just the single threaded bits from my fellowship GC. That branch will be too big a change to introduce all at once. I have a second idea for minimizing the change that might be worth investigating. If it works we'll just need to use a macro to derive a `Drop` impl for each struct that's possibly cyclic in `Rc`. That might allow us to keep using `Rc`, rather than a custom `GcRc` smart pointer. The down side is it come with a performance penalty compared to `GcRc`.

Avi-D-coder:
> Having at least _a_ benchmark suite that most people can run on their rigs is already better than having none.

I can get the 24GB benchmark branch ready to merge easily. If you think it's worth it?
Not sure what counts as runable by most people, I just upgraded to a 32GB desktop, which makes me a bit behind the times ;)

mboes:
> I can get the 24GB benchmark branch ready to merge easily. If you think it's worth it?

Well, lemme lower my criteria then. ;) Even benchmarks that effectively crash most people's machine is still better than nothing. I'm sure poor sods like me with 16GB laptops can pick out a few benchmarks that need only a few GB, if they need to.

yannham:
Superseded by #529.

PR #334, opened 2021-03-29
Update README

yannham:
Update the information in the README:

 - ~remove useless information from example commands~
 - add the `cargo run` alternative
 - mention the export and the REPL commands
 - update the "run from a file" command
 - update the roadmap

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,14 +79,15 @@ decisions and trade-offs.
 
 1. Clone the repository in a local folder:
   ```
-  git clone git@github.com:tweag/nickel.git
-  cd nickel
+  $ git clone git@github.com:tweag/nickel.git
+  $ cd nickel
+  nickel$
   ```
 
 2. Install build dependencies:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,15 +79,14 @@ decisions and trade-offs.
 
 1. Clone the repository in a local folder:
   ```
-  $ git clone git@github.com:tweag/nickel.git
-  $ cd nickel
-  nickel$
+  git clone git@github.com:tweag/nickel.git
+  cd nickel
```
edolstra, line 83: Not sure whether this is an improvement. I like prefixing shell commands by `$` or `#` because it clearly separates them from the output, i.e.
```
# command
error: bla
```
is clearer than
```
command
error: bla
```

yannham: Yeah, I initially thought it can be annoying when copy-pasting (you have to select the right part of the example), but this is not very important and I agree this is less clear. Reverted in the last commit.

PR #335, opened 2021-03-31
Change precedence of let and annotations

yannham:
Depend on #317.

## The problem

Given the ML-style lets, Nickel programs or subprograms often have the following shape:

```
let id1 = val1 in
let id2 = val2 in
...
result
```

The result usually being a configuration, a common thing to do is to add a contract:

```
let Contract = { ... } in
// or let Contract = (import "contracts.ncl").SomeContract
let id = val in
...
result | #Contract
```

The problem is that currently, annotations (type or contract) bind more loosely than anything else. Thus, the following example is parsed as `(let Contract = { ... } in ...) | #Contract`, which fails at run-time because `Contract` is not in scope of the annotation. Even if one can just add parentheses `let ... in (result | #Contract)`, in our limited tests already, it comes often and is annoying. Parsing the original example directly as `let Contract = { .. } in ... in (result | #Contract)` would have exactly the same operational semantics, while bringing more things in scope, which sounds like a better way.

## Implementation

This is not totally trivial as it may first seem, because it introduces a kind of dangling else issue: `let id = val in exp` may or may not have a trailing annotation, and thus `let id = val in let id2 = val2 in exp | #Contract` is ambiguous, as well as other constructs that can have a let on the right, such as `fun x => let id = val in ... | #Contract`. LALRPOP precedence annotations are not powerful enough to solve this, but hopefully only a few top-level rule may have let on the right (`if-then-else`, `let`, `switch` and `fun`), and thanks to LALRPOP's macros, there isn't much code duplication in this PR.

## Consistency

This change may introduce some quirks:
 1. Type and contracts annotations are treated the same in the grammar. For type annotations, this change is semantically significant, as `let id = val in exp : Type` would previously typecheck `val`, while it doesn't when parsed as `let id = val in (exp : Type)`. However, one could argue that the fact that `let id = val in exp : Type` includes `val` in the statically typed block isn't a totally obvious choice either.
 2. Let bindings eagerly "eat" annotations, while other constructs don't: `(fun x => x | forall a. a -> a)` still parses as before and as expected, but if we add a let, `(fun x => let _ign = null in x | forall a. a -> a)` is parsed as `(fun x => (let ign_ = null in x | forall a. a -> a))`. This is different and will most probably fail when applied. This different treatment may be confusing.

For 1., we could parse differently type and contract annotations, but this sounds like introducing a new inconsistency.

For 2., different stances would be:
 - Only apply the proposed change to chained lets, that is `let id1 = val1 in .. let id2 = val2 in .. exp | #Contract`, but not if there's a different construct on the left, like `fun x => let id = .. in .. exp | #Contract`. There again, I'm not sure this really less confusing.
 - Treat all of `if-then-else`, `let`, `switch` and `fun` the same way. But that means `fun x => x : forall a. a -> a` now fails, because now `x` is typechecked `x` against `forall a. a -> a` instead of the whole function.

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -15,7 +15,7 @@ use codespan::FileId;
 
 grammar<'input>(src_id: FileId);
 
-SpTerm<Rule>: RichTerm = <l: @L> <t: Rule> <r: @R> => t.with_pos(mk_pos(src_id, l, r));
+WithPos<Rule>: RichTerm = <l: @L> <t: Rule> <r: @R> => t.with_pos(mk_pos(src_id, l, r));
 
 TypeAnnot: MetaValue = ":" <l: @L> <ty: Types> <r: @R> => MetaValue {
     doc: None,
@@ -70,7 +70,7 @@ LeftOpLazy<Op, Current, Previous>: RichTerm =
     <t1: Current> <op: Op> <t2: Previous> => mk_app!(Term::Op1(op, t1), t2);
 
 pub Term: RichTerm = {
-    <t: SpTerm<RootTermNoLet>> <meta: Annot?> => {
+    <t: WithPos<RootTermNoLet>> <meta: Annot?> => {
         let pos = t.pos;
 
         if let Some(mut meta) = meta {
@@ -81,7 +81,7 @@ pub Term: RichTerm = {
             t
         }
     },
-    SpTerm<RootTermLet>,
+    WithPos<RootTermLet>,
 };
 
 pub ExtendedTerm: ExtendedTerm = {
@@ -102,7 +102,7 @@ pub ExtendedTerm: ExtendedTerm = {
 }
 
 RootTermMacro<RightRule>: RichTerm = {
-    <l: @L> "fun" <ps:Pattern+> "=>" <t: SpTerm<RightRule>> <r: @R> => {
+    <l: @L> "fun" <ps:Pattern+> "=>" <t: WithPos<RightRule>> <r: @R> => {
         let pos = mk_pos(src_id, l, r);
         ps.into_iter().rev().fold(t, |t, p| RichTerm {
             term: Box::new(Term::Fun(p, t)),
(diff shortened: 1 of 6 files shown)


aspiwack:
I do believe that `let … in` should be the loosest grammar element. So I think this is going in the right direction.

Maybe before merging we may want a conversation about precedence in general. Haskell has the convention that nothing binds tighter than application (except in one case that we don't like to think about :slightly_smiling_face: ). But I don't think that it's quite right. I think that we want, in particular, from tighter to looser:

1. Record selection `u.f`
2. Application `u v`
3. Type and contract annotations `x | c`, `x : t`
4. Function binders `fun x => u`
5. Let binders `let x = u in v`

This does appear to solve the infelicities you mention. What do you think?

yannham:
>     1. Record selection `u.f`
> 
>     2. Application `u v`
> 
>     3. Type and contract annotations `x | c`, `x : t`
> 
>     4. Function binders `fun x => u`
> 
>     5. Let binders `let x = u in v`
> 
> 
> This does appear to solve the infelicities you mention. What do you think?

Sounds good. There's also the `if-then-else`. It would be between annotations and function binders, I guess.

aspiwack:
Agreed. @edolstra do you also agree?

----

How do we proceed? Is it better to merge the PR in its present state, or to modify all the precedences in one fell swoop?

yannham:
This shouldn't be a huge patch (we may have to rewrite tests or the like, but I don't think many rely on the current semantics), and the changes looks a bit interdependent, so let me add them to this PR (or close and re-open a new one)?

edolstra:
Sounds good to me. Probably easiest to update this PR.

FWIW the [Nix precedence](https://github.com/NixOS/nix/blob/906adadacd2d1c98346a2f42c0b42a32d2806d94/src/libexpr/parser.y#L325-L395) is:

* Attribute selection
* Function application
* A bunch of operators
* if-then-else
* Lambdas

yannham:
It's implemented in the last commit. I put if-then-else, lambdas, and let-binding at the same precedence level, because there is no ambiguity. I need to fix a merge conflict now, though.

aspiwack (approved):

PR #342, opened 2021-04-08
shell: workaround locking in to default.nix

573:
When supplying the repository url to nix-shell:

    nix-shell https://github/tweag/nickel/archive/master.tar.gz

It does not attempt to load the shell.nix file contained in the tarball, instead only loading default.nix, which fails if it doesn't provide a viable shell.

I would expect the behavior to be consistent with launching from the base of a repository, namely, load shell.nix if it exists, default.nix otherwise.

(see https://github.com/divnix/devos/commit/e2172c84109751f4c457cd0914437656479c1296)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,3 +1,10 @@
-(import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
+let
+  inherit (default.inputs.nixos.lib) recurseIntoAttrs;
+
+  default = (import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
   src = builtins.fetchGit ./.;
-}).defaultNix
+}).defaultNix;
+in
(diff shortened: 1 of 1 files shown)


edolstra:
This sounds like an issue with flake-compat so it should probably be fixed there.

573:
Withdrawing this as the PR would not give anything useful at the moment, apart from a build environment.
When starting the nix-shell one still has to run the `cargo` steps manually therefor also needs the repository cloned locally.

If really inclined as a workaround one can still
```
NIX_PATH=nixpkgs=http://nixos.org/channels/nixpkgs-unstable/nixexprs.tar.xz nix-shell --pure -E 'import "${fetchTarball https://github.com/tweag/nickel/archive/master.tar.gz}/shell.nix"'
```
and in the resulting nix-shell do what's written in the project's readme. still remember you'd need a clone of the repository.

Issue #337, opened 2021-04-01
Import are relative to the current path


yannham:
**Describe the bug**
When a file `file1.ncl` imports another file `file2.ncl`, the import is resolved with respect to the current working directory, instead of being relative to the location of `file1.ncl`.

**To Reproduce**
```
$mkdir test
$echo 'import "import.ncl"' > test/test.ncl
$echo "1" > test/import.ncl
$nickel -f test/test.ncl
error: Import of import.ncl failed: No such file or directory (os error 2)
  ┌─ /home/yago/Taff/Tweag/Nickel/nickel/test/test.ncl:1:1
  │
1 │ import "import.ncl"
  │ ------------------- imported here

$cd test
test$ nickel -f test.ncl
Done: Num(1.0)
```

**Expected behavior**
Executing `test.ncl` from any location succeeds.

PR #346, opened 2021-04-13
import relative to the source file path

goolord:
resolves https://github.com/tweag/nickel/issues/337

adds the following:
- add `fn get_path(&self, file_id: FileId) -> &OsStr;` to the `ImportResolver` trait. this is the least friction way to expose this as far as I can tell
- in the first `transform_pass` of `transform`, use the path from the underlying source `FileId` of `RichTerm` as the parent path, if it exists.


yannham (approved):
LGTM, thanks! I just reformatted one line in `DummyResolver` for consistency with the rest of the code.

PR #349, opened 2021-04-19
Use GitHub settings to manage project configuration in-repo

mboes:
See https://github.com/tweag/project.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -1,0 +1,40 @@
+repository:
+  has_wiki: false
+
+labels:
+  - name: "duplicate"
+    color: cfd3d7
+  - name: "good first issue"
+    color: 7057ff
+  - name: "invalid"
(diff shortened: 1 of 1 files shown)


yannham:
Do you know how do this settings interact with existing labels?

mboes:
Existing tags that are not listed in this file will be deleted. You should add the ones you want to keep to the list. Possibly using `new_name` (see https://github.com/probot/settings) to make existing labels fit a consistent nomenclature, for example to have an `area: <area>` family of labels for syntax, types, contracts, tools etc (the latter of which should probably be called "cli" for less ambiguity). See https://github.com/bazelbuild/bazel/issues for prior art.

garbas:
Oh I didn't know this exists in GitHub. This is a really nice stuff.

mboes:
It's not native: https://github.com/probot/settings.

PR #350, opened 2021-04-19
Rationalize workflows

### LINKED ISSUE #348: Rationalize Actions workflows
There are currently separate workflows for `master` and for pull requests. The `master` workflow checks more things: in particular, it runs tests against multiple Rust channels. This is not the case for the pull requests workflow. The problem is that a pull request might be green, get merged, and then cause `master` to go red, because the `master` workflow checks more things. Another problem is that the pull request workflow only triggers on pull requests. It is therefore not possible to run CI on feature branches on CI without opening a pull request first.

We should use the same CI workflow across all branches, whether on pull requests or on the master branch. This has the added benefit of less code duplication across workflow configuration files.

cc @garbas

yannham:
Close #348. Merge previous workflows `master`, `pull_request` and `specs` in two workflows `build` and `specs`. The consequences are the following:

- Nickel is now built and tested against rust stable, rust beta and rust nightly on ~~any push~~ [EDIT: only on pushing to master, see comments below] or pull request
- Makam specs are tested on ~~any push~~ [EDIT: only on pushing to master] or pull request
- Cachix is used for all workflows

Previously:
 - Nickel would only be built and tested against rust stable on pull requests
 - ~~Nickel would only be build and tested on push to master but not to other branches~~ [EDIT: this is left unchanged]
 - Cachix was used only when pushing to master.


mboes (approved):
Nice!

.github/workflows/continuous-integration.yml
```diff
@@ -1,11 +1,11 @@
-name: "Test Nickel building on master"
+name: Continuous integration
 on:
-  push:
-    branches:
-      - master
+  - push
+  - pull_request
```
wagdav, line 4: Perhaps `push` suffices here?

If you trigger both on `push` and `pull_request` you'll run the workflow twice, for example when you open a pull-request.

I believe this is why results are duplicated in this PR too:
image

mboes: Was wondering the same thing. Thanks for the info!

regnat: That's useful for external contributions though, because for these only the `pull_request` workflow will show on the PR

yannham: > That's useful for external contributions though, because for these only the pull_request workflow will show on the PR

Good point. So maybe `push: master` and `pull_request` were an OK approximation.

 @mboes , is it really a problem in practice if topic branches without associated pull requests do not trigger the CI? Anything that will be merged probably has to go through a PR eventually, and opening a Draft PR for ongoing work on a topic branch although doesn't sound like a bad practice. In that case we could go back to `push: master` and `pull_request`, avoiding double checks but still firing on external contributions.

mboes: @yannham Yeah, that's fine by me. It's a reasonable compromise. We do lose push builds that way, but CI is easy enough to reliably run on one's own machine, and as you say, draft PR's do exist.

aspiwack: > That's useful for external contributions though, because for these only the `pull_request` workflow will show on the PR

Is it though? I assume that the push Github action will run on their fork, won't it?

regnat: It will (unless they disable it I guess), but this won't show in the PR. For example compare https://github.com/NixOS/nix/pull/4511 (made from a branch in the same repo) and https://github.com/NixOS/nix/pull/4060 (made from a branch in a fork)

aspiwack: I see. It's a pity :slightly_frowning_face:

PR #354, opened 2021-05-17
Fix broken links to examples in README.md

whentze:
The examples subdirectory of this repo was moved around at some point, so the link in README.md now dangles.

This PR fixes that and also makes the link relative so it works correctly in e.g. downstream forks (I assume this would be fine since the README already uses a relative link to refer to `RATIONALE.md`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ $ cargo test
 ### Examples
 
 You can find examples in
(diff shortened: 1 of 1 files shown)


yannham (approved):
LGTM. Thanks!

PR #358, opened 2021-06-29
initial website

garbas:


regnat:
I’ve been looking at the website, and it looks like the example in https://60deda47650c2ea0d06e104a--nickel-lang.netlify.app/getting-started#reuse doesn’t work properly: The string interpolation doesn’t seem to be taken into account in the `m#"`-delimited strings:

image

(notice the `#{version}` on the right)

(it works with simple `"`-delimited strings: 

image

)

yannham:
> I’ve been looking at the website, and it looks like the example in https://60deda47650c2ea0d06e104a--nickel-lang.netlify.app/getting-started#reuse doesn’t work properly: The string interpolation doesn’t seem to be taken into account in the `m#"`-delimited strings:

Interestingly, it also works if we remove the double quotes around `#{version}`. I think you just unveiled a bug in the lexer, let me test this against master and open a proper issue.

Radvendii:
Positive things:
1. I like the variation on the theme of interactive editors where you put "here's a bug see if you can fix it" on the front page. Very engaging.
2. The site overall looks very pretty. I wasn't sure about the icons at first, but now I'm sold. I like the purple too. It's a very nice shade.
3. Fonts are nice, too.

Problems (IMO):
1. it's not immediately clear that the main page is scrollable (as in, that there's content underneath). Although, I figured it out pretty naturally, so maybe people nowadays just assume pages have more content underneath.
1. is the footer of the main page intended to be blue as opposed to purple? I don't know that it looks bad, I just noticed it and thought I'd point it out. Same goes for REPL / JSON / YAML / TOML buttons
1. On the main page, the place where you enter code is not the same height as the place where the output shows up
2. the solarized colorscheme doesn't really fit with the overall colorscheme of the website. I think the brown output area colorscheme is somewhat better, but I'm still not 100% on it. I think the command line boxes you have on the "getting started" page fit much better, and you could maybe just reuse that. If it would be helpful I can go searching through colorschemes and try to find a few I think look better.
3. I would also add some padding on the top and bottom of the code in that box. It goes right up to the edge.
4. `[devops@nickel]` seems kind of weird and arbitrary. This isn't a huge deal, but I would maybe put just `$` or `$(pwd) $`. This would also make that take up less space, leaving more room for the content people care about
1. there are a few random words that are coloured seemingly random colours in the command line sections of the "getting started" page.
1. it doesn't scale well to small window widths, and the menu button that appears doesn't work (Also that menu botton appears at way larger sizes than it needs to)

yannham:
> Problems (IMO):
> 
>     1. it's not immediately clear that the main page is scrollable (as in, that there's content underneath). Although, I figured it out pretty naturally, so maybe people nowadays just assume pages have more content underneath.
Ok, I can try to add the kind of floating double arrow >> at the bottom.

> 
>     2. is the footer of the main page intended to be blue as opposed to purple? I don't know that it looks bad, I just noticed it and thought I'd point it out. Same goes for REPL / JSON / YAML / TOML buttons
Yes, it's two colors that were proposed during the design, so they should belong to the same color palette.

> 
>     3. On the main page, the place where you enter code is not the same height as the place where the output shows up
Ah, I experienced this but only when I load the page in a size smaller than `lg`, and then make it bigger. Could you confirm that the problem doesn't happen if you load the page at max size from the beginning?

>     4. the solarized colorscheme doesn't really fit with the overall colorscheme of the website. I think the brown output area colorscheme is somewhat better, but I'm still not 100% on it. I think the command line boxes you have on the "getting started" page fit much better, and you could maybe just reuse that. If it would be helpful I can go searching through colorschemes and try to find a few I think look better.
Yeah, the colorscheme of the editor, the output widget and the code blocks were chosen before the pink and blue of the main theme. To add to the problem, the three colorscheme are from different components, they currently don't match and I don't know if we can easily transfer a colorscheme from one to the other widget (the editor is based an ace.js, the output widget is home-made and use solarized ansi colors, and the code box on the getting started page are based on Prism.js). I'll take a look though, we can probably do better.

>     5. I would also add some padding on the top and bottom of the code in that box. It goes right up to the edge.

Noted.

> 
>     6. `[devops@nickel]` seems kind of weird and arbitrary. This isn't a huge deal, but I would maybe put just `$` or `$(pwd) $`. This would also make that take up less space, leaving more room for the content people care about

This is something handled by Prism automatically. I think it's nice because:
1. It clearly shows which line is an output and which line is an input
2. It still lets you select only the code
3. 
I can try without the host name, if it's really too long.
 
>     7. there are a few random words that are coloured seemingly random colours in the command line sections of the "getting started" page.
Yeah, it's handled by the JS highlighting library. It's not super important but I agree it looks a bit random.

>     8. it doesn't scale well to small window widths, and the menu button that appears doesn't work (Also that menu botton appears at way larger sizes than it needs to)
Right. I always have this menu problem on bootstrap, when you copy paste the canonical example it doesn't work out of the box, which is annoying. The fix is simple though, I'll do it. Otherwise, what doesn't scale well in particular?

About the menu button, the thing is that between `md` and `lg` size, there is not enough space for the menu to be displayed on one line, which I found not very nice, so I set up the breakpoint at `lg`. I can change my mind if it's not ideal, though.

Radvendii:
> Yes, it's two colors that were proposed during the design, so they should belong to the same color palette.
:+1: 

> > 3. On the main page, the place where you enter code is not the same height as the place where the output shows up
> Ah, I experienced this but only when I load the page in a size smaller than `lg`, and then make it bigger. Could you confirm that the problem doesn't happen if you load the page at max size from the beginning?
I'm not sure what `lg` means, but when I just open up the page (full size of my screen), I get this:

image

> Yeah, the colorscheme of the editor, the output widget and the code blocks were chosen before the pink and blue of the main theme. To add to the problem, the three colorscheme are from different components, they currently don't match and I don't know if we can easily transfer a colorscheme from one to the other widget (the editor is based an ace.js, the output widget is home-made and use solarized ansi colors, and the code box on the getting started page are based on Prism.js). I'll take a look though, we can probably do better.

Ah, I see how the different frameworks would make a mess of things. Are the colorschemes of all of these widgets not easily customizeable? That's a shame.

> > ```
> > 6. `[devops@nickel]` seems kind of weird and arbitrary. This isn't a huge deal, but I would maybe put just `$` or `$(pwd) $`. This would also make that take up less space, leaving more room for the content people care about
> > ```
> 
> This is something handled by Prism automatically. I think it's nice because:
> 
> 1. It clearly shows which line is an output and which line is an input
> 2. It still lets you select only the code
> 
> I can try without the host name, if it's really too long.

I don't think it's really a problem. Just for me it seems slightly... unpolished? indie? Like someone copy-pasted their command line into the website and happened to have the username "devops" and the hostname "nickel".

> Otherwise, what doesn't scale well in particular?
image
image

Also the "find the error" section just disappears. Which I guess is fine on phone-sized screens, but I'm just using half-screen-width on my laptop.

> About the menu button, the thing is that between `md` and `lg` size, there is not enough space for the menu to be displayed on one line, which I found not very nice, so I set up the breakpoint at `lg`. I can change my mind if it's not ideal, though.
image
image
It seems like there's still plenty of room there

yannham:
All of @Radvendii's issues should be addressed, at the exception of the color schemes.

Fizzixnerd:
(xpost from basecamp because I am silly)

I tried to get a screen capture, but I couldn't, probably because I'm using Wayland.  But in NixOS on chrome, the rendering of the playground doesn't work properly.  I can click on the "$" in the version, and try to delete it, and it deletes stuff like 8 characters to the left instead.

tanyabouman:
> Yes, it's two colors that were proposed during the design, so they should belong to the same color palette.
In my opinion, both the blue and purple look bad, because they're too light compared to white.  The poor contrast makes it difficult to read.  https://contrastchecker.com

Radvendii:
Two more comments on the latest version at half-laptop-screen-width:
1. The playground gets kind of squished and overly line-wrap-y. I wonder if for smaller screen widths it makes sense to stack them instead of side-by-side.
1. The 3 items at the bottom of the landing page sort of end up in an awkward configuration. With a smaller screen size, they're fine, and with a larger screen size they're fine. It's just right in the middle. I think the easiest way to deal with this would be to just have them split into three separate lines earlier on (go straight from 1 line to 3). Though I don't know if that will look the best.
image

(Ignore the weird background colour my browser tries to make everything dark mode)

tanyabouman:
About the Getting Started page:
> This program is composed of record.

This program is composed of records.
> multiline strings are useful for:

The lines should each start with "Writing", not "Write".

For this Nickel:
```
  test = m#"test.sh --option --install #{name} --version "#{version}""#m
```
I get this yaml:
```
  test: "test.sh --option --install example --version \"#{version}\""
```
It seems like the extra quotes need to be removed:
```
  test = m#"test.sh --option --install #{name} --version #{version}"#m,
```

tanyabouman:
Oh wait, the unexpected results from the string interpolation looks like a [known issue](https://github.com/tweag/nickel/issues/361).

yannham:
> In my opinion, both the blue and purple look bad, because they're too light compared to white. The poor contrast makes it difficult to read. https://contrastchecker.com

What do you find hard in particular, white on blue like in the footer or the selected mode of the playground's output, or the purple of links? both?

tanyabouman:
> > In my opinion, both the blue and purple look bad, because they're too light compared to white. The poor contrast makes it difficult to read. https://contrastchecker.com
> 
> What do you find hard in particular, white on blue like in the footer or the selected mode of the playground's output, or the purple of links? both?

A lot of it has poor contrast.

Contrast ratios for a few combinations:
Purple on white: 2.35:1
Blue on white: 1.81:1
Pink/red on white: 3.81:1

The minimum is supposed to be 4.5:1.
https://www.w3.org/TR/WCAG21/#contrast-minimum
(An easy way to find these errors is by using [WAVE](https://wave.webaim.org)).

yannham:
The last commits improve the contrast: no more white on blue/mauve (use black instead), darken a bit the text color. I tried to find a different matching color for links but this is not easy tbh: if it's too contrasted with the white background, it's poorly contrasted with the black text and it is not obvious anymore that this is a link and not just normal text.

In the end I found the darkened versions of the colors already in use don't feel too odd with the theme while being compliant. 

By the way, TIL than bootstrap's default colors are not all compliant (e.g. `text-primary` or `text-danger`).

tanyabouman:
That's a good point about contrast with the rest of the text.  [Underlining the links](https://www.w3.org/TR/WCAG20-TECHS/G183.html) would help as another way of distinguishing them.

yannham (approved):
I think it's now ready to publish.

A recent commit upgraded the vendored wasm repl to include the fix of the interpolation bug found by @regnat (#361) contained in #363.

I agree with @Radvendii that colorschemes should be uniform, but it's not a hard blocker, so let us put that in an issue. Same goes for the two-columns playground on small width.

If nobody can reproduce the problem of @Fizzixnerd unfortunately, it'll also go in an issue. We are using the [ace editor](https://github.com/ajaxorg/ace), which seems well-known and well supported. I'm not even sure we can do something at the website's level.

yannham:
I'm just fixing a last problem with quotes escaping in the serialized output of the playground, and I'll merge.

06kellyjac:
Very nice, what's the official URL?

yannham:
> Very nice, what's the official URL?

It will be available at https://nickel-lang.org, but I need @garbas's access rights to netlify in order to deploy it. At worst, it will be online tomorrow :)

PR #370, opened 2021-08-02
Grammar: use LALRPOP precedence annotations

yannham:
Simplify the definition of the grammar, and make it more easily extensible by using LALRPOP's precedence annotations (https://github.com/lalrpop/lalrpop/pull/555).


yannham:
FYI: the CI is only failing on nightly, due to the following issue https://githubmemory.com/repo/crossbeam-rs/crossbeam/issues/678. We can ignore it, as the best fix is to update the nightly version of the compiler, which will eventually land in Nixpkgs/import-cargo/whatever inputs is bringing the nightly version of rust at some point.

edolstra (approved):

PR #374, opened 2021-08-05
Closed contracts

### LINKED ISSUE #303: Closed record contracts
The preferred way of defining record contracts is to define vanilla records with meta-values, and merge them with the data to validate. #302 aims at changing this, making the pipe syntax the preferred way of enforcing contracts, for consistency with enforcing other type of contracts and better composability with builtin types (see #294 for more context). But in essence, the underlying implementation remains the same: merging.

This means that contracts defined as such are open, in the sense that they allow additional fields to sneak in: `{a = 1; b = "a"} & {a | Num}` won't raise blame, for example. Closed contracts can still be defined using row types: `{a = 1; b = "a"} | {a: Num}` *will* raise blame, but this is neither the preferred nor the more ergonomic way of defining data schema.

Thus, a direct way of specifying closed contracts as records with meta-values would be welcome. The CUE language, which Nickel's merge system is inspired from, has the same issue. CUE offers two ways of defining closed contracts (*closed structs* in CUE's terminology): [using explicitly the `close` keyword](https://cuelang.org/docs/tutorials/tour/types/closed/), or [using definitions](https://cuelang.org/docs/tutorials/tour/types/defs/).

I see two different directions to pursue in Nickel:

1. **Make the pipe operator and merge operator behave differently**: using the pipe operator aims to be the canonical way of enforcing contracts. When enforcing a contract, it seems more natural to have it closed by default. A solution is thus to decide than when using the pipe operator, contracts as records are considered closed by default. In return, it may be then necessary to re-introduce a way of specifying open contracts. Although we could also decide that to have open contracts one needs to resort to merging, we would get back the issues #302 was trying to solve in the first place for open contracts.
2. **Have a specific definition for contracts**: this corresponds to CUE's *definitions*. Have a specific definition for contracts, using e.g. the `#` character and uppercase letter: `let #Contract = {..}`. These would be closed by default, and could be made explicitly open using say an ellipsis: `{a | Num; ..}`, or an underscore `{a | Num; _}`. We can even go further and require that anything else than a function appearing in the RHS of the pipe operator (`stuff | #Contract`) must be defined this way, so that standard records wouldn't work. This would make a better distinction between record contracts and normal records (even if this distinction isn't really relevant operationally, it might be clearer conceptually).

yannham:
Depend on #370. Fix #303.

## Summary

This PR implements the fix proposed in #303: it makes the contract application operator `let Contract = { a | Num } in value | #Contract` more restrictive than just `value & Contract`, in that it now raises a contract violation if `value` has an extra field other than `a`.

Contract application having a different behavior than merge for record contracts is nice, because beside contract application now doing the correct thing (by default a schema shouldn't allow extra fields), one can still use standard merging on closed contracts to compose them:

```
let ContractA = {a | Num} in
let ContractB = {b | Num} in
let Contract= ContractA & ContractB in
{a = 1, b = 2} | #Contract
```
## Open contracts

This PR introduces a new ellipsis syntax `..` to stipulate that `Contract` allows extra field, as in the working `let Contract = {a | Num, ..} in {a = 1, b = 2} | #Contract` . With the ellipsis, a record contract behaves as before this PR.

The behavior with respect to merging is that if any of the operands is open, then the resulting contract is open. The rationale is that `..` can be imagined as a special field that is syntactically merged as any other field: `{a | Num} & {..} = {a | Num, ..}`.

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -739,8 +739,6 @@ pub enum BinaryOp {
     ListElemAt(),
     /// The merge operator (see the [merge module](../merge/index.html)).
     Merge(),
-    /// The merge operator in contract mode (see the [merge module](../merge/index.html)).
-    MergeContract(),
 
     /// Hash a string.
     Hash(),
@@ -779,17 +777,27 @@ pub enum NAryOp {
     StrReplaceRegex(),
     /// Return a substring of an original string.
     StrSubstr(),
+    /// The merge operator in contract mode (see the [merge module](../merge/index.html)). The
+    /// arguments are in order the contract's label, the value to check, and the contract as a
+    /// record.
+    MergeContract(),
 }
 
 impl NAryOp {
     pub fn arity(&self) -> usize {
         match self {
-            NAryOp::StrReplace() | NAryOp::StrReplaceRegex() | NAryOp::StrSubstr() => 3,
+            NAryOp::StrReplace()
+            | NAryOp::StrReplaceRegex()
+            | NAryOp::StrSubstr()
+            | NAryOp::MergeContract() => 3,
         }
     }
 
     pub fn is_strict(&self) -> bool {
-        true
+        match self {
+            NAryOp::MergeContract() => false,
+            _ => true,
+        }
     }
 }
 
@@ -799,6 +807,7 @@ impl fmt::Display for NAryOp {
             NAryOp::StrReplace() => write!(f, "strReplace"),
             NAryOp::StrReplaceRegex() => write!(f, "strReplaceRegex"),
             NAryOp::StrSubstr() => write!(f, "substring"),
+            NAryOp::MergeContract() => write!(f, "mergeContract"),
         }
     }
 }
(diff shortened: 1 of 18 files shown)


src/term.rs
```diff
@@ -147,6 +146,25 @@ pub enum Term {
...
+    }
+}
+
+impl RecordAttrs {
+    pub fn merge(attrs1: RecordAttrs, attrs2: RecordAttrs) -> RecordAttrs {
+        RecordAttrs {
+            open: attrs1.open || attrs2.open,
+        }
+    }
+}
```
regnat, line 166: (nit) This looks a lot like a monoid. Isn’t there a standard trait for that in Rust?

regnat (approved):
I really like this.

Just a very minor comment that you can feel free to ignore, but otherwise :+1:

src/term.rs
yannham: Good point (and typical Haskeller :stuck_out_tongue: ), but I can't find one in the standard library, just in some third-party crates, so I guess I'll leave it at is it.

PR #376, opened 2021-08-06
Fix dynamic fields not having access to recursive fields

### LINKED ISSUE #298: Record fields defined via interpolation are not recursive
**Describe the bug**
Declaring a record field using an interpolated expression makes the content of this field "non-recursive", meaning that it can't refer to other fields of the same records, as well as other fields can't refer to it.

**To Reproduce**
```
nickel> { x = 1; y = x + 1}.y
2

nickel> { x = 1; $"y" = x + 1}.y
error: Unbound identifier
  ┌─ repl-input-1:1:17
  │
1 │ { x = 1; $"y" = x + 1}.y
  │                 ^ this identifier is unbound
```

**Expected behavior**
Field defined via interpolation should have the same recursive behavior as other fields.

**Additional context**
This is due to the way such fields are elaborated: the previous example is rewritten as `{x = 1}$["y" = x + 1]` which makes clear why `y` doesn't have access to `x`.

yannham:
Depend on #374. Fix #298.

Previously, the parser was elaborating a record with fields defined dynamically as a static record followed by a sequence of dynamic extension:

```
let x = "foo" in
let y = "bar" in
{ "#{x}" = 1, "#{y}" = 2, baz = 3}

// Rewritten at parsing time as
{baz = 3} $[ x = 1] $[ y = 2 ]
```

Where `$[ ]` is the dynamic extension operator. This causes #298: because dynamic fields are separated early from their original record, they don't have access to recursive fields like `baz`

This PR fixes the problem by changing the representation of recursive records to a set of static field and a set of dynamic fields. Doing so, the dynamic fields are kept together with their original record until the first evaluation of the `RecRecord` to a standard `Record`, and can thus have access to the other recursive fields.

## Recursive dynamic fields

Only field defined statically can be recursive, that is this PR doesn't support:

```
let x = "foo" in {"#{x}" = 0, bar = foo +1} //Error
```

Doing otherwise would require to pre-evaluate some terms before we can even typecheck, and would be dubious in any case, introducing a form of dynamic scoping.

## Typechecking

If a record has a dynamic field, then it is inferred to have a type of the form `{_ : _a}` for some unification variable `_a`, which is the only type that can handle statically unknown fields.

### DIFF
--- a/tests/pass/record-defs.ncl
+++ b/tests/pass/record-defs.ncl
@@ -29,7 +29,7 @@ let Assert = fun l x => x || %blame% l in
 ({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
   == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
   | #Assert) &&
-({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
+({foo."bar"."baz" = 1, "bar"."baz"."foo" = foo.bar."baz" + 1}
   == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
   | #Assert) &&
 
--- a/tests/pass/record-defs.ncl
+++ b/tests/pass/record-defs.ncl
@@ -29,6 +29,10 @@ let Assert = fun l x => x || %blame% l in
 ({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
   == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
   | #Assert) &&
+({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
+  == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
+  | #Assert) &&
+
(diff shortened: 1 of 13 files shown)


tests/pass/record-defs.ncl
```diff
@@ -29,6 +29,10 @@ let Assert = fun l x => x || %blame% l in
 ({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
   == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
   | #Assert) &&
+({foo.bar.baz = 1, bar.baz.foo = foo.bar.baz + 1}
+  == {foo = {bar = {baz = 1}}, bar = {baz = {foo = 2}}}
+  | #Assert) &&
```
edolstra, line 34: This looks identical to the test above it?

edolstra (approved):

tests/pass/record-defs.ncl
yannham: Right, looks like a conflict resolution glitch or a bad copy/paste. Let me get rid of it

PR #377, opened 2021-08-10
Minor improvements to the README

sir4ur0n:
* Fix typos
* Fix 4th element list display
* Add the shell hint on code blocks

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,20 +84,20 @@ abstractions or just feel ad hoc. Nickel buys you more for less.
      and extra `--` as in `cargo run -- -f program.ncl`.
 
 2. Run your first program:
-  ```
+  ```shell
   $ ./nickel <<< 'let x = 2 in x + x'
   Typechecked: Ok(Types(Dyn))
   Done: Num(4.0)
   ```
   Or load it from a file:
-  ```
+  ```shell
   $ echo 'let s = "world" in "Hello, " ++ s' > program.ncl
   $ ./nickel -f program.ncl
   Typechecked: Ok(Types(Dyn))
   Done: Str("Hello, world")
   ```
-3. Start an REPL:
-  ```
+3. Start a REPL:
+  ```shell
   $ ./nickel repl
   nickel> let x = 2 in x + x
   nickel> x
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,50 +124,50 @@ for help about a specific subcommand.
...
+  ```shell
   $ cargo doc --no-deps
   ```
 2. Open the file `target/doc/nickel/index.html` in your browser.
 
 ### Examples
 
 You can find examples in
-[`examples`](./examples). Note
+[`/examples`](./examples) directory. Note
```
gkaracha, line 163: I think it was  better before, but even if it is to be changed I think we need the dot here:
```suggestion
[`./examples`](./examples) directory. Note
```

sir4ur0n: I find it weird-looking to be both an inline code (backticks) AND a link:
image

I'd go with:
* either don't make it inline code, i.e. only write [examples](./examples)
* or make it explicit it refers to the directory (in which case I agree with your suggestion of adding the dot), i.e. [`./examples`](./examples)

gkaracha: Yeah, I see. It's fine either way for me; let's see what @yannham prefers :slightly_smiling_face:

yannham (changes requested):
Thanks for the fix! I'm not sure about the `shell` hint, which I think is for shell scripts, not sessions (it randomly colors subcommand like `test` in `cargo test` because it is a shell command while `cargo doc` is not colored, it colors the `let` in the REPL, and so on). I think the `console` hint behaves more like a session but I've never really used it, maybe should we give it a try?

README.md
```diff
@@ -123,50 +124,50 @@ for help about a specific subcommand.
...
+  ```shell
   $ cargo doc --no-deps
   ```
 2. Open the file `target/doc/nickel/index.html` in your browser.
 
 ### Examples
 
 You can find examples in
-[`examples`](./examples). Note
+[`./examples`](./examples) directory. Note
```
yannham, line 163:
```suggestion
the [`./examples`](./examples) directory. Note
```

sir4ur0n:
> Thanks for the fix! I'm not sure about the `shell` hint, which I think is for shell scripts, not sessions (it randomly colors subcommand like `test` in `cargo test` because it is a shell command while `cargo doc` is not colored, it colors the `let` in the REPL, and so on). I think the `console` hint behaves more like a session but I've never really used it, maybe should we give it a try?

@yannham done in last commit, you can see the result in https://github.com/tweag/nickel/blob/e38f53a60d14d9100c2e8328ba8a49f10c1fbdbe/README.md

yannham (approved):

PR #380, opened 2021-08-24
Add :p(rint) command to REPL

### LINKED ISSUE #307: REPL: Recursively print records

Currently the REPL doesn't print the values of record fields:
```
nickel> { x = 1; y.z = 2; }
{ ... }
```

A command like `nix repl`'s `:p` to recursively print the contents, e.g.

```
nix-repl> :p { x = 1; y.z = 2; }
{ x = 1; y = { z = 2; }; }
```

ErinvanderVeen:
This alternative command fully recursively evaluates the record/list
before it is printed. To achieve this, a REPL.eval_full function is
added to the trait. For REPLImpl, a generic eval function is added that
takes the specific evaluation function as an argument.

Closes #307

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -9,6 +9,7 @@
 use crate::cache::Cache;
 use crate::error::{Error, EvalError, IOError};
 use crate::error::{ParseError, REPLError};
+use crate::eval::Environment;
 use crate::identifier::Ident;
 use crate::parser::{grammar, lexer, ExtendedTerm};
 use crate::term::{RichTerm, Term};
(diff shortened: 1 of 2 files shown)


yannham (changes requested):
Looks great, but the `print` command does not show in the list of available commands when typing `:help` or when entering an unknown command in the REPL. The list of available commands should probably not be hardcoded in a string like it is currently, but for the time being let us just add the print command to this string and think of a better way later

yannham (approved):

Issue #367, opened 2021-07-23
Performance improvement: Environment representation


Acaccia:
**Is your feature request related to a problem? Please describe.**
Using the benchmarks developed in the context of issue #332 and PR #333 , I generated flamegraphs for each of those benchmarks. On most of those, the time is spent for the majority of the Nickel evaluation on the function _Thunk::into\_closure_, which induces a huge amount of thunk cloning, and cloning of Hashmaps in _eval_closures_. Both of these seems related to the cloning of Environment.

**Describe the solution you'd like**
Environment could be a chained list of environments, with the outer ones shared and creating a new environment would add a new one on head of the list, something like 
```rust
struct Environment {
  current: HashMap<Ident, Thunk>,
  previous: Option<Rc<RefCell<Environment>>>,
}
```
This should make the cloning cheaper, by sharing the outer environments. 
However, it will make the lookup more expensive, so this solution should be benchmarked and compared with the actual implementation.

**Additional context**
The flamegraphs can be generated from the branch *benches_and_flamegraphs*. just use `cargo bench -- --profile-time=<time>` to generate these. For my flamegraps, I used `time = 5`.
(I do not think this branch will be merged, since it adds a new dependency on pprof and changes the criterion profiler)

This issue can be seen from flamegraphs of benches like _functions/church 3_ or _lists/generate normal|deepseq 30_, _lists/map normal|deepseq 30_, ...

![church_3](https://user-images.githubusercontent.com/4235083/126808574-0bb747f0-10d5-4f76-8b9f-131667ad1de6.png)
![generate_deepseq_30](https://user-images.githubusercontent.com/4235083/126808579-c3225967-a71b-45a5-9d55-d7c22bc2a26f.png)
![map_deepseq_30](https://user-images.githubusercontent.com/4235083/126808682-72ed91be-1fea-4835-876d-b0aa2aa6f37d.png)

Acaccia:
Also, using a DHAT profiling on this simple example:
```
let run = fun s =>
   let updateDict = fun dict char =>
     if records.hasField char dict then
       dict -$ char & {"#{char}" = dict."#{char}" + 1}
     else
       dict & {"#{char}" = 1} in
   lists.foldl updateDict {} (strings.chars s) in

[...]

PR #381, opened 2021-08-25
Environment refactoring

Acaccia:
See issue #367

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -2,6 +2,7 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::hash::Hash;
 use std::marker::PhantomData;
+use std::ptr::NonNull;
 use std::rc::Rc;
 
 #[derive(Debug, PartialEq, Default)]
@@ -61,6 +62,17 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
         }
     }
 
+    pub fn iter_elems(&self) -> EnvElemIter<'_, K, V> {
+        let mut env: Vec<NonNull<HashMap<K, V>>> = self
+            .iter_layers()
+            // SAFETY: all NonNull::new_unchecked comes from pointers created from Rc, so cannot be null
+            .map(|hmap| unsafe { NonNull::new_unchecked(Rc::as_ptr(&hmap) as *mut HashMap<K, V>) })
+            .collect();
+        // SAFETY: by design, env cannot be empty, and coming from an Rc, it is well aligned and initialized
+        let current_map = unsafe { env.pop().unwrap().as_ref() }.iter();
+        EnvElemIter { env, current_map }
+    }
+
     fn was_cloned(&self) -> bool {
         Rc::strong_count(&self.current) > 1
     }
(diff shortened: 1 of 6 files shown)


Acaccia:
The new environment now has severely improved the performances for all the benchmarks, except for the "serialization" one that still does not run. This time, it is not due to a OOM kill, but it seems the representation of records makes it suboptimal for the usage tested in the benchmarks, and the process runs for a very long time, and is ultimately killed.

Anyway, in the attached reports, if you open "report/index.html", you'll see how this new environment representation improves the situation compared to the previous one (commit 58d6e8171120ca6a93706a67908b1946a7db1fc5 vs commit c67b4a7879355a4e9e1822745dc73006c10020e1 in master)

[criterion.tar.gz](https://github.com/tweag/nickel/files/7099341/criterion.tar.gz)

Now there is still documentation to write before merging.

src/environment.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Creates an iterator that visits all elements from the Environment, from the oldest layer
+    /// to the most recent one. It uses this order, so calling `collect` on this iterator to create
+    /// a hashmap would have the same values as the Environment.
+    /// The element iterator type is `(&'env K, &'env V)`, with `'env` being the lifetime of the Environment.
+    pub fn iter_elems(&self) -> EnvElemIter<'_, K, V> {
+        let mut env: Vec<NonNull<HashMap<K, V>>> = self
+            .iter_layers()
+            // SAFETY: all NonNull::new_unchecked comes from pointers created from Rc, so cannot be null
+            .map(|hmap| unsafe { NonNull::new_unchecked(Rc::as_ptr(&hmap) as *mut _) })
+            .collect();
```
yannham, line 109: Why do we need to build a vector here? Can't we just pass directly the mapped `EnvLayerIter` to the element iterator, so that it can fetch the next layer only when needed? Although this is not very important.

src/environment.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    /// Creates an iterator that visits all elements from the Environment, from the current layer to the oldest one.
+    /// If values are present multiple times, only the most recent one appears.
+    /// [`iter_elems`] should be preferred, since it does not need to create an intermediary hashmap.
+    /// The element iterator type is `(&'env K, &'env V)`, with `'env` being the lifetime of the Environment.
+    ///
+    /// [`iter_elems`]: Environment::iter_elems
+    ///
+    pub fn iter(&self) -> EnvIter<'_, K, V> {
```
yannham, line 122: Is there a reason to have the `EnvIter` wrapper instead of a directly returning the underlying `hash_map::IntoIter<&'a K, &'a V>` here?

src/environment.rs
```diff
@@ -0,0 +1,364 @@
...
+        self.collapsed_map.next()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
+        pub fn deepness(&self) -> usize {
```
yannham, line 231: `depth` is more common I think (deepness is the quality of being deep, while depth is the quantity)

yannham (approved):
Looks great! I just made suggestions for a few typos and have a some questions here and there, but nothing blocking. In the end the "2-step clone trick" doesn't look too bad or too complex.

src/environment.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub fn new() -> Self {
+        Self {
+            current: Rc::new(HashMap::new()),
+            previous: RefCell::new(None),
+        }
+    }
+
+    /// Inserts a key-value pair into the Environment.
+    /// If the key was present in the _current_ Environment, the previous value is returned.
+    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
```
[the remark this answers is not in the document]
Acaccia: I actually agree. I chose this because it would be "consistent" with what HashMap::insert does, but it remains confusing.
Let's return Nothing.

src/environment.rs
Acaccia: It's always about being consistent. All the other kind of iterators have their own structures, so does this one. Ultimately, the compiler should remove it and directly use HashMap::intoiter.

src/environment.rs
Acaccia: Nope, we can't, since EnvLayerIter is not a DoubleEndedIterator. Here, during the iteration process, we access the vec frow end to start.

src/environment.rs
Acaccia: Also, `warning: bounds on generic parameters are not enforced in type aliases`

Issue #321, opened 2021-03-10
Standard library stabilization


yannham:
#254 describes the next batch of functions to add to the standard library to have a minimum viable one. While it's currently not yet closed, the implementation should end soon. This issue tracks the second phase of tasks to achieve with respect to the standard library.

- [ ] Write tests
- [x] Write documentation
- [ ] Stabilize naming, make it consistent
- [ ] Stabilize types, make them consistent
- [ ] Stabilize provided contracts, make their names and implementation consistent

silverraven691:
Now that we have `|>`, it feels like `strings.join` should be `flip`ped to `Str -> List Str -> Str`

silverraven691:
Hopefully, this is not off-topic, here's a bunch of functions I ended up re-implementing while writing some Nickel, and more that I think I may end up wanting:

```
nums.clamp : Num -> Num -> Num -> Num
functions.flip : forall a b c. (a -> b -> c) -> b -> a -> c // Although I don't suppose I'll use it as much after my PR
lists.optional : Bool -> List -> List
records.optional : Bool -> { _: Dyn } -> { _: Dyn }
records.filter : (Str -> Dyn -> Bool) -> { _: Dyn } -> { _: Dyn }
records.toList : { _: Dyn } -> List { key: Str, value: Dyn } // or: records.entriesOf
records.fromList : List { key: Str, value: Dyn } -> { _: Dyn } // or: records.fromEntries
```

yannham:
This issue is supeseded by #935

PR #383, opened 2021-08-26
Document stdlib; add a few missing functions; consistent naming

ErinvanderVeen:
Related to  #321

### DIFF
--- a/stdlib/lists.ncl
+++ b/stdlib/lists.ncl
@@ -99,16 +99,16 @@
 
         For example:
         ```nickel
-          foldl (fun acc e => acc @ [e]) [] [ 1, 2, 3 ] =>
-            [ 3, 2, 1 ]
+          foldl (fun acc e => acc + e) 0 [ 1, 2, 3 ] =>
+            6
         ```
         "#m
-      = fun f fst l =>
+      = fun f acc l =>
         if %length% l == 0 then
-          fst
+          acc
         else
-          let rest = foldl f fst (%tail% l) in
-          %seq% rest (f rest (%head% l)),
+          let newAcc = f acc (%head% l) in
+          %seq% newAcc (foldl f newAcc (%tail% l)),
 
     fold : forall a b. (a -> b -> b) -> List a -> b -> b 
       | doc m#"
@@ -116,8 +116,8 @@
 
         For example:
         ```nickel
-          fold (fun e acc => e + acc) [ 1, 2, 3 ] 0 =>
-            6
+          fold (fun e acc => acc @ [e]) [ 1, 2, 3 ] [] =>
+            [ 3, 2, 1 ]
         ```
         "#m
       = fun f l fst =>
@@ -160,7 +160,7 @@
             [ 3, 2, 1 ]
         ```
         "#m
-      = fun pred l => reverse (fold (fun x acc => if pred x then acc @ [x] else acc) l []),
+      = fun pred l => fold (fun x acc => if pred x then acc @ [x] else acc) l [],
 
     flatten : forall a. List (List a) -> List a
       | doc m#"
--- a/stdlib/lists.ncl
+++ b/stdlib/lists.ncl
@@ -95,11 +95,15 @@
 
     foldl : forall a b. (a -> b -> a) -> a -> List b -> a
       | doc m#"
-        Fold a left-associative function over a list.
+        Fold a function over a list.
+        `foldl f init [x1, x2, ..., xn]` results in `f (... (f (f init x1) x2) ...) xn`.
+
+        This function is strict in the intermediate accumulator.
 
         For example:
         ```nickel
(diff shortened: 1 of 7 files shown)


stdlib/strings.ncl
```diff
@@ -51,43 +121,305 @@
...
+      "#m
+    = fun n => %toStr% n,
 
-    fromNum | Num -> Str = fun n => %toStr% n,
     // fromEnum | < | Dyn> -> Str = fun tag => %toStr% tag,
-    fromEnum | Dyn -> Str = fun tag => %toStr% tag,
-    fromBool | Bool -> Str = fun b => %toStr% b,
+    fromEnum | Dyn -> Str
+    | doc m#"
+      Converts any basic value to its string representation.
```
yannham, line 363: It's technically true, but it's an implementation detail: this function is for enums only. We should probably make it an error to use it with anything else than enum by using a more specific contract for the input. For the time being, let's just lie in the documentation :stuck_out_tongue:

stdlib/strings.ncl
```diff
@@ -51,43 +121,305 @@
...
+      ```nickel
+        fromBool true =>
+          "true"
+      ```
+      "#m
+    = fun b => %toStr% b,
+
+    toNum | #NumLiteral -> Num
+    | doc m#"
+      Convers a string that represents an integer to that integer.
```
yannham, line 387:
```suggestion
      Converts a string that represents an integer to that integer.
```

stdlib/strings.ncl
```diff
@@ -51,43 +121,305 @@
...
+      ```nickel
+        toNum "123" =>
+          123
+      ```
+      "#m
+    = fun s => %numFromStr% s,
+
+    toBool | #BoolLiteral -> Bool
+    | doc m#"
+      Converts a string that represents a bool to that bool.
```
yannham, line 399:
```suggestion
      Converts a string that represents a boolean to that boolean.
```

stdlib/strings.ncl
```diff
@@ -51,43 +121,305 @@
...
+      "#m
+    = fun s => s == "true",
 
-    toNum | #NumLiteral -> Num = fun s => %numFromStr% s,
-    toBool | #BoolLiteral -> Bool = fun s => s == "true",
     // toEnum | #Ident -> < | Dyn> = fun s => %enumFromStr% s,
-    toEnum | #Ident -> Dyn = fun s => %enumFromStr% s,
+    toEnum | #Ident -> Dyn
+    | doc m#"
+      Converts any string that represents an identifier to an enum variant.
```
yannham, line 415:
```suggestion
      Converts any string that represents an enum variant to that enum variant.
```

The thing is that valid enum variants and valid identifiers are the same thing. I agree it's not clear to have this contract named `Ident`. Maybe we should rename it to `EnumVariant`, `Enum` or `EnumTag` as it seems to be actually used as such.

stdlib/lists.ncl
```diff
@@ -1,76 +1,263 @@
...
+        ```nickel
+          concat [ 1, 2, 3 ] [ 4, 5, 6 ] =>
+            [ 1, 2, 3, 4, 5, 6 ]
+        ```
+        "#m
+      = fun l1 l2 => l1 @ l2,
+
+    foldl : forall a b. (a -> b -> a) -> a -> List b -> a
+      | doc m#"
+        Fold a left-associative function over a list.
```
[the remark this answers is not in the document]
yannham: Also, we should mention that `foldl` is strict in the accumulator by default (it corresponds to `foldl'` in Haskell).

yannham (changes requested):
First, thanks for this, documentation for the stdlib is very much overdue :slightly_smiling_face: 

I left a bunch of comments and nitpicking, because documentation is hard. As general remarks:

- We should try to avoid using FP/PL jargon like "bottom" or "right-associative fold" and explain things in simple terms as much as possible, as the target users of Nickel are not the same as, say, Haskell users.
- It feels descriptions could sometimes be made more concise by removing unnecessary wording or changing the structure while keeping all the important information. As an example with `typeOf`:
  > Results in a value representing the type of the given dynamically typed value.
  =>
  > Return a representation of the type of the given value.

  Another solution, when you need to refer to the arguments is to give them names (adapted from Haskell documentation):
  > `map f l` is the list obtained by applying `f` to each element of `l`
- The CI is failing because the code is not formatted. To solve this, please run `cargo fmt` at the root of the repo before committing changes, or in a separate formatting commit once you're done.

stdlib/lists.ncl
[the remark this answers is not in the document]
ErinvanderVeen: I solved this by removing any notion of associativity, adding the formula as you suggested, and adding an extra rewrite step to the example.

yannham (approved):

PR #386, opened 2021-08-31
Build the WASM REPL in the flake

yannham:
This PR adds support for building the Nickel REPL WASM artifact directly via the Nickel flake, and automate its update on the website.

- Add a `buildWasm` output which builds the Nickel REPL in WASM using `wasm-pack`. This version is optimized (`-O4`)
- Add a `checks.<system>.wasm` output to check that the WASM REPL builds. Same as `buildWasm`, but without the optimization pass, which can take time.
- Add steps to build and update the WASML REPL in the website deployment workflow, such that the playground is up to date in the netlify preview.
- Add a separate workflow to do the same thing (build and update the wasm repl) and commit the changes on pushes to master, so that the WASM REPL is up-to-date in the repo too, and not only in the netlify preview.


flake.nix
```diff
@@ -1,9 +1,12 @@
 {
   inputs.nixpkgs.url = "nixpkgs/nixos-unstable";
-  inputs.nixpkgs-mozilla.url = "github:garbas/nixpkgs-mozilla/flake";
+  # We need a fixed version for wasm-bindgen-cli, that corresponds to
+  # the version of wasm-bindgen in Cargo.toml
+  inputs.nixpkgs-pinned.url = "nixpkgs/nixos-21.05";
```
garbas, line 5: I might call this input rather `nixpkgs-wasm` or `nixpkgs-website` due to all inputs are actually pinned.

garbas (approved):
:100: I have two small comments to address, but nothing major.

Next step would be to also build the website with nix :)

yannham:
@garbas  I see only one comment, did the other one get lost during the review?

PR #395, opened 2021-09-13
[Fix]Unsound typechecking of imports

yannham:
Following #393, imports are now correctly type-checked individually. However, #393 unveiled a bug: one can use imports in a typed block with any type, and it typechecks, such as in:

```
// file import.ncl
1 + 1 : Num

// file main.ncl
let x : Str = import "import.ncl" in ... // Passes!
```

This PR fixes the issue by adding a missing call to `unify` when typechecking an import in a statically typed block, fixes an existing test, and adds a regression test. In the process, we got rid of the `to_typewrapper` function, to have it rather as a trait impl `From<Types> for TypeWrapper`.

We use the apparent type of the import for the unification constraint. That is, typechecking the usage of an import in a statically typed block is equivalent to typechecking the usage of a variable coming from code external to the block:

```
// previous example now equivalent to (with respect to typechecking):
let __import = 1 + 1 : Num in
let x : Str = __import in ... // Fail
```

Concretely, an import must either be annotated at the top-level, or be a trivial expression whose apparent type can be inferred, to be assigned something else than the `Dyn` type.


regnat (approved):
Nice :)

It took me a bit to realise that all the `to_typewrapper` removals had nothing to do with the actual problem. But otherwise, everything’s great

PR #397, opened 2021-09-16
Check for unbound type variables at parsing time

### LINKED ISSUE #384: Interpreter panics on contracts with unbound type variables
**Describe the bug**
If one enforces a contract written as a type where some type variable is unbound, the interpreter panics when it tries to elaborate this contract as a concrete term.

**To Reproduce**

```
nickel> 1 | a
thread 'main' panicked at 'Unbound type variable Ident("a")', src/types.rs:202:40
```

**Expected behavior**
Two alternatives:

1. Unbound type variables should be checked at parse time, or at least during typechecking, even for types only used as contracts.
2. Automatically generalize unbound type variable, so that one could write `fst: a -> a -> a = fun x y => x` to mean `fst: forall a. a -> a -> a = fun x y => x` as in e.g. Haskell. With this, no more unbound type variables.

**Additional context**
Another related quirk: currently the parser accepts any word as a type variable. This can lead to puzzling errors:

```
nickel> 1 | num
thread 'main' panicked at 'Unbound type variable Ident("a")', src/types.rs:202:40
```

Here, if the user write `num` instead of `Num`, it is considered as a type variable. With the proposal 2. above, this won't be an error anymore, but would be elaborated as `1 | forall num. num`, which is questionable, to say the least. The conclusion is that type variables should be only 1-character.

peperunas:
Fixes #384 

A test fails, unsure if I actually found a bug with the PR or if my code is buggy.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -214,7 +214,7 @@ pub enum ParseError {
     ),
     /// Unbound type variable
     // TODO: add relevant fields
-    UnboundTypeVariable
+    UnboundTypeVariable(Vec<Ident>)
 }
 
 /// An error occurring during the resolution of an import.
@@ -1155,9 +1155,9 @@ impl ToDiagnostic<FileId> for ParseError {
                     .with_labels(labels)
             },
             // TODO: add relevant fields
-            ParseError::UnboundTypeVariable => {
+            ParseError::UnboundTypeVariable(ident) => {
                 Diagnostic::error()
-                    .with_message(format!("Unbound type variable"))
+                    .with_message(format!("Unbound type variable(s): {:?}", ident))
             }
         };
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -213,8 +213,7 @@ pub enum ParseError {
         Option<RawSpan>,
     ),
     /// Unbound type variable
-    // TODO: add relevant fields
-    UnboundTypeVariable(Vec<Ident>)
+    UnboundTypeVariables(Vec<Ident>)
 }
 
 /// An error occurring during the resolution of an import.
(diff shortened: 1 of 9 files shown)


src/parser/utils.rs
```diff
@@ -361,3 +362,53 @@ pub fn strip_indent_doc(doc: String) -> String {
         .next()
         .expect("expected non-empty chunks after indentation of documentation")
 }
+
+pub fn check_unbound(types: Types) -> Result<Types, LexicalError> {
```
yannham, line 366: It's a bit strange that `check_unbound` return `Types` in the success case. It should probably borrows immutably the `types` and return a `Result<(), LexicalError>` instead.

src/parser/lexer.rs
```diff
@@ -386,6 +387,8 @@ pub enum LexicalError {
     InvalidAsciiEscapeCode(usize),
     /// Generic lexer error
     Generic(usize, usize),
+    /// Unbound type variable
+    UnboundTypeVariables(Vec<Ident>)
```
yannham, line 391: An unbound type variable is not a lexical error. We should probably make a new custom error for LALRPOP, `CustomParseError` or whatever, that would wrap both lexical errors and unbound type variable errors.

src/error.rs
```diff
@@ -1150,6 +1155,13 @@ impl ToDiagnostic<FileId> for ParseError {
                 Diagnostic::error()
                     .with_message(format!("{} parse error: {}", format, msg))
                     .with_labels(labels)
+            },
+            ParseError::UnboundTypeVariables(idents) => {
+                Diagnostic::error()
+                    .with_message(format!("Unbound type variable(s): {}", idents.into_iter()
+                        .map(|x| format!("\"{}\"", x))
```
yannham, line 1162:
```suggestion
                        .map(|x| format!("`{}`", x))
```
It is currently customary to report identifiers, types etc. in error message quoted by `, to differentiate them from strings.

yannham (changes requested):
The approach looks good to me :+1:  there are just a few details to fix.

yannham (approved):

PR #399, opened 2021-09-17
Add positional information for unbound type variables

peperunas:


peperunas:
I am unsure how to pass more granular details about the position of the unbound var.

e.g:

```
error: Unbound type variable(s): `a`
  ┌─ repl-input-0:1:7
  │
1 │ null: a -> (forall a. a -> a)
  │       ^^^^^^^^^^^^^^^^^^^^^^^
```

Now, the whole expression is highlighted. I would like to highlight the specific variable that is unbound.

I didn't find a direct correlation between a `Term` and a `Types`.

yannham:
Indeed, `Types` do not bear position information, as opposed to `RichTerm`. I agree it could make sense to have a `RichType`, as we have a `RichTerm`, which records types with their position. However this change would pollute a lot of existing code. Another way would be to only produce a `RichType` in the parser, just to be able to report this error precisely, and if everything is fine, erase all positional information and return a `Types`.

In my opinion it is fine to just go with showing the full type. User-written types aren't gonna be very large, and this locates the problem precisely enough so that the user should be able to find the issue quickly in practice.

src/grammar.lalrpop
```diff
@@ -16,11 +16,12 @@ use crate::term::{BinaryOp, RichTerm, Term, UnaryOp, StrChunk, MetaValue,
     MergePriority, Contract, NAryOp, RecordAttrs};
 use crate::term::make as mk_term;
 use crate::types::{Types, AbsType};
+use crate::parser::utils::mk_span;
```
yannham, line 19: Nitpick: there's already a lot of `parser::utils::{...}` imports just above, we could put `mk_span` with the others.

yannham, line 19: Nitpick: there's already a lot of `parser::utils::{...}` imports just above, we could put `mk_span` with the others.

yannham (approved):

PR #400, opened 2021-09-17
Improve type instantiation

### LINKED ISSUE #360: Improve type instantiation
Currently, polymorphic types are instantiated as in a standard Hindley-Milner system, that is as soon as a variable is used. While this is customary in the Hindley-Milner type system, it's limiting in a type system with full polymorphism (à la System F) like Nickel, as it results in polymorphic types being instantiated without necessity, and thus overspecialized:

```
nickel> let f : (forall b. b -> b) -> Num = fun f => f 0 in
let id : forall a. a -> a = fun x => x in
f id : Num

error: Incompatible types
  ┌─ repl-input-8:3:3
  │
3 │ f id : Num
  │   ^^ this expression
  │
  = The type of the expression was expected to be `forall b. b -> b`
  = The type of the expression was inferred to be `_a -> _a`

```

In this example, the type of `id` is unnecessarily instantiated to a monomorphic type `_a -> _a`, causing the application to fail, while this example is expected to typecheck.

Polymorphic types should be instantiated as late as possible, at function application, or even when trying to do types unification.

- Keep the current status-quo. This means accepting that some simple terms that typecheck in System F won't typecheck in Nickel, for no deep reason.
- Introduce a syntax for explicit type application, as in Haskell. However the cost is high for limited benefits:
    1. Yet a new syntactic construct users must know about, for something that will be fairly rare (polymorphic types of rank > 1)
    2. Users have to think about type instantiation, which requires non trivial notions in type systems, that Nickel users may not know nor care about. It seems better to try to do the  natural thing silently without having the user to think about it too much.

peperunas:
Should close #360.


yannham:
Closing as this has been stalling for a while.

PR #401, opened 2021-09-22
Restore the :print command in the repl

regnat:
Accidentally removed by https://github.com/tweag/nickel/commit/a9d5255fdb15e914a40d3a55822d4c4e1d7a28c8

~~(note that I wrote this PR using the github online editor and didn’t test anything so far)~~ I’ve tested it locally, and it seems to work all-right

### DIFF
--- a/src/repl/command.rs
+++ b/src/repl/command.rs
@@ -49,6 +49,7 @@ impl FromStr for CommandType {
             "load" | "l" => Ok(Load),
             "typecheck" | "tc" => Ok(Typecheck),
             "query" | "q" => Ok(Query),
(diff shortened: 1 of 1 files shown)


yannham (approved):
Note that now you can just click on the "Details" link of the netlify workflow and test the feature directly inside the playground :)

Thanks for the fix! Rebasing the WASM PR that refactored all the REPL code a long time ago was a bit of a hassle, so sadly I'm not totally surprised something like this slipped in. Hopefully the `:print` feature is more or less the only things that happens to the REPL on master in between, so we should be fine now.

PR #403, opened 2021-09-22
split use modules to be able to compile with --no-default-features

francois-caddet:
small fix
do not try to import a module which does not exist if feature.repl is desactivated


yannham (changes requested):
Please run `cargo fmt`. Otherwise, it's good to go.

PR #404, opened 2021-09-22
do not do anymore recursive transformations in transformations.rs (im…

francois-caddet:
…ports and transformations)

rewrite cache to do itself the recurtion
now the transformations::resolve _imports return the richterm and a list of pending imports (previously the stack used to resolve imports

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -8,9 +8,10 @@ use crate::stdlib as nickel_stdlib;
 use crate::term::{RichTerm, Term};
 use crate::typecheck::type_check;
 use crate::{eval, parser, transformations};
+use crate::transformations::PendingImport;
 use codespan::{FileId, Files};
 use io::Read;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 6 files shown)


tests/imports.rs
```diff
@@ -0,0 +1,106 @@
...
+        "should_fail",
+    )
+    .unwrap();
+    prog.eval()
+        .map_err(|e| match e {
+            Error::TypecheckError(TypecheckError::TypeMismatch(..)) => Ok(()),
+            r => Err(r),
+        })
+        .unwrap_err()
+        .unwrap();
```
yannham, line 71: You should use `assert_matches!` to make those cleaner. For an example, take a look a the tests you removed in this PR.

yannham (changes requested):
LGTM mostly, however there's something fishy going on with the way you record imports in the `resolve` function. I think it only breaks when you have multiple imports in the same file, which is why it doesn't fail the current tests. Please add a test with multiple imports in the same file to check if they are all transformed. I suspect not.

PR #405, opened 2021-09-23
Language Server: MVP

ysndr:
This PR introduces a language server (and vscode client) for nickel
At the state of this PR it can parse and typecheck an opened nickel file and will dynamically recheck it whenever it changes.
The implementation is yet still very naïve, parsing and typechecking the whole file (and everything imported) in its entirety rather than reusing previous computations.
Said performance issues will be handled in following PRs.

`lsp/nls`: contains the rust written language server
`lsp/client-extension`: vscode extension

to get this running, currently youll need to have nodejs (and oobviously rust) available
Not to self: I should add node to the development shell for an easier process using `nix develop`

- go to `lsp/client-extension`
- `npm install && npm build` the extension
- there is a launch configuration "Launch Client" which should build the server and start a vscode instance with the extension loaded
- from here just create or open a nickel file


ysndr:
@yannham I added this as a PR to my fork for now
Should i  rather post this against tweag/nickel?

yannham:
> @yannham I added this as a PR to my fork for now
> Should i rather post this against tweag/nickel?

If it's not too much hassle, it's easier (I can checkout the branch without needing github-cli, one of the CI job don't run on forks, etc.). But it's also fine to keep it that way just for this one.

Cargo.toml
```diff
@@ -46,6 +46,19 @@ wasm-bindgen = { version = "=0.2.74", optional = true, features = ["serde-serial
 js-sys = { version = "0.3", optional = true }
 serde_repr = { version = "0.1", optional = true }
 
+# language-server
+lsp-server = "0.5"
+lsp-types = "0.89"
+log = "0.4"
+env_logger = "0.9"
+anyhow = "1.0"
+
```
yannham, line 55: Since you made a separate crate, I guess we can remove this, right?

lsp/nls/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "nls"
+version = "0.1.0"
+authors = ["Nicl team"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+description = "Programmable configuration files."
+edition = "2018"
+
+[features]
```
yannham, line 10:
```suggestion

```

lsp/nls/src/server.rs
```diff
@@ -0,0 +1,124 @@
...
+                        .unwrap(),
+                )
+                .unwrap();
+            }
+            _ => {}
+        }
+    }
+
+    fn handle_request(&self, req: lsp_server::Request) {
+        todo!()
```
yannham, line 122: I didn't even know this existed. I would have used `unimplemented!()`, which is fine I guess, but still it's good to know :)

yannham (approved):
Looks good! Please just clean the main `Cargo.toml` file from unnecessary dependencies and  run `cargo fmt`, as the CI currently fails because of this.

src/cache.rs
```diff
@@ -547,6 +547,12 @@ impl Cache {
         &mut self.files
     }
 
+    /// Get a mutable reference to the cached term roots
+    /// (used by the language server to invalidate previously parsed entries)
+    pub fn terms_mut(&mut self) -> &mut HashMap<FileId, (RichTerm, EntryState)> {
+        &mut self.terms
+    }
+
```
yannham, line 555: I don't really like the idea of exposing publicly the underlying hashmap. That's ok for now if it's a blocker for you, because it seems the cache is not handling file changes correctly, but let us maybe open an issue to fix the actual problem and remember to remove this once this is not needed anymore.

ysndr: Yes I see the problems with this, `pub` is remarkably open.
Alternatively we could have an `invalidate` method, but if we start that and want to use the cache term structures in future analysis we start adding a bunch of methods just for the lsp
If we don't want to do either I have to reimplement most parts of the cache (which might be the most applicable thing anyway if it turns out that the processing would require too much work on the existing nickel module)

Yes, let me open an issue..

PR #410, opened 2021-09-27
Add git hook (un)installer

ysndr:
Redo `nix develop` shell to use `mkShell`
Adds commands to install and uninstall git-hooks.

Currently only adds a git commit hook that prevents committing unformatted rust code.
Can be configured to be a push hook as well

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -208,9 +208,96 @@
           '';
         };
 
+      buildDevShell = { system, channel ? "stable" }: 
+      let 
+        pkgs = mkPkgs { inherit system; };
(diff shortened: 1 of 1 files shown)


ysndr:
Somewhat relevant XKCD

edolstra:
I wonder if git can be configured to get hooks from an environment variable. Or maybe we could patch git to support that. Modifying the hooks directory dynamically is awfully imperative...

edolstra (approved):

ysndr:
I may have another look on the tool but for now i think this approach solves what its supposed to without adding any dependencies.

PR #411, opened 2021-09-28
Redirect website to HTTPS

garbas:


garbas:
I enabled the HTTPS via Netlify, but we are still missing the redirect. I'm hope this does the trick.

website/netlify.toml
```diff
@@ -0,0 +1,29 @@
+[[redirects]]
+  from = "https://nickel-lang.netlify.com/*"
+  to = "https://nickel-lang.org/:splat"
+  status = 301
+  force = true
+
+[[redirects]]
+  from = "http://nickel-lang.netlify.com/*"
+  to = "http://nickel-lang.org/:splat"
```
yannham, line 9: Is it on purpose that the target is `http` and not `https`?

yannham (approved):
Thanks! LGTM, although please look at the comment to make sure there isn't a missing `https`.

website/netlify.toml
garbas: Fixed. Good catch!

PR #416, opened 2021-10-06
Update termimad

### LINKED ISSUE #415: The output of `query` is unreadable on light terminals
**Describe the bug**

Several parts of the output of `query` (either `nickel query` or `:query` in the repl) have colors that make it unreadable in light terminals:

Capture d’écran de 2021-10-06 16-14-19

(with the default light color of gnome terminal).

**To Reproduce**

- Get a terminal with a light colorscheme
- `nickel query <<< lists.map`

**Expected behavior**

Well… things should be readable :man_shrugging: 

**Environment**
 - OS name + version: NixOS 21.05
 - Version of the code: fecdfff7ffdd9ddfb3aed241716ccacaed0f3ac6

**Additional context**

Recent versions of `termimad` apparently have two different versions of the default skin that can be chosen using [MadSkin::default_light()](https://docs.rs/termimad/0.16.2/termimad/struct.MadSkin.html#method.default_light) or [MadSkin::default_dark()](https://docs.rs/termimad/0.16.2/termimad/struct.MadSkin.html#method.default_dark). I guess that’s a good way to go to have something that works in every case

regnat:
Fix #415 as the new version provides a style that seems to work great both on light and dark backgrounds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,9 +99,9 @@ checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


yannham (approved):
Nice! thanks.

PR #421, opened 2021-10-15
[Language Server] indexing scopes

ysndr:
Track and collect scopes for linearized items


yannham (changes requested):
If I understand correctly, you are representing scopes (that are morally a tree mimicking the structure of the AST where nodes corresponds to a local environment) as a map that associates a path from the root in this tree to a scope.

I guess there are others, maybe simpler and/or more efficient ways of doing this. One would be to store scope information in the linearization items themselves, avoiding the use of an  auxiliary data structure. A second one would be to use a tree representation directly instead of a hashmap (although that's not obviously better in term of performance, because of cache and all).

Anyway, performance is hard, and your approach is still reasonable, so let's wait to have some actual measurement before yak-shaving.

However, I couldn't compile the language server using using stable Rust. Please fix the compilation errors as well as warnings.

yannham:
By the way, you may have to rebase onto `master` to get rid of this macOS nix installation bug that affects the CI.

ysndr:
I resolved the comments you had @yannham, are you fine with this being merged?

lsp/nls/src/linearization.rs
```diff
@@ -0,0 +1,206 @@
...
+                ty,
+                scope: self.scope.clone(),
+                kind: TermKind::Structure,
+            }),
+        }
+    }
+
+    /// [Self::add_term] produces a depth first representation or the
+    /// traversed AST. This function indexes items by _source position_.
+    /// Elements are reorderd to allow efficient lookup of elemts by
```
yannham, line 141:
```suggestion
    /// Elements are reordered to allow efficient lookup of elements by
```

yannham (approved):

PR #432, opened 2021-10-26
Fix looping typechecking of cyclic imports

### LINKED ISSUE #409: Cyclic imports make the interpreter loop
**Describe the bug**
Cyclic imports make the binary overflow its stack. It's not evaluation, as it already happens just with typechecking. Spotted because of #407.

**To Reproduce**

```
echo 'let x = import "b.ncl" in null' > a.ncl
echo 'let x = import "c.ncl" in x' > b.ncl
echo 'let x = import "b.ncl" in x' > c.ncl
nickel -f a.ncl typecheck
```

**Expected behavior**
The program should evaluate successfully on `null`. Cyclic imports shouldn't be a problem. They weren't originally.

**Environment**
 - OS name + version: Linux 5.10.0-8-amd64 SMP Debian 5.10.46-4 (2021-08-03) x86_64 GNU/Linux
 - Version of the code: 269ead2a58ded6b1939e70dcd23550e92304277f

**Additional context**

A git bisect shows that the offending commit is b029763736dfa954e82bc65370b35c24117f6848
cc @francois-caddet

yannham:
Fix #409.

Do not make `type_check` functions transitively typecheck imports anymore. They now just peek at the toplevel node of the imports to see the apparent type, in order to be able to decide the type of an expression `import "foo.ncl"`, but they restrain from recursing inside `foo.ncl`.

It is now the responsibility of the caller (in practice, the cache) to call `type_check` on transitive imports of a source file, and the cache is able to see the imports it has already processed, which eliminates the loop.

Did small code cleaning on the way (such as removing the `Result` from `transform`'s return type, as now that transform does not perform import resolution, it isn't fallible anymore).

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -88,11 +88,23 @@ pub struct NameIdEntry {
 }
 
 /// The state of an entry of the term cache.
+///
+/// # Imports
+///
+/// Usually, when we apply a procedure to an entry (typechecking, transformation, ...), we also
+/// process all of its transitive imports. However, note that the state is guaranteed to be correct
+/// only for the entry, and not necessarily for its transitive imports. Under normal conditions, if
+/// the state of an entry is `Typechecked`, then all of its transitive imports should also be at
+/// least in the state `Typechecked`. This is not true if an import fails to typecheck, leaving the
+/// entry `Typechecked` while the import stays as `Parsed`.
+///
+/// Thus, the state (e.g. `Typechecked`) only guarantees in general that the corresponding
+/// procedure have been successfully applied to the entry, not to all of its transitive imports.
 #[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Copy, Clone)]
 pub enum EntryState {
     /// The term have just been parsed.
     Parsed,
-    /// The term has been parsed and the imports resolved
+    /// The term has been parsed and the imports resolved.
     ImportsResolved,
     /// The term have been parsed and typechecked.
     Typechecked,
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -5,8 +5,7 @@ use crate::parser::lexer::Lexer;
 use crate::position::TermPos;
 use crate::stdlib as nickel_stdlib;
 use crate::term::{RichTerm, Term};
-use crate::typecheck::linearization::StubHost;
-use crate::typecheck::type_check;
+use crate::typecheck::{linearization::StubHost, type_check};
 use crate::{eval, parser, transformations};
 use codespan::{FileId, Files};
 use io::Read;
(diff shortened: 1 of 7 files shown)


src/cache.rs
```diff
@@ -340,23 +339,26 @@ impl Cache {
...
             Ok(CacheOp::Cached(()))
         } else if *state >= EntryState::Parsed {
             type_check(t, global_env, self)?;
             self.update_state(file_id, EntryState::Typechecked);
+
+            if let Some(imports) = self.imports.get(&file_id).cloned() {
+                for f in imports.into_iter() {
+                    self.typecheck(f, global_env)?;
+                }
+            }
```
francois-caddet, line 360: Is it not better to typecheck deeper imports before `import` callers? I mean, put the for loop before `type_check(t, global_env, self)?;`. At least, I think we update the state to early: what append if the typechecking of a deep import fail?

src/cache.rs
```diff
@@ -367,18 +369,20 @@ impl Cache {
...
+                let t = transformations::transform(t);
+                self.terms.insert(file_id, (t, EntryState::Transformed));
+
                 if let Some(imports) = self.imports.get(&file_id).cloned() {
-                    for f in imports.iter() {
-                        self.transform(*f)?;
+                    for f in imports.into_iter() {
+                        self.transform(f)?;
                     }
                 }
```
francois-caddet, line 384: same as above

src/cache.rs
```diff
@@ -465,10 +462,12 @@ impl Cache {
             Some(_) => {
                 let (t, _) = self.terms.remove(&file_id).unwrap();
                 let (t, pending) = transformations::resolve_imports(t, self)?;
+                self.terms.insert(file_id, (t, EntryState::ImportsResolved));
+
                 for id in pending {
                     self.resolve_imports(id)?;
                 }
-                self.terms.insert(file_id, (t, EntryState::ImportsResolved));
```
francois-caddet, line 471: Same comment as commented for typechecking above

src/cache.rs
yannham: I think I had to do this to make operations on cyclic imports to succeed. Say we are typechecking `a`, and there is a cyclic import `a imports b, b imports a`. And say do as before: start typechecking `a`, then typecheck `b`, and only then finally set the state of `a` to `Typechecked`. Then, in the for loop, we will typecheck `b`. `b` will do the same thing, typecheck the content of `b`, and then typecheck imports. Including `a`. Because `a` is still in the state `Parsed`, it will go again, which creates the exact kind of loop we had in `type_check`, but just now it is in the cache layer.

Here, the `EntryState` is to understand as "is a specific operation has been performed on this specific entry". `Typechecked` would mean "this particular file has been typechecked", which is indeed what typechecks do, without telling anything about imports. Under this interpretation, we are always in a correct state. But indeed it must be documented clearly so that users of the cache don't make false assumption.

One alternative is to add intermediate states like `Typechecking`, `Transforming`, and so on. I don't know if this is very useful, but it's not very expensive to do though.

yannham: I expanded the documentation on `EntryState`. I think it should do the trick for now, if we ever need to differentiate between "the entry only is typechecked" and "the entry and all of its transitive dependencies are typechecked" at some point, we can add the missing states then. What do yo think @francois-caddet ?

francois-caddet: The issue is:
When the `EntryState` is `Resolved`, the function just ignore the import and do not resolve it. Right now it's not realy an issue because an error in any file is fatal. But if later we would like to add a recover on error system, it will not work as expected. It can be managed by checking `Cache::imports` tree and the coresponding state. But we have to document it, because if not done, we will have very wear evaluation errors.

francois-caddet (approved):

src/cache.rs
yannham: Good point, it may be more natural for `Typechecked` to mean everything went fine, imports included.

PR #436, opened 2021-10-29
[Language server] Metadata

ysndr:
Add hover information for meta data

Depends #421 

- [x] Basic contracts
- [x] Documentation
- [x] Merge priority
- [x] show information for variables that reference records with metadata
- [ ] metadata info for record fields (#447)
- [x] Jumping to contracts found in metadata (also resolves as usage)

### DIFF
--- a/lsp/nls/src/linearization.rs
+++ b/lsp/nls/src/linearization.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use nickel::{
     identifier::Ident,
     position::TermPos,
-    term::{self, MetaValue, RichTerm, Term},
+    term::{MetaValue, RichTerm, Term},
     typecheck::{
         linearization::{
             Building, Completed, Environment, Linearization, LinearizationItem, Linearizer,
--- a/lsp/nls/src/linearization.rs
+++ b/lsp/nls/src/linearization.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use nickel::{
     identifier::Ident,
     position::TermPos,
-    term::Term,
+    term::{MetaValue, Term},
     typecheck::{
         linearization::{
             Building, Completed, Environment, Linearization, LinearizationItem, Linearizer,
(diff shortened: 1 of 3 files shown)


lsp/nls/src/linearization.rs
```diff
@@ -124,14 +133,56 @@ impl Linearizer<BuildingResource, (UnifTable, HashMap<usize, Ident>)> for Analys
...
+                // Notice 1: No push to lin
+                // Notice 2: we discard the encoded value as anything we
+                //           would do with the value will be handled in the following
+                //           call to [Self::add_term]
+                let meta = MetaValue {
+                    value: None,
+                    ..meta.to_owned()
+                };
+
+                for contract in meta.contracts.iter().cloned() {
```
yannham, line 149: Do we really need to clone the iterator here? It feels like we are only matching on stuff, or at worst need to clone identifiers, but not the whole contracts.

yannham (commented):
The code looks fine. Though in practice I still can't go to the definition of a contract, while the code looks like it should handle this case. Typically:

```
let ContractFoo = ... in
{
  bar | #ContractFoo = "baz"
}
```
If I hover on `#ContractFoo` and then go to definition, nothing happens.

Also, it seems the code inside a flat type is only indexed if it is an identifier. I think we should index all such code, as any other language construct, because it can contain things more complex than just a variable. A common example is a parametrized contract:

```
let GreaterThan = fun base label value => if value >= base then value else contracts.blame label in
let myFoo | #(GreaterThan 2) = 3
```

In that case I have an application inside the flat type, but I still would like to be able to jump to the definition of `GreaterThan`.

lsp/nls/src/linearization.rs
ysndr: No, fixed in #447

ysndr:
> The code looks fine. Though in practice I still can't go to the definition of a contract, while the code looks like it should handle this case.

Here, the Contract is attached to a record field, which is not supported in this first PR, yet implemented as a consequence of #447

ysndr:
> ```
> let myFoo | #(GreaterThan 2) = 3
> ```

The Contract types are not yet unpacked, if they are not flat types

yannham (approved):

PR #437, opened 2021-10-29
Implement std::fmt::Display for Marker and OperationCont for debug printing

litchipi:
It occured severall times that I have Stack overflows errors when trying to print some variables to debug something and it happens that this variable contains everything from the stdlib for example.

To get around this, I implemented the `std::fmt::Display` trait on some elements to be able to still see the maximum informations about them, without dumping the whole memory into the console.
I think it could be much more extended than Marker and OperationCont, and I can continue to implement the trait on other elements if needed.
Thought it could be good to have a quick and easy text representation of elements inside the machine

Example using `stack.display()` in the `continuate_operation` function:
```
--- STACK ---
| Cont Op2First Plus (callstack size 1)
| Thunk
| Cont Op1 DeepSeq (callstack size 0)
| Arg
---  END  ---
--- STACK ---
| Cont Op2First Plus (callstack size 1)
| Thunk
| Cont Op1 DeepSeq (callstack size 0)
| Arg
---  END  ---
--- STACK ---
| Thunk
| Cont Op1 DeepSeq (callstack size 0)
| Arg
---  END  ---
```

### DIFF
--- a/src/stack.rs
+++ b/src/stack.rs
@@ -7,7 +7,6 @@ use crate::position::TermPos;
 use crate::term::{RichTerm, StrChunk};
 
 /// An element of the stack.
-#[derive(Debug)]
 pub enum Marker {
     /// An equality to test.
     ///
@@ -47,6 +46,21 @@ pub enum Marker {
         Environment, /* the common environment of chunks */
     ),
     Strictness(bool),
+}
+
+impl std::fmt::Debug for Marker {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Marker::Eq(_, _) => write!(f, "Eq"),
+            Marker::Arg(_, _) => write!(f, "Arg"),
+            Marker::TrackedArg(_, _) => write!(f, "TrackedArg"),
+            Marker::Thunk(_) => write!(f, "Thunk"),
+            Marker::Cont(op, sz, _) => write!(f, "Cont {:?} (callstack size {})", op, sz),
+            Marker::StrChunk(_) => write!(f, "StrChunk"),
+            Marker::StrAcc(_, _, _) => write!(f, "StrAcc"),
+            Marker::Strictness(s) => write!(f, "Strictness = {}", s),
+        }
+    }
 }
 
 impl Marker {
--- a/src/stack.rs
+++ b/src/stack.rs
@@ -94,7 +94,6 @@ impl Marker {
 }
 
 /// The evaluation stack.
-#[derive(Debug)]
 pub struct Stack(Vec<Marker>);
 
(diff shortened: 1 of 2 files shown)


yannham (commented):
That can be useful, indeed. Though given the look of the output, I tend to think it corresponds to the `Debug` trait rather than the `Display` one.

litchipi:
I rewrote them using `Debug` trait, splitted the commits for clarity, and added the implementation for `Term` which was pretty useful on evaluation debug (not very verbose, but still brings a thing).
Do not hesitate if you think of other structs who would benefit from that implementation.

src/stack.rs
```diff
@@ -97,6 +110,14 @@ impl Stack {
         Stack(Vec::new())
     }
 
+    pub fn display(&self) {
```
yannham, line 113: Why not implement the debug trait on stack too (with the same content as this function)? This let the caller decide where the output should go e.g., or print a stack as part of a bigger debug string, etc. It seems slightly arbitrary to have a `display()` function instead of implementing `Debug` just for the stack.

src/term.rs
```diff
@@ -541,6 +541,43 @@ impl Term {
     }
 }
 
+impl std::fmt::Debug for Term {
```
yannham, line 544: I'm not sure about this implementation. Usually, I use the debug trait to print a term in full (although I must confess it is far from being very readable). This implementation overrides the default implementation of `Debug` for RichTerm, but only prints one layer (for example printing `fun x => ...` only prints `Fun x`, which hides all the content of the function inside `...`). It makes sense for the stack which otherwise is way too verbose and can loop because of mutually recursive thunks, but here we are removing the only existing way to print a term in full as part of a debug process.

src/stack.rs
litchipi: Indeed, gonna fix that right away

src/term.rs
litchipi: Well I agree you cannot see what's "inside", however when dealing with stdlib problems (it loads everything into a big big `RichTerm`), you get stack overflows by displaying it (in addition of being unreadable).
The best way would be to let the choice wether to display the full thing or not, however I'm not sure of the cleanest way to implement that

yannham: I agree the current situation is not optimal, but I still think we should keep a way of printing a rich term in full. The right think to do is to implement a pretty printer for `RichTerm`s, but this is outside the scope of this PR.

litchipi: Okay then, no std::fmt::Debug for Term for now.
The issue of not being able to peek inside data (because of too much nested data) for debug purpose is still a bit annoying, it could lead to difficulties later on.

yannham (approved):

PR #438, opened 2021-11-01
[Language Server] Simple document symbols

ysndr:
The current linearization contains all variables defined in the document.
This PR makes use of that and exposes those definitions as document symbols.

- currently the whole scope of the let binding is highlighted
- we do not distinguish between types or record fields yet

(in VS-Code the document symbols can be opened with `cmd/ctrl+shift+o`)

### DIFF
--- a/lsp/nls/src/main.rs
+++ b/lsp/nls/src/main.rs
@@ -9,6 +9,7 @@ mod linearization;
 mod requests;
 mod server;
 use server::Server;
+mod term;
 
 fn main() -> Result<()> {
     env_logger::init();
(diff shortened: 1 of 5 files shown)


yannham (approved):
> we do not distinguish between types or record fields yet

I guess you mean it's not included in the symbols list.

Otherwise, one point is that this code will go over the whole linearization at each request to show document symbols. I imagine we should be caching this information. But this is a larger problem that is not limited to document symbols, so let's have the feature like this for now, and think about caching a bunch of LSP computations more generally.

Also, there are a few warnings which would be nice to fix before merging.

ysndr:
The warnings arise inside the generated grammar as far as i could tell

PR #439, opened 2021-11-02
Update and simplify flakes-based GH workflow

yannham:
The MacOS CI is failing repetitively on the install nix action. This PR tries to fix this by bumping the version of the nix install and the cachix action. Also, now that flakes are in stable nix (although as an experimental feature), this PR gets rid of the bootstraping of a flake-enabled nix version, which makes action code simpler.


yannham:
Merging without review: this is a tiny change, and as long it is not merged, the CI fails on all the PRs.

PR #443, opened 2021-11-04
Refactoring operation continuation with Strictness markers on stack

litchipi:
Using strictness markers to save and restore strictness during operation continuations.
Continuation from #433, allows to take care of operation continuation without saving a `prev_strictness` value in the `OperationCont` variants.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -527,18 +527,19 @@ where
                 }
             }
             Term::Op1(op, t) => {
-                let prev_strict = enriched_strict;
+                if !enriched_strict {
+                    stack.push_strictness(enriched_strict);
+                }
                 enriched_strict = true;
-                stack.push_op_cont(
-                    OperationCont::Op1(op, t.pos, prev_strict),
-                    call_stack.len(),
-                    pos,
-                );
+                stack.push_op_cont(OperationCont::Op1(op, t.pos), call_stack.len(), pos);
                 Closure { body: t, env }
             }
             Term::Op2(op, fst, snd) => {
-                let prev_strict = enriched_strict;
-                enriched_strict = op.is_strict();
+                let strict_op = op.is_strict();
+                if enriched_strict != strict_op {
+                    stack.push_strictness(enriched_strict);
+                }
+                enriched_strict = strict_op;
                 stack.push_op_cont(
                     OperationCont::Op2First(
                         op,
@@ -547,7 +548,6 @@ where
                             env: env.clone(),
                         },
                         fst.pos,
-                        prev_strict,
                     ),
                     call_stack.len(),
                     pos,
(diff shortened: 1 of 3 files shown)


yannham (approved):
LGTM. I think there is a remaining useless parameters `enriched_strict` still present in the signature of `continuate_operation`. This is the kind of things warnings are for, but I see there is a flood of unused warnings coming from the LSP PRs that makes new ones unnoticeable.

litchipi:
Well the `enriched_strict` argument in `continuate_operation` is still useful as it allows the popped strictness markers to be applied. Also I don't see it in the compiler warnings.

yannham:
Oh, right. But then I think popping strictness markers inside the `continuate_operation` is useless, because it will be done anyway at the very beginning of the next iteration of `eval`?

litchipi:
Indeed, didn't thought about it.
However when I remove everything, it seams that some tests doesn't pass, due to code handling `StrAcc` in         `UnaryOp::ChunksConcat()` operation processing.
I'm investigating it right now

litchipi:
I spotted the problem, but I don't know what to think of the solution I gave..

The arguments for `ChunksConcat` are added to the stack before dealing with the operation. When the `continuate_operation` is called, the stack looks like this:
```                                                                                                                                                                                                                                                              
--- STACK ---
| Cont Op1 ChunksConcat (callstack size 12)             } Term::Op1 evaluation
| Strictness = false                                    }
| StrAcc                                       } Term::StrChunks evaluation
| Thunk                                        }
| Thunk
| Cont Op2First Merge (callstack size 10) 
| Strictness = true 
| Thunk 
| Thunk
| Thunk 
| Cont Op2First Eq (callstack size 7)
| Eq
| Thunk 
| Thunk 
| Cont Op1 BoolOr (callstack size 5) 
| Arg 
| Thunk
| Cont Op1 BoolAnd (callstack size 0)
| Arg
---  END  --- 
```
Then, it pops the `Cont Op1 ChunksConcat` from the top of the stack, and call the `process_unary_operation` that expects the `StrAcc` to follow the operation continuation. But in the meantime the strictness changed (as it's the operator that set the strictness).
To fix it, I allowed the evaluation of `Term::StrChunks` to set the strictness to `true` to prepare for the unary operation in advance, and the tests passes.

**The question is:**   Is this term evaluation always followed by a unary operation ? Will it always require `enriched_strict=true` ? Or by doing this, am I breaking something in the future ?

yannham (approved):
I see. It does sound reasonable to me. The evaluation of `StrChunks` is always preparing the stack and then calling immediately to `ChunksConcat`. `ChunksConcat` operates on primitive strings, so I can't imagine a case where it shouldn't be strict. It makes sense to set the strictness to true at the first `StrChunks`, instead of waiting the execution of `ChunksConcat`.

PR #449, opened 2021-11-08
Mention that `--format` also accepts `yaml, toml`

silverraven691:
Would you like me to add a comment right next to the definition of `ExportFormat` to remind whoever will edit it to also update this documentation line?

Thanks

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -28,7 +28,7 @@ struct Opt {
 enum Command {
     /// Export the result to a different format
     Export {
-        /// Available formats: `raw, json`. Default format: `json`.
+        /// Available formats: `raw, json, yaml, toml`. Default format: `json`.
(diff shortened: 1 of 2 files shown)


yannham:
Hello,

> Would you like me to add a comment right next to the definition of ExportFormat to remind whoever will edit it to also update this documentation line?

Yes, that is a good idea. Thanks for contributing!

yannham (approved):

ysndr:
I think we could make use of the [arg_enum macro](https://docs.rs/clap/2.27.1/clap/macro.arg_enum.html) to generate this docs automatically given that clap is already a dependency through structopt

yannham:
> I think we could make use of the arg_enum macro to generate this docs automatically given that clap is already a dependency through structopt

Indeed. Let's merge this for now and open an issue for this more robust solution.

PR #451, opened 2021-11-08
Fix lsp not parsed panic

Avi-D-coder:
The lsp server no longer panics if the client makes a request before a file has been parsed.
This used to happen if a file had a syntax error in it.

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -1,13 +1,17 @@
 use codespan::ByteIndex;
 use codespan_lsp::position_to_byte_index;
 use log::debug;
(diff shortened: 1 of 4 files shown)


yannham (approved):
I let @ysndr have a look too, but LGTM, and importantly this PR solves the very annoying problem I was having just right now that the LSP was stopping after a few edits in neovim, and I had to restart coc manually quite often. It was bad enough such that I was going to investigate, but this is the fix!

ysndr (approved):
Looks good to me too.
I think a next step could be to introduce an error library, e.g. anyhow (or thiseror in nickel) to make custom errors easier to work with.

PR #454, opened 2021-11-09
[Language server] Properly resolve let and function bindings

ysndr:
The previous approach used the target type to type let idents and function declarations. During typechecking new type variables are created for inferred binding types.
This PR makes use of this fact by allowing to retype indexed idents while in scope


yannham (approved):
There is a merge conflict to solve, but otherwise it works and LGTM.

PR #458, opened 2021-11-10
Demonstrate the difference between seq and deepSeq in their documentation examples

silverraven691:
Hello,

This adds a case in which seq and deepSeq differ, hopefully demonstrating their difference

Thanks


yannham (approved):
Useful precision. Thanks!

Issue #422, opened 2021-10-19
User-defined types


thufschmitt:
**Is your feature request related to a problem? Please describe.**

Maybe I just missed it, but it seems that there’s no way to define (name) custom static types.

**Describe the solution you'd like**

A syntax allowing me to (for example):

```nickel
let type t = { foo: Num } in
[ { foo = 1 }, { foo = 2 }] : t
```

**Describe alternatives you've considered**

Use contracts everywhere instead, something like:

```nickel
let t = { foo | Num} in
[ { foo: 1 }, { foo: 2 }] | List #t
```

But that’s a bit of a shame

yannham:
Sorry, I missed this. But I think @litchipi is now on it, right?

regnat:
> Sorry, I missed this. But I think @litchipi is now on it, right?

Yes, he is :) We’re supposed to discuss this offline next week

regnat:
We’ve had a bit of chat with @litchipi about that.

This is actually a bit more involved than what I initially though, in particular because the built-in types are currently hardcoded whithin the interpreter − [the definition of types](https://github.com/tweag/nickel/blob/edb09ea346b9891abd88b7d5dc88920c8a68aa00/src/types.rs#L71) explicitly lists the existing builtin types − meaning that the issue is not just a syntactical one (define new type aliases), but also a typing one (extend the typechecking to also handle type aliases).

As a rough implementation plan, this would mean something like:
1. Extend the definition of types to also include a reference to an identifier (similar to the `Var(Indent)` case, but for concrete types rather than type variables).
2. Extend the typechecking algorithm to handle these types properly (including being careful with recursive types). Probably means passing a new “type aliases” environment to be able to resolve type aliases.
3. Add a syntax to define these aliases.
  I suspect we might also want this syntax to define a corresponding contact at the same time, although the exact semantics are a bit fuzzy.

Another option would be to inline all the type aliases before starting the actual type-checking, but that might not make things simpler, and that means forbidding recursive types altogether (because they would expand to an infinite type), which is probably too restrictive in practice (well, I guess it is, maybe I’m wrong here).

(@litchipi don’t hesitate to correct me if there’s anything wrong or missing)

litchipi:
One idea would be to have a unified type definition, and create pre-defined types `type Num = %num%`, `type Str = %str%`, etc ... to handle builtins types, while having a unified type definition system when handling variables or fields.
[...]

PR #464, opened 2021-11-17
Implement User-defined types with TypeAliasEnv to store recurrent type definitions

litchipi:
Implement user-defined types definition using a new syntax `let type ... in ...`.
Uses a TypeAliasEnv to store user-defined types and replace the `ident` given with the definition stored when performing typchecking and transformation.

The changes looks big because it adds an environment in places where they weren't before, but to keep it simple:
- If we encounter a `Term::TypeAlias(ident, type)`, we store the `type` in the environment with key `ident`
- Later on the process (typecheck or transform), if we encounter the type definition `ident`, we fetch the `type` from the environment and perform the operations on it instead.

Quick Nickel code to test it out:
```
let type recType = { foo: Num } in
let type mapType = forall a b. (a -> b) -> List a -> List b in 

let mylib = {
    mymap : mapType = fun f l => %map% l f,
    mymap_standard : forall a b. (a -> b) -> List a -> List b = fun f l => %map% l f,
    recTest1 : recType = { foo = 4 },
    recTest2 : recType = { foo = 5 },
} in 

{
    test = mylib.mymap (fun x => x * 2) [2, 4, 6, 8, 10, 12],
    test_standard = mylib.mymap_standard (fun x => x * 2) [2, 4, 6, 8, 10, 12],

    testrec1 = mylib.recTest1,
    testrec2 = mylib.recTest2,
}
```
Everything should work well, and inducing an error like `recTest2 : recType = { foo = 5, bar = false},` should return a typechecking error with the corresponding hints as if the type alias was written by hand.

Feature request #422

### DIFF
--- a/examples/user-defined-types/user-defined-types.ncl
+++ b/examples/user-defined-types/user-defined-types.ncl
@@ -1,0 +1,34 @@
+typealias Type1 = { foo: Num, bar: Str } in
+typealias Type2 = { rec: Type1, foobool: Bool } in
+typealias Type3 = { recrec: Type2, rec: Type1, foo: Num } in
+
+typealias FctType = forall a b. (a -> b) -> List a -> List b in
+
+let maptest : forall a b. (a -> b) -> List a -> List b = fun f l => %map% l f in
+//let map1 : FctType = fun f l => %map% l f in
+//let map2 : FctType = fun f l => %map% l f in
+
+let f = fun x => x in
+
+{
+    mylist0 = maptest (fun z => z*4) [4, 6],
+//    mylist1 = map1 (fun x => x*2) [3, 5],
+//    mylist2 = map2 (fun y => y*5) [5, 10],
+
+//    outofscope_type = (typealias foo = Num in (f (0 : foo)) : foo) + (f 5),
+
+    data : Type3 = {
+        recrec = {
+            rec = {
+                foo = 3,
+                bar = "recrec",
+            },
+            foobool = false,
+        },
+        rec = {
+            foo = 5,
+            bar = "rec",
+        },
+        foo = 42,
+    },
+}
--- a/src/error.rs
+++ b/src/error.rs
@@ -418,9 +418,6 @@ impl ParseError {
                 InternalParseError::Lexical(LexicalError::InvalidAsciiEscapeCode(location)) => {
                     ParseError::InvalidAsciiEscapeCode(mk_span(file_id, location, location + 2))
                 }
-                InternalParseError::UnboundTypeVariables(idents, span) => {
-                    ParseError::UnboundTypeVariables(idents, span)
-                }
             },
         }
     }
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -614,26 +614,7 @@ where
                 }
             },
             Term::Promise(ty, mut l, t) => {
-                l.arg_pos = t.pos;
-                let thunk = Thunk::new(
-                    Closure {
-                        body: t,
-                        env: env.clone(),
-                    },
-                    IdentKind::Lam(),
-                );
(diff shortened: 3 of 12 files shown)


regnat:
Thanks a lot! I’ll try to have a look at it before the end of the week.

Note that the test don’t seem to pass right now

litchipi:
Yeah, forgot to run `cargo test` before pushing :-|

To explain a bit this is what happening right now:
- There was a little mistake in the code during the `check_unbound` function in the parser, that made it ignore everything
- After it was fixed, I saw that the parser is going bottom to top, and so when the user-defined types are encountered, they are not yet in the env and are treated as unbound
- After discussion with @yannham, everything related to the type environment and `check_unbound` operation should be removed from the parser section and only set in the later `typecheck` and `transform` sections.
- Used during these sections, the `traverse` function goes bottom to top and so raises the same issue than the parser before
- Rewriting the `traverse` function top to bottom raises a nasty stack-overflow error after a recursion loop, due to the `share_normal_form::transform_one` operation
- I'm trying to fix this issue, along with rewriting `traverse` so it doesn't use recursion anymore

So before it gets merged, this feature may require a few PR for the preparatory changes first.

litchipi:
Some progress: 
- Modified the `let type` to `typealias` keyword to avoid collision with somewhat frequent `type` word in configurations
- Fixed everything so the custom records types definitions are working
- Still a bug with type inference related to #360 , that makes `cargo test` fail, along with any custom function types (exemple: `typealias FctType = forall a b. (a -> b) -> List a -> List b in`).

Needed to implement several changes that should maybe have their own PR:
- Removed `Term::Promise` from `eval` (apparently should be removed from Nickel totally)
- Implemented TypeClosures in typecheck process to keep an environment going while performing the checks (same as `eval`)
- Removed any variable binding checks from the parser. (along with the `unbound_type_variable` test in the parser)

However I'm quite stuck on the final bug, and I feel the implementations I did are heavy (a lot of clones going on), and would love some help about it.

litchipi:
Added a (somewhat ugly) type alias resolution inside the typechecking of MetaValue types (aka Promises), and that makes the example work.

However, the following code should work, and it doesn't:
```
let f = fun x => x in
{
    outofscope_type = (typealias foo = Num in (f (0 : foo)) : foo) + (f 5),
}
```
I think it has to do with environments and scope, but I don't really have a clue on how to fix this.

litchipi:
I'd like your opinions on the syntax to adopt for this new feature, random choices are:
1. `let type <ident> = <def> in <term>`: (Raises some issues because of the introduction of the `type` keyword)
2. `typealias <ident> = <dev> in <term>`
3. `let <ident> : <def> in <term>`: (Could have the confusion with variables definition there)
4. `infer <ident> = <dev> in <term>`

Any other ideas are welcomes

yannham:
1. `let type` is indeed annoying because I guess several JSON schemas have "types" field sprinkled everywhere, so we better not have it as a keyword (fields can still be quoted, but that's a bit annoying).
2. No opinion. A bit long but not bad per se
3. `let <ident> : <def> in <term>` is confusing, IMO. Usually `x : T` means `x` is of type `T`. If `x` is itself a type, then it usually refers to "types of types" (aka kinds). For example in Nickel, `let x : Type = exp in exp` already exists and has a different meaning. It would be confusing.
4. I'm not in favor of this one, because we don't infer anything, we just assign a type to an identifier.

Random proposition from existing languages (e.g. Haskell): `newtype t = Foo in ...`.

litchipi:
A new (almost working) version has been created on top of the latest code from #470 (and rebased on latest master).

The code 
```
 outofscope_type = (typealias foo = Num in (f (0 : foo)) : foo) + (f 5)
```
still doesn't work however

PR #468, opened 2021-11-19
Fix `lists.reverse`

silverraven691:
Right now, it doesn't do anything :)

```
reverse [ 1, 2, 3 ]
== f ( f ( f [] 1 ) 2 ) 3
== (([] @ [1]) @ [2]) @ [3]
```

### DIFF
--- a/stdlib/lists.ncl
+++ b/stdlib/lists.ncl
@@ -155,7 +155,7 @@
             [ 3, 2, 1 ]
         ```
         "#m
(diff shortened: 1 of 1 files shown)


yannham (approved):
Good catch. IIRC, it's because `foldl` was itself reversed at the time it was introduced, so it must have been looking fine then, but when we fixed `foldl`, it haven't been anymore. Thanks!

PR #469, opened 2021-11-19
Creating a 2-step process, top-to-bottom and bottom-to-top for transormation phase

litchipi:
Modifies the `traverse` function that is responsible for applying the contracts and sharing the normal form of a RichTerm.

Initially bottom-up, this patch sets up 2 steps, one top to bottom, one bottom to top.
Provides necessary fix for #464

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -826,6 +826,12 @@ impl fmt::Display for NAryOp {
     }
 }
 
+#[derive(Copy, Clone)]
+pub enum TraverseMethod {
+    TopDown,
+    BottomUp,
+}
+
 /// Wrap [terms](type.Term.html) with positional information.
 #[derive(Debug, PartialEq, Clone)]
 pub struct RichTerm {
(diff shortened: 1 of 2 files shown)


src/term.rs
```diff
@@ -826,6 +826,11 @@ impl fmt::Display for NAryOp {
     }
 }
 
+pub enum TraverseMethod {
+    TopToBottom,
```
yannham, line 830:
```suggestion
    TopDown,
```

src/term.rs
```diff
@@ -826,6 +826,11 @@ impl fmt::Display for NAryOp {
...
+    BottomToTop,
```
yannham, line 831:
```suggestion
    BottomUp,
```

It's a nitpick, and it can actually sounds less coherent than the original version, but those are the accepted terms.

src/term.rs
```diff
@@ -858,15 +863,24 @@ impl RichTerm {
...
-    /// passed around.
-    pub fn traverse<F, S, E>(self, f: &mut F, state: &mut S) -> Result<RichTerm, E>
+    /// Apply a transformation on a whole term by mapping a function `f` on each node in
+    /// manner defined by the method.
+    /// `f` may return a generic error `E` and use the state `S` which is passed around.
+    pub fn traverse<F, S, E>(
+        self,
+        f: &mut F,
+        state: &mut S,
+        method: &TraverseMethod,
```
yannham, line 873: We should probably make `TraverseMethod` derive `Copy` and never care anymore about references and borrows.

yannham (approved):
A few nitpicks to address, but otherwise LGTM.

litchipi:
Performed the modifications and amended the last commit

PR #471, opened 2021-11-22
[Language server] Recursive records

ysndr:


lsp/nls/src/linearization.rs
```diff
@@ -80,6 +128,146 @@ impl BuildingExt for Linearization<Building<BuildingResource>> {
...
+            }
+
+            if let Some(referenced_id) = referenced_id {
+                self.add_usage(referenced_id, *child_item);
+            }
+        }
+
+        if !unresolved.is_empty() {
+            debug!("unresolved references: {:?}", unresolved);
+            self.resolve_record_references(&unresolved);
```
yannham, line 267: It seems the following is happening:

- the caller of `resolve_record_references` is building a vector, that is then passed the function as a slice. But is not used thereafter by the caller.
- this function loops on the slice, and build a second vector, and then calls itself recursively.

To avoid a bit of fuss, it may be simpler to just pass an owned vector, that you populate and pop in a while loop (or a queue, if you need to preserve order while iterating, or a pointer to a vector). This is not really important though, just a suggestion, to avoid recursion, as you already have to loop anyway.

lsp/nls/src/linearization.rs
```diff
@@ -100,7 +288,7 @@ pub struct AnalysisHost {
     /// in their own scope immediately after the record, which
     /// gives the corresponding record field _term_ to the ident
     /// useable to construct a vale declaration.
-    record_fields: Option<(usize, Vec<Ident>)>,
+    record_fields: Option<(usize, Vec<(usize, Ident)>)>,
```
yannham, line 291: It may be  useful to use a type alias for `usize`, to know what it means. I guess those are UIDs? It's not easy to know at first sight if this is supposed to be a length, an identifier, or something else. I am not saying it has to be done in this PR, but just as a general suggestion.

lsp/nls/src/linearization.rs
```diff
@@ -347,11 +517,28 @@ impl Linearizer<BuildingResource, (UnifTable, HashMap<usize, Ident>)> for Analys
...
     fn linearize(
         self,
-        lin: Linearization<Building<BuildingResource>>,
+        mut lin: Linearization<Building<BuildingResource>>,
         (table, reported_names): (UnifTable, HashMap<usize, Ident>),
     ) -> Linearization<Completed> {
-        let mut lin_ = lin.state.resource.linearization;
         eprintln!("linearizing");
+
+        let defers: Vec<(usize, usize, Ident)> = lin
```
yannham, line 525: Does this mean we make a full pass on the linearization just to find potential deferred usage? Would there be a way to remember indices of deferred item along the way to avoid this? Not saying this should be done in this PR, but just to know.

yannham (approved):
General approach LGTM.

When trying the feature I had a few neovim freezes. But TBH it was hard to reproduce, and I can't be sure this has anything to do with the LSP. Although, I tend to see a pattern where sending a command while the LSP is rebuilding the linearization after a change may mess things up, but I can't be sure. Maybe tracing information will clear this up for us.

Otherwise it seems to work as intended.

lsp/nls/src/linearization.rs
```diff
@@ -64,6 +78,40 @@ impl BuildingExt for Linearization<Building<BuildingResource>> {
...
 
+    fn register_fields(
+        &mut self,
+        record_fields: &HashMap<Ident, RichTerm>,
+        record: usize,
+        scope: Vec<ScopeId>,
+        env: &mut Environment,
+    ) {
+        for (ident, value) in record_fields.into_iter() {
+            let id = self.id_gen().take();
```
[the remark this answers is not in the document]
ysndr: The semantics of the idGen are indeed a bit weird i admit.
`take()` was chosen as it takes the current id, after which it becomes "inaccessible " i.e. the next id is prepared. I thought about `next()` but that sounds like it would give you the increased value. In that sense `take()` resembles a `i++` while `next()` (at least in my mind) is a `++i`

lsp/nls/src/linearization.rs
```diff
@@ -64,6 +78,40 @@ impl BuildingExt for Linearization<Building<BuildingResource>> {
...
+            self.push(LinearizationItem {
+                id,
+                pos: ident.1,
+                // temporary, the actual type is resolved later and the item retyped
+                ty: TypeWrapper::Concrete(AbsType::RowEmpty()),
```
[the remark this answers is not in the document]
ysndr: Yes `dyn` would be possible too, but in the end this part only gets executed for fields  nickel will actually always retype. If for some change in nickel itself this is not done at any point, being record fields "empty row" sounded more appealing than `Dyn`.

lsp/nls/src/linearization.rs
ysndr: Yes that would be a clearer solution.

lsp/nls/src/linearization.rs
ysndr: Yes right now its filtering the linearization for deferred usages. I could track them in the building type if filtering them is too much work.

lsp/nls/src/linearization.rs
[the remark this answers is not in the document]
yannham: IMO the increased value is not relevant to the API of `IdGen`. How does `IdGen` generate unique consecutive IDs is implementation detail. The thing is, as a consumer of `IdGen`, what this function does is to provide you with the next consecutive & unique id available, which is why I tend to think `next` would make more sense. Even more because `take` is used a lot for something quite different in Rust (taking ownership of an inner value, replacing it with a default one).

lsp/nls/src/linearization.rs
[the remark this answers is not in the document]
yannham: The thing is an empty row doesn't make sense as the type of a value. For that matter, rows only make sense when they appear inside a record type. That is, a record type is a sequence of rows, where each row is a field name plus a type. Currently rows are part of `Types` for simplicity, but should probably refactor those as `Types` and `Rows` to make those constraints clearer. My point is, you can have a `Record(EmptyRow())` (this is `{}`). But that a **value** has a naked row as a type is not possible in a normal program, and is not really meaningful. Even more when this type is `EmptyRow`: what would that mean? Outside of the LSP, the typechecker would assign the type `Dyn` to fields outside of a statically typed block.

lsp/nls/src/linearization.rs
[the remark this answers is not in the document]
ysndr: Hm I think it is somewhat important to know that a call to `take` means the **next** call to `id` will give a new/different id `take`. I see though how "take" is an overloaded term in rust, but so is "next" (especially thinking about the semantics). how about "advance" or "get_and_advance"?

lsp/nls/src/linearization.rs
ysndr: Its technically in the nls crate (by implementation). It becomes very mixed up with nickel and nls as I would need to pass down the trace identifier (currently Req. Id) but that looks to create more trouble than it solves maybe

yannham: Got it, let's forget it for now then.

lsp/nls/src/linearization.rs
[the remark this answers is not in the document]
yannham: Oh, because there's also a method to get the current `id`. My bad then. I personally like `advance` better, yes (a bit like `remove`, it does something and return the state before we did the thing). But you know, names and color. Do as you think is best.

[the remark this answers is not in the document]
ysndr: I documented the IdGen methods a bit and call tehm `get`/`get_and_advance` that should be clearer now

lsp/nls/src/linearization.rs
ysndr: Yes its Inearization item id's (which are indexes during the `Building` phase but thats a different matter)
I agree, we should have these be more clear throughout the module

yannham:
@ysndr it's good to merge then, if you're done.

PR #472, opened 2021-11-22
[Overriding]Reversible thunks

yannham:
Preparatory PR for the implementation of #330. Add support for reversible thunks, although they aren't put to use yet.

## Context

#330 requires a new device for evaluation. In this model, both recursive fields and merge expressions may be interpreted differently depending on the context. The simplest example is the one of overriding:

```
let r = { foo = bar + 1, bar | default = 0} in
r.foo + (r & {bar = 1}).foo
```

Here, the same original expression inside `foo` must be evaluated differently: as `1` in `r.foo`, but as `2` in `(r & {bar = 1}).foo`. The same thing happens with merging and custom merge functions: a sub expression `x1 & x2` may evaluate differently depending on a `| merge f` annotation being there in the same merge tree.

## A call-by-name semantics with caching

At first it doesn't seem obvious how to reconcile this with the usual call-by-need model based on thunks that are updated once in a destructive manner. Rather, we should understand our semantics as fundamentally call-by-name, be it for merge or recursive records. Then, the thunks are just an external caching (or memoization) mechanism for performance matter.

In the classical call-by-need setting, it just turns out that this caching mechanism is really simple: the cache is destructively updated once with the evaluated form of an expression, and never invalidated anymore. Morally, we are memoizing functions `unit -> a`.

In our setting, the fields of a recursive record (resp. merge expressions) can be seen rather as function `Self -> a` where `Self` represents the current, final record (resp. `MergeFun -> a`). Caching must thus be slightly more general.

## Reversible thunks

The proposed solution, as implemented in this PR, is to add a second type of thunks, called reversible thunks. Those not only cache a value, but also store the original expression, that can be restored even once updated. The `Thunk` type can either contain a standard thunk or a reversible thunk, such that all the code operating on thunks doesn't need to know about them and remains unchanged: we just add a `restore` capability, that either restores the original expression for reversible thunks, or is a simple `clone` for standard thunks. Reversible thunks will then have to be introduced explicitly for expressions that need to (record fields and merge expressions) and restored explicitly before operations that are known to potentially invalidate them (merging).

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -190,10 +190,9 @@ impl ThunkData {
     pub fn into_closure(self) -> Closure {
         match self.inner {
             InnerThunkData::Standard(closure) => closure,
-            InnerThunkData::Reversible { orig, cached: None } => match Rc::try_unwrap(orig) {
-                Ok(inner) => inner,
-                Err(rc) => (*rc).clone(),
-            },
+            InnerThunkData::Reversible { orig, cached: None } => {
+                Rc::try_unwrap(orig).unwrap_or_else(|rc| (*rc).clone())
+            }
             InnerThunkData::Reversible {
                 cached: Some(cached),
                 ..
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -130,7 +130,7 @@ pub struct ThunkData {
 pub enum InnerThunkData {
     Standard(Closure),
     Reversible {
-        orig: Closure,
+        orig: Rc<Closure>,
         cached: Option<Closure>,
     },
 }
@@ -147,7 +147,10 @@ impl ThunkData {
     /// Create new reversible thunk data.
     pub fn new_rev(orig: Closure) -> Self {
         ThunkData {
-            inner: InnerThunkData::Reversible { orig, cached: None },
+            inner: InnerThunkData::Reversible {
+                orig: Rc::new(orig),
+                cached: None,
+            },
             state: ThunkState::Suspended,
         }
     }
@@ -172,13 +175,14 @@ impl ThunkData {
         match self.inner {
             InnerThunkData::Standard(ref mut closure) => closure,
             InnerThunkData::Reversible {
-                ref mut orig,
-                cached: None,
-            } => orig,
-            InnerThunkData::Reversible {
-                cached: Some(ref mut closure),
-                ..
-            } => closure,
+                ref orig,
+                ref mut cached,
+            } => {
+                if cached.is_none() {
+                    *cached = Some((**orig).clone());
+                }
+                cached.as_mut().unwrap()
+            }
         }
     }
 
@@ -186,7 +190,10 @@ impl ThunkData {
     pub fn into_closure(self) -> Closure {
         match self.inner {
             InnerThunkData::Standard(closure) => closure,
-            InnerThunkData::Reversible { orig, cached: None } => orig,
+            InnerThunkData::Reversible { orig, cached: None } => match Rc::try_unwrap(orig) {
+                Ok(inner) => inner,
+                Err(rc) => (*rc).clone(),
+            },
             InnerThunkData::Reversible {
                 cached: Some(cached),
                 ..
@@ -213,7 +220,13 @@ impl ThunkData {
                 inner: InnerThunkData::Standard(closure.clone()),
                 state: self.state,
             },
-            InnerThunkData::Reversible { ref orig, .. } => ThunkData::new_rev(orig.clone()),
+            InnerThunkData::Reversible { ref orig, .. } => ThunkData {
+                inner: InnerThunkData::Reversible {
+                    orig: Rc::clone(orig),
+                    cached: None,
+                },
+                state: self.state,
+            },
         }
     }
 }
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -131,7 +131,7 @@ pub enum InnerThunkData {
     Standard(Closure),
     Reversible {
         orig: Rc<Closure>,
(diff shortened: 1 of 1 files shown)


aspiwack:
Why do you need non-reversible thunks? This sounds kind of dangerous.

What happen if you have a field whose value refers to a non-reversible thunk whose value depends on another field of the same record. I guess we may want to restore the intermediate thunk as well? I'm not sure, is there a good theory of how all this go? Or is it just best effort while we are waiting to design a clean solution?

src/eval.rs
```diff
@@ -117,23 +117,130 @@ pub enum ThunkState {
...
+            InnerThunkData::Reversible { ref mut cached, .. } => *cached = Some(new),
+        }
+
+        self.state = ThunkState::Evaluated;
+    }
+
+    /// Create fresh unevaluated thunk data from `self`, restored to its original state before the
+    /// first update. For standard thunk data, the content is unchanged and the state is conserved: in
+    /// this case, `restore()` is the same as `clone()`.
+    pub fn restore(&self) -> Self {
```
aspiwack, line 216: I'm curious, why not the following?
```suggestion
    pub fn restore(&mut self) -> () {
```

aspiwack (commented):
Some idle comments about the code.

yannham:
>Why do you need non-reversible thunks? This sounds kind of dangerous.

The idea is to avoid retaining references to original expressions for normal thunks, when we don't need to, which should be the majority of cases.

>What happen if you have a field whose value refers to a non-reversible thunk whose value depends on another field of the same record. I guess we may want to restore the intermediate thunk as well? I'm not sure

I don't think so, at least for recursive fields. Take the following case e.g.

```
{
  foo = let y = bar + 1 in y,
  bar | default = 1,
} & {bar = 2}
```

Only the thunks captured from the outside environment might need to be reversed as well. Thunks introduced inside `foo`, like `y`, may be non-reversible thunk but they will be restored as a fresh `let` expression, and thus a fresh thunk with the original expression `bar + 1`. The thunks captured from the environment are either the other recursive fields (that are restored too, on merging), or things that can't refer to recursive fields because they are coming from the outer scope, and thus don't need to be restored.

For merging and custom functions the situation is different. I have a criterion in mind, but Iet me write it down somewhere indeed, because things are not totally trivial. At worst, making all the thunk reversible may be wasteful but should at least be indeed correct.

>I'm not sure, is there a good theory of how all this go? Or is it just best effort while we are waiting to design a clean solution?

I've put a bit of thinking into it, but I wouldn't go as far as saying there is a good theory. What about continuing merging this branch and next PRs for overriding on a separate branch `overriding`, so that I can continue to experiment while elaborating a more solid semantics?

src/eval.rs
```diff
@@ -117,23 +117,130 @@ pub enum ThunkState {
...
+/// - A reversible thunk, that can be restored to its original expression. Used to implement
+/// recursive merging of records and overriding (see the [RFC
+/// overriding](https://github.com/tweag/nickel/pull/330))
+#[derive(Clone, Debug, PartialEq)]
+pub enum InnerThunkData {
+    Standard(Closure),
+    Reversible {
+        orig: Rc<Closure>,
+        cached: Option<Closure>,
+    },
```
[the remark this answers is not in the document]
yannham: Yes, I added the `Rc` lastly by thinking "we should share original expression among sibling", but indeed we might do that for `cached` as well.

src/eval.rs
```diff
@@ -117,23 +117,130 @@ pub enum ThunkState {
...
+            }
+        }
+    }
+
+    /// Consume the data and return the cached closure.
+    pub fn into_closure(self) -> Closure {
+        match self.inner {
+            InnerThunkData::Standard(closure) => closure,
+            InnerThunkData::Reversible { orig, cached: None } => {
+                Rc::try_unwrap(orig).unwrap_or_else(|rc| (*rc).clone())
```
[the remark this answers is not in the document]
yannham: I don't even think that's the case. Because we have ownership over `self`, we can explicitly `std::mem::drop(orig)` before trying to unwrap `cached`. This should avoid cloning in the same conditions: when this thunk is only one to refer to the rc-ed expression.

src/eval.rs
yannham: Because we want to create a new, separated "cache entry" (thunk) when reverting. For example:

```
let r1 = {foo = bar + 1, bar | default = 1} in
r1.foo + (let r2 = r1 & {bar = 2} in r2.foo + r2.foo) + r1.foo
```

On the first line, we allocate `r1` with a  thunk for `foo` inside, that is forced on the first operand of `+` to `2`. Then, we allocate an `r2` where we need to revert the thunks of fields of `r1`. But there is no reason to "revert" the thunk that is currently referenced by `r1.foo`, as it will always evaluate to `2`. We can't even share the same thunk at all between `r1` and `r2`: otherwise any evaluation of `r1.foo` or `r2.foo` could poison the cache of the other one. Rather, `r2` should allocate a fresh thunk, with the original expression taken from `r1.foo`, so that it can have its own independent caching for `foo`. This is what `restore` provides. But maybe there's a better name?

This also the reason for `Rc`ing `orig`, as several thunks will potentially share the same original expression but with different, independent cached values (hence the cached value is currently owned, though your suggestion is still interesting just for the sake of making code simpler)

aspiwack: I see it now, thanks.

aspiwack (approved):
I think that, at the very least, we can state with confidence that this is not worse than the current situation. So there is no risk in merging.

Do as you feel like is best for the code review comments.

Issue #81, opened 2020-06-01
Destructuring


yannham:
Records and arrays are the two ubiquitous and fundamental data structures of a configuration language, be it Nix expressions, Jsonnet, Cue, or Nickel. It is useful in practice to be able to easily extract a few specific fields from a record (which may be located deep inside), or a few elements at specific positions from a list. To this end, we propose to add destructuring to Nickel, a syntax that allows to directly extract sub-elements from records or lists by extending let-bindings to bind not only variables, but *patterns*.

There are good examples out there to take inspiration from, such as pattern matching (restricted to records and lists) in Haskell or OCaml, or the closest to the following proposal, destructuring in Javascript.

Nix expressions also offer destructuring facilities:

- *Argument destructuring*, which can only be done on the argument of a function 
- `let inherit`, which can be confused with its twin `{ inherit ...; }`, and cannot specify default values for missing fields

Both cannot handle nested records nor bind a field to a variable with a different name. Lists cannot be destructured either. We intend to overcome these limitations in Nickel.

In the following, we provide a concrete proposal, as a starting point for discussion.

## Motivating examples
We would like to write this kind of expressions:

```
let {pname; src=srcSome; meta={description};} = somePkg in foo
/* bind pname to somePkg.pname,
 * srcSome to somePkg.src,
 * and description to somePkg.meta.description in foo */

let {foo ? "default"; bar;} = {bar=1;} in baz
/* bind foo to "default" and bar to 1 in baz */

let [fst, snd, {bar}] = [1, 2, {bar="bar";}, "rest", "is", "ignored"] in foo
/* bind fst to 1, snd to 2 and bar to "bar" in foo */

let [head, ...tail] = [1, 2, 3, 4] in foo
/* bind head to 1 and tail to [2, 3, 4] in foo */

let [_, _, third, fourth] = [1, 2, 3, 4] in foo
/* bind third to 3 and fourth to 4 in foo */

let {a; ...rest} = {a=1; b=2; c=3;} in foo
/* bind a to 1 and rest to {c=2; d=3;} in foo */

let f = fun {x; y; z;} => x + y + z in foo
/* bind x, y and z to the corresponding fields of the argument in the body of f */
```

## Description
We propose JavaScript-like destructuring, with the `@` from Nix for capturing the whole match, featuring:

- **Nested patterns**: we can destructure deeply nested records and arrays in a
    single pattern
- **Rebinding**: we can bind the field of a record to a variable with a different name
[...]

PR #473, opened 2021-11-23
Destructuring

francois-caddet:
relate to #81 
Full implementation of the record destructuring as described in #81 . The list destructuring is not implemented yet. 
Typechecking is not performed at this time. will be done in future PR


src/grammar.lalrpop
```diff
@@ -54,6 +55,14 @@ MetaAnnotAtom: MetaValue = {
     },
 };
 
+DefaultAnnot: MetaValue = "?" <t: Term> => MetaValue {
+    doc: None,
+    types: None,
+    contracts: Vec::new(),
+    priority: MergePriority::Default,
+    value: Some(t)
+};
```
yannham, line 64: Just a nitpick, but I think you can write this as
```
MetaValue {
    priority: MergePriority::Default,
    value: Some(t),
    ..Default::default()
}
```

src/term.rs
```diff
@@ -188,14 +191,16 @@ pub struct Contract {
 }
 
 #[derive(Debug, PartialEq, Clone)]
-pub struct MetaValue {
+pub struct AbsMetaValue<V> {
```
yannham, line 194: I haven't seen `AbsMetaValue` used anywhere, in the end. Did you end up not needing it?

src/term.rs
```diff
@@ -1214,6 +1232,27 @@ pub mod make {
         Term::Let(id.into(), t1.into(), t2.into()).into()
     }
 
+    pub fn let_pat<I, D, T1, T2>(id: Option<I>, pat: D, t1: T1, t2: T2) -> RichTerm
+    where
+        T1: Into<RichTerm>,
+        T2: Into<RichTerm>,
+        D: Into<Destruct>,
+        I: Into<Ident> + fmt::Debug,
```
yannham, line 1240: I imagine the `fmt::Debug` constraint is a leftover?

src/transformations.rs
```diff
@@ -10,6 +10,111 @@ use std::path::PathBuf;
...
+///         let d = x.b in ...))
+/// ```
+pub mod desugar_destructuring {
+    use super::{Ident, RichTerm, Term};
+    use crate::destruct::{Destruct, Match};
+    use crate::term::make::{op1, op2};
+    use crate::term::MetaValue;
+    use crate::term::{BinaryOp::DynRemove, UnaryOp::StaticAccess};
+
+    pub fn desugar_with_contract(rt: RichTerm) -> RichTerm {
```
yannham, line 38: Same, please document public functions.

src/transformations.rs
```diff
@@ -10,6 +10,111 @@ use std::path::PathBuf;
...
+                    destruct_term(x.clone(), pat.clone(), drop_fields(x, pat, body)),
+                ),
+                pos,
+            )
+        } else {
+            rt
+        }
+    }
+
+    fn drop_fields(x: Ident, pat: Destruct, body: RichTerm) -> RichTerm {
```
yannham, line 79: Just wondering, is it really needed to drop matched fields? Couldn't we just keep the original value around?

src/transformations.rs
```diff
@@ -10,6 +10,111 @@ use std::path::PathBuf;
...
+                    op2(DynRemove(), Term::Str(i.to_string()), x)
+                }
+            }),
+            body,
+        )
+        .into()
+    }
+
+    fn destruct_term(x: Ident, pat: Destruct, body: RichTerm) -> RichTerm {
+        let pos = body.pos.clone();
```
yannham, line 99:
```suggestion
        let pos = body.pos;
```

src/typecheck/mod.rs
```diff
@@ -616,6 +616,17 @@ fn type_check_<S, E>(
             envs.insert(x.clone(), ty_let);
             type_check_(state, envs, lin, linearizer, strict, rt, ty)
         }
+        Term::LetPattern(x, pat, re, rt) => {
```
yannham, line 619: @ysndr do we need to do something special here, with respect to the LSP?

tests/common.rs
```diff
@@ -7,3 +7,9 @@ pub fn eval(s: impl std::string::ToString) -> Result<Term, Error> {
     let mut p = Program::new_from_source(Cursor::new(s.to_string()), "test").unwrap();
     p.eval()
 }
+
+pub fn eval_file(f: &str) -> Result<Term, Error> {
+    let mut p =
+        Program::new_from_file(format!("{}/tests/{}", env!("CARGO_MANIFEST_DIR"), f)).unwrap();
```
yannham, line 13: Although it is probably not important for tests, why `unwrap()` when we can bubble up the error using `?`

yannham (approved):
LGTM globally, but I left a good chunk of comments and nitpicks. Also, your tests always start with an empty line, for some reason.

The more I think about it, and the more I imagine we will want to have strict checks instead of relying on lazy contracts for pattern matching, for early error reporting. But we decided to go this route because this was easier, and for now, this will do.

src/transformations.rs
```diff
@@ -10,6 +10,111 @@ use std::path::PathBuf;
 
 generate_counter!(FreshVarCounter, usize);
 
+/// # Desugarise destructuring.usize
```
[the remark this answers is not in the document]
francois-caddet: autocomplet problem ;)

Cargo.lock
```diff
@@ -440,9 +440,9 @@ checksum = "ab5ef0d4909ef3724cc8cce6ccc8572c5c817592e9285f5464f8e86f8bd3726e"
 
 [[package]]
 name = "heck"
-version = "0.3.3"
+version = "0.3.2"
```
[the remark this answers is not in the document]
francois-caddet: no idea... sorry... will try to revert it

src/destruct.rs
```diff
@@ -0,0 +1,75 @@
...
+                    Term::Record(
+                        self.inner()
+                            .into_iter()
+                            .map(|m| m.as_meta_field())
+                            .collect(),
+                        RecordAttrs { open },
+                    )
+                    .into(),
+                )),
+                label: Label::dummy(),
```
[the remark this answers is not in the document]
francois-caddet: Will see, but it's working like this. So may be improved in a next PR. I think it's errors reporting stuff and as you saw, almost nothing as been done for that in this PR ;)

src/grammar.lalrpop
```diff
@@ -81,13 +90,13 @@ pub ExtendedTerm: ExtendedTerm = {
             t
         };
 
-        ExtendedTerm::ToplevelLet(id, t)
+        ExtendedTerm::ToplevelLet(pat.0.unwrap(), t)
```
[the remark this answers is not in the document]
francois-caddet: Sure, I wanted to allow it but the change to do is bigger than expected. so you're right, I will change back the gramar.

src/term.rs
francois-caddet: Oups, finaly I didn't.

yannham: I mean, now that it's done it doesn't hurt to keep it, but I was just wondering.

francois-caddet: I finaly droped it. Because I was not realy convinced if the `Option` should be part of the `AbsMetaVlue` or as Type parameter. Better if we make it when we realy will need it. It's 2 lines to update and will not break the previous code.

tests/common.rs
francois-caddet: Because we can not right now ;) It's the same reason we use `unwrap` in the other function of this module. `Program::new_from_...` functions return a `std::io::Result`. It will be better to replace it by a `Result<_, errors::IOError>` but does not concern this PR I think and is not a small change. I'll open an issue for that.

src/typecheck/mod.rs
ysndr: Maybe add a new new declaration for the variables brought to scope here otherwise they are more or less ignored.

src/transformations.rs
francois-caddet: What do you mean? I actualy gave a bad meaning name to this function. I changed it to `bind_open_field`. Do you think it preferable to pass the `pat` or any other param by ref?

tests/common.rs
yannham: I thought there was a top-leve `IOError`, but it's only for imports. That makes sense

src/typecheck/mod.rs
francois-caddet: Ok, will see

src/eval.rs
```diff
@@ -875,6 +875,7 @@ pub fn subst(rt: RichTerm, global_env: &Environment, env: &Environment) -> RichT
 
                 RichTerm::new(Term::Let(id, t1, t2), pos)
             }
+            Term::LetPattern(..) => unreachable!(),
```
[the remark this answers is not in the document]
francois-caddet: You're right. I replaced it by a panic. May any error kind be better, but which one?
Actually it make me think to the old idea of having 2 ASTs (before and after transformation). ;)

src/destruct.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn as_meta_field(self) -> (Ident, RichTerm) {
+        match self {
+            Match::Assign(id, m, (_, Destruct::Empty)) | Match::Simple(id, m) => {
+                (id, Term::MetaValue(m).into())
+            }
+            Match::Assign(id, m, (_, d @ Destruct::Record(..))) => (
+                id.clone(),
+                Term::MetaValue(MetaValue::flatten(m, d.as_contract())).into(),
+            ),
+            _ => unimplemented!(),
```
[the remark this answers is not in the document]
francois-caddet: you mean in place of the `unimplemented!(),``?
No, because the only other match possibility is with a `Destruct::List` in the last `Match` field. which is realy not implement right now.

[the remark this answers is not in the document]
yannham: So there is a pattern matching that would make the interpreter panics ? What would this pattern be?

PR #474, opened 2021-11-23
[Stdlib] Add blameWith and fromPred

yannham:
This PR adds two functions to the contracts stdlib, adds missing documentation for the remaining ones, and updates the examples accordingly.

## blameWith

The current way to trigger a contract violation with an error is to compose the `blame` function and the `tag` function, which is unncessarily verbose: `contracts.blame (contracts.tag "invalid FizzBuzz number" label)`. This PR adds a `blameWith : Str -> Lbl -> a` function to write `contracts.blameWith "invalid FizzBuzz number" label` instead.

## fromPred 

Independently, we also add the `fromPred : (Dyn -> Bool) -> (Lbl -> Dyn -> Dyn)` function, that turns a boolean predicate into a contract. Currently, one has to write irrelevant boilerplate (the part about returning value and blaming in case of error):

```
let IsZero = fun label value =>
  if value == 0 then value
  else contracts.blame label in
//...
```

With `fromPred`, this becomes:

```
let IsZero = contracts.fromPred (fun value => value == 0) in
```

### DIFF
--- a/stdlib/contracts.ncl
+++ b/stdlib/contracts.ncl
@@ -107,7 +107,7 @@
         ```nickel
         let IsZero = fun label value =>
           if value == 0 then value
-          else contracts.blame label in
+          else contracts.blameWith "Not zero" label in
         0 | #IsZero
         ```
         "#m
--- a/stdlib/contracts.ncl
+++ b/stdlib/contracts.ncl
@@ -73,7 +73,80 @@
       else %blame% (%tag% "extra field `#{%head% (%fieldsOf% t)}`" l),
 
   contracts = {
-    blame = fun l => %blame% l,
-    tag = fun msg l => %tag% msg l,
+    blame
+      | doc m#"
+        Raise blame for a given label.
+
+        Type: `forall a. Lbl -> a`
+        (for technical reasons, this element isn't actually statically typed)
+
+        Blame is the mechanism to signal contract violiation in Nickel. It ends
+        the program execution and print a detailed report thanks to the
+        information tracked inside the label.
+
+        For example:
+        ```nickel
+        IsZero = fun label value =>
+          if value == 0 then value
+          else contracts.blame label
+        ```
+        "#m
+      = fun l => %blame% l,
+    blameWith
+      | doc m#"
+        Raise blame with respect to a given label and a custom error message.
+
+        Type: `forall a. Str -> Lbl -> a`
+        (for technical reasons, this element isn't actually statically typed)
+
+        Same as `blame`, but take an additional custom error message that will be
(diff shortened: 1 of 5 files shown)


stdlib/contracts.ncl
```diff
@@ -73,7 +73,80 @@
...
+
+        Same as `blame`, but take an additional custom error message that will be
+        displayed as part of the blame error. `blameWith msg l` is equivalent to
+        `blame (tag msg l)
+
+        For example:
+        ```nickel
+        let IsZero = fun label value =>
+          if value == 0 then value
+          else contracts.blame label in
```
regnat, line 110: I guess you meant 
```suggestion
          else contracts.blameWith "Not zero" label in
```

stdlib/contracts.ncl
```diff
@@ -73,7 +73,80 @@
...
+        Type: `(Dyn -> Bool) -> (Lbl -> Dyn -> Dyn)`
+        (for technical reasons, this element isn't actually statically typed)
+
+        For example:
+        ```
+        let IsZero = contracts.fromPred (fun x => x == 0) in
+        0 | #IsZero
+        ```
+        "#m
+      = fun pred l v => if pred v then v else %blame% l,
```
regnat, line 128: Should this also take a tag as argument to encourage writing meaningful error messages?

```
let IsZero = contracts.fromPred "Is not zero" (fun x => x == 0) in
...
```

regnat (changes requested):

stdlib/contracts.ncl
yannham: I wondered the same thing. But note that a predicate can only fail in one way (if it evaluates to false). Because blame errors report the original contract annotation, I think in most case a generic error message like "argument is non zero" wouldn't add much to, say, just the name of the contract `NonZero`. Or at least that probably we shouldn't force users to attach such a message when the information that the contract `NonZero` or `GreaterThan 10` failed is sufficient.

However, maybe having a variant `fromPredWith` (or whatever we call it) is a good idea. Would you like me to add it to this PR?

regnat: Oh yes, the name appears in the report, ofc.

> However, maybe having a variant fromPredWith (or whatever we call it) is a good idea. Would you like me to add it to this PR?

You’ve just convinced me that it wasn’t necessary it shouldn’t be necessary in most cases :stuck_out_tongue: . So I’d say that we can keep it out for now and see whether `tag "foo" (fromPred p)` becomes a commonly needed pattern

regnat (approved):

PR #475, opened 2021-11-25
Flip stdlib functions to accommodate use with the |> operator

silverraven691:
This is influenced by how Elm designed their standard library: https://package.elm-lang.org/packages/elm/core/latest/

These remain untouched for now:
```
lists.ncl
85:    concat : forall a. List a -> List a -> List a
nums.ncl
186:    pow : Num -> Num -> Num
```

If we compare with Elm:
- `lists.concat` is called `Array.append` and has the same order (there is a `Array.concat` is that is equal to Nickel's `lists.flatten`).
- `nums.pow` is flipped and is named `powBy`.

I've left those two out because they involve changing the name of a function, which I decided not to do until further discussion. Following Elm's method might work out, we could flip them and rename them to `lists.appendWith`, `nums.powBy`. We may add a `lists.prependWith` for symmetry.

Note that these changes may have considerable consequences on the "flavor" of the language in the long run.
It might be valid to change the documentation examples to better demonstrate usage with `|>`, especially in cases where the call site looks better than normal application

### DIFF
--- a/examples/config-gcc/config-gcc.ncl
+++ b/examples/config-gcc/config-gcc.ncl
@@ -6,7 +6,7 @@ let GccFlag =
   fun label value =>
   if builtins.isStr value then
     if strings.length value > 0 &&
-      lists.any (fun x => x == strings.substring value 0 1) available then
+      lists.any (fun x => x == strings.substring 0 1 value) available then
       value
     else
       contracts.blameWith "unknown flag #{value}" label
--- a/examples/config-gcc/config-gcc.ncl
+++ b/examples/config-gcc/config-gcc.ncl
@@ -28,7 +28,7 @@ let Path =
   let pattern = m#"^(.+)/([^/]+)$"#m in
   fun label value =>
     if builtins.isStr value then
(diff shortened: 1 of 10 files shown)


Profpatsch:
Yes please, in a currying language the last argument should always the the one “under modification”, exactly because it allows for chaining operations more easily.

yannham:
I'm personally in favor of this change in principle. Agreed that we shouldn't change `concat` and `pow`.

Let's have other opinions before moving further. If we decide to do so, we'll need to fix all the usages of those functions, both in the stdlib itself (e.g. `fold` is most probably used inside the `lists` stdlib itself) and in tests/examples/documentation. Right now the CI fails because this hasn't been done.

aspiwack:
This looks generally good to me.

yannham (changes requested):
@silverraven691 thanks again for the contribution! I think we can go further. Changes look fine but all usages of those flipped functions need to be fixed.

yannham (approved):

PR #479, opened 2021-12-02
can use arbitrary Str in enums deffinitions

### LINKED ISSUE #460: [Dogfooding] Escaping enums
While dogfooding on the [mantis repo](https://github.com/input-output-hk/mantis-ops), there are several fields with a finite say of strings as possible values (ex: [here](https://github.com/input-output-hk/mantis-ops/blob/81ad4cb0ec5ff9faeb9065a2ddac060cef6452ac/schemas/nomad/types.cue#L54)). It's nice to represent this type/contract with enums in Nickel, like:

```
{
  protocol | <http, ftp, https>
}
```

As the typechecker is then aware of those, can check match exhaustiveness, and so on. However it isn't possible when one of the alternative is not syntactically an identifier.

Having a way of escaping non-identifier characters inside enum tags, e.g. by surrounding them by quotes:

```
{
  operator | <"Foo!bar", "Foo!baz", "!="> = `"Foo!bar",
}
```

We can also encode this using a custom contract (most probably, this kind of combinator will be in the stdlib at some point):

```
let IsIn = fun values label value =>
  if lists.elem value values then 
    value 
  else
    contracts.blame label
in
{
  operator | #(IsIn ["Foo!bar", "Foo!baz", "!="])
}
```

But:

1. It is less consistent (some enums will be defined with the builtin type, some others with this custom contracts, without beginners knowing why, maybe)
2. As mentioned, the typechecker don't know what to do with that (no match, no typechecking, no pattern matching exhaustiveness, etc.)

A counter-argument could be that enums are meant for things that resembles simple labels, and if one of the alternative for a field is `"!="`, then maybe it shouldn't be encoded as an enum to begin with.

francois-caddet:
impl and close #460

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -368,7 +368,7 @@ UOp: UnaryOp = {
 };
 
 switch_case: SwitchCase = {
-    <id: Ident> "=>" <t: Term> => SwitchCase::Normal(id, t),
+    "`" <id: EnumTag> "=>" <t: Term> => SwitchCase::Normal(id, t),
     "_" "=>" <t: Term> => SwitchCase::Default(<>),
 }
 
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -505,7 +505,7 @@ subType : Types = {
             .fold(
                 Types(
                     match tail {
-                        Some(s) => AbsType::Var(s),
+                        Some(id) => AbsType::Var(id),
                         None => AbsType::RowEmpty(),
                     }
                 ),
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -183,7 +183,7 @@ Atom: RichTerm = {
     Bool => RichTerm::from(Term::Bool(<>)),
     StrChunks,
     Ident => RichTerm::from(Term::Var(<>)),
-    "`" <Ident> => RichTerm::from(Term::Enum(<>)),
+    "`" <EnumField> => RichTerm::from(Term::Enum(<>)),
     "{" <fields: (<RecordField> ",")*> <last: RecordLastField?> "}" => {
         let (last_field, attrs) = match last {
             Some(RecordLastField::Field(f)) => (Some(f), Default::default()),
@@ -317,6 +317,12 @@ ChunkExpr: StrChunk<RichTerm> = HashBrace <t: WithPos<Term>> "}" => StrChunk::Ex
 HashBrace = { "#{", "multstr #{" };
 
 StaticString: String = StringStart <s: ChunkLiteral> StringEnd => s;
+
+EnumField: Ident = {
+    <Ident>,
+    <StaticString> => <>.into(),
+}
+
 
 ChunkLiteralPart: ChunkLiteralPart<'input> = {
     "str literal" => ChunkLiteralPart::Str(<>),
@@ -488,7 +494,7 @@ subType : Types = {
     <Ident> => Types(AbsType::Var(<>)),
     "#" <WithPos<RecordOperand>> => Types(AbsType::Flat(<>)),
     "(" <Types> ")",
-    "<" <rows:(<Ident> ",")*> <last: (<Ident>)?> <tail: ("|" <Ident>)?> ">" => {
+    "<" <rows:(<EnumField> ",")*> <last: (<EnumField>)?> <tail: ("|" <EnumField>)?> ">" => {
         let ty = rows.into_iter()
             .chain(last.into_iter())
             // As we build row types as a linked list via a fold on the original
@@ -499,7 +505,7 @@ subType : Types = {
             .fold(
                 Types(
                     match tail {
-                        Some(id) => AbsType::Var(id),
+                        Some(s) => AbsType::Var(s),
                         None => AbsType::RowEmpty(),
                     }
                 ),
--- a/src/operation.rs
+++ b/src/operation.rs
@@ -321,7 +321,7 @@ fn process_unary_operation(
                 };
 
                 cases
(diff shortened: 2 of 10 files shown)


src/grammar.lalrpop
```diff
@@ -183,7 +183,7 @@ Atom: RichTerm = {
     Bool => RichTerm::from(Term::Bool(<>)),
     StrChunks,
     Ident => RichTerm::from(Term::Var(<>)),
-    "`" <Ident> => RichTerm::from(Term::Enum(<>)),
+    "`" <EnumField> => RichTerm::from(Term::Enum(<>)),
```
yannham, line 186:
```suggestion
    "`" <EnumTag> => RichTerm::from(Term::Enum(<>)),
```

src/grammar.lalrpop
```diff
@@ -318,6 +318,12 @@ HashBrace = { "#{", "multstr #{" };
 
 StaticString: String = StringStart <s: ChunkLiteral> StringEnd => s;
 
+EnumField: Ident = {
```
yannham, line 321:
```suggestion
EnumTag: Ident = {
```

src/grammar.lalrpop
```diff
@@ -488,7 +494,7 @@ subType : Types = {
     <Ident> => Types(AbsType::Var(<>)),
     "#" <WithPos<RecordOperand>> => Types(AbsType::Flat(<>)),
     "(" <Types> ")",
-    "<" <rows:(<Ident> ",")*> <last: (<Ident>)?> <tail: ("|" <Ident>)?> ">" => {
+    "<" <rows:(<EnumField> ",")*> <last: (<EnumField>)?> <tail: ("|" <EnumField>)?> ">" => {
```
yannham, line 497:
```suggestion
    "<" <rows:(<EnumTag> ",")*> <last: (<EnumTag>)?> <tail: ("|" <EnumTag>)?> ">" => {
```

src/grammar.lalrpop
```diff
@@ -499,7 +505,7 @@ subType : Types = {
             .fold(
                 Types(
                     match tail {
-                        Some(id) => AbsType::Var(id),
+                        Some(s) => AbsType::Var(s),
```
yannham, line 508: It's not very important, but why the name change?

src/operation.rs
```diff
@@ -321,7 +321,7 @@ fn process_unary_operation(
                 };
 
                 cases
-                    .remove(&en)
+                    .remove(&en.clone().into())
```
yannham, line 324: Why do we need to clone, convert to then only take a reference back?

src/operation.rs
```diff
@@ -894,7 +894,7 @@ fn process_unary_operation(
                 Term::Num(n) => Ok(Term::Str(n.to_string())),
                 Term::Str(s) => Ok(Term::Str(s)),
                 Term::Bool(b) => Ok(Term::Str(b.to_string())),
-                Term::Enum(id) => Ok(Term::Str(id.to_string())),
+                Term::Enum(s) => Ok(Term::Str(s.to_string())),
```
yannham, line 897: Same question as before. The thing inside `Term::Enum` is still of type `Ident`, not of type `String`.

src/term.rs
```diff
@@ -397,7 +397,14 @@ impl Term {
             }
             Term::Fun(_, _) => String::from("<func>"),
             Term::Lbl(_) => String::from("<label>"),
-            Term::Enum(s) => format!("`{}", s),
+            Term::Enum(Ident(s, _)) => {
```
yannham, line 400: We should avoid matching on the actual structure of `Ident`, because it may very well change in the future. Instead, rather match an `id` and then use conversion methods like `to_string`.

src/operation.rs
francois-caddet: same as before

src/operation.rs
francois-caddet: I was using string before and forgot to roll back this part.

src/term.rs
francois-caddet: Ok fixing it in next commit.

tests/pass/basics.ncl
```diff
@@ -32,6 +32,6 @@ let Assert = fun l x => x || %blame% l in
 (1/4 + 1/4 - 1/4 + 1/4 < 1/2 == false | #Assert) &&
 
 // This test checks that the terms of a switch are closured
-(let x = 3 in (switch { foo => 1, _ => x, } (3 + 2)) == 3) &&
+(let x = 3 in (switch { `foo => 1, _ => x, } (3 + 2)) == 3) &&
```
yannham, line 35:
```suggestion
(let x = 3 in (switch { `foo => 1, _ => x} (3 + 2)) == 3) &&
```

tests/pass/typechecking.ncl
```diff
@@ -53,25 +53,25 @@ let typecheck = [
...
     : Num),
 
   let f : forall r. <blo, ble | r> -> Num = fun x =>
-    switch {blo => 1, ble => 2, _ => 3} x in
+    switch {`blo => 1, `ble => 2, _ => 3} x in
             (f `bli : Num),
 
   let f : forall r. (forall p. <blo, ble | r> -> <bla, bli | p>) =
-    fun x => switch {blo => `bla, ble => `bli, _ => `bla, } x in
+    fun x => switch {`blo => `bla, `ble => `bli, _ => `bla, } x in
```
yannham, line 74:
```suggestion
    fun x => switch {`blo => `bla, `ble => `bli, _ => `bla} x in
```

yannham (approved):

PR #481, opened 2021-12-03
Apply clippy's suggestions (most of)

yannham:


src/typecheck/linearization.rs
```diff
@@ -240,11 +240,10 @@ impl Completed {
 
                 self.scope_mapping
                     .get(scope)
-                    .map_or_else(|| Vec::new(), Clone::clone)
+                    .map_or_else(Vec::new, Clone::clone)
             })
             .map(|id| self.get_item(id))
-            .filter(Option::is_some)
-            .map(Option::unwrap)
+            .flatten()
```
litchipi, line 246: I don't understand why remove the `filter + map` checks since an `Option::None` value returned from `self.get_item` is still possible. + Calling `flatten` on a non-nested vector

litchipi (approved):
Rest of it looks OK to me.
Once the CI is OK with building and testing on MacOS I think it can be merged.

src/typecheck/linearization.rs
silverraven691: `Option` is an `Iterator` so the diff ends up doing the exact same thing

silverraven691 (approved):
Nothing stood out, looks fine to me

PR #485, opened 2021-12-06
[User manual] Typing

yannham:
Add some documentation about typing. This is intended as a starting point for the user manual, and as a minimum viable documentation for the first release. Coming documents are probably: contracts, syntax, merging.


doc/manual/types-vs-contracts.md
```diff
@@ -0,0 +1,148 @@
...
+    let foo : Num =
+      let ev : ((Num -> Num) -> Num) -> Num -> Num
+        = fun f x => f (functions.const x) in
+      ev (fun f => f 0) 1
+    in ...
+    ```
+
+## Library (record of functions)
+
+You should use **type annoations** for records of functions. Currently Nickel
```
yannham, line 60:
```suggestion
You should use **type annotations** for records of functions. Currently Nickel
```

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+```
+
+Although `x` should be untyped, it works! Hmm. What about:
+
+```
+let x = 0 + 1 in
+(1 + x : Num)
+```
+
+Now, it doesn't, as we would expect originally. As it is the case here, you can
```
aspiwack, line 472: “it”

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+infer `Dyn -> Dyn` for functions, `{_: Dyn}` for records, and so on. As of now
+(INSERT VERSION HERE), it doesn't.
+
+We could add a type annotation to `x`. But sometimes we don't want to, or we
+can't. Maybe we don't want the typechecker to do its ceremony and recurse in the
+whole expression. Maybe the expression is not typeable in Nickel, but we still
+know it will evaluate to a number. In any of these situations, we can trade the
+type annotation for a contract application:
+
+Exmaple:
```
aspiwack, line 500:
```suggestion
Example:
```

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+**Take-away**: When calling to typed code from untyped code, Nickel
+automatically inserts contract checks at the boundary to enjoy clearer and
+earlier error reporting. In the other direction, an expression `exp | Type` is
+blindly accepted to be of type `Type` by the typechecker. This is a way of using
+untyped values inside typed code by telling the typechecker "trust me on this
+one, and if I'm wrong there will be a contract error anyway". Finally, while a
+type annotation switches the typechecker on, a contract annotation switches it
+back off.
+
+If you read until this point, you may feel confused. We have static types on one
```
aspiwack, line 551: I certainly hope that the reader isn't confused, otherwise the documentation hasn't done its job. At any rate, better not suggest that they do.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+side and we have contracts on the other, but type annotations and contract
+applications actually behave the same at runtime.  Although contracts seem to be
+a runtime thing, contract applications also have an effect on the typechecker.
+What's more, they look like they cover similar use cases (check that some value
+has a specific shape or behavior), and the annotations are written in the same
+type syntax, differing only in the symbols `:` and `|`. How do we know when to
+use one or the other in practice? This is what the guide [Type versus
+contracts: when to?](./types-vs-contracts.md) is for!
```
aspiwack, line 559: This paragraph is not easy to read.

doc/manual/types-vs-contracts.md
```diff
@@ -0,0 +1,148 @@
+# Type vs contracts: when to?
```
aspiwack, line 1: I'm questioning whether this should be a separate document.

I think that it's difficult to introduce all the type-related stuff in the other document without explaining why they matter. Which you do a little, but is always best put in terms of “why/when do you want to use this”.

Intuitively I'd merge the two document with a structure like: here are Nickel's things (contract application, type annotation, etc…) told briefly, the whys of each, then go into the weed of what is supported and how they articulate.

My feeling is that it would flow much more nicely, and be easier to understand.

doc/manual/types-vs-contracts.md
```diff
@@ -0,0 +1,148 @@
...
+
+Most of the time, you should use **type annotations** for functions. For
+reusable code, typing is often more adapted than contracts. You can
+exceptionally use contracts when types are not expressive enough to encode the
+property you want (such as in `#ValidUrl -> #Port -> #ValidUrl`) or if the type
+system is not powerful enough to see that your code is correct.
+
+Type annotations do have a runtime cost. If you hit contracts-related
+performances issues, you can always disable some contract checks using specific
+flags (TO BE PRECISED ONCE WE HAVE THOSE FLAGS). With annotations, code is still
```
aspiwack, line 19: I think that the performance discussion is out of place here.

doc/manual/types-vs-contracts.md
```diff
@@ -0,0 +1,148 @@
...
+    in ...
+
+    let foo : Num =
+      let ev : ((Num -> Num) -> Num) -> Num -> Num
+        = fun f x => f (functions.const x) in
+      ev (fun f => f 0) 1
+    in ...
+    ```
+
+## Library (record of functions)
```
aspiwack, line 58: This is about a gotcha. This should go in a cookbook section somewhere. Maybe said cookbook should be in a different file indeed.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+  │
+  = The type of the expression was expected to be `Str`
+  = The type of the expression was inferred to be `Num`
+  = These types are not compatible
+```
+
+That's too bad, because in practice the code of `filter` is agnostic with
+respect to the type of elements of the list. That is, `filter` is *generic*.
+This can be expressed in Nickel with so-called *parametric polymorphism*. This
+is really just a pedantic name for some flavour of generics. Generic parameters
```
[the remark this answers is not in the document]
yannham: The initial motivation to use "generics" is that it is probably more widely understood (Java, C++, Go, etc.) than "polymorphic". I agree that mixing both doesn't help, though.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+see how typed and untyped code interact.
+
+## Interaction between typed and untyped code
+
+### Using typed code inside untyped code
+
+I invite to you come back to our first love, the `filter` example. In the
+previous section, we included everything -- both the definition of `filter` and
+the call site -- in a typed block. More realistically, `filter` would be a typed
+library function (it is actually part of the stdlib as `lists.filter`), but is
```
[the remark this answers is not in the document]
yannham: Yes, it is. It was copied from Nix. But it is subject to be revised before first release, though, as some other aspects of this document.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+
+We call `filter` from an untyped location, but still get a spot-on error.
+What's is happening? The interpreter protects typed block by a
+[contract](./contracts.md), which is a runtime check. Please refer to the
+dedicated documentation for more details. For now, you can just remember that
+*any type annotation*, wherever it is, gives rise at runtime to a corresponding
+contract check. In other words, `foo: T` and `foo | T` (here `|` is contract
+application, not the row tail separator) behave exactly the same at *runtime*.
+This approach of protecting typed code from untyped code is known as *sound
+gradual typing*.
```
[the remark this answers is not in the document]
yannham: Well, I guess in practice not downcasting from `Dyn` freely means inserting safe casts instead of those downcasts? What I remember from the introduction of [this paper](https://www.cs.cornell.edu/~fabianm/papers/nomalive-oopsla17.pdf) and [this other paper](https://dl.acm.org/doi/10.1145/2837614.2837630) was the idea that in practice, the mean to retain soundness was to insert those runtime checks. But once again, maybe the jargon is not so important for this document, and we can drop it.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+1 + ((if f 10 then 1 else 0) : Num)
+```
+
+Let us try on the filter example. We want the call to be inside the typechecked
+block. The easiest way is to add an annotation at the top-level:
+
+```nickel
+(let filter | (Num -> Bool) -> List Num -> List Num = fun pred l =>
+  lists.foldl (fun acc x => if pred x then acc @ [x] else acc) [] l in
+filter (fun x => if x % 2 == 0 then x else null) [1,2,3,4,5,6]) : List Num
```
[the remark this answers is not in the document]
yannham: That was a typo, I meant to use `:` instead.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+  │
+  = The type of the expression was expected to be `Bool`
+  = The type of the expression was inferred to be `Num`
+  = These types are not compatible
+```
+
+That's already better! The error now points at the call site, and inside our
+anonymous functions, telling us it is expected to return a boolean. What's more,
+we just had to give the top-level annotation `List Num`. Nickel performs type
+inference, so that you don't have to write the type for `filter`, the filtering
```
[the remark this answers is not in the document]
yannham: Copy paste error, the example shouldn't have an annotation for `filter`.

doc/manual/typing.md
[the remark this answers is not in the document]
yannham: In the end I say "filter is generic" in a informal way and then just use polymorphic everywhere. I guess we keep the mention of generics while keeping only one technical word for polymorphism.

doc/manual/typing.md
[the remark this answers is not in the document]
yannham: I removed it, in the end. I don't think it adds useful information at this point.

doc/manual/typing.md
yannham: Got rid of this paragraph.

doc/manual/typing.md
yannham: Got rid of it.

doc/manual/typing.md
```diff
@@ -0,0 +1,559 @@
...
+higherRankId id 0 : Num
+```
+
+**Remark**: if you are a more type-inclined reader out there, you may wonder why
+the typechecker is not capable of inferring a polymorphic type for `filter` by
+itself. Indeed, Hindley-Milner type-inference can precisely infer heading
+`foralls`, such that the first example without the polymorphic annotation would
+already pass. We chose to abandon this so-called automatic generalization,
+because it makes other aspects of the type system and type inference more
+complex or limited. Requiring annotation of polymorphic functions seems like a
```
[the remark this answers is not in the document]
yannham: Good question. What would be a convincing one? On the top of my head, I don't see something obviously incompatible with let-generalization within the current type system. If this is not the original reason, I'm happy to amend.

doc/manual/types-vs-contracts.md
yannham: Moved in a sink `cookbook.md` for now.

doc/manual/types-vs-contracts.md
yannham: Removed

doc/manual/types-vs-contracts.md
yannham: One single document for contracts, typing and how/when to use them would be a bit too big, I think. We could split by "overview" speaking about both typing and contracts, and then have separate detailed sections about typing and contracts, though.

It's possible to inline this one into the respective documents on contracts and typing, or into the potential first overview. The rationale of having this as a separate guide is to have a kind of quick access, "cheat sheet"-like document, indexing sections by the situations, while the e.g. typing section is organized differently. I think it's not a problem if we're repeating things from the other sections.

doc/manual/typing.md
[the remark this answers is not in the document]
aspiwack: Honestly, I don't think that there are current or planned features of the type system which would interact with generalisation. Except maybe (limited forms of) unions and intersections. I don't know for sure. But that's sort of the point, if we don't have generalisation, then this is not a question anymore. We don't have to restrict ourselves to principal sytems with principal types (though they are always good, of course).

There are more subtle arguments against generalisation in [_Let should not be generalised_](https://lirias.kuleuven.be/retrieve/90342) and §4.2 of [_OutsideIn(X)_](https://lirias.kuleuven.be/retrieve/237824) (I think both of them say mostly the same thing).

Both of these argue against generalisation of local lets, but conclude that generalising toplevel bindings is a pragmatic middle ground. Which is still more generalisation than we do.

All that being said, maybe the most honest argument for not having implicit generalisation is that it is algorithmically complex and we didn't want to go that route. It's, of course, something that can be revisited in the future, though our inclination (well, mine really, I guess) is not to.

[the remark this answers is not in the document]
yannham: Should we still say something along the lines "it's just simpler to abandon, and doesn't buy us much"? Or should we just remove any justification attempt? After all this is a user manual, not a language design rationale document.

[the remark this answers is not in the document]
aspiwack: It's fine to keep this sentence or something approaching. It is definitely simpler to not have implicit generalisation, and it indeed doesn't cost that much.

yannham:
I made another pass, which I think addresses all the review comments, excepted the one about structure, which I answered inline.

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+  {
+    incr : Num -> Num = fun x => x + 1,
+    mkPath : Str -> Str -> Str -> Str = fun basepath filename ext =>
+      "#{basepath}/#{filename}.#{ext}",
+  }
+  ```
+
+### Polymorphism
+
+#### Plain polymorphism
```
aspiwack, line 215: Maybe

```suggestion
#### Type polymorphism
```

I'm not sure this is good, but I do fine “plan polymorphism” fairly weird sounding.

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+the record type) is polymorphic, and `a` may be substituted for arbitrary fields
+(such as `jan: Num, feb: Num`). We used two different generic parameters `a` and
+`b`, to express that the tails of the arguments may differ.  If we used `a` in
+both places, as in `forall a. {total: Num | a} -> {total: Num | a} -> Num`, we
+could still write `addTotal {total = 1, foo = 1} {total = 2, foo = 2}` but not
+`addTotal {total = 1, foo = 1} {total = 2, bar = 2}`. Using distinct parameters
+`a` and `b` gives us maximum flexibility.
+
+What comes before the tail may include several fields, is in e.g. `forall a.
+{total: Num, subtotal: Num | a} -> Num`.
```
aspiwack, line 375: There lacks an example where the row variable is actually used.

Something like
```
f : forall r. { a : Num, b : Num | r } -> {  a : Num, b : Num, sum-a-b : Num | r }
f x = x & { sum-a-b  = x.a + x.b }
```

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+
+### Take-away
+
+The type system of Nickel has a few basic types (`Dyn`, `Num`, `Str`,
+and `Bool`) and type constructors for lists, records, enums and functions.
+Nickel features generics via polymorphism, introduced by the `forall` keyword. A
+type can not only be generic in other types, but records and enums types can
+also be generic in their tail. The tail is delimited by `|`.
+
+## Interaction between statically typed and dynamically typed code
```
aspiwack, line 401: It feels like this should be before the section with the detailed type system.

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+situation, the call site escapes the typechecker. Thus, without an additional
+mechanism, static typing would only ensure that the implementation of `filter`
+doesn't violate the typing rules, but wouldn't prevent a ill-formed call from
+dynamically typed code.  At first sight, we haven't solved at all the main issue
+of delayed dynamic type errors that static typing was supposed to solve in the
+first place! Remember, the typical problem is the caller passing an value of the
+wrong type that eventually raises an error from within `filter`.
+
+Hopefully, Nickel does have a mechanism to prevent this from happening and to
+provide good error reporting in this situation.  Let us see by ourselves:
```
aspiwack, line 422:
```suggestion
Fortunately, Nickel does have a mechanism to prevent this from happening and to
provide good error reporting in this situation.  Let us see by ourselves:
```

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+### Using statically typed code inside dynamically code
+
+Until now, we have written the statically typed `filter` examples using
+statically typed blocks that enclosed both the definition of `filter` and the
+call sites. More realistically, `filter` would be a statically typed library
+function (it is actually part of the standard library as `lists.filter`) while
+being likely to be called from dynamically typed configuration files. In this
+situation, the call site escapes the typechecker. Thus, without an additional
+mechanism, static typing would only ensure that the implementation of `filter`
+doesn't violate the typing rules, but wouldn't prevent a ill-formed call from
```
aspiwack, line 415:
```suggestion
doesn't violate the typing rules, but wouldn't prevent an ill-formed call from
```

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+
+We call `filter` from a dynamically typed location, but still get a spot-on
+error. To precisely avoid dynamically code injecting values of the wrong type
+inside statically typed blocks via function calls, the interpreter protects said
+blocks by a contract. Contracts form a principled runtime verification scheme.
+Please refer to the [dedicated document](./contracts.md) for more details, but
+for now, you can just remember that *any type annotation* (wherever it is) gives
+rise at runtime to a corresponding contract application. In other words, `foo:
+T` and `foo | T` (here `|` is contract application, not the row tail separator)
+behave exactly the same at *runtime*.
```
aspiwack, line 470: What does “contract application” mean?

In fact, since you are going to explain it in the next subsection, I would simply not mention this distinction at this point.

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+
+In the other direction, we face a different issue. How can we use untyped values
+inside typed code? Let us try:
+
+```
+let x = 1 in
+(1 + x : Num)
+```
+
+Although `x` shouldn't be statically typed, this code is accepted. Hmm. What
```
aspiwack, line 485: I think that this example should be lower down. It's a facility that Nickel offers, but not an essential feature. And it makes the discussion more confused at this point.

doc/manual/typing.md
```diff
@@ -0,0 +1,629 @@
...
+contract error. Thus, if we reach `1 + x`, at this point `x` is necessarily a
+number and won't cause any type mismatch. In a way, the contract application
+acts like a type cast, but whose verification is delayed to run-time.
+
+Dually to a static type annotation, a contract annotation *turns the typechecker
+off again*. This is illustrated by the following variation being accepted:
+
+```nickel
+(1 + ((if true then 0 else "a" | Num)) : Num
+```
```
aspiwack, line 539: I would start with something like this, to be honest. Then expand to `let x | Num = …` then expand to obvious literals.

doc/manual/cookbook.md
```diff
@@ -0,0 +1,49 @@
+# Cookbook
+
+For now, this is an unorganized, temporary document not to forgot some pieces of
```
aspiwack, line 3:
```suggestion
For now, this is an unorganized, temporary document not to forget some pieces of
```

mboes:
What are the blockers to merging this? There is currently no user guide in the repo. The only documentation available is the "getting started" page on the website. We do need documentation before any release.

yannham:
It needs another pass to take into account @aspiwack second review, but then I would say we are good to merge as a first version.

doc/manual/typing.md
yannham: It's a good example, but sadly merging isn't typable currently, so this won't typecheck. For now I'll use the extension operator. We'll revisit the manual anyway just before release, if only because syntax is likely to have changed in between.

doc/manual/typing.md
yannham: I think there will be an overarching section introducing both this manual section and the one on contracts, which will sketch the fundamental principles and terms. Otherwise there's a bit of a bootstraping problems, as it's not easy to detail one without having introduced anything about the other.

yannham:
I think it's good to go for a first base. As said before nothing is set in stone, and all the manual will need a pass anyway before the first release.

Issue #462, opened 2021-11-12
Pre-release standardization: tracking issue


yannham:
This issue is meant to track a bunch of important practical usability aspects, that have been mostly discussed in issues, meetings or design documents, sometimes extensively, but without reaching a clear conclusion. The goal is to go over them and decide once and for all for the first release (or, at least, to consciously and willfully declare them out of scope for the first release, and to be debated later).

## Issues

- [x] Should we drop the `fun` keyword for something else, like just the arrow `foo bar => functionBody`, with parentheses `(foo bar) => body` or use a character as in Haskell `\foo bar => body`? Also, should we have an ML-like function syntax avoiding the anonymous function syntax most of the time: `let f x y = x + y in` ?
      Related: #207, in particular https://github.com/tweag/nickel/issues/207#issuecomment-900271443)
- [x] Should we allow a record to reference fields possibly coming from "sibling" from future merges once #330 lands, as in `{foo = 1} & {bar = foo}`? It introduces a form of dynamic binding. Alternatives were to force the second record to specify it expects a `foo` field to be defined in some way or another, as `{ bar = foo, foo | Dyn}`.
      Related: #83, #330, #459
- [x] Should we, and if yes how, differentiate custom contracts from types, both at definition and at usage?
      Related: #295, #318
- [x] Should we have a type hole syntax, to let the typechecker fill the gap, as in `let foo : _ = ... in ...` ?
      Related: #226, #430
- [ ] Should we change the file extension?
      Related: #357
- [x] Should we have a block let-binding syntax?
     Related: #218
- [ ] Standard library stabilization
     Related: #321 

**ADDENDUM**

- [x] Should we have top-level let-bindings?
      Related: #95
- [x] Case convention for values/types/contracts/stdlib

yannham:
The first release has passed, and most of those points have been decided. Open issues remain for the others.

PR #486, opened 2021-12-07
Add notes of the standardization meeting #1

yannham:
Related: #462.

### DIFF
--- a/notes/standardization-meeting-07.12.21.md
+++ b/notes/standardization-meeting-07.12.21.md
@@ -3,7 +3,7 @@ object: 1st standardization meeting
 attending:
   - Arnaud Spiwack
   - Eelco Dolstra
-  - Thépohane Huschmitt
+  - Théophane Hufschmitt
   - Yann Hamdaoui
 author: Yann Hamdaoui (yann.hamdaoui@tweag.io)
 date: "Dec 7, 2021"
@@ -14,7 +14,7 @@ date: "Dec 7, 2021"
 Attending:
 - Arnaud Spiwack
 - Eelco Dolstra
-- Théophane Huschmitt
+- Théophane Hufschmitt
 - ~~Rok Garbas~~
 - Yann Hamdaoui
 
--- a/notes/standardization-meeting-07.12.21.md
+++ b/notes/standardization-meeting-07.12.21.md
@@ -99,7 +99,7 @@ Currently, one has to write explicitly the type signature to turn typechecking o
 
 **Proposal**
 `let foo : _ = ... in ...`. Each type hole would be substituted for a
-unification variable at and resolvesd at typechecking. Unconstrained variables
+unification variable and resolved during typechecking. Unconstrained variables
 are substituted for `Dyn`.
 
 **Usages**
--- a/notes/standardization-meeting-07.12.21.md
+++ b/notes/standardization-meeting-07.12.21.md
@@ -241,6 +241,7 @@ open_contract(#(open_contract(T))) =  open_contract(T)
   Example:
   ```nickel
   //Naive intersection via merging. Would be nice. Work for records but not for
+  //contracts as functions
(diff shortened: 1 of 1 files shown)


notes/standardization-meeting-07.12.21.md
```diff
@@ -0,0 +1,283 @@
...
+  foo | #(GreaterThan 2) = ...
+  foo | #(Nullable (AllOf [Port, Unreserverd])) = ...
+  foo | #(Nullable {fst = )) = ...
+  ```
+- Contracts are bare values. It means we can't recognize them and apply special
+  behavior.
+
+  Example:
+  ```nickel
+  //Naive intersection via merging. Would be nice. Work for records but not for
```
silverraven691, line 243:
```suggestion
  //Naive intersection via merging. Would be nice. Work for records but not for
  //function contracts
```

notes/standardization-meeting-07.12.21.md
```diff
@@ -0,0 +1,284 @@
...
+But it introduces dynamic scoping, which is bad.
+
+**Proposal**
+- add it nevertheless
+- force the definition of an interface:
+  ```nickel
+  {foo = bar + 1, bar | Num} & {bar}
+  // For more complex/dynamic interfaces
+  let Interface = {bar | Num} in
+  {foo = bar + 1, ..Interface}
```
silverraven691, line 175: Does this have to be different to something like the following because scope would be calculated before the contract would be verified, and would therefore rely on dynamic scoping?
```
  let Interface = {bar | Num} in
  {foo = bar + 1} | #Interface
```

I'm worried this may feel a little redundant in a case like this:
```
  let Interface = {bar | Num} in
  let impl | #Interface = {foo = bar + 1, ..Interface},
  ..
```

yannham: >Does this have to be different to something like the following because scope would be calculated before the contract would be verified, and would therefore rely on dynamic scoping?

At first sight, I would say so. Although we could also have rule like if `Interface` is statically known to be a record with field `bar`, then `impl | #Interface` would be sufficient. We would probably want a similar constraint to allow the user to write `..Interface` anyway, so that wouldn't be very different. In the end, I think the idea to avoid the pitfalls of dynamic scoping is that `Interface` has enough static information that we can know where `bar` comes from and that it isn't any random unbound identifier.

yannham: The problem with this kind of rules though, is that they are often fragile, as they are a bit ad hoc. Any indirection or dynamic contract would break the code that was working in simple cases.

silverraven691: Yep, makes sense, thank you.

notes/standardization-meeting-07.12.21.md
```diff
@@ -0,0 +1,284 @@
...
+    f x y = ...,
+    g : Bar -> Stuff,
+    g z x = ...,
+in
+```
+
+### Type hole syntax
+[#226](https://github.com/tweag/nickel/issues/226)
+
+Currently, one has to write explicitly type signature to turn typechecking on:
```
yannham, line 97:
```suggestion
Currently, one has to write explicitly the type signature to turn typechecking on:
```

notes/standardization-meeting-07.12.21.md
```diff
@@ -0,0 +1,284 @@
...
+`let foo : SomeType = ... in ...`
+
+**Proposal**
+`let foo : _ = ... in ...`. Each type hole would be substituted for a
+unification variable at and resolvesd at typechecking. Unconstrained variables
```
yannham, line 102:
```suggestion
unification variable and resolved during typechecking. Unconstrained variables
```

yannham:
As those are notes and not an RFC or a more formal document, I'm going ahead and merging as it is.

PR #490, opened 2021-12-07
Fix #487

### LINKED ISSUE #487: The REPL doesn't show properly parse error anymore
**Describe the bug**
If we enter something that fails to parse inside the REPL, we don't get any meaningful error reporting, but rather a `<parse error>`. If we try to add contracts or type annotations, it's even worse as we just get random contract violations

**To Reproduce**
```
nickel>
let protocol : <http, ftp, sftp> = `http in
  switch {
    http -> 1,
    ftp -> 2,
    sftp -> 3
  } protocol

<parse error>

nickel>(let protocol : <http, ftp, sftp> = `http in
  switch {
    http -> 1,
    ftp -> 2,
    sftp -> 3
  } protocol) : Num

error: Blame error: contract broken by a value.
  ┌─ :1:1
  │
1 │ Num
  │ --- expected type
  │
  ┌─ repl-input-22:2:3
  │
2 │ ╭   switch {
3 │ │     http -> 1,
4 │ │     ftp -> 2,
5 │ │     sftp -> 3
6 │ │   } protocol) : Num
  │ ╰────────────' evaluated to this expression

note:
  ┌─ repl-input-22:6:17
  │
6 │   } protocol) : Num
  │                 ^^^ bound here

```

**Expected behavior**
I expect parse errors to be reported nicely via an error diagnostic, as they used to be.

**Additional context**
It is most probably caused by PR #456. For whatever reason the REPL now considers parse errors as acceptable, while it shouldn't.

cc @Avi-D-coder

Avi-D-coder:
Also adds wrapper functions around parse variants so it's easier to use.


src/error.rs
```diff
@@ -213,6 +213,20 @@ impl ParseErrors {
     pub const fn none() -> ParseErrors {
         ParseErrors { errors: Vec::new() }
     }
+
+    pub fn from_recoverable<'a>(
+        errs: Vec<
+            lalrpop_util::ErrorRecovery<usize, parser::lexer::Token<'a>, parser::error::ParseError>,
```
yannham, line 219: Nitpick: just for the sake of readability of the signature, it may look nicer to put the ugly type in a type alias.

src/parser/mod.rs
```diff
@@ -18,3 +20,65 @@ pub enum ExtendedTerm {
...
+            .map_err(|err| ParseError::from_lalrpop(err, file_id));
+
+        let parse_errors = ParseErrors::from_recoverable(parse_errors, file_id);
+        match result {
+            Ok(t) => Ok((t, parse_errors)),
+            Err(e) => Err(e),
+        }
+    }
+
+    pub fn parse_term(
```
yannham, line 42: Nitpick of nitpicks: could be named just `parse` and `parse_term`, as the name of the structure `TermParser` is already quite explicit.

yannham (approved):
LGTM. Left a few nitpicks. Feel free to ignore them or not.

src/parser/mod.rs
Avi-D-coder: Unfortunately, both structs already have a lalrpop generated method `parse`.
I'm also not happy with the naming scheme, but couldn't think of anything else?

src/error.rs
Avi-D-coder: I imported `ErrorRecovery`, `Token` to make the signature more dense.

src/parser/mod.rs
yannham: I see. Then we can either leave it as it is, or wrap LALRPOP "raw" parsers in new structs, so that we can hide the default `parse` method and only exposes the current `parse_term` and `parse_term_tolerant`. But then, you have to find a good name for the wrapping structure :stuck_out_tongue:

PR #491, opened 2021-12-08
Recursive overriding

yannham:
Implement the recursive overriding of records, as detailed in #330 and along the lines of #103. This doesn't add merge priorities, merge functions, etc., but just implement the call-by-name semantics when merging recursive records:

```
({foo | default = 0, bar = foo + 1} & {foo = 2}).bar // Equals 3 after this PR
```

## Overview

This PR implements the following changes:
1. Add a `LetRev` node to the AST. `LetRev` acts as a normal `Let`, excepted that it allocates a revertible thunk when evaluated. The share normal form pass generates those `LetRev` instead of normal `Let` precisely for the fields of recursive records.

    To be honest, it's becoming clear at this point that we need to have at least two different ASTs, one straight out of the parser, and one for runtime, as those "syntax-only" and "runtime-only" constructs start to accumulate in the current single AST. However, it is out of scope for this issue. Also, `LetRev` is not so bad as for anything else but evaluation, it is strictly equivalent to a `Let`, so most changes are mechanical.
2. When merging two records, revert all the thunks inside the fields of the operands before performing the usual recursive merging. Once done, `merge` returns a `RecRecord` instead of a standard `Record`, such that when evaluated, the abstract machine builds a fresh recursive environment again (tying the knot).
3. Incidentally, I had to implement a simple optimization for `colsurize` that has been on the plate for a long time. It avoids dumbly generating nested closures when unnecessary. It is required for overriding as the indirections introduced by those nested closures mess with the creation of the recursive environment of `RecRecord`s generated by `merge`.

### DIFF
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -322,7 +322,7 @@ pub fn merge(
         }
         // Merge put together the fields of records, and recursively merge
         // fields that are present in both terms
-        (Term::Record(m1, attrs1), Term::Record(m2, attrs2)) => {
+        (Term::Record(mut m1, attrs1), Term::Record(mut m2, attrs2)) => {
             /* Terms inside m1 and m2 may capture variables of resp. env1 and env2.  Morally, we
              * need to store closures, or a merge of closures, inside the resulting record.  We use
              * the same trick as in the evaluation of the operator DynExtend, and replace each such
@@ -330,6 +330,8 @@ pub fn merge(
              */
             let mut m = HashMap::new();
             let mut env = Environment::new();
+            rev_thunks(m1.values_mut());
+            rev_thunks(m2.values_mut());
             let (mut left, mut center, mut right) = hashmap::split(m1, m2);
 
             match mode {
@@ -430,6 +432,14 @@ fn merge_closurize(
         t2.closurize(&mut local_env, env2),
     ));
     body.closurize(env, local_env)
+}
+
+fn rev_thunks<'a, I: Iterator<Item = &'a mut RichTerm>>(map: I) {
+    for rt in map {
+        if let Term::Var(id) = std::mem::take(&mut *rt.term) {
+            *rt.term = Term::VarRev(id);
+        }
+    }
 }
 
 pub mod hashmap {
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -52,8 +52,7 @@
 //! - *Contract check*: merging a `Contract` or a `ContractDefault` with a simple value `t`
 //! evaluates to a contract check, that is an `Assume(..., t)`
 use crate::error::EvalError;
-use crate::eval::CallStack;
-use crate::eval::{Closure, Environment};
+use crate::eval::{CallStack, Closure, Environment, IdentKind, Thunk};
 use crate::label::Label;
 use crate::position::TermPos;
 use crate::term::{make as mk_term, BinaryOp, Contract, MetaValue, RecordAttrs, RichTerm, Term};
@@ -332,7 +331,7 @@ pub fn merge(
             let mut env = Environment::new();
             rev_thunks(m1.values_mut(), &mut env1);
             rev_thunks(m2.values_mut(), &mut env2);
-            let (mut left, mut center, mut right) = hashmap::split(m1, m2);
+            let (left, center, right) = hashmap::split(m1, m2);
 
             match mode {
                 MergeMode::Contract(mut lbl) if !attrs2.open && !left.is_empty() => {
@@ -345,15 +344,15 @@ pub fn merge(
                 _ => (),
             };
 
-            for (field, t) in left.drain() {
+            for (field, t) in left.into_iter() {
                 m.insert(field, t.closurize(&mut env, env1.clone()));
             }
 
-            for (field, t) in right.drain() {
+            for (field, t) in right.into_iter() {
                 m.insert(field, t.closurize(&mut env, env2.clone()));
             }
 
-            for (field, (t1, t2)) in center.drain() {
+            for (field, (t1, t2)) in center.into_iter() {
                 m.insert(
                     field,
                     merge_closurize(&mut env, t1, env1.clone(), t2, env2.clone()),
@@ -362,7 +361,7 @@ pub fn merge(
 
             Ok(Closure {
                 body: RichTerm::new(
-                    Term::Record(m, RecordAttrs::merge(attrs1, attrs2)),
+                    Term::RecRecord(m, Vec::new(), RecordAttrs::merge(attrs1, attrs2)),
                     pos_op.into_inherited(),
                 ),
                 env,
@@ -438,14 +437,15 @@ fn rev_thunks<'a, I: Iterator<Item = &'a mut RichTerm>>(map: I, env: &mut Enviro
     use crate::transformations::fresh_var;
 
     for rt in map {
-        if let Term::Var(id) = std::mem::take(&mut *rt.term) {
+        if let Term::Var(id) = rt.as_ref() {
             // This create a fresh variable which is bound to a reversed copy of the original thunk
             let reversed = env.get(&id).unwrap().restore();
             let fresh_id = fresh_var();
             env.insert(fresh_id.clone(), reversed);
             *rt.term = Term::Var(fresh_id);
         }
-        // Otherwise, after the share normal form transformations, it should be a constant
+        // Otherwise, if it is not a variable after the share normal form transformations, it
+        // should be a constant and we don't need to reverse anything
     }
 }
 
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -52,7 +52,7 @@
 //! - *Contract check*: merging a `Contract` or a `ContractDefault` with a simple value `t`
 //! evaluates to a contract check, that is an `Assume(..., t)`
 use crate::error::EvalError;
-use crate::eval::{CallStack, Closure, Environment, IdentKind, Thunk};
+use crate::eval::{CallStack, Closure, Environment};
 use crate::label::Label;
 use crate::position::TermPos;
 use crate::term::{make as mk_term, BinaryOp, Contract, MetaValue, RecordAttrs, RichTerm, Term};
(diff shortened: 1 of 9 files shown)


src/transformations.rs
```diff
@@ -368,14 +397,32 @@ impl Closurizable for RichTerm {
     /// Generate a fresh variable, bind it to the corresponding closure `(t,with_env)` in `env`,
     /// and return this variable as a fresh term.
     fn closurize(self, env: &mut Environment, with_env: Environment) -> RichTerm {
+        // If the term is already a generated variable (that is, introduced by the share normal
+        // form transformation), we don't have to create a useless intermediate closure. We just
+        // transfer the original thunk to the new environment. This is not only an optimization:
+        // this is relied upon by recursive record merging when computing the fixpoint. Change
+        // carefully.
```
aspiwack, line 404: “Change carefully” is a bit terse :slightly_smiling_face: . What are the invariants/properties that recursive record merging depends on?

src/transformations.rs
```diff
@@ -368,14 +397,32 @@ impl Closurizable for RichTerm {
...
-        };
-        env.insert(var.clone(), Thunk::new(closure, IdentKind::Record()));
 
+        let thunk = reuse_thunk.unwrap_or_else(|| {
+            let closure = Closure {
+                body: self,
+                env: with_env,
+            };
+            Thunk::new(closure, IdentKind::Record())
+        });
```
aspiwack, line 423: I would simply move all of this in the `match` above, and avoid indirecting through `Option`.

src/term.rs
```diff
@@ -146,6 +149,12 @@ pub enum Term {
     ParseError,
 }
 
+impl Default for Term {
+    fn default() -> Self {
+        Term::Null
+    }
+}
```
aspiwack, line 156: Why do we need this? It looks pretty dubious.

src/transformations.rs
```diff
@@ -169,18 +176,38 @@ pub mod share_normal_form {
...
+    enum BindingType {
+        Normal,
+        Revertible,
+    }
+
+    impl Default for BindingType {
+        fn default() -> Self {
+            BindingType::Normal
+        }
+    }
```
aspiwack, line 189: More `Default`. Does it really make sense?

src/transformations.rs
```diff
@@ -169,18 +176,38 @@ pub mod share_normal_form {
...
+    ) -> RichTerm {
         bindings.into_iter().fold(
             RichTerm {
                 term: Box::new(body),
                 pos: pos.into_inherited(),
             },
-            |acc, (id, t)| RichTerm::new(Term::Let(id, t, acc), pos),
+            |acc, (id, t)| match btype {
+                BindingType::Normal => RichTerm::new(Term::Let(id, t, acc), pos),
+                BindingType::Revertible => RichTerm::new(Term::LetRev(id, t, acc), pos),
```
aspiwack, line 209: This is a bit meh. Does it indicate that `Let` and `LetRev` ought to be the same `Let` with a `BindingType` argument?

src/typecheck/mod.rs
```diff
@@ -622,7 +622,7 @@ fn type_check_<S, E>(
             unify(state, strict, ty, mk_typewrapper::dynamic())
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))
         }
-        Term::Let(x, re, rt) => {
+        Term::Let(x, re, rt) | Term::LetRev(x, re, rt) => {
```
aspiwack, line 625: This appears to be a second indication that there should be a single `Let`.

src/merge.rs
```diff
@@ -322,15 +321,17 @@ pub fn merge(
...
             /* Terms inside m1 and m2 may capture variables of resp. env1 and env2.  Morally, we
              * need to store closures, or a merge of closures, inside the resulting record.  We use
              * the same trick as in the evaluation of the operator DynExtend, and replace each such
              * term by a variable bound to an appropriate closure in the environment
              */
             let mut m = HashMap::new();
             let mut env = Environment::new();
-            let (mut left, mut center, mut right) = hashmap::split(m1, m2);
+            rev_thunks(m1.values_mut(), &mut env1);
+            rev_thunks(m2.values_mut(), &mut env2);
```
aspiwack, line 333: This looks a bit inefficient. There will probably have to be a deep thought on data representation in the abstract machine in the future. But after the release.

yannham: Do you have a particular aspect in mind? The fact that we are not tracking dependencies of fields and just revert everything blindly, or something else?

src/transformations.rs
```diff
@@ -368,14 +397,32 @@ impl Closurizable for RichTerm {
     /// Generate a fresh variable, bind it to the corresponding closure `(t,with_env)` in `env`,
     /// and return this variable as a fresh term.
     fn closurize(self, env: &mut Environment, with_env: Environment) -> RichTerm {
+        // If the term is already a generated variable (that is, introduced by the share normal
+        // form transformation), we don't have to create a useless intermediate closure. We just
```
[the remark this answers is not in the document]
yannham: True, this is mentioned just below in the comment, but maybe not very clear from the beginning. Will amend

src/transformations.rs
yannham: Fair enough, will make more explicit.

src/term.rs
yannham: In Rust, `Default` is useful to swap values in place without violating borrowing rules. The typical use-case is to take some value out replacing it by a default value, doing what you have do, and put it back in place:

```
fn map_sth(t: Term) -> Term { ... }

fn map_in_place(t_ref: &mut Term) {
  let t = std::mem::take(t_ref); // Take content and replace it by Default::default()
  *t_ref = map_sth(t);
}
```

This is one reason it is quite liberally implemented in Rust. Also I don't find `null` to be unreasonable, when you need a dummy term for testing or something. That being said, this one is most probably an artifact for a previous use with `mem::take`, as removing it doesn't fail compilation. I can remove it for now.

src/merge.rs
aspiwack: Indeed, on principle, we don't want to be traversing an environment that we don't know the size of.

That being said, it's probably ok for now. We will just have to make a pass on the execution model with a cooler head. In the future.

src/term.rs
aspiwack: I see.

The documentation is pretty tricky

> A trait for giving a type a useful default value.

I would interpret “useful” in this sentence to mean “meaningful”. But the next sentence is

> Sometimes, you want to fall back to some kind of default value, and don’t particularly care what it is.

Which contradicts my understanding of the first sentence.

Anyway, with the disclaimer that `default()` is really a placeholder value that you never want to inspect, I have no qualm. (though, of course, if it's dead code, maybe it's best to kill it).

src/merge.rs
yannham: I'm not sure to follow the "an environment that we don't know the size of". This operation is not dependent on the size of the environment: the environment is just passed to retrieve the thunks and replace them with the reversed thunks (so one `remove` and one `insert` per field). For custom merge function reversing may be more involved, but for recursive records  it is `O(|m1| + |m2|)`, which was already the case for merging without this PR.

What we could do is to only reverse those fields that actually refers to other fields in a recursive manner. I'm not sure it would make a big difference for the operation of merging in itself, but for the evaluation of the program it would avoid invalidating and potentially recomputing thunks that don't really need it.

aspiwack: Ah I think I misunderstood. I read `rev_thunks` as reverting all the thunks in the environment. Which it isn't.

As long as the operations are only dependent on the number of fields, I think that it's fair game.

src/transformations.rs
yannham: Initially it's because in case the thunk is not found in the environment, I wanted to fallback on the same code (418 - 422). If we were to put it in the match, we would have to duplicate it (`with_env.get` returns an `Option`). But now that I think of it, not finding a generated identifier in the environment is something we want to hard fail on, and the fallback does violate the invariant relied upon by the evaluation of recursive records mentioned in the comment above. So let's rather panic! or bubble an `Err`

src/transformations.rs
```diff
@@ -368,14 +380,63 @@ impl Closurizable for RichTerm {
...
+        // {foo = %2, bar = 1} in env %2 <- (%1 in env %1 <- 1 + bar), bar <- 1
+        // ```
+        //
+        // Then, evaluating `foo` would unduly raise an unbound identifier error.
+        //
+        //
+        // We currently only do this optimization for generated variables (introduced by the share
+        // normal form). We could do it for non-generated identifiers as well, but be we would be
+        // renaming user-supplied variables by gibberish generated names. It may hamper error
+        // reporting, so for the time being, we restrict ourselves to generated identifiers.
```
aspiwack, line 421: And it will never break the recursive record merging to not handle user-defined variables this way?

yannham: TL;DR Theoretically it should be fine for closurize to handle user-defined variables either way, but you made me realize the share normal form may be breaking the invariant in some cases, which might break things if we were to optimize user-defined variables as well at some point.

Very good question. After thinking a bit more about it, I think it is actually working whether we do the optimization for user-supplied variable or not, but somehow by chance, while it should work by design.

The invariant merging is relying on is that every record field is a generated variable that itself contains the original expression (excepted for constants, that we don't really care about). Because this is a generated variable, the fact that we do or don't do the optimization for non-generated variable shouldn't matter.

But, you just made me realize this invariant may not exactly be always true. Because the share normal form transformation also tries not to introduce useless indirection, it doesn't do anything for fields that are already containing just a variable, such as `{foo = bar, bar = 1}`. There, the share normal form transformation would leave the record as it is. This violates the invariant, and the following example should intuitively not work as intended:

```
({ foo | default = 1, bar = foo} & {}).foo
```

However, it does work. I suspect it is precisely *because we do add an indirection* in `closurize` for non-generated variables currently, causing `closurize` to somehow make up for the behavior of the share normal form transformation. But by accident, and if we were to extend of the optimization to all variables, that might break the example above.

I'll try out a few things to validate those hypotheses, but if it's the case, I imagine the simplest is to remove this optimization from the share normal form so that it does enforce the sought invariant at the cost of adding a superfluous indirection for fields that contains a variable.

yannham: False alert, share normal form was already doing the right thing, actually, always introducing an indirection. I'll add a few corresponding tests, just in case.

aspiwack: So, maybe you need to adjust your comment a little before merging.

aspiwack (approved):

PR #497, opened 2021-12-09
Allow dash in identifiers

### LINKED ISSUE #492: Allow dashes in identifiers
 Dashes are common in option names. Currently one has to escape record fields with dashes inside because it is interpreted as a minus:

```nickel
{
  "foo-bar"."bar-baz" = null,
}
```

As decided during the [standardization meeting](https://github.com/tweag/nickel/blob/master/notes/standardization-meeting-07.12.21.md), alow the dash character "-" as part of an identifier, as in Nix.

```nickel
{
  foo-bar.bar-baz = null,
}
```

This means that `a-b` is not parsed as subtraction anymore. Subtraction now requires spaces. But we expect this to be a minor inconvenience (or even good practice) in a configuration language, versus the field escaping annoyance.

yannham:
Close #492.

### DIFF
--- a/examples/fibonacci/fibonacci.ncl
+++ b/examples/fibonacci/fibonacci.ncl
@@ -10,6 +10,6 @@ let fibonacci = {
   else if n == 1 then
     1
   else 
-    f (n-1) + f (n-2)
+    f (n - 1) + f (n - 2)
 }.f in
 fibonacci 10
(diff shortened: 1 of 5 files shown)


regnat (approved):
LGTM

(you could add a test with a dash in a toplevel identifier for completeness, but it’s probably not utterly important)

PR #498, opened 2021-12-09
Fix/stdlib typechecking

### LINKED ISSUE #375: Standard library functions can't be used in typed code
This problem has been described in more general terms in #226, but #226 is rather an open-ended reflection than a focused issue. This issue is dedicated to the specific problem of using stdlib functions within typed code.

**Describe the bug**
The stdlib modules hide the types of stdlib functions, making them unusable within typed code.

**To Reproduce**

```
nickel> (lists.length [] == 0) : Bool
error: Incompatible types
  ┌─ repl-input-1:1:2
  │
1 │ (lists.length [] == 0) : Bool
  │  ^^^^^ this expression
  │
  = The type of the expression was expected to be `{length: _a -> _b | _c}`
```

**Expected behavior**
The type of stdlib functions is taken into account:

```
nickel> (lists.length [] == 0) : Bool
true
```

francois-caddet:
fix typing with stdlib.
specific type environment construction for the stdlib. the records are typed on top level traversing recursively the records to infer type of fields. From each typed fields, construct a type record to type the top level record.
close #375


francois-caddet:
Replaced `cache.prepare_stdlib()` in some places by `cache.load_stdlib()` so the stdlib is not transformed and the eval environment not constructed. It's all good in nickel executable but not sure if it's the good way to do for the `lsp/nls/src/server.rs`

src/cache.rs
```diff
@@ -770,8 +791,13 @@ impl Cache {
 
     /// Load, parse, and apply program transformations to the standard library. Do not typecheck
     /// for performance reason: this is done in the test suite.
-    pub fn prepare_stdlib(&mut self) -> Result<(), Error> {
+    /// Return a global environment containing both eval and type environment. If you need only the
+    /// type environment, prefer to use `load_stdlib()` then `mk_global_type` to avoid
```
yannham, line 795:
```suggestion
    /// type environment, use `load_stdlib()` then `mk_global_type` to avoid
```

src/cache.rs
```diff
@@ -782,18 +808,37 @@ impl Cache {
...
     }
 
-    /// Generate a global environment from the list of `file_ids` corresponding to the standard
+    /// Generate a global typing environment from the list of `file_ids` corresponding to the standard
     /// library parts.
-    pub fn mk_global_env(&self) -> Result<eval::Environment, CacheError<Void>> {
-        if let Some(ids) = self.stdlib_ids.as_ref().cloned() {
-            let mut env = eval::Environment::new();
+    pub fn mk_global_types(&self) -> Result<typecheck::Environment, CacheError<Void>> {
+        let type_env = self
```
yannham, line 818: It's call `type_env` but it looks like it's a vector containing the stdlib entries, isn't it?

src/cache.rs
```diff
@@ -782,18 +808,37 @@ impl Cache {
...
+        Ok(GlobalEnv { eval_env, type_env })
     }
 
-    /// Generate a global environment from the list of `file_ids` corresponding to the standard
+    /// Generate a global typing environment from the list of `file_ids` corresponding to the standard
     /// library parts.
-    pub fn mk_global_env(&self) -> Result<eval::Environment, CacheError<Void>> {
-        if let Some(ids) = self.stdlib_ids.as_ref().cloned() {
-            let mut env = eval::Environment::new();
+    pub fn mk_global_types(&self) -> Result<typecheck::Environment, CacheError<Void>> {
```
yannham, line 817: As a nitpick, I would suggest to add `env` in the name of those functions, because I find we quite lose the meaning going from `mk_global_env` to `mk_global_eval` and `mk_global_types` (We don't really make an "eval"). Something like `mk_global_eval_env` and `mk_global_type_env` (or even dropping the `global`, if you find it too verbose, as it is a less useful information I think) may be more suggestive.

src/program.rs
```diff
@@ -95,19 +91,15 @@ impl Program {
...
+        let global_env = self.cache.prepare_stdlib()?;
         query(&mut self.cache, self.main_id, &global_env, path)
     }
 
     /// Load, parse, and typecheck the program and the standard library, if not already done.
     pub fn typecheck(&mut self) -> Result<(), Error> {
         self.cache.parse(self.main_id)?;
         self.cache.load_stdlib()?;
-        let global_env = self.cache.mk_global_env().expect("program::typecheck(): stdlib has been loaded but was not found in cache on mk_global_env()");
+        let global_env = self.cache.mk_global_types().unwrap();
```
yannham, line 102: Why removing the panic error message?

src/repl/mod.rs
```diff
@@ -70,16 +70,12 @@ pub struct REPLImpl {
...
     /// declarations and loadings made inside the REPL.
-    eval_env: eval::Environment,
-    /// The initial eval environment, without the toplevel declarations made inside the REPL. Used
+    env: GlobalEnv,
+    /// The initial type environment, without the toplevel declarations made inside the REPL. Used
     /// to typecheck imports in a fresh environment.
-    init_eval_env: eval::Environment,
-    /// The typing environment, counterpart of the eval environment for typechecking. Entries are
-    /// [`TypeWrapper`](../typecheck/enum.TypeWrapper.html) for the ease of interacting with the
-    /// typechecker, but there are not any unification variable in it.
```
yannham, line 81: I think you deleted the documentation by error?

src/repl/mod.rs
```diff
@@ -70,16 +70,12 @@ pub struct REPLImpl {
     cache: Cache,
     /// The parser, supporting toplevel let declaration.
     parser: grammar::ExtendedTermParser,
-    /// The eval environment. Contain the global environment with the stdlib, plus toplevel
+    /// The global environment. Contain the global environment with the stdlib, plus toplevel
     /// declarations and loadings made inside the REPL.
-    eval_env: eval::Environment,
-    /// The initial eval environment, without the toplevel declarations made inside the REPL. Used
+    env: GlobalEnv,
```
yannham, line 75: I'm not sure we should `GlobalEnv` here. `GlobalEnv` is used to return a pair of environment from the cache, but here I feel it's just adding an indirection that is not very useful. We could inline `eval_env` and `type_env` as before. But it's just a nitpicking suggestion, feel free to ignore.

src/typecheck/mod.rs
```diff
@@ -353,6 +352,11 @@ pub struct Envs<'a> {
     local: Environment,
 }
 
+#[derive(Clone, Debug)]
+pub enum EnvBuildError {
```
yannham, line 356: If there's only one variant (and we don't foresee additional ones), you can also use a struct `pub struct EnvBuildError(RichTerm)`. Once again, it's nitpicking.

src/typecheck/mod.rs
```diff
@@ -1023,6 +1029,31 @@ pub fn apparent_type(t: &Term, envs: Option<&Envs>) -> ApparentType {
     }
 }
 
+/// Infer type of a more complex structure.
+/// For now, it's implemented only to infer the type of a record on wich fields are typed.
+/// It's use esencialy to type the stdlib, but could be use in a future for types placeholders for
```
yannham, line 1034:
```suggestion
/// It's used essentially to type the stdlib, but could be use in a future for types placeholders for
```

src/typecheck/mod.rs
```diff
@@ -1023,6 +1029,31 @@ pub fn apparent_type(t: &Term, envs: Option<&Envs>) -> ApparentType {
...
+/// instence.
```
yannham, line 1035: It won't be used for placeholders, though. We will use full-fledged, standard type inference for those. So it's really just for the stdlib.

src/typecheck/mod.rs
```diff
@@ -1023,6 +1029,31 @@ pub fn apparent_type(t: &Term, envs: Option<&Envs>) -> ApparentType {
     }
 }
 
+/// Infer type of a more complex structure.
+/// For now, it's implemented only to infer the type of a record on wich fields are typed.
```
yannham, line 1033:
```suggestion
/// For now, it's implemented only to infer the type of a non annotated record by gathering the apparent type of the fields.
```

tests/pass/typechecking.ncl
```diff
@@ -170,7 +170,15 @@ let typecheck = [
 
   // recursive_records_quoted
   {"foo" = 1} : {foo : Num},
-] in
 
+  // stdlib typechecking:
+  strings.length : Str -> Num,
+  (strings.length "Ok") : Num,
+  (strings.length "Ok" == 2) : Bool,
+  // parcial application
```
yannham, line 178:
```suggestion
  // partial application
```

yannham (approved):
Left a few nitpicks, but otherwise look good. It seems some comments have disappeared though, and others need slight fixes, please don't forget to do a pass on documentation before merging.

src/cache.rs
francois-caddet: Yes it is ;)
Will rename it.

src/cache.rs
francois-caddet: Lol, I was thinking the oposit ;) both function return an `Environment` so the env part was useless for me. but you're right, it's not that evident actualy.

src/repl/mod.rs
francois-caddet: I don't know, I did it because:
- It keep both mutable envs grouped (separated from the init_type_env)
- It keep the `new` function simple (only the object construction and return without any variable declaration).
- I thought the indirection is small and I suppose well optimized by Rustc.

so I will let it as it is for now and if it's an issue at one point, it can be revert easily.

src/repl/mod.rs
francois-caddet: Actualy it has been moved to `cache::GlobalEnv::type`

src/typecheck/mod.rs
francois-caddet: It's the same in `eval::EnvBuildError` so to stay consistent, I used the same form. Also, will bee easyer to update with others errors types if needed.

PR #499, opened 2021-12-09
[Language Server] tracing to csv

ysndr:
Implements a tracer that can be enriched with data by different parts of the lsp
The tracer module implements how to process certain elements of data and turn them into csv records
The tracing is activated by providing `--trace FILE` on the command line.
If disabled, the implementation aims to incur as little of a overhead as possible.
An explicit feature gate is possible but not likely to influence the performance significantly.


lsp/nls/src/requests/symbols.rs
```diff
@@ -24,12 +28,15 @@ pub fn handle_document_symbols(
...
-                        .or_else(|| Some((file_id.clone(), (0usize..0usize))))
+                        .or_else(|| Some((file_id, (0usize..0usize))))
                         .map(|(file_id, range)| {
                             codespan_lsp::byte_span_to_range(server.cache.files(), file_id, range)
                                 .unwrap()
                         })
                         .unwrap();
+
+                    // `deprecated` is a required field but causes a warning although we are not using it
+                    #[allow(deprecated)]
```
yannham, line 39: Yeah, I really don't understand how you are supposed to handle those deprecated warning for struct fields. It's still in the type definition, it's not like we have the choice...

lsp/nls/src/trace.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+pub mod param {
+
+    use super::{Enrich, ResultExt, Trace};
+    use lsp_server::RequestId;
+    use nickel::typecheck::linearization::Completed;
+
+    pub trait Enrichment {}
```
yannham, line 229: What is the role of `Enrichment` with respect to `Enrich<T>` ? Could we not just use `Enrich<T>` constraints instead of the additional `Enrichment`?

yannham (approved):
For the record, to make it work in Neovim (:CocConfig), I changed

```
    "nickel-language-server": {
      "command": "nls",
      "filetypes": ["nickel"]
```

to

```
    "nickel-language-server": {
      "command": "nls",
      "args": [
          "--trace=/path/to/file"
      ],
      "filetypes": ["nickel"]
```

lsp/nls/src/trace.rs
ysndr: Yes it's a bit of a useless marker trait here
I started off with something more generic but still typesafe in mind but settled with the current impl which at least  works for our use-case.

lsp/nls/src/server.rs
```diff
@@ -124,31 +133,31 @@ impl Server {
...
+                    serde_json::from_value::<DidChangeTextDocumentParams>(notification.params)?,
                 )
-                .unwrap();
             }
-            _ => {}
-        }
+            _ => Ok(()),
+        }?;
+
+        Ok(())
```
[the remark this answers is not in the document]
ysndr: Yes I can, good catch

lsp/nls/src/trace.rs
yannham: I think you could literally replace `Enrichment` with `Enrich` and the `E` parameter of `Enrich` by `Self`? It's not very important but if it is the case it's probably better to remove superfluous code.

PR #501, opened 2021-12-11
Add curried operator syntax

yannham:
Add support for a curried operator `(op)`, such as `(==)`. This is handy to pass operators to higher-order functions or to use them in a pipeline:

```
[1,2,3,4]
|> lists.map ((+) 2)
|> lists.filter ((<=) 4)
|> (==) [4,5,6]
```

A curried operator like `(==)` is desugared at parsing time as `fun x1 x2 => x1 == x2`.

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -371,32 +371,67 @@ switch_case: SwitchCase = {
     "_" "=>" <t: Term> => SwitchCase::Default(<>),
 }
 
+InfixBOp2: BinaryOp = {
+    "++" => BinaryOp::StrConcat(),
+    "@" => BinaryOp::ListConcat(),
+}
+
+InfixBOp3: BinaryOp = {
+    "*" => BinaryOp::Mult(),
+    "/" => BinaryOp::Div(),
+    "%" => BinaryOp::Modulo(),
+}
+
+InfixBOp4: BinaryOp = {
+    "+" => BinaryOp::Plus(),
+    "-" => BinaryOp::Sub(),
+}
+
+InfixUOp5: UnaryOp = {
+    "!" => UnaryOp::BoolNot(),
+}
+
+InfixBOp6: BinaryOp = {
+    "&" => BinaryOp::Merge(),
+}
+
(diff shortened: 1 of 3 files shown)


src/grammar.lalrpop
```diff
@@ -372,6 +371,43 @@ switch_case: SwitchCase = {
     "_" "=>" <t: Term> => SwitchCase::Default(<>),
 }
 
+InfixOp: InfixOp = {
```
regnat, line 374: Having to duplicate the parsing rule for all the operators (depending on whether they are used infix or not) is a bit fishy. Isn’t there a way to factor all this with the rules of `InfixExpr`?

regnat (changes requested):
:+1: on the principle, but I’m a bit puzzled by all the duplication in the parser. Is there no way of avoiding it?

src/grammar.lalrpop
yannham: Mmh. I think the problem is that `InfixExpr` is way finer, as it differentiates between unary/binary but most importantly with respect to precedence. What I can do is to make rules for each combination in the matrix arity/precedence. That will make `InfixExpr` less verbose, although id adds one level of indirection. `InfixOp` will be a bit less verbose, but you'll still have to say that you want all levels, like:

```
UnaryOp1,
BinaryOp1,
BinaryOp2,
BinaryOp3,
...,
BinaryOp10
```

yannham: Tried in https://github.com/tweag/nickel/pull/501/commits/e6af7d9bddd836ea12e39b441eeec743884f74ad. I'm not sure how I feel about it, it does reduces duplication but also added more boilerplate/indirection, with a net addition of ~ 100 LoC (taking into account that I forgot the `!=` in the previous version)

regnat: That’s indeed more indirection than what I thought. I like the new version better (esp given the “I forgot the != in the previous version” comment which makes me fear that this duplication could be hiding some other weird bugs) but I don’t have a strong opinion either way, so leaving that up to you

regnat (approved):

Issue #502, opened 2021-12-13
Stack overflow in `test/pass/x.ncl` with `cargo test`


litchipi:
**Describe the bug**
After adding new features to the code (#470), some tests fails because their stack overflows.

**To Reproduce**
Checkout branch `timcer/typeenv` or the master after #470 is merged, and copy / paste some of the lines of `test/pass/contracts.ncl`.
Can also be done with master right now, but you'll need to copy / paste more of the lines.
Will fail on every test that is constructed as such (taken from `test/pass/basics.ncl`):
```
let Assert = fun l x => x || %blame% l in

(1+1 == 2 | #Assert) &&
(1-2+3-4 == -2 | #Assert) &&
(2-3-4 == -5 | #Assert) &&
(-1-2 == -3 | #Assert) &&
(2*2 + 2*3 - 2*4 == 2 | #Assert) &&
(...)
(let x = 3 in (switch { `foo => 1, _ => x} (3 + 2)) == 3) &&

true
```

Then perform a `cargo test`. Note that using `nickel -f` may not fail as it allocates more stack.

**Expected behavior**
No stack overflow

**Environment**
 - Ubuntu 21.10, with Linux 5.13.0-22
 - Can be reproduced with a lot of copy / paste on `master`
 - Can be reproduced with few copy / paste after PR #470

**Additional context**
For a normal evaluation, 8Mib of stack is allocated, whereas for `cargo test` it only allocates 4Mib.
The problem comes from the recursive functions in the code, that adds a lot to the stack.
The reason is that big `A && B && C && D` create a long long tree branch that gets handled in the code with
much more stack usage than the same number of nodes dispersed in a balanced tree.

**Fix proposed**
- For a quick fix in order to handle this rapidly, rewrite the `test/pass/*.ncl` tests
- For a full fix, rewrite parts of the codes that are recursive into a looping process.

litchipi:
Note: is maybe important enough to have its place in the pre-release tracking issue (#462)

steshaw:
@yannham The PR referred to here (https://github.com/tweag/nickel/pull/470) was not merged. However, it refers to a problem of stack growth on `master`. Here is a Nickel file that blows the stack on `master` currently:

https://gist.github.com/steshaw/b5f25681795699c283d4f2c4a09df662

steshaw:
[...]

PR #503, opened 2021-12-13
Rewrite tests using list of booleans instead of chained &&

litchipi:
Temporary fix for #502 necessary so #470 doesn't make test overflows randomly when modifying the code

### DIFF
--- a/tests/pass/basics.ncl
+++ b/tests/pass/basics.ncl
@@ -1,37 +1,38 @@
 let Assert = fun l x => x || %blame% l in
 
+[
 // basic arithmetic
-(1+1 == 2 | #Assert) &&
-(1-2+3-4 == -2 | #Assert) &&
-(2-3-4 == -5 | #Assert) &&
-(-1-2 == -3 | #Assert) &&
-(2*2 + 2*3 - 2*4 == 2 | #Assert) &&
-(1/2 + 1/4 - 1/8 == 0.625 | #Assert) &&
-((10 + 1/4) % 3 == 1.25 | #Assert) &&
-(10 + 1/4 % 3 == 10.25 | #Assert) &&
-(34 + (if true then 2 else 222)
-  == 36
-  | #Assert) &&
+1+1 == 2,
+1-2+3-4 == -2,
+2-3-4 == -5,
+-1-2 == -3,
+2*2 + 2*3 - 2*4 == 2,
+1/2 + 1/4 - 1/8 == 0.625,
+(10 + 1/4) % 3 == 1.25,
+10 + 1/4 % 3 == 10.25,
+34 + (if true then 2 else 222)
+  == 36,
 
 // comparisons
-(1 < 1 == false | #Assert) &&
-(1 <= 1 == true | #Assert) &&
-(1 > 1 == false | #Assert) &&
-(1 >= 1 == true | #Assert) &&
-(1 + 1/2 > 1 + 1/4 == true | #Assert) &&
-(1 + 1/2 < 1 + 1/4 == false | #Assert) &&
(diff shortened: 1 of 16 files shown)


tests/pass/basics.ncl
```diff
@@ -1,37 +1,39 @@
...
 // basic arithmetic
-(1+1 == 2 | #Assert) &&
-(1-2+3-4 == -2 | #Assert) &&
-(2-3-4 == -5 | #Assert) &&
-(-1-2 == -3 | #Assert) &&
-(2*2 + 2*3 - 2*4 == 2 | #Assert) &&
-(1/2 + 1/4 - 1/8 == 0.625 | #Assert) &&
-((10 + 1/4) % 3 == 1.25 | #Assert) &&
-(10 + 1/4 % 3 == 10.25 | #Assert) &&
+(1+1 == 2 | #Assert),
```
yannham, line 5: I guess we can remove those asserts too.

tests/pass/basics.ncl
```diff
@@ -1,37 +1,39 @@
...
-(1 + 1/2 + 1/8 < 1 + 1/4 + 1/4 == false | #Assert) &&
-(-1 - 2 < 3 - 10 == false | #Assert) &&
-(-1 - 2 > 3 - 10 == true | #Assert) &&
-(-1*2 > 1*2 == false | #Assert) &&
-(-1*2 < 1*2 == true | #Assert) &&
-(1/4 + 1/4 - 1/4 + 1/4 <= 1/2 == true | #Assert) &&
-(1/4 + 1/4 - 1/4 + 1/4 < 1/2 == false | #Assert) &&
-(1/4 + 1/4 - 1/4 + 1/4 >= 1/2 == true | #Assert) &&
-(1/4 + 1/4 - 1/4 + 1/4 < 1/2 == false | #Assert) &&
+(1 < 1 == false | #Assert),
```
yannham, line 18: Same thing.

tests/pass/builtins.ncl
```diff
@@ -1,30 +1,32 @@
 let Assert = fun l x => x || %blame% l in
 
+[
 // is_record
-(builtins.isRecord {} | #Assert) &&
+(builtins.isRecord {} | #Assert),
```
yannham, line 5: Same thing

tests/pass/contracts.ncl
```diff
@@ -1,65 +1,66 @@
...
+f `"bar:baz" == 2),
 
 (let f : forall r. <foo, "bar:baz" | r> -> Num =
   fun x => switch { `foo => 1, `"bar:baz" => 2, _ => 3, } x in
-f `"boo,grr" == 3) &&
+f `"boo,grr" == 3),
 
 // records_simple
-(({} | {}) == {} | #Assert) &&
+(({} | {}) == {} | #Assert),
```
yannham, line 56: Same thing.

tests/pass/eq.ncl
```diff
@@ -1,66 +1,67 @@
 let Assert = fun l x => x || %blame% l in
 
+[
 // basic
-(0 == 0 + 0 + 0 | #Assert) &&
-(true == (if true then true else false) | #Assert) &&
-("a" ++ "b" ++ "c" == "#{"a" ++ "b"}" ++ "c" | #Assert) &&
-(`Less == `Less | #Assert) &&
-(`small == `small | #Assert) &&
+(0 == 0 + 0 + 0 | #Assert),
```
yannham, line 5: Same thing.

yannham (changes requested):
If we use `#Assert` in the fold, we should probably get rid of all the final `#Assert`s of each individual tests (but keep different asserts like `#LocalAssert`, though).

yannham (approved):

Issue #494, opened 2021-12-09
Block syntax for let-binding


yannham:
**Is your feature request related to a problem? Please describe.**
Block let-binding are nice to express a series of order-independent bindings. See #218 for more context. 

**Describe the solution you'd like**
As decided during the [standardization meeting](https://github.com/tweag/nickel/blob/master/notes/standardization-meeting-07.12.21.md), have let-binding blocks as in Nix, just replacing semicolon `;` by `,` for consistency with the rest of the Nickel syntax.

Example:
```
let foo = ...,
    bar = ...,
    blah = ...,
in
...
```

To be consistent, the inner syntax should be exactly the same as for records, just without the braces. Those let-binding would also be mutually recursive, as records.

**Describe alternatives you've considered**
See #218 for discussion and alternatives.

yannham:
As noticed by @litchipi, it may be a bit strange that single let-binding aren't recursive by default, and block let-binding are. That is, turning

```
let x = x + 1 in
let y = y + 1 in
...
```
into
```
let x = x + 1,
    y = y + 1,
in
```
changes a benign shadowing to infinite recursion. I see three possibilities:

1. Make single let-binding recursive by default. 
2. Make block let-binding non recursive by default. Then we could have a `rec` keyword that would work for both let forms, and turn them into recursive bindings.
3. Do nothing, and have this discrepancy between single lets and let block.

I'm in favor of 2. While I understand that one may want recursive records by default, 1. sounds extreme semantically and totally disallows shadowing altogether. I've never seen a language with such a feature. 3 makes things unexpected, as shown by the example. 2 is consistent, make recursion explicit for let-binding (which is IMHO sensible, as most of the time let-binding are not recursive) while keeping it implicit for records, as we decided to do in #216.

Ping @edolstra @aspiwack

aspiwack:
I'm partial to option (2) as well.

mboes:
I'm not a fan of the proposed syntax. The parallel with records is non-obvious. It would be if the braces were there, though they are of course redundant. In records, I can have (for good reason) trailing commas. Here is would look like this:

```
let x = 2, in x
```

We could add the braces back, so e.g.

```
let { x = 1 + y, y = 1 + x, } in x + y
```

This would have the added benefit of making https://github.com/tweag/nickel/issues/494#issuecomment-992715769 a non-issue. But now people will wonder why they can't write

```
let my_rec = { x = 1 + y, y = 1 + x, } in
let my_rec in x + y
```

[...]

PR #504, opened 2021-12-13
Implementing let blocks, adding non-regression test

litchipi:
Implementation of let blocks in the form:
```
let foo = 3,
    bar : Num = 2,
    message = "Hello World",
    f = fun x => x,
    f2 : Num -> Bool = fun x => x == 2,
in
```
Adds a new node `LetBlock` in the AST that basically mimic all actions from simple `Let` term, simply looping through all the definitions and adding them to the environment before passing the environment to the final term.

Made for feature request #494 
Waiting for PR #502 and #470 to be merged.
Requesting reviews on the regression test to be sure to cover edge cases.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -591,6 +591,16 @@ where
                 env.insert(x, Thunk::new(closure, IdentKind::Let()));
                 Closure { body: t, env }
             }
+            Term::LetBlock(defs, rest) => {
+                for (id, term) in defs.into_iter() {
+                    let closure = Closure {
+                        body: term,
+                        env: env.clone(),
+                    };
+                    env.insert(id, Thunk::new(closure, IdentKind::Let()));
+                }
+                Closure { body: rest, env }
+            }
             Term::Switch(exp, cases, default) => {
                 let has_default = default.is_some();
 
@@ -944,6 +954,14 @@ pub fn subst(rt: RichTerm, global_env: &Environment, env: &Environment) -> RichT
                 let t2 = subst_(t2, global_env, env, bound);
 
                 RichTerm::new(Term::Let(id, t1, t2), pos)
+            }
+            Term::LetBlock(defs, rest) => {
+                let mut data = vec![];
+                for (id, term) in defs.into_iter() {
+                    data.push((id, subst_(term, global_env, env, Cow::Borrowed(bound.as_ref()))));
+                }
+                RichTerm::new(Term::LetBlock(
+                        data, subst_(rest, global_env, env, bound)), pos)
             }
             Term::App(t1, t2) => {
                 let t1 = subst_(t1, global_env, env, Cow::Borrowed(bound.as_ref()));
--- a/src/parser/utils.rs
+++ b/src/parser/utils.rs
@@ -106,6 +106,24 @@ pub fn elaborate_field_path(
     (fst, content)
 }
 
+pub fn build_letblock<I>(fields: I, rest: RichTerm) -> Term
+where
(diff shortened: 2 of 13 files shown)


src/eval.rs
```diff
@@ -945,6 +955,14 @@ pub fn subst(rt: RichTerm, global_env: &Environment, env: &Environment) -> RichT
 
                 RichTerm::new(Term::Let(id, t1, t2), pos)
             }
+            Term::LetBlock(defs, rest) => {
+                let mut data = vec![];
+                for (id, term) in defs.iter() {
+                    data.push((id.clone(), subst_(term.clone(), global_env, env, Cow::Borrowed(bound.as_ref()))));
```
yannham, line 961: Same remark about cloning.

src/parser/utils.rs
```diff
@@ -107,6 +107,25 @@ pub fn elaborate_field_path(
     (fst, content)
 }
 
+
+pub fn build_letblock<I>(fields: I, rest: RichTerm) -> Term
+where
+    I: IntoIterator<Item = (Ident, RichTerm)> + Debug,
```
yannham, line 113:
```suggestion
    I: IntoIterator<Item = (Ident, RichTerm)>,
```

I think the `Debug` thing is a leftover from development in the `build_record`. If it compiles without, let's get rid of it

src/term.rs
```diff
@@ -912,6 +927,17 @@ impl RichTerm {
                     pos,
                 }
             }
+            Term::LetBlock(defs, rest) => {
+                let mut data = vec![];
+                for (id, term) in defs.iter() {
+                    data.push((id.clone(), term.clone().traverse(f, state, method)?));
+                }
```
yannham, line 934: Same remark about cloning. Also just for the sake of mentioning it, you can use `collect` on a `Iterator<Item = Result<_>>` to get a `Result<Vec<_>, _>`, something like:

```
let data : Result<Vec<_>, _> = defs
    .into_iter()
    .map(|id, t| Ok((id, t.traverse(f, state, method)?)))
    .collect();
...
                     term: Box::new(Term::LetBlock(data?, rest)),
```

I'm not saying you should necessarily change it (it's a matter of taste in the end), but just FYI.

src/typecheck/mod.rs
```diff
@@ -639,6 +639,23 @@ fn type_check_<S, E>(
             envs.insert(x.clone(), ty_let);
             type_check_(state, envs, lin, linearizer, strict, rt, ty)
         }
+        Term::LetBlock(defs, rest) => {
+            for (id, term) in defs.iter() {
+                let ty_let = binding_type(term.as_ref(), &mut env, global, state.table, strict);
+                type_check_(state, global, lin, linearizer.scope(ScopeId::Left), strict, term, Closure::new(ty_let.clone(), env.clone()))?;
+                env.insert(id.clone(), Closure::new(ty_let, env.clone()));
+            }
```
yannham, line 647: Same thing as with evaluation, you need to typecheck this in a recursive environment. You can look at the `RecRecord` case to see how to it. We may even separate this typechecking of mutually recursive bindings in a separate helper as I expect the code to be almost the same for let blocks and recursive records.

yannham (commented):
The overall changes look reasonable (it misses the grammar but I imagine it's because it's still draft), but this doesn't implement the semantics of #494 that states let-blocks should be mutually recursive, as for record, and not just equivalent to a series of simple lets. See the corresponding comments.

Also, tagging @ysndr as this may need additional support in the LSP.

PR #506, opened 2021-12-14
Avi/language server/vs code client

Avi-D-coder:


lsp/README.md
```diff
@@ -46,4 +54,32 @@ The VS Code extension offers three configuration options:
 
 - `"nls.server.path"`: Path to nickel language server
 - `"nls.server.trace"`: "Enables performance tracing to the given file"
-- `"nls.server.debugLog"`: "Logs the communication between VS Code and the language server."
```
ysndr, line 49: Why should we remove that line?

lsp/README.md
```diff
@@ -46,4 +54,32 @@ The VS Code extension offers three configuration options:
...
+```
+
+### Neo(Vim) with Coc.nvim
+
+Add the `nickel_ls` `JSON` object to your `:CocConfig`/`coc-settings.json`.
+```
+{
+  "languageserver": {
+    "nickel_ls": {
+      "command": "nls",
```
ysndr, line 75: Can you add a comment, that `command` would be `nls --trace <file>` to enable performance tracing?

lsp/README.md
```diff
@@ -18,6 +18,14 @@ To enter a shell with `nickel` and `nls` run
 nix shell github:tweag/nickel
 ```
 
+Alternatively to install both using nix run:
+
+```
+git clone https://github.com/tweag/nickel.git
+cd nickel
+nix-env -f . -i
```
ysndr, line 26: Should we rather endorse the new nix command `nix build` or mention that this is the method to install without flakes?

ysndr (approved):
Thank you, generally looking good, just left some comments comments

lsp/README.md
Avi-D-coder: I have no preference.

lsp/README.md
Avi-D-coder: Good catch. That was a mistake.

lsp/README.md
ysndr: We should at least mention `nix build` or `nix profile install` as they are turning into the standard commands for nix

Avi-D-coder: Added

ysndr (approved):

PR #511, opened 2021-12-28
Do not discard position after evaluation

### LINKED ISSUE #510: Non-serializable error on top-level term shows no location
**Describe the bug**
If the top-level term one is trying to serialize (e.g. via `nickel export`) isn't serializable (typically a function), the "non-serializable term" error doesn't show any position, which makes it hard to understand where is the problem. This can typically happen when a function is not fully applied:

**To Reproduce**
```
// gen.ncl
fun param1 param2 => {
  foo = param1,
  bar = param2,
}

$nickel export <<< 'import "gen.ncl" "fooval"'
error: non serializable term
  │
1 │ <func>
  │ ^^^^^^
```

**Expected behavior**
The error should show the correct location of the function inside `gen.ncl`, as it does already for any term that is not the root node of the AST.

yannham:
Fix #510. The `eval` family previously returned a `Term`, discarding the position information of the root node of the AST. It now returns a `RichTerm`, and the caller is free of using or discarding the position information.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -1339,7 +1339,7 @@ mod tests {
 
     /// Evaluate a term without import support.
     fn eval_no_import(t: RichTerm) -> Result<Term, EvalError> {
-        eval(t, &Environment::new(), &mut DummyResolver {}).map(|rt| *rt.term)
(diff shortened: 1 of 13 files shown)


ysndr (approved):
Don't see any out of the ordinary, looks good to me.

PR #512, opened 2021-12-28
Fix uninformative "empty metavalue" error

### LINKED ISSUE #480: Better error message for missing field definition ("empty metavalue")
When applying a contract-as-a-record to a value, if we try to access a field that is required by the contract but is not defined by the value, we hit the following puzzling error message:

```
nickel> let Contract = {foo | Dyn}
nickel> let data | #Contract = {}
nickel> data.foo
error: empty metavalue
  ┌─ repl-input-0:1:17
  │
1 │ let Contract = {foo | Dyn}
  │                 ^^^^^^^^^ here
```

Beside the error message being unhelpful, the main issue is that this error shows the location of the definition of the contract, but we have no idea where is the problematic **value**. The same contract could literally be applied to hundreds of different values, for all we know. Moreover, it doesn't show when the contract has been applied to the specific value either.

A clearer error message (like `missing definition for required field bar`), as well as the location of the value that is missing this definition. Additionally, having the location of where the original contract is applied to this particular value may also be helpful.

**Additional context**
The implementation is not totally straightforward. The reason for this message is that when we finally try to unwrap the metavalue and find it is not defined, we already forgot the following pieces of information:

1. What field we were projecting (here, `foo`)
2. The label of the application of the contract to the value (here, corresponding to `data | #Contract`.

We need to remember these things in one way or another (the callstack, checking right after projecting that the value is not an empty metavalue, etc.)

yannham:
~~Close #480~~ Partially fix #480 (see comments below). When a field without a definition is accessed, the error should now show the faulty record and the location of the field access. Unfortunately, the location of the record is currently lost because of the contract application pass, which doesn't set all the positions of the additional intermediate terms. However, this is a separate issue, and should be fixed by #509 (or a superseding one).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -44,7 +44,7 @@ pub enum EvalError {
     /// A blame occurred: a contract have been broken somewhere.
     BlameError(label::Label, CallStack),
     /// A field required by a record contract is missing a definition.
-    MissingFieldDef(label::Label, CallStack),
+    MissingFieldDef(Option<label::Label>, CallStack),
     /// Mismatch between the expected type and the actual type of an expression.
     TypeError(
         /* expected type */ String,
@@ -917,8 +917,8 @@ impl ToDiagnostic<FileId> for EvalError {
 
                 diagnostics
             }
-            EvalError::MissingFieldDef(l, callstack) => {
-                use crate::eval::StackElem;
+            EvalError::MissingFieldDef(label, callstack) => {
+                use crate::eval::callstack::StackElem;
 
                 // The following code determines what was the last accessed record field by looking
                 // at the call stack. Because of recursive records though, the fields may actually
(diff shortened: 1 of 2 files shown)


yannham:
This doesn't help when fields are forced using `deepSeq`, which is precisely the case when serializing a value. Switching back to draft while coming up with a solution for this case.

yannham:
This is not as trivial as anticipated, due to the way `DeepSeq` is implemented. Let's proceed with this one and treat the case of `DeepSeq` in a different one.

src/eval.rs
```diff
@@ -1049,10 +1055,9 @@ where
                         }) => Closure { body: inner, env },
                         _ => unreachable!(),
                     }
-                }
-                // TODO: improve error message using some positions
-                else {
-                    return Err(EvalError::Other(String::from("empty metavalue"), pos));
+                } else {
+                    let label = meta.contracts.pop().unwrap().label;
```
francois-caddet, line 1059: In some cases (e.g.: destructuring) the metavalue is a dummy metavalue. So it does not contains any value but  not any contracts neither. It may be an issue related to how is performed the destructuring now. Anyway, before merging, at least avoid a panic here is better.

francois-caddet (approved):
It look good. I only have a small question before merging. It's in the discutions.

src/eval.rs
yannham: Good point. Maybe the error should take the label as an option then. That or we must enforce that any empty metavalue always come with a label, which does sound reasonable and is the case for anything that comes out of the parser. But it's still yet another invariant not verified by the compiler that we must maintain.

francois-caddet: Yes require a label for every metavalues song good. But I don't think we can do it staticaly because labels are attached to contracts which can be in 2 differents places. If we want, how to make it required? In which field (`types`/`contracts`)?

yannham: It's not as simple than that: a metavalue with a non-empty `value` may very well have no label, and it's totally fine. What's not fine is a metavalue that is both empty and without label. Bu it's current panicking on destructuring, so let's handle the case where no label is defined before merging.

yannham: Keeping in mind that, still, destructuring is not doing exactly the right thing right now.

PR #513, opened 2021-12-28
style: Use Repl* instead of REPL*

mboes:
In camelCase naming, acronyms do not remain fully capitalized unless
they are trivially short (like IO). This was pointed out by Clippy.

See https://rust-lang.github.io/api-guidelines/naming.html


yannham (approved):
Let's do the same on all the `WASM` stuff.

PR #514, opened 2021-12-28
[Style] Use Wasm* instead of WASM*

yannham:
Depend on #513. Follow-up of #513, now for WASM: use `Wasm` instead of `WASM` as prescribed by CamelCase.


mboes (commented):
Agreed with the code change. But the comments should nevertheless stay as they are.

mboes (approved):

yannham:
Indeed, I replaced too blindly. This is fixed in the last commit.

PR #516, opened 2021-12-28
[LSP] Rework the README a bit

yannham:
Reorganize the README of #505, and add a `cargo`-based installation method.

### DIFF
--- a/lsp/README.md
+++ b/lsp/README.md
@@ -12,9 +12,9 @@ it in VSCode and (Neo)Vim.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


ysndr (approved):
Should we also add  a mention to the `--trace` argument for vim users?

yannham:
> Should we also add a mention to the --trace argument for vim users?

Probably, for consistency, even if we'll probably be the only one to use it for some time. I've never used the built-in LSP client though, so let me find out how to do first.

ysndr (approved):

PR #518, opened 2021-12-28
Switch to snake_case by convention for values

### LINKED ISSUE #493: Case convention: use snake_case for values
The current informal (not enforced by the parser) case convention is to use camelCase for values and CamelCase for types and contracts.

As decided during the [standardization meeting](https://github.com/tweag/nickel/blob/master/notes/standardization-meeting-07.12.21.md), adopt the same convention as Rust: snake_case for values and CamelCase for types/contracts. For now, this means changing the stdlib, examples and tests to reflect this new convention. Enforcing this convention at parsing time is out of scope of this specific issue (and may not even be possible if we merge type and values syntax).

mboes:
This renames all builtins and standard library functions to conform to
a snake case convention, as per decision recorded in #493. This also
adapts examples and tests to this change. All functions have been
renamed the way you'd expect, except the following:

* `fromPred` -> `from_predicate`
* `typeOf` -> `typeof` (like in C)
* `fieldsOf` -> `fields` (the `_of` is redundant for a function)
* `valuesOf` -> `values` (the `_of` is redundant for a function)

Closes #493


examples/config-gcc/config-gcc.ncl
```diff
@@ -51,7 +51,7 @@ let OptLevel = fun label value =>
     contracts.blame label in
 
 let Contract = {
-  pathLibC | doc "Path to libc."
+  path_libc | doc "Path to libc."
```
yannham, line 54: Should we apply the change to configuration option? In real life, their name is dictated by the consuming tool (say, Terraform), most commonly in camelCase than in snake_case.

examples/record-contract/record-contract.ncl
```diff
@@ -11,7 +11,7 @@ let PortElt
   | doc "A contract for a port element of a Kubernetes configuration"
   = {
     name | Str,
-    containerPort | #Port,
+    container_port | #Port,
```
yannham, line 14: Illustrating the previous comment, the field names in this example are taken from actual Terraform schema, so we don't have the freedom of picking the case.

stdlib/builtins.ncl
```diff
@@ -1,90 +1,90 @@
...
-        isRecord { hello = "Hello", world = "World" } =>
+        is_record { hello = "Hello", world = "World" } =>
           true
       ```
       "#m
-    = fun x => %isRecord% x,
+    = fun x => %is_record% x,
 
-    typeOf : Dyn -> <
+    typeof : Dyn -> <
```
yannham, line 87: Should we really follow C/C++ precedent here, though? For example, [this random Rust crate](https://docs.rs/typed/1.0.0/typed/fn.type_of.html) uses `type_of`, which IMHO is more consistent with our and Rust's naming convention.

yannham (commented):
LGTM overall, but the fields exported in a configuration should not all be automatically capitalized in snake_case in the examples, I think.

stdlib/builtins.ncl
mboes: The value of any `foo` applied to `baz` is the `foo` _of_ `baz`. So any `_of` in a function is redundant and shouldn't be part of any naming convention. The only reason something like it is there in C/C++ and elsewhere is presumably because `type` is a reserved keyword. I saw the "random Rust crate" you mention, but with 480 downloads over its lifetime, it hardly has enough clout to set a meaningful precedent.

mboes (commented):
> the fields exported in a configuration should not all be automatically capitalized in snake_case in the examples, I think.

I agree that configuration should use whatever schema is imposed by upstream. But in that case we should be explicitly referencing what schema we are assuming. API fields are no more commonly camelCase than snake_case IME. It really depends.

examples/record-contract/record-contract.ncl
yannham: The official reference is https://kubernetes.io/docs/reference/kubernetes-api/. Another useful and better tailored resource is the original example it was inspired from (IIRC): https://github.com/kubernetes/examples/blob/master/guestbook-go/guestbook-controller.json.

stdlib/builtins.ncl
yannham: Ok, fair enough. By the way `type` is currently not a reserved keyword in Nickel, and we've been trying to avoid to do so when possible (such as with type aliases) as it may easily appear in configurations and would be annoying to escape everywhere (e.g. in the current schema of NixOS modules). However it is not be good idea to tie our hands now and prevent any future use of a `type` keyword, so let's go with `typeof`.

mboes: Thinking about this more: in C/C++, typeof is an operator. Probably has that name because no other keyword contains an underscore. In Python, the equivalent is a function called `type()`. Could we do the same here? `type` doesn't appear to be a reserved keyword yet and maybe we never want it to be. I have provisionally updated the PR accordingly.

mboes:
@yannham PTAL

stdlib/builtins.ncl
mboes: Hm, don't know what to make of the message timestamps above. You somehow answered my question from 24 minutes in the future. :)

I've kept `typeof` then. It's easier to go `typeof` -> `type` later, than the other way around.

yannham: I actually answered from the future indeed :stuck_out_tongue:  I wrote my reply as an answer to your first comment without seeing the following ones.

examples/record-contract/record-contract.ncl
```diff
@@ -37,7 +38,7 @@ let KubernetesConfig = {
              | doc "The number of replicas"
              | default = 1,
 
-    selector.matchLabels.app | Str,
+    selector.match_labels.app | Str,
```
yannham, line 41:
```suggestion
    selector.matchLabels.app | Str,
```

examples/record-contract/record-contract.ncl
```diff
@@ -60,7 +61,7 @@ let metadata_ = {
     replicas = 3,
     selector = {
       app.name = name_,
-      matchLabels.app = name_,
+      match_labels.app = name_,
```
yannham, line 64:
```suggestion
      matchLabels.app = name_,
```

yannham (approved):
Beside two leftover in the Kubernetes example, it's good to go.

PR #527, opened 2021-12-30
Fixup conversion to snake_case

mboes:
In #518, I had forgotten to change the names of the primitives in
`stdlib/strings.ncl`. I noticed the omission because this was breaking
the example in the nickel-lang.org playground.

### DIFF
--- a/stdlib/strings.ncl
+++ b/stdlib/strings.ncl
@@ -16,7 +16,7 @@
       ```
       "#m
     = fun l s =>
-      if %isStr% s then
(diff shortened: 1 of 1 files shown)


yannham:
Let me investigate this a bit more, because there's something really strange going on. It should have break pretty much everything, as the stdlib shouldn't even parse, but it seems the parser still understands stuff with the old case convention.

yannham (approved):
Got it, this is yet another manifestation of the parse error tolerance PR.

PR #529, opened 2021-12-31
Add benchmarks

### LINKED ISSUE #332: Benchmarks

Avi-D-coder:
Includes/Supersedes #333
> Depend on #284 #296. Close #332.
>
> -    add various basic benchmarks, run using the criterion crate.
> -    add an action to automatically run them and compare them against the result of master on each PR.
>
> Criterion is a statistical benchmarking tool. However, comparison results must be taken with prudence: I observed important, statically significant variations on my local machine between 2 consecutive runs in a relatively stable environment. It can still give us an idea about the impact of some changes.

After @ㅤyannham's recursive record optimization these only used around 24GB of memory.
~Other recent improvements on master have brought peak memory consumption down to 11GB!~ (this was likely an error causing deepseq tests to not be fully run).
The difference in peak memory consumption is on the lists benchmark. I'm unsure what caused such a large change in memory consumption. I expect usage will be further improved by #426.


.github/workflows/continuous-integration.yml
```diff
@@ -7,6 +7,14 @@ on:
...
 
 jobs:
+  run-benchmark:
+    name: run benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: jasonwilliams/criterion-compare-action@move_to_actions
```
aspiwack, line 17: This really doesn't look like something that I want to run on every push (it also does nothing on master, so let's not).

Maybe this should be in a different workflow, with a manually triggered condition (I don't know what conditions are available to us, though).

aspiwack (commented):
I haven't reviewed with any care.

I'm surprised that there is code outside of the benchmarks in this PR though. Why is this?

yannham:
>I'm surprised that there is code outside of the benchmarks in this PR though. Why is this?

We quickly implemented two little things I had in mind just on the benchmark branch to try to cut the memory leaks. 

Currently each field of a record blindly captures the whole record in its environment, even if it is not actually recursive, making basically each and every record to leak. The patch makes them only capture the recursive fields they actually use, by looking at the free variables. Then, only actually mutually recursive fields should leak. While this is what we should be doing in the end, it has been done in a slightly quick and dirty way here.

The second is to avoid building nested closures unnecessarily, when what we try to closurize is already a variable, but this one is already part of #491 anyway.

I suggest we cut both patches off this PR, to keep only benchmarks-related stuff, and re-implement the first one properly in a separate PR, while the second should land with #491.

Avi-D-coder:
Okay, we are back to running out of memory without those optimizations, and when merging master again I fixed deepseq tests not being run, so the 11GB it had very recently started using was likely still 24GB.

mboes:
> We quickly implemented two little things I had in mind just on the benchmark branch to try to cut the memory leaks.

Sounds like self-contained changes that ought to be each in their own PR? Maybe that's what you're going for with

> re-implement the first one properly in a separate PR, while the second should land with #491.

Avi-D-coder:
I have cherry picked the closurize optimization at #537.

Avi-D-coder:
The benchmark action is an expected failure.

aspiwack (approved):

yannham:
>Sounds like self-contained changes that ought to be each in their own PR? Maybe that's what you're going for with
>> re-implement the first one properly in a separate PR, while the second should land with #491.

Yes, that's what I meant.

Cargo.toml
```diff
@@ -55,9 +55,35 @@ serde_repr = { version = "0.1", optional = true }
...
+utilities = {path = "utilities"}
 
 [workspace]
 members = [
     ".",
     "lsp/nls"
 ]
+
+[profile.release]
+debug = true
```
yannham, line 69: Do we really want to include debugging symbols in the release profile? What is this for?

benches/records.rs
```diff
@@ -0,0 +1,44 @@
...
... duis ut diam quam nulla. Nisl pretium fusce id velit ut. Magnis dis parturient montes nascetur. Leo duis ut diam quam nulla porttitor massa. Non diam phasellus vestibulum lorem sed risus. Adipiscing vitae proin sagittis nisl rhoncus mattis. Imperdiet dui accumsan sit amet nulla. Venenatis a condimentum vitae sapien pellentesque habitant morbi tristique senectus.""#, // Sollicitudin ac orci phasellus egestas tellus rutrum tellus pellentesque. Venenatis cras sed felis eget velit. Tortor pretium viverra suspendisse potenti nullam ac. Pretium vulputate sapien nec sagittis aliquam malesuada. Felis eget velit aliquet sagittis id consectetur. Non consectetur a erat nam at lectus urna. Habitant morbi tristique senectus et netus. Praesent semper feugiat nibh sed. Maecenas accumsan lacus vel facilisis volutpat est velit egestas dui. Facilisi etiam dignissim diam quis. Vitae congue eu consequat ac felis donec. Maecenas sed enim ut sem viverra aliquet eget sit. Pulvinar etiam non quam lacus suspendisse faucibus. Donec pretium vulputate sapien nec sagittis aliquam malesuada. Aliquam sem fringilla ut morbi tincidunt augue interdum. Urna duis convallis convallis tellus id interdum velit laoreet.
```
yannham, line 14: We may want to put this in a separate file and include it at compile time, as we do for the the stdlib with `include_str!` (I'm not sure about the exact name)

flake.nix
```diff
@@ -155,6 +155,8 @@
               # support having different crate types depending on the target, so
               # we switch there
               sed -i 's/\[lib\]/[lib]\ncrate-type = ["cdylib", "rlib"]/' Cargo.toml
+
+              sed -i '/utilities/d' Cargo.toml
```
yannham, line 159: Is this really required? Patching `Cargo.toml` is a necessary evil but I would keep it to the minimum. If it doesn't compile to WASM because it uses io or `PathBuf` or whatever, can't we achieve the same with pure cargo, by excluding utilities when the feature `repl-wasm` (or something like that) is enabled?

Avi-D-coder: Sadly not as far as I could figure out.
Putting it behind a feature, or target did not prevent fs2 from being built.
I even tried putting every use of criterion behind another feature in the source.

There may be an incantation that works that I did not try, but I spent a silly amount of time trying to do this right, and failing. Hopefully this gets fixed on criterion end, there's already a issue for wasm. Otherwise eventually bench only deps may make it to cargo.

Cargo.toml
Avi-D-coder: I commented it out, and added a comment that says `# Enable this to use flamegraphs`.

flake.nix
yannham: I see. Then maybe you could document it quickly here (with even a link to the issue).

Avi-D-coder: Added

yannham (approved):
LGTM. It seems the WASM REPL doesn't build though. It's good to go once this is fixed.

Avi-D-coder:
> LGTM. It seems the WASM REPL doesn't build though. It's good to go once this is fixed.

Apparently, I did not get enough sleep last night ;) I just used rust comment syntax, instead of sh.

Avi-D-coder:
The lorm commit broke things.

yannham:
Something broke? I thought benchmarks not succeeding was expected and I merged.

Avi-D-coder:
Benchmarks are expected to runout of memory this last time they failed with a different error. PR on it's way.

Issue #521, opened 2021-12-29
Improve error message for empty doc string


tsmanner:
**Is your feature request related to a problem? Please describe.**
An empty `doc` string was buried in some nickel code somewhere, leading to a crash when evaluating the expression.  `nickel typecheck` has no problem with `doc ""` but other commands produce an error message which doesn't clearly indicate what, or even where, something is wrong.

Given the file `foo.ncl` with these contents
```ncl
{
  foo
    | doc ""
    = "bar",
}.foo
```

The following is the result of invoking nickel:
```
-> nickel --file foo.ncl typecheck
-> nickel --file foo.ncl
Done: ParseError
-> nickel --file foo.ncl export
error: non serializable term
  ┌─ <unknown> (generated by evaluation):1:1
  │
1 │ <parse error>
  │ ^^^^^^^^^^^^^
```

**Describe the solution you'd like**
Ideally, the error message would say something like "error: empty doc string" and print the line:column of the offending input.
At minimum, a line:column would at least point at something that can be fiddled with to fix the problem.

**Describe alternatives you've considered**
Not using `doc` anywhere.
Writing a grep or sed program to report empty or missing doc strings.

**Additional context**
A little extra musing on the topic: it seems to me that `nickel typecheck` should complain about the empty docstring, even if it crashes in the same way.  I understand that the `doc` stuff isn't part of the type system, so it doesn't necessarily make sense for it to produce an error, but the message from `export` is a parse error.  That confuses me a bit, because it implies that `typecheck` uses a different parser than the rest of nickel.

yannham:
Thanks for reporting. It sounds like yet another avatar of error-tolerant parsing. We added support for this in #456 to have the LSP still works fine when part of the source file can't be parsed. But for some reasons error-tolerant parsing seems to be activated in other places, while it was really meant to be used only for the LSP.

On current master (df33c022c30ff40f7dbd565a181c7df1fec08536), if I try a similar example in the REPL, I get a parse error instead:
```
nickel> let foo | doc "" = 1
error: Unexpected token
  ┌─ repl-input-0:1:16
  │
1 │ let foo | doc "" = 1
  │                ^
```

It could be improved, but this is at least much better than what the issue reports.
cc @Avi-D-coder

tsmanner:
Yeah, that's much better.  I'm on `66ba798` which is a couple days old.  On `df33c02` the error message from running nickel on a file are better, but still not all that informative:

`let_in.ncl`:
```ncl
[...]

PR #530, opened 2021-12-31
Fix: #521

Avi-D-coder:
Imports and doc strings will now be parsed even if they are empty.
If we want to disallow them it should be done after parsing.

This turned out not to be related to error-tolerant parsing, we just were not handling empty strings.

We now allow:
```
let foo | doc "" = 1
```

This error on empty import has been fixed as well.

Before:

```
nickel> let f = import ""
error: Unexpected token
  ┌─ repl-input-1:1:17
  │
1 │ let f = import ""
  │                 ^
```

After:

```
nickel> let f = import ""
error: Import of  failed: No such file or directory (os error 2)
  ┌─ repl-input-0:1:9
  │
1 │ let f = import ""
  │         --------- imported here
```

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -362,7 +362,10 @@ ChunkExpr: StrChunk<RichTerm> = HashBrace <t: WithPos<Term>> "}" => StrChunk::Ex
 
 HashBrace = { "#{", "multstr #{" };
 
-StaticString: String = StringStart <s: ChunkLiteral> StringEnd => s;
+StaticString: String = {
+    StringStart <s: ChunkLiteral> StringEnd => s,
+    StringStart StringEnd => String::new(),
+};
 
 EnumTag: Ident = {
     <Ident>,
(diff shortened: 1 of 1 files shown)


src/grammar.lalrpop
```diff
@@ -362,7 +362,10 @@ ChunkExpr: StrChunk<RichTerm> = HashBrace <t: WithPos<Term>> "}" => StrChunk::Ex
 
 HashBrace = { "#{", "multstr #{" };
 
-StaticString: String = StringStart <s: ChunkLiteral> StringEnd => s;
+StaticString: String = {
```
yannham, line 365: I think you can do something like this in LALRPOP, for a smaller diff:

```
StaticString: String = StringStart <s: ChunkLiteral?> StringEnd => s.unwrap_or_default();
```

yannham (approved):
I think this is fine to allow for empty strings (if we don't want to for, say, imports, it's probably easier and will give better error message to just check for emptiness separately). I let a suggestion for shorter diff.

However, I don't think this fixes #521. The issue is explicitly about the error message not pointing to the right thing, and not about Nickel accepting empty docstrings or not. It seems the REPL is handling those errors fine, but this is not the case of the basic nickel subcommands.
Example on current master:
```
$nickel <<< '1 | doc ""'
Done: RichTerm { term: ParseError, pos: Original(RawSpan { src_id: FileId(1), start: ByteIndex(0), end: ByteIndex(10) }) }
```

This shouldn't accept the program but rather report the parsing error right away, whether we accept empty docstrings or not.

PR #531, opened 2022-01-01
This change allows the criterion-compare-action to pass

Avi-D-coder:
It needs to be merged prior to #529

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,14 @@ readme = "README.md"
 description = "Programmable configuration files."
 edition = "2018"
 
+[[bin]]
+name = "nickel"
+path = "src/bin/nickel.rs"
+bench = false
+
+[lib]
+bench = false
+
 [features]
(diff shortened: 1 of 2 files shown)


yannham (commented):
So, does this disable benches for basically everything? If benches are OOM, shouldn't we rather get rid of the criterion-compare-action, at least temporarily, as it would basically do nothing?

Avi-D-coder:
This PR fixes `cargo bench -- --save-baseline` by disabling benchmarks that are not explicitly listed in `Cargo.toml`.
See https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options.

yannham (approved):
I see. I'm not sure we will keep the criterion action tbh, as benchmarking in CI may be expensive and poorly reproducible, but this fix doesn't hurt anyway.

PR #535, opened 2022-01-03
Produce a parse error diagnostic instead of a Term::ParseError in prepare

Avi-D-coder:
Addresses this issue https://github.com/tweag/nickel/pull/530#pullrequestreview-842654960
fixes with #530: #521

Both without #530
On master:
```
❯ nickel <<< '1 | doc ""'
Done: RichTerm { term: ParseError, pos: Original(RawSpan { src_id: FileId(1), start: ByteIndex(0), end: ByteIndex(10) }) }
```

On this branch:
```
❯ nickel <<< '1 | doc ""'
error: Unexpected token
  ┌─ <stdin>:1:10
  │
1 │ 1 | doc ""
  │          ^
```

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -602,8 +602,12 @@ impl Cache {
     ) -> Result<CacheOp<()>, Error> {
         let mut result = CacheOp::Cached(());
 
-        if self.parse(file_id)? == CacheOp::Done(ParseErrors::default()) {
-            result = CacheOp::Done(());
+        match self.parse(file_id)? {
+            CacheOp::Done(e) | CacheOp::Cached(e) if !e.no_errors() => return Err(e.into()),
+            CacheOp::Done(_) => {
+                result = CacheOp::Done(());
+            }
+            _ => {}
         };
 
         let import_res = self.resolve_imports(file_id).map_err(|cache_err| {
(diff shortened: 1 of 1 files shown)


yannham (changes requested):
LGTM, but while we're at it, we should also try to solve the problem when loading the stdlib as well. I have tested it and it seems parsing the stdlib still suffer from error tolerance (if I mess with a random string function the REPL builds and starts fine and only error out when I try to evaluate the specific function)

Avi-D-coder:
> (if I mess with a random string function the REPL builds and starts fine and only error out when I try to evaluate the specific function)
@yannham Can you post steps to reproduce? I'll try and debug it now.

yannham:
@Avi-D-coder 

1. go into `stdlib/strings.ncl`
2. go line 19 (inside the `BoolLiteral` contract) and replace `%is_str%` by `%%is_str%` (or whatever that is not a valid expression)
3. run `cargo run -- repl`. It shouldn't even start because the stdlib is loaded at this point.
4. execute `nickel>"true" | #strings.BoolLiteral` to finally get a strange type error.

Avi-D-coder:
Fixed

yannham (approved):

PR #537, opened 2022-01-04
Closurize optimization

Avi-D-coder:
Cherry-pick @yannham's Closurize optimization from the benchmark memory consumption optimization branch.
Original commit https://github.com/tweag/nickel/pull/529/commits/1a8fecf801b40052dd14602a844a9f15e2f1f0ba.

### DIFF
--- a/src/transformations.rs
+++ b/src/transformations.rs
@@ -489,19 +489,33 @@ pub trait Closurizable {
 }
 
 impl Closurizable for RichTerm {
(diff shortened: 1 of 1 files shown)


yannham:
It is already packed as part of #491, so it shouldn't be necessary. Let us keep just in case #491 doesn't land for some reasons, but otherwise we can close if it does.

Issue #524, opened 2021-12-29
REPL should print the contents of lists and records


mboes:
**Is your feature request related to a problem? Please describe.**
Interacting with the REPL is not as informative as one would expect. Examples:

```
$ nickel repl
repl> [1, 2, 3]
[ ... ]
repl> {x = 1, y = 2}
{ ... }
```

**Describe the solution you'd like**
Compound structures like records and lists should be fully pretty-printed.

yannham:
Said differently, should we have the current output of the `:print` (`:p`) command be the default output of the REPL? Despite the existence of the print command, I agree that the default output is quite useless.

yannham:
@mboes #538 basically uses the `:p` command as the default. However it has the consequence of forcing the evaluation and fully substituting variables inside any term you input in the REPL. Were you aware of the `:p` command? Is it not sufficient and should we still proceed? Ofc ideally forcing stuff doesn't change the semantics but maybe in the future we don't want to do it by default if someone enter something like `import "nixpkgs.ncl" {}` and import a huge, lazy record.

aspiwack:
Ghci uses the same convention (try `let x = 1 : x in x`). So does `nix repl` if I'm not mistaken. I'd say that it's definitely a reasonable choice for now.

We can consider more complex printers later if it turns out that it's a problem.

yannham:
I thought `nix repl` didn't force, so I tried. Actually it does something strange in between, maybe it just forces one level:

```
nix-repl> rec {x = x;}
error: infinite recursion encountered

nix-repl> rec {foo = {bar = foo;};}
{ foo = { ... }; }
```

Even more interestingly, the `:p` command is able to handle this kind of recursion without looping:

```
nix-repl>:p { foo = { ... }; }
{ foo = { bar = «repeated»; }; }
```

Anyway, let's proceed with full evaluation for now and see later indeed.

PR #538, opened 2022-01-04
Use deep_repr when printing evaluated term

Avi-D-coder:
fix: #524

### DIFF
--- a/src/repl/rustyline_frontend.rs
+++ b/src/repl/rustyline_frontend.rs
@@ -103,7 +103,7 @@ pub fn repl(histfile: PathBuf) -> Result<(), InitError> {
             }
             Ok(line) => {
                 match repl.eval(&line) {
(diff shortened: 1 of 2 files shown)


yannham (changes requested):
The output is not the same as for the `:p` command:

```
nickel> {foo = "bar", baz = "boo"}
{ baz = %886, foo = %885}

nickel> :p {foo = "bar", baz = "boo"}
{ baz = "boo", foo = "bar"}
```

I suspect the `:p` does a bit more magic, like getting rid of those automatically generated variables. We should probably do the same, otherwise it's just gonna print those `%XX` everywhere.

Avi-D-coder:
@yannham Print uses `eval_full`, is that an issue?

yannham:
Hmm. This means we are forcing the evaluation as well. Let me ask on the issue.

yannham (approved):

Issue #430, opened 2021-10-26
`import`s are always inferred as `Dyn`


regnat:
(I think this is essentially a more generic version of https://github.com/tweag/nickel/issues/375 − and is hopefully fixed by #392, but since the symptoms aren’t obviously related it’s worth its own issue)

**Describe the bug**

`import "foo.ncl"` is always given the `Dyn` type regardless of what’s in `foo.ncl`

**To Reproduce**

```console
$ cat <<EOF > imported.ncl
{ foo = "bar" } : { foo : Str }
EOF
$ nickel -f imported.ncl typecheck
$ cat <<EOF > main.ncl    
> let Imported = import "imported.ncl" in
> Imported.foo : Str
> EOF                                                                                                        
$ nickel -f main.ncl typecheck
error: Incompatible types
  ┌─ /tmp/tmp.iXrAfpc3uc/main.ncl:2:1
  │
2 │ Imported.foo : Str
  │ ^^^^^^^^ this expression
  │
  = The type of the expression was expected to be `{foo: Str | _a}`
  = The type of the expression was inferred to be `Dyn`
  = These types are not compatible
```

**Expected behavior**

`import "imported.ncl"` should have the same type as the equivalent inlined expression (so `{ foo : Str }` in this case)

**Environment**
 - OS name + version: NixOS 21.05
 - Version of the code: c8d608de9649f955f7c5e121b92ac96f01cfa568

EDIT: To be more precise, imports are always **inferred** as `Dyn`, but can be checked against a more precise type. For example `let Imported : { foo : Str } = import "imported.ncl" in Imported.foo : Str` typechecks just fine

francois-caddet:
Can be closed right?

PR #540, opened 2022-01-04
Get apparent type from imported term instead of defaulting to Dyn

litchipi:
Simple fix for issue #430 
Has the inconvenient of passing the ImportResolver to the `apparent_type` function.
Using `Option<_>` to allow one to simply ignore the apparent typing of the imported terms (notably useful for the `infer_type` function which doesn't have access to any ImportResolver).

Would it be needed to add a regression test for this particular issue ?

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -397,14 +397,19 @@ impl<'a> Envs<'a> {
     /// Add the bindings of a record to a typing environment. Ignore fields whose name are defined
     /// through interpolation.
     //TODO: support the case of a record with a type annotation.
-    pub fn env_add_term(env: &mut Environment, rt: &RichTerm) -> Result<(), EnvBuildError> {
+    pub fn env_add_term(
+        env: &mut Environment,
+        rt: &RichTerm,
+        resolver: &dyn ImportResolver,
+    ) -> Result<(), EnvBuildError> {
         let RichTerm { term, pos } = rt;
 
         match term.as_ref() {
             Term::Record(bindings, _) | Term::RecRecord(bindings, ..) => {
                 for (id, t) in bindings {
                     let tyw: TypeWrapper =
-                        apparent_type(t.as_ref(), Some(&Envs::from_global(env))).into();
+                        apparent_type(t.as_ref(), Some(&Envs::from_global(env)), Some(resolver))
+                            .into();
                     env.insert(id.clone(), tyw);
                 }
 
@@ -415,10 +420,10 @@ impl<'a> Envs<'a> {
     }
 
     /// Bind one term in a typing environment.
-    pub fn env_add(env: &mut Environment, id: Ident, rt: &RichTerm) {
+    pub fn env_add(env: &mut Environment, id: Ident, rt: &RichTerm, resolver: &dyn ImportResolver) {
         env.insert(
             id,
-            apparent_type(rt.as_ref(), Some(&Envs::from_global(env))).into(),
+            apparent_type(rt.as_ref(), Some(&Envs::from_global(env)), Some(resolver)).into(),
         );
     }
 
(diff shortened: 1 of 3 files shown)


yannham (commented):
LGTM but updating the documentation of updated functions (`apparent_type`, `binding_type`) and a test case would be nice.

tests/pass/import.ncl
```diff
@@ -1,3 +1,5 @@
 let Assert = fun l x => x || %blame% l in
 
-(import "imported.ncl" 3 == 3 | #Assert)
+let Imported = import "imported.ncl" in
+(Imported.f 3 == 3 | #Assert) &&
+(builtins.is_str Imported.foo | #Assert)
```
yannham, line 5: This doesn't test typing at all, as there is not type annotation. What you want to do is:

- Revert this change. I think the original test was just testing the applicative syntax for imports, and it was fine. In general it's better to add tests that to alter previous one, because unless we 100% know what we are doing, we may change what they were actually testing without knowing it.
- Add a line with a let-binding for an import and type it. Basically just after line 169, add a variant:
  ```
  // Regression test for #430 (https://github.com/tweag/nickel/issues/430)
  let x = import "typed-import.ncl"
  in x : Num,
  ```

yannham (changes requested):

litchipi:
The new regression test works as intended and doesn't pollute much the existing one, however I still don't really find it elegant but I'm not really confortable with the Nickel langage yet.
Is there a way to use `deepseq` to force the typechecking and use it elegantly with the existing term ?
Or should I let it ugly and functionnal ?

yannham:
You can put the test in `typechecking.ncl`, which is specifically for typechecking. It doesn't evaluate anything (because lists are lazy in Nickel, and this test file is just one big list), so it just typechecks.

You don't need to change the `typed-import.ncl`. You can leave it as it was, with the type annotation: in both case it has the same apparent type (`Num`), and the apparent type of the import itself doesn't even matter. What matters is that if you put this import in a variable without annotation, this variable is assigned the same type `Num`, while it was assigned `Dyn` before. So, to sum up:

1. Leave `typed-import.ncl` alone
2. Inside `tests/pass/typechecking.ncl`, add a new element at the end of the list (with the comment about it being a regression test): `let x = import "typed-import.ncl" in x : Num,`

litchipi:
Fair enough, didn't see how to add a typechecking-only test into the suite, I'll know later.
Added the test as you intended, should be okay now

src/typecheck/mod.rs
```diff
@@ -1066,6 +1091,16 @@ pub fn apparent_type(t: &Term, envs: Option<&Envs>) -> ApparentType {
             .and_then(|envs| envs.get(id))
             .map(ApparentType::FromEnv)
             .unwrap_or(ApparentType::Approximated(Types(AbsType::Dyn()))),
+        Term::ResolvedImport(f) => {
+            if let Some(r) = resolver {
+                let t = r.get(*f).expect(
+                    "Internal error: resolved import not found ({:?}) during typechecking.",
```
yannham, line 1097: `expect` takes a `&str` (and you haven't passed any parameter). So this will always print `{:?}` literally in the message. You either need to use `format!` or decide not to care about showing the specific import and drop the interpolation altogether.

yannham (approved):
LGTM but the bit on `expect`. Good to go once this is fixed.

src/typecheck/mod.rs
litchipi: Oops indeed, I just copied the line 970 (bottom of the `type_check_` function) which has exactly the same problem apparently, gonna push a fix for that too right after

yannham:
@litchipi there's a conflict now and the code hasn't been formatted properly.

Issue #523, opened 2021-12-29
Pretty print values in batch mode


mboes:
**Is your feature request related to a problem? Please describe.**
Feeding an expression to the `nickel` binary yields a value if evaluation succeeds. However, this value is not valid Nickel syntax, and sometimes is outright confusing. Examples:

```
$ nickel -f examples/fibonacci/fibonacci.ncl
Done: Num(55.0)

$ nickel -f examples/record-contract/record-contract.ncl
Done: Record(... some really big output that leaks implementation details ...)
```

**Describe the solution you'd like**
Instead of using Rust's debug output, we should output pretty-printed Nickel values. If `nickel` succeeds in evaluating an expression, then calling `nickel` on the output should yield the same output.

PR #541, opened 2022-01-04
Fix: #523

Avi-D-coder:
Fix: #523
This uses `eval_full` and `deep_repr` like print when evaluating files.
`deep_repr` output for contracts is far from ideal. This should probably be a separate issue.
 
```
❯ cargo run -- -f examples/record-contract/record-contract.ncl
{ metadata = { labels = { app = <contract,value="myApp">}, name = <contract,value="myApp">}, spec = { replicas = <doc,contract,value=3>, selector = { app = { name = "myApp"}, matchLabels = { app = <contract,value="myApp">}}, template = { metadata = { labels = { app = <contract,value="myApp">}, name = "myApp"}, spec = { containers = <contract,value=[ ... ]>}}}, apiVersion = <contract,value="1.1.0">, kind = <doc,contract,value=`ReplicationController>}
```

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -4,7 +4,7 @@ use nickel::program::Program;
 use nickel::repl::query_print;
 #[cfg(feature = "repl")]
 use nickel::repl::rustyline_frontend;
-use nickel::term::RichTerm;
+use nickel::term::{RichTerm, Term};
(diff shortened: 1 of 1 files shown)


yannham (approved):
We probably don't want to close #523 right away, as the ideal solution would be an actual pretty-printer, but this is a tiny change that does improve on the previous situation, so let's merge this.

Issue #532, opened 2022-01-01
Consider performing CI using Rust stable channel only


mboes:
We currently run builds and tests at least 6 times at every push, for each of the 3 Rust channels (stable, beta, nightly) and for two operating systems (ubuntu, macOS). The runtime of workflow jobs have high variation, and CI is only as fast as the slowest job, not to mention eating into our organization-wide [concurrency budget](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration). The latter isn't really a problem except that macOS slots are scarce.

New Rust stable releases are very frequent (every 6 weeks), and breakage rare (unlike in beta or nightly channels). AFAIK we don't specifically depend on compiler features only available in beta, so like the [Rust book](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) says, testing against beta is therefore mainly a service to the upstream Rust developers (if an issue arises in beta but not in stable then it's likely a bug in the beta release). I don't know why we furthermore test against nightly. Sounds overkill. That nightly breaks our code whenever it does is entirely expected and not terribly interesting to anyone (either us or upstream).

There is another cost to this: to retain reproducible builds, we automate tracking the head of the respective toolchains. So every day, we have at least one new commit by a bot polluting our Git history. In fact over the past 5 months, 16% of commits have been of this kind. The proportion vastly increases when development slows down. That is then more toolchains we have to push to Cachix every night.

I propose we track only `stable`. Less wasteful. We can keep the channel update workflow for stable. That only creates a bot commit every 6 weeks. I don't know that Nickel is a terribly interesting test case for Rust beta, so I propose we drop that as well as nightly. 

Benefits:
* less compute power required for our CI needs,
* 1-2 minutes shaved off CI on average.

yannham:
Fully agreed. It was proposed a while ago, and the answer was "it's doesn't hurt", but actually it isn't free either, indeed.

PR #542, opened 2022-01-04
Fix: #532, only test on rust stable channel

Avi-D-coder:


.github/workflows/continuous-integration.yml
```diff
@@ -17,16 +17,14 @@ jobs:
...
-          - nightly
         include:
           - os: ubuntu-latest
             system: x86_64-linux
             experimental: false
           - os: macos-latest
             system: x86_64-darwin
             experimental: false
-          - rust_channel: nightly
+          - rust_channel: stable
```
yannham, line 27: This is the qualifier for `experimental: true`, which specifies that the build can fail on nightly but that shouldn't abort the whole CI, because nightly is more subject to break. We don't want `stable` to be experimental, so we probably rather want to delete this line and the next one. We may keep the other references to `experimental` even if this is now useless for now I guess, as those may still serve if we want to add experimental tests at some point.

yannham (changes requested):

.github/workflows/continuous-integration.yml
```diff
@@ -17,17 +17,15 @@ jobs:
...
         include:
           - os: ubuntu-latest
             system: x86_64-linux
             experimental: false
           - os: macos-latest
             system: x86_64-darwin
             experimental: false
-          - rust_channel: nightly
-            experimental: true
+          - rust_channel: stable
```
yannham, line 27: @Avi-D-coder You can just delete them. The previous lines set `experimental` to `false` by default already.

yannham: @Avi-D-coder I'm sorry, "them" may have been unclear. What I meant is that you should delete the original lines:

```
- rust_channel: nightly
  experimenta: true
```

but I think we have to keep the other `experimental: false`

Avi-D-coder: @yannham  Updated.
It should work both ways since 
```yaml
    # The `== true` makes it work wether experimental has been defined or not.
    continue-on-error: ${{ matrix.experimental == true }}
```

yannham: Oh, you're totally right. Sorry for the noise then, either way is fine.

.github/workflows/continuous-integration.yml
```diff
@@ -17,17 +17,14 @@ jobs:
...
```
yannham, line 27:
```suggestion
```

yannham (approved):
The `rust_channel` line can also go, I think. Then it's good to go.

PR #544, opened 2022-01-04
Fix records benchmark

Avi-D-coder:


Avi-D-coder:
@yannham There we go, this is the correct benchmarks error again.
It should be good to merge.

yannham (approved):

Issue #522, opened 2021-12-29
[Tests] Examples are not fully tested


yannham:
**Describe the bug**
Currently, the test suite just tries to evaluate the examples. Nickel being lazy, this doesn't unveil errors that could be buried inside e.g. records or lists. So an example could pass the CI while fail when run with `nickel export`.

**To Reproduce**
Go to the `record-contract` example, and change `apiVersion = "1.1.0"` to `apiVersion = "1.1.0" + 1`. The tests are still passing.

**Expected behavior**
Examples should probably be forced for tests.

Avi-D-coder:
This is sort of related to https://github.com/tweag/nickel/issues/524, and https://github.com/tweag/nickel/issues/523.
With https://github.com/tweag/nickel/pull/541 we get the diagnostic bellow. If we don't want to fully evaluate in batch mode for the same reasons we may not want to in the repl https://github.com/tweag/nickel/issues/524#issuecomment-1005033127, is there any other way we could surface this error without making the user force evaluation? Having to force evaluation to get a type error like this is quite non intuitive even though if it makes sense in the context of lazy contracts.

```
error: Type error
   ┌─ /home/USER/Work/nickel/examples/record-contract/record-contract.ncl:58:16
   │
58 │   apiVersion = "1.1.0" + 1,
   │                ^^^^^^^
   │                │
   │                This expression has type Str, but Num was expected
   │                evaluated to this
   │
   = +, 1st argument
```

yannham:
This issue is a bit different, as the requested changes won't affect the user at all. It's just for the testsuite. Here it is totally fine to force stuff if we have to.

PR #545, opened 2022-01-05
Fix: #522, Use eval_full to test examples

Avi-D-coder:
Fix: #522

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -668,6 +668,29 @@ pub fn eval_full<R>(
 where
     R: ImportResolver,
 {
+    eval_internal(t0, global_env, resolver).map(|(term, env)| subst(term, global_env, &env))
+}
+
+/// Fully evaluates a Nickel term like `eval_full`, but does not substitute all variables.
+pub fn eval_deep<R>(
+    t0: RichTerm,
+    global_env: &Environment,
+    resolver: &mut R,
+) -> Result<RichTerm, EvalError>
+where
+    R: ImportResolver,
+{
+    eval_internal(t0, global_env, resolver).map(|(term, _)| term)
+}
+
+fn eval_internal<R>(
+    t0: RichTerm,
+    global_env: &Environment,
+    resolver: &mut R,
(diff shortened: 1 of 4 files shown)


tests/examples.rs
```diff
@@ -8,7 +8,7 @@ fn eval_file(file: &str) -> Result<RichTerm, Error> {
     let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     path.push(format!("examples/{}", file));
     let mut p = Program::new_from_file(path).expect("could not load file as a program");
-    p.eval()
+    p.eval_full()
```
yannham, line 11: `eval_full()` does more than we want, as it will also substitute all the variables inside the resulting term. It's maybe not immensely important for now given the size and number of examples, but it's probably better to save potential future compute power. What we want is basically to do everything that's done in `eval_full` but without the last step. It may make sense to factorize one part of `eval_full` as `eval_deep`, if that avoids duplicating code.

Avi-D-coder: Done

src/eval.rs
```diff
@@ -665,6 +665,29 @@ pub fn eval_full<R>(
...
+    global_env: &Environment,
+    resolver: &mut R,
+) -> Result<RichTerm, EvalError>
+where
+    R: ImportResolver,
+{
+    eval_internal(t0, global_env, resolver).map(|(term, _)| term)
+}
+
+fn eval_internal<R>(
```
yannham, line 686: Nitpick: `internal` is not very suggestive. Maybe `eval_deep_` or `eval_deep_closure` or something.

src/program.rs
```diff
@@ -89,6 +89,13 @@ impl Program {
         eval::eval_full(t, &global_env, &mut self.cache).map_err(|e| e.into())
     }
 
+    /// Same as `eval`, but proceeds to a full evaluation.
```
yannham, line 92:
```suggestion
```

yannham (approved):

PR #546, opened 2022-01-06
Fix expect 'import not found' message in type_check_

litchipi:
As pointed out in [a comment of #540](https://github.com/tweag/nickel/pull/540#discussion_r779407722), the message displayed by `expect` will not be formatted properly with the `{:?}` element.
We could pass it through a `format!` or simply drop it.

As the element to display is a simple file number (doesn't give us much informations on the bug itself), I propose to simply drop it.


yannham (approved):
Yes, I guess we can add more information if we need to debug something like this at some point.

Issue #548, opened 2022-01-06
MetaValue flatten function doesn't merge types


litchipi:
**Describe the bug**
`MetaValue::flatten` function doesn't merge the `types` attributes of the inner into the outer if outer's types are `None`

**To Reproduce**
```rust
let mut inner = MetaValue::new();
inner.types = Some(Contract { types: Types(AbsType::Num()), label: Label::dummy() });
let outer = MetaValue::new();
let res = MetaValue::flatten(outer, inner);
// res.types == None here
```

**Expected behavior**
`outer` should have included the types from `inner` in its own types

**Environment**
 - Ubuntu 21.10
 - On master commit `13fb3edadcc2bf7db9e09d32df6ba2f984b7a555`

**Additional context**
Encountered while finishing resolution of issue #496

yannham:
Fixed #549.

PR #549, opened 2022-01-06
Fix types merging in Metavalue::flatten function

litchipi:
Solving issue #548

Would like a review to ensure the behaviour when fixed is the expected one.


yannham (approved):
LGTM

PR #553, opened 2022-01-06
Fix/explicit destructuring typechecking

francois-caddet:
Update comments to remember where we are missing typecheck behaviours.
also fix typechecking of RHS in not typed context.

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -652,7 +652,7 @@ fn type_check_<S, E>(
         }
         Term::LetPattern(x, pat, re, rt) => {
             use crate::destruct::*;
-            // TODO: The insertion of values in the type environement is done but everything is
+            // TODO: The insertion of values in the type environment is done but everything is
             // typed as `Dyn`.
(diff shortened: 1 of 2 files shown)


src/typecheck/mod.rs
```diff
@@ -650,8 +650,10 @@ fn type_check_<S, E>(
             envs.insert(x.clone(), ty_let);
             type_check_(state, envs, lin, linearizer, strict, rt, ty)
         }
-        Term::LetPattern(x, pat, re, _rt) => {
+        Term::LetPattern(x, pat, re, rt) => {
             use crate::destruct::*;
+            // TODO: The insertion of values in the type environement is done but everything is
```
yannham, line 655:
```suggestion
            // TODO: The insertion of values in the type environment is done but everything is
```

yannham (approved):
Does this change make the LSP now work fine with destructuring?

francois-caddet:
> Does this change make the LSP now work fine with destructuring?

Yes It should at least for the symbols.
I didn't tried all cases but I think it is ok.

PR #554, opened 2022-01-07
[Refactor] Break typecheck into submodules

yannham:
In the same vein as #547 and #551, break `typecheck/mod.rs` into more files:

- move already existing submodules into their own file (`reporting`, `mk_typewrapper`)
- move error definitions and related code into a new `error` submodule
- move `get_XXop_type`, returning the type of primitive operations, into a new `operation` submodule
- Fixed a few typos/unclear comments along the way. Minor changes.

Went from ~2,4kLoC to 1,4kLoC for `mod.rs`.


aspiwack:
This doesn't look particularly urgent. Is it blocking something?

yannham:
Not particularly, I just took the occasion of having the lowest count of open PRs for a while to do this kind of cleaning, that as moving things around in files can make rebasing painful.

aspiwack:
I see.

That being said, I think that now is the time to focus on the upcoming release, and prioritise this sort of work over all other (as we always find that there is more work than anticipated).

litchipi (approved):
Looks good to me, could maybe even isolate `type_check` and `unify` as they are already big and may grow even more, and occasions of breaking changes may not appear again that soon.
However it's still nice like that, OK for me

yannham:
For now I lean on the side of keeping all the actual typechecking-related logic inside the core module. We could split it further later on if that's becoming a problem.

PR #559, opened 2022-01-10
Add --nostdlib option in debug releases to skip stdlib loading

litchipi:
When debugging a problem, the debugging messages often gets unreadable because of the loading of the standard library.

As usually we debug against a minimal example containing the bug and want to trace what happends inside the software, we usually do not need the standard library (and could copy manually parts of it if needed).

So I propose to add an option `nostdlib` in **debug releases only** that would skip the loading of the standard library to allow this to be easily done.

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -19,6 +19,12 @@ struct Opt {
     #[structopt(short = "f", long)]
     #[structopt(parse(from_os_str))]
     file: Option<PathBuf>,
+
+    #[cfg(debug_assertions)]
+    /// Skip the standard library import, for debugging only, does not affect REPL
+    #[structopt(long)]
+    nostdlib: bool,
+
     #[structopt(subcommand)]
     command: Option<Command>,
 }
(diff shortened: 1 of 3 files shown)


src/bin/nickel.rs
```diff
@@ -19,6 +19,12 @@ struct Opt {
     #[structopt(short = "f", long)]
     #[structopt(parse(from_os_str))]
     file: Option<PathBuf>,
+
+    #[cfg(debug_assertions)]
+    /// Do not include the standard library in the scope
```
yannham, line 24:
```suggestion
    /// Do not include the standard library in the scope (for debugging purposes only)
```

yannham (commented):
Can be useful indeed. However it doesn't work for the REPL right now, but just for the default command. If it is intended, the documentation of the flag should at least informs about this.

src/bin/nickel.rs
litchipi: I tried to make an explicit, complete and short message like so:
```
Skip the standard library import, for debugging only, does not affect REPL
```

yannham (approved):

PR #561, opened 2022-01-10
Convert TermPos to Raw spans throughout nls

ysndr:
`TermPos` is guaranteed to be not-none after linearization is done
We can use `RawSpan` from that point on instead ofunwraping the `TermPos` everywhere.

Alternative:
Keep the `TermPos` but unwrap whenever used
This could retain semantics of the different cases of `TermPos`

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -68,7 +68,7 @@ impl Building {
             let id = self.id_gen().get_and_advance();
             self.push(LinearizationItem {
                 id,
-                pos: ident.pos,
+                pos: ident.pos.unwrap(),
                 // temporary, the actual type is resolved later and the item retyped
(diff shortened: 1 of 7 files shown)


yannham:
>TermPos is guaranteed to be not-none after linearization is done

Why is it the case? Do you mean that this happens to be currently the case, or is there a technical reasons like you ignore term without position or something?

ysndr:
Yes all items without position are either ignored or receive a new position (I.e. Record fields)
There was a step in the post-processing in which all None-positions panic so it was already expected that those won't exist afterwards

yannham (approved):
Ok. Then it could be a good addition to write explicitly this invariant at the point it has to be enforced, if this is not already the case. Or as documentation of the Linearization. Somewhere it makes to most sense to avoid accidentally breaking it.

ysndr:
Ill update the documentation on top of #558

PR #562, opened 2022-01-10
Disable record benchmarks

Avi-D-coder:
These are two slow, and are one of the things preventing the benchmarks GitHub action from succeeding.


yannham (approved):
We should maybe remember the fact that we need to investigate this and fix it someday, such as in an issue.

Issue #52, opened 2019-11-20
Encoding Sums with Π and Σ


teofr:
_**Current status**: Intersection and Union types ended up being much more complex than anticipated, on standby for now. We are looking for a simpler way to encode Sum types, this issue presents a possible solution._

The idea would be to encode sums by using Pi and Sigma types: `A+B` would become `Sigma b: Bool, if b then A else B`, which could be constructed with `Pi b: Bool, (if b then A else B) -> Sigma c: Bool, if c then A else B`.

For this, we'd need to implement two new kind of contracts:
 * Dependent functions, `x: A -> \x. B`, which allow the codomain contract to depend on the input, as a reference we have [1] Racket's functions contracts, [2] is also a good reference. There are two challenges/decisions:
   * _strictness_, a function that ignores it's input may evaluate it anyways if the target contract does
   * _lax vs picky_, as pointed in [3], we can decide to wrap or not the argument on the domain contract when passed to the codomain, I think it should be checked there as well.
 * Dependent records, `{idx: Bool, val: if idx then A else B}` would allow us to have Sigma types quite naturally, together with some more complex stuff.

Ideas/problems around this:
 * We may be good enough with a Pi indexed only by simple types (Bool, Num, enums), I think this would simplify some of the concerns regarding strictness. Dually, some similar restriction might be helpful for Sigma. If we also drop Num, type checking might become much easier.
 * Maybe we can get rid of Sigma, and construct them with `Pi b: Bool, (if b then A else B) -> (Pi c: Bool, if c then Bool else if b then A else B)`, where `c` is the accessor for the pair (`true == fst`, `false == snd`).
 * A big question is what to do with type checking  for these types.
 * [4] talks about some problems with dependent function contracts, I'll try to understand in more detail what they're saying and check whether it makes any difference for us.

cc @aspiwack 

[1] [Racket's contracts, in particular `->i`](https://docs.racket-lang.org/reference/function-contracts.html#%28form._%28%28lib._racket%2Fcontract%2Fbase..rkt%29._-~3ei%29%29)
[2] [On Contract Satisfaction in a Higher-Order World, Dimoulas & Felleisen](http://www.cs.umd.edu/class/spring2014/cmsc631/papers/dimoulas-contract-sat.pdf)
[3] [Contracts Made Manifest, Greenberg et. al.](https://www.cis.upenn.edu/~bcpierce/papers/contracts-popl.pdf)
[...]

PR #568, opened 2022-01-12
Save original notes on dependent records

yannham:
Save the notes of #52, such that we can close the long-lived and inactive issue without loosing the thoughts.

### DIFF
--- a/notes/sum-as-dependent-records.md
+++ b/notes/sum-as-dependent-records.md
@@ -9,7 +9,7 @@ The idea would be to encode sums by using Pi and Sigma types: `A+B` would become
 
 For this, we'd need to implement two new kind of contracts:
  * Dependent functions, `x: A -> \x. B`, which allow the codomain contract to depend on the input, as a reference we have [1] Racket's functions contracts, [2] is also a good reference. There are two challenges/decisions:
-   * _strictness_, a function that ignores it input may evaluate it anyways if the target contract does
+   * _strictness_, a function that ignores it input may evaluate it anyway if the target contract does
    * _lax vs picky_, as pointed in [3], we can decide to wrap or not the argument on the domain contract when passed to the codomain, I think it should be checked there as well.
  * Dependent records, `{idx: Bool, val: if idx then A else B}` would allow us to have Sigma types quite naturally, together with some more complex stuff.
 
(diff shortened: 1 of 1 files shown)


notes/sum-as-dependent-records.md
```diff
@@ -0,0 +1,43 @@
+# Notes on sum encoding with dependent types
+
+**This is an archive of thoughts originally written in issue [#54](https://github.com/tweag/nickel/issues/52)**
+
+
+_**Current status**: Intersection and Union types ended up being much more complex than anticipated, on standby for now. We are looking for a simpler way to encode Sum types, this issue presents a possible solution._
+
+The idea would be to encode sums by using Pi and Sigma types: `A+B` would become `Sigma b: Bool, if b then A else B`, which could be constructed with `Pi b: Bool, (if b then A else B) -> Sigma c: Bool, if c then A else B`.
+
+For this, we'd need to implement two new kind of contracts:
```
Acaccia, line 10: Should we allow contractions in documents ?

notes/sum-as-dependent-records.md
```diff
@@ -0,0 +1,43 @@
...
+ * Dependent functions, `x: A -> \x. B`, which allow the codomain contract to depend on the input, as a reference we have [1] Racket's functions contracts, [2] is also a good reference. There are two challenges/decisions:
+   * _strictness_, a function that ignores it's input may evaluate it anyways if the target contract does
```
Acaccia, line 12: its

Acaccia: anyways

Acaccia (approved):
There are several contractions in the text, should we allow them for these kind of notes ?

notes/sum-as-dependent-records.md
yannham, line 12:
```suggestion
   * _strictness_, a function that ignores it input may evaluate it anyways if the target contract does
```

yannham:
```suggestion
   * _strictness_, a function that ignores it's input may evaluate it anyway if the target contract does
```

aspiwack:
> There are several contractions in the text, should we allow them for these kind of notes ?

Contractions are always ok in English text. It's really not something that should should worry about.

Issue #207, opened 2020-11-13
The Syntax Bikeshedding Dojo, round 4: Basics


yannham:
@garbas asked me if the function syntax was final. Since nothing is final yet, I think it is a good idea to create a dedicated issue for the basic constructs of the language, even if most seem non controversial, such that people can voice concerns or make alternative proposals.

1. **Functions**
2. **Records**
3. **Comments**
4. **Others**

### Definition
Currently, Nickel uses an ML-like anonymous function syntax `fun arg1 .. argn => body`. There are other possibilities:
- *Nix-like*: `arg1: arg2: ..: body`, but `:` will be used for typing, so it's probably a bad idea (see section below).
- *C-like*: `fun(arg1,..,argn) ..`. [ReasonML](https://reasonml.github.io/docs/en/function) does this for example, although they also adopt the corresponding call syntax `f(arg1,..,argn)` which looks less natural in languages with currying. This can be confused with tuples, but Nickel does not have tuples currently.
- *C-like body*: `.. { body }`. Independently from the syntax of arguments, the body could be written as a block.
- *variation*: use a different keyword than `fun`: `fn`, `def`, etc.

We could also define a non-anonymous function in a more concise way. For example, in OCaml, you can do:
```
let f arg1 .. argn = body in
//sugar for
let f = fun arg1 ... argn => body in
```

### Typing
It has not been discussed, but a nice thing to have is a syntax to define the arguments, their type and the return type directly. In ML-derived languages, you can usually write:
```
let f (x: Num) (y: Num) (z: Num) : Num = body
//instead of
let f : Num -> Num -> Num -> Num = fun x y z => ...
```

As in the previous section, there are many variations and possible choices:
- Rust-y return type syntax: `let f (x: Num) (y: Num) (z: Num) -> Num = body`
- and others depending on the choice of the function definition syntax

One Nickel-specific issue would be to have a similar syntax for function contracts. For example, we can replace `: Num` with `| type Num` or `| contract Num` (see #183):
```
let f (x | type Num) (y | type Num) (z | type Num) | type Num = body
```
to indicate that pre and post-conditions must be checked at run-time, but that we don't wish to statically typecheck the body.

Note that there's the issue #81 for destructuring, which would also be available directly on function arguments.
## Records and Lists
Nickel uses Nix-like record definition:
```
{
  field = content;
  ..
  field = content;
}
```
[...]

PR #569, opened 2022-01-12
let function syntax

francois-caddet:
Extend the language to accept ML like functions declaration:
```text
let f a b = a + b in f 3 5
```
discuced in #207 
Can accept patterns params like in #81 but with this new syntax. Only the first element after the let has to be an identifier, because it's the name of the function but the others can be patterns.
@yannham  is there an other issue to link?

### DIFF
--- a/lsp/nls/src/linearization/completed.rs
+++ b/lsp/nls/src/linearization/completed.rs
@@ -109,7 +109,7 @@ impl Completed {
 
         let item = match item.kind {
             TermKind::Usage(UsageState::Resolved(usage)) => {
-                usage.and_then(|u| self.get_item(u + 1)).unwrap_or(item)
+                usage.and_then(|u| self.get_item(u)).unwrap_or(item)
             }
             TermKind::Declaration(_, _) => self.get_item(item.id + 1).unwrap_or(item),
             _ => item,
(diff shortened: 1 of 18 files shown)


yannham:
Oh, and we also want those when defining record fields, like `{ foo : Num -> Num, foo x y = x + y}`

mboes:
I think we want to punt on this to post release and until we have more code written in Nickel, for the same reason as discussed in https://github.com/tweag/nickel/issues/494#issuecomment-1004781660.

yannham:
Even in the light of https://github.com/tweag/nickel/issues/494#issuecomment-1002950315 ?

francois-caddet:
I don't know well this syntax but will see. If it's not a lot to implement I will do it for this PR.

mboes:
Sorry, what I meant in that comment is that this clearly needs a solution eventually. But since this is only syntactic sugar, would be good to gain experience first. If there was only one possible design here that would be one thing. But multiple designs are reasonable, including changing the syntax for functions everywhere, so that no syntactic sugar for lets is necessary. Further, it feels to me like there are higher priorities ahead of the release.

francois-caddet:
So we'll see later.

PR #570, opened 2022-01-12
[language server] Fix resolving definition for meta

ysndr:
Fixes an issue where metadata for usages was not correctly resolved


yannham (approved):
Seems to fix the issue for me indeed.

PR #572, opened 2022-01-13
[language server] Fix off-by-one issue in resolution of declarations

ysndr:
Function arguments and let bindings should show the correct type when hovered

### DIFF
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -139,8 +139,40 @@ impl Linearizer for AnalysisHost {
 
         let id = id_gen.get();
         match term {
+            Term::FunPattern(ident, destruct, _) => {
+                if let Some(ident) = ident {
+                    self.env.insert(ident.to_owned(), id);
+                    lin.push(LinearizationItem {
+                        id: id_gen.get_and_advance(),
+                        ty,
+                        pos: ident.pos.unwrap(),
+                        scope: self.scope.clone(),
+                        kind: TermKind::Declaration(ident.to_owned(), Vec::new()),
+                        meta: self.meta.take(),
+                    });
+                }
+                for matched in destruct.to_owned().inner() {
+                    let (ident, term) = matched.as_meta_field();
+                    self.env.insert(ident.to_owned(), id_gen.get());
+                    lin.push(LinearizationItem {
+                        id: id_gen.get_and_advance(),
+                        // TODO: get type from pattern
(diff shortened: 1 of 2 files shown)


ysndr:
That's odd 
Parameter types definitely worked for me
Can you add an example?
Either way I'll make sure to look for your issues

yannham:
Ok, this has nothing to do with this PR it seems, but with #567, after more tests. Don't why it works on your machine, as I checked this branch specifically, but after checking the problem is also on master but not before the merge commit of #567. I can't judge for the correctness of this specific fix, so if you tested it and it seems logically sound, go ahead and merge. Then we'll take a look at what broke with function argument destructuring.

yannham (approved):

ysndr:
I implemented lineaerization for fun patterns which should catch AEG names and at least simple record fields
Record fields are typed as Dyn as I don't know where to get their type from right now

yannham:
Yes, the type of the record fields being `Dyn` is a current issue with destructuring, so this needs to be fixed first.Will take a look ASAP

lsp/nls/src/linearization/mod.rs
```diff
@@ -139,8 +139,40 @@ impl Linearizer for AnalysisHost {
...
+            Term::FunPattern(ident, destruct, _) => {
+                if let Some(ident) = ident {
+                    self.env.insert(ident.to_owned(), id);
+                    lin.push(LinearizationItem {
+                        id: id_gen.get_and_advance(),
+                        ty,
+                        pos: ident.pos.unwrap(),
+                        scope: self.scope.clone(),
+                        kind: TermKind::Declaration(ident.to_owned(), Vec::new()),
+                        meta: self.meta.take(),
```
yannham, line 151: Do we want to use the metavalue of the function for the parameter? I think this is causing the last strange issue I have: when I hover on the first parameter of a function, it shows the documentation of the whole function as metadata.

yannham (approved):
Now the parameter hovering seems fix. There's just one quirk thing remaining, which is that when I hover on the first parameter of a multi parameter function, I get the right type, but I get the metavalue (documentation in particular) associated to the function istelf, which feels odd. See the associated comment.

yannham:
@ysndr should we merge this? I just wanted to make sure you saw the comment above about the metadata of the first parameter of a function.

ysndr:
The metadata thing should be handled differently, sure.
I’m ok with merging this now and make another or looking at the metadata

PR #573, opened 2022-01-13
[RFC] Merge types and terms syntax proposal

yannham:


rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,287 @@
...
+
+Which looks more natural, and can be given a reasonable semantics.
+
+## Challenges
+
+Such a change brings in two main challenges:
+
+### Semantics
+
+The first challenge is to give a semantics for all the strange term we can now
```
thufschmitt, line 101:
```suggestion
The first challenge is to give a semantics for all the strange terms we can now
```

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,287 @@
...
+write:
+
+```nickel
+let foo : forall a. a -> (fun x => x + 1) -> a + 1
+```
+
+In order to tame 1., we'll try to rely on the fact that the current language
```
thufschmitt, line 108:
```suggestion
In order to tame it, we'll try to rely on the fact that the current language
```

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,287 @@
...
+`!: Term + Type -> UniTerm` and `~` is some flavour of operational equivalence
+(we won't prove those rules formally, as this a mere design guideline, so we
+also let the definitions be rough). It basically just erases `#` and `§`.
+
+1. For any `t: Term`, `term(!t) ~ t` and for any `T: Type`, `type(!T) = T` .
+   That is, a term or a type in the current syntax with translation operator
+   removed will be interpreted the same in the new syntax.
+2. For any `T: Type`, `term(!T) ~ §T`. That is, the interpretation as a term of
+   something that is a type in the current syntax must be the corresponding
+   contract.
```
thufschmitt, line 184: Don’t we also want:

3. For any `t: Term`, `type(!t) = #t` ?

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,287 @@
...
+  for lists `term [t1, ..., tn] = [ term(t1), ..., term(tn)]`).
+- For type-only constructs (ground types, `->`, and `{_: T}`), it is
+  `term(T) = §(type(T))`, that is the contract associated to the type.
+
+#### Records
+
+It remains to treat the case of records, that can mix previously type and term
+constructs.
+
+- a row declaration `foo : Num` is translated as it is. This is actually already
```
thufschmitt, line 202: What about `foo : Num = 1` ?

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,287 @@
...
+  `type(fun x => body)` = `#term(fun x => body)`.
+
+#### Records
+
+As for terms, the case of records remain. For records that are actually the
+embedding of a `Type`, such as `{foo : Num, bar: {baz: Num}}`, the translation
+is obvious. For records that are the embedding of a `Term`, we can simply
+proceed as above: `type({foo = 1, bar = "baz"}) = #{foo = 1, bar = "baz"}`.
+
+The case for records that mixes the both `Term` and `Type` is less clear. Take
```
thufschmitt, line 242: Would it make sense to consider `foo: Num = 1` as equivalent to `foo = (1: Num)` and translate it as such? (Maybe that’s equivalent to your first proposal, but I’m not 100% sure)

thufschmitt (commented):
Just a few comments after a quick read.

But great stuff overall :+1:

yannham:
Arg, I did quite a bit of rework and only saw your comments. I'll try to see if some of your questions still need answers

rfcs/002-merge-types-terms-syntax.md
yannham: Yes, I think it's equivalent.At any rate `{foo : Num = 1}` is actually desugared to `{foo = (1 : Num)}` by the parser, and this proposals says than then this is a term and the corresponding type will be `#{foo = (1 : Num)}`.

rfcs/002-merge-types-terms-syntax.md
yannham: I think this part has now more explanations about this case. Let me know if this is the case.

yannham: Sorry, actually here we are inside the translation of terms. So `{foo : Num = 1}` is already a totally valid term, no issue.

Profpatsch:
This looks (on a first glance) amazing if it works. One thing I’ve always hated about learning Haskell is how the same syntax in a different context can make it a totally different thing semantically, and sometimes also the other way around.

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,454 @@
...
+{fail : forall a. a -> a}
+```
+
+But is not very worrisome: it's clear that the local definition is the one that
+matters.
+
+To translate a variable, we thus have to know if a variable is a type variable
+or not. We sidestep this question as an external implementation detail, assuming
+here that there is a predicate `is_type_var` available during the translation.
+We can no set:
```
yannham, line 356:
```suggestion
We can now set:
```

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+Let's now define the `term` and `type` translations. Our design principle is to
+try to have them just be simple syntactic translations that insert the implicit
+`#` and `§` appropriately, without having to recurse into terms or do anything
+complex, whenever possible.
+
+The two functions must satisfy the following coherence laws with `! : Term +
+Type -> UniTerm` the embedding of the current syntax to the new unified syntax
+(erasing `#` and `§`, and changing `|` for tails to `;`) and `~` some flavour of
+operational equivalence (we won't prove those rules formally, as this a mere
+design guideline, so we also restrain from giving a formal definition for `~`).
```
aspiwack, line 232: I didn't understand the part of the paragraph starting at “`~` some flavour of”.

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+
+1. For all `t: Term`, `term(!t) ~ t` and for all `T: Type`, `type(!T) = T` .
+   That is, a term or a type in the current syntax with `#` removed will be
+   interpreted the same in the new syntax.
+2. For all `T: Type`, `term(!T) ~ §T`. That is, the interpretation of something
+   that is a type in the current syntax as a term in the new syntax must behave
+   the same as the corresponding contract.
```
aspiwack, line 239: Isn't there an equivalent law for `#`?

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+
+The case for records that mixes the both `Term` and `Type` is less clear. Take
+for example: `x : {foo : Num = 1}`. Should we consider this as a totally opaque
+record, that is `type({foo : Num = 1}) = #{foo : Num = 1}` or take advantage
+from the information that `foo` is necessarily a number and set `type({foo : Num
+= 1}) = {foo: Num}`? The problem with the latter approach is that we loose the
+property of *blame safety*, which is that "well-typed program can't be blamed".
+Indeed, the extracted type is less precise that the actual contract. For
+example, `{foo = 2} : {foo : Num, foo = 1}` would pass typechecking but fails at
+runtime on the failure of merging `1` and `2`. There are also multiple
```
aspiwack, line 291: It probably wouldn't, actually. With the proposed semantics, the runtime contract would presumably be `§{foo: Num}`.

It doesn't make it a good idea, though.

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+approximation possible: we could create an opaque type encoding the tuple `(Num,
```
aspiwack, line 292:
```suggestion
approximations possible: we could create an opaque type encoding the tuple `(Num,
```

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+Eq 2)`, we could set `foo : Dyn`, etc.
```
aspiwack, line 293: I didn't understand the tuple `(Num, Eq 2)` bit.

rfcs/002-merge-types-terms-syntax.md
yannham: It's a convoluted way of saying: I don't want to use term equality as it is too strict, but some kind of operational equivalence instead. However I don't want to define it formally. It wouldn't really make sense, given there is no formal specification of Nickel at this point anyway. Maybe I can just drop what's inside the parentheses.

rfcs/002-merge-types-terms-syntax.md
yannham: I think it is too strict to ask, and is e.g. violated by the fact that currently `{foo : Num}` is a valid term. Thus, according the a dual hypothetical law for `#`, we should have `type({foo : Num}) = #{foo : Num}`. But we certainly want `type({foo: Num}) = {foo : Num}`. Maybe less artificially, and although this is mentioned as a future improvement only, we may want to extract static type information for other terms too, such as `type({foo | Num}) = {foo : Num}`, which would also violate this law.

rfcs/002-merge-types-terms-syntax.md
yannham: Your remark makes me think about something. The `type()` translation erases metadata, so it may have some unwanted consequences, such as `{foo = 1} | {foo : Num | doc "Foo field"}` loosing its documentation at runtime (`type({foo : Num | doc "Foo field"}) = {foo : Num}`). Here maybe the contract feels slightly off, mixing type annotation and doc, but this would be even more natural if we set `type({foo | Num}) = {foo : Num}` at some point, as we would certainly write `{foo = 1} | #{foo | Num | doc "..."}`.

There's also some edgy details about `§{foo : Num}` and `#{foo | Num}` not behaving exactly the same right now. I see `type()` as extracting static type information from an annotation, but being quite irrelevant to runtime (as, at most, it will drop metadata, but never add anything helpful to evaluation). Thus, it may make sense to use the translation uniquely at typechecking time, to extract static type information, but keep the original input for runtime. I may had this in mind implicitly when I wrote this paragraph.

rfcs/002-merge-types-terms-syntax.md
yannham: The idea would be to reintroduce the missing information inside the static type: `foo` is not only a number, but it must be equal to `2` (more generally, `mergeable with 2` is probably the right predicate). That would be an opaque type in the same way `x | #{foo = 2}` is an opaque type saying `x` must have a field `foo` that is mergeable with `2`.

rfcs/002-merge-types-terms-syntax.md
```diff
@@ -0,0 +1,452 @@
...
+
+Because of blame safety, and the different possible trade-offs, **we propose to only translate as
+types records that are already record types in the current syntax**. That is:
+
+```
+type({f1: t1, .., fn : tn}) = {f1: type(t1), .., fn: type(tn)}
```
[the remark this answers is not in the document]
yannham: This is discussed in the `Record fields` section. This makes sense, and wouldn't have any fundamental problem.

But right now it turns out `§{foo : Num}` and `{foo | Num}` are behaving differently as contracts. The former is more eager, in that `{} | {foo: Num}` will fail immediately, while `{} | #{foo | Num}` will have a `foo` field without definition, that needs to be projected before the contract fails. I imagine this comes originally from the idea of contracts being based on merging. The laziness of `|` allows to specify data incrementally, without having the invalid intermediate values to break the contract: `{foo | #{bar | Num, baz | Str}} & {foo.bar = 1} & {foo.baz = "ha"}` will work, but `{foo | {bar : Num, baz : Str}} & {foo.bar = 1} & {foo.baz = "ha"}` won't because the intermediate value `{foo.bar = 1}` is missing the `baz` field.

I'm not sure this was a totally conscious design decision, or rather that contract-as-a-term-record and contract-from-a-record-type were implemented at different times with different flavors. This surely needs to be made uniform, but here I just wanted to avoid having to make the eager vs lazy decision in this RFC and rather postpone it, as long as the choice is forward-compatible and is not a regression with respect to the current situation (currently, `foo | #{bar | Num}` makes the type of `foo` to be totally opaque, without meaningful static information)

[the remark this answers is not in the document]
yannham: Well, to be fair, the choice is not totally forward-compatible, if we choose to either make `#{foo | Num}` or `{foo : Num}` lazy: the semantics of contracts written in on type or the other would be altered. What I meant is that we can make this choice later, and then upgrade the `type()` translation in a forward-compatible manner. Let's say the choice eager vs lazy commute with the change `type({foo | Num}) = #{foo | Num})` => `type({foo | Num}) = {foo : Num}`.

rfcs/002-merge-types-terms-syntax.md
aspiwack: It does sound tricky. At any rate, I agree with being as conservative as we can get away with. We can discover interesting usage patterns with experience.

rfcs/002-merge-types-terms-syntax.md
aspiwack: I think this is adding more confusion than it helps. Probably better to remove this bit.

rfcs/002-merge-types-terms-syntax.md
aspiwack: Maybe cut the sentence in smaller pieces as well. Now that you've told me what to look for, I managed to parse it (after two or three failed attempts). It's perfectly grammatical, but I got quite lost along the way.

rfcs/002-merge-types-terms-syntax.md
yannham: Thinking about this again, we may ask for a weaker law, that is `term(!type(!t)) ~ t` for `t : Term`. Unfortunately, the quirk around `{foo | Num}` and `{foo : Num}` behaving slightly differently is probably breaking this law. We may however mention it as a goal, as the issue is in the current state of the implementation, and is not inherent to this RFC.

aspiwack (approved):
It's in a good enough state as it is, I think. Let's merge and get the implementation done.

PR #574, opened 2022-01-14
Automatically update github actions and flake inputs

garbas:


garbas:
@aspiwack Goal is to keep our dependencies and CI up-to-date and to automate this process. Alternative is doing this manually, which is what we were doing already so far, but not really good.

This PR keeps up-to-date two things:
- github actions 
- flake inputs

Once an update is available a PR is going to be created. This is a common practice to keep your dependencies and CI latest while still requiring some review before it lands in master.

aspiwack:
Does it make sure to only open a single PR? What if we decide to skip an update and close the PR, will it open a new one tomorrow?

yannham (commented):
The idea LGTM, as long as there is a manual verification step (here implemented as the action creating a PR). @aspiwack has a good point: if some update is breaking, will the action repost every week? Another question is that it seems PRs generated from GitHub actions [don't trigger workflows](https://github.com/DeterminateSystems/update-flake-lock#running-github-actions-ci). Do you know if this means we have to do the gym mentioned in the DeterminateSystems' README? Or we can just run workflows manually with a click on the PR (which would then be fine)?

garbas:
@aspiwack @yannham The last time I checked (~6-8 months ago) the updates would be combined in one PR. I don't think they changed the defaults, giving it a quick look. But this can be definitely tuned up if it doesn't match our expectations.

I'm also aware that there will be some changes needed to get everything configured as we want (eg. will github actions trigger workflows), but I would do this after this PR gets merged since this things is hard to know before.

yannham (approved):
Sounds fair.

PR #576, opened 2022-01-18
Doc/syntax

Acaccia:
This PR adds a document in the manual explaining nickel syntax.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -121,11 +121,46 @@ Examples:
 []
 ```
 
-### Record
-<!-- TODO -->
+Lists can be concatenated with the operator `@`:
+```
+> [1] @ [2, 3]
+[ 1, 2, 3 ]
+```
 
-### Dynamic Record
-<!-- TODO -->
+### Record
+Records are key-value storage. They are delimited with by `{` and `}`, elements are separated by `,`.
+key-value elements are noted as `key = value`.
+The keys can be either identifiers, or strings. Values can be of any type.
+Elements inside a record are unordered.
+
+Examples:
+```
+{}
+{a = 3}
+{my_id_n5 = "my id number 5", my_id_n4 = }
+{"5" = 5, six = 6}
+```
+
+A record cannot contain the same key twice, or the values must be identical (see section about _merge_).
+```
+> {a = 3, a = 3}
+{ a = 3 }
+
+> {a = 3, "a" = 3}
+{ a = 3 }
+```
+
+Accessing a record field can be done using the `.` operator :
+```
+> { a = 1, b = 5 }.a
+1
+
+> { a = 1 }.b
+error: Missing field
+
+> { "1" = "one" }."1"
+"one"
+```
 
 ### Enum
 <!-- TODO -->
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -219,7 +219,7 @@ Lists can be concatenated with the operator `@`:
 ```
 
 ### Record
-Records are key-value storage, or in Nickel terms, field-value storage. They are delimited by `{` and `}`, and elements are separated by `,`.
+Records are key-value storage, or in Nickel terms, field-value storage. They are delimited by `{` and `}`, and elements are separated with `,`.
 Field-value elements are noted as `field = value`.
 The fields are strings, but can be written without quotes `"` if they respect identifiers syntax. Values can be of any type.
 Elements inside a record are unordered.
@@ -313,7 +313,7 @@ true
 ```
 
 ## Functions
-A function is declared using the `fun` keyword, then arguments separated by spaces, and finally an arrow `=>` to add the body of the function.
+A function is declared using the `fun` keyword, then arguments separated with spaces, and finally an arrow `=>` to add the body of the function.
 To call a function, just add the arguments after it separated with spaces.
 Functions in Nickel are curried, meaning that a function taking multiple arguments is actually a function that takes a single argument and returns a function taking the rest of the arguments, until it is applied.
 
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -257,6 +257,15 @@ It is possible to write records of records via the *piecewise syntax*, where we 
 { a = { b = 1, c = 2 }, b = 3 }
 ```
 
+Since all fields are strings, we can use string interpolation to create keys or accessing:
+```
+> let k = "a" in { "#{k}" = 1 }
+{ a = 1 }
+
+> let k = "a" in { a = 1 }."#{k}"
+1
+```
+
 ## Constructs
 
 ### If-Then-Else
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -1,8 +1,8 @@
 # Nickel Syntax
 
-## Simple types
+## Simple values
 
-There are tree basic kind of values in Nickel : 
+There are three basic kind of values in Nickel : 
  1. numeric values, 
  2. boolean values,
  3. strings.
@@ -46,7 +46,7 @@ Numbers are also comparable, using the dedicated comparison operators, returning
 
 The boolean values in Nickel are denoted `true` and `false`.
 
-To work with boolean values, we have a the classic logical operators *AND* (&&), *OR* (||) and *NOT* (!).
+To work with boolean values, we have the classic logical operators *AND* (&&), *OR* (||) and *NOT* (!).
 The _AND_ and _OR_ operators are lazy in the evaluation of the second argument. The truth table below gives the results for using these operators:
 
 Examples:
@@ -108,10 +108,10 @@ false
 ```
 
 
-## Composite types
+## Composite values
 
 ### List
-Lists allows grouping of values into sequences. They are delimited by `[` and `]`, and elements are separated with `,`.
+A list is a sequence of values. They are delimited by `[` and `]`, and elements are separated with `,`.
 
 Examples:
 ```
@@ -128,8 +128,8 @@ Lists can be concatenated with the operator `@`:
 ```
 
 ### Record
-Records are key-value storage. They are delimited with by `{` and `}`, elements are separated by `,`.
-key-value elements are noted as `key = value`.
+Records are key-value storage. They are delimited by `{` and `}`, and elements are separated by `,`.
+Key-value elements are noted as `key = value`.
 The keys can be either identifiers, or strings. Values can be of any type.
 Elements inside a record are unordered.
 
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -286,17 +286,6 @@ Examples:
 ["1", "2", "3"]
 ```
 
-The `then` and `else` branches are only evaluated accordingly to the result of the expression inside the `if`.
-
-Examples:
-```
-> ["1"] @ (if true then ["2"] else "error") @ ["3"]    // Here only the expression in 'then' is evaluated, which is correct
-["1", "2", "3"]
-
-> ["1"] @ (if false then ["2"] else "error") @ ["3"]    // Here only the expression in 'else' is evaluated, which results in a type error
-error: Type error
-```
-
 ### Let-In
 Let-in allows the binding of an expression. It is used as `let <ident> = <expr> in <expr>`.
 
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -31,11 +31,11 @@ There are a some predefined operators for working with numbers :
 | /        | The division operator                                | `1 / 2 = 0.5` |
 | %        | The modulo operator (returns the *signed* remainder) | `5 % 3 = 2`   |
 
-> **Remark about the `-` operator:***  
-> Since `-` can be used inside of an identifier, the subtraction operators **needs** to be surrounded by spaces,
+> **Remark about the `-` operator:**  
+> Since `-` can be used inside an identifier, the subtraction operators **needs** to be surrounded by spaces,
 > write `1 - 1`, not `1-1`.
 
-Numbers are also comparable, using the dedicated comparison operators, returning a boolean value :
+Numbers can be compared using the following operators :
 | Operator | Description      | Example   |
 |:--------:|:----------------:|:---------:|
 | ==       | Equal            | `5 == 5`  |
@@ -45,10 +45,10 @@ Numbers are also comparable, using the dedicated comparison operators, returning
 | >=       | Greater or Equal | `1 >= 1`  |
 | <=       | Smaller or Equal | `-1 <= 6` |
 
-All those operators have precedence with each others. In the table below, you will find the operators sorted from highest to lowest precedence:
+In the table below, you will find the operators sorted from highest to lowest precedence:
 |       Operators      | Associativity | Remark                                        |
 |:--------------------:|:-------------:|-----------------------------------------------|
-|       `( ... )`      |               | parentheses always have to highest precedence |
+|       `( ... )`      |               | parentheses always have the highest precedence |
 |          `-`         | right-to-left | unary negation (as in `-1`)                   |
 |     `*`, `/`, `%`    | left-to-right |                                               |
 |       `+`, `-`       | left-to-right | binary addition and subtraction               |
@@ -59,8 +59,8 @@ All those operators have precedence with each others. In the table below, you wi
 
 The boolean values in Nickel are denoted `true` and `false`.
 
-To work with boolean values, we have the classic logical operators *AND* (&&), *OR* (||) and *NOT* (!).
-The _AND_ and _OR_ operators are lazy in the evaluation of the second argument. The truth table below gives the results for using these operators:
+Nickel features the classical boolean operators *AND* (&&), *OR* (||) and *NOT* (!).
+The _AND_ and _OR_ operators are lazy in the evaluation of the second argument: for example, in `exp1 && exp2`, `exp2` is only evaluated if `exp1` evaluates to `false`. The truth table of these operators is given below:
 
 Examples:
 ```
@@ -81,7 +81,7 @@ Strings are enclosed by `" ... "` for a single line string or by `m#" ... "#m` f
 They can be concatenated with the operator `++`.
 Strings must be UTF-8 valid.
 
-String interpolation exists in Nickel, via the notation `"#{ < expression that evaluates to a string > }"`.
+The string interpolation syntax is `"#{ < expression that evaluates to a string > }"`.
 
 Examples:
 ```
@@ -178,7 +178,7 @@ error: Missing field
 "one"
 ```
 
-It is possible to write records of records via the *pieces syntax*, where we separate fields by dots:
+It is possible to write records of records via the *piecewise syntax*, where we separate fields by dots:
 ```
 > { a = { b = 1 } }
 { a = { b = 1 } }
@@ -261,10 +261,9 @@ let r : { _ : Num } = { a = 1, b = 2 }
 ```
 
 ## Metadata
-Metadata can be used to add contracts (more information in relevant documentation), documentation or default values.
-It is introduced with a `|`. Multiple metadata can be chained.
+Metadata are used to attach contracts (more information in relevant documentation), documentation or priority to values.
+A metadata is introduced with the syntax `<value> | <metadata>`. Multiple metadata can be chained.
 
-For adding a regular type as contract, the syntax is `| < type >`. For a defined contract, we use the `| #< defined type >`.
 
 Examples:
 ```
@@ -286,7 +285,7 @@ error: Blame error: contract broken by a value.
 3
 ```
 
-Adding documentation can be done with `| doc < a string >`.  
+Adding documentation can be done with `| doc < string >`.  
 Examples:
 ```
 > 5 | doc "The number five"
@@ -301,7 +300,7 @@ Examples:
 true
 ```
 
-Default values are used for records contracts, to give a default values to fields.
+Record contracts can set default values using the `default` metadata:
 It is noted as `| default = < default value >`.  
 Examples:
 ```
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -145,7 +145,7 @@ Examples:
 "Hello World"
 
 > let w = "World" in m##"Hello #{w}"##m
-"Hello #{"World"}"
+"Hello #{w}"
 
 > let w = "World" in m##"Hello ##{w}"##m
 "Hello World"
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -428,7 +428,8 @@ true
 ```
 
 Record contracts can set default values using the `default` metadata:
-It is noted as `| default = < default value >`.  
+It is noted as `| default = < default value >`.
+This is especially useful when merging records (more about this in the dedicated document about merge).  
 Examples:
 ```
 > let Ais2ByDefault = { a | default = 2 } in {} | #Ais2ByDefault
@@ -436,4 +437,10 @@ Examples:
 
 > let Ais2ByDefault = { a | default = 2 } in { a = 1 } | #Ais2ByDefault
 { a = 1 }
+
+> { foo | default = 1, bar = foo + 1 }
+{ foo = 1, bar = 2 }
+
+> {foo | default = 1, bar = foo + 1} & {foo = 2} {foo = 2, bar = 3}
+{ foo = 2, bar = 3 }
 ```
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -229,7 +229,7 @@ Examples:
 ```
 {}
 {a = 3}
-{my_id_n5 = "my id number 5", my_id_n4 = }
+{my_id_n5 = "my id number 5", "my id n4" = "my id number 4" }
 {"5" = 5, six = 6}
 ```
 
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -1,0 +1,113 @@
+# Nickel Syntax
+
+## Simple types
+
+There are tree basic kind of values in Nickel : 
+ 1. numeric values, 
+ 2. boolean values,
+ 3. strings.
+ 
+### Numeric values
+
+Nickel has a support for numbers, positive and negative, with or without decimals.
+Internally, those numbers are stored as 64-bits floating point numbers, following the IEEE 754 standard.
+
+Examples:
+
+```
+1
+0.543
+42
+-1000000
+-6.8
+```
+
+There are a some predefined operators for working with numbers :
+| Operator | Description                                          | Example       |
+|:--------:|:----------------------------------------------------:|:-------------:|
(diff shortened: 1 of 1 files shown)


doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
+# Nickel Syntax
+
+## Simple types
+
+There are tree basic kind of values in Nickel :
```
Avi-D-coder, line 5:
```suggestion
There are three basic kind of values in Nickel : 
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+| >        | Greater than     | `1 > -5`  |
+| >=       | Greater or Equal | `1 >= 1`  |
+| <=       | Smaller or Equal | `-1 <= 6` |
+
+
+### Boolean values
+
+The boolean values in Nickel are denoted `true` and `false`.
+
+To work with boolean values, we have a the classic logical operators *AND* (&&), *OR* (||) and *NOT* (!).
```
Avi-D-coder, line 49:
```suggestion
To work with boolean values, we have the classic logical operators *AND* (&&), *OR* (||) and *NOT* (!).
```

Avi-D-coder (changes requested):

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+Numbers are also comparable, using the dedicated comparison operators, returning a boolean value :
+| Operator | Description      | Example   |
+|:--------:|:----------------:|:---------:|
+| ==       | Equal            | `5 == 5`  |
+| !=       | Not Equal        | `5 != 4`  |
+| <        | Smaller than     | `2 < 3`   |
+| >        | Greater than     | `1 > -5`  |
+| >=       | Greater or Equal | `1 >= 1`  |
+| <=       | Smaller or Equal | `-1 <= 6` |
+
```
yannham, line 43: It might be important to underline here that because we allow `-` (dash) in identifiers, you must always put space around subtraction. That is, `a-b` is parsed as the identifier `a-b`, not the subtraction. One must write `a - b`.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+
```
yannham, line 44: We should maybe talk about operator precedence. Or have a table at the end. Even then, recalling the precedence of `+`, `*` and `/`  here may be a good idea.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+```
+> true && false
+false
+
+> false || true
+true
+
+> ! true
+false
+```
```
yannham, line 62: Same, something about `&&` taking precedence over `||` could be nice.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+"Well, if this isn't a string?
+Yes it is, indeed it is"
+
+> "Hello" ++ "World"
+"HelloWorld"
+```
+
+## Equality Operators
+
+We have seen that the operators `==` and `!=` works for comparing numbers, but they can actually compare any values together. Two values are equal if they have the same _type_ and the same _value_.
```
yannham, line 87: Saying that "two values are equal if they have the same value" is slightly tautological, even if I see where you are coming from. Maybe getting rid of the `and the same _value_`, and add rather something more explicit about `==` not doing implicit conversions? If that's the point you want to make.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+
+> 5 == "Hello"
+false
+
+> true == "true"
+false
+```
+
+
+## Composite types
```
yannham, line 111:
```suggestion
## Composite values
```
You start with `values`, which is a good idea, because in a typed language `type` already means something. So better to stick to it.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+
+### List
+Lists allows grouping of values into sequences. They are delimited by `[` and `]`, and elements are separated with `,`.
```
yannham, line 114:
```suggestion
A list is a sequence of values. They are delimited by `[` and `]`, and elements are separated with `,`.
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+Lists can be concatenated with the operator `@`:
+```
+> [1] @ [2, 3]
+[ 1, 2, 3 ]
+```
+
+### Record
+Records are key-value storage. They are delimited with by `{` and `}`, elements are separated by `,`.
```
yannham, line 131:
```suggestion
Records are key-value storage. They are delimited by `{` and `}`, and elements are separated by `,`.
```

yannham, line 131: You should probably introduce the word `field`. This is the one used in most of the code/documentation.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+key-value elements are noted as `key = value`.
```
yannham, line 132:
```suggestion
Key-value elements are noted as `key = value`.
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+The keys can be either identifiers, or strings. Values can be of any type.
```
yannham, line 133: Pedantically, all keys are strings. However they can be written like identifiers, or escaped if needed. I'm sorry, this is not a great way of saying it either, but the current sentence feels like strings and identifiers are two different, separate kind of keys.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+
+Examples:
+```
+{}
+{a = 3}
+{my_id_n5 = "my id number 5", my_id_n4 = }
+{"5" = 5, six = 6}
+```
+
+A record cannot contain the same key twice, or the values must be identical (see section about _merge_).
```
yannham, line 144: That's true for primitive values, but not for records. For example `{foo = {bar = 1}, foo = {baz = 2}}` is totally fine, and is what `{foo.bar = 1, foo.baz = 2}` is desugared to. You can either say that exactly (primitive values must be the same, but records are treated differently) or totally offload the discussion to the merging section, just saying that multiple definitions are merged.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+```
+> {a = 3, a = 3}
+{ a = 3 }
+
+> {a = 3, "a" = 3}
+{ a = 3 }
```
yannham, line 150: maybe add an example of something that fails?

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+> { a = 1, b = 5 }.a
+1
+
+> { a = 1 }.b
+error: Missing field
+
+> { "1" = "one" }."1"
+"one"
+```
+
```
yannham, line 164: This section could have too:

- the definition by pieces syntax, like `{foo.bar.baz = value}`. This is desguared to `{foo = {bar = {baz = value}}}`. As mentioned above, you can define one same record in several pieces: `{foo.bar = 1, foo.baz = 2}`. The two desugared versions of `foo` will be then merged.
- metadata. Or this may be a different section, but we must have that the `|` operator is used to attach metadata to values. And list the one available: contract annotation (without keyword), `doc`, `default`.
- merging. We won't define it there, but just says that `&` recursively merge records, and point to the relevant section.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
...
+> let a = 1 in let b = 2 in a + b
+3
+```
+
+The identifier in the `let` part can be typed (more about typing in the dedicated document):
+```
+> let a : Num = 6 in 6 * 7
+42
+```
```
yannham, line 201: I'm not sure this should be in the let section. The thing is, any value can have a type annotation `5 : Num`. Then, any *binding* (record fields and let) have this syntactic sugar `let foo : Type = ...` or `{foo : Type = ...}` as syntactic sugar for respectively `let foo = (... : Type)` and `{foo = (... : Type)}`. It could be in a separate section "Annotations" or something, together with metavalues.

doc/manual/syntax.md
```diff
@@ -0,0 +1,221 @@
+# Nickel Syntax
+
+## Simple types
```
yannham, line 3:
```suggestion
## Simple values
```

doc/manual/syntax.md
Acaccia: added note

doc/manual/syntax.md
Acaccia: added

doc/manual/syntax.md
Acaccia: added

doc/manual/syntax.md
Acaccia: rephrased

doc/manual/syntax.md
Acaccia: removed the section, the reader should read the documentation about merge.

doc/manual/syntax.md
Acaccia: section was removed

doc/manual/syntax.md
Acaccia: done

doc/manual/syntax.md
Acaccia: Moved typing to its own section

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+Examples:
+```
+> "Hello, World!"
+"Hello, World!"
+
+> m#"Well, if this isn't a multiline string?
+Yes it is, indeed it is"#m
+"Well, if this isn't a string?
+Yes it is, indeed it is"
```
yannham, line 94: It could be nice to have multiline strings inside indented code, to see the difference:
```
{
  program = {
    shell_code = m#"
      if [ -f /var/log/messages ]; then
          echo "/var/log/messages exists."
      fi
    "#m
  }
}
```
then `program.shell_code` is:
```
"if [ -f /var/log/messages ]; then
  echo "/var/log/messages exists."
fi"

File: doc/manual/syntax.md [hunk 18]
```diff
@@ -0,0 +1,313 @@
...
+1
+
+> { a = 1 }.b
+error: Missing field
+
+> { "1" = "one" }."1"
+"one"
+```
+
+It is possible to write records of records via the *pieces syntax*, where we separate fields by dots:
```
yannham, line 181:
```suggestion
It is possible to write records of records via the *piecewise syntax*, where we separate fields by dots:
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+{ a = { b = 1 } }
+
+> { a.b = 1, a.c = 2, b = 3}
+{ a = { b = 1, c = 2 }, b = 3 }
+```
+
+## Statements
+
+### If-Then-Else
+This statement allows conditional branching in your code. You can use it as `if <bool expr> then <expr> else <expr>`. Both expressions in the `then` and `else` branches are lazily evaluated, and must have the same type.
```
yannham, line 196: What do you mean by `lazily evaluated`? If the `else` is evaluated even when the condition is `false`, then I guess it's not really an `if-then-else`.

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+```
+> if true then "TRUE :)" else "false :("
+"TRUE :)"
+
+> if false then "Not this one" else "This one"
+"This one"
+
+> if "forty-two" == 42 then "equal?" else "unequal"
+"unequal"
+```
```
yannham, line 208: Could add an example showing that `if` is an expression (there is no statement in Nickel), like
```
>["1"] @ (if 42 == "42" then ["3"] else ["2"]) @ ["3"]
["1", "2", "3"]
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+let add = fun a b => a + b in add 1 2
+3
+
+> let add = fun a b => a + b in
+    let add1 = add 1 in
+        add1 2
+3
+```
+
```
yannham, line 243: Should also talk about the curried operator syntax. For each infix operator `+`, `-`, `@`, etc. there is an associated function `(op)`, such as `(@)`, and `(+)` that turns it into a function. It can be (partially) applied and passed as a function:

```
>let plus_one = (+) 1 in
let times_two = (*) 2 in
plus_one (times_two 2)
5

>let flatten = lists.fold (@) [] in
flatten [[1,2], [3,4]]
[ 1, 2, 3, 4 ]
```

We can also introduce the reverse application operator `|>` here. `x |> f` is `f x`. It is left associative, such that `x |> f |> g` is `g (f  x)`. It is useful to chain operations on a value:

```
[1,2,3,4,5,6]
|> lists.map ((+) 1)
|> lists.filter ((<=) 5) // this is fun x =>  5 <= x, so elements GREATER than 5
|> lists.elem_at 0
// 5
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+{a: Num = 1, b: Bool = true, c : List Num = [ 1 ]}
+let r : {a : Num, b : Bool, c : List Num} = { a = 1, b = true, c = [ 1 ] }
+
+{ a = 1, b = 2 } : { _ : Num }
+let r : { _ : Num } = { a = 1, b = 2 }
+```
+
+## Metadata
+Metadata can be used to add contracts (more information in relevant documentation), documentation or default values.
```
yannham, line 264:
```suggestion
Metadata are used to attach contracts (more information in relevant documentation), documentation or priority to values.
```

Avi-D-coder (approved):

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+It is introduced with a `|`. Multiple metadata can be chained.
```
yannham, line 265:
```suggestion
A metadata is introduced with the syntax `<value> | <metadata>`. Multiple metadata can be chained.
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+For adding a regular type as contract, the syntax is `| < type >`. For a defined contract, we use the `| #< defined type >`.
```
yannham, line 267:
```suggestion
```
I think we can just get rid of this one, since this won't be true anymore once we have merged the type and contract syntax.

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+> let SmallNum = contracts.from_predicate (fun x => x < 5) in 10 | #SmallNum
+error: Blame error: contract broken by a value.
+
+> let SmallNum = contracts.from_predicate (fun x => x < 5) in
+    let NotTooSmallNum = contracts.from_predicate (fun x => x >= 2) in
+        3 | Num | #SmallNum | #NotTooSmallNum
+3
+```
+
+Adding documentation can be done with `| doc < a string >`.
```
yannham, line 289:
```suggestion
Adding documentation can be done with `| doc < string >`.  
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+> true | Bool | doc m#"
+    If something is true,
+    it is based on facts rather than being invented or imagined,
+    and is accurate and reliable.
+    (Collins dictionary)
+    "#m
+true
+```
+
+Default values are used for records contracts, to give a default values to fields.
```
yannham, line 304:
```suggestion
Record contracts can set default values using the `default` metadata:
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,313 @@
...
+It is noted as `| default = < default value >`.  
+Examples:
+```
+> let Ais2ByDefault = { a | default = 2 } in {} | #Ais2ByDefault
+{ a = 2 }
+
+> let Ais2ByDefault = { a | default = 2 } in { a = 1 } | #Ais2ByDefault
+{ a = 1 }
+```
```
yannham, line 313: Could add an example of the usage of default values for merging/overriding:
```
> {foo | default = 1, bar = foo + 1} & {foo = 2}
{foo = 2, bar = 3}

File: doc/manual/syntax.md [hunk 28]
```diff
@@ -0,0 +1,313 @@
...
+{ a = { b = 1 } }
+ 
+> { a.b = 1 }
+{ a = { b = 1 } }
+
+> { a.b = 1, a.c = 2, b = 3}
+{ a = { b = 1, c = 2 }, b = 3 }
+```
+
+## Statements
```
yannham, line 193: There is no statement in Nickel, everything is an expression. Control-flow maybe? But `let .. in ..` is arguably not control flow. Maybe just `Constructs`, or something vague and general enough? Blocks?

doc/manual/syntax.md
Acaccia: @yannham I thought about this, but I have the impression this is out of scope for the syntax document.

doc/manual/syntax.md
Acaccia: I didn't know they existed, this is so cool

doc/manual/syntax.md
Acaccia: I have no idea what you don't understand here. I will rephrase, but I'm not sure it will be better...

doc/manual/syntax.md
Acaccia: Constructs seems good for me, I'll use it.

doc/manual/syntax.md
Acaccia: added an example

doc/manual/syntax.md
Acaccia: I shamelessly cut/paste your example.

doc/manual/syntax.md
yannham: Mmh. What about just putting the example, but without explaining it further? The idea is be to reflect how `default` can be encountered in the wild, but deferring any deep explanation to the relevant manual section?

doc/manual/syntax.md
yannham: Sorry, my comment is not very clear. My question is: is the "lazy evaluated" part meaningful or useful? What would be an `if-then-else` that is not  "lazy", in another language?

doc/manual/syntax.md
yannham: Yes, this is why this is a good thing you're writing this document :slightly_smiling_face: the knowledge won't be hidden in the code and our heads anymore. This syntax exists in other ML descendants, like Haskell and OCaml.

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+"This one"
+
+> if "forty-two" == 42 then "equal?" else "unequal"
+"unequal"
+
+> ["1"] @ (if 42 == "42" then ["3"] else ["2"]) @ ["3"]
+["1", "2", "3"]
+```
+
+The `then` and `else` branches are only evaluated accordingly to the result of the expression inside the `if`.
```
yannham, line 289: Sorry to insist, but I still don't understand the why of this paragraph. To me it is the very definition of an if-then-else, and as such it is obvious. Unless I'm misunderstanding something? What point do you want to make here, that would be true for Nickel but not for other languages?

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+let add = fun a b => a + b in add 1 2
+3
+
+> let add = fun a b => a + b in
+    let add1 = add 1 in
+        add1 2
+3
+```
+
+All existing infix operators in Nickel can be turned into prefix functions by putting them inside parentheses.
```
yannham, line 334:
```suggestion
All existing infix operators in Nickel can be turned into functions by putting them inside parentheses.
```
I see what you mean, but to someone especially into parsing or PL (which Nickel users may be), it may sound a bit strange, as if we now have different types of functions, prefix functions, and other. At this point, I think it sufficient and clearer to just say this turns operators into functions.

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+
+Functions might be composed using the *pipe operator*. The pipe operator allows for a function application `f x` to be written as `x |> f`.
+This operator is left-associative, so `x |> f |> g` will be interpreted as `g (f x)`.
+
+Examples:
+```
+> "Hello World" |> strings.split " "
+["Hello", "World"]
+
+> "Hello World" |> strings.split " " |> lists.head
```
yannham, line 364:
```suggestion
> "Hello World"
  |> strings.split " "
  |> lists.head
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+"Hello"
+
+> "Hello World" |> strings.split " " |> lists.head |> strings.uppercase
```
yannham, line 367:
```suggestion
> "Hello World"
  |> strings.split " "
  |> lists.head
  |> strings.uppercase
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+
+Examples:
+```
+> 5 | Num
+5
+
+> 5 | Bool
+error: Blame error: contract broken by a value.
+
+> let SmallNum = contracts.from_predicate (fun x => x < 5) in 1 | #SmallNum
```
yannham, line 403:
```suggestion
> let SmallNum = contracts.from_predicate (fun x => x < 5) in
1 | #SmallNum
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+1
+
+> let SmallNum = contracts.from_predicate (fun x => x < 5) in 10 | #SmallNum
```
yannham, line 406:
```suggestion
> let SmallNum = contracts.from_predicate (fun x => x < 5) in
10 | #SmallNum
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+error: Blame error: contract broken by a value.
+
+> let SmallNum = contracts.from_predicate (fun x => x < 5) in
+    let NotTooSmallNum = contracts.from_predicate (fun x => x >= 2) in
```
yannham, line 410:
```suggestion
  let NotTooSmallNum = contracts.from_predicate (fun x => x >= 2) in
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+        3 | Num | #SmallNum | #NotTooSmallNum
```
yannham, line 411:
```suggestion
  3 | Num
    | #SmallNum
    | #NotTooSmallNum
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+    "#m
+true
+```
+
+Record contracts can set default values using the `default` metadata:
+It is noted as `| default = < default value >`.
+This is especially useful when merging records (more about this in the dedicated document about merge).  
+Examples:
+```
+> let Ais2ByDefault = { a | default = 2 } in {} | #Ais2ByDefault
```
yannham, line 435:
```suggestion
> let Ais2ByDefault = { a | default = 2 } in
  {} | #Ais2ByDefault
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+{ a = 2 }
+
+> let Ais2ByDefault = { a | default = 2 } in { a = 1 } | #Ais2ByDefault
```
yannham, line 438:
```suggestion
> let Ais2ByDefault = { a | default = 2 } in
  { a = 1 } | #Ais2ByDefault
```

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+{ a = 1 }
+
+> { foo | default = 1, bar = foo + 1 }
+{ foo = 1, bar = 2 }
+
+> {foo | default = 1, bar = foo + 1} & {foo = 2} {foo = 2, bar = 3}
```
yannham, line 444:
```suggestion
> {foo | default = 1, bar = foo + 1} & {foo = 2}
```

yannham (approved):
It now seems to include everything we need to cover, and after having addressed the comments of this review, should be in good shape for being merged. We will do a last pass after the merging types and syntax RFC is fully implemented.

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+```
+
+Lists can be concatenated with the operator `@`:
+```
+> [1] @ [2, 3]
+[ 1, 2, 3 ]
+```
+
+### Record
+Records are key-value storage, or in Nickel terms, field-value storage. They are delimited by `{` and `}`, and elements are separated by `,`.
```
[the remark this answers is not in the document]
Acaccia: @yannham Why do you prefer "with" to "by" in this case ?

doc/manual/syntax.md
Acaccia: I don't see why it is obvious, especially for a configuration language, that someone might use without any prior knowledge of programming.

doc/manual/syntax.md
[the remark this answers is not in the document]
yannham: I did a shallow google search and both seemed to be totally valid in this case, so the choice was quite arbitrary, to be honest. Picking `by` is totally fine if you prefer to. The suggestion is to pick one and use it consistently across this document.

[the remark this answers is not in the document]
Acaccia: Okay, then I will change it everywhere, with the almighty power of sed

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+Field-value elements are noted as `field = value`.
+The fields are strings, but can be noted as identifiers if they respect identifiers syntax (internally, they will be stored as strings). Values can be of any type.
+Elements inside a record are unordered.
+Two records can be _merged_ together using the operator `&`. The reader can find more information about merging in the relevant documentation.
+
+Examples:
+```
+{}
+{a = 3}
+{my_id_n5 = "my id number 5", my_id_n4 = }
```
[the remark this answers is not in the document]
Acaccia: oopsie

doc/manual/syntax.md
Acaccia: meh, removed anyway

doc/manual/syntax.md
```diff
@@ -0,0 +1,446 @@
...
+Strings must be UTF-8 valid.
+
+The string interpolation syntax is `"#{ < expression that evaluates to a string > }"`.
+
+Examples:
+```
+> "Hello, World!"
+"Hello, World!"
+
+> m#"Well, if this isn't a multiline string?
```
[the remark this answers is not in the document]
Acaccia: I already evoke multiline strings above, so it seems normal I give a small example ?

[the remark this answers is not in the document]
yannham: Oh right, I thought they were only introduced for the first time later. Then one possibility is to make a first paragraph mentioning only strings, and in a second step talk about multiline strings. It just felt slightly odd to mention them and give a first example but without really knowing what they are about, and then actually explain what they are for. But it's my very arbitrary opinion. You can also keep it as it is, if you prefer to. Examples usually don't hurt

PR #580, opened 2022-01-18
Manual/merging

francois-caddet:
Relate to #573 
Document merging and fields overwriting for records.

### DIFF
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -4,7 +4,7 @@ When you have a big configuration, the way to split it in several files
 (e.g.: separated by cathegorie or level of abstraction) is the merging operator.
 In nickel this is the `&` operator.
 
-For instence, having a server config and a firewall config, the network config
+For instance, having a server config and a firewall config, the network config
 could be:
 
 ```text
@@ -15,8 +15,7 @@ in
 server & firewall
 ```
 
-In the simple case where both records do not have common fields, it gives a
-record which is the union of both.
+In the simple case where records `x` and `y` have no common fields, `x&y` is the union of `x` and `y`.
 
 If you have to merge records with fields in commons, you will use two new concepts:
 
@@ -95,7 +94,7 @@ depend on.
 
 To simplify, when merging two records, all contracts of both left and right one
 are applied to the resulting one.
-For instence:
+For instance:
 
 ```text
 
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -277,9 +277,10 @@ depend field for clarity but both behave the same.
 
 ### Contracts crossvalidation
 
-When merging two records, all contracts of both left and right one
-are applied to the resulting one.
-For instance:
+When merging records with a common field foo and different contracts attached on
+each side (say `A1`, `A2` on the left and `B` on the right), then merging
+ensures that the final value will respect the combination of all the contracts
+(`A1`, `A2`, `B`). For instance:
 
 ```text
 let Port | doc "A contract for a port number"
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -1,0 +1,66 @@
+# Merging records
+
+When you have a big configuration, the way to split it in several files
+(e.g.: separated by cathegorie or level of abstraction) is the merging operator.
+In nickel this is the `&` operator.
+
+For instence, having a server config and a firewall config, the network config
+could be:
+
+```text
+let
+server = import "server.ncl",
+firewall = import "firewall.ncl",
+in
+server & firewall
+```
+
+In the simple case where both record does not have common fields, it give a
+record which is the union of both ones.
+
+If you have to merge with fields in commons, you will use two new concepts:
+
+- default annotation,
+- overwriting,
+
+## Default annotation.
+
+To be able to merge records with commons fields, at least one side field has to
+be annotated as default. It means, if you have two records:
+
+```text
+let left = {
+  firewall.enabled = true,
+  firewall.type = "iptables",
+  firewall.open_ports = [21, 80, 443],
+} in
+let right = {
+  firewall.enabled = false,
+  server.host.options = "TLS",
+} in
+left & right
+```
+
+Like it is, it's impossible to merge and will throw an unmergeable terms error.
+Here, the issue is on the field `firewall.enabled`. What you have to do to fix it
+is to annotate this field as default in one of the records:
+
+```text
+let left = {
+  firewall.enabled | default = true,
+  firewall.type = "iptables",
+  firewall.open_ports = [21, 80, 443],
+} in
+let right = {
+  firewall.enabled = false,
+  server.host.options = "TLS",
+} in
+left & right
+// => {firewall.enabled = false, ...}
+```
+
+The default annotation is generaly to give a default value to a record field.
+So, this value can be changed afterward. Saying it in a more abstract way,
+default indicate a smaller priority to a field in case of merging. Saying that,
+If both side would have been annotated `default`, the merge is not possible
+neither.
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -1,15 +1,13 @@
 # Merging records
 
-The merging concept in nickel is a very powerfull behaviour. It's performed by
-the `&` operator.
+Merging in Nickel allows to compose small, logical, and manageable blocks into a
+potentially complex final configuration.
 
-Merging is used in several contexts, from the simplest to the more powerfull.
 In this part we will try to described this concept exaustively as possible.
 
 Warning: The given examples are clearly not the only context on which merging
 can be used. But it's impossible to expose all the cases in a user manual.
-You can check the examples in github for further readings. Also, you can check
-RFC002 in the github repository.
+You can check the examples in github for further readings.
 
 Warning: Nickel beeing in a pre 1st release state now. Merging is a feature
 which can recieve breacking updates until passing in 1.0.0.
@@ -28,10 +26,22 @@ If you have to merge records with fields in commons, you could be in following c
 The simplest merging case is when both records does not have any common fields.
 It can be used to merge records from differents files in one only record.
 
+### Usecases
+
 You generaly will use this feature to be able to split a config in subparts.
 When your config has a small set of big subparts, they can themself be either
 a one top key record but also a multikeys record. The idea is to group the top
 level records by topic.
+
+### Description
+
+
+Having two records, `x` and `y`. With `x = {x0 = vx0, x1 = vx1, ..., xn = vxn}` and
+`y = {y0 = vy0, y1 = vy1, ..., yn = vyn}`. The merge of both give:
+`x & y = {x0 = vx0, y0 = vy0, x1 = vx1, y1 = vy1, ..., xn = vxn, yn = vyn}`
+In other terms, `x & y` is the union of `x` and `y`.
+
+### Example
 
 For instance, having a server config and a firewall config, the network config
 could be:
@@ -50,28 +60,44 @@ could be:
     open_ports = [23, 80, 443],
 }
 
-// file: server.ncl
-let
-server = import "server.ncl",
-firewall = import "firewall.ncl",
-in
+// file: network.ncl
+let server = import "server.ncl" in
+let firewall = import "firewall.ncl" in
 server & firewall
 ```
 
 In this simple case where records `x` and `y` have no common fields, `x&y` is the
-union of `x` and `y`.
+union of `x` and `y`:
+
+```text
+{
+    host_name = "example",
+    host = "example.org",
+    ip_addr = "0.0.0.0",
+    enable_firewall = true,
+    open_ports = [23, 80, 443],
+}
+```
 
 ## Recursivity of the merge
 
 It's a behaviour you will use in similar cases as before but when the split is
-at a deeper level (e.g.: in two different files you have unintersecting fields
-but with the same "root path" `open_ports.` or even `firewall.open_ports`).
+at a deeper level.
+
+### Usecases
+
+In two records you have unintersecting fields but with common
+"root path" (e.g.: `open_ports.` as in following example; can also even be
+deeper as `firewall.open_ports.`).
+
+### Description
+
 When merging, in the case of intersection of fields, Nickel will try to
 recursively merge them. That mean: if you have tow records, `x` and `y`, both
 having a field `a`; if `a` is a record, the merge will be a record `z` with a
 field a beeing the merge between `x.a` and `y.a` or `z = {a = x.a & y.a}.
 
-Example:
+### Example
 
 ```text
 // file: firewall/udp.ncl
@@ -85,11 +111,9 @@ Example:
 }
 
 // firewall.ncl
-let
-udp = import "firewall/udp.ncl",
-tcp = import "firewall/tcp.ncl",
-in
-udp & tcp
+let udp = import "firewall/udp.ncl" in
+let tcp = import "firewall/tcp.ncl" in
+{ firewall= udp & tcp}
 ```
 
 In the above example, we merge two records, both with a field `open_ports`. The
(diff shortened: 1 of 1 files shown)


doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+```text
+let
+server = import "server.ncl",
+firewall = import "firewall.ncl",
+in
+server & firewall
+```
+
+In the simple case where both record does not have common fields, it give a
+record which is the union of both ones.
```
Avi-D-coder, line 19:
```suggestion
record which is the union of both.
```

Avi-D-coder: Could also be "record which is the union of both records".

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+If you have to merge with fields in commons, you will use two new concepts:
```
Avi-D-coder, line 21:
```suggestion
If you have to merge records with fields in commons, you will use two new concepts:
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+- default annotation,
+- overwriting,
+
+## Default annotation
+
+To be able to merge records with commons fields, at least one side field has to
+be annotated as default. It means, if you have two records:
```
Avi-D-coder, line 29:
```suggestion
be annotated as default. This means, if you have two records:
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+  firewall.enabled = false,
+  server.host.options = "TLS",
+} in
+left & right
+// => {firewall.enabled = false, ...}
+```
+
+The default annotation is generaly to give a default value to a record field.
+So, this value can be changed afterward. Saying it in a more abstract way,
+default indicate a smaller priority to a field in case of merging. Saying that,
```
Avi-D-coder, line 64:
```suggestion
default indicate a lower priority to a field in case of merging. Saying that,
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+If both side would have been annotated `default`, the merge is not possible
```
Avi-D-coder, line 65:
```suggestion
If both sides have been annotated `default`, the merge is not possible.
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+neither.
```
Avi-D-coder, line 66:
```suggestion
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Overwriting
+
+The overwriting is the concept specifying the behaviour of a merge when you
+overwrite a field on which depend an otherone. This  feature is described in
```
Avi-D-coder, line 71:
```suggestion
overwrite a field on which depends an other one. This  feature is described in
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+["#573 [RFC] Merge types and terms syntax proposal"](https://github.com/tweag/nickel/pull/573)
+in more details.
+
+Here, we will simply explain what appen in the following case:
+We have a record with some fields annotated `default` and others depending on
+these ones. An exemple could be:
```
Avi-D-coder, line 77:
```suggestion
these ones. An example could be:
```

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+neither.
+
+## Overwriting
+
+The overwriting is the concept specifying the behaviour of a merge when you
+overwrite a field on which depend an otherone. This  feature is described in
+["#573 [RFC] Merge types and terms syntax proposal"](https://github.com/tweag/nickel/pull/573)
+in more details.
+
+Here, we will simply explain what appen in the following case:
```
Avi-D-coder, line 75: This sentence should be reworded.

doc/manual/merging.md
```diff
@@ -0,0 +1,118 @@
...
+    firewall.open_proto.ftp | default = true,
+    firewall.open_ports = []
+        @ (if firewall.open_proto.ftp then [21] else [])
+	@ (if firewall.open_proto.http then [80] else [])
+	@ (if firewall.open_proto.https then [443] else []),
+} in // security => {firewall.open_ports = [21, 80, 443]
+security & {firewall.open_proto.ftp = false} // => {firewall.open_ports = [80, 443]
+```
+
+We then see that depending fields are updated when you overwrite the fields they
```
Avi-D-coder, line 92:
```suggestion
We then see that dependent fields are updated when you overwrite the fields they
```

Avi-D-coder (changes requested):

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
+# Merging records
+
+When you have a big configuration, the way to split it in several files
+(e.g.: separated by cathegorie or level of abstraction) is the merging operator.
+In nickel this is the `&` operator.
```
aspiwack, line 5: We'll have to change this paragraph, it's a little awkward. And probably doesn't hammer strongly enough why the reader should read on.

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+
+For instence, having a server config and a firewall config, the network config
```
aspiwack, line 7:
```suggestion
For instance, having a server config and a firewall config, the network config
```

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+```text
+let
+server = import "server.ncl",
+firewall = import "firewall.ncl",
+in
+server & firewall
+```
+
+In the simple case where both records do not have common fields, it gives a
+record which is the union of both.
```
aspiwack, line 19:
```suggestion
In the simple case where records `x` and `y` have no common fields, `x&y` is the union of `x` and `y`.
```

In addition, give an example.

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+
+If you have to merge records with fields in commons, you will use two new concepts:
+
+- default annotation,
+- overwriting,
```
aspiwack, line 24: Are these two concepts or just one?

There are other merging cases:
- merging two record fields.
- merging a field against a field which merely has a contract attached (this emulate mix-ins in particular).

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Default annotation
+
+To be able to merge records with commons fields, at least one side field has to
+be annotated as default. This means, if you have two records:
```
aspiwack, line 29: As I was saying, this isn't the case.

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+  firewall.open_ports = [21, 80, 443],
+} in
+let right = {
+  firewall.enabled = false,
+  server.host.options = "TLS",
+} in
+left & right
+```
+
+Like it is, it's impossible to merge and will throw an unmergeable terms error.
```
aspiwack, line 44: Explain why.

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+The default annotation is generaly to give a default value to a record field.
+So, this value can be changed afterward. Saying it in a more abstract way,
+default indicate a lower priority to a field in case of merging. Saying that,
+If both sides have been annotated `default`, the merge is not possible.
+
+## Overwriting
+
+The overwriting is the concept specifying the behaviour of a merge when you
+overwrite a field on which depends an other one. This  feature is described in
+["#573 [RFC] Merge types and terms syntax proposal"](https://github.com/tweag/nickel/pull/573)
```
aspiwack, line 71: Don't link to pull requests in the documentation. Don't link to merged RFCs either (they are internal documentation).

Well, let me temper that. You can link to them as further reading for the curious reader (links are good!). But they can't be the introduction, and the understanding of the documentation can't depend on reading them.

You are allowed to copy large bits of them though!

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+} in // security => {firewall.open_ports = [21, 80, 443]
+security & {firewall.open_proto.ftp = false} // => {firewall.open_ports = [80, 443]
+```
+
+We then see that dependent fields are updated when you overwrite the fields they
+depend on.
+
+## A word about contracts
+
+To simplify, when merging two records, all contracts of both left and right one
```
aspiwack, line 96: Why do you want to “simplify”?

doc/manual/merging.md
```diff
@@ -0,0 +1,117 @@
...
+are applied to the resulting one.
+For instence:
```
aspiwack, line 98:
```suggestion
For instance:
```

aspiwack (commented):
There is still a lot of work to do to make this document useful.

It is good that it has a lot of examples (though I point at one place where documentation is missing).

I would order the document, probably in the following order
1. Merging records with no common fields
2. Merging records when the common fields are themselves record
3. Explain what is actually stored in a field, demonstrate by showing how the docstring is preserved by merging
4. Merge “pure contract” with a value (the case of mixins)
5. Default values.

francois-caddet:
> There is still a lot of work to do to make this document useful.
> 
> It is good that it has a lot of examples (though I point at one place where documentation is missing).

Thanks for your usefull review.

> I would order the document, probably in the following order
> 
> 1. Merging records with no common fields
> 2. Merging records when the common fields are themselves record
> 3. Explain what is actually stored in a field, demonstrate by showing how the docstring is preserved by merging
> 4. Merge “pure contract” with a value (the case of mixins)
> 5. Default values.
I reworked this page as you proposed except for the 3rd point.
What is stored in a field would probably have been described before in the Syntax page and in Contract one also. So I was thinking of having more a part explaining how behave the merging in case of contracts and annotated fields.
Right now it's in the last part and realy lake details and cases. Also it could probably be moved uper in the doc.
Do you think I'm going on right way?

yannham (commented):
The content is getting there, but the wording, not yet. Some paragraphs don't flow very well. I would suggest to make a pass and:
- try to be straight-to-the point stylistically (avoid useless adverbs, vague references to previous points made, passive forms, etc.)
- try to think, for each section/paragraph/blob/whatever, what is the key information to convey. Usually there is a motivation, then a high-level explanation of the behavior of the feature, then probably a more detailed explanation of the behavior (we are in a manual section, after all), and finally, examples.
- cut everything that is vague, a bit out of context, doesn't add useful information, etc. In general, anything that doesn't serve the purpose you established at the previous step.

doc/manual/merging.md
```diff
@@ -0,0 +1,258 @@
...
+} & {
+    port | #Port = 80,
+} // blame because 80 < 1024
+```
+
+In the case the second record would contains `port=8888` it does not have blame.
+
+TODO:
+
+- case of top level contracts,
```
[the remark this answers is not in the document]
francois-caddet: When the contract is attached to the full record:

```text
let x = {a = 1} | SomeContract in
let y = {b = 2} | SomeOthers in
x & y
```

If I'm not mistaking, contracts are simply merged no?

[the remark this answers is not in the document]
yannham: Merging contracts is not well defined in the general case (`SomeContract` and `SomeOthers` could very well be custom contracts defined as functions, for example). So no, contracts are not merged. In fact, merging doesn't know if it is "at top-level" or not, so the semantics has to be the same for recursive merging of fields and for "top-level" merges, like merging two variables.

yannham:
I reworked the section, re-organizing the content and rewording the text, in the hope of improve it. @francois-caddet could you make a pass on it and tell me what you think please? Thanks

doc/manual/merging.md
```diff
@@ -0,0 +1,522 @@
...
+Merging two records with no common fields results in a record with the fields
+from both operands. That is, `{foo = 1, bar = "bar"} & {baz =false}` evaluates
+to `{foo = 1, bar = "bar", baz = false}`.
+
+### Specification
+
+Technically, if we write the left operand as:
+
+```
+letf = {
```
thufschmitt, line 42:
```suggestion
left = {
```

(there’s a couple extra instances of `left`, probably because of copy-pasting. You might want to `s/letf/left` directly)

doc/manual/merging.md
```diff
@@ -0,0 +1,522 @@
...
+= {
+  common = left          if p(left) > p(right)
+           right         if p(left) < p(right)
+           left & right  if p(left) = p(right)
+}
+```
+
+Currently, there are only two priorities, `normal` (by default, when nothing is
+specified) and the `default` one, with `default < normal`. We plan to add more
+in the future (see [RFC001][rfc001]).
```
thufschmitt, line 281: Might be nicer to directly link to the relevant section

```suggestion
in the future (see [RFC001](https://github.com/tweag/nickel/blob/c21cf280dc610821fceed4c2caafedb60ce7177c/rfcs/001-overriding.md#priorities)).
```

doc/manual/merging.md
```diff
@@ -0,0 +1,522 @@
...
+The section describes the behavior and use-cases of merge, by considering the
+following situations:
+
+- Merging two records without common fields
+- Merging records with common fields
+- Merging records with metadata
+  * Default values
+  * Contracts
+  * Documentation
+- Recursive overriding
```
thufschmitt, line 29: Could these link to the relevant sections (and maybe make it more explicit that it’s a toc, I didn’t notice it at first)?

doc/manual/merging.md
```diff
@@ -0,0 +1,522 @@
...
+We've seen in the section on default values that they are useful to override
+(update) a single field with a different value. The combo of merging and default
+values can do more. In Nickel, records are recursive by default, in order to
+express easily dependencies between the different fields of the configuration.
+Concretely, you can refer to other fields of a record from within this record:
+
+```nickel
+let base_config = {
+  version | default = "20.09",
+  input.url | default = url = "nixpkgs/nixos-%{version}",
```
thufschmitt, line 472: Is that really syntactically correct?

thufschmitt (commented):
A few suggestions, but it reads OK overall (though I have a very hard time judging that properly in general, so someone else might want to differ on that).

The main thing I’m missing is a sentence or paragraph at the beginning that would handwave a unified semantics to merging as right now there’s several semantics depending on the shape of the record, but nothing that explains how they actually make sense together. Maybe something like

> Merge is an operation that combines two records. Fields common to both records will be recursively merged if possible using the semantics described below

doc/manual/merging.md
yannham: Nope, it's inconsistent consecutive editions, good catch.

doc/manual/merging.md
```diff
@@ -0,0 +1,526 @@
...
+  and the final result would have an additional field `other_subfield` ?
+
+Nickel chooses to answer **no** to both. In general, when a contract is attached
+to a field `foo`, merging ensures that whatever is this field merged with,
+including being dropped in favor of another value, the final value for `foo` has
+to respect the contract as well or the evaluation will fail accordingly.
+
+#### Specification
+
+For two operands with one field each, which the same on both side, with respective
```
yannham, line 379:
```suggestion
For two operands with one field each, which is the same on both side, with respective
```

doc/manual/merging.md
```diff
@@ -0,0 +1,526 @@
...
+& {version = "unstable"}` will evaluate to:
+
+```nickel
+{
+  version = "unstable",
+  input = {url = "nixpkgs/nixos-unstable"},
+}
+```
+
+Currently, one can only use it on a field that have been marked as default. A
```
yannham, line 503:
```suggestion
Currently, one can only override a field that has been marked as default beforehand. A
```

Acaccia (approved):

doc/manual/merging.md
```diff
@@ -0,0 +1,526 @@
...
+  top_right = 2,
+  common = {left = "left", right = "right"}
+}
+```
+
+When one or both of the common fields are not records, the merge will fail
+unless one of the following condition hold:
+
+- They are both of a primitive data type `Num`, `Bool`, `Enum`, or they are null, and they are equal
+- They are physically equal, meaning that they point to the same location in
```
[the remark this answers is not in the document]
yannham: To be clear, this pointer equality stuff is not implemented yet (but very soon). On one side, I can see why relying on this may be fragile on the user side. It means that `{x = ["foo"]} & {x = ["foo"]}` doesn't merge, but `let y = ["foo"] in {x = y} & {x = y}` does. On the other hand, I guess as long as you under approximate real equality, it's still better than rejecting everything. In particular in Nickel, where I had problems when applying the same contract several times because any default value that is not primitive (a list, a function, etc.) wouldn't merge with itself). Also it probably helps performance because you just shortcut merging big records if they are actually the same.

But @edolstra may have a point: maybe what we want is to do it nevertheless, to make multiple contract application to never fail, but not document it in the specification, to avoid having people relying on this, and pretending that only equal primitive values can merge.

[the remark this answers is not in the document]
Infinisil: > It means that `{x = ["foo"]} & {x = ["foo"]}` doesn't merge, but `let y = ["foo"] in {x = y} & {x = y}` does.

Oh, yeah that actually sounds pretty bad, breaking referential transparency..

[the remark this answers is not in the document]
yannham: So, I just tried to implement this, and recursive records/revertible thunks make the thing even more tricky (in that it doesn't short-cut even when sub-element of the record are the same, which was my original motivation). So let us just get rid of that for now totally, both in the documentation and in the implementation :slightly_smiling_face:

PR #582, opened 2022-01-19
`flake.nix` improvements

garbas:
Each commit in this PR represents an improvement for `flake.nix`. If needed I can submit each commit separately if that would make things easier to review.

Things that were improved in this PR:

- Use overlay to pin wasm-bindgen-cli (no need to depend on older nixpkgs)
- Simplify flake.nix with flake-utils
- Switching from nixpkgs-mozilla overlay to rust-overlay
- Replace custom git hooks with pre-commit-hooks.nix and also add format check for `*.nix` files

This all work was done in preparation to make it easier (possible) produce statically linked binaries, which I hope is the next PR :)


lsp/client-extension/node-env.nix
```diff
@@ -1,6 +1,6 @@
 # This file originates from node2nix
```
yannham, line 1: Was this file automatically re-generated, or did you manual edit it and I need to review it too?

yannham (approved):
I haven't reviewed the node stuff, because I suspect it has been automatically generated, but otherwise looks great! Feels good to get rid of those `channels.toml`, the custom precommits and the somehow hacky bindgen pinning :slightly_smiling_face:

lsp/client-extension/node-env.nix
garbas: This was all formatted by `nixpkgs-fmt`. But I don't think we should do this for generated nix files such as this. I'll add it to the ignore list and reverted the changes.

yannham:
@garbas you probably need to re-delete `channel_beta.toml`, but otherwise you can merge.

Issue #583, opened 2022-01-19
Cannot use in a contract function a key from the same record as argument.


Acaccia:
**Describe the bug**
An unbound identifier error occur when using an identifier from the same record in a contract. 
Here is a minimal example:
```
let Name = fun b label name => 
    if b then name else contracts.blame label in

let Config = {
    b | Bool,
    name | #(Name b),
} in

{
    b = true,
    name = "name",
} | #Config
```
which gives the error: 
```
error: Unbound identifier
  ┌─ bug.ncl:6:19
  │
6 │     name | #(Name b),
  │                   ^ this identifier is unbound

```

**To Reproduce**
Check the minimal example in description.

**Expected behavior**
No unbound error, this should work.

**Environment**
- Rust version: 1.58.0 
 - Version of the code: commit 5a2fb2446e78810482ab1260da77962dbe416337

**Additional context**
Could be linked to issue #579

Acaccia:
@yannham nope, it was a mistake, the initial example did not work.

Issue #579, opened 2022-01-18
"Unbound identifier" when overriding nested records


yannham:
**Describe the bug**
It seems overriding nested records is doing something wrong, as it throws undue `Unbound identifier` errors.

**To Reproduce**

```nickel
>({foo.bar = foo.baz, foo.baz | default = 2} & {foo.baz = 1})
error: Unbound identifier
  ┌─ repl-input-7:1:13
  │
1 │ ({foo.bar = foo.baz, foo.baz | default = 2} & {foo.baz = 1})
  │             ^^^ this identifier is unbound
```

This requires `foo` to be overridden on the right to trigger. If we don't do a merge, or just merge with different fields, everything is working fine:

```nickel
>{foo = {bar = foo.baz}, foo = {baz | default = 2}}.foo.bar
2

>({foo.bar = foo.baz, foo.baz | default = 2} & {zorg = 2}).foo.bar
2
```

**Expected behavior**
We expect the first example to run without error:
```nickel
>({foo.bar = foo.baz, foo.baz | default = 2} & {foo.baz = 1}).foo.bar
2
```

PR #587, opened 2022-01-20
Fix recursive overriding when merging records with intersecting fields

yannham:
Fix both #579 and #583. When introducing recursive overriding, what we did when merging was to revert the thunks of the two records being merged, build a new recursive record `RecRecord` with the result of the merging, and let `eval` compute the fixpoint of this record at the next tick. 

To compute the fixpoint, `eval` builds the so-called recursive environment with the content of each field, and then go through each field again in order to extend their environment with this recursive environment. For this to work, we were careful to maintain the invariant that the content of each field of a recursive record was a closure (stored as a variable) whose content was _immediately_ the original expression, without any additional indirection.

This works as long as we are not actually recursively merging fields: that is, when the fields of the two operands of `&` are disjoints like in `{foo = 1} & {bar = 2}`, or when the common fields have different priorities `{foo | default = 1, bar = foo} & {foo = 2}`. There, each field of the end record is directly taken from one of the operand.

However, this isn't so true anymore when we are actually recursively merging fields, as in `{foo | default = 1, m={baz=foo}} & {foo = 2, m={bar=foo}}`. Here, `merge` will closurize the `m` coming from both operands (let's call the result `%m1` and `%m2`), and set `m = %m1 & %m2` in the final result. This breaks the invariant, as `%m1 & %m2` is not an unevaluated expression coming from a recursive record anymore. `eval` would then patch `m` with the recursive environment. Alas, with no effect: `m` doesn't use any free variable from the recursive environment. What we need is to actually patch the environment of `%m1` and `%m2`, not `m`.

In this PR, we:
- extract the record fixpoint computation code from `eval` to a submodule, to avoid duplicating the logic in `eval` and `merge`
- in `merge`, we keep a list of `Rc`s to the fields of both operands _before_ we merge them. We then build the final record, build the final recursive environment, and can eventually patch the environment of the original fields as wanted.

### DIFF
--- a/src/eval/fixpoint.rs
+++ b/src/eval/fixpoint.rs
@@ -1,0 +1,54 @@
+//! Helper to compute the fixpoint of a recursive record.
+use super::*;
+
+/// Build a recursive environment from an iterator of bindings. For each binding, extract the thunk
+/// from the passed environment, and return it in the result. The resulting environment can then be
+/// passed to the [`patch_field`] function.
+pub fn rec_env<'a, I: Iterator<Item = (&'a Ident, &'a RichTerm)>>(
+    bindings: I,
+    env: &Environment,
+) -> Result<Vec<(Ident, Thunk)>, EvalError> {
+    bindings
+        .map(|(id, rt)| match rt.as_ref() {
+            Term::Var(ref var_id) => {
+                let thunk = env
+                    .get(var_id)
+                    .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
+                Ok((id.clone(), thunk))
+            }
+            _ => {
+                // If we are in this branch, the term must be a constant after the
+                // share normal form transformation, hence it should not need an
+                // environment, which is why it is dropped.
+                let closure = Closure {
+                    body: rt.clone(),
+                    env: Environment::new(),
+                };
+                Ok((id.clone(), Thunk::new(closure, IdentKind::Let)))
+            }
+        })
+        .collect()
+}
+
+fn patch_thunk(mut thunk: Thunk, rec_env: &Vec<(Ident, Thunk)>) {
+    thunk.borrow_mut().env.extend(rec_env.iter().cloned());
+}
+
+/// Take a [`RichTerm`] that is stored inside a field, and patch its environment by enriching it
+/// with the recursive environment.
+pub fn patch_field(
+    rt: &RichTerm,
+    rec_env: &Vec<(Ident, Thunk)>,
+    env: &Environment,
+) -> Result<(), EvalError> {
+    if let Term::Var(var_id) = &*rt.term {
+        let thunk = env
+            .get(var_id)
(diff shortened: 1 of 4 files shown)


src/eval/merge.rs
```diff
@@ -329,8 +329,17 @@ pub fn merge(
              * the same trick as in the evaluation of the operator DynExtend, and replace each such
              * term by a variable bound to an appropriate closure in the environment
              */
+            // Merging recursive record is the one operation that may overrides recursive fields.
```
yannham, line 332:
```suggestion
            // Merging recursive record is the one operation that may override recursive fields.
```

src/eval/merge.rs
```diff
@@ -329,8 +329,18 @@ pub fn merge(
...
              */
+            // Merging recursive record is the one operation that may override recursive fields. To
+            // have the recursive fields depend on the updated values, we need to revert the thunks
+            // first.
             rev_thunks(m1.values_mut(), &mut env1);
             rev_thunks(m2.values_mut(), &mut env2);
+
+            // We save the original fields before they are potentially merged in order to patch
+            // their environment in the final record. Note that we are only cloning shared terms
+            // (`Rc`s) here.
```
aspiwack, line 340: What does it mean to “patch their environments”?

src/eval/fixpoint.rs
```diff
@@ -0,0 +1,51 @@
...
+    bindings
+        .map(|(id, rt)| match rt.as_ref() {
+            Term::Var(ref var_id) => {
+                let thunk = env
+                    .get(var_id)
+                    .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
+                Ok((id.clone(), thunk))
+            }
+            _ => {
+                // If we are in this branch, the term must be a constant after the
```
aspiwack, line 21: Is “the term” `rt`? If so, say so.

src/eval/fixpoint.rs
```diff
@@ -0,0 +1,51 @@
...
+                    body: rt.clone(),
+                    env: Environment::new(),
+                };
+                Ok((id.clone(), Thunk::new(closure, IdentKind::Let)))
+            }
+        })
+        .collect()
+}
+
+/// Patch the environment of the content of a recursive record field by extending it with a
```
aspiwack, line 34: I don't know what “Patch” means here either.

aspiwack (commented):
You lost me with the patching business. What are we doing, here, really?

yannham:
"Patching" is "tying the knot". When we are evaluating a recursive, we first allocate thunks for each one of the fields in a standard (i.e. non-recursive, as a normal record) way. Then we go over the said thunks and mutate their environment to add precisely those thunks we have just allocated, which is where the recursion is really introduced. This second step of mutating environments of thunks after the fact is what is called patching.

aspiwack (approved):
Ok, I get it, mostly.

This code is getting pretty complex. We really want to sit down and figure out a more principled strategy.

PR #543, opened 2022-01-04
Use free variables to avoid memory cycles in rec records

Avi-D-coder:
cherry pick of https://github.com/tweag/nickel/pull/529/commits/836e01d17a5f8bdeb568b5489f66ad021cea79f7

PR #588, opened 2022-01-20
Use free variables to avoid memory cycles in rec records (take 2)

Avi-D-coder:
Replaces https://github.com/tweag/nickel/pull/543.
This rework reduces the ugliness and improves the performance.
It's analogous to the [second](https://github.com/tweag/nickel/pull/543#issuecomment-1013229170) approach, but at the term level. It can be made prettier, by the refactor I described there in the future.


src/eval/mod.rs
```diff
@@ -567,12 +577,24 @@ where
...
-
-                                    thunk.borrow_mut().env.extend(
-                                        rec_env
-                                            .iter_elems()
-                                            .map(|(id, thunk)| (id.clone(), thunk.clone())),
-                                    );
+                                    let mut clos = thunk.borrow_mut();
+                                    if let Some(Some(free_vars)) = free_vars
+                                        .as_ref()
+                                        .map(|fr| fr.get(&Ident::from(String::new())))
```
yannham, line 583: Nitpick: you can avoid the double `Some` by using `and_then` (aka flatmap) instead of `map`.

yannham (changes requested):
Not a final solution, but this approach looks way better and is fine for the time being.

src/eval/mod.rs
```diff
@@ -524,18 +524,28 @@ where
...
                             // We already checked for unbound identifier in the previous fold,
                             // so function should always succeed
                             let mut thunk = env.get(var_id).unwrap();
-                            thunk.borrow_mut().env.extend(
-                                rec_env
-                                    .iter_elems()
-                                    .map(|(id, thunk)| (id.clone(), thunk.clone())),
-                            );
+                            let mut clos = thunk.borrow_mut();
+                            if let Some(Some(free_vars)) = free_vars.as_ref().map(|fr| fr.get(id)) {
```
yannham, line 535: See the other similar comment about `and_then` below.

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,62 @@
...
+            ffv.as_mut()
+                .unwrap()
+                .insert(Ident::from(String::new()), dyn_fields_free_vars);
+
+            map.iter().for_each(|(id, _)| {
+                free_vars.remove(id);
+            });
+        }
+        Term::Switch(..) => {
+            // TODO switch should be revisted.
```
yannham, line 53: Why? The free variables of a switch is just the combination of the free variables of all cases + the default one. Currently switch are not destructured, and thus don't bind any variable.

src/transform/mod.rs
```diff
@@ -34,12 +39,14 @@ pub fn transform(rt: RichTerm) -> RichTerm {
...
         .unwrap();
 
-    rt.traverse(
-        &mut |rt: RichTerm, _| -> Result<RichTerm, ()> {
-            let rt = share_normal_form::transform_one(rt);
+    rt.traverse_with_parent(
+        false,
+        &mut |rt: RichTerm, parent, (free_vars, field_free_vars)| -> Result<RichTerm, ()> {
+            let mut rt = share_normal_form::transform_one(rt);
+            collect_free_vars(&mut rt, parent, free_vars, field_free_vars);
```
yannham, line 46: We probably want to do the transformation before `share_normal_form`, just because the latter introduces a bunch of let-binding and thus make the free variables calculation slower (while they don't change the free variables of recursive record fields).

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        _ => {}
```
yannham, line 56: You forgot `FunPattern` and `Fun`, which also removes free variables from the set i.e. `free_vars(fun x => body) = free_vars(body) - {x}`

src/transform/collect_free_vars.rs
Avi-D-coder: Destructuring is what I was wondering about.

src/term.rs
```diff
@@ -93,6 +96,11 @@ pub enum Term {
         HashMap<Ident, RichTerm>,
         Vec<(RichTerm, RichTerm)>, /* field whose name is defined by interpolation */
         RecordAttrs,
+        // Cached free variables calculation
+        // An empty `Ident` is used for all dynamic fields.
```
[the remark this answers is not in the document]
Avi-D-coder: So don't track dyn fields free variables set?
I considered trying to name them by position, but I'm not sure there ordering is stable.

If we don't need to track them then we don't need `VecDeque`.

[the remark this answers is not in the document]
yannham: > So don't track dyn fields free variables set?

To first approximation, I would say the opposite: at this point, that's not much harder to track them like any statically named field, and that's the more logical thing to do. Indexing by position is fine, as not only the order should be stable across all the transformations the AST go through, but here we are moreover talking about stability between two call to `traverse` on consecutive nodes. You could have a separate `Vec` for that in the field of `RecRecord`, say that would now be `(HashMap<_, _>, Vec<_>)`. Putting everything in the hashmap might be trickier than it first appears to, because you can define arbitrary record name `{"__%1" = "mystring"}`, so you have to handle conflicts with your naming schema.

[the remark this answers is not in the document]
yannham: Sorry, that's wrong, we indeed need the order of dynamic fields to be stable between the free vars transformation and the evaluation. This is currently the case, I believe, and I can't see right now a good reason this wouldn't. But it's true that it's a new invariant to document and to maintain.

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,62 @@
...
+) {
+    match SharedTerm::make_mut(&mut rt.term) {
+        Term::Var(id) => {
+            free_vars.insert(id.clone());
+        }
+        Term::Let(id, _, _, _) | Term::Fun(id, _) => {
+            free_vars.remove(id);
+        }
+        Term::LetPattern(id, _, _, _) => {
+            // We can ignore the `Destruct`, since all values lhs are in the rhs.
```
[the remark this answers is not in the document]
Avi-D-coder: I was wrong about this. It's only true in the top down version.

src/transform/mod.rs
Avi-D-coder: Switching their order produces an error.
Not sure why, any guesses?

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,62 @@
...
+            id.as_ref().map(|id| free_vars.remove(id));
+        }
+
+        Term::RecRecord(map, _dyn_fields, _, ffv) => {
+            *ffv = Some(
+                map.iter()
+                    .map(|(id, _)| (id.clone(), fields_free_vars.pop_front().unwrap()))
```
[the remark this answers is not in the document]
Avi-D-coder: For some reason I thought drain didn't work here, but your right it does, I must have misread the docs.

src/term.rs
```diff
@@ -91,8 +93,14 @@ pub enum Term {
...
     RecRecord(
         HashMap<Ident, RichTerm>,
-        Vec<(RichTerm, RichTerm)>, /* field whose name is defined by interpolation */
+        Vec<(RichTerm, RichTerm, Option<HashSet<Ident>>)>, /* field whose name is defined by interpolation */
         RecordAttrs,
+        // Cached free variables calculation
+        // An empty `Ident` is used for all dynamic fields.
+        // It would be nice to use a `None`, but that does not play well with the borrow checker.
+        // TODO When RecRecord is refactored store a fields free var set next to it's `RichTerm`.
+        // TODO ignore free vars in impl PartialEq, and Ord.
```
yannham, line 102: Those comments probably need to be updated.

src/eval/mod.rs
```diff
@@ -498,11 +498,20 @@ where
...
             }
-            Term::RecRecord(ts, dyn_fields, attrs) => {
+            Term::RecRecord(ts, dyn_fields, attrs, free_vars) => {
                 let rec_env = fixpoint::rec_env(ts.iter(), &env)?;
 
-                ts.iter()
-                    .try_for_each(|(_, rt)| fixpoint::patch_field(rt, &rec_env, &env))?;
+                let filter_env = |(id, _): &(Ident, Thunk)| -> bool {
+                    if let Some(free_vars) = free_vars.as_ref().and_then(|fr| fr.get(id)) {
+                        free_vars.contains(id)
```
yannham, line 506: I don't think this is what we want. Here the logic of should we add field `id` to the environment of field `foo` doesn't depend on the field `foo` being patched, which is fishy.

You should have two `id`s: the one of the field we patch, and the one we are currently filtering on. That is, the filter function should be different for each field:

```
let filter_env = |field: Ident| {
  if let Some(free_vars) = free_vars.as_ref() {
    |(id, _)| free_vars.contains(id);
  }
  else {
    |_| true
  }
};
...
ts.iter().try_for_each(|(id, rt)| {
    fixpoint::patch_field(rt, &rec_env, &env, filter_env(id))
})?;
```

(May need type annotations or some `fn` instead of closures with additional parameters, but you get the idea)

yannham: This may or may not explain why my counter-example below `{x = 1, let y = let x = x + 1 in x}` still passes.

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,80 @@
...
+// `fields_free_vars` is a mapping between position of a field in a RecRecord`,
+//  and the free variables used in the body of the field.
+pub fn collect_free_vars(
+    rt: &mut RichTerm,
+    parent_rec_record: TermType,
+    free_vars: &mut HashSet<Ident>,
+    fields_free_vars: &mut Vec<HashSet<Ident>>,
+) {
+    if parent_rec_record == TermType::RecRecord {
+        fields_free_vars.push(free_vars.clone());
```
yannham, line 24: Doesn't this mean that you keep accumulating the free variables of the previous fields? Like in:

```
{x = y, foo = 1}
```

Assuming you iterate on `foo` in second, I think you will have `y` as free variable of `foo`, unduly. I suspect you need to set `*free_vars = HashSet::new()` here. It does mean that to compute the free vars of a recursive record, you need to merge all the free vars into one again. You could have yet another variable to keep track of the full set of free variables, but I think it's premature optimization, and it's fine to have to merge hashsets on a recursive record.

yannham (commented):
I'm not sure about the handling of let-binding, see the corresponding comment. In general there are a number of places where the logic doesn't look exactly right, but it's not easy to see the effects, because as long as you over-approximate the free vars, or if you happen not to set the set of free variables to `Some`, evaluation won't break. It may be useful to add a few unit tests and `debug_assert!` to ensure basic invariants and that it's working as expected on various base cases

The handling of the free variables of dynamic fields is slightly unfortunate. I think you added the free variables of static fields as a separate field in order to avoid disrupting the code that operates on the content of records, but did the opposite for the free variables of dynamic fields, and in the end it feels like having disadvantages of both. I get that the advantage is to avoid relying on an additional invariant about ordering. FWIW, I think it's fine to require that the order of dynamic fields doesn't change between during transformations, as long as it is documented, and put all the free variables in one dedicated field of the `RecRecord` variant, something like `RecordFreeVars = {static: HashMap<Ident, HashSet<Ident>>, dynamic: Vec<HashSet<Ident>>}`. But do as you think is best.

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,80 @@
...
+        Term::Let(id, _, _, _) | Term::Fun(id, _) => {
+            free_vars.remove(id);
+        }
+        Term::LetPattern(id, d, _, _) | Term::FunPattern(id, d, _) => {
+            id.as_ref().map(|id| free_vars.remove(id));
+            remove_destruct(d, free_vars);
+        }
+
+        Term::RecRecord(map, dyn_fields, _, ffv) => {
+            for f in dyn_fields.iter_mut() {
```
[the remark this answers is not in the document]
Avi-D-coder: Good catch, I forgot to remove that one.

src/transform/collect_free_vars.rs
```diff
@@ -0,0 +1,80 @@
...
+) {
+    if parent_rec_record == TermType::RecRecord {
+        fields_free_vars.push(free_vars.clone());
+    }
+
+    match SharedTerm::make_mut(&mut rt.term) {
+        Term::Var(id) => {
+            free_vars.insert(id.clone());
+        }
+        Term::Let(id, _, _, _) | Term::Fun(id, _) => {
```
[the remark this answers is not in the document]
Avi-D-coder: I thought it worked, because above the match on line 2, we store the set of free vars.
Since we store the set to the `RecRecord` before removing the binding.

```
    if parent_rec_record == TermType::RecRecord {
        fields_free_vars.push(free_vars.clone());
    }
```

Looking at it now I would expect it to fail with one more level of nesting, but it doesn't. 🤔
```
{
  x = 2,
  y = let z = 1 in let x = x + z in x,
}
```

I need to recheck my assumptions, and add tests.

Avi-D-coder:
This approach is not viable, fixing the aforementioned issues replicates the previous ugly code.

yannham:
@Avi-D-coder could you explain what in particular is causing trkoublres? The logic was off but it didn't seem to be fundamentally a bad idea, requiring maybe one additional `ParentType` (or whatever it is called).

yannham:
I would have thought the removing `x` from the set of free variables is not the job of the children nodes (`x` and `x+1`), but the job of the parent node `let`. At this point, it should know that the first element on the stack inside the state is the left node and thus doesn't need any removal, while the second one should indeed be patched. The children nodes just have to know that their parent node is special, a bit like for record fields, and thus that they need to save their free variables inside the state separately, instead of sharing just one `HashSet`. This would require just an additional `ParentNode::Let` as new machinery. Or am I missing something here?

Avi-D-coder:
I believe your correct, we could transfer positional info that way. I'll see if that works.

Avi-D-coder:
It intuitively feels like it should work, but it does not. The child must produce the free var set to avoid mixing state between branches. It's the same problem I hit with version 1.

yannham:
I haven't got the time to look at it, but I still don't get it. If the traversal is depth first, and you know your direct parent, children nodes should be able to take the required disposition: save the current free variables (which should correspond to the current branch under the let, either left or right) in the state, and allocate a fresh working set. Then, it will recurse in the second branch, which would do the same.

When coming back to the let, we should have both free vars sets pushed on a stack or a tuple or whatever you chose inside the state, separated. Then you can pop the second one, remove the let-bound variable, merge the two, clean the stack/tuple/state/foo, and return the result.

Is there some hidden issue? What didn't work exactly?

Avi-D-coder:
@yannham if we're saving state prior to each traversal branch we are reimplementing something equivalent to the monoid like behavior of the previous version. Using a stack to to push our let state to would be an option, but the calculation would become even more dependent on terms not being added or removed by other transformations in the same bottom up pass `(share_normal_form)`. Since, even the more resilient monoid like free vars pass was not associative with `share_normal_form` my assumption is a stack based version will break.

If I have time today or on my own time in the next couple weeks, I'll try and implement this as a separate adhoc pass. It should not take long to do when not incorporating it into this bottom up pass. Then maybe a method of merging it with this bottom up pass will be more apparent.

yannham:
> @yannham if we're saving state prior to each traversal branch we are reimplementing something equivalent to the monoid like behavior of the previous version

I think it's fine to do it for let-binding. We already have to do it for records. It's still far from the naive monoid approach: for something like `let x = (fun y => x + y + z + (foo x y z)) in x (10 + (if 5 == z then 10 else 20)))`, it will only have to do one split and one merge of hashsets, while the naive monoid approach would generate and merge like the number of node of the AST (at a quick glance, at least 20~30).

If you're really, really worried about that, you can still pass a `ParentType::LeftLet(id)` to let the left child, that could then store a boolean indicating if it had `id` as a free variable in the state (thus before recursing in the right child). Then if the boolean is true, the parent let doesn't remove anything.

But I'm not sure this is worth the trouble. After all, we can encode arbitrary nested function calls in the state, so we can probably simulate pretty much any traversal we want, but that doesn't mean we should. What I like about this approach is that it doesn't look too complex, doesn't disrupt the existing code too much, and eliminates all of undue memory cycles and thus memory leaks that are not actually mutually recursive values. I think we can afford a small non-optimality with respect to let-bindings.

On the other hand, maybe the "efficient" solution is less complex. In that case, I would go for it.

PR #589, opened 2022-01-21
[RFC002] Unified syntax #1: one record to rule them all

yannham:
First step of the implementation of RFC002 (#589). This PR adds the `UniTerm` syntax for records, which should be the most involved part. The plan is to avoid introducing another duplicate AST for the `UniTerm` syntax in general, and just convert on-the-fly to `Types` or `Terms` when parsing. Records and records types, as noted in the RFC, should be the only intersecting part of the union of the previous syntaxes, and thus do need a bit more machinery.

The most interesting changes are in `parser/uniterm.rs` and `grammar.lalrpop`. Other changes include fixing the syntax for row tails in tests and examples, and propagating unbound variables errors at different stage of the interpreter pipeline. Indeed, we could previously afford to check at parsing time if type variables were unbound, and then always assume there were in the rest of the code, but this isn't the case anymore with the `UniTerm` syntax.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -291,6 +291,8 @@ pub enum ParseError {
     ),
     /// Unbound type variable
     UnboundTypeVariables(Vec<Ident>, RawSpan),
+    /// TODO
+    InvalidUniRecord(),
 }
 
 /// An error occurring during the resolution of an import.
@@ -444,6 +446,7 @@ impl ParseError {
                 InternalParseError::UnboundTypeVariables(idents, span) => {
                     ParseError::UnboundTypeVariables(idents, span)
                 }
+                InternalParseError::InvalidUniRecord(..) => ParseError::InvalidUniRecord(),
             },
         }
     }
@@ -1195,6 +1198,10 @@ impl ToDiagnostic<FileId> for ParseError {
                         .join(",")
                 ))
                 .with_labels(vec![primary(span)]),
+            ParseError::InvalidUniRecord(..) => {
+                Diagnostic::error().with_message(format!("invalid record literal"))
+            }
+            // .with_labels(vec![primary(span)]),
         };
 
         vec![diagnostic]
--- a/src/error.rs
+++ b/src/error.rs
@@ -1219,6 +1219,10 @@ impl ToDiagnostic<FileId> for ParseError {
                     primary(span),
                     secondary(illegal_span).with_message("can't use this record construct"),
                     secondary(tail_span).with_message("in presence of a tail"),
+                ])
+                .with_notes(vec![
+                    String::from("Using a polymorphic tail in a record `{ ..; a}` requires the rest of the record to be only composed of type annotations, of the form `<field>: <type>`."),
+                    String::from("Value assignements, such as `<field> = <expr>`, metadata, etc. are forbidden."),
                 ]),
         };
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -1200,8 +1200,7 @@ impl ToDiagnostic<FileId> for ParseError {
                 .with_labels(vec![primary(span)]),
             ParseError::InvalidUniRecord(..) => {
                 Diagnostic::error().with_message(format!("invalid record literal"))
-            }
-            // .with_labels(vec![primary(span)]),
+            } // .with_labels(vec![primary(span)]),
         };
 
         vec![diagnostic]
--- a/src/error.rs
+++ b/src/error.rs
@@ -291,10 +291,18 @@ pub enum ParseError {
     ),
     /// Unbound type variable
     UnboundTypeVariables(Vec<Ident>, RawSpan),
-    /// Illegal record literal in the uniterm syntax (see RFC002). In practice, this is a record
-    /// mixing a tail and non-type constructs.
+    /// Illegal record literal in the uniterm syntax. In practice, this is a record with a
+    /// polymorphic tail that contains a construct that wasn't permitted inside a record type in
+    /// the original syntax. Typically, a field assignment:
+    ///
+    /// ```nickel
+    /// forall a. {foo : Num; a} // allowed
+    /// forall a. {foo : Num = 1; a} // InvalidUniRecord error: giving a value to foo is forbidden
+    /// ```
+    ///
+    /// See [RFC002](../../rfcs/002-merge-types-terms-syntax.md) for more details.
     InvalidUniRecord(
-        RawSpan, /* non-type construct position */
+        RawSpan, /* illegal (in conjunction with a tail) construct position */
         RawSpan, /* tail position */
         RawSpan, /* whole record position */
     ),
@@ -1209,7 +1217,7 @@ impl ToDiagnostic<FileId> for ParseError {
                 .with_message(format!("invalid record literal"))
                 .with_labels(vec![
                     primary(span),
-                    secondary(illegal_span).with_message("can't use non-type construct"),
+                    secondary(illegal_span).with_message("can't use this record construct"),
                     secondary(tail_span).with_message("in presence of a tail"),
                 ]),
         };
--- a/src/error.rs
+++ b/src/error.rs
@@ -291,8 +291,13 @@ pub enum ParseError {
     ),
     /// Unbound type variable
     UnboundTypeVariables(Vec<Ident>, RawSpan),
-    /// TODO
-    InvalidUniRecord(),
+    /// Illegal record literal in the uniterm syntax (see RFC002). In practice, this is a record
+    /// mixing a tail and non-type constructs.
+    InvalidUniRecord(
+        RawSpan, /* non-type construct position */
+        RawSpan, /* tail position */
+        RawSpan, /* whole record position */
+    ),
 }
 
 /// An error occurring during the resolution of an import.
@@ -446,7 +451,9 @@ impl ParseError {
                 InternalParseError::UnboundTypeVariables(idents, span) => {
                     ParseError::UnboundTypeVariables(idents, span)
                 }
-                InternalParseError::InvalidUniRecord(..) => ParseError::InvalidUniRecord(),
+                InternalParseError::InvalidUniRecord(illegal_pos, tail_pos, pos) => {
+                    ParseError::InvalidUniRecord(illegal_pos, tail_pos, pos)
+                }
             },
         }
     }
@@ -1198,9 +1205,13 @@ impl ToDiagnostic<FileId> for ParseError {
                         .join(",")
                 ))
                 .with_labels(vec![primary(span)]),
-            ParseError::InvalidUniRecord(..) => {
-                Diagnostic::error().with_message(format!("invalid record literal"))
-            } // .with_labels(vec![primary(span)]),
+            ParseError::InvalidUniRecord(illegal_span, tail_span, span) => Diagnostic::error()
+                .with_message(format!("invalid record literal"))
+                .with_labels(vec![
+                    primary(span),
+                    secondary(illegal_span).with_message("can't use non-type construct"),
+                    secondary(tail_span).with_message("in presence of a tail"),
+                ]),
         };
 
         vec![diagnostic]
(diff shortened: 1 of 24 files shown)


src/error.rs
```diff
@@ -291,6 +291,13 @@ pub enum ParseError {
     ),
     /// Unbound type variable
     UnboundTypeVariables(Vec<Ident>, RawSpan),
+    /// Illegal record literal in the uniterm syntax (see RFC002). In practice, this is a record
+    /// mixing a tail and non-type constructs.
```
aspiwack, line 295: Make `RFC002` be a link.

I don't understand what “mixing a tail and non-type constructs” means. But probably this would be cleared out by an example.

src/error.rs
```diff
@@ -1195,6 +1205,13 @@ impl ToDiagnostic<FileId> for ParseError {
                         .join(",")
                 ))
                 .with_labels(vec![primary(span)]),
+            ParseError::InvalidUniRecord(illegal_span, tail_span, span) => Diagnostic::error()
+                .with_message(format!("invalid record literal"))
+                .with_labels(vec![
+                    primary(span),
+                    secondary(illegal_span).with_message("can't use non-type construct"),
```
aspiwack, line 1212: Similarly, I doubt that “can't use non-type construct” would make for a useful error message.

src/grammar.lalrpop
```diff
@@ -1,30 +1,48 @@
...
-CheckUnbound<Rule>: Types = <l: @L> <t: Rule> <r: @R> =>? check_unbound(&t, mk_span(src_id, l, r)).map_err(|e| lalrpop_util::ParseError::User{error: e}).and(Ok(t));
+WithPos<Rule>: Rule = <l: @L> <t: Rule> <r: @R> => t.with_pos(mk_pos(src_id, l, r));
+
+AsTerm<Rule>: RichTerm = <ut: WithPos<Rule>> =>?
+    RichTerm::try_from(ut)
+        .map_err(|e| lalrpop_util::ParseError::User{error: e});
+
+AsTypes<Rule>: Types = <ut: WithPos<Rule>> =>?
+    Types::try_from(ut)
+        .map_err(|e| lalrpop_util::ParseError::User{error: e});
```
aspiwack, line 42: This is elegant.

src/grammar.lalrpop
```diff
@@ -175,19 +220,9 @@ Atom: RichTerm = {
     Bool => RichTerm::from(Term::Bool(<>)),
     StrChunks,
     Ident => RichTerm::from(Term::Var(<>)),
+    AsTerm<UniRecord>,
```
aspiwack, line 223: I assume that at the end of the refactoring, you will only have a `Term: Richterm = AsTerm<TheWholeSyntax>`, correct?

src/parser/error.rs
```diff
@@ -19,4 +18,11 @@ pub enum ParseError {
     Lexical(LexicalError),
     /// Unbound type variable(s)
     UnboundTypeVariables(Vec<Ident>, RawSpan),
+    /// Illegal record literal in the uniterm syntax (see RFC002). In practice, this is a record
+    /// mixing a tail and non-type constructs.
+    InvalidUniRecord(
+        RawSpan, /* non-type construct position */
+        RawSpan, /* tail position */
+        RawSpan, /* whole record position */
+    ),
```
aspiwack, line 27: I've seen this before…

Do we need the error to be defined both in `error` and `parser::error`?

Same remarks as previously on the docstring.

src/position.rs
```diff
@@ -23,6 +23,22 @@ pub struct RawSpan {
     pub end: ByteIndex,
 }
 
+impl RawSpan {
+    /// Fuse two spans if they are from the same source file. The resulting span is the smallest
+    /// span that contain both `span1` and `span2`.
+    pub fn fuse(span1: RawSpan, span2: RawSpan) -> Option<RawSpan> {
+        if span1.src_id == span2.src_id {
+            Some(RawSpan {
+                src_id: span1.src_id,
```
aspiwack, line 32: What is the `src_id`?

src/error.rs
yannham: A "non-type construct" is anything that wasn't allowed in the original syntax inside a record type, but is allowed in a record literal. Typically a field definition `{foo = 1}`. In the RFC we discussed that it isn't immediatly obvious what semantics to give to `{foo : Num = 1; a}` (either as a type or as a term), so we planned to forbid using a polymorphic tail `; a` for anything that isn't a record type (technically, for any `t  : UniTerm` that isn't of the form `t = !r` for `r: Types` a record type in the original syntax and `! : Term + Type -> UniTerm` the embedding).

aspiwack: Don't explain it to me, instead explain it to someone working in the codebase. In 6 months.

src/grammar.lalrpop
yannham: In the long term, yes. My current plan is however to try to just do that for records (more or less the only part that is "common" between terms and types), and otherwise get away by doing the translation on the fly, which would most of those definitions additionally to `AsTerm<>`. The goal is to avoid to have to implement a separate, mostly duplicate AST definition for the uniterm syntax before first release.

src/position.rs
yannham: It's a identifier for the source file (differentiating e.g. positions in "foo.ncl" and "bar.ncl").

src/parser/error.rs
yannham: `error::ParseError` is a superset of `parser::error::ParseError`. The latter are custom errors that we can raise from within LALRPOP: it includes lexing errors (because we use a custom lexer), and now the `InvalidUniRecord`. Then LALRPOP can raise those errors or generic ones (unexpected end of file, unexpected token, etc.). On level up, `error::ParseError` also includes additional parsing errors when deserializing e.g. json files, and also wrap LALRPOP generic and custom errors with additional information, such as the file identifier.

For modeling errors, I've often missed structural typing, something like OCaml polymorphic variants. Right now we duplicate quite a few constructors to model the inclusion of some errors in a larger error set but with additional information. One alternative would be to reuse the previous level when possible, adding data if needed, something like `LALRPOPError(lalrpop::SomeErrorType, FileId)`. One disadvantage is that in this precise case, the error is much less nice to handle because `lalrpop::SomeErrorType` has usually a different and LALRPOP-specific representation. In practice the only part consuming errors is the one turning them into a diagnostics, so that's maybe not a big deal.

aspiwack (approved):
Do make sure the documentation, and in particular the error messages are built to last. But aside from that, this is all looking quite good.

Acaccia (changes requested):

src/parser/uniterm.rs
```diff
@@ -0,0 +1,134 @@
...
+                                        Ok(Types(AbsType::RowExtend(id, Some(Box::new(ctrt.types)), Box::new(acc)))),
+                                }
+                                else {
+                                    // Position of identifiers must always be set at this stage
+                                    // (parsing)
+                                    let span_id = id.pos.unwrap();
+                                    let term_pos = rt.pos.into_opt().unwrap_or(span_id);
+                                    Err(InvalidRecordTypeError(TermPos::Original(RawSpan::fuse(span_id, term_pos).unwrap())))
+                                }
+                            }
```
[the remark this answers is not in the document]
yannham: But with your suggestion, we are doing a copy in any case, even if the pattern doesn't match. While the `assert_matches!` first checks before doing the clone, doesn't it?

It is probably cheap 99.9% of the time, but just for the sake of pretending: if `rt.term` is a `List` of size `10000`, then in the first case we don't copy anything, while with your suggestion, we would copy the whole array, I think?

tests/unbound_type_variables.rs
```diff
@@ -0,0 +1,35 @@
...
+        assert_matches!(res, Err(Error::ParseErrors(_)));
+
+        match res {
+            Err(Error::ParseErrors(ParseErrors {errors})) => {
+                assert_matches!(errors.as_slice(), [ParseError::UnboundTypeVariables(vars, _)] if vars.contains(&Ident::from($var)) && vars.len() == 1)
+            }
+            _ => unreachable!(),
+        };
+    }
+}
```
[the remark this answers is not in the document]
yannham: The problem with a function is that if the test fails, it won't show the callsite (hence the actual line of the tests that is failing), but will always point to inside the function, which had proved annoying for debugging. This is why we did duplicate some macros throughout the tests. I agree that it isn't a great solution, maintenance-wise, and I would be happy to find a better one.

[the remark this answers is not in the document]
Acaccia: oh this is why :o Okay cool, interesting. I never knew about this (yeah, I guess I should read the rust book sometime).
In this case, I completely agree with the approach.

src/parser/uniterm.rs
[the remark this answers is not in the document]
Acaccia: That's not really what the `match_sharedterm` is used for. You would use it for cases where you still need access to the `SharedTerm` in case no match arm is selected, so that you avoid to duplicate a `SharedTerm` to another (maybe it's not very clear and I should clarify the documentation for the macro).

Here since you don't need it, the regular match is enough. I'm quite sure that being one of the classic construct of the language, there wasn't some optimization applied (not cloning a List of 10000 if not needed for example). I would be  surprised in this code if anything else than `ctry.types` was moved/cloned.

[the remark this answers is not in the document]
Acaccia: Just to clarify: it is not wrong to use `match_sharedterm` here, it just adds a level of indirection that makes it a bit more difficult to read (in my opinion, of course). So I would advise to only use it when it brings a real benefit, like access to the `SharedTerm` later or a clear indication with a benchmark that we would need it there.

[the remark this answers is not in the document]
yannham: > That's not really what match_sharedterm is used for. You would use it for cases where you still need access to the SharedTerm in case no match arm is selected, so that you avoid to duplicate a SharedTerm to another (maybe it's not very clear and I should clarify the documentation for the macro).

I don't understand why using or not the shared term in the `else` makes any difference. My understanding is that the goal of `matched_sharedterm` is to allow to take ownership of the underlying data *only if it matches a specific pattern*, instead of having to first perform a clone upfront that will be dropped without being used in some cases. Something that you can do transparently in a standard `match`, but not with something that is behind a `Rc`. If you need ownership in all branches, then on the contrary, you can call `into_owned` upfront as you will have to do it anyway, and use a standard match. It rather seems that `matched_sharederm` is useful only when you don't need to take ownership everywhere. Whether you reuse the original `Rc` or not when you don't take ownership doesn't seem very relevant to me.

Take the following case: a `RichTerm` is pointing to a `10000` elements `Term::List(ts)`, and it is a 2-Rc pointer. With the second solution, we will first clone blindly the underlying `Term` when calling the `into_owned()` function, and thus a `10000` elements array before matching. This won't match the pattern (that expects a `MetaValue`), and we will go to the `_` case: because we don't use the matched term, we will drop it right away. I'm afraid this won't be optimized away by `rustc/llvm`, as this is a pretty complex case. We do need the cloned term in the case it is a metavalue, which is only known at runtime. So it would need to push the `clone` inside all the branches, and then detects that one branches is unwraping an `Rc` and then dropping it without using it. That sounds like something [Koka](https://www.microsoft.com/en-us/research/uploads/prod/2020/11/perceus-tr-v1.pdf) would do, but from my (limited, let's be honest) knowledge of rust optimizations with respect to reference counting, I don't think it is anywhere near that level. But I may be wrong.

All of that being said, there's actually an important point to make, that I just realized only now: we are in the parser, and thus all `Rc` should be 1-counted, and no clone should ever take place when calling to `into_owned()`. Now, that does sound like a compelling case for having a simple `match` instead of using our own macro.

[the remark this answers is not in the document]
yannham: In the end I changed it anyway, because I realized this must be a 1-Rc, and in this case the `match_shared` is indeed useless. If you do compare the results, I'm curious about it too :slightly_smiling_face: but to be honest this does look like reuse analysis with `Rc`s, that is described in the paper linked above. I would be very very surprised if Rust was able to do something like that.

PR #591, opened 2022-01-22
Merge enum variants properly

### LINKED ISSUE #590: Enum variants don't merge even if they were the same
**Describe the bug**
Enum variants don't merge even if they were the same.

**To Reproduce**
```
nickel> `A & `A
error: Non mergeable terms
  ┌─ repl-input-1:1:1
  │
1 │ `A & `A
  │ ^^---^^
  │ │    │
  │ │    with this expression
  │ merged here
  │ cannot merge this expression
```

**Expected behavior**
It should result in
```
`A
```

**Environment**
 - OS name + version: `Linux xps 5.10.89-1-MANJARO #1 SMP PREEMPT Wed Dec 29 18:09:17 UTC 2021 x86_64 GNU/Linux`
 - Version of the code: a48713d

**Additional context**
I came across the issue when I applied the same contract (which specifies default enum variant) twice, something like
```
let T = {a | <A, B> | default = `A } & {a = `A} in {} | #T | #T
```

MiSawa:
Resolves #590

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -199,6 +199,26 @@ pub fn merge(
                 ))
             }
         }
+        (Term::Enum(i1), Term::Enum(i2)) => {
+            if i1 == i2 {
+                Ok(Closure::atomic_closure(RichTerm::new(
+                    Term::Enum(i1),
+                    pos_op.into_inherited(),
+                )))
+            } else {
+                Err(EvalError::MergeIncompatibleArgs(
+                    RichTerm {
+                        term: SharedTerm::new(Term::Enum(i1)),
+                        pos: pos1,
+                    },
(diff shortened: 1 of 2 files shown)


yannham (approved):
LGTM. Thanks a lot for reporting and contributing! Reviewing this also made me realize we could factor out all those `else { Err(MergeIncompatibleArgs(...)) }` in the last match case, because they are all the same. But that's another change for another PR.

yannham:
The deployment test is broken on the CI, but let's not accumulate PRs because of this, as it seems to be the case on each and every PR opened since a few days, even just documentation. The website builds fine when pointing to the last head, and even if something goes wrong, nothing will break until we manually update the website. So I'm merging this now.

PR #598, opened 2022-01-27
[Language server] walk contract types

ysndr:
Linearize contract annotations

Previously only the direct reference of a contract was linearized.
That means for annotations like below, `AlwaysTrue` could be referenced and navigated to

```nickel
let AlwaysTrue = EqualsTo true in
let value | #AlwaysTrue = true in
in value
```

This PR allows much more general constructs on the RHS of `|` by traversing the AST of the contract expression.
Therefore, now `AlwaysTrue` can be accessed even in code like this:

```nickel
let AlwaysTrue = EqualsTo true in
let contrs = {
  AlwaysFalse = EqualsTo false,
} in
let t | #AlwaysTrue = true in
let not | #AlwaysTrue -> #contrs.AlwaysFalse = fun x => !x in
```

### DIFF
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -298,12 +298,17 @@ impl Linearizer for AnalysisHost {
                 // recursively linearize flat types
                 fn walk_types(
                     lin: &mut Linearization<Building>,
-                    outer_host: &AnalysisHost,
+                    outer_host: &mut AnalysisHost,
                     t: &nickel::types::Types,
                 ) {
+                    let mut scope = outer_host.scope.clone();
+                    let (scope_id, next_scope_id) = outer_host.next_scope_id.next();
+                    outer_host.next_scope_id = next_scope_id;
+                    scope.push(scope_id);
+
                     let inner_host = AnalysisHost {
                         env: outer_host.env.clone(),
-                        scope: outer_host.scope.clone(),
+                        scope,
                         ..Default::default()
                     };
                     match &t.0 {
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -81,6 +81,19 @@ impl AnalysisHost {
     }
 }
 
+impl Default for AnalysisHost {
+    fn default() -> Self {
+        Self {
+            env: Environment::new(),
+            scope: Default::default(),
+            next_scope_id: Default::default(),
+            meta: Default::default(),
+            record_fields: Default::default(),
+            access: Default::default(),
+        }
+    }
+}
+
 impl Linearizer for AnalysisHost {
     type Building = Building;
     type Completed = Completed;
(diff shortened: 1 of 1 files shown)


lsp/nls/src/linearization/mod.rs
```diff
@@ -247,35 +260,72 @@ impl Linearizer for AnalysisHost {
...
+                            .filter_map(|chunk| match chunk {
+                                nickel::term::StrChunk::Literal(_) => None,
+                                nickel::term::StrChunk::Expr(rt, _) => Some(rt),
+                            })
+                            .for_each(|rt| walk_terms(lin, host.scope(), rt)),
+                        Term::App(rt1, rt2) => {
+                            walk_terms(lin, host.scope(), rt1);
+                            walk_terms(lin, host.scope(), rt2);
+                        }
+                        Term::Record(rts, _) | Term::RecRecord(rts, _, _) => rts
```
Avi-D-coder, line 291: Do we need to walk dynamic fields?

lsp/nls/src/linearization/mod.rs
```diff
@@ -247,35 +260,72 @@ impl Linearizer for AnalysisHost {
...
+                fn walk_types(
+                    lin: &mut Linearization<Building>,
+                    outer_host: &AnalysisHost,
+                    t: &nickel::types::Types,
+                ) {
+                    let inner_host = AnalysisHost {
+                        env: outer_host.env.clone(),
+                        scope: outer_host.scope.clone(),
+                        ..Default::default()
+                    };
```
ysndr, line 308: I wonder if it will bite us that we use the same scope and environment here instead of working in their own scope.

lsp/nls/src/linearization/mod.rs
ysndr: For stupid Contracts like 
```
... | #({some.contract = Contract}.some.contract)
```

or what other kind of dynamic fields do you mean?

lsp/nls/src/linearization/mod.rs
ysndr: Ah it will if we have a `let` expression here such as

```
#(let Contract = ... in Contract)
```

contract will leak into the outer environment (only on the linearization but it hurts autocompletion once finally implemented)

fixed in follow-up commit

lsp/nls/src/linearization/mod.rs
Avi-D-coder: I'm not sure if it would come up in contracts.
I'm wondering if we need to traverse the second field of `RecRecord`.
```rust
    RecRecord(
        HashMap<Ident, RichTerm>,
        Vec<(RichTerm, RichTerm)>, /* field whose name is defined by interpolation */
        RecordAttrs,
    ),
```

lsp/nls/src/linearization/mod.rs
Avi-D-coder: No idea

lsp/nls/src/linearization/mod.rs
ysndr: Oh, yeah they are not typechecked so i never dealt with those guys
As the TC doesnt touch them either, we should do so in the linearizer which would mean pulling up this walk method to be used by the linearizer.
I'm not sure if i want to do that here or in a later PR

Avi-D-coder: Either is fine

Avi-D-coder (approved):

PR #599, opened 2022-01-27
[Language Server] Explicitly link values

### LINKED ISSUE #585: Hovering on first parameter of a function shows the wrong metadata
**Describe the bug**
When hovering on the *first* parameter of a function, the LSP shows the metadata of the whole function.

**To Reproduce**
Write the following example in a file:

```
let f : Num -> Num
      | doc "foo" =
      fun x => x + 1 in
f 1
```
Open it with an NLS-enabled editor. Hovering on the `x` inside `fun x` shows the right type, but it also shows the documentation `foo` of `f`.

**Expected behavior**
Hovering on a parameter shouldn't show the metadata of the whole function being defined.

ysndr:
Fixes #585

Let identifier is associated with the referenced value explicitly.
Yet if the value is a function the generated item is the argument binding which receives the `meta` data.

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -5,7 +5,7 @@ use nickel::{
     identifier::Ident,
     term::{MetaValue, RichTerm, Term},
     typecheck::{
-        linearization::{LinearizationState, Scope, ScopeId},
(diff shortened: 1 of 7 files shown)


Avi-D-coder (approved):
Nice!

PR #608, opened 2022-02-08
Label: make types field a Rc

### LINKED ISSUE #603: Cloning and droping of Labels take a schoking amount of time
**Describe the bug**
Sorting a list currently takes an insane amount of time with Nickel. After generating flamegraphs, we can notice that a big amount of time is spend on cloning and dropping `Label`s, and especially their field `type`.
fg-label
(complete flamegraph linked in the `Files` section)

**To Reproduce**
In the linked PR (soon to be open), I added a new stupid benchmark sorting a list of 100 elements.
Running it with `cargo bench sort -- --profile-time=120`, we obtain a flamegraph similar to the one linked below.

**Expected behavior**
Better performances by avoiding a lot of clones.

**Files**
[flamegraph-benchmark.tar.gz](https://github.com/tweag/nickel/files/7994645/flamegraph-benchmark.tar.gz)

Acaccia:
Fixes #603 

This was an easy win, simply changing the `types` field in `Label` to an `Rc<Types>` gave instant improvement in performances, as seen on those benchmarks:
[criterion.tar.gz](https://github.com/tweag/nickel/files/8024099/criterion.tar.gz) (decompress and open _report/index.html_)

The flamegraph generated from the same benchmark doesn't even show any `Label` operation anymore:
image


aspiwack:
How much time did the benchmark gain overall?

yannham:
I believe there is no global gain reported in criterion, but randomly clicking through the benchmarks shows that some did not move a lot, and others got anything between roughly a few percents and 30~40%. Several are in the 30-40% zone, which is a nice return for slapping one `Rc` on top of a struct field. In the end we will still want to fully `Rc` `Types`, though, if only for typechecking.

PR #609, opened 2022-02-08
[RFCOO2] Part II: revenge of the type

yannham:
# Uniterm syntax

*Note*: this PR is for merging in the `uniterm` branch, and not directly into `master`, because it doesn't pass the CI. Fixing the CI has been done in a separate branch and will come in a subsequent PR for `uniterm`. We'll then merge the whole `uniterm` branch into `master`.

This PR implements the rest of RFC002. My initial plan was to get away with just the `UniRecord` introduced in #589, using the existing `Types` and `RichTerm`, and doing the conversions as needed on-the-fly. However, the intersection of the term and types syntax (variables and records) makes this seemingly quite difficult: it's then hard to avoid having different derivations in the grammar (aka ambiguities) when those elements are used in a compound expression (schematically, there is a derivation first going trhough type rules and then term rules, and vice-versa).

## Content

This PR:

- introduces a new AST for `UniTerms`, but that is in the end quite simple, as it reuses existing types definitions. It just wraps `Types` and `RichTerm` in dedicated variants, for those uniterms that are already determined to be of one kind, plus variants for common constructs for both terms and types that can still be converted as both later (`UniRecord` and `Var`).
- adds conversions facilities from and to `RichTerm` and `Types`
- adds helpers to handle type variable fixup (see paragraph at the end)
- patches the grammar to:
  1. Return and convert to and from `UniTerm` when needed 
  2. Mix type constructors and term constructors in the same grammar (e.g. `->` is now a normal operator in the `InfixExpr` rule, `forall` can bind a term on the right)
 
We took the occasion to add documentation to the grammar definition.
 
## Choices

The following choices and restrictions have been made:

- A type can now be a bare term, such as in `foo : MyContract a (b + 1)`. However, we restricted terms appearing in type position to forbid rules with potential top-level annotations on the right, which are annotated terms `foo | Annot`, let-bindings `let foo = 1 in foo | Num`, `switch` and `if-then-else`. Otherwise, something like `foo : let x = 1 in x : Num` is ambiguous. We could have set an associativity to accept it but while fixing one deterministic parsing, however this kind of expression is to be honest ugly and far enough from the spirit of allowing terms in types to just forbid such rules altogether. Any complex term expression with e.g. let-binding should be better put in a separate variable if one needs to use it as a contract, I think.
- The original plan was to remove enum types, but they actually appear in quite a few important functions of the stdlib. We have to discuss this but for now I just picked a temporary and arbitrary replacement `[|`/`|]` for `<`/`>` as enum type delimiters.
- The precedence of the arrow constructor `->` is set to the lowest (highest level in the grammar), such that it binds more loosely that other infix operators. `1 + 1 -> 2 + 2` doesn't make much sense right now, but it makes more sense as `(1 + 1) -> (2 + 2)` than as `1 + (1 -> 2) + 2` (who knows, in the future constants like `2` could have the semantics of `EqualsTo 2` when used as contracts, in which case the former parsing does make sense) `.

## Type var fixup

Now that both type variables and standard variables share the same namespace, the nature of a variable may depend on its surrounding context (or, more precisely, of enclosing `forall`s). In `foo | a -> b`, both `a` and `b` are parsed as `#a` and `#b` in the old syntax (i.e. term variables seen as contract). In `foo | forall a b. a -> b` both are type variables, and in `foo | forall a. a -> b` the first is a type variable and the second is a contract `#b`.

Since parsing is done bottom-up, when parsing `a -> b`, we don't know this yet. Thus, we parse everything as type variables, and then when reaching an annotation `:`, we know that no enclosing `forall`s will be able to bound variables anymore, and we traverse the newly created type to turn unbound type variables to term variables. This is implemented in `parser::uniterm::fix_type_vars`.

### DIFF
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -152,10 +152,10 @@ pub struct UniRecord {
 pub struct InvalidRecordTypeError(pub TermPos);
 
 impl UniRecord {
-    /// Try to convert a `UniRecord` to a type. The strict part means that if the `UniRecord` must
-    /// be a plain record type, uniquely containing fields of the form `fields: Type`. Currently,
-    /// it doesn't support the field path syntax: `{foo.bar.baz : Type}.into_type_strict()` returns
-    /// an `Err`.
+    /// Try to convert a `UniRecord` to a type. The strict part means that the `UniRecord` must be
+    /// a plain record type, uniquely containing fields of the form `fields: Type`. Currently, it
+    /// doesn't support the field path syntax: `{foo.bar.baz : Type}.into_type_strict()` returns an
+    /// `Err`.
     pub fn into_type_strict(self) -> Result<Types, InvalidRecordTypeError> {
         let ty = self
             .fields
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -339,7 +339,7 @@ impl TryFrom<UniRecord> for Types {
 /// parsing `fst: a` if `a` will be a type variable or a term variable (while, for all other
 /// constructs, an annotation is a boundary that `forall` binders can't cross). In this example,
 /// there is indeed an enclosing forall binding `a`. With unirecords, before fixing type variables,
-/// we have wait until we eventually convert the unirecord to a term (in which case we fix all the
+/// we have to wait until we eventually convert the unirecord to a term (in which case we fix all the
 /// top-level annotations) or a type (in which case we do nothing: the enclosing type will trigger
 /// the fix once it's fully constructed). Fixing a unirecord prior to a conversion to a term is
 /// done by [`fix_fields_types`].
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -16,8 +16,11 @@ use std::convert::TryFrom;
 ///
 /// During parsing, some constructs are common to both terms and types, such as variables or record
 /// literals ([`UniRecord`]s). They may be the source of ambiguities in the grammar, if we add two
-/// different derivation for them (e.g. one going first through term rules then types rules, and
-/// vice-versa).
+/// different derivation for them. This happens inside rules that accepts both a term and and a
+/// type, say the operands of an arrow `lhs -> rhs`. Take for example `lhs` a variable: `lhs := a`.
+/// It's hard not to end up having two possible derivation for `a`, one going first through term
+/// rules and then types rules (a custom contract `#a` in the old syntax), and vice-versa (a type
+/// variable `a` in the old syntax.
 ///
 /// To avoid the issue, we parse the source as a `UniTermNode`s, and give only one possible
 /// derivation - we will take a variable as an example in the following - through the parsing
@@ -71,13 +74,6 @@ impl TryFrom<UniTerm> for Types {
             UniTermNode::Record(r) => Types::try_from(r),
             UniTermNode::Types(ty) => Ok(ty),
             UniTermNode::Term(rt) => Ok(Types(AbsType::Flat(rt))),
-            // => {
-            //     ty.contract().map_err(|UnboundTypeVariableError(id)| {
-            //         // We unwrap the position of the identifier, which must be set at this stage of parsing
-            //         let pos = id.pos;
-            //         ParseError::UnboundTypeVariables(vec![id], pos.unwrap())
-            //     })
-            // }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


yannham:
I thought I could get away without having a new separate AST for `UniTerm`, but I'm not so sure now. The issue is that some basic constructs (records and variables for example) can be both interpreted as terms and types, and thus introduce ambiguity in the places in the grammar that accepts both terms and types. To keep the grammar unambiguous requires to butcher it until it has multiple version of many rules (a version without variables included, a version without records included, etc.: same as you would have to solve the dangling else problem in a LALR(1) grammar without precedence modifiers), across all the rules hierarchy, and for multiple combinations. It doesn't look it is worth it, so I'm going to rather parse everything as a new `UniTerm` enum, and then convert it to a term accordingly in the end.

src/parser/uniterm.rs
```diff
@@ -141,3 +307,105 @@ impl TryFrom<UniRecord> for Types {
...
+///
+/// ```
+/// let mk_pair : forall a b. a -> b -> {fst: a, snd: b} = <exp>
+/// ```
+///
+/// Since this unirecord will eventually be interpreted as a record type, we can't know yet when
+/// parsing `fst: a` if `a` will be a type variable or a term variable (while, for all other
+/// constructs, an annotation is a boundary that `forall` binders can't cross). In this example,
+/// there is indeed an enclosing forall binding `a`. With unirecords, before fixing type variables,
+/// we have wait until we eventually convert the unirecord to a term (in which case we fix all the
```
aspiwack, line 342:
```suggestion
/// we have to wait until we eventually convert the unirecord to a term (in which case we fix all the
```

src/parser/uniterm.rs
```diff
@@ -141,3 +307,105 @@ impl TryFrom<UniRecord> for Types {
...
+/// top-level annotations) or a type (in which case we do nothing: the enclosing type will trigger
+/// the fix once it's fully constructed). Fixing a unirecord prior to a conversion to a term is
+/// done by [`fix_fields_types`].
```
aspiwack, line 345: We don't have time to do this before the release. But honestly, I think this is starting to get into awkward hack territory.

It will be more straightforward, eventually, to have a syntax tree for uniterm, and two for terms and types with mutual inclusion constructors. We generate uniterms for the parsers, and then convert the uniterm to a term (or type) recursively (keeping an environment of binders to be able to substitute the variables properly).

src/parser/uniterm.rs
```diff
@@ -141,3 +307,105 @@ impl TryFrom<UniRecord> for Types {
...
+pub fn fix_type_vars(ty: &mut Types) {
+    fn fix_type_vars_aux(ty: &mut Types, mut bound_vars: Cow<HashSet<Ident>>) {
```
aspiwack, line 347: Ah, yes, it is annoying to represent an environment in Rust (by which I mean a set or a map, or even a list which is updated locally).

Are there documented best practices? My guess would be that the best way is to implement the local update by a mutation and the inverse mutation on the way out, which I find tricky to get right (and precludes parallelisation, but that's a bit of a minor concern here).

aspiwack (approved):

src/parser/uniterm.rs
aspiwack: I'm guessing, with esprit d'escalier, that one could rely on a purely `Rc`-d data structure (Or `Arc`-d, which makes it a touch tricky to implement such a data structure, if it always has to be implemented twice).

src/parser/uniterm.rs
```diff
@@ -43,35 +181,55 @@ impl UniRecord {
...
+                                acc.zip(pos)
+                                    .and_then(|(acc, span)| RawSpan::fuse(acc, span))
+                            })
+                            .flatten();
+
+                        Err(InvalidRecordTypeError(
+                            span.map_or(TermPos::None, TermPos::Original),
+                        ))
+                    } else {
+                        let elem = path.pop().unwrap();
```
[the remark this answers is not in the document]
yannham: It's true that it's currently not statically enforced, but the path is the LHS of a field declaration `foo.bar.[...] = value`, and it can't be empty in the grammar, so the parser should never outputs empty paths. On one side we could make sure this code never panics by changing the condition from `path.len() > 1` to `path.len() != 1`, and everything would work fine. On the other hand, it's an illegal state, so I'm tempted to keep it like this and let it fail hard if this invariant were to be violated.

PR #619, opened 2022-02-23
Make the builtin contract implementations inaccessible from user-code

yannham:
The implementation of builtin contracts is pure Nickel code written as part of the stdlib. Currently, they pollute the user environment at the top-level by introducing identifiers like `list`, `record`, etc. everywhere accessible. Beside the risk of confusion stemming from those identifiers seemingly defined nowhere but still being present in the environment, this caused clashes when trying to change the name of some stdlib modules. Since the uniterm syntax, those contracts are not to be used directly, since one can use `Num`, `List`, etc. directly as terms instead.

This PR prefixes the identifiers of the builtin contract implementations with a `$` (willingly distinct from the `%` used to introduce intermediate variables holding thunks at run-time) which makes them inaccessible from user code directly, because there is no way to write an identifier that starts with `$`.

### DIFF
--- a/src/stdlib.rs
+++ b/src/stdlib.rs
@@ -34,14 +34,14 @@ pub mod contracts {
     macro_rules! generate_accessor {
         ($value:ident) => {
             pub fn $value() -> RichTerm {
-                mk_term::var(stringify!($value))
+                mk_term::var(format!("${}", stringify!($value)))
             }
         };
     }
 
     // `dyn` is a reserved keyword in rust
     pub fn dynamic() -> RichTerm {
-        mk_term::var("dyn")
+        mk_term::var("$dyn")
     }
 
     generate_accessor!(num);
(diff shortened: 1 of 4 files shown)


thufschmitt (approved):
Looks good 👍🏻

Small question/suggestion though: Wouldn’t it be cleaner to put all these “hidden” values in their own (possibly hidden) module rather than hiding each of them individually?

yannham:
That's a good point. The idea of having not in a module was that builtin contracts may be often invoked and that putting them in a module would add one level of indirection. On the other hand, I'm far from being positive about it being noticeable at all, performance wise. Let met at least separate them from the visible `contract` module by putting them in a separate file.

yannham:
Arg, it's going to cause a painful rebasing with #620 and #622. Will do in a future PR.

PR #622, opened 2022-02-23
Rename lists to arrays

### LINKED ISSUE #533: Rename List to Array or Vector
Nickel lists are not lists in the sense that functional programming understands them (a datatype with inductive structure, effectively a linked list). I believe the same remark holds regarding Nix. The algorithmic complexity of operations like `elem_at` are not what one would expect, while `tail` might either take more time or leak memory. @aspiwack noted very early in the design of Nickel that lists are a misnomer. Though not likely, we may well want a list type too down the road, in which case it would be better for `List` not to mean array.

Change the type primitive type `List` to `Array` or `Vector`. Either is fine I think. "Array" is the term used in JavaScript and other configuration languages like CUE. "Vector" is perhaps more specific to C++, Rust, Haskell (the latter uses both terms for essentially the same thing), but perhaps rolls off the tongue better.

**Additional context**

ReScript [is an example](https://rescript-lang.org/docs/manual/latest/array-and-list) of a language that draws the distinction between lists and arrays as we do here.

yannham:
Depends on #620. Fix #533.


francois-caddet:
If it deppend on #620 why to name it `arrays` with the `s`?

yannham:
Yeah, the title is misleading, sorry. It's renaming `list` to `array`, there is no `s`.

francois-caddet:
In the diff, right now, I have `s` everywhere. Probably because #620 is not merged.

yannham:
There is of course still plural `arrays` in the codebase, and that's normal: name of benches, plural sentences in the documentation, etc. This branch was already based onto #620, so #620 being merged or not shouldn't make a difference. This PR does use the singular `array` for the stdlib module, which is what matters. If there may be individual mistake, could you please point at them directly with a review comment, to make the fix easier? Thanks!

francois-caddet (approved):

PR #624, opened 2022-02-24
Replace `#` by `%` for multilines strings and interpolation

francois-caddet:
To be able to use `#` for comments instead of `//` as proposed in (TODO: which issue?) we should use a different symbol for multilines strings.
In (this comment, `%` is proposed)[#]. This change imply to change also the character used for interpolation because its linked to it (same number of character to delimit a multiline string than used to start interpolation).


yannham:
Cf #207.

src/parser/lexer.rs
```diff
@@ -296,14 +294,14 @@ pub enum StringToken<'input> {
...
     // Has lower matching priority than `HashBrace` according to Logos' rules.
-    #[token("#")]
+    #[token("%")]
     Literal(&'input str),
 
     #[token("\"")]
     DoubleQuote,
-    #[token("#{")]
+    #[token("%{")]
     HashBrace,
```
yannham, line 305:
```suggestion
    Interpolation,
```
I guess `HashBrace` doesn't make sense anymore.

yannham (approved):
Nitpick: change `HashBrace` to `Interpolation`, and `"#{"` to "%{" in the parser.

yannham:
@francois-caddet I think you have something to change in the grammar.larlpop too. Grepping for HashBrace could a good idea. Also, there seems to be a conflict

PR #625, opened 2022-02-24
Remove Array aliasing Array Dyn

yannham:
Depends on #622.

We originally transitioned from only one list type `List` to a parametrized list type `List Dyn`. At that time, we introduced an alias `List` for `List Dyn`, now being `Array` for `Array Dyn`, to let existing code (albeit tiny) untouched. It turns out this requires special casing in the parser today with the uniterm syntax, and takes syntax real estate without being strongly motivated (to this day, we've mostly written arrays parametrized by something more precise than `Dyn`).

This PRs removes the alias, forcing an `Array` type to always have a parameter. If this turns out to be really useful in practice later, we can always introduce it again.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -365,7 +365,7 @@ impl Types {
 
         match &self.0 {
             Dyn() | Num() | Bool() | Str() | Var(_) => true,
-            Array(ty) if ty.0 == AbsType::Dyn() => true,
+            Flat(rt) if matches!(*rt.term, Term::Var(_)) => true,
             _ => false,
         }
     }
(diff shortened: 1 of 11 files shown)


src/types.rs
```diff
@@ -365,7 +365,7 @@ impl Types {
 
         match &self.0 {
             Dyn() | Num() | Bool() | Str() | Var(_) => true,
-            Array(ty) if ty.0 == AbsType::Dyn() => true,
+            Flat(rt) if matches!(*rt.term, Term::Var(_)) => true,
```
aspiwack, line 368: Oh, did you fix a but, while you were at it?

aspiwack (approved):

src/types.rs
yannham: The removal is just consequence of removing the alias, but yes the second line is a change to a tiny detail I noticed before. When printing an `Array` type, unneeded parentheses were added when the parameter was a contract, like `Array (Foo)` instead of `Array Foo`, which is fixed here.

PR #628, opened 2022-03-01
Fix missing field definition `?` when forcing a record

### LINKED ISSUE #480: Better error message for missing field definition ("empty metavalue")

yannham:
Actually close #480. #512 fixed the issue for direct record access, but not when the missing field definition is encountered while forcing recursively a record:

```
$ nickel export -f foo.ncl
error: missing field definition for `?`
```

Which is quite unfortunate, because `nickel export` uses `deep_seq` under the hood for the serialization, hence #480 wasn't fixed for one of the main use-case of contracts in Nickel (validate a configuration when exporting).

#512 made all record access operations to push a new element on the call stack indicating that a field is under evaluation. Doing so, when an empty metavalue is encountered, it's possible to go through the callstack to find the currently active field and report it as missing (as well as other positional information).

Deep sequencing doesn't go through the standard operations for field access. What's more, when encountering a record, it stores at once a continuation with several fields to evaluate, thus we need to remember the specific context (field name, positions, etc.) of each field to push the right thing on the call stack before evaluating said field.

This PR parametrizes `DeepSeq` with an optional call stack element to be pushed before evaluation. When generating deep seq calls from the deep sequencing of a record, we pass the corresponding `Field` stack element. There are better approaches than this PR. We should probably store this information on the stack rather than inside the term itself. Also, we should probably avoid allocating a new nested big `DeepSeq` expression when recursing inside a record or a list, and rather store the continuation on the stack, as we do for equalities.

However, with the first release milestone in mind, this seems to be the quickest route to deliver the fix. Will open an issue later to think about it again when having more time.


litchipi (approved):
Maybe not the most optimal implementation for sure, but clean enough for now so that LGTM.
You do mix a bit of formatting changes to the commit, but as it's just imports it shouldn't give headaches to anyone rebasing on top of the patch

Issue #389, opened 2021-09-02
Memory Leak


Acaccia:
**Describe the bug**
On current master, running memcheck on any example gives us a memory leak, as shown on this example below:
```
==32847== Memcheck, a memory error detector
==32847== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==32847== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==32847== Command: ./target/release/nickel -f examples/lists/lists.ncl export
==32847== 
5040==32847== 
==32847== HEAP SUMMARY:
==32847==     in use at exit: 190,648 bytes in 1,469 blocks
==32847==   total heap usage: 183,786 allocs, 182,317 frees, 22,555,037 bytes allocated
==32847== 
==32847== 190,648 (104 direct, 190,544 indirect) bytes in 1 blocks are definitely lost in loss record 384 of 384
==32847==    at 0x483E7C5: malloc (vg_replace_malloc.c:380)
==32847==    by 0x264765: _ZN6nickel4eval12eval_closure17heb48e97a080aa304E.llvm.1067354156114795460 (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x2630EE: nickel::eval::eval_full (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x1DDA39: nickel::program::Program::eval_full (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x160488: nickel::main (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x173FF2: std::sys_common::backtrace::__rust_begin_short_backtrace (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x174008: _ZN3std2rt10lang_start28_$u7b$$u7b$closure$u7d$$u7d$17h996a4f3e0abafbe7E.llvm.3445810120371879798 (in /home/acaccia/Workspace/nickel/target/release/nickel)
==32847==    by 0x3E9CB8: call_once<(),Fn<()>> (function.rs:259)
==32847==    by 0x3E9CB8: do_call<&Fn<()>,i32> (panicking.rs:401)
==32847==    by 0x3E9CB8: try<i32,&Fn<()>> (panicking.rs:365)
==32847==    by 0x3E9CB8: catch_unwind<&Fn<()>,i32> (panic.rs:434)
[...]

PR #631, opened 2022-03-02
Reduce memory leaks by attaching free variables to record fields

yannham:
# Free vars pass

Partially address #389. Continuation of #543 and #588. The idea is to annotate the fields of recursive records with their free variables (or, more precisely, with the other fields of the record that syntactically appear in their definition, that is the intersection of their free variables and the set of fields of the record) during a transformation pass, and leverage this information when building the fixpoint to avoid undue leaks.

Thanks to this data, we can add to the local environment of each fields only those fields that they use (or, at least, that appear in their definition). Before, every field was added, creating cycles of reference-counted pointers (and thus leaks) in pretty much any record:

```nickel
{
  a = "foo",
  b = a ++ " bar",
  c = b ++ " baz",
}
```

Before this PR, this recursive record would evaluate to an object where each field `a`, `b` and `c` retains a reference (`Rc<RefCell<_>>`) to all other fields, thus creating cycles that wouldn't be reclaimed (`a` -> `a`, `a` -> `b` -> `a`, etc).

After this PR, `a` retains no reference to its siblings, `b` only retains a reference to `a`, and `c` only to `b`. This is a DAG that should be deallocated without problem. Only mutually recursive fields like `{foo = fun a => bar (a +1), bar = fun a => foo (a + 1)}` should leak.

## Content

This PR:

- Augment the `RecRecord` variant to store additional free variables data
- Add a free_vars pass to fill those data during program transformation. In #543 and #588, @Avi-D-coder  tried various approaches to avoid having to do a separate pass or suffer a quadratic complexity for this pass. It turns out it is difficult to both avoid adding a whole new traversal just for free variables and not butcher the current term traversal interface. In the end, I went for a separate traversal. Given the current state, I don't expect it to have bad consequences, and that most of the time is taken by evaluation anyway. This makes the change simpler and a better separation of matters.
- Use the additional data when creating the fixpoint of a recursive records. This requires to persist the free vars data beyond the first evaluation of a recursive records, because since overriding, we produce new recursive records in the middle of evaluation when merging. The free vars are thus directly attached to the revertible thunks introduced by recursive records. This brings a bonus: we can now detect at transformation time expressions that don't actually depend on the other fields of a record recursively, and don't bother introducing a revertible thunk in this case, that would be recomputed at each subsequent merge and consume more memory.
- add unit tests for the computation of free variables.

### DIFF
--- a/src/eval/fixpoint.rs
+++ b/src/eval/fixpoint.rs
@@ -48,19 +48,16 @@ pub fn patch_field(
             .get(var_id)
             .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
 
-        // Because `deps()` borrows thunk, we have to first build an extension, release the borrow,
-        // and only then we can use `borrow_mut()` and patch the environment.
-        let extension = match thunk.deps() {
-            ThunkDeps::Known(deps) => rec_env
-                .iter()
-                .filter(|(id, _)| deps.contains(id))
-                .cloned()
-                .collect(),
-            ThunkDeps::Unknown => rec_env.clone(),
-            ThunkDeps::Empty => Vec::new(),
-        };
+        let deps = thunk.deps();
 
-        thunk.borrow_mut().env.extend(extension);
+        match deps {
+            ThunkDeps::Known(deps) => thunk
+                .borrow_mut()
+                .env
+                .extend(rec_env.iter().filter(|(id, _)| deps.contains(id)).cloned()),
+            ThunkDeps::Unknown => thunk.borrow_mut().env.extend(rec_env.clone()),
+            ThunkDeps::Empty => (),
+        };
     }
 
     // Thanks to the share normal form transformation, the content is either a constant or a
--- a/src/eval/fixpoint.rs
+++ b/src/eval/fixpoint.rs
@@ -33,6 +33,11 @@ pub fn rec_env<'a, I: Iterator<Item = (&'a Ident, &'a RichTerm)>>(
 
 /// Update the environment of the content of a recursive record field by extending it with a
 /// recursive environment.
+///
+/// For each field, retrieve the set set of dependencies from the corresponding thunk in the
+/// environment, and only add those dependencies to the environment. This avoid retaining
+/// reference-counted pointers to unused data. If no dependencies are available, conservatively add
+/// all the recursive environment. See [`transform::free_var`].
 pub fn patch_field(
     rt: &RichTerm,
     rec_env: &Vec<(Ident, Thunk)>,
@@ -42,8 +47,16 @@ pub fn patch_field(
         let mut thunk = env
             .get(var_id)
             .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
-        thunk.borrow_mut().env.extend(rec_env.iter().cloned());
+
+        let extension: Vec<_> = rec_env
+            .iter()
+            .filter(|(id, _)| thunk.deps().map(|fv| fv.contains(id)).unwrap_or(true))
+            .cloned()
+            .collect();
+
+        thunk.borrow_mut().env.extend(extension);
     }
+
     // Thanks to the share normal form transformation, the content is either a constant or a
     // variable. In the constant case, the environment is irrelevant and we don't have to do
     // anything in the `else` case.
--- a/src/eval/fixpoint.rs
+++ b/src/eval/fixpoint.rs
@@ -48,11 +48,17 @@ pub fn patch_field(
             .get(var_id)
             .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
 
-        let extension: Vec<_> = rec_env
-            .iter()
-            .filter(|(id, _)| thunk.deps().map(|fv| fv.contains(id)).unwrap_or(true))
-            .cloned()
-            .collect();
+        // Because `deps()` borrows thunk, we have to first build an extension, release the borrow,
+        // and only then we can use `borrow_mut()` and patch the environment.
+        let extension = match thunk.deps() {
+            ThunkDeps::Known(deps) => rec_env
+                .iter()
+                .filter(|(id, _)| deps.contains(id))
+                .cloned()
+                .collect(),
+            ThunkDeps::Unknown => rec_env.clone(),
+            ThunkDeps::Empty => Vec::new(),
+        };
 
         thunk.borrow_mut().env.extend(extension);
     }
--- a/src/transform/share_normal_form.rs
+++ b/src/transform/share_normal_form.rs
@@ -34,6 +34,7 @@ use crate::{
     position::TermPos,
     term::{BindingType, RichTerm, Term},
 };
+use std::collections::HashSet;
 
 /// Transform the top-level term of an AST to a share normal form, if it can.
 ///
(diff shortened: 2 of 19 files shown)


yannham:
@Avi-D-coder I can't request your review anymore, but if by any chance you happen to have both the desire and time to give a look, I would be happy to have your feedback :slightly_smiling_face:

Avi-D-coder:
Looks like the benchmarks no longer run on my machine. I'd also like to see what happens if you add the `run-benchmark` label to this PR.

```
nickel on  task/free-vars is 📦 v0.1.0 via 🦀 v1.59.0 
❯ nix develop         
warning: Using saved setting for 'extra-substituters = https://nickel.cachix.org' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = nickel.cachix.org-1:ABoCOGpTJbAum7U6c+04VbjvLxG9f0gJP5kYihRRdQs=' from ~/.local/share/nix/trusted-settings.json.
warning: ignoring untrusted substituter 'https://nickel.cachix.org'
warning: ignoring untrusted substituter 'https://nickel.cachix.org'
nix-pre-commit-hooks: updating /home/USER/nickel repo
pre-commit uninstalled
pre-commit installed at .git/hooks/pre-commit
=== Nickel development shell ===
Info: Git hooks can be installed using `pre-commit install`

[avi@desktop:~/nickel]$ cargo bench
   Compiling nickel v0.1.0 (/home/USER/nickel)
   Compiling pprof v0.4.5
warning: unused imports: `MultiStringToken`, `NormalToken`, `StringToken`
  --> /home/USER/nickel/target/release/build/nickel-7e46679926825c8c/out/grammar.rs:13:20
   |
13 |     lexer::{Token, NormalToken, StringToken, MultiStringToken},
   |                    ^^^^^^^^^^^  ^^^^^^^^^^^  ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

   Compiling utilities v0.1.0 (/home/USER/nickel/utilities)
warning: `nickel` (lib) generated 1 warning
    Finished bench [optimized] target(s) in 27.30s
     Running unittests (target/release/deps/arrays-fa52b9323ad17e22)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

Benchmarking foldr strings 50: Warming up for 3.0000 sthread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TypecheckError(UnboundIdentifier(Ident { label: "strings", pos: Original(RawSpan { src_id: FileId(9), start: ByteIndex(67), end: ByteIndex(74) }) }, Original(RawSpan { src_id: FileId(9), start: ByteIndex(67), end: ByteIndex(74) })))', utilities/src/lib.rs:139:55
```

yannham:
Ah, yes. The fix of #630 is required. I'll try to get it reviewed, merged, and rebase over it soon.

yannham:
Some additional data. Following the advice of @Acaccia, I forged a particularly pathological example (but that can really appear in practice!) which folds over a list and builds a big final record using merge. Then, I ran [bytehound](https://github.com/koute/bytehound) on it, both on `master` and on this branch.

Here is the example:

```nickel
let n = 100 in
array.generate function.id n
|> array.foldl (fun acc x => acc & {"%{string.from_num x}" = x}) {}
```

I expect this example to leak each and every intermediate record during folding. Let's run bytehound on this.

## On Master

- total allocated memory: 34.398 MB
- total leaked memory: ~ 34MB

Almost all the memory is leaked, increasing as a straight line over time, without a plateau. The memory consumed seems to be roughly proportional to the `n` parameter. Here is a plot of the memory consumption (above is general usage, below is leaked memory).

nickel-memory-master

The flamegraph show thats 88% of leaked memory is due to `fixpoint::patch`, which is building new recursive records during merging:

nickel-memory-flamegraph-master

## With this PR

- total allocated memory: 636.810 KB
- total leaked memory: 570.523KB

It seems there is a big leak a the beginning, and then the leaked memory stays constant. It doesn't depend on `n` (using `50` and `100` gave the same total leaked memory). I suspect this is the standard library there, and associated data.

nickel-memory-free-vars

The flamegraph shows that indeed most memory is leaked in the `prepare_eval` function (71%):

nickel-memory-flamegraph-free-vars

## Conclusion

On this example, we see a 50x reduction in memory usage and leaked memory with the proposed approach, but more importantly the leaked memory seems to be constant after an initial bump.

Avi-D-coder:
Nice!

yannham:
The last commit reduces further allocated memory to 530.078 KB and allocation counts by roughly 25%. It further distinguishes between "the free variables haven't been filled, so I must be conservative and provide the whole recursive environment" and "this thunk is not revertible so it doesn't need a recursive environment" case, reducing further `Rc`s cloning.

src/transform/share_normal_form.rs
```diff
@@ -89,7 +92,21 @@ pub fn transform_one(rt: RichTerm) -> RichTerm {
                         if !t.as_ref().is_constant() {
                             let fresh_var = fresh_var();
                             let pos_t = t.pos;
-                            bindings.push((fresh_var.clone(), t));
+                            let field_deps = deps.as_ref().and_then(|deps| deps.stat_fields.get(&id)).cloned().map(Rc::new);
```
Acaccia: The same thing is written a second time a bit further, shouldn't it be in a function ?

src/eval/fixpoint.rs
```diff
@@ -42,8 +47,16 @@ pub fn patch_field(
         let mut thunk = env
             .get(var_id)
             .ok_or_else(|| EvalError::UnboundIdentifier(var_id.clone(), rt.pos))?;
-        thunk.borrow_mut().env.extend(rec_env.iter().cloned());
+
+        let extension: Vec<_> = rec_env
+            .iter()
+            .filter(|(id, _)| thunk.deps().map(|fv| fv.contains(id)).unwrap_or(true))
+            .cloned()
+            .collect();
```
Acaccia: This vec allocation frustrates me so much...
Here, since deps is in an Rc, I think I would make a `deps_clone` in `Thunk` methods to get it directly without the `Ref`, so that I could pass the iterator directly to the extend.

Acaccia (approved):
Approved, but with 2 small comments :+1:

src/transform/share_normal_form.rs
yannham: It's slightly different, because this one `stat_fields` while the other uses `dyn_fields`. Let me think 5min if I can write a function generic enough. That being said, you made me realize I forgot to do the `is_non_rec` optimization in the second case, so this duplicated code is not harmless, in the end...

src/eval/fixpoint.rs
yannham: That's funny because in the meantime (last commits) I added a comment before your review to explain it. The problem here is that `deps` is borrowing the thunk, so we can't keep a lazy iterator and use it in the `extend`, because the `borrow_mut()` would fail saying there is already an immutable borrow active on the thunk's `refcell`.

yannham: Sorry, I haven't read your comment carefully. But then your are cloning the `deps` while this code doesn't. Arguably the `deps` should be smaller than the `rec_env`, but I don't think there is a case between the two that is obviously better in every respect between the two. Also, I'm not sure to see how that would go: even with a `deps_clone`, you still have to call it on `thunk`, which borrows thunk, which would still fail when doing it while there's the mutable borrow active? What would be the difference with just calling `thunk.deps().clone()`?

yannham: You can ignore my previous comments, I was mistaken. `deps` should way smaller, and its size is bounded by the one of `rec_env`, so you're right, it's better to clone `deps()` than the full `rec_env`.

Acaccia: I'm not sure I was understandable, because in my explanation, I wanted to write the the deps because it is in a Rc, not the reference to deps that you get here.
Still not sure I'm understandable...

yannham:
And the last correction (doing the non_rec optimization also for non recursive fields) brings down allocated memory to 413.586 KB, leaked memory to 396K, and allocation counts down by another 25-30%.

src/eval/fixpoint.rs
Acaccia: Nope, but that works too :D

yannham: Ok, now I get it. We can use an `Rc` indeed and not clone anything. I don't expect it to make a big difference, but it does make the code simpler by getting rid of the `D` parameter of `ThunkDeps` and the `cloned()` method, so here it is: f8d24a2637e8ba445a2c35f0b978af2abc5805e4

Issue #295, opened 2021-02-15
The Syntax Bikeshedding Tipi: prefix character for contracts (#)


yannham:
As mentioned by @mboes in https://github.com/tweag/nickel/issues/207#issuecomment-775198557, we may want to follow the convention of using `#` to denote comments. This shouldn't clash with the interpolation syntax which only applies inside strings, but it does clash with the user-defined contract syntax `value | #Contract`.

Another consideration is that other interpolation-ish syntax still uses the old `$`, like dynamically defining a record field from an expression: `{ $"my_" ++ "field" = value }`, or the field removal operator `record -$ field`. It would be nice to use a consistent character/syntax for interpolation, however using `#` for these cases would clash with using `#` for comments. I see the following solutions:

- Keep things as they are
- Find a new character for contracts, set `#` to be comments, and keep the `$` for field interpolation, at the price of some inconsistency.
- Find a new character for both contracts and interpolation, and only use `#` for comments.

It seems `%` is sometimes used for interpolation in templates languages. For contracts, any common prefix character would do: `%`, `@`, `&`, `$`, `~`, etc. Although some may be too closely associated with another semantics in the wild such as `$` for interpolation or `&` for intersection types/merging. In that case, the more neutral ones like `@`, `%` or `~` may be preferable.

mboes:
Agreed. `%` looks reasonable to me.

yannham:
Closed by the uniterm syntax, by #634 and #624. Took `#` for comments, `%` for interpolation, and contracts don't need a character anymore.

PR #634, opened 2022-03-03
Syntax/hashtag for comment

francois-caddet:
now use `#` for comments
relate to #295

### DIFF
--- a/tests/pass/builtins.ncl
+++ b/tests/pass/builtins.ncl
@@ -4,8 +4,8 @@ let Assert = fun l x => x || %blame% l in
   # is_record
   builtin.is_record {},
 
-  #/ This currently do not check that subexpressions are actually forced,
-  #/ just that the evaluation succeeds
+  # This currently does not check that subexpressions are actually forced,
+  # just that the evaluation succeeds
   %seq% 1 true,
   let x = (1 + 1) in %seq% x x == 2,
   let r = {a=(1 + 1),} in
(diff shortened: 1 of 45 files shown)


tests/pass/builtins.ncl
```diff
@@ -1,11 +1,11 @@
 let Assert = fun l x => x || %blame% l in
 
 [
-  // is_record
+  # is_record
   builtin.is_record {},
 
-  /// This currently do not check that subexpressions are actually forced,
-  /// just that the evaluation succeeds
+  #/ This currently do not check that subexpressions are actually forced,
```
yannham, line 7:
```suggestion
  # This currently does not check that subexpressions are actually forced,
```

tests/pass/builtins.ncl
```diff
@@ -1,11 +1,11 @@
...
+  #/ just that the evaluation succeeds
```
yannham, line 8:
```suggestion
  # just that the evaluation succeeds
```

yannham (changes requested):
The search and replace is a bit too brutal:

1. It turned all URLs from `https://` to `https:#`. I guess another `s/http\(s?\):#/http\1:\/\//` (or alike) should fix that
2. The RFC001 uses the `//` operator from Nix a lot, which is turned into `#` incorrectly. The simplest is probably to just revert all the changes done to this file: I think it's fine for technical RFCs, which are dated, to use older syntax that made sense at the time of writing.

yannham (approved):

PR #637, opened 2022-03-04
[User manual] Add introduction

yannham:
Add an introduction for the user manual. It just takes part of the README and link to the getting started section of the website to get something up and running. Ideally the getting started section would be also in the manual, but right now it uses the playground on the website, so I think for the time being we can keep it separated.

This will be the landing page for `https://nickel-lang.org/user-manual/`.

### DIFF
--- a/doc/manual/introduction.md
+++ b/doc/manual/introduction.md
@@ -1,0 +1,85 @@
+---
+slug: introduction
+---
+
+# The Nickel User Manual
+
+Nickel is a generic configuration language.
+
+Its purpose is to automate the generation of static configuration files - think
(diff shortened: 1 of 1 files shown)


garbas (approved):
I like he overall intro, one thing that I'm missing (also on the website) is setting of expectations for the user, eg. Can I use Nickel tomorrow in production? Or when can I use it? What other features are in the works? I guess I'm missing some sort of roadmap that would set expectations. Maybe this should be done in a separate PR.

Other than that the intro is good.

yannham:
You're right. Will do in a separate PR.

PR #641, opened 2022-03-07
[Manual] Set language for markdown code blocks

yannham:
Set `nickel` language for markdown code blocks of the manual when appropriate, to get syntax coloring in code editor and on the website.


ysndr (approved):
Esp. in syntax.md are a few repl/-ish lines thatcould possibly benefit from highlighting as well. Maybe its worth tagging them as `nickel` or `nickel-repl` as well

yannham:
Yes, I didn't set them as `nickel` because it's not strictly nickel code, and we would need a syntax definition for `nickel-repl` to have it be useful somewhere. Thus left as a future TODO.

PR #644, opened 2022-03-08
Hide enum types in the stdlib and the manual

yannham:
The enum type syntax has been changed temporarily since the unified syntax
change. While we bikeshed on a new solid syntax, we don't want users to rely
on the temporary one.

This PR:

- Hides enum types from stdlib types by wrapping them as contracts with meaningful names and error messages.  When a user write, say, ``builtin.serialize `NonExistentFormat "foo"``, the error they will face is understandable without using the enum type syntax.
- Documents enum tags but not enum types in the syntax section of the manual.
- Hide documentation about enum types in the typing section of the manual.

That way, they can still be used under the hood the stdlib, but without facing users.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -6,10 +6,11 @@ slug: syntax
 
 ## Simple values
 
-There are three basic kind of values in Nickel :
- 1. numeric values,
- 2. boolean values,
- 3. strings.
+There are four basic kind of values in Nickel :
+ 1. numeric values
+ 2. boolean values
+ 3. strings
+ 4. enum tags
 
 ### Numeric values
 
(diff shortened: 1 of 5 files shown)


doc/manual/syntax.md
```diff
@@ -178,6 +179,40 @@ def concat(str_array, log=false):
...
+#       "
+
+```
+
+While it's technically possible to just use strings in place of enum tags, using
+an enum tag insists on the fact that only a finite number of alternatives can be
+used for the corresponding value.
+
+Enum are also handled more finely by the typechecker, however, as of the version
+0.1 of Nickel, enum types have been temporary disabled for incidental reasons.
```
aspiwack, line 214: Be more specific: “there are only two enum types[…]”

stdlib/array.ncl
```diff
@@ -254,7 +269,7 @@
         "%m
       = fun f n => %generate% n f,
 
-    sort : forall a. (a -> a -> [| Less, Equal, Greater |]) -> Array a -> Array a
+    sort : forall a. (a -> a -> ComparisonResult) -> Array a -> Array a
```
aspiwack, line 272: Is it not statically typechecked? The (static) type checker currently reads this as an opaque type, why doesn't it complain?

yannham: This is why the CI is failing currently, I missed this one. Indeed we need a contract application.

doc/manual/syntax.md
yannham: You mean, the one from the stdlib? Those are standard custom contracts. I wrote them using enum types just because I can, because it's more concise and doesn't cost anything, but any user can write:

```nickel
MyComparisonResult = contract.from_predicate (fun value => value == `Less || value == `Equal || value == `Greater)
```

Should we really mention them as "enum types" in the doc, although they are just opaque types, statically?

stdlib/array.ncl
aspiwack: I don't understand your answer, I'm afraid :slightly_smiling_face:

doc/manual/syntax.md
aspiwack: I changed my mind, now that we've discussed the details. Let's say

```suggestion
Enum will also handled more finely by the typechecker, however, as of the version
0.1 of Nickel, enum types are not yet supported.
```

aspiwack (approved):
I see. Good application of gradual typing.

Honestly, I would have rather we hid the syntax of enum types completely, and implemented `ComparisonResult` & co as magic constants. But it's probably the more pragmatic choice.

Maybe label every use with a comment saying “internal syntax, don't rely on it”.

PR #648, opened 2022-03-10
Update code samples in stdlib docs to use renamed stdlib modules

silverraven691:
This follows changes made in #620


yannham (approved):
Missed those, indeed. Thanks!

PR #658, opened 2022-03-15
Note on possible lossless parsers

ErinvanderVeen:
I would like this PR to serve as a discussion for the recommendation and options laid out in the note, and options I failed to include.


notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,137 @@
...
+of rust-analyzer. Ungrammar has not seen use outside of rust-analyzer. I
+personally do not see a need for Nickel to base its SyntaxNodes on ungrammar.
+
+Since we can no longer use lalrpop, the parser would have to be written by
+hand. Parser combinators are not very suitable to this approach, as many
+are based around a parser succeeding or failing, negating rowan's advantage of
+being able to produce an AST even when there is an error.
+
+There is of course nothing fundamentally impossible about using some parser
+combinator library.
```
yannham, line 76: Another argument used usually against parser combinators is that they are not very efficient, but I don't know if that claim is fair in practice. I imagine that it depends a lot on the particular parser combinator library.

yannham (commented):
As a general question: being comment/newline-aware would indeed require to wrap many rules. Actually you don't need 4 rules per operation, but you do need some duplication. Also, we need to be careful not to introduce ambiguities: 
```nickel
(+)
 1 # foo
 2
```

Should `#foo` be associated with `1` or `2`? Or the whole application?

Something that is not totally clear though is how this issue would be solved by other technologies. Whatever formalism you use, you still have to specify that comments and newlines may be associated with some (many) elements of your grammar. If there is a reason why this is easier with, say, a PEG parser generator or parser combinators than with LALRPOP, then I think this deserves to be explained in this document.

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,137 @@
...
+Of course, rust-analyzer is greatly benefitted by a lossless ast (CSTs as
+they call them). To this extend they created two libraries specifically for
+this purpose. [rowan][rowan] and [ungrammar][ungrammar].
+
+Rowan is a widespread library that has already seen use in existing languages.
+For instance for the
+[slint](https://github.com/slint-ui/slint/) language that make use of rowan for
+their language server and
+[compiler](https://github.com/slint-ui/slint/tree/master/internal/compiler).
+Hence we can safely say rowan is likely suitable for nickel.
```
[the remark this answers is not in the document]
ErinvanderVeen: I will reword this section, but what I wanted to imply is that:
- rowan is used for quite a few programming tools (mainly language servers)
- at least one compiler

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,137 @@
...
+
+Ungrammar is a grammar that works to construct the SyntaxNodes as described in
```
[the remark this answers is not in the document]
ErinvanderVeen: I have reworded this section to:

Ungrammar is "a ... formalism for describing concrete syntax trees". Concretely,
ungrammar takes a
[file written in its grammar](https://github.com/rust-analyzer/ungrammar/blob/master/rust.ungram)
and produces
[a file](https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/syntax/src/ast/generated/nodes.rs)
that contains the functions (RedNodes) needed to interact with the GreenNodes
(typically produced by rowan). Ungrammar has seen limited/no use outside of
rust-analyzer, and I do not think Nickel would benefit greatly from it.

aspiwack:
What would be the advantage of using something like Ungrammar rather than writing your concrete syntax tree? It's not obvious to me.

aspiwack:
@tek You have experience with tree-sitter, maybe you have an opinion on whether a tree-sitter parser can be use as the actual front end for a language? This would most definitely reduce the complexity for this issue.

ErinvanderVeen:
> What would be the advantage of using something like Ungrammar rather than writing your concrete syntax tree? It's not obvious to me.

@aspiwack I have tried to clarify what ungrammar is used for, hopefully this clears it up.

aspiwack:
What difference do you make between a typed syntax tree and a untyped syntax tree?

PS: I think we should refrain from using the moniker AST (which stands for _Abstract_ Syntax Tree) here as we are speaking of _concrete_ syntax trees instead.

ErinvanderVeen:
In the most recent commit I have now made a glossary that defines the terms I use, and made the use of these terms consistent.

milahu:
> whether a tree-sitter parser can be use as the actual front end for a language?

related issue

* https://github.com/tree-sitter/tree-sitter/issues/1631
  * parse errors are not helpful
  * "Usually tree-sitter is used as a fast first-pass parser before the slower actual parser comes up with the syntax errors. I think there are benefits to keeping tree-sitter simple so grammars stay easy to write."
  * "Tree-sitter is also a syntax-only parser and does not have any capability for adding semantic analysis like variables not being defined or incorrect function arity; full parsers include these things with their error messages."

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,193 @@
...
+Lossless/full fidelity: A lossless CST is a tree that contains all information
+to recreate the original string/file exactly.
+
+Typed: A Typed Syntax Tree is a tree where the type of the parent implies the
+type of the children. Every node has a specific type, with specific fields,
+nodes are not interchangeable. Most ASTs are typed.
+
+Untyped: An Untyped Syntax Tree is a tree where the type/kind of a parent node
+does not force the type/kind of the children nodes. Lossless syntax trees are
+typically untyped.
```
yannham, line 42:
```suggestion
- **Concrete Syntax Tree (CST)/Parse Tree**: A tree that results from parsing, and
represents the syntactic structure of a string/file. May be lossless/full
fidelity.
- **Abstract Syntax Tree (AST)**: A tree that represents the abstract syntactic
structure of a string/file. Used for further processing of the file. Not
lossless/full fidelity. 
- **Lossless/full fidelity**: A lossless CST is a tree that contains all information
to recreate the original string/file exactly.
- **Typed**: A Typed Syntax Tree is a tree where the type of the parent implies the
type of the children. Every node has a specific type, with specific fields,
nodes are not interchangeable. Most ASTs are typed.
- **Untyped**: An Untyped Syntax Tree is a tree where the type/kind of a parent node
does not force the type/kind of the children nodes. Lossless syntax trees are
typically untyped.
```

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,193 @@
...
+replacing the syntax tree with a [rowan][rowan] based
+alternative. Unfortunately, while it is possible to integrate logos, rowan, and
+lalrpop, comments and whitespace would still be lost. This is an inherent
+problem with lalrpop in which grammar rules match directly on the token input.
+
+# The problem with comments/whitespace
+Above, it was stated that the way lalrpop consumes the token input stream is
+not well suited for a formatter (or other programming tools). This is a general
+problem with parsers that attempt to construct a typed syntax tree directly
+(there is no place for comments in a typed tree. There are ways around this,
```
yannham, line 59:
```suggestion
(there is no place for comments in a typed tree). There are ways around this,
```

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,193 @@
...
+### Disadvantages
+- By far the most work
+- Rapid development is harder when implementing a parser from scratch
+
+## [Rowan][rowan] + [Ungrammar][ungrammar] (optional) + [Logos][logos]
+Rowan and Ungrammar are two libraries created for the
+[rust-analyzer][rust-analyzer] project.
+
+Of course, rust-analyzer is greatly benefited by a lossless concrete syntax tree.
+To this extend they created two libraries specifically for
```
yannham, line 84:
```suggestion
To this extent they created two libraries specifically for
```

notes/lossless-ast-parsing.md
```diff
@@ -0,0 +1,193 @@
...
+nice to share a grammar for editors and nickel itself. Tree-sitter's concept is
+not too far removed from rowan (an untyped tree that is easy to update) with
+similar purposes (to ease the creation of development tools). However,
+tree-sitter does not provide the nice interface that rowan provides. For
+instances, there are no SyntaxNodes/RedNodes and no included trait to convert
+the CST to a suitable AST.
+Aside from
+[toy-examples](https://github.com/vanhtuan0409/toon-lang) there are, no
+existing projects that utilize tree-sitter as a parser for a rts/compiler/etc.
+I don't see a fundamental problem that would stop us from using tree-sitter as
```
yannham, line 168: Now I think you can add something about error reporting being too limited.

yannham (changes requested):
I think those notes are useful to keep, and aren't really expected to change much by now. The main issue is probably the lack of update about the error reporting of tree-sitter, but once fixed, we should merge this.

yannham:
ping @ErinvanderVeen

ErinvanderVeen:
Will be updated. But need to figure out how to word Xavier's approach.

PR #667, opened 2022-03-21
Preserve record location when applying record contract

yannham:
When working on the dogfooding codebase, I noticed that a lot of errors (and in particular, the `missing field definition`) weren't able to correctly locate the offending value, in particular when the value is a record with a record contract applied (which is expected to be pretty frequent in Nickel):

nickel-error-record-without-pos

This is due to record contract application being desugared to a (variant of) merge, with no position information. A first step is to give the same position to the elaborated merge as the contract parameter. However, this position is the position of the argument before evaluation, which isn't always helpful, because it can be proxied by variables or parameters to built-in contracts:

nickel-record-loc-1

A better solution was to do more or less the same thing, but after the evaluation of the (modified) merge: that is, when applying a record contract using `MergeContract(contract, record)`, we set the position of the final result to be the one of record. Now we get something more helpful:

nickel-record-loc-2

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -400,10 +400,16 @@ pub fn merge(
                 .iter()
                 .try_for_each(|rt| fixpoint::patch_field(rt, &rec_env, &env2))?;
 
+            let final_pos = if mode == MergeMode::Standard {
+                pos_op.into_inherited()
+            } else {
+                pos1.into_inherited()
+            };
+
             Ok(Closure {
                 body: RichTerm::new(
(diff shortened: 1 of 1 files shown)


steshaw:
Should there be a test for this?

yannham:
> Should there be a test for this?

Ideally, yes. But there's currently no tests whatsoever on error messages (which are also not trivial to make robust to small changes), so I would tend to think that this is a separate issue that need to be solved in a dedicated PR. It would be a bit inconsistent to have just one lone test for this specific case.

ErinvanderVeen (approved):

PR #679, opened 2022-04-08
Add CONTRIBUTING.md

yannham:


CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+At the root of the repository, run:
+
+```shell
+$ nix build .#buildWasm
+```
+
+The corresponding NPM package is put in `result/`.
+
+#### Using Cargo
```
aspiwack, line 170: These instructions are horrible ([remember](https://discourse.tweag.io/t/make-your-instructions-snap/157)). Is there really no better way?

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+- Unit tests, located directly in the corresponding module.
+- Integration tests, located in the dedicated crate `tests`.
+
+You can take inspiration from the existing tests to add your own. By convention,
+tests expected to pass are written in a standalone Nickel file in `tests/pass/`.
+Each `.ncl` file defines a list of expressions that must individually evaluate
+to the boolean `true`. The whole file is an expression that returns true if and
+only if every tests pass, or fail with a contract failure to help locating the
+failing test (instead of returning just `false`).
```
aspiwack, line 223: It's unrelated to this document, but I'm curious as to why you care about the returned value if failure is signified by a contract violation?

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+If a test expected to pass is failing, run it directly using nickel with `nickel
+-f tests/pass/test_that_doesnt_pass.ncl` to get better error messages than
+`cargo test`.
```
aspiwack, line 227: Is that a bug or misfeature in `cargo test`? Is it something that we could considering fixing upstream?

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+Tests expected to fail are often embedded directly into rust source code,
+because you usually want to additionally check that the error is the one you
+expect.
```
aspiwack, line 231: How? Maybe it's enough to give a link to an example.

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Note that a full run takes some time, up to a dozen of minutes. You can run
+specific benchmarks instead of the full suite. Please refer to the documentation
+of [`cargo bench`][doc-cargo-bench].
+
+# How to submit changes
+
+Once you have made sure the maintainers are aware of your changes, you can start
+the implementation, eventually submitting changes as a pull request.
```
aspiwack, line 252: This is remarkably condescending. Let's avoid.

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+    * run `cargo test` on the root crate `nickel-lang`
+    * run `cargo check` inside auxiliary crates: `lsp/nls` and
+        `utilities/`, to make sure that your change didn't impact them
+    * run `cargo fmt` at the root of each modified crate
+    * (optional) check that the WebAssembly REPL is building. See [building the
+      WebAssembly REPL](#webassembly-repl). This step may take some time, and
+      should only be required if you added, upgraded or downgraded
+      dependencies.
+6. Once all the previous steps are completed, you can push your branch to your
+   fork and make a pull request on GitHub.
```
aspiwack, line 278: This is almost all standard github workflow or checked by the CI.
```suggestion
1. Documentation: If you added new items (functions, modules) to the public API
   of a crate, please document those items in-code. If you made an user-facing
   change (syntax, stdlib, language features, etc.), please update the existing
   documentation (in particular the user-manual) in consequence.
2. Tests: be it for bug fixing or adding new features, try to write extensive
   tests as much as possible, to ensure the correctness of your changes as well
   as avoiding regressions.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+Your pull request will need at least one approval from a maintainer to be
+merged. Maintainers will review your pull request and may request changes as
+well as leave comments, that needs to be addressed in order to the process to
+move forward.
```
aspiwack, line 283:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+
+Even when a code review is accepting, it may still come with questions or
+comments. Please consider all the review comments and decide to act in
+consequence. Once done, resolve the corresponding conversations.
```
aspiwack, line 287:
```suggestion
```
Instead, just let people know when you review their PRs what you expect of them.

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+There are two ways to set up a development environment for Nickel: using
+[Nix][nix], or directly via your preferred system package manager. Nix is able
+to drop you in a development shell with everything needed (the Rust toolchain,
+mostly) to hack on Nickel in one command, without installing anything globally
+on your system. _While Nix is also capable of building Nickel by itself, using
+cargo (either a system-wide installation or the one given by the Nix development
+shell) is the recommended way of building when working on the Nickel repository
+itself_. The reason is that incremental compilation for Rust and Nix is not
+there yet, and incremental rebuilds using only Nix are going to be painfully
+long.
```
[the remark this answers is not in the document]
yannham: I think most `CONTRIBUTING.md` rather offload those instructions to a `HACKING.md` (or separate guide) document (Haskell, Rust, OCaml, etc.), that they link to in `CONTRIBUTING.md`. I may have started thinking that this would be very small, and thus could inline it, but the WASM stuff ended up being non-trivial, a separate guide may be a better approach.

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+### NLS (nickel-lang-lsp)
+
+To build NLS, the LSP server for Nickel, build the crate inside `lsp/nls`:
+
+```shell
+$ cd lsp/nls
+lsp/nls $ cargo build
+```
```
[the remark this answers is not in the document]
yannham: Yes, with `cargo build --all`. Let me add this information, although having per-crate instruction is still useful (if you're working on the language server, it's probably more natural to do check and build only on NLS during development than going back to the root of the repo)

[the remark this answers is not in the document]
aspiwack: But I'd expect to call `cargo build lsp` or something to that effect. Rather than `cd`-ing into a subdirectory.

CONTRIBUTING.md
yannham: It's fundamentally annoying, because of this patching Cargo.toml. I guess we could just drop one `sed` invocation that do steps 2 and 3., which is what the flake is doing during the build phase. I'll try to make it better.

aspiwack: I want to express my puzzlement at the fact that we need to patch the `Cargo.toml`. Why do we need to do something like that?

CONTRIBUTING.md
yannham: I don't think that's really the job of `cargo` (which is lower-level). It's just that tests are usually based on `assert!` or the like, and when those fail, we get a message that two Rust objects (usually printed in a verbose and exhaustive way) are not equal. In that case, probably that `Ok(true)` is not equal to `Err(...some contract error)`.

On the other hand, running the nickel interpreter will pretty print the error as usual, pointing the right locations and so. I imagine this is technically possible to make the error reporting of tests to be more similar to the interpreter. I just don't know if it's really worth the effort: failing tests is a rather rare event, and in that case running `nickel` once separately for debugging purpose doesn't sound like a prohibitive cost.

aspiwack: I see. I understand how this isn't a most urgent situation. It kind of depends on the cost to fix. It's better to have informative error messages right in the test framework, but it's also not so much better as to justify spending a lot of time on it.

CONTRIBUTING.md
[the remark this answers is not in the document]
yannham: Ah, yes, there is `cargo build -p nickel-lang-lsp`. Will add.

CONTRIBUTING.md
```diff
@@ -0,0 +1,303 @@
...
+$ nls --version
+nickel-lang-lsp 0.1.0
+```
+
+### WebAssembly REPL
+
+This repository is able to build WebAssembly (WASM) version of the REPL, which
+is used for the online playground on [nickel-lang.org][nickel-lang.org]. While
+using `cargo` directly is the recommended way for building, both methods have
+advantages for the WASM REPL. Building through Cargo alone is possible, but
```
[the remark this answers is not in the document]
yannham: Ah, too many moving sentences around. Initial mention to Nix must have been gone.

CONTRIBUTING.md
yannham: Let me reproduce the comments inside the corresponding build phase in our `flake.nix`:

```
# Wasm-pack requires to change the crate type. Cargo doesn't yet
# support having different crate types depending on the target, so
# we switch there
sed -i 's/\[lib\]/[lib]\ncrate-type = ["cdylib", "rlib"]/' Cargo.toml
                                                                                
# This is a hack to prevent the fs2 crate from being compiled on wasm.
# This may be able to be removed once one or more of these issues are resolved:
# https://github.com/bheisler/criterion.rs/issues/461
# https://github.com/rust-lang/cargo/issues/1596
# https://github.com/rust-lang/cargo/issues/1197
# https://github.com/rust-lang/cargo/issues/5777
sed -i '/nickel-lang-utilities/d' Cargo.toml
```

The related issue for the first problem is e.g. https://github.com/rust-lang/cargo/issues/4881. May be useful to add it in the comments, by the way.

yannham: Building the crate in "normal" mode and for WASM requires two different incompatible configurations (the crate type), and Cargo doesn't allow to have this configuration depends on the target currently.

For the second patch, it's related to `utilities` not being used for the WASM build whatsoever (it's just a bunch of helpers for tests and benchmarks), but cargo still insist on building it in WASM because it appears in the dependencies. But `utilities` has some filesystem related helpers that doesn't please the WASM build.

CONTRIBUTING.md
yannham: It is just an arbitrary convention, I'm afraid. We started with just the boolean version, and slapped contracts later to have better location for failing tests. In some sense, the current situation is rather than the boolean thing is the actual specification as far as `cargo test` is concerned, and the contract thing is a bonus for better debugging.

CONTRIBUTING.md
yannham: The idea was to avoid having contributor wasting their time on something that has no chance to land, but maybe this has been already made clear a the beginning. I've removed it.

yannham:
I implemented the suggestions and moved the setup development environment part to a separate `HACKING.md`.

CONTRIBUTING.md
aspiwack: How about maintaining to `cargo.toml` files? (the second one is called, say `cargo.wasm.yaml`) Would that be possible?

yannham: A third possibility is to have a separate branch `wasm`, that follows `closely` with the patched manifest. We'll have to keep the branch in sync, but that may be the less annoying solution for users.

aspiwack: No, a separate branch sounds awful. Having two nearly identical cargo file sounds merely annoying (it can be solved easily after CI). But Cargo not letting me switch between files basically kills this idea (we can make two different cargo files, and use a wrapper around cargo which symlinks whichever file is relevant for the build; this is terrible, let's not).

The shell script is not a great solution. Not even a good solution. But it's not a bad solution either. So we should probably go with that. It still runs the risk of people committing the wasm `cargo.yaml` by mistakes, though. But it's not too hard to deal with.

What do other project which compile to wasm do? Do they patch the `cargo.toml` file as well?

yannham: After digging around a bit, I find a lot of people complaining but not many better solutions. There is even a cargo plugin to do the dirty patching automatically: https://crates.io/crates/cargo-crate-type. In some situation you can specify both crate types `["cdylib", "rlib"]`, but now I remember that I tried this trick and it broke macOS build because something is not fully supported there. I gave a look at some repositories that compile to WASM but it's not always clear to see what's happening on big multi-crate repo to be honest. Some use other build system on top of Cargo (NPM, thunk, Makefile), some use emscripten instead of wasm-pack as a WASM toolchain. I suspect some of them just have a separate crate with additional code for the WASM version anyway, beside the `crate-type` difference.

One technique mentioned is to have a proxy crate, say `nickel-lang-wasm` that is just wrapping `nickel-lang` as a dependency and change the crate type to `cdylib`. It's possible that it'll also solve the problem of `utilities` being compiled, because it's a dev-dependency and I guess that shouldn't propagate to the wrapping crate. That may be a less awful solution to try.

aspiwack: > One technique mentioned is to have a proxy crate, say `nickel-lang-wasm` that is just wrapping `nickel-lang` as a dependency and change the crate type to `cdylib`. It's possible that it'll also solve the problem of `utilities` being compiled, because it's a dev-dependency and I guess that shouldn't propagate to the wrapping crate. That may be a less awful solution to try.

This very much sounds like progress, actually. Let me know whether this works. (I'll look at the rest of your changes independently, no need to hold this PR for that. But I'm happy to see that we have potential solutions).

garbas (approved):
I gave it a quick read but I didn't test all of the command. The whole thing looks very professional :) 

Would it be possible to include this on the website? Or at least point to this markdown files.

yannham:
I haven't had time to try to improve the WASM situation. This guide is still useful, so I propose to merge it as it is and open an issue to remember to try a better solution for building the WASM REPL.

@garbas given this is markdown and part of the nickel flake, it wouldn't be hard to either include it as a page or point to it. I wonder where that should go, though? It doesn't concern random prospective users visiting the website, but it may fit somewhere indeed.

garbas:
> @garbas given this is markdown and part of the nickel flake, it wouldn't be hard to either include it as a page or point to it. I wonder where that should go, though? It doesn't concern random prospective users visiting the website, but it may fit somewhere indeed.

It is fine that this is here. It would be trivial to just copy this two files during the build process of documentation.

yannham:
Sorry, I didn't make my question clear enough: I meant where should it appear on the website, as a page or a link.

garbas:
I think having a link with some short description would be a good first step for now.

aspiwack:
I really doubt that it has value to add either document to the website.

PR #681, opened 2022-04-20
Recursive let

### LINKED ISSUE #525: Add recursive Let-bindings
There is no way currently to define recursive functions using let-binding alone. One has to work around this using records. Example:

```
let fact = {
 go = fun n => if n == 0 then 1 else go (n - 1)
}.go in
fact 10
```

Add `let-rec`, so that the following works:

```
let rec fact = fun n => if n == 0 then 1 else fact (n - 1) in
fact 10
```

Alternatively, we could make all let-bindings recursive (see below).

We could alternatively make all let-bindings recursive. However, for most use cases of a configuration language, recursive let-bindings should be rare, if needed at all. Further, there is a cost to recursive let-bindings. Code that would fail at scope checking time could instead be unintentionally recursive. It also makes shadowing the same variable repeatedly in a chain of let-bindings not work the way some users would expect, e.g. `let x = 1 in let x = x + 1 in x` would need to be rewritten as `let x = 1 in let x' = x + 1 in x'`.

A downside of explicit recursion is that it's not consistent with records, which are implicitly recursive.

ErinvanderVeen:
Resolves #525. Unfortunately, this PR adds "rec" as a keyword, meaning that tests that use "rec" as a shorthand for "record" were  changed to use "r" instead.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -328,7 +328,8 @@ Examples:
 ```
 
 ### Let-In
-Let-in allows the binding of an expression. It is used as `let <ident> = <expr> in <expr>`.
+Let-in allows the binding of an expression. It is used as `let <rec?> <ident> = <expr> in <expr>`.
+The `rec` keyword in Let-in constructs allows the let binding to become recursive, enabling the use of the `<ident>` within the first `<expr>`.
 
 Examples:
 ```
@@ -340,6 +341,15 @@ true
 
 > let a = 1 in let b = 2 in a + b
 3
+
+> let rec f = fun n => if n == 0 then n else n + f (n - 1) in f 10
+55
+
+> let rec fib = fun n => if n <= 2 then 1 else fib (n - 1) + fib (n - 2) in fib 9
+34
+
+> let rec repeat = fun n x => if n <= 0 then [] else repeat (n - 1) x @ [x] in repeat 3 "foo"
+["foo", "foo", "foo"]
 ```
 
 ## Functions
--- a/src/parser/utils.rs
+++ b/src/parser/utils.rs
@@ -7,7 +7,10 @@ use std::rc::Rc;
 
 use codespan::FileId;
 
+use super::error::ParseError;
+
 use crate::{
+    destruct::Destruct,
     identifier::Ident,
     label::Label,
     mk_app, mk_fun,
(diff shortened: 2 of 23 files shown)


src/term.rs
```diff
@@ -185,6 +185,12 @@ pub struct RecordAttrs {
     pub open: bool,
 }
 
+#[derive(Debug, Default, Eq, PartialEq, Clone)]
```
yannham, line 188: This structure could have a short description, at least of the fields (`binding_type` and `rec`).

tests/unbound_type_variables.rs
```diff
@@ -57,4 +57,8 @@ fn unbound_type_variables() {
                        f null",
         "a"
     );
+    assert_unbound!(
+        "let rec f = fun x => if x == 0 then x else f (x - n) in f null",
+        "n"
+    );
```
yannham, line 63: This test is succeeding a bit by accident, I think. This file is testing for unbound *type variables*, that is unbound variables that happens in a type position (right hand side of a type or contract annotation). Here there is no annotation whatsoever, but because some unbound type variables manifest as normal unbound variables, the macro `assert_unbound!` is liberal enough to make it work.
Beside this test being useful or not, it shouldn't be here. Additionally, I'm not sure it is testing something new that isn't already tested (there is no reason recursive lets would make unbound variables to be suddenly bound). I would suggest to just get rid of it.

To see if the `free_var` transform is doing its job correctly on recursive lets, you can add new test cases in  `tests/free_var.rs`.

tests/typecheck_fail.rs
```diff
@@ -230,3 +230,10 @@ fn piecewise_signature() {
         Err(TypecheckError::TypeMismatch(..))
     );
 }
+#[test]
+fn recursive_let() {
+    assert_matches!(
+        type_check_expr("let rec f : Num -> Num = fun x => f \"hoi\" in null"),
+        Err(TypecheckError::TypeMismatch(..))
+    );
+}
```
yannham, line 239: It's good to add failing tests, but passing tests are nice too :slightly_smiling_face: I would suggest to add a few tests in `pass/typechecking.ncl` to check that recursive functions are correctly typechecked, including polymorphic recursive functions, something like:

```nickel
let rec f : forall a. a -> Num -> a = fun x n => if n == 0 then a else if f "0" n == "1" then f x (n - 1) else f x (f 1 n)
```

doc/manual/syntax.md
```diff
@@ -340,6 +341,15 @@ true
...
 3
+
+> let rec f = fun n => if n == 0 then n else n + f (n - 1) in f 10
+55
+
+> let rec fib = fun n => if n <= 2 then 1 else fib (n - 1) + fib (n - 2) in fib 9
+34
+
+> let rec repeat = fun n x => if n <= 0 then null else { head = x, tail = repeat (n - 1) x } in repeat 3 "foo"
+{ head = "foo", tail = { head = "foo", tail = { head = "foo", tail = null }}}
```
yannham, line 352: I think using built-in arrays instead of encoding linked lists through records could read better for a neophyte going through the manual.
```suggestion
> let rec repeat = fun n x => if n <= 0 then [] else repeat (n - 1) x @ [x] in repeat 3 "foo"
["foo", "foo", "foo"]
```

yannham (changes requested):
The implementation looks good, but this PR could benefit from a tad more tests and documentation, see inline comments.

src/parser/utils.rs
```diff
@@ -314,6 +317,42 @@ pub fn mk_label(types: Types, src_id: FileId, l: usize, r: usize) -> Label {
     }
 }
 
+/// Create a let/let pattern term. If the parsed let statement has a pattern we check if the term
+/// is also parsed as being recursive, we throw an error. This error is thrown because Nickel does
+/// not implement recursive let patterns, although it might in the future.
+///
+/// If no pattern was found, an, optionally recursive, let binding is constructed.
```
yannham, line 324:
```suggestion
/// Generate a `Let` or a `LetPattern` (depending on `pat` being empty or not) from a the parsing
/// of a let definition. This function fails if the definition has both a non-empty pattern and
/// is recursive (`pat != Destruct::Empty && rec`), because recursive let-patterns are currently
/// not supported.
```

yannham (approved):
For the free variable tests, I was thinking about adding something like:

```rust
#[test]
fn recursive_let() {
    assert!(check_stat_vars(
        "{
          a = let rec b = b + a + h in b,
          b = let rec a = a + b in f (a + 1) z,
          c = let rec foo = b in let rec bar = c in if a.r then [b] else {foo = c}
        }",
        HashMap::from([
            ("a", vec!["a"]),
            ("b", vec!["b"]),
            ("c", vec!["a", "b", "c"])
        ])
    ));
}
```

in `tests/free_var.rs`. This checks that the free_var transform correctly handles the case of recursive lets bindings.

francois-caddet (approved):

PR #683, opened 2022-04-22
Feature/pretty print

francois-caddet:
Pretty print of the AST (implemented as a method of `RichTerm`)

Will help for debuging stufs


src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+                } else {
+                    allocator.nil()
+                })
+                .nest(2)
+                .append(allocator.line())
+                .group()
+                .braces(),
+            RecRecord(
+                fields,
+                inter_fields, /* field whose name is defined by interpolation */
```
yannham, line 243: it's a detail, but `inter_fields` is called `dyn_fields` elsewhere in the codebase for dynamically defined fields. Unless you specifically care that this one be called `inter_fields`, it's nice to keep naming consistent.

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+                            allocator.nil()
+                        }
+                        .append(allocator.nil()),
+                    );
+                    rt = t;
+                }
+                allocator
+                    .text("fun ")
+                    .append(allocator.intersperse(params, allocator.space()))
+                    .append(allocator.text(" =>"))
```
yannham, line 140: Sometimes you use literal spaces as here, and sometimes you use `allocator.space()`. Is there a reason to not do:

```rust
.append(allocator.space())
.append(allocator.text("=>"))
```

Here?

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+                    allocator
+                        .intersperse(rts.iter().map(|rt| allocator.atom(rt)), allocator.line())
+                        .nest(2),
+                )
+                .group(),
+
+            Sym(sym) => allocator.text(format!("<symbol: {}>", sym)),
+
+            // TODO
+            Wrapped(i32, RichTerm) => unimplemented!(),
```
yannham, line 355: Same. Better just print `<wrapped>` than panic here.

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+
+            MetaValue(mv) => mv.to_owned().pretty(allocator),
+            Import(f) => allocator
+                .text("import ")
+                .append(allocator.as_string(f.to_string_lossy()).double_quotes()),
+            ResolvedImport(id) => allocator.text(format!("import <file_id: {:?}>", id)),
+            ParseError => allocator.text("## PARCE ERROR! ##"),
```
yannham, line 362: Let's be consistent and use the same `<`/`>` for special constructs.

```suggestion
            ParseError => allocator.text("<parse error>"),
```

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+            Div() => allocator.text("/"),
+            Modulo() => allocator.text("%"),
+
+            Eq() => allocator.text("=="),
+            LessThan() => allocator.text("<"),
+            GreaterThan() => allocator.text(">"),
+            GreaterOrEq() => allocator.text(">="),
+            LessOrEq() => allocator.text("<="),
+
+            Merge() => allocator.text("&"),
```
yannham, line 69:
```suggestion
            Merge() => allocator.text("&"),
            
            ArrayConcat() => allocator.text("@"),
```

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+
+        match self.as_ref() {
+            Null => allocator.text("null"),
+            Bool(v) => allocator.as_string(v),
+            Num(v) => allocator.as_string(v),
+            Str(v) => allocator.as_string(v),
+            StrChunks(chunks) => allocator
+                .text("m%\"")
+                .append(allocator.intersperse(
+                    chunks.iter().map(|c| {
```
yannham, line 93: You need to reverse chunks, as they are stored in the the reverse direction (as a stack of chunks to be popped). For example, the current implementation gives:

```console
$ nickel expend <<< ' "unkown flag %{flag}" '
m%"%{flag}
unkown flag "%m
```

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+                        match c {
+                            crate::term::StrChunk::Literal(s) => allocator.as_string(s),
+                            crate::term::StrChunk::Expr(e, i) => allocator
+                                .text("%{")
+                                .append(e.to_owned().pretty(allocator))
+                                .append(allocator.text("}")),
+                        }
+                    }),
+                    allocator.line_(),
```
yannham, line 102: You don't want to insert new lines that don't exists inside strings, as that changes the semantics.

src/pretty.rs
yannham: Also, you may want to special case when `chunks` is of size 1 to use standard string delimiters, otherwise `m%"` is always used, even for small strings.

yannham (changes requested):
It's working already quite well, but I had issues with strings in particular (see inline comments).

If-then-else are also rendered in strange way: sometimes the `else` is indented more than the `if`, sometimes it's not. Maybe it's due to how the `if` is printed in a strange way: first `if` alone, then `then case_1 else` that is grouped, then the last term, which may mess up the layout.

Nitpick: empty lists are printed with two spaces inside.

Types are printed sometimes with additional braces:

```console
$ nickel expend <<< '1 | forall a b c. Num -> Num -> (Num -> a) -> {_ : Dyn}'
1 | forall a b c. Num -> Num -> (Num -> a) -> {{ _: Dyn }}
```

src/bin/nickel.rs
```diff
@@ -31,6 +31,17 @@ struct Opt {
 /// Available subcommands.
 #[derive(StructOpt, Debug)]
 enum Command {
+    /// Expend the nickel source after requiered transformation.
+    /// Could be used for debugging.
+    /// By default only format the code with comments droped.
+    Expend {
+        /// Output file. Standard output by default
+        #[structopt(short = "o", long)]
```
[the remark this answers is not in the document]
francois-caddet: Yes it's actually useless for a one file. But we may want to perform the printing for the imported files also. So the output would be a dir. Not sure it's a potential use case. What do you think?

src/pretty.rs
```diff
@@ -0,0 +1,471 @@
...
+            Null => allocator.text("null"),
+            Bool(v) => allocator.as_string(v),
+            Num(v) => allocator.as_string(v),
+            Str(v) => allocator.as_string(v),
+            StrChunks(chunks) => allocator
+                .text("m%\"")
+                .append(allocator.intersperse(
+                    chunks.iter().map(|c| {
+                        match c {
+                            crate::term::StrChunk::Literal(s) => allocator.as_string(s),
```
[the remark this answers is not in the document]
francois-caddet: Only escaping `"\%"` should work right?
Should I try to count consecutives `%` signs in `Literal`s to adapt the `%` quantity of interpolation as it could be done to avoid having to escape characters? May be a futur improvement.

src/pretty.rs
francois-caddet: It's equivalent actually. Less to write ;). But you're right, the second version is more clear.
The `space()` is a simple space, writing it as a function make it easy to replace by another type of blank. Not absolutely sure but I think it's the only reason. But yes it would be more clear to use the `space` function everywhere.

src/pretty.rs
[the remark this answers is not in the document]
yannham: Your proposal sounds maybe like a good thing to do, but indeed is a bit more involved, so let's postpone it to a later iteration indeed. Escaping with `\` doesn't work in multiline strings, so the answer to your first question is it depends. If you print something within double quotes, yes, `\%` suffices, but in a multiline string you need to use interpolation to print a `%`, as in `%{"%"}`.

src/pretty.rs
yannham: Do they follow the same rules about breaking, indentation and such? I would expect `space()` to have a specific semantics (like this can be replaced by a newline), but I don't know, I'm just asking.

francois-caddet: It's actualy `line()` and `softline()` wich can be either a line or a space. `space()` is only a space if I don't mistake.

francois-caddet:
Ready to merge.

2 futur improvements to look in:
- Patterns destructuring is prettyprinted only for the simple form `let {a, b, c} = ...`
- do not use the `LetAttr` it mean that recursive lets are printed as standard lets but may cause others issues.
- make better tests

Issue #584, opened 2022-01-20
Implement a proper bidirectional algorithm


yannham:
**Is your feature request related to a problem? Please describe.**
The current implementation of the typechecker is more or less based on a bidirectional typechecking algorithm, but mixes both inference, checking and a third nickel-specific mode, the non-strict mode (where we just walk the AST looking for the next statically typed block to handle). The information flow is not always clear, the algorithm is hard to extend, and some edge cases are not handled well (higher rank types checking and polymorphic type instantiation, mainly). A proper bidirectional algorithm would make it easier to add subtyping as well, with the relation induced by `T <: Dyn`, for example. The walking phase is also wasteful: we continue to generate unification variable, but just don't actually unify them. 

**Describe the solution you'd like**
Have a formal bidirectional formulation of the type system on paper. Implement it and break the `type_check_` function into three mutually recursive functions (or, constraints generating function later, if this is too stack-hungry), `walk`, `infer` and `check`, with clear roles and information flow.

**Additional context**
Such an enterprise was already tried in #392, but it hit some non-trivial issues with polymorphic instantiation and order of arguments (dubbed the "tabby-first problem" in section 5.1 of [Krishnaswami and Dunfield](https://arxiv.org/abs/1908.05839)). A first step could be to forget about `Dyn`-induced subtying for now, and adapt [Complete and Easy Bidirectional Typechecking for Higher-Rank Polymorphism](https://arxiv.org/abs/1306.6032), which doesn't have this issue.

yannham:
I'm setting the milestone to 1.0, which doesn't mean "implement RFC004", but rather refactor and clarify a bit the algorithm around clear notion of inference and checking modes.

yannham:
Post-poned to after 1.0. Variable levels are non trivial to implement and the feature is not critical.

PR #691, opened 2022-04-28
[RFC004] Type system RFC + first draft of specification

yannham:
Step toward #584.

Contain an RFC to describe the rationale of the proposal and an OTT spec. `nix build` builds the final PDF from the OTT specification.


rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+# would fail, a is instantiated to Num, but foo is not of type Num
+uni_pair 1 foo : _
+```
+
+This problem is dubbed the Tabby-first problem in [the review of bidirectional
+typing of Dunfield and Krishnaswami](https://arxiv.org/abs/1908.05839).
+Surprisingly, it doesn't seem to have been deeply studied (outside of the
+subtyping relation induced by polymorphism). For subtyping induced by
+polymorphism, there are different trade-offs: Dunfield and Krishnaswami
+restricts their system to impredicative polymorphism, which is then
```
aspiwack, line 407:
```suggestion
restricts their system to predicative polymorphism, which is then
```
I think.

rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+order-independent [^3]. In exchange, the arrow type has the expected variance
+(that is, eta-expansion preserves typeability). On the other hand, Haskell
+deemed impredicative polymorphism to be more important and chose to abandon the
+variance of arrows and other type constructors. Coupled with a so-called
+QuickLook phase, Haskell is able to most common impredicative instantiations.
```
aspiwack, line 412: Missing verb.

rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+We can't apply the Haskell approach naively, because in the presence of `Dyn`,
+an impredicative instantiation is not the only possibility anymore, even if we
+don't do deep instantiation or deep skolemization:
+
+```nickel
+let foo : Dyn = null in
+let id : forall a. a -> a = fun x => x in
+let concat : forall b. Array b -> Array b -> Array b = (@) in
+
+concat [id] [null]
```
aspiwack, line 423: I don't this example is correct. This you can absolutely instantiate `concat [id]` with `b = Bool -> Bool`.

An example that would work is, I believe
```suggestion
let foo : Dyn = null in
let aid : Array (forall a. a -> a) = [fun x => x] in
let concat : forall b. Array b -> Array b -> Array b = (@) in

concat aid [null]
```

rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+
+We may extend the QuickLook phase to also cover `Dyn`, but this needs more
+investigation. Another fix could be to make all type constructors invariant also
+for the subtyping relation generated by `Dyn`, but that is quite restrictive in
+practice (several examples in this document would need `Array T <: Array Dyn`,
+to get rid of the casts).
+
+### Packing up
+
+To sum up (**TODO**: better wording):
```
aspiwack, line 443: I think that the wording is fine
```suggestion
To sum up:
```

rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+The final proposal is a system that:
+
+- Has subtyping generated by the dynamic type and dictionaries.
+- Support higher-rank polymorphism
+- Is predicative
+- Doesn't do deep skolemization/instantiation, or equivalently, in which type
+  constructors are invariant with respect to subtyping induced by polymorphism
+
+A declarative specification as well as a constraint-solving algorithm are
+provided in the file [type-system.pdf](./type-system.pdf).
```
aspiwack, line 468: This link is dead. And I don't think that we want to host a pdf in the git repository anyway. We will have to find a place to host the generated pdf, probably on the website?

rfcs/004-typechecking.md
```diff
@@ -0,0 +1,496 @@
...
+so doesn't seem to require new developments or very fancy types. The current
+implementation, which supports polymorphism and row polymorphism, appears to
+already cover most cases.
+
+However, the co-existence of statically typed code and dynamically typed code
+makes Nickel different from most of the aforementioned inspirations. While we
+often brand Nickel as a gradually typed language for simplicity, it isn't
+really, technically speaking. A cornerstone of _gradual type systems_, derived
+from the original work of Siek and Taha[^1], is to statically accept potentially
+unsafe conversions from and to the dynamic type `Dyn`, and more complex types
```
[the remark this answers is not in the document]
yannham: I usually take Gradual typing to mean the system of Siek and Taha (and derived systems), but Typed Racket is a fair example. Will update.

rfcs/004-typechecking.md
yannham: Ah, because the typechecker would eagerly instantiate `id` to `?a -> ?a` when writing `[id]`, giving `[?a -> ?a]`, allowing it to be unified with a concrete type?

aspiwack: Right.

spec/type-system/notes-richard-2.md
```diff
@@ -0,0 +1,44 @@
+# Call w/ Richard #2
```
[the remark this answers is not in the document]
yannham: Initially yes, for backup purpose, but I've moved them elsewhere now.

yannham:
@aspiwack beside the document content in itself, any thought on which direction to take (amongst the ones proposed in this RFC) with respect to subtyping?

aspiwack:
It doesn't seem that it goes into much more details than what we've discussed previously. If there are options outlined, I don't see them, what am I missing?

yannham:
I would say the packing up + alternatives section. Either going for subtyping, no subtyping, or subtyping for dicts only, very broadly.

aspiwack:
I see. I must admit that the dict-only subtyping is an attractive alternative. But it's not fully an alternative: it's on the path to `Dyn` subtyping. Let's still plan for `Dyn` subtyping for the time being, but let's make sure to implement dict-only subtyping first, and see where it gets us. It commits us to much less. We can revisit the plan then, based on experience gathered, and use-cases.

yannham:
The specification is not complete or fully well-rounded, but I think it's worth merging this PR for the RFC. I added a warning in the README of the specification, and I think the status of each part has been made clear enough that we can merge this in the main repo.

PR #693, opened 2022-04-29
[RFC003] Nix-nickel (Draft, WIP) 

yannham:
[Rendered](https://github.com/tweag/nickel/blob/rfc/nix-nickel/rfcs/003-nix-nickel.md)

### DIFF
--- a/rfcs/003-nix-nickel.md
+++ b/rfcs/003-nix-nickel.md
@@ -32,7 +32,7 @@ flakes and modules.
 ### Constraints
 
 Because Nix is a distinct project from Nickel which has existed for quite some
-time now, we need operate under the following constraints:
+time now, we need to operate under the following constraints:
 
 - _Do not require unreasonable changes to Nix itself_. While it's probably
     illusory to avoid any change in Nix at all, we must strive to keep them
@@ -51,8 +51,8 @@ time now, we need operate under the following constraints:
 - _Do not lock ourselves in the current Nixpkgs architecture_. Nixpkgs had
     made a number of design choices that are, in hindsight, not optimal (stdenv,
     describing packages as functions instead of data, etc.). While we have to
-    keep some form of backward-compatibility, we want to do so while avoiding to
-    tie ourselves to the current design and implementation. In other words, this
+    keep some form of backward-compatibility, we want to do so while avoiding
+    tying ourselves to the current design and implementation. In other words, this
     document must propose a solution that stay compatible with a future radical
     departure from e.g. the Nixpkgs architecture.
 
@@ -65,7 +65,7 @@ goals and constraints:
 - **Flakes**: new format for decentralized and composable packages.
 - **NixOS modules**: NixOS system configuration.
 
-In the long term, we aims at handling all those cases, but the scope of such an
+In the long term, we aim at handling all those cases, but the scope of such an
 undertaking appears very large for a single RFC. We decide to focus on the first
 item: writing derivations and Nixpkgs-style packages.
 
@@ -88,7 +88,7 @@ possibilities, so we choose to include it as well.
 
 ### Interaction with Nixpkgs
 
-There is a bunch of scenarios that require leveraging Nixpkgs:
+There are a bunch of scenarios that require leveraging Nixpkgs:
 
 - **PKG**: Using a package from Nixpkgs, for example as a dependency
 - **LIB**: Use one of the myriad of helpers from Nixpkgs: `mkShell`,
@@ -101,10 +101,10 @@ JSON.
 
 **LIB** is more involved. Some of Nixpkgs helpers take either functions, but the
 sole laziness of expressions means that an interface (in the form of a FFI)
-between Nix and Nickel would needs to handle going back and forth between the
+between Nix and Nickel would need to handle going back and forth between the
 two languages transparently.
 
-**OVD** is technically not very different than **LIB**, since it mostly amounts
+**OVD** is technically not very different from **LIB**, since it mostly amounts
 to calling Nix functions like `override`, `overrideAttrs`, etc (for any override
 that doesn't operate directly on a derivation, which most are). That is, solving
 **LIB** would solve **OVD** as well.
@@ -264,7 +264,7 @@ dependencies as record fields without definition:
 }
 ```
 
-The contract may be optional as the infrastructure of an hypothetical Nickelpkg
+The contract may be optional as the infrastructure of a hypothetical Nickelpkg
 would already apply the contract Package.
 
 <!-- In this design, it would be nice to avoid having to repeat those
@@ -322,7 +322,7 @@ derivations.
 
 However, this only concerns a few top-level attributes. Other fields of the
 derivation, such as `version`, `license`, etc. are still hidden under the output
-function. Flakes are not a alternative to Nixpkgs, but a schema working on top
+function. Flakes are not an alternative to Nixpkgs, but a schema working on top
 of it. Building derivations in a flake still relies on Nixpkgs mechanisms like
 `mkDerivation`, and the overriding mechanisms are the same as well. Flakes
 solves related but distinct issues (package composition and reproducibility).
@@ -379,8 +379,8 @@ Compilation could be done on the fly, as the compilation process would be rather
 straightforward, or have a Nixpkgs snapshot pre-compiled to Nickel, or a mix of
 both.
 
-This solution requires to reimplement Nix builtins in Nickel (the nickel-nix
-compatibility layer) . This is an interesting milestone in itself, because even
+This solution requires reimplementing Nix builtins in Nickel (the nickel-nix
+compatibility layer). This is an interesting milestone in itself, because even
 without a Nix-to-Nickel compiler, the compatibility layer would already make
 writing derivations in pure Nickel possible.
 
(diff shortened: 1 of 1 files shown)


MagicRB:
Oh my! It's here

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,330 @@
...
+
+Thus, tackling derivations first makes sense. From there, Nixpkgs is probably
+the thinnest layer over derivations of the three, and already open a lot of
+possibilities, so we choose to include it as well.
+
+## Challenges
+
+### Interaction with Nixpkgs
+
+There a bunch of actions that require leveraging Nixpkgs:
```
yannham, line 91:
```suggestion
There is a bunch of scenarios that require leveraging Nixpkgs:
```

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,330 @@
...
+      name = "%{name}-%{version}";
+      inherit builder args;
+    })
+    |> builtin.derivation
+}
+
+# ... other definitions building on this one
+
+# actual package
+pkgs.hello = builders.unix_package & {
```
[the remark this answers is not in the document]
yannham: True, and flakes are one example I think, where `output` is still a function but that's fine because the other metadata are one level up and therefore accessible without having to evaluate the output function. And back to our example, evaluating a merge and a simple function (which is morally a parametrized record) shouldn't make a big difference in practice.

One reason I could see to do that would be for consistency, as using both functions and merging may end up being confusing to the user, in particular which one is to be used in which case.

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+
+  buildInputs = if enable_gui then [ super.gtk ] else [],
+
+}
+```
+
+**TODO**: other solutions? The pkg subfield, but seems like a lesser version of
+the first proposal. Bazel-like includes? From experience, the semantics is
+confusing, it's tied to the filesystem structure. And in the end it's still a
+strange form of dynamic scoping: if we are to do it, let go for a proper version.
```
thufschmitt, line 305: I agree with the confusing-ness of bazel targets, **but** they have a very strong point for them in that because since they are at a meta-language level, they are only resolved at the very last moment, meaning that
1. The actual path of the dependency really stays fully opaque, which
    1. Is more robust. For example, there was a couple of occurrences of code manipulating the output path of a dependency in nixpkgs, and that broke with CA derivations so I had to change it − like https://github.com/NixOS/nixpkgs/pull/124227
    2. (potentially) makes the final UX nicer because you don’t have ugly store paths leaking everywhere whenever you want to do something
2. Evaluating a target doesn’t require evaluating its dependencies (meaning that the whole discussion about packages-as-data vs packages-as-functions doesn’t even have to exist)

We anyways couldn’t do something like what bazel does (because in the Nix model you do need to resolve the paths of your dependencies to generate the build script, which isn’t the case in bazel), but maybe an in-between solution could be interesting

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+straightforward, or have a Nixpkgs snapshot pre-compiled to Nickel, or a mix of
+both.
+
+This solution requires to reimplement Nix builtins in Nickel (the nickel-nix
+compatibility layer) . This is an interesting milestone in itself, because even
+without a Nix-to-Nickel compiler, the compatibility layer would already make
+writing derivations in pure Nickel possible.
+
+- TODO: what about dynamic imports?
+- TODO: what about using a Nickel pkgs from Nix?
```
thufschmitt, line 388: Unless (until?) we can get to a point where the Nix evaluator is just Nickel with a Nix frontend (in which case that problem won’t exist anymore because Nix could just call Nickel code transparently), I think that could be a limiting factor for the adoption of Nickel − at least for open-source projects, but these are the one with the most visibility. So we’ll want to have a decent story around this too

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+
+#### Using Nixpkgs in the PARM
+
+Using a package from Nixpkgs in the PARM with a Nix-to-Nickel compiler would be
+straightforward, as we could evaluate anything to a derivation whenever needed.
+One would need to use Nixpkgs functions and idioms to do e.g. overrides though,
+but this seems pretty hard to avoid, as a systematic translation from the
+Nixpkgs model to the PARM doesn't seem trivial, if even doable.
```
thufschmitt, line 396: I don’t think that’s a bad thing in the long term tbh. Having an interface that would be too transparent could cause the path of least resistance to lead to a weird spaghetti of Nix and Nickel code intertangled, which would probably be a very bad thing.

At least, having some annoying constraints at the boundaries (even if just because both worlds have different idioms) mean that people will care about these boundaries, and things will stay clearly separated

rfcs/003-nix-nickel.md
yannham: As mentioned during the meeting, it seems going from a data representation to a function representation is easier, so we can hope for a simple generic wrappers around a Nickel package. But that needs to be investigated further before proceeding.

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+  # Interface
+
+  name
+    | doc "Name of the derivation, used in the Nix store path."
+    | Str,
+
+  version
+    | doc "Version of the derivation, used in the Nix store path."
+    | Str,
+    | default = "",
```
milahu, line 160: comma

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+  src = builders.fetchurl & { url = ..., sha256 = ... },
+
+  buildInputs = if enable_gui then [ gtk ] else [],
+}
+```
+
+Thanks to laziness, `nix` could extract fields like `name` or `version` directly
+without having to provide inputs or to evaluate `drv`.
+
+We wall call this approach the **PARM** (**P**ackage **A**s **R**records
```
milahu, line 217: wall -> will

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+##### Explicit list
+
+One possibility is to specify explicitly all the
+dependencies as record fields without definition:
+
+```nickel
+# file hello.ncl
+{
+  # dependencies
+  gtk | NickelPackage,
```
milahu, line 248: this could allow inversion of control, where a child package can modify a parent package

```nickel
# file hello.ncl
{
  # dependencies
  gtk | NickelPackage = {
    version = "2.3.4", # bottom-up pinning. nixpkgs: gtk_2_3_4
    enable_somefeature = true, # require feature from parent
    # nixpkgs: hello = callpackage ./hello.nix { gtk = gtk.override { enable_somefeature = true; }; };
  },
```

problem: pollution &rarr; move to scope

```nickel
# file hello.ncl
{
  pkgs = {
    gtk | NickelPackage,
  },
```

rfcs/003-nix-nickel.md
yannham, line 217:
```suggestion
We call this approach the **PARM** (**P**ackage **A**s **R**records
```

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+
+  # Interface
+
+  name
+    | doc "Name of the derivation, used in the Nix store path."
+    | Str,
+
+  version
+    | doc "Version of the derivation, used in the Nix store path."
+    | Str,
```
yannham, line 159:
```suggestion
    | Str
```

rfcs/003-nix-nickel.md
yannham: Good point. Although this could also be solved by merge priorities, the Nickelpkg layer erasing the local definitions. But it does sound fragile. Another issue is that having everything living at the same level forces all the versions of gtk to be the same in the final fixpoint, preventing local overrides (we want `this` package to have `python=python27` but `that` package to have `python=python3` for example).

For those reasons, I think your first suggestion sounds like the best way forward. Also, it could avoid having to repeat dependencies as say `inputs` and `build_inputs`: we could kill two bird with one stone and declare both inputs for Nix and the "merging interface" for Nickel as in

```nickel
{
  inputs = {
    gtk,
  },

 build_inputs = {
    foocompiler,
  }

  # ... inputs.gtk ...
}
```

milahu: bikeshed:
inputs sounds too general, could be packages and options (enable_somefeature)
packages is too long
pkgs is short & sweet (but also used in nixpkgs)

or we identify "package scopes" by type

```
{
  p | Pkgs = {
    gtk,
  },
  rundeps | Pkgs = [
    apkg, bpkg, cpkg, dpkg,
  ],
  drv = {
    build = m%"
      cp %{p.gtk}/lib/libgtk.so $out/lib/
      echo PATH=%{lib.makePath rundeps} >$out/bin/wrapper.sh
    "%m,
  },
}
```

such short names would compensate for nickel's missing `with` feature

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+  derivation, etc.), which should be fine.
+
+Compilation could be done on the fly, as the compilation process would be rather
+straightforward, or have a Nixpkgs snapshot pre-compiled to Nickel, or a mix of
+both.
+
+This solution requires to reimplement Nix builtins in Nickel (the nickel-nix
+compatibility layer) . This is an interesting milestone in itself, because even
+without a Nix-to-Nickel compiler, the compatibility layer would already make
+writing derivations in pure Nickel possible.
```
steshaw, line 385: If it's a lot of work to reimplement builtins, I was wondering whether it may be possible to leverage the builtins within Nix itself (i.e. the executable / C++ source code) by using Nickel as a library in Nix. I'm not familiar with the Nix implementation, so it may be cumbersome, as often extension functions are tied closely to the interpreter. However, with some refactoring, a shared interface may be extracted for use in Nix proper and Nix-Nickel.

Using Nickel as a library seems like a good path for Nix in the future, whatever happens. Once that's in place, it makes sense to slowly port Nix from C++ to Rust. I believe there is already some interest in that direction.

rfcs/003-nix-nickel.md
steshaw: Are we back to considering a functional representation for Nickel pkgs?

yannham: I think what is being discussed here is having a wrapper of some sort to use a Nickel package from Nix code, but just as a legacy compatibility mechanism. Using Nickel packages from another Nickel package would use the data representation and leverage all the benefits. However, if for some reason you need to use a package defined in Nickel from a Nixpkgs-like Nix codebase, you could, thanks to (very schematically) `to_nixpkg : Package -> NixPkgLegacyFunction`.

rfcs/003-nix-nickel.md
yannham: Yes, that is a possible path forward, but it is in opposition with the constraint `_Do not require unreasonable changes to Nix itself_`. We could debate if those changes are unreasonable, but I think that for a first approach, the less changes to Nix, the better. Not only from a disturbance/community point of view, but also I fear that even once accepted, the Nix side of the changes would be very long to land given the current development process, and that would hold us back.

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+- Different kind of strings are incompatible, to avoid accidentally forgetting
+  or losing contexts. Also, we don't know a priori how to convert one context to
+  another.
+- There would be a contract to distinguish the different kind of
+  strings in order enforce the usage of e.g. Nix style contexts for writing
+  Nickel for Nix. We want to avoid users loosing or missing context unknowingly.
+
+#### G-exps
+
+Alternative: something like `g-exp`. No magic, no extension, but less ergonomic.
```
steshaw, line 445: What are `g-exprs`? Sounds like something Scheme-like (ala s-expression / fexprs). Perhaps a reference and an example of how it is less ergonomic would be helpful.

yannham: Those are the Guix equivalent of string contexts, somehow: [manual page](https://guix.gnu.org/manual/en/html_node/G_002dExpressions.html). At this point this paragraph is just draft notes, but it'll deserve a proper link indeed once finalized. At a quick glance, it's less ergonomic because:

- Use a special syntax to start a g-exp `#~` (gexp)
- Use a special syntax to interpolate other g-exp `#$` (ungexp)

As opposed to Nix where string contexts just work without having the user to even think about it or convert between strings and derivations. On the other hand, the explicitness of g-exp may actually be a good point: arguably, it makes things less magic. But it would still be one additional thing that Nix users don't need to do today. Although, I think what was written at the time is not true: something akin to g-expr in Nickel using standard syntax for strings, like interpolation `%{}`, would still need a bit of magic and an extension mechanism, probably.

steshaw: As I mentioned in IRL, my thinking is that the ideal solution is to have overloaded and customisable interpolated strings. Some languages do this such as Scala and Haskell (via quasiquotes and TH). In Scala it requires a prefix. Haskell doesn't have builtin interpolation, so quasiquotation kind of _is_ the way :man_shrugging:. I experimented a little bit with the the support in Idris the other night. It seems doable. Though I seem to have hit a limitation, misunderstanding, or bug in Idris.

steshaw: In Idris, for instance, you can overload the standard string interpolation syntax. It should be possible to make something like this work:

```idris
postInstall : PkgExp
postInstall = """
    substituteInPlace \{out}/bin/my-script \\
      --replace /usr/bin/python \{python}/bin/python \\
      --replace /usr/bin/perl "\{perl}/bin/perl -I \{libfoo}"
  """
```

I'm not fond of that backslash as it evokes the idea of a "lambda". This is simply how string interpolation is defined in Idris (strange as `$` was discarded because of it's special meaning).

In [this short experiment](https://gist.github.com/steshaw/b3916d42f49f5b4e4ecc2922513cd7ad#file-pkg-idr-L50-L57), `pkg3a` should be equivalent to the desugared `pkg3` but I ran into a snag.

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,330 @@
...
+    proposed solution should be implementable in limited time (in the order of
+    magnitude months, up to half a year). If possible, the plan should have
+    intermediate milestones enabling new capabilities in an incremental way.
+- _Backward-compatible_. We will realistically need to be able to leverage
+    existing Nix code (especially from Nixpkgs) in some way: otherwise, even in
+    the unlikely best case scenario where everybody suddenly switches to Nickel
+    right away, it'll still take up a very long time to migrate everything.
+- _Do not lock ourselves in the current Nixpkgs architecture_. Nixpkgs had
+    made a number of design choices that are, in hindsight, not optimal (stdenv,
+    describing packages as functions instead of data, etc.). While we have to
```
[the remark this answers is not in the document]
blaggacao: https://github.com/nixpkgs-architecture/issues#1

rfcs/003-nix-nickel.md
```diff
@@ -0,0 +1,460 @@
...
+  src = builders.fetchurl & { url = ..., sha256 = ... },
+
+  buildInputs = if enable_gui then [ gtk ] else [],
+}
+```
+
+Thanks to laziness, `nix` could extract fields like `name` or `version` directly
+without having to provide inputs or to evaluate `drv`.
+
+We call this approach the **PARM** (**P**ackage **A**s **R**records
```
bew, line 217: Tiny typo :) 
```suggestion
We call this approach the **PARM** (**P**ackage **A**s **R**ecords
```

PR #696, opened 2022-05-12
Add `doc` command to construct markdown for (nested) record

### LINKED ISSUE #650: Let me generate some markdown documentation
In order to consult the stdlib's documentation, I have to read its source

I'd like to be able to generate just enough markdown from the metadata of values (such as the stdlib, but not only) in other to stitch together a documentation page.

I think this could be implemented cheaply for now, piggy-backing off `nickel query` and `QueryPrinter` in `src/repl/query_print.rs`, maybe adding a flag to use `SimpleRenderer` instead of `MarkdownRenderer`. From there, it'd be fine to concatenate a big document together from multiple calls, perhaps as Nix expression.

I'm happy to provide some PRs if this is accepted.

- Formats other than Markdown would be fine, it's just ubiquitous
- We could make `nickel query` emit JSON and people could format the documentation however they want, that doesn't seem insanely helpful as the documentation strings would still be markdown
- We could start a `nickel doc` and generate full blown docs, similarly to `cargo doc`, but that seems like a lot of work with the more-structural-than-nominal approach of Nickel, and we're still dealing with some instability
- Add metadata introspection capability to the language in order to generate the documentation via exprs, I doubt that's enough of a justification though

ErinvanderVeen:
Recursively walks the record in the provided file to collect all documentation and outputs a markdown file. The paths are used to construct the headers. Any headers present in the documentation are increased to a level above the current heading level.

For example:
``nickel doc -f stdlib/array.ncl`` produces the file `.nickel/doc/array.md`.

Resolves #650

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -62,6 +62,9 @@ enum Command {
         #[structopt(long)]
         history_file: Option<PathBuf>,
     },
+    /// Generate the documentation files for the specified nickel file
+    #[cfg(feature = "doc")]
+    Doc {},
 }
 
 fn main() {
@@ -127,6 +130,8 @@ fn main() {
             }
             Some(Command::Typecheck) => program.typecheck().map(|_| ()),
             Some(Command::Repl { .. }) => unreachable!(),
+            #[cfg(feature = "doc")]
+            Some(Command::Doc { .. }) => program.output_doc(),
             None => program
                 .eval_full()
                 .map(|t| println!("{}", Term::from(t).deep_repr())),
--- a/src/program.rs
+++ b/src/program.rs
@@ -126,6 +126,12 @@ impl Program {
         report(&mut self.cache, error)
     }
 
+    /// Create a markdown file with documentation for the specified program in `.nickel/doc/program_main_file_name.md`
+    #[cfg(feature = "doc")]
+    pub fn output_doc(&mut self) -> Result<(), Error> {
+        doc::output_doc(&mut self.cache, self.main_id)
+    }
+
     #[cfg(debug_assertions)]
     pub fn set_skip_stdlib(&mut self) {
         self.cache.skip_stdlib = true;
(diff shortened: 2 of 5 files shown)


doc/array.md
```diff
@@ -0,0 +1,237 @@
...
+``` nickel
+  filter (fun x => x <= 3) [ 4, 3, 2, 5, 1 ] =>
+    [ 3, 2, 1 ]
+```
+
+## `elem`
+
+Results in true if the given element is a member of the array, false otherwise.
+
+### This is a toplevel header
```
yannham, line 219: Leftover?

src/program.rs
```diff
@@ -207,6 +217,132 @@ where
     };
 }
 
+/// Create a markdown file with documentation for the specified FileId.
+fn output_doc(cache: &mut Cache, file_id: FileId) -> Result<(), Error> {
+    // Parse the file into the cache
```
yannham, line 222: Nitpick: comments are a good thing, but some of those in this function are just very descriptive and don't add much to code understanding (parse the file in the cache: `cache.parse(file_id)?`, for every term: `for(..)`, etc.). It's a fine balance though, and a matter of taste, so please do as you see fit. Also, some do add value (like the one saying that the default settings for Comrak is essentially CommonMark).

src/program.rs
```diff
@@ -207,6 +217,132 @@ where
...
+        let file_path: &Path = cache.files().name(*file_id).as_ref();
+
+        // We are only interested in the original name
+        let file_name = file_path
+            .file_name()
+            .expect(format!("Could not get file name of: {}", file_path.display()).as_ref());
+
+        // Attempt to create the documentation directory
+        let docpath: &Path = Path::new(".nickel/doc/");
+        fs::create_dir_all(docpath).expect("Could not create documentation destination directory");
```
yannham, line 249: We should restrict `expect` and `unwrap` to things we are sure really happen (or if it does happen, a fundamental invariant is being broken and it's justified to panic). Here this case may very well happen for various reasons and it's not nice to the user to print an austere panic message. There is an `Error::IOError` variant than you can use for that.

src/program.rs
```diff
@@ -207,6 +217,132 @@ where
...
+
+        // Create pathbuf for the target markdown file
+        let mut markdown_file = docpath.to_path_buf();
+        markdown_file.push(file_name);
+        markdown_file.set_extension("md");
+
+        // Create and write the generated markdown to the file
+        let mut file = File::create(markdown_file.clone().into_os_string()).expect(
```
yannham, line 257: Same remark about `expect`.

src/program.rs
```diff
@@ -207,6 +217,132 @@ where
     };
 }
 
+/// Create a markdown file with documentation for the specified FileId.
```
yannham, line 220: Wouldn't functions from this point on be better in their own module, or at least submodule? They feel only tangentially related to `program` (which is, in itself, just a kind of interface bundling several moving parts of the codebase)

doc/array.md
```diff
@@ -0,0 +1,237 @@
+# `array`
```
yannham, line 1: Is there a reason for just having `array` but not the other stdlib modules? I guess this is a test or something.

yannham (commented):
The WASM build is failing, but it's hard to see why on the runner. You may have more logs by trying locally.

Another point is that we could add a `buildStdlibDoc` or something to the flake, to generate the latest documentation for the stdlib. But we can dot that in a subsequent PR.

ErinvanderVeen:
> The WASM build is failing, but it's hard to see why on the runner. You may have more logs by trying locally.

I think this was related to comrak not being able to target wasm. To fix this issue I have hidden everything related to documentation generation behind a "doc" feature flag.

src/program.rs
```diff
@@ -126,6 +126,12 @@ impl Program {
         report(&mut self.cache, error)
     }
 
+    /// Wrapper for [`output_doc`]
```
yannham, line 129: Would that reference work? I suspect it may refer to the very same function. Maybe you'll need `[doc::output_doc]` instead?

src/bin/nickel.rs
```diff
@@ -62,6 +62,9 @@ enum Command {
         #[structopt(long)]
         history_file: Option<PathBuf>,
     },
+    /// Generate the documentation files for the specified nickel file
+    #[cfg(feature = "doc")]
```
yannham, line 66: We may want to provide a path with a flag like `--dir` or `--out`, but that can wait for another PR.

yannham (approved):

yannham (changes requested):
Sorry I approved a bit quickly, there are still those `.expect` to get rid of.

src/program.rs
ErinvanderVeen: Both won't work since `doc::output_doc` is in a private module. I simply removed the reference and changed the documentation.

PR #698, opened 2022-05-16
Change expand to print-ast

yannham:
#683 added a new `expand` command to pretty-print the AST of a program. While it is close to a formatter, it looses too much information (newlines, spacing, parenthesis, comments) to be one, and is rather intended to be used as a debugging tool (and for error reporting, internally).

As mentioned already in https://github.com/tweag/nickel/pull/683#discussion_r858939532, I find expand to be not very explicit, too common to indicate that this features isn't really meant to be used for most people, and takes command real estate if one day we have another notion of expansion.

This PR renames `expand` to the more descriptive and purposingly technical `print-ast`. This PR also adds `s` to action verbs to harmonize command descriptions, because Clap does by default for e.g. `help`.

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -20,7 +20,7 @@ struct Opt {
     file: Option<PathBuf>,
 
     #[cfg(debug_assertions)]
-    /// Skip the standard library import, for debugging only, does not affect REPL
+    /// Skips the standard library import. For debugging only. This does not affect REPL
     #[structopt(long)]
     nostdlib: bool,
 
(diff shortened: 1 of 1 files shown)


francois-caddet:
Only a detail ;)
I feel print-ast can be confusing with the raw printing as done until now in the debug. so, adding a `formated` or `pretty` in the comand may disambigate it.
Just

yannham:
Other idea: `unparse`/`unparse-ast`. I think it conveys better the idea of not just printing the AST as it is, but to go back to the source representation. It seems to exist as a word and be used in this context (although I wouldn't go as far as saying this is THE accepted term).

aspiwack:
I think that prettyprint is common enough that it ought not to be controversial.

yannham:
For the record, because I don't know if this has been made clear, but this command does print the AST back as Nickel source code, not as a tree representation. In any case, that's hardly very important, because this command is not really intended to be user-facing. I'm fine with `pprint-ast`

Issue #701, opened 2022-05-19
Undue `ill-formed type` error when using a record field as a type


yannham:
**Describe the bug**
When using a record access like `record.Contract` as a type annotation, one get an enigmatic ill-formed type error.

**To Reproduce**

```nickel
nickel> let foo = {Bar = Dyn}
nickel> (null | foo.Bar) : foo.Bar
error: ill-formed type
  ┌─ :1:1
  │
1 │ <unevaluated>
  │ ------------- ill-formed type
```

This fires whatever is on the RHS of the contract annotation `null | _`, as long as it is a custom contract., but requires a record access in the outer type annotation. For example, `(null | Baz) : foo.Bar` also triggers the error, but neither `(null | Bar) : Baz` nor `(null | Dyn) : foo.Bar` does.

**Expected behavior**
I expect the example program to be accepted and to return `null`.

**Additional context**
`ill-formed type` exception are almost always internal errors: they indicate that a non-row type ended up somewhere in a row position, but I believe those kind of types are impossible to write in the source syntax. There must be a mishandling of custom contracts used as a type somewhere, but it's surprising that anything depends on the structure of the custom contracts, which shouldn't ever be inspected by the typechecker...

yannham:
After a bit more investigation, the typechecker doesn't like when flat type (contracts) are not variables, which is of course really limiting. Not throwing this error anymore is easy to solve; however the larger issue at head is that this code appears in `unify`, where the typechecker has to decide if two flat types unify. And this is not a trivial matter. I'll open a separate issue for that.

yannham:
(Separate issue: #724)

yannham:
Closed by quickfix in #769. Keeping #724 open for a better long term solution.

PR #702, opened 2022-05-19
[Fix] Typing of stdlib items annotated with a contract

### LINKED ISSUE #700: Stdlib functions with contract annotation don't have the right type
**Describe the bug**
Stdlib functions which have contract annotation in place of static type annotations end up with the type `Dyn` in the interface of the module. For example:

```console
nickel>:typecheck builtin
Ok: {
  serialize: Dyn,
  hash: Dyn, 
  is_str: Dyn -> Bool, 
  ValueType: Dyn,
  is_record: Dyn -> Bool,
  deep_seq: forall a. Dyn -> a -> a,
  deserialize: Dyn, HashAlgorithm: Dyn,
  seq: forall a. Dyn -> a -> a, typeof: Dyn,
  ExportFormat: Dyn, is_array: Dyn -> Bool,
  is_fun: Dyn -> Bool, is_bool: Dyn -> Bool,
  is_num: Dyn -> Bool
}
```

While querying say, `serialize`, shows that it has a much more precise annotation:

```console
nickel>:q builtin.serialize
• contract: ExportFormat -> Dyn -> Str
[..]
```

**To Reproduce**
One easy way to get bugged by this issue is to just try to use such functions in a typed context:

```console
nickel>let data = {foo = "bar"} in
let as_json : Str = builtin.serialize `Json data in
as_json

error: incompatible types
  ┌─ repl-input-6:2:21
  │
2 │ let as_json : Str = builtin.serialize `Json data in
  │                     ^^^^^^^^^^^^^^^^^ this expression
  │
  = The type of the expression was expected to be `_a -> _b -> Str`
```
  

**Expected behavior**
The stdlib items annotated with contracts should have their type properly inferred.

**Additional context**
The fault lies, with high probability, in the function `typecheck::infer_type`, that do its magic to bubble up the types from the record fields to the top-level record.

yannham:
Fix #700. However, require a fix to #701 which currently causes tests to fail (and any meaningful test that we can think of, unfortunately, will have the same issue) before getting green CI and being mergeable.

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -116,7 +116,7 @@ impl<'a> Envs<'a> {
                 if let Term::RecRecord(rec, ..) = rt.as_ref() {
                     Ok(rec
                         .iter()
-                        .map(|(id, rt)| (id.clone(), infer_type(rt.as_ref()))))
+                        .map(|(id, rt)| (id.clone(), infer_record_type(rt.as_ref()))))
                 } else {
                     Err(EnvBuildError::NotARecord(rt.clone()))
                 }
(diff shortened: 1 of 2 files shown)


src/typecheck/mod.rs
```diff
@@ -1042,23 +1042,30 @@ pub fn apparent_type(
...
+            types, contracts, ..
+        }) if (types.is_some() || !contracts.is_empty()) => apparent_type(t, None, None).into(),
+        // We unwrap meta-values without any type information. Otherwise, something like a
+        // top-level module documentation would stop this function from doing its job.
+        Term::MetaValue(MetaValue {
+            value: Some(rt),
+            types: None,
+            contracts,
+            ..
+        }) if contracts.is_empty() => infer_record_type(rt.as_ref()),
```
fuzzypixelz, line 1058: Shouldn't this be `if types.is_none() && contracts.is_empty()` instead?

fuzzypixelz (approved):
LGTM.

src/typecheck/mod.rs
yannham: Ah, good catch!

yannham: In fact, it's already covered by the pattern `types: None`, because types is just an option.

PR #705, opened 2022-05-20
add mantis benches

francois-caddet:
come from a port of mantis ops by @yannham

### DIFF
--- a/benches/mantis.rs
+++ b/benches/mantis.rs
@@ -1,0 +1,21 @@
+use criterion::{criterion_main, Criterion};
+use nickel_lang_utilities::{ncl_bench_group, EvalMode};
+use pprof::criterion::{Output, PProfProfiler};
+
+ncl_bench_group! {
+    name = benches;
+    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    {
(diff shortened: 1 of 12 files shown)


benches/mantis.rs
```diff
@@ -0,0 +1,21 @@
...
+        name = "mantis",
+        path = "mantis/run",
+        args = (r#"{namespace = "mantis-staging", job="miner", role=`miner}"#),
+        eval_mode = EvalMode::DeepSeq,
+    }, {
+        name = "mantis serialize",
+        path = "mantis/run",
+        subtest = "serialize",
+        args = (r#"{namespace = "mantis-staging", job="miner", role=`miner}"#),
+        eval_mode = EvalMode::DeepSeq,
```
yannham, line 18: I think you don't need to `DeepSeq` here, because the final value you get is a string anyway, so it's going to be in evaluated form.

benches/mantis/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
yannham, line 1: I think we can get rid of that because it's exactly the same license as the current repo.

yannham (approved):

PR #709, opened 2022-05-24
first draft of typechecker benches.

francois-caddet:
proved on a rewrite of nixpkgs lists lib in nickel

right now, the benchmarked code is very small but it put bases to create typechecker benchmarks.
As oposit to others benches, this one should bench only the typecheck phase.


Cargo.toml
```diff
@@ -101,3 +101,7 @@ harness = false
 [[bench]]
 name = "mantis"
 harness = false
+
+[[bench]]
+name = "nixpkgs"
```
yannham, line 106: This is not a really descriptive name, if not deceptive. I would insist on it benchmarking typechecking, like`typecheck-nixpkgs-lib` or something.

benches/nixpkgs/lists.ncl
```diff
@@ -0,0 +1,119 @@
...
+
+  imap0: forall a b. (Num -> a -> b) -> (List a) -> (List b)
+  | doc m%"
+      Map with index starting from 0
+
+     Example:
+       imap0 (fun i v => "%{v}-%{%to_string% i}") ["a" "b"]
+       >> [ "a-0", "b-1" ]
+  "%m
+  imap0 = fun f list => %gen% (fun n => f n (%elem_at% list n)) (%%length%% list),
```
yannham, line 98: Those double `%%` are not valid syntax, and cause parse errors.

benches/nixpkgs/lists.ncl
```diff
@@ -0,0 +1,119 @@
...
+      Apply the function to each element in the list. Same as `map`, but arguments
+      flipped.
+
+      Example:
+        forEach [ 1, 2 ] (fun x =>
+          toString x
+        )
+        >> [ "1" "2" ]
+    "%m
+  = fun xs f => %map% xs f,
```
yannham, line 27: Why not use the stdlib, instead of low-level primops?

```suggestion
  = fun xs f => array.map f xs,
```

yannham (changes requested):
There are parse error in the provided file (try to run it with `nickel -f benches/nixpkgs/list.ncl`). For some reason the benchmarking code load them in an error-tolerant way, and probably just ignore the typechecking of error nodes.

You should also probably use the stdlib functions `array.xxx` rather than the builtin operators `%xxx%. This shouldn't make a difference with respect to typechecking, but would make the code way more readable.

PR #712, opened 2022-05-25
bench stdlib typechecking

francois-caddet:
As part of the intent to bench the typechecker, this PR add a bench of stdlib.
It should as possible only bench the typechecking phase and not the previous ones (loading and parsing).

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -802,14 +802,22 @@ impl Cache {
         // 1. The stdlib is meant to stay relatively light.
         // 2. Typechecking the standard library ought to occur only during development. Once the
         //    stdlib is stable, we won't have typecheck it at every execution.
+        let global_env = self.mk_types_env().map_err(|err| match err {
+            CacheError::NotParsed => CacheError::NotParsed,
+            CacheError::Error(_) => unreachable!(),
+        })?;
+        self.typecheck_stdlib_(&global_env)
+    }
+
+    /// Internal function to typecheck stdlib. Has to be public because it's used in benches.
+    pub fn typecheck_stdlib_(
+        &mut self,
+        global_env: &typecheck::Environment,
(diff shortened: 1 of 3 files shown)


src/cache.rs
```diff
@@ -802,14 +802,22 @@ impl Cache {
...
         // 2. Typechecking the standard library ought to occur only during development. Once the
         //    stdlib is stable, we won't have typecheck it at every execution.
+        let global_env = self.mk_types_env().map_err(|err| match err {
+            CacheError::NotParsed => CacheError::NotParsed,
+            CacheError::Error(_) => unreachable!(),
+        })?;
+        self.typecheck_stdlib_(&global_env)
+    }
+
+    /// Internal function to typecheck stdlib. Has to be public because it's used in benches.
```
yannham, line 812: Nitpick: if it's public, it's not totally internal :stuck_out_tongue: so maybe a short description could be added, like "Typecheck the stdlib, provided the initial typing environment."

yannham (changes requested):
The time measurement is suspiciously low (in the order of hundreds of microseconds). I suspect that typechecking is only done the first time, and then subsequent calls to `self.typecheck` do nothing because the state of the entries corresponding to the stdlib is `TYPECHECKED`, which invalidates criterion's measurements.

yannham (approved):
My bad, I don't think this is the case in the end, because you clone the initial cache at each iteration. For comparison the other typechecking bench is the order of a few microseconds, which is a huge difference. So it seems typechecking is just that fast on small programs.

francois-caddet:
Yes, I noticed that also. But I actualy think it's correct. As you said, the cache is cloned at each iteration and the type_env is not mutable. So we should not have the typechecking cached. Just be happy, your typechecker is so fast ;)

PR #722, opened 2022-06-03
full translation of nixpkgs/lib/lists.nix

francois-caddet:
Complet the typechecking bench with the full translation of `nixpkgs/lib/lists.nix` to Nickel

### DIFF
--- a/benches/nixpkgs/lists.ncl
+++ b/benches/nixpkgs/lists.ncl
@@ -116,4 +116,404 @@
        >> [ "a", "z", "b", "z" ]
   "%m
   = (fun f list => array.flatten (array.map f list)),
+
+  flatten: Dyn -> Array Dyn
+  | doc m%"
+      Flatten the argument into a single list, that is, nested lists are
+     spliced into the top-level lists.
+
+     Example:
+       flatten [1, [2, [3], 4], 5]
+       >> [1, 2, 3, 4, 5]
+       flatten 1
+       >> [1]
+  "%m
+  = fun x =>
+    if builtin.is_array x
+    then concatMap (fun y => flatten y) (x | Array Dyn)
+    else [x],
+
+  remove: forall a. a -> Array a -> Array a
+  | doc m%"
+      Remove elements equal to 'e' from a list.  Useful for buildInputs.
+
+     Example:
+       remove 3 [ 1, 3, 4, 3 ]
+       >> [ 1, 4 ]
+  "%m
+  = # Element to remove from the list
+    fun e => array.filter (fun x => x != e),
+
+  findSingle: forall a. (a -> Bool) -> a -> a -> Array a -> a
+  | doc m%"
+      Find the sole element in the list matching the specified
+     predicate, returns `default` if no such element exists, or
+     `multiple` if there are multiple matching elements.
+
+     Example:
+       findSingle (fun x => x == 3) "none" "multiple" [ 1, 3, 3 ]
+       >> "multiple"
+       findSingle (fun x => x == 3) "none" "multiple" [ 1, 3 ]
+       >> 3
+       findSingle (fun x => x == 3) "none" "multiple" [ 1, 9 ]
+       >> "none"
+  "%m
+  = fun
+    # Predicate
+    pred
+    # Default value to return if element was not found.
+    def
+    # Default value to return if more than one element was found
+    multiple
+    # Input list
+    list
+    => let found = array.filter pred list in
+       let len = array.length found in
+       if len == 0 then def
+       else if len != 1 then multiple
+       else array.head found,
+
+  findFirst: forall a. (a -> Bool) -> a -> Array a -> a
+  | doc m%"
+      Find the first element in the list matching the specified
+     predicate or return `default` if no such element exists.
+
+     Example:
+       findFirst (fun x => x > 3) 7 [ 1, 6, 4 ]
+       >> 6
+       findFirst (fun x => x > 9) 7 [ 1, 6, 4 ]
+       >> 7
+  "%m
+  = fun
+    # Predicate
+    pred
+    # Default value to return
+    def
+    # Input list
+    list
+    => let found = array.filter pred list in
+       if found == [] then def else array.head found,
+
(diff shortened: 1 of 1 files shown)


benches/nixpkgs/lists.ncl
```diff
@@ -116,4 +116,554 @@
...
+  remove: forall a. a -> Array a -> Array a
+  | doc m%"
+      Remove elements equal to 'e' from a list.  Useful for buildInputs.
+
+     Example:
+       remove 3 [ 1, 3, 4, 3 ]
+       >> [ 1, 4 ]
+  "%m
+  = # Element to remove from the list
+    fun e => array.filter (fun x => x != e),
```
yannham, line 145:
```suggestion
    fun e => array.filter ((!=) e),
```

Nitpick, but just to show how cool curried operator are :stuck_out_tongue: (and to test their typechecking too)

benches/nixpkgs/lists.ncl
```diff
@@ -116,4 +116,554 @@
...
+     list.  If you're using this, you should almost certainly
+     reconsider if there isn't a more "well-typed" approach.
+
+     Example:
+       toList [ 1, 2 ]
+       >> [ 1, 2 ]
+       toList "hi"
+       >> [ "hi "]
+  "%m
+  = fun x => if builtin.is_array x then (x | Array Dyn) else [x],
```
yannham, line 280: Ah, interesting example of how flow typing could be useful here (would eliminate the need for a contract application).

benches/nixpkgs/lists.ncl
```diff
@@ -116,4 +116,554 @@
...
+            mate  = [ { name = "mate",  script = "gnome-session &" } ],
+            xfce  = [ { name = "xfce",  script = "xfce4-session &" } ],
+          }
+  "%m
+  = fun pred => foldl_ (fun r e =>
+       let key = pred e in
+         { "%{key}" = (r."%{key}" @ [e]) } & (record.remove key r)
+    ) {},
+
+  groupBy_: forall a b. (b -> a -> b) ->  b -> (a -> Str) -> Array a -> { _: b}
```
yannham, line 343:
```suggestion
  groupBy_ : forall a b. (b -> a -> b) ->  b -> (a -> Str) -> Array a -> { _: b}
```

benches/nixpkgs/lists.ncl
```diff
@@ -116,4 +116,554 @@
...
+     Example:
+       naturalSort ["disk11", "disk8", "disk100", "disk9"]
+       >> ["disk8", "disk9", "disk11", "disk100"]
+       naturalSort ["10.46.133.149", "10.5.16.62", "10.54.16.25"]
+       >> ["10.5.16.62" "10.46.133.149" "10.54.16.25"]
+       naturalSort ["v0.2", "v0.15", "v0.0.9"]
+       >> [ "v0.0.9", "v0.2", "v0.15" ]
+  "%m
+  = fun lst =>
+    let vectorise = fun s => map (fun x => if builtin.is_array x then (head x) | Num else x) (array.split "(0|[1-9][0-9]*)" s) in
```
yannham, line 540: I don't think this function exists?

yannham (commented):
I haven't looked a everything yet. But it seems the source file doesn't typecheck on my machine?

yannham (approved):
Ah, yes, `toposort` is annoying because the result has a variable number of fields and they don't have the same type. For now I don't think we can do much better.

benches/nixpkgs/lists.ncl
```diff
@@ -116,4 +116,557 @@
...
+    fun e => array.filter ((!=) e),
+
+  findSingle: forall a. (a -> Bool) -> a -> a -> Array a -> a
+  | doc m%"
+      Find the sole element in the list matching the specified
+     predicate, returns `default` if no such element exists, or
+     `multiple` if there are multiple matching elements.
+
+     Example:
+       findSingle (fun x => x == 3) "none" "multiple" [ 1, 3, 3 ]
```
bew, line 154: (beginner question)
Don't you have an issue with the types for this example?
The function type says that `def`, `multiple` & the elements of the array should have the same type, but in your example the elements of the array are integers and the others are string.

Or am I missing something?

yannham: @bew Yes, you're right. You could still use this function with `a = Dyn` indeed, but Nickel won't infer this kind of upper bound for you automatically, so you would need a bunch of contract or type annotation to make it accepted. The above type is wrong (I mean it is sound, but arbitrarily too restrictive), and the provided example probably doesn't typecheck as it is.

I guess a more reasonable type is `forall a b. (a -> Bool) -> b -> b -> Array a -> b`

bew: Well no, because if the element is found only once, it should be returned..
And `def` & `multiple` could (for whatever reason) be of different types..

Is there a plan to add union types to Nickel and allow this to precisely return a kind of `Union[ElemType, NoneType, MultiType]` ?

yannham: For some reasons I wrote an answer but probably forgot to hit "Enter"... @bew you're right, again, I just inferred a type from the example too quickly. Indeed, in a static type system I don't think we can do much better than the current type. As you say, the example is morally relying on union types.

Note that the example is fine if it is in dynamically typed code: the contract won't complain. In typed code, you can use some type annotations to make the example pass. Sorry, I've written a working example in the previous ghost comment, but I can't find it. It merely amounted to use a `: Dyn` for the result of the call and `| Dyn` and `| Array Dyn` for arguments. The boolean function argument didn't need an annotation, so that's not too bad. The RFC on typechecking could even help remove some of the `Dyn` annotations on the arguments.

Union types are not planned in the near future. We actually tried to implement them, but encountered difficulties: in Nickel, each static type must have a contract as a counter-part (for so-called *soundness* of the gradual typing), but it turns out combining generic unions with lazy contracts is really hard, with undesirable consequences. We even presented [a paper about this misadventure](https://arxiv.org/abs/2106.06278) at SPLASH21 and ECOOP22, which is condensed in [a blog post](https://www.tweag.io/blog/2022-04-28-union-intersection-contracts/).

PR #725, opened 2022-06-03
Use `Environment::disjoin_union` instead of reconstructing environments

fuzzypixelz:
The `disjoint_union` method recursively .. joins two environments by interleaving their HashMaps. In the worst cases this produces bloated environments but in the best cases it helps avoid reconstructing a new environment by copying elements to it.

A `debug_assert` verifying the disjoint nature would be a good idea at this point.

Using the unit tests and a hacky elapsed time measurement to compare this against the existing implementation yields:

```
extend took 17.003µs
disjoint_union took 4.2µs

extend took 109.418µs
disjoint_union took 4.628µs

extend took 7.598µs
disjoint_union took 3.541µs

extend took 15.054µs
disjoint_union took 3.003µs

extend took 15.438µs
disjoint_union took 4.023µs

extend took 12.13µs
disjoint_union took 2.965µs
```

Clearly handling environment layers is more efficient than moving around elements themselves. Still have to look into how this impacts the wider codebase.

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -129,6 +129,41 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
     fn was_cloned(&self) -> bool {
         Rc::strong_count(&self.current) > 1
     }
+
+    /// Layer the `left` environement on top of `right`.
+    ///
+    /// # Issues
+    ///
+    /// - This only makes sense for disjoint environements.
+    /// - The resulting environement is list whoose size is the sum of
+    /// the inputs' sizes. Which means longer "non-local" lookups.
+    pub fn disjoint_union(left: &Self, right: &Self) -> Self {
+        // NOTE: it doesn't make sense to take an owned environment because
+        // it would probably be already cloned, with an empty `current`.
+        if left.previous.borrow().is_none() {
+            Environment {
+                current: left.current.clone(),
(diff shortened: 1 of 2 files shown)


fuzzypixelz:
The benchmark vary from slightly faster (-3%) to slightly slower (+4%, for folding), so I'm not quite sure how to interpret this.

If we're evaluating `a1 @ a2 @ a3 ... @ an` and each array environment the same size `s`, then at step `i` we get an environment with `s * (2**i)` layers, which doesn't sound very good.

yannham:
Ah, indeed, we traded speed of merging (in `O(depth_right_env)` for the disjoint_union versus `O(elem_count)` for the current extend approach) for speed of access (O(`depth_right_env`) after fusion for the disjoint_union versus `O(1)` for the current extend approach). And then of course there's cache effects, etc. (I honestly don't know how much Hashmap iterators are (in)efficient, but in the extend approach the changing layers most probably invalidate the cache). It was worth trying, but in retrospective this might not be a clear-cut win...

yannham:
@fuzzypixelz do you agree with the overall conclusion, and then I think we can proceed and close this PR, or do you think differently/have more insights now to discuss?

fuzzypixelz:
@yannham I think we can close this PR. Overall I don't think altering the structure of environments this way is worth the minuscule (if any) speedups. We may even get regressions later if we use this without a lot of care.

PR #729, opened 2022-06-08
Mark attrs1 as unused to resolve rust warning

ErinvanderVeen:


ErinvanderVeen:
@fuzzypixelz I assigned you because this warning originates in a PR you created. The fact that is unused might be an issue, I'll leave that up to you.

fuzzypixelz (approved):
You are right, it should be unused. In the original PR it was used for the resulting attribute, but then I changed it to be explicit following Yann's feedback and forgot to mark `attrs1` as unused.

Thanks!

PR #730, opened 2022-06-08
WIP: reduce closurization calls in record merging

fuzzypixelz:
The current results aren't very impressive. On very particular loads like the following:

```
let rec generate = fun n f =>
  if n == 0 then {}
else generate (n - 1) f & { "%{f n |> string.from_num}" = null }
in
  generate 999 (fun x => x + 1)
```
I can see a speedup of about 30%, but this doesn't transfer well to benches like `mantis`, where the difference is minuscule.

I think that with record-merging, we need to closurize the merged fields to maintain the invariant "all record fields are closurized upon merging", which makes it hard to avoid some of the current calls. It's almost as if the overhead we avoid on `.closurize` calls is reintroduced by the environment manipulation.

### DIFF
--- a/src/destruct.rs
+++ b/src/destruct.rs
@@ -65,7 +65,10 @@ impl Destruct {
                             .into_iter()
                             .map(|m| m.as_meta_field())
                             .collect(),
-                        RecordAttrs { open },
+                        RecordAttrs {
+                            open,
+                            closurized: false,
+                        },
                     )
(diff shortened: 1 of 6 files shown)


yannham:
Unless there are new developments, I think we can mark this as an inconclusive experiment, and close for the time being. Is that correct @fuzzypixelz ?

PR #732, opened 2022-06-10
Avoid recompiling regex in %str_is_match%

fuzzypixelz:
# Rationale

When using functions like `string.is_match regex str`, the interpreter recompiles the regular expression `regex` on every use. This is because the underlying implementation in the standard library relies on the binary operator `%str_is_match%`. After discussing this with @francois-caddet and @yannham, we can make `%str_is_match%` a unary operator which takes a (regex) string and generates a Nickel function matching on this string.

# Implementation

The above sounds straightforward, but I couldn't write it out using the current AST. In particular, our "generated" function (let's call it the matcher), would need to refer to the same compiled `Regex` and _dynamically_ call `is_match` on it. 

We _can_ use closures:

- We can force them to be `Sync + Send`, which will keep the entire AST thread safe.
- In our case, we can move the compiled regular expression into a closure and have it call `is_match` dynamically.

This comes in the form of a new `UnaryOp::FFI(<closure>)`, which means "call this Rust closure on a Nickel Term", it's an operator because this is less intrusive than adding it as a term.

# Overly-specific benchmark

```
let pattern = "^[\\w\\.-]+@([\\w-]+\\.)+[\\w-]{2,4}$" in
let Email = contract.from_predicate (string.is_match pattern) in 
[
  "nickel-lang@tweag.io",
  # -- snip --
] | Array Email
```
With a list of ~70 emails, the difference is `0.20s` vs `0.02s` on my machine, but this is a simple measure with `time`.

Even though the mantis benchmark motivated this (as it did ~20 calls to `Regex::new`), the results weren't impressive there.

### DIFF
--- a/stdlib/string.ncl
+++ b/stdlib/string.ncl
@@ -272,15 +272,33 @@
     | doc m%"
       `is_match regex str` checks if `str` matches `regex`.
 
-      Note that this function may perform better by sharing its partial application between multiple calls,
-      because in this case the underlying regular expression will only be compiled once.
-
       For example:
       ```nickel
         is_match "^\\d+$" "123" =>
           true
         is_match "\\d{4}" "123" =>
           false
+      ```
+
+      For example, in the following program, the whole call to
+      `string.is_match "[0-9]*\\.?[0-9]+ x"` is re-evaluated at each invocation of
+      `is_number`. The regexp will be recompiled 3 times in total:
+
+      ```nickel
+      let is_number = fun x => string.is_match "[0-9]*\\.?[0-9]+" x in
+      ["0", "42", "0.5"] |> array.all is_number =>
+        true
+      ```
+
+      On the other hand, in the version below, the partial application of
+      `string.is_match "[0-9]*\\.?[0-9]+"` is evaluated once, returning a
+      function capturing the compiled regexp. The regexp will only be compiled
+      once and for all:
+
+      ```nickel
+      let is_number' = string.is_match "[0-9]*\\.?[0-9]+" in
+      ["0", "42", "0.5"] |> array.all is_number' =>
+        true
       ```
       "%m
     = fun regex => %str_is_match% regex,
@@ -290,9 +308,6 @@
       `match regex str` matches `str` given `regex`. Results in the part of `str` that matched, the index of the
       first character that was part of the match in `str`, and a arrays of all capture groups if any.
 
-      Note that this function may perform better by sharing its partial application between multiple calls, 
-      because in this case the underlying regular expression will only be compiled once.
-
       For example:
       ```nickel
         match "^(\\d).*(\\d).*(\\d).*$" "5 apples, 6 pears and 0 grapes" =>
@@ -300,6 +315,10 @@
         match "3" "01234" =>
           { match = "3", index = 3, groups = [ ] }
       ```
+
+      Note that this function may perform better by sharing its partial application between multiple calls,
+      because in this case the underlying regular expression will only be compiled once (See the documentation
+      of `string.is_match` for more details).
       "%m
     = fun regex => %str_match% regex,
 
--- a/stdlib/string.ncl
+++ b/stdlib/string.ncl
@@ -41,9 +41,10 @@
       ```
       "%m
     = let pattern = m%"^[+-]?(\d+(\.\d*)?(e[+-]?\d+)?|\.\d+(e[+-]?\d+)?)$"%m in
+      let is_num_literal = %str_is_match% pattern in
       fun l s =>
         if %is_str% s then
-          if %str_is_match% s pattern then
+          if is_num_literal s then
             s
           else
             %blame% (%tag% "invalid num literal" l)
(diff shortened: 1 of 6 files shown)


fuzzypixelz:
I left out the dead code from the binary operator version intact, and much of the new code isn't very clean. I'd like to first see what you think about this approach and if there is a simpler alternative.

yannham (commented):
The base approach is reasonable I think (make `is_match` a unary op that compiles the regex and return another function/unary op), but I think storing a closure inside an op sounds overkill and way too generic/powerful: we don't really know what exactly could be in there (well, probably pretty much anything), and I fear this will encourage further hacks to use this `Code` to push magic in it that wasn't intended at all initially.

My take would be to just have a variant of operators `is_match` and `match` that are parametrized by a compiled regular expression (and nothing else). That is, instead of FFI, having `IsMatchCompiled(Regex)` and `MatchCompiled(Regex)` that only generated during evaluation. This is similar, for example, to the `StaticAccess` operator (record field access) which is in fact a family of operators `StaticAccess(id)` corresponding to `_.id`. This is not great, but as a rule of thumb, it's probably fine to attach parameters to operators as long as those are pure data that don't require any further evaluation and don't retain any reference to terms: strings, numbers, identifiers, compiled regex should be fine.

Another solution is to make compiled regex first-class, or even regex object first class. I don't know if we want to do that right now, so for the time being, the parametrized operators approach (duly documented) will probably do.

fuzzypixelz:
> I think storing a closure inside an op sounds overkill and way too generic/powerful: we don't really know what exactly could be in there (well, probably pretty much anything), and I fear this will encourage further hacks to use this `Code` to push magic in it that wasn't intended at all initially.

I went a bit overboard with this and over-engineered the solution. In the back of my mind, this seemed like a "useful trick for future problems", but as you say closures can have arbitrary effects and will cause headaches later on.

> That is, instead of FFI, having IsMatchCompiled(Regex) and MatchCompiled(Regex) that are only generated during evaluation.

I was all too focused on putting the `Regex::is_match` call into the AST that I didn't really consider this!

I'll rectify the code and make sure to squash the commits.

fuzzypixelz:
I have added a sentences in the stdlib documentation for `is_match` an `match`, explaining why it's better to partially apply them. Then I noticed this wasn't properly done in the `NumLiteral` contract so I fixed it.

There is a conflict because we removed `Ident` on `master`. We would simply keep `master`'s version.

stdlib/string.ncl
```diff
@@ -294,6 +295,9 @@
     | doc m%"
       `is_match regex str` checks if `str` matches `regex`.
 
+      Note that this function may perform better by sharing its partial application between multiple calls,
+      because in this case the underlying regular expression will only be compiled once.
+
```
yannham, line 300: I think it deserves a small example, say:

```
let is_integer = fun kind x =>
  if kind == `Integer then
    string.is_match "[0-9]+" x
  else
    string.is_match "[0-9]*\\.?[0-9]+" x in
is_number `Integer "1" && is_number `Float "0.55"
```
vs
```
let is_integer = fun kind =>
  if kind == `Integer then
    string.is_match "[0-9]+"
  else
    string.is_match "[0-9]*\\.?[0-9]+" in
...
```

Or something like this (the code may need to be fixed, I haven't tried it).

yannham (commented):
I think the doc would benefit from a small example. Let's just rebase on current master with respect to the conflict.

stdlib/string.ncl
fuzzypixelz: Would it make sense to include two examples (in both `is_match` and `match`)?

fuzzypixelz:
This should be good now.

stdlib/string.ncl
```diff
@@ -278,8 +279,24 @@
...
+      let is_number = fun x => string.is_match "[0-9]*\\.?[0-9]+" x in
+      ["0", "42", "0.5"] |> array.all is_number =>
+        true
+
+      # This function does not saturate `is_match`'s arguments
+      # and is therefore more efficient than `is_number`.
+      let is_number' = string.is_match "[0-9]*\\.?[0-9]+" in
+      ["0", "42", "0.5"] |> array.all is_number' =>
+        true
+      ```
```
yannham, line 297:
```suggestion
      For example, in the following program, the whole call to
      `string.is_match "[0-9]*\\.?[0-9]+ x"` is re-evaluated at each invocation of
      `is_number`. The regexp will be recompiled 3 times in total:
      
      ```nickel
      let is_number = fun x => string.is_match "[0-9]*\\.?[0-9]+" x in
      ["0", "42", "0.5"] |> array.all is_number =>
        true
      ```

      On the other hand, in the version below, the partial application of
      `string.is_match "[0-9]*\\.?[0-9]+"` is evaluated once, returning a
      function capturing the compiled regexp. The regexp will only be compiled
      once and for all:

      ```nickel
      let is_number' = string.is_match "[0-9]*\\.?[0-9]+" in
      ["0", "42", "0.5"] |> array.all is_number' =>
        true
      ```
```

yannham (approved):
Suggested a slightly more explicit wording for the documentation. Do as you see fit, and you can merge :slightly_smiling_face:

PR #733, opened 2022-06-13
first draft of nix parser to nickel AST

francois-caddet:
- parse a let binding of type `let <IDENT> = <LITERAL>; in <IDENT>`
- can parse when multiple bindings in the let
- for now, concider every let as recursive let.


src/nix.rs
```diff
@@ -0,0 +1,338 @@
...
+    let span = mk_span(file_id, pos.start().into(), pos.end().into());
+    println!("{:?}", node);
+    let node: ParsedType = node.try_into().unwrap();
+    match node {
+        ParsedType::Error(_) => Term::ParseError.into(),
+        ParsedType::Root(n) => translate(n.inner().unwrap(), file_id),
+        ParsedType::Paren(n) => translate(n.inner().unwrap(), file_id),
+        ParsedType::Dynamic(n) => translate(n.inner().unwrap(), file_id),
+
+        ParsedType::Assert(n) => unimplemented!(),
```
yannham, line 149: We can keep it like this for now, but in the future `assert` can be implemented as a contract annotation, as it is done in tests.

src/nix.rs
```diff
@@ -0,0 +1,338 @@
...
+        ParsedType::Value(n) => match n.to_value().unwrap() {
+            value::Value::Float(v) => Term::Num(v).into(),
+            value::Value::Integer(v) => Term::Num(v as f64).into(),
+            value::Value::String(v) => Term::Str(v).into(),
+            // TODO: How to manage Paths in nickel?
+            value::Value::Path(a, v) => Term::Str(v).into(),
+        },
+        ParsedType::Str(n) => {
+            let parts = n.parts();
+            //TODO: Do we actualy need the `Term::Str` variant in nickel AST?
```
yannham, line 160: Yes, for evaluation. A `StrChunks` evaluate to a `Str` in the end, and operator like concatenation can only operate on strings. It's also useful to determine e.g. quoted fields that are actually statically known, as `{"myfield%bar" = 1}` from actualy dynamic fields, etc.

However, as far as compiling Nix is concerned, you can indeed stick to always producing `StrChunks`.

src/nix.rs
```diff
@@ -0,0 +1,338 @@
...
+        // definition.
+        ParsedType::Ident(n) => match n.as_str() {
+            "true" => Term::Bool(true),
+            "false" => Term::Bool(false),
+            "null" => Term::Null,
+            id => Term::Var(id.into()),
+        }
+        .into(),
+        ParsedType::LegacyLet(_) => unimplemented!(), // Probably useless to suport it in a short term.
+        ParsedType::LetIn(n) => {
```
yannham, line 216: What could be a nice addition: a comment explaining that because `let` are mutually recursive in Nix, you translate them as recursive records + pattern matching in Nickel. It might take some head scratching to understand it as it is :slightly_smiling_face:

src/nix.rs
```diff
@@ -0,0 +1,338 @@
+use crate::cache::Cache;
+use crate::parser::utils::mk_span;
+use crate::term::{BinaryOp, UnaryOp};
+use crate::term::{MergePriority, MetaValue, RichTerm, Term};
+use codespan::FileId;
+use rnix;
+use rnix::types::{BinOp, EntryHolder, TokenWrapper, TypedNode, UnaryOp as UniOp};
+use std::collections::HashMap;
+
+impl From<(UniOp, FileId)> for RichTerm {
```
yannham, line 10: It sounds a bit overstretched to implement the from trait from a tuple of an unary op and a file id into a `RichTerm`. Usually `From/Into` are for somehow straightforward conversions between. E.g. in Rust I think you can't do `let s : String = 5.into()`, because it implies a non trivial operation. You need to do `5.to_string()`, where `to_string()` has exactly the same signature as `into()`, but has a different name. What about implementing something like `translate_nix_op` or whatever rather than the `From` trait?

src/nix.rs
```diff
@@ -0,0 +1,338 @@
...
+                Term::Op1(UnaryOp::Ite(), translate(n.condition().unwrap(), file_id)).into(),
+                translate(n.body().unwrap(), file_id),
+            )
+            .into(),
+            translate(n.else_body().unwrap(), file_id),
+        )
+        .into(),
+        ParsedType::BinOp(n) => (n, file_id).into(),
+        ParsedType::UnaryOp(n) => (n, file_id).into(),
+        ParsedType::Select(n) => match ParsedType::try_from(n.index().unwrap()).unwrap() {
```
yannham, line 303: Ugh, that's a lot of optionals :stuck_out_tongue:

src/nix.rs
```diff
@@ -0,0 +1,338 @@
...
+            .into(),
+        },
+        ParsedType::OrDefault(n) => unimplemented!(),
+
+        // TODO: what are these?
+        ParsedType::PatBind(_) | ParsedType::PathWithInterpol(_) => {
+            unimplemented!()
+        }
+
+        // Is not a `RichTerm` so is transformed as part of actual ones.
```
yannham, line 323: I didn't understand what this comment mean.

src/nix.rs
francois-caddet: Ok so a later optimisation could be done but we let it as it is for now.

src/nix.rs
francois-caddet: What do you mean? The return type of many functions of `rnix`? Yes in many cases for me it make no sens to return an option...

src/nix.rs
francois-caddet: Mmmmh... me neither ;)
let me think...
Actualy, these variants are not "Terms" in the sens they can not be parsed independently of a Term. They can only be used in a term. Our `translate` function return a term so these can not be parsed at this level. they are parsed as parts of actual terms.
...
Is it more clear?

src/bin/nickel.rs
```diff
@@ -31,6 +31,10 @@ struct Opt {
 /// Available subcommands.
 #[derive(StructOpt, Debug)]
 enum Command {
+    /// translate Nix input to Nickel code.
+    /// Only a POC, main target is to be able to run Nix code on nickel.
+    /// May never be a complet source to source transformation.
+    Nixin,
```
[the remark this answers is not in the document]
yannham: Ah, I haven't even noticed the palindrome. Fair enough. Maybe we will have a feature flag too, because we don't want to ship a full nix compiler for:

1. People that don't care about Nix and want a small, embedabble conflang interpreter
2. The WASM REPL

src/nix.rs
francois-caddet: I don't realy know. Why to return an `Option` and not an error.
Here, `try_from` return a `Result` but I didn't manage it for now. But `index()` return an `Option` what will never be `None` if `try_from` don't fail. I think. Anyway if it could, for me it should be concidered as an error (How a record access can have `None` for the accessed field?). It may be done like this for the programaticaly generated trees.

src/nix.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+impl ToNickel for BinOp {
+    fn translate(self, file_id: FileId) -> RichTerm {
+        use rnix::types::BinOpKind::*;
+        let lhs = self.lhs().unwrap().translate(file_id);
+        let rhs = self.rhs().unwrap().translate(file_id);
+        match self.operator().unwrap() {
+            // TODO: Should be fixed using a nickel function `compat.concat` of type `a -> a -> a`
+            // using `str_conct` or `array_concat` in respect to `typeof a`.
```
yannham, line 31:
```suggestion
            // using `str_concat` or `array_concat` in respect to `typeof a`.
```

src/nix.rs
```diff
@@ -0,0 +1,279 @@
...
+                        Home => format!("~/{}", v),
+                        Store => format!("<{}>", v),
+                    };
+                    Term::Str(strpath)
+                }
+            }
+            .into(),
+            ParsedType::Str(n) => {
+                let parts = n.parts();
+                //TODO: Do we actualy need the `Term::Str` variant in nickel AST?
```
yannham, line 93:
```suggestion
```

src/nix.rs
```diff
@@ -0,0 +1,279 @@
...
+                            .collect();
+                        elaborate_field_path(p, val)
+                    })
+                    .collect();
+                build_record(fields, Default::default()).into()
+            }
+
+            // In nix it's allowed to define vars named `true`, `false` or `null`.
+            // But we prefer to not suport it. If we try to redefine one of these builtins, nickel
+            // will panic (see below in the `LetIn` arm.
```
yannham, line 139:
```suggestion
            // will panic (see below in the `LetIn` arm).
```

src/nix.rs
```diff
@@ -0,0 +1,279 @@
...
+                            })
+                            .collect();
+                        elaborate_field_path(p, val)
+                    })
+                    .collect();
+                build_record(fields, Default::default()).into()
+            }
+
+            // In nix it's allowed to define vars named `true`, `false` or `null`.
+            // But we prefer to not suport it. If we try to redefine one of these builtins, nickel
```
yannham, line 138:
```suggestion
            // But we prefer to not support it. If we try to redefine one of these builtins, nickel
```

src/nix.rs
```diff
@@ -0,0 +1,279 @@
...
+                let mut destruct_vec = Vec::new();
+                let mut fields = HashMap::new();
+                for kv in n.entries() {
+                    // In `let` blocks, the key is suposed to be a single ident so `Path` exactly one
+                    // element.
+                    let id = types::Ident::cast(kv.key().unwrap().path().next().unwrap()).unwrap();
+                    // Check we don't try to redefine builtin values. Even if it's possible in Nix,
+                    // we don't suport it.
+                    let id: Ident = match id.as_str() {
+                        "true" | "false" | "nul" => panic!(
```
yannham, line 163:
```suggestion
                        "true" | "false" | "null" => panic!(
```

yannham (approved):

src/nix.rs
milahu: > either `rowan` or `rnix` leads to a steady flow of `unwrap` that hurt our little heart of type-safers.

alternatives ...?

imho this should be an incremental parser like tree-sitter or lezer-parser,
so that nickel core can provide the language server

PR #734, opened 2022-06-14
[Refactoring] Separate walk phase during typechecking

yannham:
As part of refactoring the typechecker to make the different phase better separated, this PR extract a "walk" phase in a separate function. Walk mode correspond to the non-strict mode, outside of a statically typechecked block, where the typechecker simply walk the AST while looking for typed blocks (as well as saving the apparent type of bound variables).

Previously, the `unify` function used to branch on the `strict` parameter for this non-strict phase. Still, the typechecker was spinning useless fresh unification variables and calling to `unify` needlessly (that would return immediately). Separating the walk phase allowed to get rid of `unify_` and the branching in `unify`, and should offer better performances as walking is lighter and doesn't allocate garbage unification variables. The benchmarks show indeed a constant improvements in the order of a few percents up to 10%. It would be interesting to see the difference on the complete typechecking bench of #722 once it lands.

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -41,6 +41,7 @@
 //!
 //! In non-strict mode, all let-bound expressions are given type `Dyn`, unless annotated.
 use crate::cache::ImportResolver;
+use crate::destruct::*;
 use crate::environment::Environment as GenericEnvironment;
 use crate::error::TypecheckError;
 use crate::identifier::Ident;
@@ -208,8 +209,7 @@ where
 {
     let (mut table, mut names) = (UnifTable::new(), HashMap::new());
     let mut building = Linearization::new(LL::Building::default());
-    let ty = table.fresh_unif_var();
-    let mut wildcard_vars = vec![];
+    let mut wildcard_vars = Vec::new();
 
     {
         let mut state: State = State {
@@ -220,19 +220,17 @@ where
             wildcard_vars: &mut wildcard_vars,
         };
 
-        type_check_(
+        walk(
             &mut state,
             Envs::from_global(global_env),
             &mut building,
             linearizer.scope(),
-            false,
             t,
-            ty.clone(),
         )?;
     }
 
     let result = TypeCheckingOutput {
-        types: to_type(&table, ty),
+        types: Types(AbsType::Dyn()),
         wildcards: wildcard_vars_to_type(wildcard_vars, &table),
     };
     let lin = linearizer.complete(building, (table, names)).into_inner();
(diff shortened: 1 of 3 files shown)


src/typecheck/mod.rs
```diff
@@ -248,41 +242,271 @@ where
...
+                );
+                envs.insert(id.clone(), binding_type.clone());
+                linearizer.retype_ident(lin, id, binding_type);
+            }
+
+            stat_map
+                .iter()
+                .try_for_each(|(_, t)| -> Result<(), TypecheckError> {
+                    walk(state, envs.clone(), lin, linearizer.scope(), t)
+                })
```
francois-caddet, line 413: This is not a recursive record so inserting the fields idents into the type environment before type checking the values could cause Issues right? I mean issues like extra completions for the LSP and unbound variables not reported.

src/typecheck/mod.rs
```diff
@@ -1259,7 +1396,7 @@ pub fn unify_(
 
 /// Try to unify two row types. Return an [`RowUnifError::IllformedRow`] error if one of the given
 /// type is not a row type.
-pub fn unify_rows(
+pub fn unifyrows(
```
xlambein, line 1399: Did you do `s/unify_/unify/g`? :smile:

xlambein (approved):

src/typecheck/mod.rs
yannham: You're right. I think it wouldn't cause any issue in practice because all records are recursive currently, so probably this match case is just never entered. But still, it's incorrect.

src/typecheck/mod.rs
yannham: Ah, guilty as charged. Well spotted :smiley:

src/typecheck/mod.rs
```diff
@@ -1104,28 +1262,7 @@ fn row_add(
...
-/// Try to unify two types.
-pub fn unify_(
-    state: &mut State,
-    mut t1: TypeWrapper,
-    mut t2: TypeWrapper,
-) -> Result<(), UnifError> {
+pub fn unify(state: &mut State, mut t1: TypeWrapper, mut t2: TypeWrapper) -> Result<(), UnifError> {
     if let TypeWrapper::Ptr(pt1) = t1 {
         t1 = state.table.root(pt1);
     }
```
francois-caddet, line 1225: Does this function really update t1?

src/typecheck/mod.rs
francois-caddet: Yes and for generated AST it can be an issue

yannham: But I don't think we'll typecheck a generated AST, though. I mean if you are thinking about the Nix to Nickel compilation.

src/typecheck/mod.rs
yannham: Yes? Why not? It's a mutable variable in scope. If it wouldn't, then what `t1` would it update? (Also, if it wouldn't, most of the typechecking process would probably be utterly broken). On the other hand, I personally don't really like this way of having a mutable parameter (I would shadow `t1` with another `t1` both being immutable). But this code wasn't introduce by this PR, so if we prettify it, let's do that in another one :slightly_smiling_face:

PR #735, opened 2022-06-15
Separate the WASM build in a wrapper crate

### LINKED ISSUE #695: Make a proxy crate for the WASM REPL
As mentioned in a [discussion](https://github.com/tweag/nickel/pull/679#discussion_r847100964), the instructions to manually build (understand without Nix) the WebAssembly REPL are complex and error-prone. Even with Nix, while usage is easy, the manual process is just incorporated as a build script, and thus it's not free either in maintenance or onboarding new developers.

The situation is due in part to current limitations of Cargo, but one technique that would make things slightly better and seems to be used elsewhere is to make a separate crate which just exposes the `nickel` crate but with the right setting for being built by `wasm-pack`. See the full discussion for details. Let's try that.

yannham:
Close #695. Make a new crate that simply wraps `nickel-lang` and re-export its API, but with the right setting for being build using WASM to get rid of all the monkey patching that needed to be done before. It has worked quite straightforwardly, and greatly ease the procedure of building the WASM REPL.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1245,6 +1245,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "nickel-repl"
+version = "0.1.0"
+dependencies = [
+ "nickel-lang",
+]
+
+[[package]]
 name = "nix"
 version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


garbas (approved):
Very nice idea to use a separate cargo workspace to work around Cargo.toml drawbacks.

thufschmitt (approved):
Nice :) As much as I like Nix, it's cool to keep the build logic where it belongs

PR #737, opened 2022-06-16
Syntax: allow ' in identifers

### LINKED ISSUE #708: Let use `'` in identifiers names
In Nix and also Haskell I think, `'` is a valid character for identifiers.

Example from `nixpkgs/lib/lists.nix`:

```
foldr = op: nul: list:
    let             
      len = length list;
      fold' = n:     
        if n == len
        then nul                                                                  
        else op (elemAt list n) (fold' (n + 1));
    in fold' 0;
```

Is it a good improvement for nickel? It shouldn't conflict with the rest of the syntax.

Does not receve a unexpected tocken error when righting something like:

```
let fold' = fun n => ...
```

Don't make changes of nickel and use `_` instead of `'` in these cases.

**Additional context**

fuzzypixelz:
Resolves #708.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -4,6 +4,10 @@ slug: syntax
 
 # Nickel Syntax
 
+## Identifiers
+
+Nickel identifiers start with an alphabetic character, followed by zero or more alphanumeric characters, `_` (underscores) or `'` (single quotes). For example, `this-isn't-invalid` is a valid identifier.
+
 ## Simple values
 
 There are four basic kind of values in Nickel :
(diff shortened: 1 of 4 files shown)


doc/manual/syntax.md
```diff
@@ -4,6 +4,10 @@ slug: syntax
 
 # Nickel Syntax
 
+## Identifiers
+
+Nickel identifiers start with an alphabetic character, followed by alphanumeric characters in addition to `_` (underscore) and `'` (single quote). For example, `this-isn't-invalid` is valid syntax.
```
yannham, line 9:
```suggestion
Nickel identifiers start with an alphabetic character, followed by zero or more alphanumeric characters, `_` (underscores) or `'` (single quotes). For example, `this-isn't-invalid` is valid identifier.
```

yannham (approved):

PR #745, opened 2022-06-22
Fix random order of fields in `nickel doc`

yannham:
Depends on #743. `nickel doc` doesn't sort the fields of records, and because Rust hashmaps are randomized, two invocations of on the same file will order fields in a different way. This PR fixes that by always sorting fields before processing them. Passing by, I replaced a few `sort_by` calls by `sort_by_key` elsewhere, which are arguably simpler.

### DIFF
--- a/src/pretty.rs
+++ b/src/pretty.rs
@@ -26,7 +26,7 @@ fn min_interpolate_sign(text: &str) -> usize {
 
 fn sorted_map<'a, K: Ord, V>(m: &'a HashMap<K, V>) -> Vec<(&'a K, &'a V)> {
     let mut ret: Vec<(&K, &V)> = m.iter().collect();
-    ret.sort_by(|x, y| x.0.cmp(&y.0));
+    ret.sort_by_key(|(k, _)| *k);
     ret
 }
 
--- a/src/program.rs
+++ b/src/program.rs
@@ -283,8 +283,12 @@ mod doc {
             Term::MetaValue(MetaValue { doc: Some(md), .. }) => {
                 document.append(&mut parse_documentation(header_level, arena, &md, options))
             }
-            Term::Record(hm, _) | Term::RecRecord(hm, _, _, _) => {
-                for (ident, rt) in hm {
+            Term::Record(map, _) | Term::RecRecord(map, _, _, _) => {
+                // Sorting fields for a determinstic output
+                let mut entries: Vec<(_, _)> = map.iter().collect();
+                entries.sort_by_key(|(k, _)| *k);
+
+                for (ident, rt) in entries {
                     let header = mk_header(&ident.label, header_level + 1, arena);
                     document.append(header);
                     to_markdown(rt, header_level + 1, arena, document, options)?;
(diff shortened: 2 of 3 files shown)


ErinvanderVeen (approved):

francois-caddet:
Around the more general issue of map ordering, we could have a look to this crate:
https://lib.rs/crates/indexmap
May it be used for nickel? They are using https://lib.rs/crates/hashbrown backend. It seams to be the implementation used in `std::collections:HashMap` but not sure if they use the same hasher.

yannham:
I wonder if we want to preserve order of insertion, or enforce alphabetic order (which has the advantage of being reliable and deterministic, at the cost of performance of iteration probably).

ErinvanderVeen:
I personally would prefer the order of insertion. This allows user to enforce their preferred ordering, keeping similar functions close together.

francois-caddet:
Yes and throw errors in same order as they apeer in the file. At least if we are sure we parse sequencialy top to botom. May transformations break this idea? Also parallelism could break it but I don't think parallelism at this level is usefull. @yannham In what sens you think its more reliable?

yannham:
It is more reliable because it is preserved by more transformations. Typically if I re-order the fields of a record inside a source file, that won't change anything if we sort alphabetically, but it will if we use order of insertion. Alphabetical order is predictable: I can know in advance, just knowing the name of the fields, in what order they will be iterated over. This is not the case for order of insertion, which depends on the definition of the record and the various operations applied to it therafter.

You could imagine that both the `.ncl` files and the `.json` files are gited in some situations (support users that don't have nickel). Then, with insertion order, reordering some definitions would create a diff for the JSON output.

The drawback I can see is that you have less control on the layout of the generated files, indeed. I tend to think that generated files ought to be read by machines though, and that the source of truth is the nickel source. For errors appearing in the same order as they appear in the source file, I wonder if that's really important? Maybe because users are used to that, and that would be strange to report first an error on the last defined field, and then another one at the beginning? But I feel like that isn't a very strong argument either: they will have to fix both anyway. What's important is that the order is deterministic so that they know for sure when they fixed one error and made progress.

I personally don't have a very strong opinion right now, but I don't see an objectively strong argument in favor of order of insertion. It impacts generated files and the order of error messages, which I don't think are really important (beside, once again, that such a deterministic order exist). You can still organize your Nickel source files as you please.

For the record, the order of iteration of object properties is not exactly the insertion order (a mix of insertion order + natural ordering on numeric keys, IIRC).

aspiwack:
> The drawback I can see is that you have less control on the layout of the generated files, indeed. I tend to think that generated files ought to be read by machines though, and that the source of truth is the nickel source.

This is true of Json files, but not if we output a documentation (either in standard output or in an html document à la `cargo doc`).

In Json documents, though, alphabetic ordering is superior, when it comes to diffing (it doesn't matter if you have a dedicated tree differ, but in this day and age, we cannot assume that).

aspiwack:
I pressed « comment » before I read the title of this PR. If this is for `nickel doc`, then this is for human-readable formats not computer-readable format, isn't it?

PR #749, opened 2022-06-24
WIP: Make Array contracts lazy

fuzzypixelz:


fuzzypixelz:
Some of the methods on `ArrayAttrs` might need to be changed.

fuzzypixelz:
We will need to store both the array/record contracts _and_ the labels inside `ArrayAttrs`.

fuzzypixelz:
`ArrayLazyAssume` is currently a binary operator. It takes a contract and an array, then returns the same array carrying the contracts. Ideally we would like it to also take a label but that would require making it a ternary operator. I'll go ahead with this now.

fuzzypixelz:
Now `%array_lazy_assume% : forall a. Contract -> Label -> Array a -> Array a`, a bit like the actual `Assume` but now the third argument is strict.

A few more issues surfaced:
- How do we store `Contract + Label` in `ArrayAttrs`? Of course we could just put in a tuple but a struct with proper fields would be more readable, I just need a name for it :)
- We need to somehow preserve the environment of the contract. We cannot simply put the full closure into the AST so merging with the array's environment would be a good (naive) idea.

A couple notes:
- All `NAry` operators are ternary operators; time for another special case?
- Are operators stylized in `camelCase` or `snake_case` in error output?

fuzzypixelz:
I just rebased against `master` so that the conflicts don't get out of hand.

fuzzypixelz:
At this point, we've mostly recovered the original behaviour. For example:

```nickel
([-1] | Array Pos) @ ([1] | Array Pos) |> array.elem_at 1 =>
    1
    
([-1] | Array Pos) @ ([1] | Array Pos) |> array.elem_at 0 =>
  ┌─ :1:7
  │
1 │ Array Pos
  │       --- expected array element type
  │
  ┌─ repl-input-4:1:3
  │
1 │ ([-1] | Array Pos) @ ([1] | Array Pos) |> array.elem_at 0
  │   ^^ applied to this expression
  │
  │
1 │ -1
  │ -- evaluated to this value

note: 
  ┌─ repl-input-4:1:9
  │
1 │ ([-1] | Array Pos) @ ([1] | Array Pos) |> array.elem_at 0
  │         ^^^^^^^^^ bound here
...
```

Now onto the failing tests, I guess.

fuzzypixelz:
I've been hunting down a regression in record evaluation. The minimal example to observe the bug is this:

```nickel
builtin.serialize `Json ([{y = "hello"}] | Array {y | Str})
```

On the latest version of this branch, c9720101b5910fb4ee3eeadc70f81e4d0428d2c4, the above fails with:

```
error: non serializable term
  ┌─ repl-input-0:1:32
  │
1 │ builtin.serialize `Json ([{x = "hello"}] | Array {x | Str})
  │                                ^^^^^^^
```

More specifically, the `StrChunks` corresponding to `"hello"` isn't evaluated and therefore isn't serializable (`validate` fails).

This only seems to happen when we have a contract on the array.

yannham:
Ah, I think I see the issue. We have to maintain the invariant during the whole evaluation that a data structure (record or array) only contains variables inside. This is ensured after parsing by the `share_normal_form` transformation, and maintained by the various evaluation steps. You probably break this invariant when mapping the contract over arrays before closurizing them, because you replace their element with something of the form `%apply_contract% _ _ _ `. The solution is to closurize the result of `apply_contracts`.

fuzzypixelz:
> You probably break this invariant when mapping the contract over arrays before ~closurizing~ concatenating them, because you replace their elements with something of the form `%apply_contract% _ _ _ `. The solution is to closurize the result of `apply_contracts`.

You are completely right, I did break the invariant. But even after closurizing the elements and making sure that `DeepSeq` is applied to variables (via debug-printing), the issue persists.

This problem doesn't happen with `Array {x | Num}` or `Array {x : Str}` or indeed `Array {x | Array Num}`.

fuzzypixelz:
I took the liberty of modifying `seq_terms`'s signature to sidestep an ownership issue. Basically the iterator applying contracts has a mutable reference to `shared_env` and so I can't pass this iterator _and_ the `shared_env` to `seq_terms`. It's easy enough to remove `env` from `seq_terms`'s parameters.

fuzzypixelz:
Work is continued on #764.

PR #751, opened 2022-06-24
[Refactor] Get rid of is_xxx primops in favor of typeof

yannham:
I needed to add a missing `builtin.is_enum` function, and the workflow is quite heavy:

- add a `is_enum` token in the lexer
- add corresponding cases in the grammar
- add an `IsEnum()` unary op and its implementation
- add the type of `IsEnum()` to `typecheck::operation` so that we can typecheck it
- add a case in `builtin.typeof`

It turns out our primops are all the `is_xxx` functions, and `builtin.typeof` is defined as a big if-then-else applying those functions in order. This is not really optimal.

This PR instead implements only one primop `typeof` once and for all. The `builtin.is_xxx` are just defined as standard functions using `typeof`.

Incidentally improve the macros `mk_ty_enum_rows/mk_ty_enum` slightly by making the tail optional.


src/grammar.lalrpop
```diff
@@ -502,6 +502,15 @@ StaticString: String = StringStart <s: ChunkLiteral?> StringEnd => s.unwrap_or_d
 EnumTag: Ident = {
     <Ident>,
     <StaticString> => <>.into(),
+    // Allow tags of the form `Str, `Bool, etc. This is ad-hoc and should
+    // eventually be replaced with a proper lexing of enum tags, but this is not
+    // a trivial task. For now, to make builtin.typeof usable, we just
+    // special-case builtin types.
```
ErinvanderVeen, line 508: Is there/Should there be an issue for this?

PR #757, opened 2022-06-28
Allow unused imports in LALRPOP

fuzzypixelz:
This may be unwarranted, but I would like the warning to go away.

it seems that if one has a grammar with multiple top-level pub parsers. Then LALRPOP creates modules for each of them, and inlines the grammar's imports _inside each of the modules_, but still keeps the global use imports present in the final rust program.

So in our case, we end up with something like:

```rust
use imports;

mod __parse__ExtendedTerm {
  use imports;
}

mod  __parse__Term {
  use imports;
}
```

Now you could easily see where the redundant imports come from.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -7,6 +7,7 @@ use lalrpop_util::lalrpop_mod;
 lalrpop_mod!(
     #[allow(clippy::all)]
     #[allow(unused_parens)]
+    #[allow(unused_imports)]
     pub grammar);
 
 pub mod error;


fuzzypixelz:
The problem with this is that we may end up with _actual_ unchecked redundant imports.

yannham:
I'm a bit divided. I don't know if those warnings are really an issue: we quickly get used to it, and know that we should ignore it. it's a bit annoying, true. But the proposed solution is at risk of missing unused imports, which is not tragic either, but I tend to think that having false negatives is slightly worse than having false positives?

steshaw:
Those unused import warnings really bothered me in the past, also. It's like a constant distraction, perhaps exacerbated at the time because the lalrpop seemed to run on every change.

aspiwack:
The standard Tweag stance on the matter: CI should run with `-Werror` (turn warnings into errors). So there is no warning on master. If you don't intend to address the source of the warnings, then silence the warnings. You can de-silence them when/if you decide to address them.

yannham (approved):

PR #762, opened 2022-07-04
Share the data among reverted normal thunks

yannham:
When merging two records, the implementation of `merge` reverts all the thunks of both operands. If a thunk is revertible, `revert` create a new, independent thunk, with its value reset to the original expression. This is expected. For non revertible thunks (which may happen often: any field that is not depending recursively on other fields will be allocated as a standard, non-revertible thunk), the operation should just be a clone.

However, `revert` is currently cloning the inner value of the thunk, while it could just clone the `Rc` and share the same value. There is no strong reason to do that: it's more expensive, and if one thunk is evaluated, the result is not shared with the other. We discovered this behavior with @fuzzypixelz while working on a unrelated bug. This PR clones the outer reference instead.

### DIFF
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -107,22 +107,27 @@ impl ThunkData {
         self.state = ThunkState::Evaluated;
     }
 
-    /// Create fresh unevaluated thunk data from `self`, reverted to its original state before the
-    /// first update. For standard thunk data, the content is unchanged and the state is conserved:
-    /// in this case, `revert()` is the same as `clone()`.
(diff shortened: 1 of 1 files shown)


ErinvanderVeen (approved):

yannham:
For the record, I ran the benchmarks showing mostly positive improvements between 6-12%, higher for the Mantis benchmark. But sometimes a few negative hits on things that shouldn't be affected by revertibility. Honestly the variance is often high, although I killed the window manager for the sake of the experience, so I wouldn't put too much trust in this measurement. At least it didn't seem to make a clear negative impact.

PR #763, opened 2022-07-05
Implement type equality for contracts

### LINKED ISSUE #724: Deciding type equality for two flat types (aka contracts aka opaque types)
Contracts (and more generally, even if that may be meaningless at runtime, any valid term) can appear inside a type annotation. The semantics, as far as the typechecker is concerned, is that those terms are seen as black-box opaque types, which can only unify with themselves. For example:

```nickel
let FooContract = ... in
{
  x : FooContract = let bar = (5 | FooContract) in bar
}
```

This typechecks, because the contract application `5 | FooContract` ensures to either fail with a contract error or that `5` satisfies `FooContract`. Thus, indeed, `bar : FooContract`.

However, the current check is very fragile, and even unsound:

- unifying two flat types requires them to be variables, or fails otherwise (see #701). This is far too limiting for many use-cases.
- if they are variables, it just compares their names. But names may be shadowed, so this is unsound. For example, the following passes:

        let FooContract = Str in
        let foo : FooContract = (let FooContract = Num in (5 | FooContract)) in
        foo

  This typechecks, but fails at runtime with a contract error (ouch).

This situation leads to the following question: how should we decide type equality between two contracts (flat types)? For variables, we must have a way to check that they refer to the same definition. We may also want to support more complex expressions, like contract application: `Nullable Str` should be unifiable with `Nullable Str`. On the other hand, we probably don't want to do general substitution and allow `let BarContract = FooContract` to be equal to `FooContract`, because this is a slippery slope that could lead to arbitrary computations being run during typechecking.

yannham:
Closes #724.


fuzzypixelz (approved):
Looks good to me. I left some nitpicks here and there.

src/typecheck/eq.rs
```diff
@@ -0,0 +1,356 @@
...
+                .map(|v2| f(state, v1, env1, v2, env2))
+                .unwrap_or(false)
+        })
+}
+
+/// Convert rows (represented as linked list for the needs of unification) to a hashmap to perform
+/// the type equality comparison.
+///
+/// Require the rows to be closed (i.e. the last element must be RowEmpty), otherwise `None` is
+/// returned. `None` is returned as well if a type encountered is not a row type, a if it is a enum
```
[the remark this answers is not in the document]
yannham: Not really, because I think this function is called only for rows of a record. However, you made me realize I didn't treat enums properly, and the current version probably requires all the tags to be the same in the exact same order...

src/typecheck/eq.rs
```diff
@@ -0,0 +1,356 @@
...
+//! want to equate `Alias` with e.g. `Foo "bar" "baz"` if `Alias` has been defined as
+//! `let Alias = Foo "bar" "baz" in ...`.
+//!
+//! We also want to equate different aliases with the same definition: `Alias type_eq Alias'` if
+//! `let Alias' = Foo "bar" "baz" in ...`, or `let Alias' = Alias in ...`.
+//!
+//! ## Recursion
+//!
+//! We must refrain from following all variables links blindly, as their could be cycles in the
+//! graph leading to an infinite loop:
```
ErinvanderVeen, line 18:
```suggestion
//! We must refrain from following all variables links blindly, as there could be cycles in the
//! graph leading to an infinite loop:
```

src/typecheck/eq.rs
```diff
@@ -0,0 +1,356 @@
...
+//! ```nickel
+//! {
+//!   Foo = Bar,
+//!   Bar = Foo,
+//! }
+//! ```
+//!
+//! Because we just follows variables, and don't apply functions, we can detect cycles while
+//! walking the graph. Still, as it is performed potentially many times during typechecking, type
+//! equality ought to stay reasonably cheap. We choose to just set an arbitrary limit (the gas) on
```
ErinvanderVeen, line 29:
```suggestion
//! walking the graph. Still, as it is potentially performed many times during typechecking, type
//! equality ought to stay reasonably cheap. We choose to just set an arbitrary limit (the gas) on
```

src/typecheck/eq.rs
```diff
@@ -0,0 +1,356 @@
...
+            }
+            _ => return None,
+        }
+    }
+}
+
+/// Perform the type equality comparison on types. Structurally recurse into type constructors and test
+/// that subtypes or subterms (contracts) are equals.
+///
+/// Currently, this function morally operates on `Types` rather than `TypeWrapper`s as it is called
```
ErinvanderVeen, line 279: `morally`?

src/typecheck/eq.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Used to generate temporary rigid type variables for substituting type variables when
+    /// comparing foralls. Those ids never escape the type equality computations and are used
+    /// solely as rigid type variables: this is why they don't need proper allocation in the
+    /// unification table or to care about those ids clashing with the one generated by the
+    /// typechecker. Generated type constants simply needs to be unique for the duration of the
+    /// type equality computation.
+    var_uid: usize,
+    /// The current gas remaining for variable substitutions. Once it reaches zero and we encounter
+    /// a variable, we abort the computation and return false.
+    gas: u8,
```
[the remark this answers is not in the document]
yannham: After testing, indeed u32/u32 and u32/u8 are the same. I think the latter is best, because `gas` ought to stay small. All in all you are right, but in this specific case I don't think 8 bytes for a single parameter passed around inside a casual function like this one really matter (this state is not held very long in memory). I would do the change if it were atomic, but unification variables are already `usize` elsewhere (as they are used to index inside the unification table), so using an u32 here would either need some conversions or to change the type of `TypeWrapper::XXX` variants. I propose to  keep it like this for now

src/typecheck/eq.rs
yannham: In the sense that the official type is `TypeWrapper`, but "in fact", "actually", it is operating on the subset `Types`. Is this use of `morally` strange, uncommon or totally invalid? I've used it for a long time, but it wouldn't be the first time one discovers they've been done something wrong their whole life :sweat_smile:

ErinvanderVeen: I have not seen it used this way. Feel free to resolve this conversation anyway.

PR #764, opened 2022-07-06
Make Array contract application lazy

fuzzypixelz:
# The problem

Right now applying an Array contract does something that resembles:

```
%map% t (fun value => %assume% elt (%go_array% l) value)
```

This essentially puts the array elements behind a mapping of the contract on each element. Meaning that simply taking the head of the array will trigger `n = array.length xs` operations, even though we only need one (this is because `map` is a strict primitive operation implemented in the interpreter). Also, if we take the head of the array multiple times, we will needlessly apply the contract again and again.

# How this improves the situation

When an Array receives a contract application, the contract is simply stored in its attributes and only applied when data leaves the Array, or when we merge with another Array with potentially incompatible contracts.

# Issues

This code seems to break `DeepSeq`, especially when serializing terms. This is addressed in https://github.com/tweag/nickel/pull/767 by adding a proper `Force` operator.


fuzzypixelz:
Sure. I'll close the original PR.

src/term.rs
```diff
@@ -183,14 +183,69 @@ impl Default for BindingType {
...
+        self
+    }
+
+    /// Drop the pending contracts.
+    pub fn contracts_cleared(mut self) -> Self {
+        self.pending_contracts.clear();
+        self
+    }
+
+    /// Extend contracts from an iterator of `PendingContract`.
```
yannham, line 234:
```suggestion
    /// Extend contracts from an iterator of `PendingContract`.
    /// De-duplicate equal contracts. Note that current contract
    /// equality testing is very limited, but this may change in the
    /// future
```

src/term.rs
```diff
@@ -836,17 +891,34 @@ pub enum UnaryOp {
     /// Transform a string to an enum.
     EnumFromStr(),
     /// Test if a regex matches a string.
-    /// Like [`StrMatch`], this is a unary operator because we would like a way to share the
+    /// Like [`UnaryOp::StrMatch`], this is a unary operator because we would like a way to share the
```
yannham, line 894: Just out of curiosity, didn't `` [`StrMatch`] `` work? I thought we tested those links when building the documentation of 0.1.0.

src/term.rs
```diff
@@ -836,17 +891,34 @@ pub enum UnaryOp {
...
     /// Match a regex on a string, and returns the captured groups together, the index of the
     /// match, etc.
     StrMatch(),
-    /// Version of `StrIsMatch` which remembers the compiled regex.
+    /// Version of [`UnaryOp::StrIsMatch`] which remembers the compiled regex.
     StrIsMatchCompiled(CompiledRegex),
-    /// Version of `StrMatch` which remembers the compiled regex.
+    /// Version of [`UnaryOp::StrMatch`] which remembers the compiled regex.
     StrMatchCompiled(CompiledRegex),
+    /// Force "full" evaluation of a term and return it.
```
yannham, line 905:
```suggestion
    /// Force full evaluation of a term and return it.
```

src/term.rs
```diff
@@ -836,17 +891,34 @@ pub enum UnaryOp {
...
+    ///
+    /// This was added in the context of [`BinaryOp::ArrayLazyAssume`],
+    /// and may not make much sense on its own.
```
yannham, line 908:
```suggestion
    /// This was added in the context of [`BinaryOp::ArrayLazyAssume`],
    /// in particular to make serialization work with lazy array contracts.
```

I'm not sure it doesn't make sense. It may make more sense than the current `deep_seq`, somehow (it evaluates more). If I may make a suggestion, we could stick to the motivation :slightly_smiling_face:

src/typecheck/operation.rs
```diff
@@ -195,6 +195,8 @@ pub fn get_uop_type(
                 ("groups", mk_typewrapper::array(AbsType::Str()))
             ),
         ),
+        // Dyn -> Dyn
+        UnaryOp::Force(_) => (mk_typewrapper::dynamic(), mk_typewrapper::dynamic()),
```
yannham, line 199: Why does this appear in this PR?

src/eval/operation.rs
```diff
@@ -1851,15 +1964,53 @@ fn process_binary_operation(
...
+                                apply_contracts(t, ctrs_left.clone(), pos1)
+                                .closurize(&mut env, env1.clone())
+                            ));
+
+                            ts.extend(ts2.into_iter().map(|t|
+                                apply_contracts(t, ctrs_right.clone(), pos2)
+                                .closurize(&mut env, env2.clone())
+                            ));
+
+                            let ctrs_common = attrs1
```
yannham, line 2001: I think you can use `std::iter::partition` above to compute `ctrs_left` and `ctrs_common` at once above, instead of doing another pass here. Maybe that will saved one `cloned()` as well

src/eval/operation.rs
```diff
@@ -2145,6 +2301,58 @@ fn process_binary_operation(
...
+                        term: t2,
+                        pos: pos2,
+                    },
+                ))
+            };
+
+            match_sharedterm! {t2,
+                with {
+                    Term::Array(ts, attrs) => {
+
```
yannham, line 2331:
```suggestion
```

src/term.rs
```diff
@@ -183,14 +183,69 @@ impl Default for BindingType {
...
+    }
+
+    /// Drop the pending contracts.
+    pub fn contracts_cleared(mut self) -> Self {
+        self.pending_contracts.clear();
+        self
+    }
+
+    /// Extend contracts from an iterator of `PendingContract`.
+    pub fn with_contracts<I>(mut self, iter: I) -> Self
```
yannham, line 235: I find the name to be slightly ambiguous. Usually `with_xxx` means that `xxx` will be set to the new parameter (a form of `set` + return self), while here it _extends_ the current one. Maybe a name like `with_extra_contracts` or the like would be clearer?

src/eval/operation.rs
```diff
@@ -2408,11 +2616,55 @@ fn eq(env: &mut Environment, c1: Closure, c2: Closure) -> EqResult {
...
         }
-        (Term::Array(l1, _), Term::Array(l2, _)) if l1.len() == l2.len() => {
+        (Term::Array(l1, a1), Term::Array(l2, a2)) if l1.len() == l2.len() => {
             // Equalities are tested in reverse order, but that shouldn't matter. If it
             // does, just do `eqs.rev()`
-            let eqs = l1.into_iter().zip(l2.into_iter());
-            gen_eqs(eqs, env, env1, env2)
+
+            // We should apply all contracts here, otheriwse we risk having wrong values,
+            // think record contrats with default contracts, Wrapped terms, etc.
```
yannham, line 2624:
```suggestion
            // think record contrats with default contracts, wrapped terms, etc.
```

yannham (approved):

src/term.rs
fuzzypixelz: No, they're only highlighted as code. I discovered this while adding documentation for `Force` and testing how it renders.

Just doubled-checked on master.

src/term.rs
fuzzypixelz: Oh I see! Is there some definitive reference for Rust API naming conventions?

src/typecheck/operation.rs
fuzzypixelz: Because we merged in `feature/force-operator`?

src/pretty.rs
```diff
@@ -238,7 +238,7 @@ where
                         }, // be sure we have at least 1 `%` sign when an interpolation is present
                     )
                     .max()
-                    .unwrap();
+                    .unwrap_or(1);
```
yannham, line 241: Can we get rid of the `else { 1 }` above if we use `1` as a default value here? But I think those kind of improvements, albeit small, would be better off as a separate PR.

src/term.rs
```diff
@@ -183,14 +183,70 @@ impl Default for BindingType {
     }
 }
 
+/// A contract with its associated data.
+#[derive(Debug, PartialEq, Clone)]
+pub struct PendingContract {
+    /// The "pending" contract, can be a function or a record.
```
yannham, line 189:
```suggestion
    /// The pending contract, can be a function or a record.
```

src/interner.rs
```diff
@@ -0,0 +1,10 @@
+//! A string interner.
```
yannham, line 1: I guess this is unintended :slightly_smiling_face:

yannham (approved):

src/interner.rs
fuzzypixelz: For sure!

PR #765, opened 2022-07-06
Benchmark array contracts

fuzzypixelz:
Add benchmarks in the context of #764 so we can evaluate results.

### DIFF
--- a/benches/arrays/primes.ncl
+++ b/benches/arrays/primes.ncl
@@ -1,0 +1,42 @@
+let range
+  | doc "Generate an array of integers in the range [`start`, `end`)."
+  | Num -> Num -> Array Num
+  = fun start end =>
+    if end <= start then 
+      []
+    else
+      array.generate (fun x => x + start) (end - start) 
+in
+
+let is_prime 
+  | doc "Returns true if the argument is a prime number."
+  = fun x => x > 1 && array.all (fun d => x % d != 0) (range 2 (x - 1))
+in
+
+let Prime = fun label value =>
+  if is_prime value then
+    value
+  else
+    contract.blame_with "%{string.from_num value} is not a prime number." label 
+in
+
+let primes 
+  | doc "Generate `max` primes using Sieve of Eratosthenes."
+  | Num -> Array Prime
+  = fun max =>
+    let limit = num.pow max (1 / 2) in # sqrt(max)
+    let drop_multiples = fun x xs => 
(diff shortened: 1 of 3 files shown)


yannham (changes requested):
It seems to be missing the `arrays/sum.ncl` file at least, or am I missing something?

fuzzypixelz:
Just a quick mental note.

We need benchmarks that specifically target

benches/arrays/sum.ncl
```diff
@@ -0,0 +1,12 @@
+let sum
+  | Array Num -> Num
+  = let rec aux = fun step xs =>
```
yannham, line 3: If the goal of this bench is a case where a lot of array contracts are applied at each recursive call, I would not care about stack space/tail calls and write the naive sum function `sum [] = 0, sum x:xs = 1 + sum xs`. Here, the `Array Num` input contract is only applied on the call to `sum`, but not to actual recursive calls to `aux` (although there's still `array.length` which applies an array contract each times), I think.

benches/arrays/primes.ncl
```diff
@@ -0,0 +1,42 @@
...
+let is_prime 
+  | doc "Returns true if the argument is a prime number."
+  = fun x => x > 1 && array.all (fun d => x % d != 0) (range 2 (x - 1))
+in
+
+let Prime = fun label value =>
+  if is_prime value then
+    value
+  else
+    contract.blame_with "%{string.from_num value} is not a prime number." label
```
yannham, line 20: Not very important for a benchmark, but the idiomatic way would be to write 

```nickel
let Prime = contract.from_predicate is_prime in
```

yannham (approved):

benches/arrays/primes.ncl
fuzzypixelz: I was aware of that. But I thought the error message could not be customized. Would this make for a new library function? i.e. `contract.from_predicate_with`? (the name is not the strongest argument here :P)

yannham:
@fuzzypixelz it's a bit harder to follow and to review when adding new commits to this PR, which has already been approved twice. What about dropping the last commit (I think it adds a new benchmark? Also the commit contains trailing git conflict markers), merge as it is once the CI is green, and open a new PR for other benchmarks?

yannham:
Thanks! LGTM :)

PR #771, opened 2022-07-22
Implement type equality for enums

yannham:
Add missing type equality computations for enum types (which are just compared as sets of identifiers).

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -1149,6 +1149,15 @@ impl TypeWrapper {
             Ptr(x) => Ptr(x),
         }
     }
+
+    /// Create an iterator on rows represented by this type.
+    ///
+    /// The iterator continues as long as the next item is of the form `RowExtend(..)`, and
+    /// stops once it reaches `RowEmpty` (which ends iteration), or something else which is not a
+    /// `RowExtend` (which produces a last item [`typecheck::RowIteratorItem::Tail`]).
+    pub fn iter_as_rows(&self) -> RowIterator<'_> {
+        RowIterator { next: Some(self) }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/typecheck/mod.rs
```diff
@@ -1149,6 +1149,15 @@ impl TypeWrapper {
...
     }
+
+    /// Create an iterator on rows represented by this type.
+    ///
+    /// The iterator continues as long as the next item is of the form `RowExtend(..)`, and
+    /// stops once it reaches `RowEmpty` (which ends iteration), or something else which is not a
+    /// `RowExtend` (which produces a last item [`typecheck::RowIteratorItem::Tail`]).
+    pub fn iter_as_rows(&self) -> RowIterator<'_> {
+        RowIterator { next: Some(self) }
+    }
```
francois-caddet, line 1160: Should we check the variant to be sure it's a Row? Or do we consider any type can give an iterator but with a single element? It could make sens if we have an iterator for Arrow also? May not be useful now except for the pretty printer ;)

francois-caddet (approved):

src/typecheck/mod.rs
yannham: Actually, I initially wrote this function as returning an `Option<RowIterator<'_>>`. But it is meant to be called on the *inner* value of a `DynRecord` or a `Enum`, where currently we can have e.g. `Dyn`, and may have more in the future. So in the end I choose to let the responsibility of a potential check to the caller, because as far as `iter_as_rows` is concerned, it can produce an iterator in any case. It will just not be very meaningful in some cases.

yannham: For arrows and other things, let's do on a need basis: if that makes your life simpler in the pretty printer, don't hesitate to do so!

PR #773, opened 2022-07-25
quick and dirty fix of lsp error when position is None

### LINKED ISSUE #741: NLS (LSP) fails when putting types inside terms
**Describe the bug**
When using a type inside a term (that is, outside of an annotation), the LSP panics.

**To Reproduce**
Open a file with the following content:

```nickel
let foo = Str in
5
```

The LSP client will panic.

**Expected behavior**
The LSP should handle those cases just fine.

**Additional context**

This is caused by a call to `unwrap` inside `linearization/mod.rs` on the position of identifiers. All identifiers were expected to have a position set just after parsing. However, since #595, types inside terms are actually converted to contracts on the fly. The first line of the above example will typically be rewritten by the parser as `let foo = $string in`, where `$string` is a special identifier with no position. Even if we could fix this case by assigning the same position as the original expressoin `Str`, I don't think this will work as easily with more complex types having `Str` as components, because types don't have a position right now.

francois-caddet:
should fix #741

### DIFF
--- a/lsp/nls/src/linearization/completed.rs
+++ b/lsp/nls/src/linearization/completed.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use codespan::ByteIndex;
 use nickel_lang::{
+    position::TermPos,
     term::MetaValue,
     typecheck::linearization::{LinearizationState, Scope},
 };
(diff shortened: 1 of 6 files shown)


lsp/nls/src/linearization/completed.rs
```diff
@@ -80,19 +81,37 @@ impl Completed {
...
+                let pos = item.pos.unwrap();
+                (pos.src_id, pos.start)
+            }
+        }) {
             // Found item(s) starting at `locator`
             // search for most precise element
             Ok(index) => linearization[index..]
                 .iter()
-                .take_while(|item| (item.pos.src_id, item.pos.start) == *locator)
+                .take_while(|item| {
```
yannham, line 96: same remark as above

lsp/nls/src/linearization/completed.rs
```diff
@@ -80,19 +81,37 @@ impl Completed {
...
+                        (pos.src_id, pos.start)
+                    };
+                    (src_id, start) == *locator
+                })
                 .last(),
             // No perfect match found
             // iterate back finding the first wrapping linearization item
             Err(index) => linearization[..index].iter().rfind(|item| {
-                let (istart, iend, ifile) = (item.pos.start, item.pos.end, item.pos.src_id);
+                let (istart, iend, ifile) = if item.pos == TermPos::None {
```
yannham, line 109: Same remark as above. You can also use `into_opt`, or just plain pattern matching, to avoid resorting to `unwrap`.

lsp/nls/src/linearization/mod.rs
```diff
@@ -437,7 +437,14 @@ impl Linearizer for AnalysisHost {
             scope,
         } = lin.into_inner();
 
-        linearization.sort_by_key(|item| (item.pos.src_id, item.pos.start));
+        linearization.sort_by(|it1, it2| match (it1.pos, it2.pos) {
+            (TermPos::None, TermPos::None) => std::cmp::Ordering::Equal,
+            (TermPos::None, _) => std::cmp::Ordering::Less,
+            (_, TermPos::None) => std::cmp::Ordering::Greater,
+            (pos1, pos2) => (pos1.unwrap().src_id, pos1.unwrap().start)
+                .partial_cmp(&(pos2.unwrap().src_id, pos2.unwrap().start))
```
yannham, line 445: Why do we need a `partial_cmp` that is then unwrapped? Doesn't a tuple `(T, U)` implements `Ord` if `T: Ord, U: Ord` ?

lsp/nls/src/requests/goto.rs
```diff
@@ -123,21 +123,25 @@ pub fn handle_to_usages(
...
-                        uri: Url::parse(&server.cache.name(src_id).to_string_lossy()).unwrap(),
-                        range: Range::from_codespan(
-                            &src_id,
-                            &(start as usize..end as usize),
-                            server.cache.files(),
-                        ),
-                    },
-                };
-                locations.push(location);
+                if reference.pos == TermPos::None {
```
yannham, line 126: Nitpick: I would use pattern matching or `into_opt` (even if the latter may cause an allocation, if not optimized away).

yannham (commented):
I get a new error when trying on Mantis, but it doesn't seem to be related to your change. Even on master, opening `stdlib/array.ncl` crashes the LSP on some other part. I'm bisecting right now

yannham (approved):
Other bug filed in another issue (#775). After a quick, local fix to #775, this PR indeed seems to fix #741.

PR #776, opened 2022-07-27
Array slices

fuzzypixelz:
This changes the Array term to use a combination of a range and an `Rc<[RichTerm]>` instead of `Vec<RichTerm>`. This makes it cheaper to:

- Clone `Term::Array`
- Take the tail of `Term::Array`

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -21,14 +21,14 @@ use crate::{
     position::TermPos,
     serialize,
     serialize::ExportFormat,
-    term::{make as mk_term, PendingContract},
+    term::{array::Array, make as mk_term, PendingContract},
     term::{ArrayAttrs, BinaryOp, NAryOp, RichTerm, StrChunk, Term, UnaryOp},
     transform::{apply_contracts::apply_contracts, fresh_var, Closurizable},
 };
 use md5::digest::Digest;
 use simple_counter::*;
-use std::collections::HashMap;
 use std::iter::Extend;
+use std::{collections::HashMap, rc::Rc};
 
 generate_counter!(FreshVariableCounter, usize);
 
@@ -472,6 +472,7 @@ fn process_unary_operation(
                         .collect();
                     fields.sort();
                     let terms = fields.into_iter().map(mk_term::string).collect();
+
                     Ok(Closure::atomic_closure(RichTerm::new(
                         Term::Array(terms, ArrayAttrs::new().closurized()),
                         pos_op_inh,
@@ -503,7 +504,7 @@ fn process_unary_operation(
                     Ok(Closure {
                         // TODO: once sure that the Record is properly closurized, we can
                         // safely assume that the extracted array here is, in turn, also closuried.
-                        body: RichTerm::new(Term::Array(terms, Default::default()), pos_op_inh),
+                        body: RichTerm::new(Term::Array(terms, ArrayAttrs::default()), pos_op_inh),
                         env,
                     })
                 }
@@ -730,7 +731,7 @@ fn process_unary_operation(
         }
         UnaryOp::ArrayHead() => {
             if let Term::Array(ts, attrs) = &*t {
-                if let Some(head) = ts.first() {
+                if let Some(head) = ts.get(0) {
                     let head_with_ctr = apply_contracts(
                         head.clone(),
                         attrs.pending_contracts.iter().cloned(),
@@ -755,10 +756,9 @@ fn process_unary_operation(
         }
         UnaryOp::ArrayTail() => match_sharedterm! {t, with {
                     Term::Array(ts, attrs) => {
-                        let mut ts_it = ts.into_iter();
-                        if ts_it.next().is_some() {
+                        if !ts.is_empty() {
                             Ok(Closure {
-                                body: RichTerm::new(Term::Array(ts_it.collect(), attrs), pos_op_inh),
+                                body: RichTerm::new(Term::Array(ts.advance_by(1), attrs), pos_op_inh),
                                 env,
                             })
                         } else {
@@ -858,6 +858,7 @@ fn process_unary_operation(
                     .chars()
                     .map(|c| RichTerm::from(Term::Str(c.to_string())))
                     .collect();
+
                 Ok(Closure::atomic_closure(RichTerm::new(
                     Term::Array(ts, ArrayAttrs::new().closurized()),
                     pos_op_inh,
@@ -1090,7 +1091,7 @@ fn process_unary_operation(
                 let capt = regex.captures(s);
                 let result = if let Some(capt) = capt {
                     let first_match = capt.get(0).unwrap();
-                    let groups: Vec<RichTerm> = capt
+                    let groups = capt
                         .iter()
                         .skip(1)
                         .filter_map(|s_opt| {
@@ -1111,7 +1112,10 @@ fn process_unary_operation(
                     mk_record!(
                         ("match", Term::Str(String::new())),
                         ("index", Term::Num(-1.)),
-                        ("groups", Term::Array(Vec::new(), Default::default()))
+                        (
+                            "groups",
+                            Term::Array(Array::default(), ArrayAttrs::default())
+                        )
                     )
                 };
 
@@ -1190,7 +1194,7 @@ fn process_unary_operation(
                             // It's important to collect here, otherwise the two usages below
                             // will each do their own .closurize(...) calls and end up with
                             // different variables, which means that `cont` won't be properly updated.
-                            .collect::<Vec<_>>();
+                            .collect::<Array>();
 
                         let terms = ts.clone().into_iter();
                         let cont = RichTerm::new(Term::Array(ts, attrs), pos.into_inherited());
@@ -2003,6 +2007,13 @@ fn process_binary_operation(
                             debug_assert!(attrs1.closurized, "the left-hand side of ArrayConcat (@) is not closurized.");
                             debug_assert!(attrs2.closurized, "the right-hand side of ArrayConcat (@) is not closurized.");
 
+                            // NOTE: To avoid the extra Vec allocation, we could use Rc<[T]>::new_uninit_slice()
+                            // and fill up the slice manually, but that's a nightly-only experimental API.
+                            // Note that collecting into an Rc<[T]> will also allocate a intermediate vector,
+                            // unless the input iterator implements the nightly-only API TrustedLen, and Array's iterator currently doesn't.
+                            // Even if we could implement TrustedLen we would have to contend with the fact that .chain(..) tends to be slow.
+                            // - Rc<[T]>::from_iter docs: https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-FromIterator%3CT%3E
+                            // - chain issue: https://github.com/rust-lang/rust/issues/63340
                             let mut ts: Vec<RichTerm> = Vec::with_capacity(ts1.len() + ts2.len());
 
                             let mut env = env1.clone();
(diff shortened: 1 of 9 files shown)


yannham (commented):
Left a few remarks, but the approach looks good and sound to me. @fuzzypixelz, however you told me that the benchmarks weren't conclusive?

benches/mantis/run.ncl
```diff
@@ -3,5 +3,5 @@
   run = import "deploy.ncl",
   serialize.run = fun args => (
         builtin.serialize `Json (run args)
-    )
-}
+    ),
+}.run { namespace = "mantis-staging", job="miner", role=`miner }
```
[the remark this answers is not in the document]
fuzzypixelz: Oh not a all. That's just a silly workaround I do to run a benchmark outside of Criterion.

src/eval/operation.rs
```diff
@@ -500,11 +512,11 @@ fn process_unary_operation(
                     // require to copy the identifiers because of the lack of HKT. See
                     // https://github.com/rust-lang/rust/issues/34162.
                     values.sort_by(|(id1, _), (id2, _)| id1.cmp(id2));
-                    let terms = values.into_iter().map(|(_, t)| t).collect();
+                    let terms = Array::from_iter(values.into_iter().map(|(_, t)| t));
```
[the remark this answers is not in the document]
fuzzypixelz: Oh you're completely right. Honestly I hadn't made the connection between `FromIterator` and `collect()`. But it turns out:

fuzzypixelz:
> Left a few remarks, but the approach looks good and sound to me. @fuzzypixelz, however you told me that the benchmarks weren't conclusive?

Oh the first PR comment is from a much older implementation, which was run on the faulty benchmarks. I should rewrite it.

src/environment.rs
```diff
@@ -62,6 +62,11 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
         }
     }
 
+    /// Returns the number of elements in the current layer.
+    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
fuzzypixelz: I just realized this is no longer needed for the PR. Might as well remove it ... naming things is hard.

src/eval/operation.rs
```diff
@@ -1069,15 +1070,18 @@ fn process_unary_operation(
...
                     )
                 } else {
                     //FIXME: what should we return when there's no match?
                     mk_record!(
                         ("match", Term::Str(String::new())),
                         ("index", Term::Num(-1.)),
-                        ("groups", Term::Array(Vec::new(), Default::default()))
+                        (
+                            "groups",
+                            Term::Array(std::iter::empty().collect(), Default::default())
```
yannham, line 1083: Maybe we should have an `Array::new()` constructor, as well as a `Default` impl for arrays?

src/eval/operation.rs
```diff
@@ -1850,17 +1854,16 @@ fn process_binary_operation(
                             debug_assert!(attrs1.closurized, "the left-hand side of ArrayConcat (@) is not closurized.");
                             debug_assert!(attrs2.closurized, "the right-hand side of ArrayConcat (@) is not closurized.");
 
-                            let mut ts: Vec<RichTerm> = Vec::with_capacity(ts1.len() + ts2.len());
-
-                            ts.extend(ts1.into_iter());
-                            ts.extend(ts2.into_iter());
+                            // TODO: chain tends to be slow, but Array currently can only be constructed from iterators.
+                            // it's possible to use Rc<[T]>::new_uninit_slice() (nightly-only) and fill up the array manually.
+                            let ts = ts1.into_iter().chain(ts2.into_iter()).collect();
```
yannham, line 1859: I wonder if one can get an owned slice out of a Vector (even it that implies `mem::swap` or the like). I couldn't find something with a quick research though. If it were to be true, we could use an array as a buffer and then just get back the slice to build a Nickel array.

src/program.rs
```diff
@@ -398,30 +386,21 @@ mod tests {
...
+        let expd = mk_array!(
+            ArrayAttrs::new().closurized(),
+            Term::Num(2_f64),
+            Term::Str(String::from("ab")),
+            mk_array!(
+                ArrayAttrs::new().closurized(),
+                Term::Num(1_f64),
+                mk_array!(ArrayAttrs::new().closurized(), Term::Num(3_f64))
+            )
+        );
```
yannham, line 403: Ugh, good catch, this test was strange. Must have been pasted from the parser.

src/term.rs
```diff
@@ -1485,6 +1624,18 @@ pub mod make {
         };
     }
 
+    /// Array for types implementing `Into<ArrayAttrs>` (for the attributes and `Into<RichTerm> for elements.
```
yannham, line 1627:
```suggestion
    /// Array for types implementing `Into<ArrayAttrs>` (for the attributes) and `Into<RichTerm> (for elements).
```

src/term.rs
```diff
@@ -1485,6 +1624,18 @@ pub mod make {
...
+    /// `mk_array!(ArrayAttrs::default(), Term::Num(42)) corresponds to `[42]`. Though the evaluated array may have different attributes.
+    #[macro_export]
```
yannham, line 1629: A few questions:
1. Does the `Into<ArrayAttrs>` make sense? For RichTerm it does, because we have all those different nested types (`RichTerm`, `SharedTerm`, `Term`). However I can't think of something interesting that would be `Into<ArrayAttrs>`.
2. It could be useful to have the array attrs optional, by using a different separator, like `;` at the end (a bit like the macro for row types does about the tail).

src/term.rs
```diff
@@ -1485,6 +1624,18 @@ pub mod make {
...
+    macro_rules! mk_array {
+        ( $attrs:expr, $( $terms:expr ),* ) => {
+            {
+                let ts = [$( crate::term::RichTerm::from($terms) ),*] .into_iter().collect();
```
yannham, line 1633: You can convert freely an array to a slice, without having to recollect them:

```rust
    let baz = [1, 2, 3];
    let booz : Rc<[i64]>= Rc::new(baz);
```

src/transform/free_vars.rs
```diff
@@ -144,7 +144,7 @@ fn collect_free_vars(rt: &mut RichTerm, free_vars: &mut HashSet<Ident>) {
             *deps = Some(new_deps);
         }
         Term::Array(ts, _) => {
-            for t in ts {
+            for t in ts.get_mut().unwrap().iter_mut() {
```
yannham, line 147: Same remark about panicking. You probably want `make_mut()`. By the way, this is what this very function does about the RC inside the RichTerm at the beginning: we use `SharedTerm::make_mut`, which is just `Rc::make_mut` under the hood, and not `get_mut().unwrap()`.

src/typecheck/mod.rs
```diff
@@ -527,6 +528,7 @@ fn type_check_<L: Linearizer>(
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))?;
 
             terms
+                .as_ref()
```
yannham, line 531: It's the 4th time I see `as_ref().iter()` I think. Maybe `Array` should get an `iter()` function?

yannham (changes requested):

fuzzypixelz:
@yannham I think we should reopen this PR as I no longer have access to this branch.

yannham:
Ah, I haven't thought about that. Let's do that then. Do I let you reopen the PR? And from there we can close this one.

yannham:
@fuzzypixelz , you've been granted write access to the repo, and should be able to access this branch and this PR.

src/typecheck/mod.rs
fuzzypixelz: Added!

src/term.rs
```diff
@@ -408,9 +545,10 @@ impl Term {
                 func(t1);
                 func(t2);
             }
-            OpN(_, ref mut terms) | Array(ref mut terms, _) => terms.iter_mut().for_each(|t| {
+            OpN(_, ref mut terms) => terms.iter_mut().for_each(|t| {
                 func(t);
             }),
+            Array(ref mut terms, _) => terms.get_mut().unwrap().iter_mut().for_each(|t| func(t)),
```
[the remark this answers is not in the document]
fuzzypixelz: `Rc::make_mut` doesn't typecheck when calling it on `Rc<[T]>` because "`[T]` doesn't implement `Clone`"
I was able to work around this and left a comment.

src/eval/operation.rs
fuzzypixelz: It looks like because an `Array` iterator does not implement `TrustedLen` (an experimental nightly-only API), collecting into an `Rc<[RichTerm]>` will go through a vector anyway. So might as well use a vector as a buffer and then use `Rc::from` like the standard library does.

https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-FromIterator%3CT%3E

fuzzypixelz:
Rebased onto master, should be ready to go.

src/term.rs
```diff
@@ -201,7 +202,157 @@ impl PendingContract {
...
+        }
+
+        /// Makes a mutable slice into the given `Array`.
+        pub fn make_mut(&mut self) -> &mut [RichTerm] {
+            // NOTE: trying to use `Rc::make_mut` will result in the following compiler error:
+            // > the trait `Clone` is not implemented for `[term::RichTerm]`
+            // This seems to be an edge-case in the standard library.
+            // As a workaround, if `get_mut` fails we recollect the array into a new `Rc` by cloning all terms.
+            // Thus we make sure that the second call to `get_mut` won't fail.
+            if Rc::get_mut(&mut self.inner).is_none() {
```
yannham, line 255: Wouldn't a call to something like `Rc::strong_count(&self.inner) != 1` be cheaper? Or at least more natural.

yannham: After digging, you also need to take into account the weak cout. In the stdlib, `make_mut` use the non public function `Rc::is_unique`, which is defined as:

```rust
    #[inline]
    fn is_unique(this: &Self) -> bool {
        Rc::weak_count(this) == 0 && Rc::strong_count(this) == 1
    }
```

src/term.rs
```diff
@@ -586,7 +738,8 @@ impl Term {
             }
             Term::Array(elements, _) => {
                 let elements_str: Vec<String> = elements
-                    .iter()
+                    .clone()
```
yannham, line 741: While I understand that you need to keep a Rc to the backing array while iterating, it's a bit disturbing as a code change to add a `clone()` here. Nitpick: I would encapsulate this function as an `iter()` function directly on `Array` (and I think that would gives us for loops iteration for free?)

src/transform/free_vars.rs
```diff
@@ -144,7 +144,7 @@ fn collect_free_vars(rt: &mut RichTerm, free_vars: &mut HashSet<Ident>) {
             *deps = Some(new_deps);
         }
         Term::Array(ts, _) => {
-            for t in ts {
+            for t in ts.make_mut().iter_mut() {
```
yannham, line 147: Nitpick, but same here: I would let user code inchanged (consumer of arrays), and add an `iter_mut()` wrapper inside `Array`, such that using them feel almost transparent, as if we didn't do any change. Future contributors also won't have the additional mental load of having to know those small discrepancies between using our `Array` and the standard `Vec`.

src/term.rs
yannham: In fact, after re-reading, I think we don't have a proper `iter()` replacement. Here the iterator will clone each rich term upon iteration, if I'm not mistaken. Maybe we want a proper `&RichTerm` iterator for `Array`?

src/transform/free_vars.rs
yannham: Actually, I take that back: the difference is now that `iter_mut()` may clone the entire backing array, which is not the case with a usual vector. So it may actually be a good thing that the caller is aware of that difference.

yannham:
I'm taking over this PR, as there is not much to do to bring it down the finish line.

yannham (approved):

Issue #671, opened 2022-03-23
The Syntax Bikeshedding Dojo, round 6: enum types


yannham:
Prior the implementation of RFC002 (see #595), Nickel used the following syntax for enum types: `<foo, bar, baz>`. Once we merged the term and type syntax, this became ambiguous with the comparison operators. We are currently temporarily using an undocumented syntax, just for the stdlib, but users can't rely on a stable way to write enum types, which is too bad.

## What other languages do

I am not aware of that many languages with structural enum types (in Nickel, we don't have to give a name to our enums prior to using them: it's all based on the structure of the enum type, i.e. what are the alternative it permits). C/C++/Rust's enums are nominal, algebraic data types (OCaml, Haskell, Scala, etc.) are nominal as well.

### OCaml's polymorphic variants

Ocaml uses the `[`/`]` delimiters and `|` as a separator for its structural enum types (polymorphic variants):

```ocaml
type 'a vlist = [`Nil | `Cons of 'a * 'a vlist]
```

At first sight, this wouldn't play very well in Nickel (including variants):
- ``[`foo | `bar]`` is already valid syntax: it's a one element list with tag `` `foo `` and attached contract `` `bar ``
-  ``[`foo, `bar]`` is a two-element list
- ``[`foo ; `bar]`` would clash with the row tail syntax `forall a. <foo, bar | a>`.

There is a twist though, that could make using ``[`foo, `bar]`` possible and consistent. See the proposals below.

### Unions: TypeScript, CUE

Typescript (and CUE, for that matter) doesn't have specific enum types, but can emulate them with union types and strings. That is, the old `<foo, bar, baz>` in Nickel would be written `"foo" | "bar" | "baz"`. This doesn't provide much inspiration, as:

1. We don't have union types, for a [good reason](https://dl.acm.org/doi/10.1145/3486602.3486767)
2. `|` and its variations are already taken in Nickel (contracts application, metadata, reverse application)

### Dhall's tagged unions

Dhall uses the following syntax for a flavor of union/enum types, which is close to the old Nickel one:

```dhall
< Number : Natural | Boolean : Bool >
```

It would have the same ambiguity issue.

That's the ones I know from the top of my head, feel free to add other examples you know.

## Proposals

The following is an non exhaustive list of random ideas, don't hesitate to add your own suggestions.

### The `[| foo, bar |]` syntax

[...]

PR #780, opened 2022-07-28
Restore syntax for enum types

yannham:
The enum types synax was temporary and somehow hidden. For version `0.2.0`, we introduce them again. This PR also changes the syntax such that alternatives must be prefixed with a backtick `, as for literal enum tags, for visual consistency. See #671 for more context.

This PR also updates the manual accordingly.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -214,8 +214,9 @@ While it's technically possible to just use strings in place of enum tags, using
 an enum tag insists on the fact that only a finite number of alternatives can be
 used for the corresponding value.
 
-Enum will also be handled more finely by the typechecker, however, as of the
-version 0.1 of Nickel, enum types are not yet supported.
+Additionally, the typechecker is aware of enums and can for example statically
+enforce that only valid tags are passed to a function within a typed block. See
+[the manual section on typing](./typing.md) for more details.
 
 ## Equality
 
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -1,22 +1,5 @@
 {
   array = {
-    ComparisonResult
-      | doc m%"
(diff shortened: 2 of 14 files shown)


doc/manual/syntax.md
```diff
@@ -214,8 +214,9 @@ While it's technically possible to just use strings in place of enum tags, using
 an enum tag insists on the fact that only a finite number of alternatives can be
 used for the corresponding value.
 
-Enum will also be handled more finely by the typechecker, however, as of the
-version 0.1 of Nickel, enum types are not yet supported.
+Additionnally, the typechecker is aware of enums and can for example statically
```
fuzzypixelz, line 217:
```suggestion
Additionally, the typechecker is aware of enums and can for example statically
```

stdlib/array.ncl
```diff
@@ -271,7 +254,7 @@
         "%m
       = fun f n => %generate% n f,
 
-    sort | forall a. (a -> a -> ComparisonResult) -> Array a -> Array a
+    sort | forall a. (a -> a -> [| `Lesser, `Equal, `Greater |]) -> Array a -> Array a
```
fuzzypixelz, line 257: Would it be desirable to make this accessible to lib consumers? If someone wants to define a sorting function and statically type it, wouldn't they need to spell out the enum type?

But I suppose this would require some kind of static type aliases.

fuzzypixelz (approved):
LGTM!

stdlib/array.ncl
yannham: Yes, I initially kept an alias for this, but then the typechecker doesn't understand as an enum type statically. If we have type alias at some point (and actually contract equality may end up doing that with a slight enhancement) this may be a better API. For now, I fear we have no choice, if we want static typing.

PR #784, opened 2022-08-01
Intern (mostly static) strings using `Rc<str>`

fuzzypixelz:
This is still very much a work in progress.


yannham:
@fuzzypixelz I don't remember what was the conclusion about this experience. Is it just incomplete? Or did show no improvement? What action should we take?

yannham:
Without further information, let's close this experiment for now. We can probably consider it being superseded by #835.

PR #785, opened 2022-08-01
 Improve error messages for some record contract violations

### LINKED ISSUE #517: [Dogfooding] Improve error messages for some record contract violations
When applying a contract written as a record to a value, the implementation is mostly to merge the contract with the value (although with a few minor differences). If the value is not a record, then the error obtained is then `Non-mergeable terms`, which is slightly unintuitive.

That is, non-record contracts violation and record contract violation give different error messages:

```
let Contract = contracts.fromPred (fun x => builtins.isNum x && x % 2 == 1) in
null | #Contract
error: Blame error: contract broken by a value.
  ┌─ :1:1
  │
1 │ #Contract
  │ --------- expected type
  │
  ┌─ repl-input-3:2:1
  │
2 │ null | #Contract
  │ ---- evaluated to this expression

note: 
  ┌─ repl-input-3:2:8
  │
2 │ null | #Contract
  │        ^^^^^^^^^ bound here

nickel> let Contract = {foo | Num} in
null | #Contract
error: Non mergeable terms
  ┌─ repl-input-2:1:16
  │
1 │ let Contract = {foo | Num} in
  │                ^^^^^^^^^^^ with this expression
2 │ null | #Contract
  │ ^^^^ cannot merge this expression
``` 

That record contract violation gives a standard blame error message when the operand is not a record.

ebresafegaga:
Closes #517

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -91,6 +91,7 @@ pub fn merge(
     mut env2: Environment,
     pos_op: TermPos,
     mode: MergeMode,
+    call_stack: &CallStack,
 ) -> Result<Closure, EvalError> {
     // Merging a simple value and a metavalue is equivalent to first wrapping the simple value in a
     // new metavalue (with no attribute set excepted the value), and then merging the two
@@ -425,6 +426,24 @@ pub fn merge(
                 ),
                 env,
             })
+        }
+        // We want to merge a non-record term with a record contract
+        (t1_, t2_ @ Term::Record(..)) => {
+            if let MergeMode::Contract(label) = mode {
+                Err(EvalError::BlameError(label, call_stack.clone()))
+            } else {
+                Err(EvalError::MergeIncompatibleArgs(
+                    RichTerm {
+                        term: SharedTerm::new(t1_),
+                        pos: pos1,
+                    },
+                    RichTerm {
+                        term: SharedTerm::new(t2_),
+                        pos: pos2,
+                    },
+                    pos_op,
+                ))
+            }
         }
         //The following cases are either errors or not yet implemented
         (t1_, t2_) => Err(EvalError::MergeIncompatibleArgs(
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -427,36 +427,25 @@ pub fn merge(
                 env,
             })
         }
-        // We want to merge a non-record term with a record contract
-        (t1_, t2_ @ Term::Record(..)) => {
-            if let MergeMode::Contract(label) = mode {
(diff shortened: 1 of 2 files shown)


src/eval/merge.rs
```diff
@@ -426,6 +427,24 @@ pub fn merge(
...
+                        pos: pos1,
+                    },
+                    RichTerm {
+                        term: SharedTerm::new(t2_),
+                        pos: pos2,
+                    },
+                    pos_op,
+                ))
+            }
+        }
```
yannham, line 447: I think you can reuse the case below with an `if` guard, instead of duplicating the code creating the `MergeIncompatibleArgs`. There are several solutions to nest the matching, but here is an example:

```suggestion
        (t1_, t2_ @ Term::Record(..)) if matches!(mode, MergeMode::Contract(..)) => {
            if let MergeMode::Contract(label) = mode {
                Err(EvalError::BlameError(label, call_stack.clone()))
            } else {
                unreachable!()
            }
        }
```

Another solution is to keep the generic match `(t1_, t2_)` and then match again, which saves the redundant match on `mode` and the `unreachable!()` at the price of a nested pattern matching:

```rust
(t1_, t2_) => {
    match ((mode, t2_)) {
        (MergeMode::Contract(label), Term::Record(..)) => Err(EvalError::BlameError(label, call_stack.clone())),
        _ => {
            Err(EvalError::MergeIncompatibleArgs(
                RichTerm {
                    term: SharedTerm::new(t1_),
                    pos: pos1,
                },
                RichTerm {
                    term: SharedTerm::new(t2_),
                    pos: pos2,
                },
                pos_op,
            ))
        }
    }
}
```

Yet another solution is to make a local function `mk_non_mergeable_error` or the like.

yannham (commented):
The logic looks good to me. I left a comment about code duplication.

src/eval/merge.rs
ebresafegaga: I don't like the `unreachable!()`, I'll go with the later.

src/eval/merge.rs
```diff
@@ -426,18 +427,25 @@ pub fn merge(
...
-                pos: pos2,
-            },
-            pos_op,
-        )),
+        (t1_, t2_) => match (mode, &t2_) {
+            // We want to merge a non-record term with a record contract
+            (MergeMode::Contract(label), Term::Record(..)) => {
+                Err(EvalError::BlameError(label, call_stack.clone()))
+            }
+
```
yannham, line 435:
```suggestion
```

yannham (approved):

PR #789, opened 2022-08-03
Fix fold benchmarks

fuzzypixelz:
It seems like the fold benchmarks weren't running at all, in that the arguments to `array.fold` are not saturated, the signature of `array.generate` is wrong, and `letter`'s contract is incorrect.

I got slowdowns of nearly 10000% after the changes.

Or am I missing something?

### DIFF
--- a/benches/arrays/fold.ncl
+++ b/benches/arrays/fold.ncl
@@ -1,34 +1,34 @@
-let letter | string.CharLiteral -> string.CharLiteral = fun n =>
-  string.char_code "a" + (n % 26)
-  |> string.char_from_code in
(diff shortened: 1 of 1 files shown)


yannham:
Ah, that's bad. We should maybe have reference values to test the benchmarks suite against, although we don't want to include that in the benchmark. Maybe we can just reuse the `.ncl` files from tests, quite simply. But that wouldn't protect new benchmarks, though. Another solution would be to have, in addition to the `run` function, an `expected` function returning constants that would be compared against the function. Although the `n` parameter makes this slightly more complex.

yannham (approved):

PR #790, opened 2022-08-03
Initial implementation of Deserializer to export Nickel terms to Rust.

giacomocariello:
Ad discussed in Discussed in https://github.com/tweag/nickel/discussions/786, I've approached a first implementation of a Deserializer to convert Nickel terms into Rust plain types.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -348,6 +348,17 @@ pub enum SerializationError {
     Other(String),
 }
 
+/// An error occurred during deserialization to Rust.
+#[derive(Debug, PartialEq, Clone)]
+pub enum RustDeserializationError {
+    InvalidType { expected: String, occurred: String },
+    MissingValue,
+    EmptyMetaValue,
+    UnimplementedType { occurred: String },
+    InvalidRecordLength(usize),
+    Other(String),
+}
+
 /// A general I/O error, occurring when reading a source file or writing an export.
 #[derive(Debug, PartialEq, Clone)]
 pub struct IOError(pub String);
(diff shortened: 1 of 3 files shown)


yannham:
Thanks a lot for tackling this! I haven't had the time to look at it yet in detail, but maybe a naive question: do you really need to implement a deserializer by hand? The structure `term::Term` already implement `Serialize` (which is used to write it down to JSON for example) and `Deserialize` (which is used to import directly JSON as if it was Nickel code). A `RichTerm` is just a thin layer over a `Term` (mostly, an `Rc` and a position), so you can derive a (de)serializer for the former quite easily, I suspect. Or am I missing something?

yannham:
I see. It felt like this was re-implementing something that was already possible, but I see now that it would require to use an intermediate data format indeed, which is not great. We'll have a look at it soon.

yannham (commented):
Look good in general! I just left a few comments. Also, I'm wondering about the rationale behind the `#inline` annotations: some are on non-trivial functions (`deserialize_enum`), some on trivial functions (`deserialize_newtype_struct`), but then a lot of other deserialization functions don't get an annotation.

Is there a strong motivation to deviate from the default behavior for inlining? Is the goal to enable cross-crate inlining?

giacomocariello:
> I'm wondering about the rationale behind the `#inline` annotations: some are on non-trivial functions (`deserialize_enum`), some on trivial functions (`deserialize_newtype_struct`), but then a lot of other deserialization functions don't get an annotation.
> 
> Is there a strong motivation to deviate from the default behavior for inlining? Is the goal to enable cross-crate inlining?

Initially I took serde_json::Value's Deserializer as reference for drafting the implementation and I noticed the inlines seemed not consistent. Parsing through their repo's history and issues, it seems they acknowledged there was no relevant performance gain through inlining for cross-crate usage anyways. I eventually forgot to remove them from my code, but I'll do that now.

src/error.rs
```diff
@@ -1583,3 +1595,37 @@ impl ToDiagnostic<FileId> for ReplError {
         }
     }
 }
+
+impl std::fmt::Display for RustDeserializationError {
```
yannham, line 1599: I wonder if this error shouldn't stay in `deserializer`. There is no strong rationale (unfortunately) with respect to what error should go in `error` and what can stay in the module itself, but I would say that `error.rs` is concerned with top-level errors that the user may face directly (evaluation, IO, typechecking, etc.). For example I think they all implement the `ToDiagnostic` trait to print them nicely using codespan.

This is not the case of `RustDeserializationError`, which seems rather to be directed toward crates consuming the deserializer implementation.

yannham (approved):
I think the error can stay in the `deserializer` module itself. Otherwise it's good to go.

giacomocariello:
> Hey @giacomocariello, it's good to merge but there seems to be a compilation error on the CI probably caused by the shuffling of `RustDeserializationError` code

Whoops! I forgot to edit tests module accordingly. Now it should be ok!

jmgilman:
Hi @giacomocariello, do you have an example of using this in the wild? I'm interested in working with Nickel configs from within an existing Rust application.

PR #791, opened 2022-08-04
[Fix] LSP freezing on some commands

yannham:
We noticed with @fuzzypixelz than some hover were taking ages with the LSP. It turns out debug print statement were left out. This PR fixes this problem. This is crippling enough that it mandates a new minor release.


fuzzypixelz (approved):
LGTM. I think there are no more print calls.

PR #792, opened 2022-08-04
Enable markdown linting for doc

yannham:
What started as a non-trivial enterprise, adding markdown linting checks to the CI, was actually very easy to do thanks to pre-commit-hooks.nix. This PR:

- makes a pass on the various markdown documentation to appease the linter
- adds a configuration file for markdownlint to customize a couple rules
- enables markdown linting in pre-commit-hooks.nix, which has the effect of running the linter as part of the checks of the flake (and hence, on the CI).

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,10 @@
+{
+    "default": true,
+    "MD013": {
+        "code_blocks": false,
+        "tables": false
+    },
+    "MD024": {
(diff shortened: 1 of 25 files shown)


thufschmitt (approved):
I didn't thoroughly checked that the markdown changes were really semantically meaningless, but at least there's no big bad thing that I can see.

:+1: on the Nix side

PR #793, opened 2022-08-05
[CI] Add clippy check

yannham:
Add a clippy check to the CI (and flake checks, and pre commit hooks). The CI disables two rules that, in our experience, don't always make sense to implement. This PR doesn't disable them on the whole crate because sometimes they are justified, so we don't want to silence them either when a contributor is running clippy locally: we just avoid having them blocking the CI.

* Add Clippy to the CI run
* Add a Clippy pre-commit hook
* CI logs are fully visible in Github Action
* All Clippy errors and warnings are fixed (most automatically via `clippy --fix`, some manually)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -50,4 +50,4 @@ jobs:
 
     - name: Run all checks
       run: |
-        nix flake check
+        nix flake check --print-build-logs
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ rustyline-derive = { version = "0.6.0", optional = true }
 
 # The `wasm-bindgen` version is pinned (`=`) because it must be the same as `wasm-bindgen-cli` in `flake.nix`
 wasm-bindgen = { version = "=0.2.83", optional = true, features = ["serde-serialize"] }
+serde-wasm-bindgen = "0.4.5"
 js-sys = { version = "0.3", optional = true }
 serde_repr = { version = "0.1", optional = true }
 pretty = "0.11.3"
--- a/flake.nix
+++ b/flake.nix
@@ -138,92 +138,82 @@
           '';
         };
 
-      buildNickel =
-        { channel ? "stable"
-        , isDevShell ? false
-        , target ? pkgs.rust.toRustTarget pkgs.stdenv.hostPlatform
-        ,
-        }:
-        let
-          rustProfile =
-            if isDevShell then "default"
-            else "minimal";
-
-          rust = mkRust { inherit rustProfile channel target; };
-
-          pre-commit-builder = { isHook }: pre-commit-hooks.lib.${system}.run {
-            src = self;
-            hooks = {
-              nixpkgs-fmt = {
-                enable = true;
-                excludes = [
-                  "lsp/client-extension/default.nix"
-                  "lsp/client-extension/node-env.nix"
-                  "lsp/client-extension/node-packages.nix"
-                ];
-              };
-
-              rustfmt = {
-                enable = true;
-                entry = pkgs.lib.mkForce "${rust}/bin/cargo-fmt fmt -- --check --color always";
-              };
-
-              markdownlint = {
-                enable = true;
-                excludes = [
-                  "notes/(.+)\\.md$"
-                  "^RELEASES\\.md$"
-                ];
-              };
-
-              custom-clippy =
-                let
-                  allow = [
-                    "clippy::new-without-default"
-                    "clippy::match_like_matches_macro"
-                  ];
-
-                  allowOpts = builtins.concatStringsSep " "
-                    (builtins.map (lint: "--allow \"${lint}\"") allow);
-
-                  # Clippy requires access to the crate registry because many lints require dependency information.
-                  # Similarly to `buildNickel`, we support 2 modes:
-                  # - when run as a shell hook, we want to be fast, so we let Cargo do incremental compilation, and download dependencies as needed.
-                  #   This is mainly used by developers, who care about speed at the expense of hermeticity.
-                  # - when run as a flake check, we want to be fully reproducible, so we Nixify Cargo dependency resolution.
-                  #   This is mainly used by CI, where the extra time is not such a big deal.
-                  offlineOpts = if isHook then "" else "--frozen --offline";
-
-                  # Since `pre-commit-hooks` does not execute the entries (commands) as part of a Nix derivation,
-                  # we have to do some extra setup work.
-                  # - `stdenv` provides standard environment similarly to `mkDerivation`, including a C compiler (`cc`) required by Rust.
-                  # - `cargoHome` provides Rust with the right `CARGO_HOME` environment variable to discover built dependencies.
-                  cargoHomeHookSetup = if isHook then "" else ''
-                    source ${pkgs.stdenv}/setup
-                    source ${cargoHome}/nix-support/setup-hook
-                  '';
-                in
-                {
-                  enable = true;
-                  files = "\\.rs$";
-                  pass_filenames = false;
-                  entry =
-                    "${pkgs.writeShellScript "clippy-hook" ''
-                    export PATH="${pkgs.lib.makeLibraryPath missingSysPkgs}:$PATH"
-                    ${cargoHomeHookSetup}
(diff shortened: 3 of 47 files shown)


flake.nix
```diff
@@ -175,6 +175,21 @@
...
+                    "clippy::new-without-default"
+                    "clippy::match_like_matches_macro"
+                  ];
+                  allowOpts = builtins.concatStringsSep " "
+                    (builtins.map (lint: "-A \"${lint}\"") allow);
+                in
+                {
+                  enable = true;
+                  entry = pkgs.lib.mkForce ''
+                    ${rust}/bin/cargo-clippy clippy --frozen --offline -- -D warnings ${allowOpts}
```
fuzzypixelz, line 190: It seems cargo doesn't like `--offline` in clippy, it cannot find `ansi_term` (presumably to recursively run clippy on it?). The same thing happens locally.
I'm not familiar with how the flake works, but it seems like passing `--offline` here is superfluous.
I suggest disabling the run on dependencies. Then it _shouldn't_ need to access to the network.

Nix sure is something.
```suggestion
                    ${rust}/bin/cargo-clippy clippy --frozen --no-deps -- -D warnings ${allowOpts}
```

src/eval/callstack.rs
```diff
@@ -219,6 +219,10 @@ impl CallStack {
         self.0.len()
     }
```
fuzzypixelz, line 221:
```suggestion

/// Returns `true` if the callstack contains no elements.
```

Small nitpick.

src/identifier.rs
```diff
@@ -59,6 +59,15 @@ where
...
 }
 
+// We can't satisfy clippy and implement `From<Ident> for String`. Otherwise, the generic
+// implementation above will give a second way of deriving `From<Ident> for Ident`:
+//
+// - the identity, provided by core (impl From<T> for T)
+// - `String::from::<Ident>` -> `Ident::from::<String>`, given by the implementation above.
+//
+// And the compiler is unhappy (and the second implementation would silently erase the position).
+// Hence, we disable the clippy lint, because being able to write `Ident::from("foo")` is nice.
```
fuzzypixelz, line 69: I love this sentence.

fuzzypixelz (dismissed):
I have a suggestion to fix the flake check, it's currently failing with a weird error.

Other than that, LGTM.

sir4ur0n:
@yannham what is the status of this PR? Should it be closed? Taken over by someone?

yannham:
I'd like the assistance of a Nix Guru, but hasn't had the time to do that. It does need some action to be taken.

aspiwack:
> I'd like the assistance of a Nix Guru, but hasn't had the time to do that. It does need some action to be taken.

Make a project opening issue, maybe?

sir4ur0n:
I am starting to take a look at this PR

.github/workflows/continuous-integration.yml
```diff
@@ -50,4 +50,4 @@ jobs:
 
     - name: Run all checks
       run: |
-        nix flake check
+        nix flake check --print-build-logs
```
sir4ur0n, line 53: Nice little change that makes all errors readable in Github Action logs :tada:

Cargo.toml
```diff
@@ -54,6 +54,7 @@ rustyline-derive = { version = "0.6.0", optional = true }
 
 # The `wasm-bindgen` version is pinned (`=`) because it must be the same as `wasm-bindgen-cli` in `flake.nix`
 wasm-bindgen = { version = "=0.2.83", optional = true, features = ["serde-serialize"] }
+serde-wasm-bindgen = "0.4.5"
```
sir4ur0n, line 57: This new dependency was necessary to replace some deprecated code, see e.g. `src/repl/wasm_frontend.rs`

src/eval/callstack.rs
```diff
@@ -219,6 +219,11 @@ impl CallStack {
...
 
+    /// Return whether the callstack is empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
```
sir4ur0n, line 225: Clippy was complaining that this module has a `len` method without a `is_empty` method. @yannham said in a future PR we probably should change the visibility of many/all methods of this module to `pub(super)` but this is rather unrelated to this PR, so in the short term we add this function.

src/eval/merge.rs
```diff
@@ -527,13 +535,16 @@ fn rev_thunks<'a, I: Iterator<Item = &'a mut RichTerm>>(map: I, env: &mut Enviro
...
+    }
+
     /// Split two hashmaps m1 and m2 in three parts (left,center,right), where left holds bindings
     /// `(key,value)` where key is not in `m2.keys()`, right is the dual (keys of m2 that are not
     /// in m1), and center holds bindings for keys that are both in m1 and m2.
-    pub fn split<K, V1, V2>(
-        m1: HashMap<K, V1>,
-        m2: HashMap<K, V2>,
-    ) -> (HashMap<K, V1>, HashMap<K, (V1, V2)>, HashMap<K, V2>)
+    pub fn split<K, V1, V2>(m1: HashMap<K, V1>, m2: HashMap<K, V2>) -> SplitResult<K, V1, V2>
```
sir4ur0n, line 547: Clippy was complaining that the return type was too verbose, and both @yannham and I agreed with Clippy here, so we extracted the return type in a `struct`, which makes the code clearer in our opinion

vkleen (approved):
Nice work! 👍 Running clippy as a pre-commit hook takes a little time but not enough to really annoy me, I think. In any case, I think it's a sane default and not too hard to temporarily disable locally if it gets annoying at some point.

For the record, if you're working on a lot of commits and want to disable the pre-commit hooks (if they're enabled in your development shell), just run `pre-commit uninstall`.

sir4ur0n:
@vkleen an alternative is to pass `--no-verify`/`-n` option to Git, or `-n` to Magit

sir4ur0n:
We have made another round of refactoring with @balsoft and I think the `flake.nix` looks better now, but apparently the Darwin `nix flake check` fails:
```
cp: cannot stat '/private/tmp/nix-build-nickel-0.2.1_20221115_df7b5da.drv-1/vendor': Permission denied
```

I have no idea where to even begin investigating :shrug: 
I have requested access to Tweag's Apple M1 builder, I'm waiting for the infra team to grant me access, then I'll try to reproduce remotely... But I nonetheless don't know what to look at :sweat_smile:

sir4ur0n:
CI is now green for both Linux and Darwin, so it's ready for the final round of reviews :tada:

flake.nix
```diff
@@ -183,22 +187,26 @@
...
-                  echo $CARGO_HOME
-                  echo "ls12"
-                  ls -laa
+                  NIX_BUILD_TOP=$nix_build_top
+                  TEMP=$nix_build_top
+                  TEMPDIR=$nix_build_top
+                  TMP=$nix_build_top
+                  TMPDIR=$nix_build_top
                   source $stdenv/setup
                 '';
```
vkleen, line 211: That's some dark magic 🪄 I like it 👍

vkleen (approved):
The new changes to please clippy seem positive and safe to me. I think, we should merge this (since @sir4ur0n managed to fix CI 🎉) before another 150 changes become necessary.

yannham (commented):
LGTM! (I can't approve because I'm technically the original author of the PR). The new flake looks  better and is well documented :+1:

src/typecheck/mod.rs
```diff
@@ -1996,7 +1993,7 @@ pub fn unify_erows(
                 }
                 (EnumRowsF::Extend { row: id, tail }, erows2 @ EnumRowsF::Extend { .. }) => {
                     let t2_tail = erows_add(state, &id, UnifEnumRows::Concrete(erows2))?;
-                    unify_erows(state, ctxt, *tail, t2_tail)
+                    unify_erows(state, *tail, t2_tail)
```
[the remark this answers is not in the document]
sir4ur0n: The `&Context` argument was removed because it was useless: https://github.com/tweag/nickel/pull/793/files#diff-138af75d9d5a9700f6695900b1a0a4454d2acbfd263999a573553beae6ca2096L1980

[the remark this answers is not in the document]
yannham: Ah, so your first clippy pass removed context, and then a snippet coming from the record rows refactoring still had it. I see, thanks!

PR #794, opened 2022-08-05
Migrate to the latest edition (2021)

yannham:
`cargo fix --edition` didn't find anything to do. Let's just see if the CI is happy (I haven't tried it locally, given how long it is :confused:  )


fuzzypixelz (approved):
CI is happy!

I guess the code is _that good_.

PR #796, opened 2022-08-05
Fix sort benchmark and stdlib function

fuzzypixelz:
The sort benchmark was crashing because of a contraction violation. It still used ``` `Less``` instead of ``` `Lesser``` after it was renamed.  Though `array.sort`'s implementation itself still used ``` `Less```.

I took the opportunity to make the sort benchmark depend on the input, using some questionable random number generation methods.

I also added a test for `array.sort` so this kind of thing doesn't happen in the future.


benches/arrays/sort.ncl
```diff
@@ -1,5 +1,27 @@
...
+    [next] @ array_random next (n - 1)
+in
+
+let ascending = fun x y => 
+  if x < y then `Lesser 
+  else if x == y then `Equal 
+  else `Greater 
+in
+
+{ run = fun n => array.sort ascending (array_random 1337 n) }
```
yannham, line 27: The code is rather simple, a good surprise. But the issue is now that we don't benchmark sorting, we benchmark random array generation + sorting. One possibility is to do that on the rust side, and then pass it to the `run` function. Or make a custom function to run this benchmark which first evaluate `array_random 1337 n`, before starting the measurement, and then pass it to run?

But now that I write it, the latter seems to imply the same pain of writing a custom benchmark function, and is just more complicated. Maybe the simplest is to generate the array on the Rust side and pass it to the run function instead of `n`.

yannham (changes requested):

benches/arrays/sort.ncl
fuzzypixelz: That's fair. Maybe I could also split the number generation stuff into its own benchmark?

yannham: Yes, since the code is written, let's add it

fuzzypixelz: Should be all good now. 

Though I used mostly the same code on the Rust side to avoid adding `rand` to the dependencies. We probably already have it indirectly but this is such a minor requirement.

benches/arrays.rs
```diff
@@ -2,6 +2,24 @@ use criterion::{criterion_main, Criterion};
...
+    let mut numbers = Vec::with_capacity(len);
+    numbers.push(1337);
+
+    for _ in 0..len {
+        let x = *numbers.last().unwrap();
+        numbers.push((a * x + c) % m);
+    }
+
+    // HACK: It so happens that this is valid Nickel syntax.
+    format!("{:?}", numbers)
```
yannham, line 20: May it be more robust to instantiate an `Term::Array` and use the ast pretty-printing facilities, so that if the syntax change, this won't break? Honestly I don't think the array syntax will ever become invalid... but, if it's only a few line...

yannham (approved):

PR #801, opened 2022-08-09
Refactoring of global typing environment

yannham:
Get rid of the useless separation between the global typing environment and the current typing environment. It was done for performance reasons, when cloning environments was expensive. But now, environments support sharing and cheap clone: instead of having two separate environments, we can just start from the initial (previously "global") and insert directly into it. Doing so, we get rid of the `typecheck::Envs` as well as simplify a bit the API of `typecheck`.

This PR also renames "global" to "initial", which is now more faithful.

Note that one could be tempted to the same to the eval environment: alas, the situation is more complex. We are creating new environments quite a lot (through closurization) during evaluation. Those environment are totally empty: if we would get rid of the current separate initial environment, we would need to create new environments by cloning this initial environment, instead of currently creating entirely empty ones, which is cumbersome (`initial_env` would have to be threaded through all evaluation sub functions). Thus, for evaluation, it's simpler to keep the initial environment separate.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -846,7 +846,7 @@ impl Cache {
     }
 
     /// Load, parse, and apply program transformations to the standard library. Do not typecheck
-    /// for performance reason: this is done in the test suite. Return an initial environment
+    /// for performance reasons: this is done in the test suite. Return an initial environment
     /// containing both the evaluation and type environments. If you only need the type environment, use
     /// `load_stdlib` then `mk_type_env` to avoid transformations and evaluation preparation.
     pub fn prepare_stdlib(&mut self) -> Result<Envs, Error> {
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -847,7 +847,7 @@ impl Cache {
 
     /// Load, parse, and apply program transformations to the standard library. Do not typecheck
     /// for performance reason: this is done in the test suite. Return an initial environment
-    /// containing both eval and type environment. If you need only the type environment, use
+    /// containing both the evaluation and type environments. If you only need the type environment, use
     /// `load_stdlib` then `mk_type_env` to avoid transformations and evaluation preparation.
     pub fn prepare_stdlib(&mut self) -> Result<Envs, Error> {
         #[cfg(debug_assertions)]
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -6,8 +6,8 @@ use crate::position::TermPos;
 use crate::stdlib as nickel_stdlib;
 use crate::term::{RichTerm, SharedTerm, Term};
 use crate::transform::import_resolution;
+use crate::typecheck::type_check;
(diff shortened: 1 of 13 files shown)


src/cache.rs
```diff
@@ -847,17 +846,16 @@ impl Cache {
     }
 
     /// Load, parse, and apply program transformations to the standard library. Do not typecheck
-    /// for performance reason: this is done in the test suite.
-    /// Return a global environment containing both eval and type environment. If you need only the
-    /// type environment, use `load_stdlib()` then `mk_global_type` to avoid
-    /// transformations and evaluation preparation.
-    pub fn prepare_stdlib(&mut self) -> Result<GlobalEnv, Error> {
+    /// for performance reason: this is done in the test suite. Return an initial environment
```
fuzzypixelz, line 849:
```suggestion
    /// for performance reasons: this is done in the test suite. Return an initial environment
```

src/cache.rs
```diff
@@ -847,17 +846,16 @@ impl Cache {
...
+    /// containing both eval and type environment. If you need only the type environment, use
```
fuzzypixelz, line 850:
```suggestion
    /// containing both the evaluation and type environments. If you only need the type environment, use
```

src/eval/mod.rs
```diff
@@ -275,8 +280,8 @@ where
 /// # Arguments
 ///
 /// - `clos`: the closure to evaluate
-/// - `global_env`: the global environment containing the builtin functions of the language. Accessible from anywhere in the
-/// program.
+/// - `initial_env`: the initial environment containing the builtin functions of the language.
```
fuzzypixelz, line 283: The global envrironment doesn't only contain functions. I can't think of a better word here. Small nitpick.

src/typecheck/mod.rs
```diff
@@ -191,16 +149,30 @@ pub struct State<'a> {
 
 /// Typecheck a term.
 ///
-/// Return the inferred type in case of success. This is just a wrapper that calls `type_check_`
-/// with a fresh unification variable as goal.
+/// Return the inferred type in case of success. This is just a wrapper that calls
+/// `type_check_linearize` with an blanket implementation for the linearizer.
```
fuzzypixelz, line 153:
```suggestion
/// `type_check_linearize` with a blanket implementation for the linearizer.
```

fuzzypixelz (approved):
Seems like a straightforward enough change.

src/eval/mod.rs
yannham, line 283:
```suggestion
/// - `initial_env`: the initial environment containing the builtins of the language.
```

francois-caddet (commented):
Looks great. The only question is, in eval, there are many init_env as functions params. Are they not simply envs? Are they only the stdlib env or the full environment?

PR #802, opened 2022-08-15
Fix polymorphic contracts that change semantics

### LINKED ISSUE #727: Polymorphic contracts don't fail but silently change the semantics
**Describe the bug**
Polymorphic functions in statically typed languages (and in the statically typed fragment of Nickel) enjoy a property called [parametricity](https://en.wikipedia.org/wiki/Parametricity). When derived from a static type, a contract tries to dynamically enforce the invariants that would be guaranteed by their static counterpart. Thus, contracts for polymorphic functions enforce parametricity using a technique called dynamic sealing (see for example the [Blame for All](https://dl.acm.org/doi/10.1145/1926385.1926409) paper).

Concretely, polymorphic arguments are wrapped or sealed in a special node (`Wrapped`) when they enter a function, which prevents any inspection. They are unwrapped when exiting the function (either passing the argument to another function of the right type, or when returning a value).

The issue is that in practice, trying to use a wrapped value has surprising and different behaviors. For example, functions operating on `Dyn` values such as `builtin.is_xxx` or `==` won't raise any error, but just return unexpected results. On the other hand, operators that require a specific type for their argument will fail, but often with an unhelpful error message. This is all made worse by the fact that this parametricity issue is not trivial for non FP folks (and even for FP folks!) and uncommon in other languages, so it deserves a particularly good error message.

**To Reproduce**
Trying to use equality on wrapped values just return false without failing:

```
nickel> let peq | forall a. a -> a -> Bool = (==) in peq 0 0
false

nickel>
```

Same for `builltin.is_xxx`, which returns false without failing:

```
nickel> let p_is_num | forall a. a -> Bool = builtin.is_num in p_is_num 0
false

nickel>
```

On the other hand, using a numeric operator gives a (strange) error:

```nickel
nickel>let p_plus | forall a. a -> Num -> Num = (+) in p_plus 0 1
error: type error
  │
1 │ <wrapped>
  │ ^^^^^^^^^ this expression has type Wrapped, but Num was expected
  │
  = +, 1st argument
```

**Expected behavior**
Trying to use (i.e. to evaluate) a wrapped value for anything else than wrapping/unwrapping should raise a proper contract error, referring to the original contract and with proper advice on why the issue is happening.

**Environment**
 - Version of the code: f7e0ced275a71907bfd9b19072bffb92a3c8c376

**Additional context**
This probably implies to store a reference to the original label inside `Wrapped`, and when reaching a wrapped value, checking if the next continuation is a `Wrap/Unwrap` operator to decide if we must stop the execution with a contract error.

ebresafegaga:
This is a WIP to fix #727. Feedback is needed to continue the implementation.


src/eval/stack.rs
```diff
@@ -230,6 +230,20 @@ impl Stack {
         }
     }
 
+    pub fn peek_op_cont(&mut self) -> Option<&OperationCont> {
```
yannham, line 233: Why does a peek operation need a mutable reference?

src/eval/stack.rs
```diff
@@ -230,6 +230,20 @@ impl Stack {
...
+        // I think this would fail if the stack is empty
+        // When else can this fail?
+        let mut last_idx = self.0.len() - 1;
+        while let Marker::Thunk(..) = self.0[last_idx] {
+            last_idx -= 1;
+        }
+        match &self.0[last_idx] {
+            Marker::Cont(cont, _, _) => Some(cont),
+            _ => None,
+        }
```
yannham, line 244: Suggestion using a more functional, iterator-y style:

```suggestion
        let it = self.0.iter().rev().skip_while(|marker| matches!(marker, Marker::Thunk(..));
        match it.next() {
            Marker::Cont(cont, _, _) => Some(cont),
            _ => None,
        }
```

src/term.rs
```diff
@@ -144,7 +144,7 @@ pub enum Term {
     /// type variable. In our example, the last cast to `a` finds `Sealed(2, "a")`, while it
     /// expected `Sealed(1, _)`, hence it raises a positive blame.
     #[serde(skip)]
-    Sealed(SealingKey, RichTerm),
+    Sealed(SealingKey, RichTerm, RichTerm),
```
yannham, line 147: Is there a reason to use a `RichTerm` here instead of using `Label` directly? It incurs some pattern matching that is useless and has unreachable cases, I imagine.

```suggestion
    Sealed(SealingKey, Label, RichTerm),
```

src/term.rs
```diff
@@ -435,7 +435,7 @@ impl Term {
             Term::Record(..) | Term::RecRecord(..) => Some("Record"),
             Term::Array(..) => Some("Array"),
             Term::SealingKey(_) => Some("SealingKey"),
-            Term::Sealed(_, _) => Some("Sealed"),
+            Term::Sealed(_, _, _) => Some("Sealed"),
```
yannham, line 438: Nitpicking, but using ellipsis for more than one element is more future proof (who knows, one day `Sealed` may become a 7-ary operator :sweat_smile: )

```suggestion
            Term::Sealed(..) => Some("Sealed"),
```

src/term.rs
```diff
@@ -487,7 +487,7 @@ impl Term {
             Term::Record(..) | Term::RecRecord(..) => String::from("{ ... }"),
             Term::Array(..) => String::from("[ ... ]"),
             Term::SealingKey(_) => String::from("<sealing key>"),
-            Term::Sealed(_, _) => String::from("<sealed>"),
+            Term::Sealed(_, _, _) => String::from("<sealed>"),
```
yannham, line 490:
```suggestion
            Term::Sealed(..) => String::from("<sealed>"),
```

src/term.rs
```diff
@@ -574,7 +574,7 @@ impl Term {
             | Term::Op1(_, _)
             | Term::Op2(_, _, _)
             | Term::OpN(..)
-            | Term::Sealed(_, _)
+            | Term::Sealed(_, _, _)
```
yannham, line 577:
```suggestion
            | Term::Sealed(..)
```

src/term.rs
```diff
@@ -614,7 +614,7 @@ impl Term {
...
```
yannham, line 617:
```suggestion
            | Term::Sealed(..)
```

src/typecheck/eq.rs
```diff
@@ -124,8 +124,10 @@ fn contract_eq_bounded(
         (Str(s1), Str(s2)) => s1 == s2,
         (Enum(id1), Enum(id2)) => id1 == id2,
         (SealingKey(s1), SealingKey(s2)) => s1 == s2,
-        (Sealed(key1, inner1), Sealed(key2, inner2)) => {
-            key1 == key2 && contract_eq_bounded(state, inner1, env1, inner2, env2)
+        (Sealed(key1, inner1, lbl1), Sealed(key2, inner2, lbl2)) => {
+            key1 == key2
+                && contract_eq_bounded(state, lbl1, env1, lbl2, env2)
```
yannham, line 129: I think we don't want to compare the labels to decide contract equality.

```suggestion
```

yannham (commented):
Preliminary review. The approach seems to be the correct one. But I think you are making your life more complex by using a `RichTerm` instead of a `Label` directly: that's fine it this means that `%seal%` must become a binary op, and be strict in the label. There is not much need for lazy labels, and those operators are intended for internal use in the stdlib only, anyway.

Once the PR is cleaned up, we can also add some tests.

src/term.rs
ebresafegaga: Using `Label` directly means we have to change the seal operator `%seal%` from a unary operator to a binary operator. I was trying to avoid modifying the parser. But I will have to use a `Label` now, and modify the parser a bit.

yannham: For what it worth, it's a pretty straightforward change. A few lines of boiler plate here and there, but that should be a small change, and we did this (changing operator arity, or adding new operators) several times already :slightly_smiling_face:

src/eval/mod.rs
```diff
@@ -312,6 +312,47 @@ where
...
+                        println!("the label: {:?}", term);
+                        if let Term::Lbl(lbl) = &**term {
+                            return Err(EvalError::BlameError(lbl.clone(), call_stack.clone()));
+                        } else {
+                            // Not a label, should be an error
+                            panic!("")
+                        }
+                    }
+                    None => {
+                        // Is this even possible?
```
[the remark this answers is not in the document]
ebresafegaga: Then maybe we need to write tests were this happens? Currently, none of the tests take this branch.

src/eval/mod.rs
```diff
@@ -317,6 +317,45 @@ where
...
+                        // of the `Seq` operation. See also: https://github.com/tweag/nickel/issues/123
+                        update_thunks(&mut stack, &update_closure);
+                        // Then, evaluate / `Seq` the inner value.
+                        Closure {
+                            body: inner.clone(),
+                            env,
+                        }
+                    }
+                    None | Some(..) => {
+                        // This `Continuation` should not be allowed to evaluate a sealed term
```
yannham, line 354:
```suggestion
                        // This operation should not be allowed to evaluate a sealed term
```

src/eval/mod.rs
```diff
@@ -317,6 +317,45 @@ where
...
+                                term: shared_term.clone(),
+                                pos,
+                            },
+                            env: env.clone(),
+                        };
+                        // Update the original thunk (the thunk which holds the result of the `Seq`) immediately.
+                        // We do this because  we are on a `Sealed` term, and this is in WHNF, and if we don't,
+                        // we will be unwrapping a `Sealed` term and assigning the "unsealed" value to the result
+                        // of the `Seq` operation. See also: https://github.com/tweag/nickel/issues/123
+                        update_thunks(&mut stack, &update_closure);
```
yannham, line 346: Nitpick: if we are updating thunks in both cases (or erroring out), we can do it before the match.

```suggestion
                let closure = Closure {
                    body: RichTerm {
                        term: shared_term,
                        pos,
                    },
                    env,
                 };
                 // Update the original thunk (the thunk which holds the result of the op) in both cases,
                 // even if we continue with a seq.
                 // We do this because  we are on a `Sealed` term, and this is in WHNF, and if we don't,
                 // we will be unwrapping a `Sealed` term and assigning the "unsealed" value to the result
                 // of the `Seq` operation. See also: https://github.com/tweag/nickel/issues/123
                update_thunks(&mut stack, &clos);
                match stack_item {
                    Some(OperationCont::Op2Second(BinaryOp::Unseal(), _, _, _)) => {
                        continuate_operation(closure, &mut stack, &mut call_stack)?
                    }
                    Some(OperationCont::Op1(UnaryOp::Seq(), _)) => {
```

src/eval/operation.rs
```diff
@@ -1125,6 +1111,36 @@ fn process_binary_operation(
...
+                        RichTerm {
+                            term: t2,
+                            pos: pos2,
+                        },
+                    ))
+                }
+            } else {
+                Err(EvalError::TypeError(
+                    String::from("Sym"),
+                    String::from("wrap"),
```
yannham, line 1135:
```suggestion
                    String::from("%seal%, 1st argument"),
```

src/term.rs
```diff
@@ -1180,10 +1179,11 @@ impl RichTerm {
                     pos,
                 )
             },
-            Term::Sealed(i, t) => {
-                let t = t.traverse(f, state, order)?;
+            Term::Sealed(i, t1, lbl) => {
+                let t1 = t1.traverse(f, state, order)?;
+                // Should we ingore the label? (i.e lbl)
```
yannham, line 1184: Yes, we should ignore the label. The label is just error reporting information. It is currently ignore while traversing a metavalue anyway, so let's be consistent here.

```suggestion
```

src/transform/free_vars.rs
```diff
@@ -99,7 +99,7 @@ fn collect_free_vars(rt: &mut RichTerm, free_vars: &mut HashSet<Ident>) {
                 collect_free_vars(t, free_vars);
             }
         }
-        Term::Sealed(_, t) => collect_free_vars(t, free_vars),
+        Term::Sealed(_, t, _lbl) => collect_free_vars(t, free_vars),
```
yannham, line 102:
```suggestion
        Term::Sealed(_, t, _) => collect_free_vars(t, free_vars),
```

src/typecheck/eq.rs
```diff
@@ -124,7 +124,7 @@ fn contract_eq_bounded(
         (Str(s1), Str(s2)) => s1 == s2,
         (Enum(id1), Enum(id2)) => id1 == id2,
         (SealingKey(s1), SealingKey(s2)) => s1 == s2,
-        (Sealed(key1, inner1), Sealed(key2, inner2)) => {
+        (Sealed(key1, inner1, _lbl1), Sealed(key2, inner2, _lbl2)) => {
```
yannham, line 127:
```suggestion
        (Sealed(key1, inner1, _), Sealed(key2, inner2, _)) => {
```

src/typecheck/mod.rs
```diff
@@ -382,7 +382,7 @@ fn walk<L: Linearizer>(
                 _ => Ok(()),
             }
         }
-        Term::Sealed(_, t) => walk(state, env, lin, linearizer, t),
+        Term::Sealed(_, t, _lbl) => walk(state, env, lin, linearizer, t)
```
yannham, line 385:
```suggestion
        Term::Sealed(_, t, _) => walk(state, env, lin, linearizer, t)
```

src/typecheck/mod.rs
```diff
@@ -815,7 +815,7 @@ fn type_check_<L: Linearizer>(
         }
         Term::SealingKey(_) => unify(state, ty, mk_typewrapper::sym())
             .map_err(|err| err.into_typecheck_err(state, rt.pos)),
-        Term::Sealed(_, t) => type_check_(state, env, lin, linearizer, t, ty),
+        Term::Sealed(_, t, _lbl) => type_check_(state, env, lin, linearizer, t, ty),
```
yannham, line 818:
```suggestion
        Term::Sealed(_, t, _) => type_check_(state, env, lin, linearizer, t, ty),
```

src/pretty.rs
```diff
@@ -676,7 +676,7 @@ where
                 .append(allocator.hardline()),
 
             // TODO
-            Sealed(_i, _rt) => allocator.text("#<sealed>").append(allocator.hardline()),
+            Sealed(_i, _rt, _lbl) => allocator.text("#<sealed>").append(allocator.hardline()),
```
yannham, line 679:
```suggestion
            Sealed(_i, _rt, _) => allocator.text("#<sealed>").append(allocator.hardline()),
```

src/eval/operation.rs
```diff
@@ -1415,7 +1431,7 @@ fn process_binary_operation(
                 // Return a function that either behaves like the identity or
                 // const unwrapped_term
 
-                Ok(if let Term::Sealed(s2, t) = t2.into_owned() {
+                Ok(if let Term::Sealed(s2, t, _lbl) = t2.into_owned() {
```
yannham, line 1434:
```suggestion
                Ok(if let Term::Sealed(s2, t, _) = t2.into_owned() {
```

yannham (approved):
Left a bunch of non-essential remarks, but LGTM otherwise. Feel free to do as you see fit with those remarks and then merge on green CI.

PR #803, opened 2022-08-16
Feat/simplify merge operation

Acaccia:
_THIS IS A DRAFT_

This pull request tries to improve the merge operation on records, by reducing the amount of allocations needed for this operation.
As of now, this update is visible on flamegraphs, but the improvement it brings is to light in comparison to the complexity it adds to the code.

I will keep investigating this subject.


yannham:
@Acaccia I know we discussed that internally, but for the record, could you update here what is the status of this PR?

Acaccia:
This PR is a tough choice for me. As I said in the first comment, this seems to improve the performances _slightly_, while clearly making the code more complex.

As we discussed it internally with @yannham, I would advise we check if fixing the current bottleneck, the Environment lookup (see attempts with PRs #807 or #822), makes this the new bottleneck in heavy-merge nickel code, and decide then.

yannham:
This PR is unfortunately stalling, and we aren't planning on changing the environment in the short term. @Acaccia do you think there is anything to be extracted as small, localized changes (like improvement to the `split` function) which don't leak elsewhere in the APIs to make another lighter PR? In any case, this one will keep diverging from master, so we may have to close it at some point.

PR #808, opened 2022-08-26
Handle and report parse errors that cause trouble during evaluation.

### LINKED ISSUE #799: Parse errors in imports are not reported
**Describe the bug**
If a user imports a module that has a parse error, Nickel responds with `this expression has type <unevaluated>, but Type was expected`, rather than reporting the parse error.

**To Reproduce**

Create two Nickel files, `main.ncl` and `lib.ncl`:

```nickel
# main.ncl
let x = import "lib.ncl" in "Hello, " ++ x.name ++ "!"
```

```nickel
# lib.ncl
{
  name = "Nickel",,
}
```

**Expected behavior**

I would expect Nickel to report the parse error:

```console
error: unexpected token
  ┌─ lib.ncl:2:19
  │
2 │   name = "Nickel",,
  │                   ^
```

And indeed, this is exactly what we get when running `lib.ncl` directly. The issue is that running `main.ncl` yields:

```console
error: type error
  ┌─ main.ncl:1:40
  │
1 │ let x = import "lib.ncl" in "Hello, " ++ x.name ++ "!"
  │                                        ^ this expression has type <unevaluated>, but Record was expected
  │
  ┌─ lib.ncl:1:1
  │
1 │ ╭ {
2 │ │   name = "Nickel",,
3 │ │ }
  │ ╰─' evaluated to this
  │
  = field access

```

I presume that after import resolution, `x == Term::ParseError`. This could explain why a type error is raised when the evaluator tries to coerce it into a record.

**Environment**
 - OS name + version: openSUSE Tumbleweed x86_64
 - Version of the code: 0e6cc76f562c8423c4b629c4e188be8eaa82f53c

**Additional context**

Perhaps we could enrich the `Term::ParseError` variant to include the parse error itself, and take extra care to check for this special case before coercing.

But I have a question, why does `Term::ParseError` exist at all? Isn't it easier to error out when resolving the import?

torhovland:
Fixes #799.

Here's a suggested fix. What do you think about this?

The following commands

```
nickel -f main.ncl
nickel -f lib.ncl
```

now both report the same error message:

```
error: unexpected token
  ┌─ /Users/tor/projects/work/nickel/lib.ncl:2:19
  │
2 │   name = "Nickel",,
  │                   ^
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -53,7 +53,7 @@ pub enum EvalError {
         /* evaluated expression */ RichTerm,
     ),
     /// Tried to evaluate a term which wasn't parsed correctly.
-    ParseError(/* parse error */ ParseError),
+    ParseError(ParseError),
     /// A term which is not a function has been applied to an argument.
     NotAFunc(
         /* term */ RichTerm,
--- a/src/error.rs
+++ b/src/error.rs
@@ -52,7 +52,7 @@ pub enum EvalError {
         /* position of the original unevaluated expression */ TermPos,
         /* evaluated expression */ RichTerm,
     ),
-    /// A wrapped parse error that causes trouble during evaluation.
+    /// Tried to evaluate a term which wasn't parsed correctly.
     ParseError(/* parse error */ ParseError),
     /// A term which is not a function has been applied to an argument.
     NotAFunc(
--- a/src/error.rs
+++ b/src/error.rs
@@ -52,6 +52,8 @@ pub enum EvalError {
         /* position of the original unevaluated expression */ TermPos,
         /* evaluated expression */ RichTerm,
     ),
+    /// A wrapped parse error that causes trouble during evaluation.
+    ParseError(/* parse error */ ParseError),
     /// A term which is not a function has been applied to an argument.
     NotAFunc(
         /* term */ RichTerm,
@@ -1036,6 +1038,7 @@ impl ToDiagnostic<FileId> for EvalError {
                     .with_labels(labels)
                     .with_notes(vec![msg.clone()])]
             }
+            EvalError::ParseError(parse_error) => parse_error.to_diagnostic(files, contract_id),
             EvalError::NotAFunc(t, arg, pos_opt) => vec![Diagnostic::error()
                 .with_message("not a function")
                 .with_labels(vec![
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -465,6 +465,8 @@ fn process_unary_operation(
                         pos_op,
                     )), //TODO include the position of operators on the stack
                 }
+            } else if let Term::ParseError(parse_error) = &*t {
+                Err(EvalError::ParseError(parse_error.clone()))
             } else {
                 Err(EvalError::TypeError(
                     String::from("Record"),
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -465,8 +465,6 @@ fn process_unary_operation(
                         pos_op,
                     )), //TODO include the position of operators on the stack
                 }
-            } else if let Term::ParseError(parse_error) = &*t {
-                Err(EvalError::ParseError(parse_error.clone()))
             } else {
                 Err(EvalError::TypeError(
                     String::from("Record"),
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -696,7 +696,7 @@ pub fn subst(rt: RichTerm, global_env: &Environment, env: &Environment) -> RichT
             })
             .unwrap_or_else(|| RichTerm::new(Term::Var(id), pos)),
         v @ Term::Null
-        | v @ Term::ParseError
+        | v @ Term::ParseError(_)
         | v @ Term::Bool(_)
         | v @ Term::Num(_)
         | v @ Term::Str(_)
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -627,6 +627,9 @@ where
                     env: local_env,
                 }
             }
+            Term::ParseError(parse_error) => {
+                return Err(EvalError::ParseError(parse_error.clone()));
+            }
             // Continuation of operations and thunk update
             _ if stack.is_top_thunk() || stack.is_top_cont() => {
                 clos = Closure {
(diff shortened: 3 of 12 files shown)


src/eval/operation.rs
```diff
@@ -465,6 +465,8 @@ fn process_unary_operation(
                         pos_op,
                     )), //TODO include the position of operators on the stack
                 }
+            } else if let Term::ParseError(parse_error) = &*t {
```
fuzzypixelz, line 468: The problem with this is that Nickel imports aren't necessarily behind `StaticAccess`. Most code examples use a top level record to simulate some notion of a module. But you could put in any other expression.

So if you write `import "lib.ncl"` where:

```nickel
# lib.ncl
"Nickel"$
```

The evaluator will escape this check.

Maybe it's better to check for parse error terms in `nickel_lang::eval::eval_closure` instead, as it contains the main loop. I'm not entirely sure if this will cover all cases. 

Another option would be to refactor the `if let Term ... else { Err ...` pattern to make sure all of them test for `ParseError`. Lots of code editing :P

Everything else LGTM.

fuzzypixelz (changes requested):
Thanks for the PR!

src/eval/operation.rs
torhovland: Thanks for the feedback. It was easy to move the check to `eval_closure`, and that covers both parse error examples. Could you take a look at the latest commit?

yannham (changes requested):
Thanks for tackling this :slightly_smiling_face: 

I think you are making your life slightly more complicated that what it needs to be. The `Term::Error` variant is really useful only for the LSP. It shouldn't even be created in the first place during normal execution. Usually, those kind of issues of errors being silently ignored during parsing all boil down to the wrong flavor of the parse function being used (the wrong one being the error tolerant one). 

For example, with this change, if I bury my syntax error inside a term, and don't use the import, it still goes undetected:

```nickel
# file: test.ncl
{ foo = {bar = [2 @ ],}, baz = 2 }
```
```console
$ nickel <<< 'let _ign = import "test.ncl" in 1'
1
```

It's not tragic, but slightly surprising. And since we parse test.ncl in any case because of typechecking, I think we should report the error as soon as possible, not only when trying to evaluate it. This would also let the `eval` module untouched. In particular, it seems `cache::parse_multi` function is error-tolerant, while it should not. As for `parse`, we probably want a `parse_multi` and `parse_multi_lax` function or the like.

One annoying bit is, because parsing imports is done inside the `impl ImportResolver for Cache`, we may have to create newtype wrappers around `Cache` to get two different implementation, one which is error tolerant for the LSP, and the other which is strict for the normal execution workflow. Then, we may also need a `resolve_import` and `resolve_import_lax` function.

Tangentially, one option is to have an option directly stored inside the cache when creating it to toggle error-tolerant parsing in general, so that we don't have to duplicate all those functions with a `_lax` version.

torhovland:
@yannham Thank you for a detailed response. That clears things up. I've verified that I do get a good error message if I create an error-intolerant `parse_multi`:

```
error: unexpected token
  ┌─ /Users/tor/projects/work/nickel/lib.ncl:2:19
  │
2 │   name = "Nickel",,
  │                   ^
  │
  ┌─ /Users/tor/projects/work/nickel/main.ncl:1:9
  │
1 │ let x = import "lib.ncl" in "Hello, " ++ x.name
  │         ---------------- imported here
```

However, if we let the LSP run the error-tolerant version, we're back at the problematic error message originally reported in this issue. Shouldn't that also be fixed, or does it not matter when running on an LSP?

yannham:
It doesn't matter when running the LSP, or, rather, it's a feature™.

Code analysis relies on typechecking to run, but with a parse error, we don't even reach typechecking, and thus the LSP can't provide any information, even if the syntax error is localized deep into a source file (and usually, when one is typing, the state is most often a parse error). So, in order to still provide goto definition, types, etc. for all other part of a program, we made the LSP error tolerant. The LSP still gather those parse errors and display them, it just needs that they don't abort the execution of the pipeline.

I think that you may have the opposite issue, actually: just making `parse_multi` not error tolerant probably prevents the LSP from being error tolerant, which is not what we want. Hence, you may have to create appropriate lax versions (error-tolerant) for `parse_multi` and co, and use them in the LSP pipeline.

torhovland:
Yes, I understand the importance of error-tolerant parsing for the LSP, but I was wondering if bad eval errors caused by parse errors would then be problematic, but I suppose it won't be, because in an LSP context no Nickel code is going to be evaluated anyway, is it?

yannham:
Yes, that is correct. It may happen in the future, if the LSP starts to do smart things like evaluating some contract applications in the background, but we're not there yet.

yannham:
That being said, if we remove the tests, your change is rather small, innocuous, and somehow more future-proof. Maybe we should merge this anyway, and then implement a non error-tolerant parse multi.

torhovland:
Well, on one hand it does feel slightly dirty to wrap parse errors in eval errors. On the other hand, it does improve those eval errors 🙂 

I'm happy to remove those changes and update with the two versions of `parse_multi`. Or I can keep everything. Either way is fine with me. Although removing the tests seems odd.

yannham:
Yes, sorry, I didn't mean to say that we should remove the tests, but rather that they shouldn't be taken into account for the size of the diff. `EvalError` is already wrapping so other errors if I remember correctly, I think it's fine: it's something that can happen during evaluation, or may happen in the future (we will also have to implement a lazy import primitive for Nix, which may also produce such parsing error during evaluation).

Let's keep this PR as it is, and do the `parse_multi` change in a second one?

src/error.rs
```diff
@@ -1036,6 +1038,7 @@ impl ToDiagnostic<FileId> for EvalError {
                     .with_labels(labels)
                     .with_notes(vec![msg.clone()])]
             }
+            EvalError::ParseError(parse_error) => parse_error.to_diagnostic(files, contract_id),
```
yannham, line 1041: I think this is fine for now, but in the future we may want to display an error message that makes it explicit that the error was somehow "dynamic", i.e. happened during evaluation. The reason is that otherwise, it can be confusing to toggle a parse error on and off depending on what get evaluated at runtime (I imagine user expect parse error to always happen).

yannham (approved):

PR #809, opened 2022-08-29
Merge lazy array contracts

fuzzypixelz:
Since the time the `topic/lazy-array-contracts` branch was finalized, master has diverged quite a bit and it's time merge it in finally. 

Let's also use this chance to pick up on any issues that might've been missed before.


yannham:
@fuzzypixelz maybe you've noticed, but there are conflicts to solve before being able to give it a review.

src/eval/operation.rs
```diff
@@ -2412,11 +2611,55 @@ fn eq(env: &mut Environment, c1: Closure, c2: Closure) -> EqResult {
...
             }
         }
-        (Term::Array(l1, _), Term::Array(l2, _)) if l1.len() == l2.len() => {
+        (Term::Array(l1, a1), Term::Array(l2, a2)) if l1.len() == l2.len() => {
             // Equalities are tested in reverse order, but that shouldn't matter. If it
             // does, just do `eqs.rev()`
-            let eqs = l1.into_iter().zip(l2.into_iter());
-            gen_eqs(eqs, env, env1, env2)
+
+            // We should apply all contracts here, otheriwse we risk having wrong values,
```
yannham, line 2618:
```suggestion
            // We should apply all contracts here, otherwise we risk having wrong values,
```

yannham (approved):

PR #810, opened 2022-08-30
Add a "cache" to the Environment

Acaccia:
This PR adds some kind of cache to the environment. It remembers the location of last accessed variables in its layers, for a faster access.

From the first results of experiments, we have a lot of "cache hit" even with a little environment. Now the issue comes with the representation of this cache, which is not "clone-friendly" (the gain we have from direct access to the variables is lost by the clone processes).

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -222,6 +222,65 @@ impl<'a, K: 'a + Hash + Eq, V: 'a + PartialEq> Iterator for EnvIter<'a, K, V> {
     }
 }
 
+mod cache {
+    use std::cell::RefCell;
+    use std::collections::HashMap;
+    use std::hash::Hash;
+    use std::rc::{Rc, Weak};
+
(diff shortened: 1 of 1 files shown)


yannham:
@Acaccia same thing, could you please write the conclusions of this experiment here, for the record? And we can decide what to do with this PR then.

Acaccia:
As I tried with different implementations, I could not come with a representation of the cache which would appear to be "clone-friendly".

- I tried with different sizes for the cache, it seems that my random guess of 5 was the minimum admissible, anything below would result in too frequent cache-miss.

- I tried with a shared cache, but this implementation failed for cases where variables could be shadowed.

- I tried using these with the regular rust `Vec` and a [SmallVec](https://crates.io/crates/smallvec), It did not really have any impact.

Thus, this experiment seems to be a failure and I will close this PR with this comment.

PR #811, opened 2022-08-30
Add capped pretty printing function

yannham:
Implement the display trait for terms using the pretty printer, and add an helper to pretty print a term within a given limited width. Preliminary work for solving #724 (used for reporting failed unification of contracts).

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -1298,6 +1298,28 @@ impl RichTerm {
             TraverseOrder::BottomUp => f(result, state),
         }
     }
+
+    /// Pretty print a term capped to a given max length (in characters). Useful to limit the size
+    /// of terms reported e.g. in typechecking errors. If the output of pretty printing is greater
+    /// than the bound, the string is truncated to `max_width` and the last character after
+    /// truncate is replaced by the ellipsis unicode character U+2026.
+    pub fn pretty_print_cap(&self, max_width: usize) -> String {
+        let output = format!("{}", self);
+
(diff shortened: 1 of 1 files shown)


francois-caddet (approved):

src/term.rs
```diff
@@ -1298,6 +1298,28 @@ impl RichTerm {
...
+    /// than the bound, the string is truncated to `max_width` and the last character after
+    /// truncate is replaced by the ellipsis unicode character U+2026.
+    pub fn pretty_print_cap(&self, max_width: usize) -> String {
+        let output = format!("{}", self);
+
+        if output.len() <= max_width {
+            output
+        } else {
+            let (end, _) = output.char_indices().nth(max_width).unwrap();
+            let mut truncated = String::from(&output[..end]);
```
Acaccia, line 1313: Shouldn't we care about unicode here? What if the term is `Term::String("y̆es")` ?
`"y̆es"` is 3 graphemes but 4 chars...

yannham: I tend to think that we don't really care about cutting off less characters that expected: it's mostly an helper to avoid going over `max_width`. So as long as the string is strictly smaller, I guess that's fine? To my knowledge, iterating over graphemes also requires an external crate (and blow the size of the executable because it needs to include Unicode tables), which doesn't seem to be worth the trouble.

Acaccia: Fine by me

PR #814, opened 2022-08-31
Toggle error tolerance on/off in Cache.

torhovland:
Related to #799 and #808.

Not entirely happy with this. See comment below. What do you think, @yannham?


yannham:
Just to be sure, so the flag only controls what function get called in the implementation of `ImportResolver`, but as far as parsing functions are concerned, it's their name that decide if the parsing is error-tolerant?

torhovland:
@yannham Correct. I think if we have such a flag in `Cache`, it would be confusing to continue serving all those functions as well.

yannham:
Yes, I think so. I think we should either have the flag control everything, or have multiple variants for `parse` and two different implements of `ImportResolver`.

If having just one parsing function disturbs you because both parsing functions have different return type, another solution is to push the flag at the type-level and parametrize the cache by this type-flag:

```rust
// Trait to determine which type parameters are acceptable for Cache 
pub trait CacheMode {}

pub struct ErrorTolerant;
pub struct ErrorStrict;

impl CacheMode for ErrorTolerant {}
impl CacheMode for ErrorStrict {}

pub struct Cache<Mode : CacheMode> {
    // ..
    mode: PhantomData<Mode>
}

impl<T> Cache<T> {
   // all common functions
   // ..
}

impl<ErrorTolerant> Cache<ErrorTolerant> {
  // parse function for error-tolerant parsing, with the right signature
}

impl<ErrorTolerant> Cache<ErrorTolerant> {
  // parse function for non-error-tolerant parsing, with the right signature
}
```

That's a bit more heavier, and now `Cache` has a parameter, but that's also clean, somehow. Otherwise, you can also just use a signature that would work for both (I guess returning ParseErrors works for both mode, even if it is meant to be always empty for non error tolerant parsing)

torhovland:
I tried the parameterized Cache, but I'm not sure I like it, because the different parse functions are polluting other parts of the codebase, such as `load_stdlib` and `prepare`. Perhaps it is possible to implement this cleanly, but I don't see how. The `error_tolerant` flag should be much easier, if we can live with type signatures that aren't ideal.

torhovland:
@yannham OK, so I suggest the changes in this PR, then. Feel free to discuss more!

src/cache.rs
```diff
@@ -70,6 +70,8 @@ pub struct Cache {
     stdlib_ids: Option<Vec<FileId>>,
     /// The inferred type of wildcards for each `FileId`.
     wildcards: HashMap<FileId, Wildcards>,
+    /// Whether processing should try to continue even in case of errors.
+    error_tolerant: bool,
```
yannham, line 74: Nitpick, but I think using a bespoke enum instead of booleans makes the intent clearer:

```rust
pub enum ParsingMode {
  ErrorTolerant,
  ErrorStrict,
}
...
    error_tolerant: ParsingMode
```

Then, when you read e.g. the code above in the LSP `Cache::new(ParsingMode::ErrorTolerant)`, you immediately know what is happening without having to jump to the Nickel codebase (which is a separate crate) and read the documentation of `Cache`. Cf "boolean blindness".

yannham (approved):

yannham:
LGTM! Indeed I haven't thought about the fact that the type signature would propagate the type parameter to all callers of `parse`, in a transitive way. Indeed, the flag looks like the better trade-off overall

PR #815, opened 2022-08-31
[Feature] Optional fields

### LINKED ISSUE #373: Optional fields in contracts
Nickel can't currently easily express optional fields when writing a schema as a record contract. Any field mentioned in the contract must be defined at some point, or this will cause an error at serialization time.

Add a syntax to specify that a field is optional in a record contract. This can be e.g. `?` or `| optional`. This attribute would make the serializer skip such a field in a record instead of raising an error.

Example:

```
let Schema = {
  a | Num,
  b | Num
    | optional,
} in

builtins.serialize `Json Schema // Err: a is not defined
builtins.serialize `Json ({a = 1} | #Schema) // Ok: {"a" : 1}
builtins.serialize `Json ({a = 1, b = "a"} | #Schema) //Err: b's contract fails
builtins.serialize `Json ({a = 1, b = 2} | #Schema) //Ok: {"a": 1, "b": 2}
```

Trying to access directly an optional field that is not defined would raise an error, as for a non-optional one. However, it should probably not appear in `records.hasField`, `records.fieldsOf` and `records.map` if they are not defined.

As we can write custom contracts as functions, an equivalent `Schema` contract can already be implemented manually. However, we lose all the advantages of writing schema simply as records: this is tedious, less-efficient, and this is not composable, because we can't apply record operations (like merging, field access, and so on.) on it.

yannham:
# Optional fields

Closes #373.

## Overview

This PR adds the keyword `optional` (subject to bikeshed) to the metadata syntax to describe a field that may be filled, but is not required.

The general guideline for the semantics of optional fields is that an optional field without definition should be transparent to all record operations, and serialization. Once it get merged with a definition, or with the same field without definition but that is not optional, it acts like a normal field.

In particular, an empty optional field won't appear when calling to `record.fields` and will return false when calling to `record.has_field`. The rationale is that we don't want `record.has_field "foo" r` to return `true` on an empty optional field and then have `r.foo` to fail. We may want functions like `fields_all` or `fields_optional` that include empty optional fields later, but I don't see an obvious usage for now (dynamic contract inspection/building, maybe?), so let's wait for someone to actually ask for them.

## Deciding "is an empty optional"

The way we handle metadata in Nickel is by nature very dynamic: any value can have metadata, and in principle we can't fetch the full metadata of an expression, or decide properties like "is it defined" without performing a form of evaluation, which can involve arbitrary computations. We can have nested metadata, etc.

This means that deciding "is this field an empty optional" is potentially arbitrary long, and causes the evaluation of the field, which sounds bad. For example, `record.fields` requires to decide "is empty optional" on each field, to filter the list. It feels more natural to store this information at the field level, as a simple flag, and update upon merge, such that we have a direct access in constant time to this property for any given field.

## Solution implemented in this PR

However, this would make optional fields quite special, without a good reason. The same dilemma applies to many other metadata. This is why I think we should have a general reflection for metadata (should they only be allowed on record fields and stored there, basically), instead of special casing optional. In the meantime, this PR uses an heuristics to try to detect quickly an empty optional, which I hope will be more than enough for most concrete cases (indeed, while it may be arbitrarily complex to detect in theory, the natural use of optional fields should stay rather simple). The lack of optional fields have been limiting in several experimentation already, when trying to write contracts for concrete use-cases, so in my humble opinion, it's better than nothing to have them with some limitations now, and potentially make them better later, rather than waiting for a restructuring of metadata.

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -241,6 +241,7 @@ pub fn merge(
                 doc: doc1,
                 types: types1,
                 contracts: contracts1,
+                opt: opt1,
                 priority: priority1,
                 value: value1,
             } = meta1;
@@ -248,6 +249,7 @@ pub fn merge(
                 doc: doc2,
                 types: types2,
                 contracts: contracts2,
+                opt: opt2,
                 priority: priority2,
                 value: value2,
             } = meta2;
@@ -345,6 +347,9 @@ pub fn merge(
                 doc,
                 types,
                 contracts,
+                // If one of the record requires this field, then it musn't be optional. The
+                // resulting field is optional iff both are.
+                opt: opt1 && opt2,
                 priority,
                 value,
             };
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -878,5 +878,60 @@ pub fn subst(rt: RichTerm, initial_env: &Environment, env: &Environment) -> Rich
     }
 }
 
+/// Checks if the given term is a chain of nested and/or merged metavalues such that:
+///
+/// 1. At least one metavalue has the `optional` flag set
+/// 2. The final value is undefined
+///
+/// This function is used to determine if a record field is an optional field without definition,
+/// and should thus be ignored by record operations.
+///
+/// Beware: correctly checking that a value is an empty optional can incur arbitrary computations
+/// in principle, but this function is supposed to be a quick peek. It's a terminating
+/// approximation and hence is necessarily incomplete. In practice, this function is able to
+/// traverse merge expressions and follow variables links, but within a limit (in the number of
+/// variables followed).
+pub fn is_empty_optional(rt: &RichTerm, env: &Environment) -> bool {
+    fn is_empty_optional_aux(rt: &RichTerm, env: &Environment, is_opt: bool, gas: &mut u8) -> bool {
+        match rt.as_ref() {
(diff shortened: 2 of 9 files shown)


Acaccia (approved):

thufschmitt:
It feels a bit weird to have `optional` be a metadata field and not a property of the structure of the record itself. What's the reason for that? You mention that “this would make optional fields quite special, without a good reason”, but it intuitively seems to me that they have to be a first-class construct for records :thinking:

yannham:
@thufschmitt to be clear, I think having optional be a property of the record itself is eventually the right solution. But to me, it seems that this is not specific to `optional`: this is the same dilemma for e.g. `doc`, the priority (`default`, etc.), possibly the future `hidden`, `serialize_with`, custom merge function, etc. The rationale is then not to introduce an ad-hoc parameter for optionals right now, but to do like we've done until now temporarily, and to think about basically moving all metadata but annotations to be attached directly to record fields, at once.

I'm just not 100% convinced the somehow hacky solution of this PR is fine, even temporarily. But still a bit convinced.

thufschmitt:
Fair point. It's less obvious (to me at least) for stuff like `doc`, but it makes sense. Let's :+1: this and wait for https://github.com/tweag/nickel/pull/819/ to have a better representation

thufschmitt (approved):

PR #819, opened 2022-09-07
[RFC005] Metada representation and semantics

yannham:
Add RFC005, which proposes a rework of the representation and the semantics of metadata.

### DIFF
--- a/rfcs/005-metadata-rework.md
+++ b/rfcs/005-metadata-rework.md
@@ -1,0 +1,611 @@
+---
+feature: representation and semantics of metadata
+start-date: 2022-09-07
+author: Yann Hamdaoui
+---
+
+# The representation and semantics of metadata
+
+The goal of this RFC is to rethink how metadata (documentation, priority, custom
+merge function, etc.) are represented, accessed, and propagated, and what is
+their semantics with respect to merging.
+
+## Motivation
+
+This RFC has two main motivations:
+
+1. **Merge semantics**: Fix a number of strange and unintuitive behaviors
+   observed in the wild while prototyping a Nix package set written in Nickel. A
+   detailed report is given in
+   [#710](https://github.com/tweag/nickel/issues/710). Most of the examples and
+   explanations from this report are reproduced here for this RFC to be
+   self-contained.
+2. **Representation**: Simplify the implementation and make it more efficient.
+   The current representation of metadata implies that fetching the metadata of
+   a value may involve evaluating arbitrary Nickel expressions, which doesn't
+   really fit well with the idea of metadata. One striking example is optional
+   fields, which is detailed later in this document.
+
+### Merge semantics
+
+This section explores the first motivation: the current semantics of merging
+metadata is fragile.
+
+The idea of merging is mainly inspired from two models[^1], the [NixOS module
+system](https://nixos.wiki/wiki/Module) and the [CUE programming
+language](https://cuelang.org/docs/concepts/logic/).
+
+The NixOS module system and the CUE merge system are fundamentally based on the
+idea of composing small logical data blocks into a final result. Those blocks
+are not necessarily independent: a value defined in a block may depend on other
+values that will only be available later in the big merged blob (for NixOS
+modules, those are typically options). In Nickel terms, this is often referred
+to as _recursive overriding_.
+
+This capability is what makes merging powerful, but comes with its share of
+questions. What should be the result of the following program?
+
+```nickel
+{
+  foo | Num
+      | default = 5,
+  bar = foo,
+} & { foo = "a" }
+```
+
+If anything, this expression should evaluate to `{foo = "a", bar = "a"}`.
+However, the fact that `foo` has been annotated with a `Num` contract but may
+end up with a string in the final result without an error is disturbing. This
+example is small and artificial, but in real life the two merged blocks could be
+defined in different files, and the reason as to why an `"a"` ends up here where
+a number was expected could be hard to investigate.
+
+In CUE, the underlying model guides the answer to this question. Types and
+values are not distinct entities (roughly, values can be seen as types with only
+one inhabitant, i.e. singletons), but live all in the same space. Merging
+intuitively corresponds to intersection. In that interpretation, trying to merge
+`"a"` with something of type `Num` will throw an error. Similarly, in NixOS
+modules, a configuration option of type string won't be merged with a number
+without an error (Nix doesn't have builtin types per se, but the NixOS module
+system is a library with a notion of type similar in spirit to Nickel's
+contracts).
+
+Thus, it seems reasonable to do the same in Nickel. Let us write a first
+specification that we would like our system to enjoy:
+
+**Eventual validity `(EvVal)`**: let `e` be a merge expression
+`x_1 & ... & x_n`, where each `x_i`s evaluate to a record. Then, if there exists
+`j` such that `x_j` has a field `f` with a contract `C` attached, then `e.f | C`
+(`e.f` respects the contract `C`)
+
+#### Cross-application
+
+Nickel does respect `(EvVal)`. Actually, it does something stricter in the sense
+that more expressions are blamed than what would be strictly required by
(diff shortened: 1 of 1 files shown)


thufschmitt (commented):
I like this a lot!

A few questions though:

- How will this interact with custom merge functions? With the current design, I intuitively expect that I'll be able to define say `let foo = 1 | merge (fun x y => x + y) in foo & foo`. I assume that wouldn't be possible anymore?
- The difference between applying a contract to a record (`{ foo = 5 } | { foo | Num }`) and applying a contract to the fields (`{ foo = 5 | Num }`) is not very intuitive – esp. since `{ foo = 5 } & { foo | Num }` has a different meaning than the first one IIUC. Not sure what can be done about it, but it's a bit confusing

rfcs/005-metadata-rework.md
```diff
@@ -0,0 +1,562 @@
...
+```text
+nickel>{
+  foo | {
+    bar : Num,
+    baz : Str
+  }
+}
+& {foo = {}}
+& {foo.bar = 1}
+& {foo.baz = "a"}
```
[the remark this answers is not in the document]
yannham: Ah, I haven't even thought about that, good catch!

yannham:
> How will this interact with custom merge functions? With the current design, I intuitively expect that I'll be able to define say let foo = 1 | merge (fun x y => x + y) in foo & foo. I assume that wouldn't be possible anymore?

Indeed, as for other metadata, you can only specify custom merge functions at field level. But you can still encode your example anyway, as `let r = {foo | merge (+) = 1} in (r & r).foo`. That being said, merge functions outside of record fields sound a bit unnatural to me, or even abusing the semantics of merge (like most metadata outside of field definitions). Or do you have a particular use-case in mind?

> The difference between applying a contract to a record ({ foo = 5 } | { foo | Num }) and applying a contract to the fields ({ foo = 5 | Num }) is not very intuitive – esp. since { foo = 5 } & { foo | Num } has a different meaning than the first one IIUC. Not sure what can be done about it, but it's a bit confusing

It may need clarification in the document, but actually `({foo = 5} | {foo | Num})` should behave the same as `{foo | Num = 5}` and `{foo | Num} & {foo = 5}`. Indeed, application of record contract is merging (there are a few differences with respect to closed record contracts and top-level metadata, but they are irrelevant here). So, even with this proposal, `({foo | default = 5} | {foo | Num}) & {foo = "a"}` will blame, because it will evaluate to `{foo | Num | default = 5} & {foo = "a"}` and then be cross-applied as a metadata.

The difference with the current situation is that right now, not only `{foo = 5} | {foo | Num}` reduces more or less to `{foo | Num = 5}`, but `{foo | Num}` also *cross-applies* to other merge operands, giving the strange failure reported in "Outer cross-application doesn't make sense":

```text
nickel>({foo = 5} | {foo | Num})
& {bar = "bar"}
error: contract broken by a value: extra field `bar`
  ┌─ :1:1
  │
1 │ { ... }
  │ ------- expected type
  │
  ┌─ repl-input-1:2:3
  │
2 │ & {bar = "bar"}
  │   ^^^^^^^^^^^^^ applied to this expression
  │
  │
1 │ { ... }
  │ ------- evaluated to this value
note:
  ┌─ repl-input-1:1:14
  │
1 │ ({foo = 5} | {foo | Num})
  │              ^^^^^^^^^^^ bound here
```

In the end, the proposed solution looks to me like the most natural. The issue is maybe that I insisted too much on "contract application outside of fields has nothing to do with metadata anymore", while this isn't entirely true, because a record contract annotation can result in metadata being propagated?

thufschmitt:
> It may need clarification in the document, but actually ({foo = 5} | {foo | Num}) should behave the same as {foo | Num = 5} and {foo | Num} & {foo = 5}.

Aaaah yes, get it now. So it's actually very natural. Thanks :)

> merge functions outside of record fields sound a bit unnatural to me, or even abusing the semantics of merge

Fair enough, it definitely is :)

One related thing that comes to my mind though is the floating idea of having Nix-style contexts as metadata. I assume that wouldn't work as nicely anymore with this representation (because in this case you really want it to be associated with the value and not a specific record field). Which is mildly annoying since it means that we'd probably either have to give-up on them or have some special-cased AST construct

vkleen:
The proposal looks good to me. The issues it discusses have bitten me regularly with the Terraform+Nickel project and it seems to me that metadata attached to fields together with the new merging strategy would alleviate a lot of problems.

Regarding contexts as metadata, I think I would second @yannham's suggestion of introducing a separation between "field metadata" and "value metadata" at some point. It seems to me that the current metadata approach tries to address two complementary usecases and this leads to all the unintuitive behaviour and complications that we are seeing.

rfcs/005-metadata-rework.md
```diff
@@ -0,0 +1,562 @@
...
+   different node (say, `Annotated`) and wouldn't impact the behavior of
+   merging.
+2. Instead of implementation `(CrossApp)`, we implement a laxer strategy,
+   lazy propagation (which corresponds closely to `(EvVal)`). As we've done with
+   [lazy array contracts](https://github.com/tweag/nickel/pull/764), contracts
+   would be accumulated at the field level instead of being eagerly applied.
+   Only when the field is extracted (or evaluated by any other mean: equality
+   testing, deep sequing, etc.), we can apply the pending contracts. As for lazy
+   array contracts, this also enables new potential optimizations, like
+   eliminating redundant contract applications.
```
aspiwack, line 411: We're adding contract conjunction again, aren't we? This may be an issue. We a least need to clarify the semantics here.

rfcs/005-metadata-rework.md
```diff
@@ -0,0 +1,562 @@
...
+when that happens: no later arbitrary evaluation, no nesting, just static data.
+Allowing them anywhere doesn't bring clear practical benefits, whereas it comes
+with substantial drawbacks, exposed in the previous sections.
+
+### Specification
+
+The proposed change comes in to aspects:
+
+1. Metadata can only be written next to record fields, and are stored only there
+   in the AST. Their semantics with respect to merging is left unchanged, except
```
aspiwack, line 399: As I've told you in off-channel conversations, I think they could be attached to let bindings as well. Not sure that it's quite as useful though. So maybe we don't have to think about it quite now.

aspiwack (commented):
You are making a quite comprehensive case. It's all well-documented and convincing.

rfcs/005-metadata-rework.md
yannham: In some sense, yes, but in fact, we already do this currently (when cross-applying contracts). Here we just propose to do that a bit more lazily. The semantics is that this should be the same as annotating a value with several contracts at once:

```
foo | A
    | B
    | C
```

Which just means applying the contract in sequence, id est the "naive" intersection. Do you think it needs to be emphasized?

aspiwack: Yes, it's probably worth it, and it can't hurt.

On function it's more or less `(A → B) ∩ (C → D) = (A∩C) → (B∩D)`, correct?

rfcs/005-metadata-rework.md
yannham: It could make sense indeed, to attaching documentation for example. But I think it may bring the same issue discussed here if we allow for example default values. I would vouch for let's implement this for record fields only, and consider later an extension if this seems to be really needed.

rfcs/005-metadata-rework.md
yannham: > On functions it's more or less (A → B) ∩ (C → D) = (A∩C) → (B∩D), correct?

Correct.

yannham:
Added a paragraph on broken associativity ([bc63c03](https://github.com/tweag/nickel/pull/819/commits/bc63c03047004fe0a681f04e47e2391b159b89d4)) and on the intersection being naive ([f67694a](https://github.com/tweag/nickel/pull/819/commits/f67694afa120b34668d973d52afd918ca566e2df)). Modulo a rebase, I think it's in good shape for being merged.

PR #807, opened 2022-08-22
Improve environment lookup times

fuzzypixelz:

PR #822, opened 2022-09-09
Perf/environment as hamt

Acaccia:
This PR is an alternative to #807 to improve the lookup time in the `Environment`, without the side effect of coupling the `Environment` code with transformations during the nickel code parsing.

Here we try an HAMT, Being the default representation of hashmaps in functional languages like Clojure or Scala (source: Wikipédia :p ).
Those structures allows for a fast lookup and a O(1) clone operation.

First flamegraph (with a program which resembles to `countLetters` benchmark) is encouraging, with "negligible" lookup and clone, but big insertion time.
image


Acaccia:
As we can see in the previous comment image, we can try improving the insertion time by making our `Ident` "clone-friendlier". Two alternatives were tried in this end:

- using the [SmolStr](https://crates.io/crates/smol_str) crate
- replacing the String by an Rc<String>.

In both cases, running the benchmarks of array had worse results than the current implementation.

The insertion seems to be too expensive with the HAMT for it to be a viable solution.

yannham:
@Acaccia I wonder if we should explore this more. On paper, HAMT seem like the perfect data structure for our use-case: O(1) insertion, O(1) lookup, and O(1) sharing. The current implementation probably inserts a lot of generated variable in the environment for closurization, but that could probably be made in a different way. Maybe we should at least write an issue, with a link to this PR, to remind us to try again if we change the representation of closure?

PR #826, opened 2022-09-12
Nix with parsed and transpiled with let shadowing:

francois-caddet:
First impl of the nix with keyword.
Shadowing by Outer `let ... in` blocks is working.
Will add functions and rec-records shadowing in a later PR.
Many clones are performed but it will be fixed in a code cleaning PR.

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -1,6 +1,7 @@
 use crate::cache::Cache;
 use crate::conversion::State;
 pub use crate::conversion::ToNickel;
+use crate::identifier::Ident;
 use crate::mk_app;
 use crate::parser::utils::mk_span;
 use crate::term::make::{self, if_then_else};
@@ -142,7 +143,22 @@ impl ToNickel for rnix::SyntaxNode {
                 "true" => Term::Bool(true),
                 "false" => Term::Bool(false),
                 "null" => Term::Null,
-                id => Term::Var(id.into()),
+                id => {
+                    if state.env.contains(id) || state.with.is_empty() {
+                        Term::Var(id.into())
+                    } else {
+                        Term::App(
+                            mk_app!(
+                                make::op1(
+                                    UnaryOp::StaticAccess("with".into()),
+                                    Term::Var("compat".into()),
+                                ),
+                                Term::Array(state.with.clone(), Default::default())
+                            ),
+                            Term::Str(id.into()).into(),
+                        )
+                    }
+                }
             }
             .into(),
             ParsedType::LegacyLet(_) => panic!("Legacy let form is not supported"), // Probably useless to suport it in a short term.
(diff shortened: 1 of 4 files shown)


src/nix.rs
```diff
@@ -142,7 +143,22 @@ impl ToNickel for rnix::SyntaxNode {
...
+                id => {
+                    if state.env.contains(id) || state.with.is_empty() {
+                        Term::Var(id.into())
+                    } else {
+                        Term::App(
+                            mk_app!(
+                                make::op1(
+                                    UnaryOp::StaticAccess("with".into()),
+                                    Term::Var("compat".into()),
+                                ),
```
yannham, line 155: Suggestion: do as for builtin contracts. Create a module named `compat` in the `stdlib` module (the Rust one) that exports this term as `with()`, and just replace this code here `stdlib::compat::with()`, or even `stdlib::compat::with(array)` to avoid cluttering the code and be more resilient to refactoring of the Nickel stdlib.

stdlib/compat.ncl
```diff
@@ -0,0 +1,12 @@
+{
+compat | doc "Nix compatibility layer. should not be used by Nickel program."
+= {
+  with: Array {_: Dyn} -> Str -> Dyn
```
yannham, line 4: I don't know if we want a type annotation right now. In spirit, this is great. In practice, it incurs a contract check each time. Maybe just typecheck it once and then remove it? I guess it could be nice to have pragmas to say "don't generate contract application for this file", but we don't have that yet

src/nix.rs
```diff
@@ -180,10 +215,19 @@ impl ToNickel for rnix::SyntaxNode {
...
-            ParsedType::With(n) => unimplemented!(),
+            ParsedType::With(n) => {
+                let State { file_id, env, with } = state;
+                let mut with = with.clone();
+                with.push(n.namespace().unwrap().translate(state));
+                n.body().unwrap().translate(&State {
+                    file_id: file_id.clone(),
+                    env: env.clone(),
+                    with,
+                })
```
yannham, line 229:
```suggestion
                let mut state = state.clone();
                state.with.push(n.namespace().unwrap().translate(state));
                n.body().unwrap().translate(&state)
```

stdlib/compat.ncl
```diff
@@ -0,0 +1,12 @@
...
+  = fun envs field => (
+    array.fold (fun prev current =>
+      if record.has_field field current
+      then current
+      else prev
```
yannham, line 9:
```suggestion
    array.fold (fun current prev =>
      if record.has_field field prev
      then prev
      else current
```

I think you swapped the two. I was confused at first, because given the definition of `fold`, I thought this code would incorrectly select the oldest environment with "field" (the first one in the list), instead of the latest. But actually it is correct.

But then, you are testing n times if the field is inside, even if you already know it is. I would do something like:

```
    array.fold (fun current acc =>
      if !acc.found && record.has_field current field
      then {value = current."%{field"}", set = true}
      else acc
```

Then you have to project `.value` instead of `."%{field}"` and use a starting value `{value = null, found = false}` or just `{found = false}`.

yannham (approved):
It's missing tests, and tangentially, we could use a better data structure than hashmaps and hashsets for environment-like requirements, but let's postpone those for now. The logic seems correct.

stdlib/compat.ncl
francois-caddet: Yes, better like this :) but I have to revert the array right?

yannham: No, because you first look into the accumulator "have the field been already found in the tail" ? If not, set this one as found, otherwise, use the one of the tail. Because this is a right fold, this should be in the right direction.

PR #828, opened 2022-09-13
[Refactor] Pull out Nickel test helpers in testlib.ncl

yannham:
A couple functions are defined and used again and again throughout tests on the Nickel side. This PR puts them in a separate file that is then imported in each test, avoiding having to redefine them each time.


francois-caddet (approved):
Good :).
May Assert be added in the stdlib? it's really used everywhere.

yannham:
Mmh, maybe, yeah. It gives nice error reporting, as opposed to a simple failing function. But left for a future PR :)

PR #829, opened 2022-09-13
[Feature] Merge priorities

yannham:
Depends on #198. Implement merge priorities as of [RFC001](https://github.com/tweag/nickel/blob/master/rfcs/001-overriding.md). The special operator `default rec` and `force rec` are not implemented in this PR but left as future work.

### DIFF
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -254,16 +254,64 @@ final record:
 ## Merging record with metadata
 
 Metadata can be attached to values thanks to the `|` operator. Metadata
-currently includes contract annotations, default value, and documentation. We
-describe in this section how metadata interacts with merging.
+currently includes contract annotations, default value, merge priority, and
+documentation. We describe in this section how metadata interacts with merging.
 
-### Default values
+### Merge priorities
+
+Priorities are specified using the `priority` annotation, followed by a number
+literal. There are also two other special priorities, the bottom priority, specified
+using the `default` annotation, and the top priority, specified using the
+`force` annotation.
+
+Priorities dictate which values take precedence over the other values. By
+default, values are given the priority `0`. Values with the same priority are
+recursively merged as specified in this document, which can mean failure if the
+values can't be meaningfully merged:
+
+```text
+nickel>{foo | default = 1} & {foo = 2}
+error: non mergeable terms
+  ┌─ repl-input-1:1:8
+  │
+1 │ {foo = 1} & {foo = 2}
+  │        ^           ^ with this expression
+  │        │
+  │        cannot merge this expression
+  │
+  = Both values have the same merge priority but they can't be combined
+```
+
+On the other hand, if the priorities differ, this value with highest priority
+silently erases the other in the final result:
+
+```text
+nickel>{foo | priority 1 = 1} & {foo = 2}
+{ foo = 1 }
+
+nickel>{foo | priority -1 = 1} & {foo = 2}
+{ foo = 2 }
+```
+
+The priorities are ordered in the following way:
+
+- bottom is the lowest priority
+- numeral priorities are ordered as usual numbers (priorities can be any valid Nickel
+  number, including fractions and negative values)
+- top is the highest priority
+
+#### Default values
 
 A `default` annotation can be used to provide a base value, but let it be
 overridable through merging. For example, `{foo | default = 1} & {foo = 2}`
-evaluates to `{foo = 2}`. Without the default value, this merge would have
-failed with a `non mergeable fields` error, because merging being symmetric, it
-doesn't know how to combine `1` and `2` in a generic and meaningful way.
+evaluates to `{foo = 2}`. Default values are just a special case of priorities,
+being the lowest possible one.
+
+#### Forcing values
+
+Dually, values with the `force` annotation are given the highest priority. Such
+a value can never be overriden, and will either take precedence over another
+values or be tentatively combined if merged with another forcing value.
 
 #### Specification
 
@@ -283,10 +331,6 @@ same on both side:
            left & right  if p(left) = p(right)
 }
 ```
-
-Currently, there are only two priorities, `normal` (by default, when nothing is
-specified) and the `default` one, with `default < normal`. We plan to add more
-in the future (see [RFC001](https://github.com/tweag/nickel/blob/c21cf280dc610821fceed4c2caafedb60ce7177c/rfcs/001-overriding.md#priorities)).
 
 #### Example
 
--- a/examples/merge-priorities/main.ncl
+++ b/examples/merge-priorities/main.ncl
@@ -1,0 +1,20 @@
+# Merge several blocks into one final configuration. In a real world case, one
+# would also want contracts to validate the shape of the data.
+let server = import "server.ncl" in
+let security = import "security.ncl" in
+# Disabling firewall in the final result
+server & security & {
+  #As opposed to the simple merge example, this would now fail
+  #firewall.enabled = false
+
+  firewall.open_ports | priority 10 = [80],
+  firewall.type = "superiptables",
+  server.host.ip = "89.22.11.01",
+
+  # this will only be selected if no values with higher priority (or no priority
+  # annotation, which is the same as priority 0) is ever defined
+  # because there's a definite value below, this won't be selected.
+  server.host.name | priority -1 = "hello-world.backup.com",
+} & {
+  server.host.name = "hello-world.main.com",
+}
--- a/src/error.rs
+++ b/src/error.rs
@@ -1116,7 +1116,10 @@ impl ToDiagnostic<FileId> for EvalError {
 
                 vec![Diagnostic::error()
                     .with_message("non mergeable terms")
-                    .with_labels(labels)]
+                    .with_labels(labels)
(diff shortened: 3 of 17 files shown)


src/term.rs
```diff
@@ -287,15 +295,131 @@ pub struct RecordDeps {
...
+    /// The priority of default values that are overridden by everything else.
+    Bottom,
+    /// The priority by default, when no priority annotation (`default`, `force`, `priority`) is
+    /// provided.
+    ///
+    /// Act as the value `MergePriority::Numeral(0.0)` with respect to ordering and equality
+    /// testing. The only way to discriminate this variant is to pattern match on it.
+    Neutral,
+    /// A numeral priority. The inner value should never be `NaN`. Comparing a `MergePriority` with
+    /// a `NaN` value will panic.
```
yannham, line 357: This comment isn't needed anymore, now that I've defined a proper `NumeralPriority` wrapper around floats.

```suggestion
    /// A numeral priority.
```

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+using the `default` annotation, and the top priority, specified using the
+`force` annotation.
+
+Priorities dictate which values take precedence over other values. By default,
+values are given the priority `0`. Values with the same priority are recursively
+merged as specified in this document, which can mean failure if the values can't
+be meaningfully merged:
+
+```text
+nickel>{foo | default = 1} & {foo = 2}
```
francois-caddet, line 273:
```suggestion
nickel> {foo | default = 1} & {foo = 2}
```

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+error: non mergeable terms
+  ┌─ repl-input-1:1:8
+  │
+1 │ {foo = 1} & {foo = 2}
+  │        ^           ^ with this expression
+  │        │
+  │        cannot merge this expression
+  │
+  = Both values have the same merge priority but they can't be combined
+```
```
francois-caddet, line 283: Mistake here no? the above exemple has different priority, first record has `foo` annoted default.

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+
+On the other hand, if the priorities differ, the value with highest priority
+simply erases the other in the final result:
+
+```text
+nickel>{foo | priority 1 = 1} & {foo = 2}
```
francois-caddet, line 289:
```suggestion
nickel> {foo | priority 1 = 1} & {foo = 2}
```

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+{ foo = 1 }
+
+nickel>{foo | priority -1 = 1} & {foo = 2}
```
francois-caddet, line 292:
```suggestion
nickel> {foo | priority -1 = 1} & {foo = 2}
```

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+
+#### Default values
 
 A `default` annotation can be used to provide a base value, but let it be
 overridable through merging. For example, `{foo | default = 1} & {foo = 2}`
-evaluates to `{foo = 2}`. Without the default value, this merge would have
-failed with a `non mergeable fields` error, because merging being symmetric, it
-doesn't know how to combine `1` and `2` in a generic and meaningful way.
+evaluates to `{foo = 2}`. A default value is just a special case of a priority,
+being the lowest possible one.
```
francois-caddet, line 308: Even if the implementationementation doesn't manage as such, may we say it is like having `-inf` value? Actually, how nickel manage infinites?

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+
+#### Forcing values
+
+Dually, values with the `force` annotation are given the highest priority. Such
+a value can never be overriden, and will either take precedence over another
+value or be tentatively merged if the other value is forcing as well.
```
francois-caddet, line 314: same idea as default but with `+inf` here.

examples/merge-priorities/main.ncl
```diff
@@ -0,0 +1,20 @@
+# Merge several blocks into one final configuration. In a real world case, one
+# would also want contracts to validate the shape of the data.
+let server = import "server.ncl" in
+let security = import "security.ncl" in
+# Disabling firewall in the final result
+server & security & {
+  #As opposed to the simple merge example, this would now fail
+  #firewall.enabled = false
```
francois-caddet, line 8:
```suggestion
  # As opposed to the simple merge example, this would now fail
  # firewall.enabled = false
```

src/error.rs
```diff
@@ -1116,7 +1116,10 @@ impl ToDiagnostic<FileId> for EvalError {
 
                 vec![Diagnostic::error()
                     .with_message("non mergeable terms")
-                    .with_labels(labels)]
+                    .with_labels(labels)
+                    .with_notes(vec![String::from(
+                        "Both values have the same merge priority but they can't be combined",
```
francois-caddet, line 1121:
```suggestion
                        "Both values have the same merge priority so they can't be combined",
```

src/parser/uniterm.rs
```diff
@@ -211,7 +211,7 @@ impl UniRecord {
                                         types: Some(ctrt),
                                         contracts,
                                         opt: false,
-                                        priority: MergePriority::Normal,
+                                        priority: MergePriority::Neutral,
```
francois-caddet, line 214: An alternative could be to use `Option<float>`.
then `Bottom` become `Some(-float::inf())`, `Top` become `Some(float::inf())`, and `Neutral` become `None`
Sorry not sure how to write infinite in Rust but that's the idea.
May easy the priority comparaisons but may be less clear in the code...

src/repl/query_print.rs
```diff
@@ -235,18 +235,19 @@ fn write_query_result_<R: QueryPrinter>(
...
                     renderer.write_metadata(out, "default", &t.as_ref().shallow_repr())?;
                     found = true;
                 }
                 MetaValue {
-                    priority: MergePriority::Normal,
+                    priority: MergePriority::Numeral(n),
                     value: Some(t),
                     ..
                 } if selected_attrs.value => {
+                    renderer.write_metadata(out, "priority", &format!("{}", n))?;
```
francois-caddet, line 250: an improvement could be to check 0 equality and don't print anything if it's the case

src/term.rs
```diff
@@ -287,15 +295,130 @@ pub struct RecordDeps {
...
+            (MergePriority::Bottom, MergePriority::Bottom)
+            | (MergePriority::Neutral, MergePriority::Neutral)
+            | (MergePriority::Top, MergePriority::Top) => true,
+            (MergePriority::Numeral(p1), MergePriority::Numeral(p2)) => p1 == p2,
+            (MergePriority::Neutral, MergePriority::Numeral(p))
+            | (MergePriority::Numeral(p), MergePriority::Neutral)
+                if p == &NumeralPriority::zero() =>
+            {
+                true
+            }
```
francois-caddet, line 380:
```suggestion
                if p == &NumeralPriority::zero() => true,
```

francois-caddet (approved):
A nice improvement for nickel with a clear implementation :)
Only a few changes proposed.
Just to be sure, are we really not needing `Merge` token anymore? maybe for RFC001 (custom merge functions)?

doc/manual/merging.md
yannham: Infinite can't be represented as a float literal (they can be obtained by operations on float, I guess, but not written down), so `+inf` and `-inf` aren't currently possible as numeral priorities. I don't know, is it really that clearer to say that it is `-inf`, rather than spelling out loud "this is the lowest possible priority"? To me the latter requires less maths notion (arguably, `+inf` and `-inf` are not exactly rocket science, but still). I don't have a strong opinion, honestly.

src/term.rs
yannham: That's the formatted that has decided about the braces, not me :slightly_smiling_face:

src/repl/query_print.rs
yannham: Not really, because `0.0` may have been written down. We could check for `Neutral` indeed. On the other hand, it can be useful to know the actual priority, even if it hasn't been annotated?

src/error.rs
yannham: I would keep the `but`. 2 values with the same priority may very well be combined, like records. Here the problem is that they have the same priority AND we don't know how to merge them.

doc/manual/merging.md
```diff
@@ -254,16 +254,64 @@ final record:
...
+using the `default` annotation, and the top priority, specified using the
+`force` annotation.
+
+Priorities dictate which values take precedence over other values. By default,
+values are given the priority `0`. Values with the same priority are recursively
+merged as specified in this document, which can mean failure if the values can't
+be meaningfully merged:
+
+```text
+nickel> {foo | default = 1} & {foo = 2}
```
yannham, line 273:
```suggestion
nickel> {foo = 1} & {foo = 2}
```

doc/manual/merging.md
yannham: Ah, yes, well-spotted.

PR #831, opened 2022-09-15
Fix polarity for polymorphic contract failure

### LINKED ISSUE #821: Improve error messages and fix blame for polymorphic contract failure
Currently, the error message when we have polymorphic contract failure is not particularly helpful:
```
nickel> let f | forall a. a -> a -> Bool = (==) in f 1 2

error: contract broken by the caller
  ┌─ :1:11
  │
1 │ forall a. a -> a -> Bool
  │           - expected type of the argument provided by the caller
  │
  │
1 │ x
  │ - evaluated to this value
  │
    1. A function `f` is bound by a contract: e.g. `Num -> Num`.
    2. `f` is called with an argument of the wrong type: e.g. `f false`.
  = Either change the contract accordingly, or call `f` with an argument of the right type.
  = Note: this is an illustrative example. The actual error may involve deeper nested functions calls.

note: 
  ┌─ repl-input-0:1:9
  │
1 │ let f | forall a. a -> a -> Bool = (==) in f 1 2
  │         ^^^^^^^^^^^^^^^^^^^^^^^^ bound here

note: 
  ┌─ repl-input-0:1:44
  │
1 │ let f | forall a. a -> a -> Bool = (==) in f 1 2
  │                                            ----- (1) calling f

```
We can see here that the caller is given the blame, but this is not correct. The blame should be on the implementation. In particular, `(==)` does not have the type `forall a. a -> a -> Bool` 

The expected behaviour is to blame the implementation rather than the context; it should also give a *better* error message, stating how to *fix* the implementation. 

This is a sequel of #727

ebresafegaga:
Closes #821

### DIFF
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -334,7 +334,13 @@ where
                     }
                     None | Some(..) => {
                         // This operation should not be allowed to evaluate a sealed term
-                        return Err(EvalError::BlameError(lbl.clone(), call_stack.clone()));
+                        let l = crate::label::Label {
+                            // We are trying to evaluate a sealed term, so the blame
+                            // should be on us because we're not respecting polymorphism.
+                            polarity: true,
+                            ..lbl.clone()
+                        };
+                        return Err(EvalError::BlameError(l, call_stack.clone()));
                     }
                 }
             }
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -334,7 +334,9 @@ where
                     }
                     None | Some(..) => {
                         // This operation should not be allowed to evaluate a sealed term
-                        return Err(EvalError::BlameError(lbl.clone(), call_stack.clone()));
+                        let mut l = lbl.clone();
+                        l.polarity = true;
+                        return Err(EvalError::BlameError(l, call_stack.clone()));
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


yannham (changes requested):
The polarity isn't always positive:

```nickel
let evaluate_incr_at_1 : (forall a. (a -> a) -> a -> a) -> Num = fun ev =>
  ev ((+) 1) 1 in
evaluate_incr_at_1 (fun f x => null)
```

This should blame the caller, not `evaluate_incr_at_1`. The point is, the polarity is not the one of the _occurrence_ of the type variable, but is the polarity of the introducing `forall _`.

yannham (approved):
LGTM. Because the logic is not obvious, a little comment in the code could help our decedent (important information: we set the polarity of the label to `lPol` because this is what's important for blaming polymorphic contracts).

In the future it could be nice to have an even more specialized error message. But that can come in a later PR. In general I've been thinking lately about enriching the interface of `label` in the Nickel language itself, so that custom contracts can set "notes" and additional messages by themselves, instead on relying only on the erase-on-write `msg` field

ebresafegaga:
> In general I've been thinking lately about enriching the interface of `label` in the Nickel language itself, so that custom contracts can set "notes" and additional messages by themselves, instead on relying only on the erase-on-write `msg` field

Interesting. This would probably require adding a new operator to set the "notes" field.

PR #833, opened 2022-09-15
doc: link to typing document

asymmetric:
Adds missing link.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -468,7 +468,7 @@ Examples:
 ## Typing
 
 To give a type to a value, we write it with `< value > : < type >`. More
-information on typing in the relevant document.
+information on typing in the [relevant document](./typing.md).
 
 Examples:
 


yannham (approved):
Thanks! It still needs to be backported to the right branch to be updated on the website. Will do.

asymmetric:
@yannham what's the branch? So I can do it myself for the next cases (which I've already found :sweat_smile:)

yannham:
@asymmetric it's `stable`, but the workflow is a bit constraining, because we also need the change on master. Right now I individually cherry-pick commits on master that update the doc and should be visible on the website as well to backport them on `stable`, then need to update the flake.lock of the website repo, and deploy. Sorry, it just got out of my mind, but it's at least done for this change.

In the general case, unfortunately, I fear you can't do many of these actions by yourself.

asymmetric:
Ok, my takeaway is that it’s better to lump these changes together rather than in many small commits, since the overhead is constant :)

PR #835, opened 2022-09-16
String interner for identifiers

### LINKED ISSUE #774: String interning
# Why?

Nickel terms contain two _kinds_ of strings. The first one corresponds to string literals in source code (and even interpolated strings, through `StrChunk`). The second kind is `Ident`, which is used for both user-defined variables and symbols generated through closurization.

Interning strings would provide the following benefits:
- Reduce allocations and memory usage through de-duplication.
- Make comparisons cheap by using a `Symbol` type such as `u32`.

# Integration in the current pipeline

Ideally, LALRPOP would be supplied an `Interner` data structure, which it will during parsing to generate an AST where `Ident` and `String` are replaced by a `Symbol` type. Subsequent consumers of the AST would have access to the same `Interner` to resolve `Symbol -> &str` as needed.

This implies at least two methods on `Interner`: `intern` and `resolve`. Another operation might be `static_intern`, which would be useful for symbols present in the standard library.

# Solution space

## Simple

The simplest possible interner would look something like this:

```rs
pub struct Interner {
    dedup: HashMap<String, Symbol>,
    backend: Vec<String>,
}
```

The `dedup` HashMap would be checked at every `intern` call to check if we already encountered this string before. If so, we return the corresponding `Symbol` directly. Otherwise, we need to store the string _two_ times. Here the cost of `intern` is the cost of hashing `String` at best. At worst, we add to that two memory allocations.

Resolving a symbol would index through `backend` and return a `&str`. The cost is a single memory access. Easy enough.

It seems that [lalrpop-intern](https://docs.rs/lalrpop-intern/0.15.1/src/lalrpop_intern/lib.rs.html#14-17) actually goes this route. Though I'm not entirely sure if it's actually used in `LALRPOP` itself at the moment.

## Don't Move

This design by [matklad](https://matklad.github.io/2020/03/22/fast-simple-rust-interner.html) avoids redundant allocations by storing a `&'static str` references. How could they be static? This data structure has the additional variant that the contents of `buffer` _are never moved on the heap_, which guarantees that the references in `dedup` and `backend` will be valid for the lifetime of `Interner`. This means that when `buffer` is almost full, it is moved into `buffers_full` and replaced with a new `String` of twice the capacity. 

This technique is called "buckets" and it's present in the `Bucket` backend of the [string-interner](https://docs.rs/string-interner/latest/src/string_interner/backend/bucket/mod.rs.html#23) crate, [string-cache](https://github.com/servo/string-cache) as part of the Servo project, [boa-interner](https://github.com/boa-dev/boa/blob/main/boa_interner/src/lib.rs#L91) used in the Boa language, and the [lasso](https://github.com/Kixiron/lasso/blob/master/src/arenas/single_threaded.rs#L8) single-threaded backend. Something similar is also present in [artichoke](https://github.com/artichoke/intaglio/blob/trunk/src/str.rs#L249), a Ruby implementation.

```rs
pub struct Interner {
    dedup: HashMap<&'static str, Symbol>,
    backend: Vec<&'static str>,
    buffer: String,
    buffers_full: Vec<String>,
}
```

Not only does this solve our issues with the redundant allocation, but it also reduces overall allocations because each new bucket's initial capacity is twice that of the previous one. So we enjoy periods where `intern`'s cost is hashing plus cloning, separated by moments of (potentially) sizable allocations. Overall, the downside of this approach seems to be _the long_ lifetime of the strings. Nickel however shouldn't suffer from this.

## Fancy Raw Entries

This technique does a bit of (nightly) magic with `HashMap`, and looks like this:

```rs
pub struct Interner<H = RandomState> {
    dedup: HashMap<Symbol, (), ()>,
    hasher: H,
   // .. cropped
}
```

What happens here is that the `HashMap` itself has no hashing algorithm, instead, we insert `Symbol` _with the hash of its corresponding string_ so that a subsequent de-duplication lookup would use the Raw Entry API in `std` to retrieve the `Symbol`  using the hash of it given string. Almost like a `HashSet` where we get to control the hash of all entries, this experimental API however is nightly-only not available on `HashSet`, though it is available in the `hashbrown` crate.

I found that this commonly used, for example in [string-interner](https://github.com/Robbepop/string-interner/blob/master/src/interner.rs#L51), [strena](https://github.com/CAD97/strena/blob/main/src/lib.rs#L175), and [lasso](https://github.com/Kixiron/lasso/blob/master/src/rodeo.rs#L41).

## One Giant Buffer

In this setup, everything is stored in one `String`. So since our buffer can, and will, move several times, we can't use references like in the buckets approach. Moreover, since we have to use `Spans` in this case, we add an extra level of indirection when resolving symbols, since we now make the extra step of looking up the `end` of each string. Resolving now is two memory accesses instead of one. And without some `unsafe`, we would have to check that our span is valid UTF-8.

```rs
pub struct Interner<H = RandomState> {
    dedup: HashMap<Symbol, (), ()>,
    hasher: H,
    ends: Vec<usize>,
    buffer: String,
}
```

This can be seen in the `string` backend of [string-interner](https://github.com/Robbepop/string-interner/blob/master/src/backend/string.rs#L50) and the [strena](https://github.com/CAD97/strena/blob/main/src/lib.rs#L171) crates.

## Honorable Mentions

- [ustr](https://github.com/anderslanglands/ustr) is a global, immutable, static (everything is leaked) string cache.
- [smol-str](https://github.com/rust-analyzer/smol_str) is a string type with `O(1)` cloning and is stack-allocated when less than 22 bytes. It's part of rust-analyzer.

# Conclusion

The solutions outlined above come with different trade-offs in terms of allocations and speed of both resolving and interning. What's left now is to carve out a special solution for Nickel based on our specific requirements.

Acaccia:
This PR implements a String interner for Nickel identifiers.

As of now, this is a naive implementation, with the interner being a simple `Hashmap<String, Arc<String>>`.

Most of the benchmarks showed a performance improvement, while a few others showed severe regression (like the "round_trip" benchmark).
[criterion.zip](https://github.com/tweag/nickel/files/9585128/criterion.zip)
To open the benchmark results, decompress the archive and open in your browser `report/index.html`.

From the flamegraphs comparison with master, it seems the insertion in the interner is quite heavy (top is master, bottom is with string-interner).
flamegraph-master
flamegraph-interner

From here, we could try using different implementations to make the string interner more useful. For example, as stated in #774, using `u32` symbols could make comparison cheaper, and would allow the `Ident` type to implement `Copy`. Changing the `Interner` struct to use it should be trivial.

The crate [string-interner](https://crates.io/crates/string-interner) possesses a few different implementations that might be worth a try.

Closes #774.

### DIFF
--- a/src/deserialize.rs
+++ b/src/deserialize.rs
@@ -70,7 +70,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
             Term::Num(v) => visitor.visit_f64(v),
             Term::Str(v) => visitor.visit_string(v),
             Term::Enum(v) => visitor.visit_enum(EnumDeserializer {
-                variant: v.label,
+                variant: v.into(),
                 rich_term: None,
             }),
             Term::Record(v, _) => visit_record(v, visitor),
@@ -117,7 +117,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
         V: Visitor<'de>,
     {
         let (variant, rich_term) = match unwrap_term(self)? {
-            Term::Enum(ident) => (ident.label, None),
+            Term::Enum(ident) => (ident.into(), None),
             Term::Record(v, _) => {
                 let mut iter = v.into_iter();
                 let (variant, value) = match iter.next() {
@@ -135,7 +135,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
                         occurred: "Record with multiple keys".to_string(),
                     });
                 }
-                (variant.label, Some(value))
+                (variant.into(), Some(value))
             }
             other => {
                 return Err(RustDeserializationError::InvalidType {
@@ -429,7 +429,7 @@ impl<'de> MapAccess<'de> for RecordDeserializer {
         match self.iter.next() {
             Some((key, value)) => {
                 self.rich_term = Some(value);
-                seed.deserialize(key.label.into_deserializer()).map(Some)
+                seed.deserialize(key.label().into_deserializer()).map(Some)
             }
             None => Ok(None),
         }
--- a/src/deserialize.rs
+++ b/src/deserialize.rs
@@ -70,7 +70,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
             Term::Num(v) => visitor.visit_f64(v),
             Term::Str(v) => visitor.visit_string(v),
             Term::Enum(v) => visitor.visit_enum(EnumDeserializer {
-                variant: v.into(),
+                variant: v.into_label(),
                 rich_term: None,
             }),
             Term::Record(v, _) => visit_record(v, visitor),
@@ -117,7 +117,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
         V: Visitor<'de>,
     {
         let (variant, rich_term) = match unwrap_term(self)? {
-            Term::Enum(ident) => (ident.into(), None),
+            Term::Enum(ident) => (ident.into_label(), None),
             Term::Record(v, _) => {
                 let mut iter = v.into_iter();
                 let (variant, value) = match iter.next() {
@@ -135,7 +135,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
                         occurred: "Record with multiple keys".to_string(),
                     });
                 }
-                (variant.into(), Some(value))
+                (variant.into_label(), Some(value))
             }
             other => {
                 return Err(RustDeserializationError::InvalidType {
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -33,6 +33,10 @@ impl Ident {
     pub fn label(&self) -> &str {
         INTERNER.lookup(self.symbol)
     }
+
+    pub fn into_label(self) -> String {
+        self.label().to_owned()
+    }
 }
 
 /// Special character used for generating fresh identifiers. It must be syntactically impossible to
@@ -82,7 +86,7 @@ where
 
 impl Into<String> for Ident {
     fn into(self) -> String {
-        self.as_ref().to_owned()
+        self.into_label()
     }
 }
 
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -189,4 +189,34 @@ mod interner {
             self.vec[sym.0 as usize]
         }
     }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        #[test]
+        fn test_intern_then_lookup() {
+            let interner = Interner::new();
+            let test_string = "test_string";
+            let sym = interner.intern(test_string);
+            assert_eq!(interner.lookup(sym), test_string);
+        }
+
+        #[test]
+        fn test_intern_twice_has_same_symbol() {
+            let interner = Interner::new();
+            let test_string = "test_string";
+            let sym1 = interner.intern(test_string);
+            let sym2 = interner.intern(test_string);
+            assert_eq!(sym1, sym2);
+        }
+
+        #[test]
+        fn test_intern_two_different_has_different_symbols() {
+            let interner = Interner::new();
+            let sym1 = interner.intern("a");
+            let sym2 = interner.intern("b");
+            assert_ne!(sym1, sym2);
+        }
+    }
 }
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -108,20 +108,30 @@ mod interner {
 
     use typed_arena::Arena;
 
+    /// A symbol is a correspondance between an [Ident](super::Ident) and its string representation stored in the [Interner].
     #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
     pub struct Symbol(u32);
 
+    /// The interner, which serves a double purpose: it pre-allocates space
+    /// so that [Ident](super::Ident) labels are created faster
+    /// and it makes it so that labels are stored only once, saving space.
     pub(crate) struct Interner<'a>(RwLock<InnerInterner<'a>>);
 
     impl<'a> Interner<'a> {
+        /// Creates an empty [Interner].
         pub(crate) fn new() -> Self {
             Self(RwLock::new(InnerInterner::new()))
         }
 
+        /// Stores a string inside the [Interner] if it does not exists, and returns the corresponding [Symbol].
         pub(crate) fn intern(&self, string: impl AsRef<str>) -> Symbol {
             self.0.write().unwrap().intern(string)
         }
 
+        /// Looks up for the stored string corresponding to the [Symbol].
+        ///
+        /// This operation cannot fails since the only way to have a [Symbol] is to have [interned](Interner::intern)
+        /// the corresponding string first.
         pub(crate) fn lookup(&self, sym: Symbol) -> &str {
             // SAFETY: We are making the returned &str lifetime the same as our struct,
             // which is okay here since the InnerInterner uses a typed_arena which prevents
@@ -131,6 +141,12 @@ mod interner {
         }
     }
 
+    /// The main part of the Interner.
+    ///
+    /// It is made out of 3 parts:
+    /// - the arena, that preallocates space where strings are stored
+    /// - the map, which prevents the arena from creating different [Symbols](Symbol) for the same string,
+    /// - the vec, which allows for retrieving a string from a [Symbol].
     struct InnerInterner<'a> {
         arena: Mutex<Arena<u8>>,
         map: HashMap<&'a str, Symbol>,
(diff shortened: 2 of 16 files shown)


src/identifier.rs
```diff
@@ -1,14 +1,42 @@
...
-    pub label: String,
-    pub pos: TermPos,
+    label: interner::Symbol,
+    pos: TermPos,
+    generated: bool,
+}
+
+impl Ident {
+    pub fn new_with_pos(label: impl AsRef<str>, pos: TermPos) -> Self {
+        let generated = label.as_ref().starts_with(GEN_PREFIX);
```
yannham, line 21: I wonder if we couldn't do something a bit cleaner here, by avoiding allocating a string at all for generated symbols, by:

- having a method to print a symbol as a number (`Symbol::to_string()`)
- move the method to create a fresh identifiers to `Ident`, which would set `generated` to `true` in this case
- if `label()` is called, just return `format!("{}{}", GEN_PREFIX, symbol`.

I can see that the last one may be problematic because `label` return a `&str`. But at least we can move the code to create a fresh generated identifier into `Ident` and handle `generated` in a less fragile way (rather than relying on the presence of some character).

src/identifier.rs
```diff
@@ -1,14 +1,42 @@
...
+
+    pub fn new(label: impl AsRef<str>) -> Self {
+        Self::new_with_pos(label, TermPos::None)
+    }
+
+    pub fn label(&self) -> &str {
+        INTERNER.lookup(self.label)
+    }
+
+    pub fn pos(&self) -> TermPos {
```
yannham, line 37: Is there a good reason not to make `pos` public? For `symbol` it makes sense, but `pos` can be freely modified without consequences. I would tend to avoid adding a boilerplate `pos()` function, changing all the code that uses `.pos` somewhere, without a motivation to not use the language's accessor (random example: [codespan's diagnostic has all fields public](https://docs.rs/codespan-reporting/0.11.1/codespan_reporting/diagnostic/struct.Diagnostic.html))

src/deserialize.rs
```diff
@@ -70,7 +70,7 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
             Term::Num(v) => visitor.visit_f64(v),
             Term::Str(v) => visitor.visit_string(v),
             Term::Enum(v) => visitor.visit_enum(EnumDeserializer {
-                variant: v.label,
+                variant: v.into(),
```
yannham, line 73: Why not replacing this with `v.label()` ? I think it's more explicit than `into()`, which depends on the target type expected.

yannham (commented):
I wonder why we need `Interner`, `InnerInterner`, a `RwLock` and a `Mutex`, as the interpreter is currently fully sequential? If this is the case, I think we should just drop all of that altogether: it has a cost, it makes the code more complex and it's not yet clear if we will need it. Maybe a future parallel implementation will have one interner per thread, for example.

It seems `Arena` is implementing `!Sync`, so an interner with an arena inside inherits this constraint and prevent any unsound usage. The day we need the interner to be `Sync`, we can ask ourselves this question again :slightly_smiling_face:

src/identifier.rs
```diff
@@ -1,14 +1,42 @@
...
 
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+static INTERNER: Lazy<interner::Interner> = Lazy::new(interner::Interner::new);
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
 #[serde(into = "String", from = "String")]
 pub struct Ident {
-    pub label: String,
-    pub pos: TermPos,
+    label: interner::Symbol,
```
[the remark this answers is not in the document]
Acaccia: oof, yes completely, changing that

src/identifier.rs
Acaccia: Not sure this is relative to this PR, but it could be a nice improvement, definitely :)

src/identifier.rs
Acaccia: Okay, I'll answer to that with another question: does it make any sense to be able to modify `pos`? Using an access method makes it clear that we can only read the field. If it makes sense to be able to modify this field, then it should be reverted.
(I don't see the connection with `codespan`...)

Acaccia:
@yannham For `once_cell` to allow for the creation of a static variable, we need for our struct to be Send and Sync. Here I use a Mutex on the `InnerInterner`'s arena to make it Send and Sync, so that the Interner using a RwLock would be Send and Sync too. I choose this implementation, so that a Mutex is only active if we need to intern a string inside the arena that does not exists yet. Like this, we don't have to fully lock the Interner for each and any operation.


> Maybe a future parallel implementation will have one interner per thread, for example.

I tried replacing the current implementation with a thread_local + RefCell, it came with a global performance regression.
After that I tried using a global mutex, with `once_cell` and `lazy_static`, both solutions were also slower according to our current benchmarks.

Hence, after our chat on Slack, we decided to keep the current implementation.

Acaccia:
Here is a new benchmark comparison between the commit from which this branch came (793d3d214b58739ee7d1c1c1c3c546c66f337c76) and this branch. We can see that we have a performance increase for nearly all benchmarks of ~15-20%.
[criterion-arena.zip](https://github.com/tweag/nickel/files/9685803/criterion-arena.zip)

yannham:
>  For once_cell to allow for the creation of a static variable, we need for our struct to be Send and Sync. 

Indeed, I totally missed that.

> I tried replacing the current implementation with a thread_local + RefCell, it came with a global performance regression.
After that I tried using a global mutex, with once_cell and lazy_static, both solutions were also slower according to our current benchmarks.

And this has been quite surprising to me :open_mouth: (is thread local storage so costly?) , but the benchmarks have spoken! Let's go with your initial approach, which will also allow synchronization for free if we go multi-threaded and decide to have one shared interner.

src/identifier.rs
yannham: The connection with codespan is the following: they have the same choice to make for `Diagnostic`. They could go with method accessors instead, but they didn't. This is just one datapoint.

I tend to see types like Ident etc. as describing data. You should be able to manipulate data freely, as long as it doesn't have any bad consequences. If no methods and no other part of the code rely on the fact that `pos` is immutable, then I don't see a reason to impose this constraint, which moreover needs this additional accessor. For example, the `pos` of RichTerm is not a method, and we often modifes the position afterward, because we build them in several steps (granted, we usually use `with_pos`, but just because it's nicer to chain, and not always: the error handling code does modify the position of a rich term directly because it makes the overall code simpler, see e.g. https://github.com/tweag/nickel/blob/54bfa6eb3074305ab289cec98fe2d6bedd1fdaf2/src/error.rs#L878). On paper, we could also say that it doesn't make sense to modify the position of a `RichTerm`, but in practice it can come in handy.

It also makes things inconsistent with the rest of the codebase (for `RichTerm` you use `.pos`, for `Label` you use `.arg_pos`, but for `Ident` you use a function call `.pos()`).

Acaccia: okay then, I'm changing this back

src/deserialize.rs
Acaccia: I added a method in Ident `into_label`, to make conversions clearer than using plain `into`

yannham (commented):
Rereading the unsafe parts, I'm not convinced the interner offers a safe interface. Because `'a` is free in the implementation, what stops me from doing that?

```
fn main() {
  let dangling : &'static str = get_static_dangling_reference();
  println!("{}", dangling);
}

fn get_static_dangling_reference() : &'static str {
  let interner = Interner::<'static>::new();
  let ref : &'static str = interner.intern("oh, no, UB :(");
  ref // interner will be dropped, returning a dangling ref
}
```

From the top of my head, I can see two solutions:

- say well, our usage of the interner is safe, but it needs to respect some invariant (that `'self` outlives `'a`). Then we have to mark more of the interner implementation unsafe, probably at least `intern`
- Do some more complex stuff with a self referential struct that holds a phantom type `&'a Interner` or something that guarantees that `'a` can't outlive `arena`

Acaccia:
@yannham Testing your example with correct Rust syntax:
```rust
fn main() {
    let dangling: &'static str = get_static_dangling_reference();
    println!("{}", dangling);
}

fn get_static_dangling_reference() -> &'static str {
    let interner = interner::Interner::<'static>::new();
    let sym = interner.intern("oh, no, UB :(");
    let refer: &'static str = interner.lookup(sym);
    refer // interner will be dropped, returning a dangling ref
}
```

Rust gives the error:
```
let refer: &'static str = interner.lookup(sym);
           ------------   ^^^^^^^^^^^^^^^^^^^^ borrowed value does not live long enough
           |
           type annotation requires that `interner` is borrowed for `'static`
refer
}
- `interner` dropped here while still borrowed
```

Thus the borrow checker won't allow it.

yannham:
You're right, the lifetime of the returned value is constrained in `lookup` actually. My bad

yannham (approved):

PR #839, opened 2022-09-19
Maintain a term environment during an REPL session

yannham:
Follow up of #766 and #834. Correctly check for contract equality during an REPL session, by providing an initial context (typing environment + term environment), instead of just a typing environment, to various typechecking.

Sadly, we don't really use the machinery introduced in #834, because that would require to use `GenericTypeWrapper<E>` instead of `TypeWrapper` in a lot of places inside the `typecheck` module, which is a verbose and heavy change. Thus we somehow maintain two environments which contain the same information but in slightly different form (`eval_env` and `term_env`), but the overhead is expected to be insignificant and it lets the meat of the `typecheck` module untouched.

Depends on #834

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -91,17 +91,21 @@ pub enum ErrorTolerance {
 pub struct Envs {
     /// The eval environment.
     pub eval_env: eval::Environment,
-    /// The typing environment, counterpart of the eval environment for typechecking. Entries are
-    /// [`crate::typecheck::TypeWrapper`] for the ease of interacting with the typechecker, but
-    /// there are not any unification variable in it.
-    pub type_env: typecheck::Environment,
+    /// The typing context.
+    ///
+    /// The typing context includes the typing environment, counterpart of the eval environment for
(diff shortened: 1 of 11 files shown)


src/cache.rs
```diff
@@ -91,17 +91,21 @@ pub enum ErrorTolerance {
...
-    /// [`crate::typecheck::TypeWrapper`] for the ease of interacting with the typechecker, but
-    /// there are not any unification variable in it.
-    pub type_env: typecheck::Environment,
+    /// The typing context.
+    ///
+    /// The typing context includes the typing environment, counterpart of the eval environment for
+    /// typechecking. Entries are of type [`crate::typecheck::TypeWrapper`] for the ease of
+    /// interacting with the typechecker, but there are not any unification variable in it.
+    ///
+    /// The context also includes the term environment, used to check for contract equality.
```
Acaccia, line 100: Why is the documentation for `Context` here instead of near its definition?

Acaccia (changes requested):

Acaccia (approved):

PR #845, opened 2022-09-22
Recursive priorities (or push-priorities, or leafy priorities)

### LINKED ISSUE #279: "Leafy" defaults

yannham:
Closes #279. Implement push priority operators as described in RFC001. The general idea is that:

```
let foo = {
  bar = 1,
  baz = {
    x = "a",
    y = "b",
  }
} in
foo | _push_default
```

is equivalent to

```
{
  bar | default = 1,
  baz = {
    x | default = "a",
    y | default = "b",
  }
}
```

The push priority operators are dynamic and not just syntactic sugar. They operate on arbitrary expressions and are thus not limited to record literals. The syntax `_push_default` and `_push_force` is temporary (and ugly on purpose), to proceed with the technical implementation without being blocked by bikeshedding considerations, which can happen after, prior to the next release.


src/parser/utils.rs
```diff
@@ -100,6 +104,114 @@ impl InfixOp {
...
+/// (ex: `push force`/`push default`). Those are collected in this extended annotation and then
+/// desugared into a standard metavalue.
+#[derive(Clone, Debug, Default)]
+pub struct ExtdAnnot {
+    /// Standard metadata.
+    pub meta: MetaValue,
+    /// Presence of an annotation `push force`
+    pub push_force: bool,
+    /// Presence of an annotation `push default`
+    pub push_default: bool,
```
matthew-healy, line 167: Does it ever make sense for a value to be annotated with both `_push_force` and `_push_default`? If not, I'm wondering if it could make more sense to model this in a way where it's not possible for both to be `true` at the same time, e.g.

```rust
pub push_prio: Option<PushPriority>
```

matthew-healy (commented):
I think this looks good, with the obvious caveat that I'm still very new to the codebase. I left some small comments, but neither needs to block the PR.

matthew-healy (approved):

dpl0a (commented):
Looks ok to me. I second matthew's comments though.

src/parser/utils.rs
yannham: It's indeed fishy semantically, but the thing is that nothing currently prevent the user from doing so:

```
x | _push_default | _push_force = ...
```

because annotations are just parsed one by one and combined. In some sense, you can see `ExtdAnnot` as an AST for meta-annotation, and the source syntax does allow both to be set, just as they were flags. One it's actually transformed into a term by `attach()`, it does have to chose an actual `push_prio` to use.

PR #846, opened 2022-09-26
Implement "occurs check" to prevent infinite loop during type-checking

### LINKED ISSUE #277: Infinite loop during typechecking (recursive types)
**Describe the bug**
Following #270, which has been fixed for usual cases, some terms still make the typechecker loop. The occurs check is not implemented, hence the inference of terms introducing recursive equations crashes.

**To Reproduce**
```
nickel typecheck <<< '{g = g 0}.g : Num'

```

**Expected behavior**
The typechecker should reject such terms without crashing.

ebresafegaga:
### Summary
Currently, the Nickel type-checker does not perform "occurs check" when unifying a type variable with another type . Occurs checks normally prevents us from creating "infinite" type substitutions. For example, if we want to unify a type variable `a` with another type 
`a -> b`, occurs checks prevents `a = a -> b` from being a type substitution equation. 

In Nickel, this means that programs which result in a type variable unifying with another type which includes that variable results in a stack overflow.

### Example:
```
let f :  _ = fun x => x x in f 
```
```
let f :  _ =  fun esv es e => e esv e (es e) in f
```
```
{g = g 0}.g : Num
```

### Actual output:
```
```

### Desired output:

Nickel should give a type error, stating the exact problem in a friendly way; and maybe even "tips" on how to resolve it.
```
nickel> let f :  _ = fun x => x x 

The expression x has the inferred type a -> b. 
x cannot be the argument of x, because this will lead to the construction of the infinite type a = a -> b.
Maybe try using concrete types to guide your implementation instead of wildcards?
```

The error message generation is not *yet* like as I've described above.

Closes #277

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -1314,60 +1314,43 @@ fn row_add(
     }
 }
 
-/// Get the set of all unification variables in a type.
-fn get_free_unif_vars(wildcards: &Vec<TypeWrapper>, t: TypeWrapper) -> HashSet<usize> {
-    let mut set: HashSet<usize> = HashSet::new();
-    fn get_free_unif_vars_inner(
-        set: &mut HashSet<usize>,
-        wildcards: &Vec<TypeWrapper>,
-        t: TypeWrapper,
-    ) {
-        match t {
-            TypeWrapper::Ptr(i) => {
-                set.insert(i);
-                ()
-            }
-            TypeWrapper::Contract(..) | TypeWrapper::Constant(_) => (),
-            TypeWrapper::Concrete(ty) => match ty {
-                AbsType::Dyn()
-                | AbsType::Num()
-                | AbsType::Bool()
-                | AbsType::Str()
-                | AbsType::Sym()
-                | AbsType::Flat(_)
-                | AbsType::Var(_)
-                | AbsType::RowEmpty() => (),
-                AbsType::Wildcard(id) => wildcards
-                    .get(id)
-                    .map(|ty| get_free_unif_vars_inner(set, wildcards, ty.clone()))
-                    .unwrap_or(()),
-                AbsType::Arrow(ty1, ty2) => {
-                    get_free_unif_vars_inner(set, wildcards, *ty1);
-                    get_free_unif_vars_inner(set, wildcards, *ty2)
-                }
-                AbsType::Forall(_, ty)
-                | AbsType::Enum(ty)
-                | AbsType::StaticRecord(ty)
-                | AbsType::DynRecord(ty)
-                | AbsType::Array(ty) => get_free_unif_vars_inner(set, wildcards, *ty),
-                AbsType::RowExtend(_, ty_row, tail) => {
-                    ty_row
-                        .map(|ty| get_free_unif_vars_inner(set, wildcards, *ty))
-                        .unwrap_or(());
-                    get_free_unif_vars_inner(set, wildcards, *tail)
-                }
-            },
-        }
-    }
-    get_free_unif_vars_inner(&mut set, wildcards, t);
-    set
-}
-
-/// Check if the type variable `ptr` occurs in `t`
+/// Checks if the "root type" of the unification variable `ptr` is also 
+/// the root type of any unification variable in `t`
 fn occurs_check(state: &State, ptr: usize, t: TypeWrapper) -> bool {
-    get_free_unif_vars(state.wildcard_vars, t)
-        .iter()
-        .any(|other_ptr| state.table.root(ptr) == state.table.root(*other_ptr))
+    match t {
+        TypeWrapper::Ptr(i) if i == ptr => state.table.root(i) == state.table.root(ptr),
+        TypeWrapper::Ptr(_) => false,
+        TypeWrapper::Contract(..) | TypeWrapper::Constant(_) => false,
+        TypeWrapper::Concrete(ty) => match ty {
+            AbsType::Dyn()
+            | AbsType::Num()
+            | AbsType::Bool()
+            | AbsType::Str()
+            | AbsType::Sym()
+            | AbsType::Flat(_)
+            | AbsType::Var(_)
+            | AbsType::RowEmpty() => false,
+            AbsType::Wildcard(id) => state
+                .wildcard_vars
+                .get(id)
+                .map(|ty| occurs_check(state, ptr, ty.clone()))
+                .unwrap_or(false),
+            AbsType::Arrow(ty1, ty2) => {
+                occurs_check(state, ptr, *ty1) || occurs_check(state, ptr, *ty2)
+            }
+            AbsType::Forall(_, ty)
+            | AbsType::Enum(ty)
+            | AbsType::StaticRecord(ty)
+            | AbsType::DynRecord(ty)
+            | AbsType::Array(ty) => occurs_check(state, ptr, *ty),
+            AbsType::RowExtend(_, ty_row, tail) => {
+                ty_row
+                    .map(|ty| occurs_check(state, ptr, *ty))
+                    .unwrap_or(false)
+                    || occurs_check(state, ptr, *tail)
+            }
+        },
+    }
 }
 
 /// Try to unify two types.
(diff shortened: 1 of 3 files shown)


src/typecheck/mod.rs
```diff
@@ -1314,6 +1314,52 @@ fn row_add(
...
+                .union(&get_free_variables(wildcards, *tail))
+                .map(|i| *i)
+                .collect(),
+        },
+        TypeWrapper::Contract(..) | TypeWrapper::Constant(_) => HashSet::new(),
+    }
+}
+
+/// Check if the type variable `ptr` occurs in `t`
+fn occurs_check(ptr: usize, state: &State, t: TypeWrapper) -> bool {
```
matthew-healy, line 1357: Nitpick: from what I can see it seems like the pattern elsewhere has been to pass the `state` as the first argument. It possibly makes sense to continue that here?

src/typecheck/mod.rs
```diff
@@ -1314,6 +1314,52 @@ fn row_add(
...
+        TypeWrapper::Concrete(ty) => match ty {
+            AbsType::Dyn()
+            | AbsType::Num()
+            | AbsType::Bool()
+            | AbsType::Str()
+            | AbsType::Sym()
+            | AbsType::Flat(_)
+            | AbsType::Var(_)
+            | AbsType::RowEmpty() => HashSet::new(),
+            // TODO: replace with get_wilccard_var(..)
```
matthew-healy, line 1330: Is there a reason this hasn't been done as part of this PR? From what I can tell, calling `get_wildcard_var` here could change the behaviour of this function (by returning any new unification variables created), but I'm struggling to work out whether that would have any impact on `occurs_check`.

matthew-healy (commented):
The change looks good to me, though I've only been looking at the Nickel codebase for a few days and I'm mainly reviewing to get some more exposure to the codebase. :grin: 

Do you think it might make sense to add test cases using the example from the PR description (e.g. in `tests/typecheck_fail.rs`) to prevent future regressions?

src/typecheck/mod.rs
```diff
@@ -1314,6 +1314,52 @@ fn row_add(
     }
 }
 
+/// Get the set of all unification variables in a type.
+fn get_free_variables(wildcards: &Vec<TypeWrapper>, t: TypeWrapper) -> HashSet<usize> {
+    match t {
+        TypeWrapper::Ptr(i) => HashSet::from([i]),
```
yannham, line 1320: I don't think this is sound. If `?i` has been unified with another type with variables inside before, for example `?a -> ?b`, then the free unification variablrs of `?i` shouldn't be `[?i]`, but `[?a,?b]` (if they are fresh, not-yet-unified variables). So, you should follow (and ignore with respect to occur checks, as if the checked variable appears here there is something wrong with the typechecking algorithm) unification variables that point to something in the unification table.

src/typecheck/mod.rs
yannham: I think both version are correct, because `get_wildcard_vars` may only create new fresh unification variables, which don't matter (as they will never fail the occurs check). Somehow the current implementation is lazier, as it doesn't generate the wildcard vars yet, but will the next time they are actually needed by unification. It probably doesn't make any noticeable difference, and if anything, I expect this version to be slightly better, as it doesn't pollute the hashset with new fresh unification variables.

yannham (commented):
It may be interesting to run the benchmarks on typechecking the nix list lib to see if that affects performance. My occurs check knowledge is rusty but I remember that running it without any form of memoization like this can turn out to be expensive. For the record, although this is not the main subject of the article, [efficient and insightful generalization](https://okmij.org/ftp/ML/generalization.html) also mentions how doing it in a "lazy" way (and not at each unification).

I remember why this issue has been open for quite some time, now. I think my original plan was to implement path compression first (i.e., when you call `get_root`, you remember all the unification variable on the way, like if there is a link `?a -> ?b -> ?c -> ?d -> Num` and at the end makes the table look like `?a -> Num, ?b -> Num, ?c -> Num, ?d -> Num`). Doing so, we can avoid the occurs check at unification, and perform it only when calling to `get_root`, potentially grouping several traversal into one. This probably requires to remember for each unification variable if the check has been done or not.

But to implement path compression efficiently, you want `TyperWrapper` to use `Rc`, not bare `Box`, otherwise you will deep-copy your types a lot. But this hasn't happened yet :confused: 

We can first see what happens on the benchmark: if the numbers are reasonable, this can be a first implementation, and then we can work toward shareable types (Rc) + path compression + occurs check.

src/typecheck/mod.rs
```diff
@@ -1314,6 +1314,52 @@ fn row_add(
     }
 }
 
+/// Get the set of all unification variables in a type.
+fn get_free_variables(wildcards: &Vec<TypeWrapper>, t: TypeWrapper) -> HashSet<usize> {
```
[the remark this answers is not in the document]
ebresafegaga: Yes, I initially wanted to do that. But I thought I'd just write this once, because there might be other places which might need to get all the free unification variables.

src/typecheck/mod.rs
ebresafegaga: I think if `?i` points to `?a -> ?b`, then `?a -> ?b` will be the root of this set because it is no longer a variable: it now a ` TypeWrapper::Concrete(AbsType::Arrow(?a, ?b))`

This is why I call root when checking if two unification variables are the same:
```rust 
fn occurs_check(state: &State, ptr: usize, t: TypeWrapper) -> bool {
    get_free_unif_vars(state.wildcard_vars, t)
        .iter()
        .any(|other_ptr| state.table.root(ptr) == state.table.root(*other_ptr))
}
```

src/typecheck/mod.rs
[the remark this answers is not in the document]
ebresafegaga: I think I'd rather create an inner function that threads a mutable argument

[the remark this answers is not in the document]
yannham: Do you have anything in mind in particular that would need the free unification variables? If yes, then go. Otherwise, I think you probably aren't gonna need it :slightly_smiling_face: and I would implement the hashset-free version. If really we were wrong and need this later, this isn't very hard to move to the hashset version.

src/typecheck/mod.rs
yannham: It's still not enough, because if `?a` points to `?c -> ?d`, and `?b` is not yet unified, you should have `?c, ?d, ?b` as free variables. Which your `root` call isn't sufficient to catch (it only does one unrolling), if I'm not mistaken.

Even if it would be sufficient, the function is still unsound, because it requires the caller to do something additional which is far from obvious. If it's called `get_free_variables`, so it should return the free unification variables for the standard definition of free unification variables. If someone else comes and refactor the code, something could go wrong. Otherwise, it deserves another name, like impartial, lazy, on_step, etc. and be precisely documented.

src/typecheck/mod.rs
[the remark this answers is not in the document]
ebresafegaga: I don't think I have, I'll go with the check now

src/typecheck/mod.rs
ebresafegaga: Yes. I guess this is a another reason to just use a check, instead of generating an intermediate set of free unification variables.

yannham: I'm not sure I understand why this issue and hashset vs simple check are related. In both cases, I think you will have to potentially perform arbitrary-deep unrolling of free variables?

ebresafegaga: What does "arbitrary-deep unrolling of free variables" mean? Also, see the latest commit to see what I mean by using a simple check instead.

yannham:
@ebresafegaga Could you please report the result of the benchmarks here, to document the decision with respect to this PR? Thanks!

ebresafegaga:
On average, we're seeing a 68% regression in performance while type-checking ` typecheck-nixpkgs-lib`. 

Here's an example of such benchmark: 
```
nixpkgs lists           time:   [2.8098 ms 2.8272 ms 2.8450 ms]
                        change: [+79.283% +81.642% +83.967%] (p = 0.00 < 0.05)
                        Performance has regressed.

```

yannham:
We are cheating by not doing the occurs check right now, so a hit is to be expected, but 80% doesn't sound acceptable. I propose to close this PR (but keep the branch somewhere, or maybe just commit the checking function somewhere) and try to do something a bit lazier/smarter

PR #849, opened 2022-10-03
Deprecate `fresh_var` for a new method in Ident

Acaccia:
Depends on #835 

The concept came from a comment from @yannham:

	I wonder if we couldn't do something a bit cleaner here, by avoiding allocating a string at all for generated symbols, by:
		- having a method to print a symbol as a number (Symbol::to_string())
	    - move the method to create a fresh identifiers to Ident, which would set generated to true in this case
	    - if label() is called, just return format!("{}{}", GEN_PREFIX, symbol.
	I can see that the last one may be problematic because label return a &str. But at least we can move the code to create a fresh generated identifier into Ident and handle generated in a less fragile way (rather than relying on the presence of some character).

Here, we deprecated the method `fresh_var`, responsible for creating new `Ident` with a generated label, and created a method directly inside `Ident`.

### DIFF
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -31,7 +31,7 @@ impl Ident {
         Self::new_with_pos(label, TermPos::None)
     }
 
-    pub fn generate() -> Self {
+    pub fn fresh() -> Self {
         Self::new(format!("{}{}", GEN_PREFIX, GeneratedCounter::next()))
     }
 
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -6,6 +6,7 @@ use std::hash::Hash;
 
 use crate::position::TermPos;
 
+simple_counter::generate_counter!(GeneratedCounter, usize);
 static INTERNER: Lazy<interner::Interner> = Lazy::new(interner::Interner::new);
 
 #[derive(Debug, Clone, Copy, Deserialize, Serialize)]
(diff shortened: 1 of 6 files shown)


src/transform/mod.rs
```diff
@@ -66,11 +62,10 @@ pub fn transform_no_free_vars(
         .unwrap())
 }
 
+#[deprecated = "use [Ident::generate] instead"]
```
yannham, line 65: That's clean, and the right thing to do once we offer stability guarantees. But as we are in version 0.2.1 < 1.0.0, hence I think we might get away with just deleting this function altogether.

Acaccia: Yep, I was wondering what would be our process here. Okay, deleting it then :D

src/identifier.rs
```diff
@@ -30,6 +31,10 @@ impl Ident {
         Self::new_with_pos(label, TermPos::None)
     }
 
+    pub fn generate() -> Self {
```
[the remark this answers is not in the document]
Acaccia: Sure, that works for me

yannham (approved):
Thanks! I don't know why netlify deploy action is running, and the error message is not very informative. Maybe just re-running the action will be sufficient?

yannham:
It's not. The error seems to be related to action access rights, which sound totally orthogonal: let's ignore for now, as it won't have any impact on the website as long as we don't backport anything in the `stable` branch.

PR #854, opened 2022-10-04
[Fix] Type to term conversion

### LINKED ISSUE #758: Undue "unbound type variable" when alternating types and contracts
**Describe the bug**
In some cases the parser throw undue unbound type variables errors:

1. When using a contract as the parameter of a type, itself part of a contract, the parser throws an undue unbound type variable error
2. When using a contract as the parameter of a type outside of an annotation

**To Reproduce**

To trigger the  bug, when need a parametrized type (`Array`), which parameter is a variable (`B`), and this type must istelf be part of a contract (`A _`) (Case 1.):

```nickel
nickel> let A = fun param label value => value in
let B = fun label value => value in
[1] | A (Array B)
error: unbound type variable(s): `B`
  ┌─ repl-input-0:3:16
  │
3 │ [1] | A (Array B)
  │                ^
```

When there's no annotation, for example if we store the contract inside a variable, the outer layer is not even needed: only one alternation between a type (`Array`) and a contract (`B`) is enough (Case 2.):

```nickel
nickel>  let B = fun label value => value in
let Contract = Array B in
null
error: unbound type variable(s): `B`
  ┌─ repl-input-1:2:22
  │
2 │ let Contract = Array B in
  │                      ^
```

**Expected behavior**
`B` should be interpreted as a variable referring to the corresponding contract, and not as a type variable. Both programs should run fine.

**Environment**
 - Version of the code: 34b3b8ca8b867912e7abaf579f838a9f8777d721

**Additional context**
This is due to the way the type variable fix up phase is implemented. It is not correct outside of annotations, and it is not doing the right thing on flat types.

francois-caddet:
fix #758

Does not break any tests. But has to be check deeper to besure it does not cause unexpected type var to term var transformation.

### DIFF
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -88,7 +88,8 @@ impl TryFrom<UniTerm> for RichTerm {
         let rt = match node {
             UniTermNode::Var(id) => RichTerm::new(Term::Var(id), pos),
             UniTermNode::Record(r) => RichTerm::try_from(r)?,
-            UniTermNode::Types(ty) => {
+            UniTermNode::Types(mut ty) => {
+                fix_type_vars(&mut ty);
                 ty.contract().map_err(|UnboundTypeVariableError(id)| {
                     // We unwrap the position of the identifier, which must be set at this stage of parsing
                     let pos = id.pos;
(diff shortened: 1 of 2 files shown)


yannham (commented):
After thinking a bit about it, it does look sensible. Because `forall`s don't bind inside flat types anyway, I can't think of a case when fixing type vars when converting to a term would be too early. It also doesn't seem to incur additional work, because places where we already fix types are mostly at annotations, and such types won't be converted to a term.

Let's add at least the examples of the issue as regression tests, though.

francois-caddet:
Regression tests added.
It fix the issue at least.
Do we open an other issue when we will find an example where it break? or should we wait a bit before merging?

yannham (approved):

PR #856, opened 2022-10-05
Add language server support for type wildcards

### LINKED ISSUE #699: Add language server support for type wildcards
Currently, type wildcards aren't properly supported by the language server.  The LS should display the inferred type for each wildcard.  Instead, it displays `_`, the representation of a wildcard.

It might be enough to call `linearizer.retype_ident` on type-annotated `MetaValue` to replace wildcards with their unification variable.

ebresafegaga:
Closes #699

### DIFF
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -90,10 +90,11 @@ impl Default for AnalysisHost {
     }
 }
 
+use nickel_lang::types::Types;
 impl Linearizer for AnalysisHost {
     type Building = Building;
     type Completed = Completed;
-    type CompletionExtra = (UnifTable, HashMap<usize, Ident>, Vec<TypeWrapper>);
+    type CompletionExtra = (UnifTable, HashMap<usize, Ident>, Vec<Types>);
 
     fn add_term(
         &mut self,
@@ -347,11 +348,7 @@ impl Linearizer for AnalysisHost {
     fn complete(
         self,
         mut lin: Linearization<Building>,
-        (table, reported_names, wildcard_vars): (
-            UnifTable,
-            HashMap<usize, Ident>,
-            Vec<TypeWrapper>,
-        ),
+        (table, reported_names, wildcard_vars): (UnifTable, HashMap<usize, Ident>, Vec<Types>),
     ) -> Linearization<Completed> {
         debug!("linearizing");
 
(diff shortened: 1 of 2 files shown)


lsp/nls/src/linearization/mod.rs
```diff
@@ -93,7 +93,7 @@ impl Default for AnalysisHost {
 impl Linearizer for AnalysisHost {
     type Building = Building;
     type Completed = Completed;
-    type CompletionExtra = (UnifTable, HashMap<usize, Ident>);
+    type CompletionExtra = (UnifTable, HashMap<usize, Ident>, Vec<TypeWrapper>);
```
yannham, line 96: Nitpick: it may be time to turn this into a struct, just to have explicit names to describe what each component represents.

src/typecheck/mod.rs
```diff
@@ -249,8 +249,10 @@ where
         )?;
     }
 
-    let result = wildcard_vars_to_type(wildcard_vars, &table);
-    let lin = linearizer.complete(building, (table, names)).into_inner();
+    let result = wildcard_vars_to_type(wildcard_vars.clone(), &table);
+    let lin = linearizer
+        .complete(building, (table, names, wildcard_vars.clone()))
```
yannham, line 254:
```suggestion
        .complete(building, (table, names, result.clone()))
```

I think you want to use `result` here. `wildcard_vars` only maps wildcard ids to unification variables, not to the final, resolved type, if I remember correctly.

lsp/nls/src/linearization/mod.rs
ebresafegaga: Yeah, that makes sense. Since this struct would be a subset of the main type-checking `State` type, would it be better to just make the `CompletionExtra` associated type be the whole type-checking state, as in: 
```rust
type CompletionExtra = State
```

yannham: I don't think that will work so well, because `State` is tied to a lifetime and retains mutables references and not owned values. I fear you'll have lifetime annotations that bubble up everywhere.

lsp/nls/src/linearization/mod.rs
```diff
@@ -410,6 +417,10 @@ impl Linearizer for AnalysisHost {
                     meta,
                 },
             )
+            .map(|item| LinearizationItem {
+                ty: transform_wildcard(wildcard_vars.clone(), item.ty),
```
yannham, line 421: Do you really need to clone the whole array? `transform_wildcard` should rather take an immutable reference to `wildcard_vars`.

yannham (changes requested):
Does it work better now, with `result`?

ebresafegaga:
> Does it work better now, with `result`?

Yes it does, but I think it could have worked with `wildcard_vars`. I was initially testing with the `nls` binary from the crate registry rather than the locally built `nls`

yannham (approved):

src/typecheck/mod.rs
```diff
@@ -225,11 +233,11 @@ pub fn type_check_linearize<LL>(
     mut linearizer: LL,
 ) -> Result<(Wildcards, LL::Completed), TypecheckError>
 where
-    LL: Linearizer<CompletionExtra = (UnifTable, HashMap<usize, Ident>)>,
+    LL: Linearizer<CompletionExtra = Extra>,
 {
     let (mut table, mut names) = (UnifTable::new(), HashMap::new());
     let mut building = Linearization::new(LL::Building::default());
-    let mut wildcard_vars = Vec::new();
+    let mut wildcard_vars: Vec<TypeWrapper> = Vec::new();
```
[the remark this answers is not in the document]
ebresafegaga: We don't need it. I used it to guide my implementation at the time.

PR #857, opened 2022-10-06
Pretty print the optional metavalue annotation

vkleen:
The pretty printer used to ignore the optional annotation for metavalues.


vkleen:
That's a good point about the priorities. I guess, I'll merge this first and look at those next.

PR #860, opened 2022-10-06
Improve error messages to differentiate between flat types and static types

### LINKED ISSUE #718: Bad type error reporting mixing type vars and type id with the same name
**Describe the bug**

Sometimes the typechecker can throw  something like:

```
  = The type of the expression was expected to be `a`
  = The type of the expression was inferred to be `a`
``` 

it can actualy append when one is a type variable and the other one is a type identifier. 

**To Reproduce**
```
let f : forall a. {_: a} -> {_: a} -> {_: a}
  = fun fst snd => ((fst & s nd) | {_: a})
```

When no variable `a` is defined, it throw a unbound identifier. That's ok. But if we define `a` before the function (e.g.: `let a = Num`), it throw the confusing error shown above. 
the function below isn't very usefull ;) for a more complet and real case example you can check my comment in #423 

**Expected behavior**
I'm not sure if this error can appen in others contexts than the one described by #423 
But if it can or if fixing #423 is more dificult, we should probably find a way to distinguish types variables and identifiers in errors.

My first idea would be to force casing of type identifier to start with uper case letter and variables with lower case. But I realy do not think of what changes it imply in nickel.

Thank to share example if you have betters 👍 

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

francois-caddet:
Depending if type is flat or not, print a different note in the error.

fix #718

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1327,21 +1327,27 @@ impl ToDiagnostic<FileId> for TypecheckError {
                     ])]
             ,
             TypecheckError::TypeMismatch(expd, actual, span_opt) => {
-                fn get_note(ty: &Types) -> String {
+                fn addendum(ty: &Types) -> String {
                     if ty.0.is_flat() {
-                        String::from("the term")
+                        String::from(" (a contract)")
                     } else {
-                        String::from("a type expression")
+                        String::from("")
                     }
                 }
+                let last_note = if expd.0.is_flat() ^ actual.0.is_flat() {
+                    String::from("Static types and contracts are not compatible")
+                } else {
+                    String::from("These types are not compatible")
+                };
+
                 vec![
                     Diagnostic::error()
                         .with_message("incompatible types")
                         .with_labels(mk_expr_label(span_opt))
                         .with_notes(vec![
-                            format!("Expected an expression typed by {} `{}`", get_note(expd), expd),
-                            format!("Inferred this expression to be typed by {} `{}`", get_note(actual), actual),
-                            String::from("These types are not compatible"),
+                            format!("The type of the expression was expected to be `{}`{}", expd, addendum(expd)),
+                            format!("The type of the expression was inferred to be `{}`{}", actual, addendum(actual)),
+                            last_note,
                         ])]
             }
             TypecheckError::RowKindMismatch(ident, expd, actual, span_opt) => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -1326,17 +1326,24 @@ impl ToDiagnostic<FileId> for TypecheckError {
                         format!("Maybe you forgot to put a `forall {}.` somewhere in the enclosing type ?", ident),
                     ])]
             ,
-            TypecheckError::TypeMismatch(expd, actual, span_opt) =>
+            TypecheckError::TypeMismatch(expd, actual, span_opt) => {
+                fn get_note(ty: &Types) -> String {
+                    if ty.0.is_flat_type() {
+                        String::from("the term")
+                    } else {
+                        String::from("a type expression")
+                    }
+                }
                 vec![
                     Diagnostic::error()
                         .with_message("incompatible types")
                         .with_labels(mk_expr_label(span_opt))
                         .with_notes(vec![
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -1326,17 +1326,24 @@ impl ToDiagnostic<FileId> for TypecheckError {
...
+                }
                 vec![
                     Diagnostic::error()
                         .with_message("incompatible types")
                         .with_labels(mk_expr_label(span_opt))
                         .with_notes(vec![
-                            format!("The type of the expression was expected to be `{}`", expd),
-                            format!("The type of the expression was inferred to be `{}`", actual),
+                            format!("Expected an expression typed by {} `{}`", get_note(expd), expd),
+                            format!("Inferred this expression to be typed by {} `{}`", get_note(actual), actual),
```
yannham, line 1343: A couple remarks:

- The new formulation (the beginning of the sentence) works nicely, but we use the formula "The type of the expression was expected to be" in a lot of places. I'm not too keen on having different error messages for no apparent reasons, so I think we have to change them all or to change none.
- "an expression typed by the term" doesn't sound right. We don't really have a standard definition for what is a type expression neither.

Here is a suggestion for a minimal change: keep the original structure of the error message, and only add something if one is a flat type with a variable inside and the other is a type variable:

```rust
                fn addendum(ty: &Types, other: &) -> &str {
                    match ty.0 {
                        AbsType::Flat(t) if !matches(t.as_ref(), Term::Var(_)) => " (contract)",
                        AbsType::Var(_) => " (type variable)",
                        _ => "",
                    }
                }
                vec![
                    Diagnostic::error()
                        .with_message("incompatible types")
                        .with_labels(mk_expr_label(span_opt))
                        .with_notes(vec![
                            format!("The type of the expression was expected to be `{}`{}", expd, addendum(expd)),
                            format!("The type of the expression was inferred to be `{}`{}", actual, addendum(actual)),
                            String::from("These types are not compatible"),
                        ])]
```

src/error.rs
```diff
@@ -1326,17 +1326,24 @@ impl ToDiagnostic<FileId> for TypecheckError {
                         format!("Maybe you forgot to put a `forall {}.` somewhere in the enclosing type ?", ident),
                     ])]
             ,
-            TypecheckError::TypeMismatch(expd, actual, span_opt) =>
+            TypecheckError::TypeMismatch(expd, actual, span_opt) => {
+                fn get_note(ty: &Types) -> String {
+                    if ty.0.is_flat_type() {
```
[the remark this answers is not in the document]
francois-caddet: Why? this function is called later with both parts of the error.

src/error.rs
francois-caddet: Qre we sure vars are the only things which can cause confusion? could we have it with records?

src/error.rs
```diff
@@ -1326,17 +1326,30 @@ impl ToDiagnostic<FileId> for TypecheckError {
                         format!("Maybe you forgot to put a `forall {}.` somewhere in the enclosing type ?", ident),
                     ])]
             ,
-            TypecheckError::TypeMismatch(expd, actual, span_opt) =>
+            TypecheckError::TypeMismatch(expd, actual, span_opt) => {
+                fn addendum(ty: &Types) -> String {
```
yannham, line 1330: Nitpick: I think you can use `&str` instead of `String` here if you just need to print them, avoiding an allocation.

yannham (approved):

src/error.rs
```diff
@@ -1326,17 +1326,30 @@ impl ToDiagnostic<FileId> for TypecheckError {
...
-            TypecheckError::TypeMismatch(expd, actual, span_opt) =>
+            TypecheckError::TypeMismatch(expd, actual, span_opt) => {
+                fn addendum(ty: &Types) -> &str {
+                    if ty.0.is_flat() {
+                        " (a contract)"
+                    } else {
+                        ""
+                    }
+                }
+                let last_note = if expd.0.is_flat() ^ actual.0.is_flat() {
```
yannham, line 1337: TIL that Rust has a syntax for exclusive or

Issue #859, opened 2022-10-06
Pretty printer identifier quoting is too lax


vkleen:
**Describe the bug**
The pretty printer doesn't quote enough record keys.

**To Reproduce**
```
❯ nickel pprint-ast <<<'{ "priority" = 5 }'
{ priority = 5, }
❯ nickel repl
nickel> { priority = 5, }
error: unexpected token
  ┌─ repl-input-0:1:3
  │
1 │ { priority = 5, }
  │   ^^^^^^^^
```

**Expected behavior**
A pretty-printed AST should always be parsable.

**Environment**
 - OS name + version: NixOS
 - Version of the code: master fcaa7669047b6d80b01d1eebd21c335b5c3dd0a2

yannham:
As Viktor has left, and is unlikely to make this progress, I'm going ahead and closing this. The original surface bug is fixed anyway; the discussion is more about a nicer implementation internally.

PR #861, opened 2022-10-07
Add a keywords list for use by the pretty printer

vkleen:
This is a quick fix for #859. I'm not convinced it's the best approach, especially because it asks people to keep updating a list of strings as the syntax evolves. But that should be an infrequent issue.

As it is, this PR gets the pretty printer to produce parseable records even if the record field names coincide with keywords.

### DIFF
--- a/src/parser/lexer.rs
+++ b/src/parser/lexer.rs
@@ -33,6 +33,8 @@ use crate::parser::error::{LexicalError, ParseError};
 use logos::Logos;
 use std::ops::Range;
 
+// When adding or removing tokens that might be parsed as identifiers please update the KEYWORDS
+// list
 /// The tokens in normal mode.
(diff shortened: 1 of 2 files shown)


src/parser/lexer.rs
```diff
@@ -33,6 +33,8 @@ use crate::parser::error::{LexicalError, ParseError};
 use logos::Logos;
 use std::ops::Range;
 
+// When adding or removing tokens that might be parsed as identifiers please update the KEYWORDS
```
yannham, line 36:
```suggestion
// **IMPORTANT**
// When adding or removing tokens that might be parsed as identifiers,
// please update the KEYWORDS array
```

yannham (approved):
That's not ideal, but indeed I don't see right now an easy way to generate them automatically from logos, so let's do like this for the time being.

PR #862, opened 2022-10-07
Refactor/virtual machine now owns resolver +  repl and program have own virtual machine

dpl0a:
`VirtualMachine` no longer depends on a lifetime parameter now and it owns its `import_resolver`field.
Moreover the `cache` field in both `ReplImpl` and `Program` have been replaced by an instance of `VirtualMachine` holding the cache. 
Now instead of creating a new `VirtualMachine` each time, the `vm` from a `ReplImpl` or a `Program` gets its `stack` and `call_stack` cleared and reused.

### DIFF
--- a/src/program.rs
+++ b/src/program.rs
@@ -43,7 +43,7 @@ pub struct Program {
     /// The cache holding the sources and parsed terms of the main source as well as imports.
     //cache: Cache,
     ///
-    vm: VirtualMachine<Cache>
+    vm: VirtualMachine<Cache>,
 }
 
 impl Program {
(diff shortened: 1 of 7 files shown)


src/program.rs
```diff
@@ -40,7 +41,9 @@ pub struct Program {
     /// The id of the program source in the file database.
     main_id: FileId,
     /// The cache holding the sources and parsed terms of the main source as well as imports.
-    cache: Cache,
+    //cache: Cache,
+    ///
```
yannham, line 45: Looks like the comments have been messed up a bit.

src/repl/mod.rs
```diff
@@ -77,35 +78,39 @@ pub struct ReplImpl {
...
+    ///
+    vm: VirtualMachine<Cache>,
 }
 
 impl ReplImpl {
     /// Create a new empty REPL.
     pub fn new() -> Self {
         ReplImpl {
-            cache: Cache::new(ErrorTolerance::Strict),
+            //cache: Cache::new(ErrorTolerance::Strict),
```
yannham, line 89:
```suggestion
```

yannham (approved):

PR #867, opened 2022-10-11
Add basic LSP completion for record terms

### LINKED ISSUE #453: Complete record fields and list them as symbols
LSP clients supporting document symbols should receive record field symbols.

It should also be possible to get completions for record fields if the completion is invoked after typing a record.

```
let record = {
  foo = 0,
  bar = 1
} in record.
            ^
            resolves `foo` and `bar`
```

ebresafegaga:
Closes #453


lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +26,151 @@ pub fn handle_completion(
...
+                .take_while(|c| c.is_ascii_alphabetic() || *c == '.')
+                .skip_while(|c| *c == '.') // skip . (if any)
+                .collect::<String>()
+                .chars()
+                .rev()
+                .collect::<String>();
+
+            linearization
+                .linearization
+                .iter()
```
yannham, line 88: Does this mean we are iterating through the whole linearization each time we want to trigger completion? The linearization can be quite big for big files.

Also, what if the there is another identifier in scope? For example, if I do:

```
let x = {foo = 1} in
let x = {bar = 1} in
x.
```

I get `foo` and `bar` as suggestions, but `foo` will cause an error at runtime.

yannham (commented):
I may see a few issues with this approach:

1. We have to go through the whole linearization each time we trigger record completion. One of the point of the linearization system is to be able to make efficient requests based on the position in a file. We probably want to do something more like the goto definition mechanism (which is, in some sense, quite similar: we have an identifier in some scope, and we need to get the information about its definition), and use the scope for that. This probably means storing the record completion information directly in the `Declaration` and `RecordField` nodes instead of storing them in their own node.
2. I haven't try hard, but I fear that you can't add a new type of node freely like this without introducing bugs in the hover or goto definition/usage. The way the linearization works is that it uses a binary search to find an item at a given position. For this to work, you need to have at most one item with the exact same position. It can be worked around if there are several of them but we are careful to put the one that must be found in last, but still. In addition to 1., I think this is a second argument to not add a new type of linearization item, but rather add new fields in the existing one. For example I see that `usages` is both in `Declaration` and `RecordField`. Maybe it's time to have a separate datatype for the information the must be associated to a binding, and attach that to both a `Declaration` and `RecordField`. But this can be done in a separate/subsequent PR.

ebresafegaga:
> 
> 1. We have to go through the whole linearization each time we trigger record completion. One of the point of the linearization system is to be able to make efficient requests based on the position in a file. We probably want to do something more like the goto definition mechanism (which is, in some sense, quite similar: we have an identifier in scope, and we need to get the information about its definition), and use the scope for that. This probably means storing the record completion information directly in the `Declaration` and `RecordField` nodes instead of storing them in their own node.

The main reason it is like this is because at the cursor position, when completion happens, there is no term there; so if we the linearization tries to get the item there, it results in `None`. I think we can fix this by  finding a linearization item which has the same ID as the identifier before the completion `.` and get its scope. This was like a quick hack to get it working immediately.

ebresafegaga:
>
> 2. I haven't tried hard, but I fear that you can't add a new type of node freely like this without introducing bugs in the hover or goto definition/usage. The way the linearization works is that it uses a binary search to find an item at a given position. For this to work, you need to have at most one item with the exact same position. It can be worked around if there are several of them but we are careful to put the one that must be found in last, but still. In addition to 1., I think this is a second argument to not add a new type of linearization item, but rather add new fields in the existing one. For example I see that `usages` is both in `Declaration` and `RecordField`. Maybe it's time to have a separate datatype for the information the must be associated to a binding, and attach that to both a `Declaration` and `RecordField`. But this can be done in a separate/subsequent PR.

Yes, I introduced a bug with the hover when working on this, but I fixed it. Apart from that, I don't see any reason why this new `TermKind`  should be problematic. You're also right, it can be merged with into "data about bindings" in `TermKind::Declaration(..)` , which would be better.

lsp/nls/src/linearization/mod.rs
```diff
@@ -104,6 +104,41 @@ impl Linearizer for AnalysisHost {
...
+        fn get_record_bind_item(
+            ty: TypeWrapper,
+            scope: Scope,
+            ident: &Ident,
+            id: usize,
+            term: &RichTerm,
+        ) -> Option<LinearizationItem<Unresolved>> {
+            let mut t = term.as_ref();
+            match t {
+                Term::MetaValue(MetaValue {
```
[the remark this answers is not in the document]
ebresafegaga: I can't seem find it. Do you know where it is located?

lsp/nls/src/requests/completion.rs
ebresafegaga: This is kinda tricky. I have noticed that the parser is a bit tolerant some cases, and this is one of such cases. When we type this: `x.` , the parser returns a `ParseError` node. This node might not have the scope of the previous identifiers. Usually , we'd do this `linearization.get_in_scope(&item)` to get the scope, where `item` would be a linearization item that represents the parse error node, but it won't return anything in scope. So searching the whole linearization is a fix  that works for now. 

The second problem can be fixed by checking against unique ids instead of identifier labels, I'll fix it.

yannham: Hm, do you know why it doesn't return anything in scope? Is it because `ParseError` is treated specifically? My hunch with this PR's solution is that it adds a linear cost in the size of the AST for something that is automatically triggered. If record completion was something that needed an explicit action from the user, then that's fine if it starts slow, it's an additional feature. We can polish it on the way. But this is not the case here, as you get completion as soon as you type a `.`. We tried on very small examples but I wonder how that goes on something like the mantis example, or even bigger.

If the scope solution is complex, then we can go with the current solution for now, but I just want to make sure and to understand why the scope way is _actually_ hard.

ebresafegaga: I'm looking into it

ebresafegaga: We use IDs in scope for completion now, except when we want to check for `RecordFields { .. }`, because those `LinearizationItem`s have their own special scope (only visible to the themselves)

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +29,194 @@ pub fn handle_completion(
...
+            .get_in_scope(&item)
+            .iter()
+            .filter_map(|i| match i.kind {
+                TermKind::Declaration(ref ident, _, _) => Some((vec![ident.clone()], i.ty.clone())),
+                _ => None,
+            })
+            .collect::<Vec<_>>(),
+    };
+
+    // TODO: remove duplicates
```
sir4ur0n, line 202: This TODO should probably be addressed in this PR, or an issue should be created and referenced here to track this remaining work

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +29,194 @@ pub fn handle_completion(
...
     let item = item.unwrap().to_owned();
 
-    let in_scope: Vec<_> = linearization
-        .get_in_scope(&item)
+    Trace::enrich(&id, linearization);
+
+    let trigger = params.context.and_then(|context| context.trigger_character);
+    let in_scope = match trigger {
+        // Record Completion
+        Some(s) if s == "." => {
```
sir4ur0n, line 116: Silly question from a total Rust noob: can't we do this instead?
```suggestion
        Some(".") => {
```

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +29,194 @@ pub fn handle_completion(
...
+            // Get the ID before the .
+            // if the current source is `let var = {a = 10} in var.`
+            // name = "var"
+            let name = text[..start - 1]
+                .chars()
+                .rev()
+                .skip_while(|c| *c == '.') // skip . (if any)
+                .take_while(|c| c.is_ascii_alphabetic())
```
sir4ur0n, line 124: This test to find an identifier (if my understanding is correct: this piece of code is looking for the identifier before the `.`) looks suspicious to me.

E.g. `_` is a valid character in some Rust identifiers yet it will not be accepted here. So `foo_bar.` would look for the identifier `bar`, right?

More generally, I would argue that this can be extracted in a pure function: take a string and extract the identifier.

And I also suggest writing some tests to easily check the behavior is as expected. Currently there is no automated way to ensure the behavior is correct right?

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +29,194 @@ pub fn handle_completion(
...
+                .iter()
+                .find(|(ident, _)| ident.label() == name)
+                .map(|(_, id)| *id);
+            if name_id == None {
+                // NOTE: this might result in a case where completion is not working.
+                // This happens when an item cannot be found from the linearization.
+                server.reply(Response::new_ok(id, Value::Null));
+                return Ok(());
+            }
+            let name_id = name_id.unwrap();
```
sir4ur0n, line 140: Once again this might be a total noob question, please bear with me :pray: 

If I understand this code correctly, we unsafely extract from the `Some` after handling the `None` case and early-returning.

This might be dangerous in the face of refactoring: e.g. moving the `if` below would throw at runtime.

Can we use pattern matching instead? That way, there is a guarantee of safety "for free":
```rust
match name_id {
  None => {
    server.reply(Response::new_ok(id, Value::Null));
    return Ok(());
  }
  Some(name_id) => {...}
}
```

sir4ur0n (commented):
I am about 2-hours-old on the Nickel codebase so I dare not `Request changes` :smile: 

I did ask some questions - mostly related to Rust, but also a few about the feature - but feel free to discard any and all as "RTF-Rust-M", I will not take any offense :rofl:

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +29,194 @@ pub fn handle_completion(
...
+
+            linearization
+                .get_in_scope(&item)
+                .iter()
+                .filter_map(|i| {
+                    let (ty, _) = linearization.resolve_item_type_meta(i);
```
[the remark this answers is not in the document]
yannham: I would say using `_` is common in pattern matching and destructuring in Rust (and other functional languages). As a datapoint, I personally do that in OCaml and Rust, because somehow if you don't use this variable, it should be irrelevant to understanding the code (if the thing you extract is correctly named), same as using ellipsis `..` in Rust when matching on a big struct:

```rust
if let Foo { bar: "bar", baz, ..} = value {
 ..
}
```

If we follow your convention, and `Foo` has 10 fields, we should know and write them all, which I don't personally find that useful. But, as with many things, it's also a matter of taste :) I think both approach are used and reasonable.

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes we can do this, but we need to map `trigger` with a the transformation: `String -> &str `, because string literals have the type `&str`, but the LSP is giving us a `String`

lsp/nls/src/requests/completion.rs
ebresafegaga: You're right! Currently `foo_bar` would not get a completion. 

Yes, I think using a pure function would be better.

But for the tests, I don't think there is any part of the LSP that has tests right now; I think it would be better to make another issue for that.

ebresafegaga: We should be matching this for identifiers: 

```rust 
 #[regex("_?[a-zA-Z][_a-zA-Z0-9-']*")]
    Identifier(&'input str)
```

lsp/nls/src/requests/completion.rs
ebresafegaga: I think the reason it is like that is to avoid the nesting that comes with that pattern matching. But I think we can also have this safety guarantee by using the `?` short-circuiting operator:
```rust
 let item = linearization.item_at(&locator);
 let item = item.ok_or(ResponseError {
        code: ErrorCode::InternalError as i32,
        message: "Couldn't get the linearization item at the cursor position".to_owned(),
        data: None,
    })?;
```

But this way, it doesn't return an `Ok(..)` response

lsp/nls/src/requests/completion.rs
sir4ur0n: > But for the tests, I don't think there is any part of the LSP that has tests right now; I think it would be better to make another issue for that.

What reasons lead you to do it in another PR?

Does it perhaps require a lot of work to create a first test in `./lsp/nls`? Besides the usual:
* Add test lib in `Cargo.toml`
* Add `cargo test` in CI
 
If not, I'd say it's ok to add the first test as part of this PR, because adding feature X and adding tests about feature X belong in the same PR.

If it is a lot of work, I suggest we shortly press pause on this PR, open a new PR to add test infrastructure for the LSP, then rebase this branch and add tests

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,99 @@
...
+/// Get the identifier before the `.`, for record completion.
+fn get_identifier(text: &str) -> Result<String, ResponseError> {
+    let name: String = text
+        .chars()
+        .rev()
+        .skip_while(|c| *c == '.') // skip . (if any)
+        .collect();
+
+    // unwrap is safe here because we know this is a correct regex.
+    let regex = regex::Regex::new(r"[_a-zA-Z0-9-']*[a-zA-Z]_?").unwrap();
```
sir4ur0n, line 87: I don't know for Rust, but in many languages, compiling a regex is not trivial, so it's usually advised to create it once as a constant. What's the status for Rust? Because this will be evaluated every time someone presses `Ctrl Space` on a dot

sir4ur0n, line 87: I think this regexp is wrong: there are many other characters allowed in identifiers in Rust, according to https://doc.rust-lang.org/reference/identifiers.html
Counter-examples:
* `京`
* `к`

E.g. I just tested with `fooкbar.` and it returns `bar` rather than `fooкbar`.

Tests, in particular PBTs, would be quite helpful in checking this function is well-behaved

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +104,149 @@ pub fn handle_completion(
...
+    let text = server.cache.files().source(file_id);
     let start = position_to_byte_index(
         server.cache.files(),
         file_id,
         &params.text_document_position.position,
     )
     .unwrap();
 
-    let locator = (file_id, ByteIndex(start as u32));
+    let locator = (file_id, ByteIndex((start - 1) as u32));
```
sir4ur0n, line 115: Why `- 1`? Maybe this should be documented, because future readers might ask the same question

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +104,149 @@ pub fn handle_completion(
...
+        context
+            .trigger_character
+            .as_ref()
+            .map(|trigger| trigger.as_str())
+    });
+
+    let in_scope = match trigger {
+        // Record Completion
+        Some(".") => {
+            let name = get_identifier(&text[..start - 1])?;
```
sir4ur0n, line 143: Another occurrence of `start - 1`.

I think this emphasizes my comment above: this probably deserves being extracted in a variable, and maybe a short comment on that variable to explain the `- 1`

lsp/nls/src/requests/completion.rs
sir4ur0n: So what is the idiom in Rust, or in the Nickel codebase in particular?
I would pick "safe and future proof" over "less indentation level" every day of the year, but I don't want to enforce my opinion on the codebase :smile:

sir4ur0n: Side note: maybe the problem is also that this function is too big, and takes care of too many things? It is currently 156 lines long, which might be an indicator.

lsp/nls/src/requests/completion.rs
ebresafegaga: Well, this is the reverse (because we reversed the source text) of what is in the Nickel lexer, so I think it should be good for Nickel source files: 
```rust
 #[regex("_?[a-zA-Z][_a-zA-Z0-9-']*")]
    Identifier(&'input str),
```

lsp/nls/src/requests/completion.rs
ebresafegaga: `-1` because at the current cursor position, there is no *linearization* item there, so we try to get the item just before the cursor.

lsp/nls/src/requests/completion.rs
ebresafegaga: I don't think Nickel supports unicode characters in identifiers

lsp/nls/src/requests/completion.rs
ebresafegaga: This one was to remove the `.`, but I don't think we need it anymore since `get_identifier` skips `.`s

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +104,149 @@ pub fn handle_completion(
...
+                .iter()
+                .find(|seen_item| seen_item.label == item.label)
+                .is_none()
+            {
+                seen.push(item.clone())
+            }
+        }
+
+        seen
+    }
```
sir4ur0n, line 230: Suggestions:
* use `.any()` rather than `.find().is_none()`
* use `.fold()` rather than a `for` loop

This should give something like
```rust
    fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
        items.iter().fold(Vec::with_capacity(0), |mut seen, item| {
            if !seen.iter().any(|seen_item| seen_item.label == item.label) {
                seen.push(item.clone())
            }
            seen
        })
    }
```

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +104,149 @@ pub fn handle_completion(
...
+            .iter()
+            .filter_map(|i| match i.kind {
+                TermKind::Declaration(ref ident, _, _) => Some((vec![ident.clone()], i.ty.clone())),
+                _ => None,
+            })
+            .collect::<Vec<_>>(),
+    };
+
+    // O(n^2)
+    fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
```
sir4ur0n, line 217: What are the benefits of declaring this function inside `handle_completion` function, rather than on its own?

Doesn't it create a new function each time `handle_completion` is called?

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +104,149 @@ pub fn handle_completion(
...
+        .map(|(idents, _)| {
+            idents
+                .iter()
+                .map(|ident| CompletionItem {
+                    label: ident.into_label(),
+                    ..Default::default()
+                })
+                .collect::<Vec<_>>()
         })
+        .flatten()
```
sir4ur0n, line 243: Can we merge this `.map().flatten()` into a single `.flat_map()`?
```suggestion
        .flat_map(|(idents, _)| {
            idents
                .iter()
                .map(|ident| CompletionItem {
                    label: ident.into_label(),
                    ..Default::default()
                })
                .collect::<Vec<_>>()
        })
```

lsp/nls/src/server.rs
```diff
@@ -50,7 +50,7 @@ impl Server {
             definition_provider: Some(OneOf::Left(true)),
             references_provider: Some(OneOf::Left(true)),
             completion_provider: Some(CompletionOptions {
-                trigger_characters: Some(vec![]),
+                trigger_characters: Some(vec![".".to_owned()]),
```
sir4ur0n, line 53: This `"."` refers to exactly the same character as the one in `completion.rs` on lines 83 and 142. They refer to "Hey, the user just typed `.` and would like completion". This is not a coincidence.

Yet currently, it is not trivial for the reader to identify if they refer to the same thing, or where this `"."` comes from. They might as well have completely different meanings.

There is a **domain** coupling between:
* `server.rs` registers the trigger `.`
* `completion.rs` handles the trigger `.`

As a rule of thumb, it is healthy to have static (type checked) coupling between things that have a domain coupling (i.e., things that **should be coupled**). It adds semantic and clarity (and a teeny tiny bit of safety, but this is negligible here).

As such, I suggest extracting into a constant here (or in a better-suited place), and refer to that constant in `completion.rs`.

lsp/nls/src/requests/completion.rs
yannham: I tend to agree with @sir4ur0n on this one. It doesn't feel right to do an `unwrap` when we can prove easily at compile time that there is no way of panicking (as a side note, although in this case the compiler may be able to optimize the check away, it also has a potential runtime cost, because `unwrap` need to check again that the value is not `None`. It's probably highly negligible in practice, so it's more of a psychological thing, but still)

lsp/nls/src/requests/completion.rs
yannham: https://docs.rs/regex/latest/regex/#example-avoid-compiling-the-same-regex-in-a-loop

lsp/nls/src/requests/completion.rs
yannham: I don't think I agree with the `fold` one. Returning the `seen` at the end feels artificial. It's common when filling a container to declare a local mutable variable and fill it using a `for` loop (and there might be a performance difference, but that's harder to assess). I did the fold thing initially too, coming from langauges without a `for`, but after some time I think a `for` is actually more meaningful and readable if you're mutating the accumulator in place.

lsp/nls/src/requests/completion.rs
yannham: IIRC, such functions are actually lifted to the top-level (because the compiled code has no notion of "local" functions). So in term of code generation, this should be strictly equivalent. Usually you use such local functions that because for specialized helpers, just as a way of better organizing your function implementation. Your helper may not make sense in any other context, or you don't expect it to be reused somewhere else. In this case, a local function don't pollute the top-level module API.

lsp/nls/src/requests/completion.rs
sir4ur0n: How about linking to https://nickel-lang.org/user-manual/syntax in a comment here for reference? To let the reader know exactly what characters form valid identifiers

sir4ur0n, line 87: According to @matthew-healy in https://tweag.slack.com/archives/C0133ME1F4P/p1666780348629259?thread_ts=1666780145.455399&cid=C0133ME1F4P the regex used by the lexer is `_?[a-zA-Z][_a-zA-Z0-9-']*` which is different from the one here.

We must be consistent.

I would even go 1 step further and argue we should reuse the lexer's regexp rather than copying it here, if possible

lsp/nls/src/requests/completion.rs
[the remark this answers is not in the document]
sir4ur0n: While I won't push further on this, I think the ellipsis is a different topic.

Tuple fields are unnamed: there is no trivial way to distinguish the parts other than the order, which is brittle. Even worse when the different tuple fields are of the same type. A change of order of fields in the type and return of the function returning a tuple could have dramatic, undetected effects.

Ellipsis happens on named fields (to the best of my knowledge) so the risk is significantly lower: unless one changes both the fields of a record AND also swaps field names (which is an extremely dangerous thing to do :smile: ), an ellipsis can't break the code at runtime in the future.

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes, this one is the reverse of what we have in the lexer because the source text is reverse.

lsp/nls/src/requests/completion.rs
sir4ur0n: My comment was more about code organization and clarity than performance or compilation :sweat_smile: 

Having this function **in the middle** of the algorithm feels weird and disrupts the reader in my opinion. Since the function name is rather clear, there isn't a need to understand that function implementation to understand the algorithm here. So by moving it out of the algorithm (or moving it at the very beginning before the start of the algorithm) the algorithm would be clearer for readers

lsp/nls/src/requests/completion.rs
ebresafegaga: For the reuse, I don't think constants literals are allowed for defining the regex in the lexer:
```rust
  #[regex("_?[a-zA-Z][_a-zA-Z0-9-']*")]
  Identifier(&'input str),
```

The `#regex` macro only accepts a string literal, I think. 

Even the lexer doesn't share common regexes: 
``` rust
 // **IMPORTANT**
// This regex should be kept in sync with the one for RawEnumTag below.
#[regex("_?[a-zA-Z][_a-zA-Z0-9-']*")]
Identifier(&'input str),

// **IMPORTANT**
// This regex should be kept in sync with the one for Identifier above.
#[regex("`_?[a-zA-Z][_a-zA-Z0-9-']*", |lex| lex.slice().split_at(1).1)]
RawEnumTag(&'input str),
```

Maybe the best we can do is add a comment also stating that the LSP should be updated as well, if there is any change to be made.

sir4ur0n: Then we probably want to document this in both places, since this is a hidden domain coupling between both regexes: This regexp **must** be the reverse of the regexp in `<add link to lexer code here>` to detect/parse/identify identifiers

Also add a comment in the lexer regexp: any change in the lexer regex **must** also be backported in this regex.


We could probably go 1 step further but I suspect not everyone will agree: store regex components, store the list, and use the reverse list here:
```rust
// lexer.rs or whatever file contains the original regex :D
let optional_starting_underscore = r"_?"
let first_character = r"[a-zA-Z]"
let other_characters = r"[_a-zA-Z0-9-']*"
let pub identifier_regex_components = [optional_starting_underscore, first_character, other_characters]
let identifier_regex = regex::Regex::new(identifier_regex_components.join()).unwrap();

// completion.rs
use nickel::lexer::identifier_regex_components

/// Now the domain coupling between `lexer.rs` and this file is not only made explicit, but it's checked.
let reverse_identifier_regex = regex::Regex::new(identifier_regex_components.reverse().join()).unwrap();
```

lsp/nls/src/linearization/building.rs
```diff
@@ -100,6 +101,7 @@ impl Building {
                 },
             });
             env.insert(ident.clone(), id);
+            self.env.insert(ident.clone(), id);
```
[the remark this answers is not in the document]
ebresafegaga: So this has to do with the way the `linearizer` (the "thing" that traverses an AST) and a `linearization` (data gotten from traversing an AST) works in the LSP. The linearizer traverses the AST, and it need an environment to be able to resolve variables, but all the *persistent* data are actually stored in the `linearization`. When we get to a new child of an  AST node, we create a new `linearizer`, with a fresh environment (among other things), and that new `linearizer` traverses the child node; the only way to have access to this environment after traversing the AST is through the `linearization`.  So one of those environments is for the `linearizer` while the other is for the `linearization` 

This was just a quick hack to get things working though, I think we can completely remove the environment from the `linearizer`, and depend on the persistent environment in the `linearization`

lsp/nls/src/requests/completion.rs
yannham: But you asked 

> Doesn't it create a new function each time handle_completion is called?

which I interpreted as does it allocate a new closure each time the function is called? What did you originally mean by that, then?

> Having this function in the middle of the algorithm feels weird and disrupts the reader in my opinion. Since the function name is rather clear, there isn't a need to understand that function implementation to understand the algorithm here. So by moving it out of the algorithm (or moving it at the very beginning before the start of the algorithm) the algorithm would be clearer for readers

Ah, then yes, I totally agree. Subfunctions should be declared only at the beginning of a function, IMHO.

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
 use codespan::ByteIndex;
```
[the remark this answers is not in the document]
sir4ur0n: Scenario 2 will be addressed in https://github.com/tweag/nickel/issues/877

Scenario 1 remains a valid question for this PR

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
...
+    server::{self, Server},
     trace::{Enrich, Trace},
 };
 
+/// Find record fields for an item with the specified id.
+fn find_record_fields(
+    linearization: &Vec<LinearizationItem<Types>>,
+    id: usize,
+) -> Option<Vec<Ident>> {
+    linearization.iter().find_map(|item| match item.kind {
```
yannham, line 29: I may be missing something again, but we still do a linear search through the whole linearization, don't we? The point of indices is to have constant time access. Couldn't we rather use `Completed::get_item` instead of doing this `iter()`?

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
...
+) -> Result<Vec<CompletionItem>, ResponseError> {
+    let lin_env = &linearization.lin_env;
+    let in_scope = match trigger {
+        // Record Completion
+        Some(server::DOT_COMPL_TRIGGER) => {
+            let name = get_identifier(source)?;
+            let name_id = lin_env
+                .iter()
+                .find(|(ident, _)| ident.label() == name)
+                .map(|(_, id)| *id);
```
yannham, line 128: Why do we still do an `iter()` here, which is linear-time, instead of using `Environment::get` ?

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
...
+                                        find_record_fields(&linearization.linearization, *body_id)
+                                            .map(|idents| (idents, item.ty.clone()))
+                                    }
+                                }
+                            }
+                            _ => None,
+                        }
+                    })
+                    .collect(),
+                None => return Ok(Vec::with_capacity(0)),
```
yannham, line 166:
```suggestion
                None => return Ok(Vec::new()),
```

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
...
+                            }
+                            (TermKind::Declaration(_, _, ValueState::Known(body_id)), _)
+                                if name_id == item.id =>
+                            {
+                                match find_record_contract(&linearization.linearization, *body_id) {
+                                    // Get record fields from contract metadata
+                                    Some(contract) => {
+                                        let fields = match &contract.types {
+                                            Types(AbsType::Record(row)) => extract_ident(row),
+                                            _ => Vec::with_capacity(0),
```
yannham, line 151:
```suggestion
                                            _ => Vec::new(),
```

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,198 @@
...
+    let in_scope: Vec<_> = in_scope
+        .iter()
+        .flat_map(|(idents, _)| {
+            idents
+                .iter()
+                .map(|ident| CompletionItem {
+                    label: ident.into_label(),
+                    ..Default::default()
+                })
+                .collect::<Vec<_>>()
```
yannham, line 189: Do we necessarily need to collect here? Isn't the outer collect sufficient?

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +203,45 @@ pub fn handle_completion(
...
         server.cache.files(),
         file_id,
         &params.text_document_position.position,
     )
     .unwrap();
 
-    let locator = (file_id, ByteIndex(start as u32));
+    // -1 because at the current cursor position, there is no linearization item there,
+    // so we try to get the item just before the cursor.
+    let locator = (file_id, ByteIndex((start - 1) as u32));
```
yannham, line 216: Nitpick: I guess there may be an item at the cursor (for example partial identifiers are valid identifiers). I think you rather want this for record completion to work, and it turns out it's fine for normal completion too, because if you started typing an identifier, the position one byte before is the same token as well?

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +203,45 @@ pub fn handle_completion(
...
     let linearization = server.lin_cache_get(&file_id)?;
-
-    Trace::enrich(&id, linearization);
+    // Cloning to avoid mutable borrow and shared borrow at the same time,
+    // which won't compile
+    let linearization = linearization.clone();
```
yannham, line 220: It looks really fishy to clone the whole linearization. I fear this is a very costly operation. What is exactly the problem? I think we can probably solve it by using the right approach and borrowing.

lsp/nls/src/requests/completion.rs
yannham: Oh, because the `lin_env` is an env of `Ident`. Then maybe it should be an environment of type  `Environment<String, usize>`? Otherwise it feels a bit useless to use an `environment` but to iterate over it as a vector.

yannham (changes requested):
It's in better shape!

But there are still `clones` and other iteration which are bothering.

Tangentially, I feel like we do a bit of hacky heuristics each time we need to do record completion in the LSP. Like look at the contract (but not at the `types`, by the way), or at the definition, etc. I would rather have a proper datastructre, tailored to the need of completion (well, that would just be a vector of identifiers, I suppose) that is built statically during the analysis phase, by pasting together all the information we have from contracts, types and the definition of the record. That being said, we can discuss this in a follow-up work. For now, I think I would like to be sure that we are not doing whole linearization iteration that we don't need to do.

lsp/nls/src/requests/completion.rs
yannham: A second solution is to create an Ident from the string. Thanks to string interning both approaches should be close, so pick whichever is simpler from a code point of view.

lsp/nls/src/requests/completion.rs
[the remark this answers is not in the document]
yannham: You're absolutely right about the advantage of named fields vs unnamed fields. I think it does speak for using them as much as possible when defining datatypes with more than one argument (which is something we haven't always done in the past). For example, you could argue that giving a proper type to the result of `resolve_item_type_meta` is a good idea.

However, once we have to deal with a tuple, as the caller, I fail to see how giving a name to an unused variable protects against any risk you mention. Program A with `_` and program A' with `_named` are strictly equivalent with respect to runtime behavior and compilation. So if you refactor the order of fields, your program will either silently compile and do the wrong thing at runtime, or break at compile time in the same way, whether you give names to unused fields or not.

In the end, naming an unused field is merely adding a non-checked comment for readers. Which is arguably not a bad thing to do, but I just wanted to give another perspective, as I personally prefer not to add mental load for something that can be ignored, and to say that it's more a matter of taste. I trust Gaga to make an informed choice now :slightly_smiling_face:

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes. It works well with normal completion. This is just for record completion to work

lsp/nls/src/requests/completion.rs
ebresafegaga: Fixed!

lsp/nls/src/requests/completion.rs
ebresafegaga: Nice, thank you! Fixed.

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes. Nice catch, thanks!

ebresafegaga:
>Tangentially, I feel like we do a bit of hacky heuristics each time we need to do record completion in the LSP. Like look at the contract (but not at the types, by the way), or at the definition, etc. I would rather have a proper datastructre, tailored to the need of completion (well, that would just be a vector of identifiers, I suppose) that is built statically during the analysis phase, by pasting together all the information we have from contracts, types and the definition of the record. That being said, we can discuss this in a follow-up work.

Yes, I agree with this. Should I create an issue to track this?

lsp/nls/src/requests/completion.rs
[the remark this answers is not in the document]
ebresafegaga: Fixed scenario 1!

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,230 @@
...
+    server::{self, Server},
     trace::{Enrich, Trace},
 };
 
+/// Find record fields for an item with the specified id.
+fn find_record_fields(linearization: &Completed, id: usize) -> Option<Vec<Ident>> {
+    let item = linearization.get_item(id)?;
+    match item.kind {
+        TermKind::Record(ref fields) if item.id == id => {
+            Some(fields.keys().map(|ident| ident.clone()).collect())
```
yannham, line 29:
```suggestion
            Some(fields.keys().cloned().collect())
```

yannham (changes requested):
I think I'm not sure to understand what the new environments - `lin_env` and `env` - are referring to (or, put differently, to which scope they correspond).

My understanding is that the **Linearization** is a data structure akin to the AST, used by the sever to answer LSP queries. During typechecking, we are building this structure step by step. As such, there is only one linearization, which is threaded through the various function calls of the typecheck module.

The linearizer, in practice implemented by `AnalysisHost`, corresponds to local state. In particular, when walking a node like `(let x = f in x) (let x = arg in x)`, two different analysis hosts with their own environment are spawned (through the `scope()` function). They will have their own definition for `x`, each. But they will operate on the same linearization.

Thus, it makes sense for `AnalysisHost` to carry around an `Environment`: it corresponds to the environment at the current point of the AST. It is related to a scope. However, you are moving an environment from this local structure to a global structure, the linearization. To which scope correponds `lin_env`?

My understanding is that `Building`'s `env` and `Completed`'s `lin_env` are just the environment of the last added item. Which is probably not what we want. First, just when building the linearization, we may end up with additional stuff in the environment: for example, when walking `(let y = f in y) (let x = arg in ..)`, I'm pretty sure the implementation of this PR will have a `{y = f, x = arg}` environment when walking the `..`, but `y` wasn't even defined in this scope. I suspect everything still works fine, but a bit by accident: because Nickel checks early for undefined variable, you can't refer to this `y`. If you define a new `y`, it will shadow the old one in the environment. Still, it's not the right environment.

The second issue is that you use this environment to find e.g. a record to do completion. I suspect it works in simple cases because we are writing at the end of the file, so the environment of the last item added is actually the current environment. However, if you edit right into the middle of a file, you don't necessarily get the right thing:

Capture d’écran du 2022-11-02 15-19-18

I must admit the current notion of scope, analysis host, scope ids and so on is quite convoluted and hard to understand. We'll have to simplify all of that at some point. For now, I can see two possible solutions:

1. Don't use the environment, but only rely on scope. For each location in the file, you can get a vector of item id corresponding to the linearization items in scope (this is how simple completion is implemented). You can then traverse this list, look for declarations with a matching ident, and do what you currently do if that turns out to be a record.
2. Store an environment with linearization items. That's probably going to take up some memory, but an environment is just an `Rc`, and it should at least make things quick. If we do that, we can probably get rid of scope altogether (listing the keys of the environment would replace `scope`), but we can do that in a forthcoming PR.

In fact, I think 2. is both way simpler to do for you and will actually _spare_ memory once we've eliminated `scope` (that we need to store and clone today anyway, and as opposed to `Env`, it's not behind an `Rc`). Concretly, I would revert your change (put the environment back in `AnalysisHost`), add an `env` field to `LinearizationItem`, store the local environment there. Then, when asking for record completion, you are already finding the smallest enclosing linearization item: from it, you will get your env directly.

ebresafegaga:
Thank you, all fixed!

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +232,209 @@ pub fn handle_completion(
...
     .unwrap();
 
-    let locator = (file_id, ByteIndex(start as u32));
+    // -1 because at the current cursor position, there is no linearization item there,
+    // so we try to get the item just before the cursor.
+    let locator = (file_id, ByteIndex((start - 1) as u32));
     let linearization = server.lin_cache_get(&file_id)?;
+    // Cloning to avoid mutable borrow and shared borrow at the same time,
+    // which won't compile
+    // let linearization = linearization.clone();
```
yannham, line 249:
```suggestion
```

lsp/nls/src/requests/completion.rs
```diff
@@ -21,42 +232,209 @@ pub fn handle_completion(
...
 
+    let item = linearization.item_at(&locator).cloned();
```
yannham, line 251: Why do we need a `cloned` here? It seems we are only doing immutable borrows below.

src/identifier.rs
```diff
@@ -114,7 +114,7 @@ mod interner {
     use typed_arena::Arena;
 
     /// A symbol is a correspondance between an [Ident](super::Ident) and its string representation stored in the [Interner].
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
yannham, line 117: Why do you need a default for `Symbol`? I fear it won't do what you expect. Symbols should be strictly generated by the string interner, they are indices inside an allocation table. Here deriving `Default` will give you index `0`, so this will probably just point to the first identifier ever allocated since Nickel started, which will probably not be empty. If you want `Ident` to implement `Default` (although it sounds a bit fishy, do you need that for tests?), I think you have to implement it by hand as `Ident::from("")` or something like that.

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,227 @@
...
+                return Ok(Vec::new());
+            }
+        }
+        Some(..) | None => {
+            // This is also record completion, but it is in the form
+            // <record var>.<partially-typed-field>
+            // we also want to give completion based on <record var> in this case.
+            if let Some(name) = get_identifier_before_field(source) {
+                let name = Ident::from(name);
+                if let Some(name) = item.env.get(&name).cloned() {
```
yannham, line 183:
```suggestion
                if let Some(name) = item.env.get(&name).copied() {
```

It's a nitpick, as both function are doing exactly the same thing. Still, it makes clear the fact that this is a very cheap operation: I was initially wondering while you need to clone a name, as you will probably just is it as a key or an index.

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,227 @@
...
+    let in_scope = match trigger {
+        // Record Completion
+        Some(server::DOT_COMPL_TRIGGER) => {
+            let name = get_identifier_before_dot(source).ok_or(ResponseError {
+                code: ErrorCode::InternalError as i32,
+                message: "Couldn't get identifier for record completion".to_owned(),
+                data: None,
+            })?;
+            let name = Ident::from(name);
+            if let Some(name) = item.env.get(&name).cloned() {
```
yannham, line 171:
```suggestion
            if let Some(name) = item.env.get(&name).copied() {
```

(see below)

yannham, line 171: I find a bit strange to call something `name` which has type `usize`. Especially when it was previously an ident. Wouldn't `id` be a better name for it?

lsp/nls/src/requests/completion.rs
```diff
@@ -1,16 +1,227 @@
...
+) -> Vec<(Vec<Ident>, Types)> {
+    linearization
+        .get_in_scope(item)
+        .iter()
+        .filter_map(|item| {
+            let (ty, _) = linearization.resolve_item_type_meta(item);
+            match (&item.kind, ty) {
+                // Get record fields from static type info
+                (TermKind::Declaration(..), Types(AbsType::Record(row))) if name == item.id => {
+                    Some((extract_ident(&row), item.ty.clone()))
```
yannham, line 137: Why do you need to `clone` here if `extract_ident` is taking a reference?

yannham (commented):
Almost there! I think the `Default` implementation of `Ident` may be unsound. Otherwise, I may not have caught them all, but there are still `clones` that don't look like they are required. You may want to do a `cargo clippy` (and ignore suggestion not related to your change) to see if there are interesting suggestions (which include removing unnecessary clones).

lsp/nls/src/requests/completion.rs
ebresafegaga: This `clone` is for the type, not the boxed row type that is passed as an argument to `extract_ident`.

src/identifier.rs
ebresafegaga: Thanks, fixed!

yannham: @gaga Did you push your changes? I still see a `Default` implementation for `Symbol` and `Ident`.

lsp/nls/src/requests/completion.rs
yannham: Ah yes, sorry, I misread the parentheses.

yannham (approved):
LGTM. Really happy to see record completion finally landing :tada: !

ysndr:
great work 👏🏼

PR #869, opened 2022-10-12
Rename DynRecord to Dict

ebresafegaga:
Currently, dictionaries types in Nickel are represented with something called a dynamic record: `DynRecord(Ty)` . This maps `Str` keys to the type `Ty`. This in not particularly bad, but it would be better to just have it be called a *dictionary*: `Dict(Ty)`, to better portray it's meaning.

This change renames all occurrences of `DynRecord(..)` to `Dict(..)`

### DIFF
--- a/doc/manual/typing.md
+++ b/doc/manual/typing.md
@@ -190,9 +190,7 @@ The following type constructors are available:
   pair.fst : Num
   ```
 
-- **Dynamic record**: `{_: T}`. A record whose field
-  names are statically unknown but are all of type `T`.  Typically used to model
-  dictionaries.
+- **Dictionary**: `{_: T}`. A dictionary mapping string keys to `T`.
 
   Example:
 
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -487,7 +487,7 @@ fn walk_type<L: Linearizer>(
        }
        AbsType::Flat(t) => walk(state, ctxt, lin, linearizer, t),
        AbsType::Enum(ty2)
-       | AbsType::DynRecord(ty2)
+       | AbsType::Dict(ty2)
        | AbsType::StaticRecord(ty2)
        | AbsType::Array(ty2)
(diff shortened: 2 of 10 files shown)


yannham (commented):
While we're at it, could we also rename `StaticRecord` to simply `Record`?

ebresafegaga:
> While we're at it, could we also rename `StaticRecord` to simply `Record`?

Yes, I will make a separate PR for that once this one is merged.

doc/manual/typing.md
```diff
@@ -190,9 +190,7 @@ The following type constructors are available:
   pair.fst : Num
   ```
 
-- **Dynamic record**: `{_: T}`. A record whose field
-  names are statically unknown but are all of type `T`.  Typically used to model
-  dictionaries.
+- **Dictionary**: `{_: T}`. A dictionary mapping string keys to `T`.
```
matthew-healy, line 193: Nitpick: "[...]to values of type `T`." might be _slightly_ clearer.

src/typecheck/mod.rs
```diff
@@ -767,8 +767,8 @@ fn type_check_<L: Linearizer>(
                 ty.clone()
             };
 
-            if let TypeWrapper::Concrete(AbsType::DynRecord(rec_ty)) = root_ty {
-                // Checking for a dynamic record
+            if let TypeWrapper::Concrete(AbsType::Dict(rec_ty)) = root_ty {
+                // Checking for a dictionaries
```
matthew-healy, line 771: Typo: `dictionary`

matthew-healy (approved):
Looks good! Two very minor comments, but overall this seems like a sensible change.

doc/manual/typing.md
yannham: I think the new version is actually losing some information. There's no notion of "dictionaries" among primitive values (records are the only notion of key-value maps), so if I'm a beginner, I could think that there is something else out there, called dictionary, with a potentially different syntax.

I would suggest to just change

>Dynamic record

to Dictionary and keep the original sentence.

yannham (approved):

PR #872, opened 2022-10-13
[Fix] reset state for thunks on the stack

### LINKED ISSUE #868: Spurious infinite recursion in the REPL
**Describe the bug**
The REPL keeps too much state between evaluations. When an evaluation throws an error, long lived thunks created by let-bindings may remain in the blackholed state. If they are evaluated again, the evaluator thinks it encountered an infinite recursion.

**To Reproduce**
```
nickel repl <<EOF
let contract = { var | { foo | Str | optional } } & { var | { bar | Num | optional } }
{} | contract
{ var = {} } | contract
EOF
```
This is almost certainly not a minimal example. I haven't found a shorter one yet, though.

**Expected behavior**
No inifinite recursion error. Thunks touched by the evaluator need to be reset into an unevaluated state if an error occurs.

**Environment**
 - Version of the code: a0025284af0aa1830238d152869c03b14d31b250

**Additional context**
Thanks to @yannham for a quick initial diagnosis!

dpl0a:
Changed `VirtualMachine::reset` to also reset the state of all thunks in the stack to `Suspended`.
Fixes #868

### DIFF
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -158,8 +158,7 @@ impl<R: ImportResolver> VirtualMachine<R> {
     pub fn reset(&mut self) {
         self.eval_mode = Default::default();
         self.call_stack.0.clear();
-        self.stack.reset_thunks();
-        self.stack.clear(); //Do we even need to call this at this point?
+        self.stack.reset();
     }
 
     fn set_mode(&mut self, new_mode: EvalMode) {
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -158,8 +158,8 @@ impl<R: ImportResolver> VirtualMachine<R> {
     pub fn reset(&mut self) {
         self.eval_mode = Default::default();
         self.call_stack.0.clear();
-        reset_thunks(&mut self.stack);
-        self.stack.clear();
+        self.stack.reset_thunks();
+        self.stack.clear(); //Do we even need to call this at this point?
     }
 
     fn set_mode(&mut self, new_mode: EvalMode) {
@@ -176,10 +176,6 @@ impl<R: ImportResolver> VirtualMachine<R> {
 
     pub fn import_resolver_mut(&mut self) -> &mut R {
         &mut self.import_resolver
-    }
-
-    pub fn stack_mut(&mut self) -> &mut Stack {
-        &mut self.stack
     }
 
     /// Evaluate a Nickel term. Wrapper around [eval_closure] that starts from an empty local
@@ -764,12 +760,6 @@ pub fn env_add(env: &mut Environment, id: Ident, rt: RichTerm, local_env: Enviro
 fn update_thunks(stack: &mut Stack, closure: &Closure) {
     while let Some(thunk) = stack.pop_thunk() {
         thunk.update(closure.clone());
-    }
-}
-
-fn reset_thunks(stack: &mut Stack) {
-    while let Some(thunk) = stack.pop_thunk() {
-        thunk.reset_state();
     }
 }
 
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -154,10 +154,11 @@ impl<R: ImportResolver> VirtualMachine<R> {
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/eval/mod.rs
```diff
@@ -177,6 +178,10 @@ impl<R: ImportResolver> VirtualMachine<R> {
         &mut self.import_resolver
     }
 
+    pub fn stack_mut(&mut self) -> &mut Stack {
```
yannham, line 181: Wouldn't that be easier to implement stack unwiding directly in the `Stack` or `VirtualMachine`? We wouldn't have to expose the internals of `VirtualMachine` publicly.

src/eval/mod.rs
```diff
@@ -762,6 +767,12 @@ fn update_thunks(stack: &mut Stack, closure: &Closure) {
     }
 }
 
+fn reset_thunks(stack: &mut Stack) {
+    while let Some(thunk) = stack.pop_thunk() {
+        thunk.reset_state();
+    }
+}
```
yannham, line 774: I fear that won't be enough. The stack could very much look like:

```
ThunkUpdateFrame
Op2Cont
...
ThunkUpdateFrame
```

The second thunk won't be released. I think you have to iterate through all the stack until the end, and treat thunk update frames, and not only the first consecutive thunks.

dpl0a: Hmm true, I somehow overlooked this

src/eval/mod.rs
dpl0a: I'll look into it

src/eval/mod.rs
dpl0a: Should work correctly now

src/eval/stack.rs
```diff
@@ -132,6 +132,20 @@ impl Stack {
         self.0.clear();
     }
 
+    /// Pops all items in the stack and resets the state of the thunks it encounters.
+    pub fn reset_thunks(&mut self) {
```
yannham, line 136: I like recursion as much as the next functional programmer :stuck_out_tongue: but this will use stack space (the binary process stack, not our heap-allocated `Stack`) linear in the size of `stack` (our `Stack`, `self`).

More idiomatic Rust would be:

```rust
    pub fn reset_thunks(self) {
        for elt in self.0 {
            if let Some(Marker::Thunk(thunk)) = elt {
                thunk.reset_state();
             }
        }
    }
```

src/eval/mod.rs
```diff
@@ -154,11 +154,12 @@ impl<R: ImportResolver> VirtualMachine<R> {
...
     }
 
-    /// Reset the state of the machine (stacks and eval mode) to prepare for another evaluation round.
+    /// Reset the state of the machine (stacks, eval mode and thunk state) to prepare for another evaluation round.
     pub fn reset(&mut self) {
         self.eval_mode = Default::default();
         self.call_stack.0.clear();
-        self.stack.clear();
+        self.stack.reset_thunks();
+        self.stack.clear(); //Do we even need to call this at this point?
```
yannham, line 162: Well, you either clear inside `reset_thunks` (but then maybe the name is not really adapted anymore, because it's doing more?) as you do now, and you don't have to call to `clear()`. Or you make `reset_thunks` more atomic, by just iterating immutably on the stack, and keep the separate `clear()` function. I don't have a strong opinion, beside the naming issue.

src/eval/stack.rs
dpl0a: Your solution is way better. Actually I wanted to write something like this but somehow I was convinced this would cause some problems with moving out of self.0, but self doesn't need to be mutable so there's no problem. My bad for not realizing that (and not trying it out before writing a non-idiomatic solution)

dpl0a: Yeah but doesn't this move out stuff behind a mutable reference?

yannham: Yes, you need to adapt for the borrowing you want.

yannham (approved):

PR #874, opened 2022-10-14
Fix bad lexing of enum tags

### LINKED ISSUE #756: Bad lexing of enum tags
**Describe the bug**
Enum tags are lexed in a strange way. Although they start with the backtick character `` ` ``, the interpreter currently refuses to parse an enum tag which is a keyword, e.g. `` `if ``, while that should be totally ok. Another strangeness is that we can add whitespaces between the backtick and the identifier, which is non intuitive.

**To Reproduce**

```
nickel> `if
error: unexpected token
  ┌─ repl-input-1:1:2
  │
1 │ `if
  │  ^^

nickel> `      foo
`foo

nickel>
```

**Expected behavior**
An enum tag should allow arbitrary keyword, as it can be distinguished syntactically by the starting backtick. It also shouldn't allow whitespaces between the backtick and the identifier. Technically, the backtick is currently lexed as a standalone lexeme, followed by either an identifier or a quoted string. It should probably be lexed as one enum tag token instead.  

**Environment**
 - Version of the code: d952614e4c7b7c12f2d7cbfa141be9fa0a06729c

matthew-healy:
Enum tag parsing previously disallowed tags whose names overlapped with language keywords (e.g. `` `if `` was rejected), while also allowing tags to have an arbitrary amount of space between the backtick and the identifier.

This commit fixes both of these issues by handling raw identifier tags and string tags in slightly different ways.

- For raw tags, the entire enum tag is now a single `RawEnumTag` token. This means that the identifier part can never be confused with a keyword during parsing, while also ensuring that there is no extra whitespace between the backtick and identifier.
- For string tags, we lex a `StringEnumTagBegin` token, which is just a backtick followed by a double quote. This puts the lexer into string mode, and we parse the remainder of the identifier as a regular static string.

Note that this means that multiline string enum tags are now rejected. Indeed we explicitly return a lexer error when we encounter a backtick followed by an attempt to open a multiline string, as without handling this case we create the possiblility for a term like `` `m%"test"%m`` to successfully parse, but with a result that is almost certainly not what the user intended.

Merging this will close #756.

### DIFF
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -506,23 +506,18 @@ ChunkExpr: StrChunk<RichTerm> = Interpolation <t: WithPos<Term>> "}" => StrChunk
 
 Interpolation = { "%{", "multstr %{" };
 
-StaticString: String = StringStart <s: ChunkLiteral?> StringEnd => s.unwrap_or_default();
+// A construct which looks like a string, but is generic over its delimiters.
+// Used to implement `StaticString` as well as `StringEnumTag`.
+DelimitedStaticString<Start, End>: String = Start <s: ChunkLiteral?> End => s.unwrap_or_default();
 
-EnumTag: Ident = "`" <EnumTagContent> => <>;
+StaticString = DelimitedStaticString<StringStart, StringEnd>;
 
-EnumTagContent: Ident = {
-    <Ident>,
-    <StaticString> => <>.into(),
-    // Allow tags of the form `Str, `Bool, etc. This is ad-hoc and should
-    // eventually be replaced with a proper lexing of enum tags, but this is not
-    // a trivial task. For now, to make builtin.typeof usable, we just
-    // special-case builtin types.
-    "Dyn" => "Dyn".into(),
-    "Num" => "Num".into(),
-    "Bool" => "Bool".into(),
-    "Str" => "Str".into(),
-    "Array" => "Array".into(),
-}
+StringEnumTag = DelimitedStaticString<"`\"", "\"">;
+
+EnumTag: Ident = {
+  "raw enum tag" => <>.into(),
+  <StringEnumTag> => <>.into(),
+};
 
 ChunkLiteralPart: ChunkLiteralPart<'input> = {
     "str literal" => ChunkLiteralPart::Str(<>),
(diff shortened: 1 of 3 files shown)


src/parser/grammar.lalrpop
```diff
@@ -565,7 +554,8 @@ UOp: UnaryOp = {
 };
 
 SwitchCase: SwitchCase = {
-    "`" <id: EnumTagContent> "=>" <t: Term> => SwitchCase::Normal(id, t),
+    <id: EnumTag> "=>" <t: Term> => SwitchCase::Normal(id, t),
+    // "`" <id: EnumTagContent> "=>" <t: Term> => SwitchCase::Normal(id, t),
```
yannham, line 558:
```suggestion
```

src/parser/grammar.lalrpop
```diff
@@ -508,21 +508,10 @@ Interpolation = { "%{", "multstr %{" };
...
-    // special-case builtin types.
-    "Dyn" => "Dyn".into(),
-    "Num" => "Num".into(),
-    "Bool" => "Bool".into(),
-    "Str" => "Str".into(),
-    "Array" => "Array".into(),
-}
+EnumTag: Ident = {
+  "raw enum tag" => <>.into(),
+  "`\"" <s: ChunkLiteral?> "\"" => s.unwrap_or_default().into(),
```
yannham, line 513: It's not very important, but I wonder if we can factor this out using the `StaticString` rule, which is more or less the same. The issue is that the start and end rule for the literal are different, but we can parametrize that with a macro rule in LARLPOP:

```rust
StaticString<Start, End> : String = Start <s: ChunkLiteral?> StrigEnd => s.unwrap_or_default();
```

Then replace all current occurrences of `StaticString` with `StaticString<StringStart, StringEnd>` and the one above with

```rust
StaticString<"`\"", "\""> => <>.into(),
```

Or use a different name `StaticStringDelim` and just write `StaticString = StaticStringDelim<StringStart, StringEnd>`. I'm not sure any of these are better than the current one, but once again, it's rather to let you know that those are also possibilities :slightly_smiling_face: 

It wouldn't change much the complexity of the current code, but could be a tad useful if we decide that multiline string delimiters for `import` also don't make sense, using e.g. `StaticStringDelim<"\"", "\"">` instead of `StaticString`. It's also more abstraction for something quite simple, though.

yannham (approved):

src/parser/grammar.lalrpop
matthew-healy: Thank you! The duplication was bothering me, but it hadn't occurred to me that I could just factor it out a macro. Will definitely do this before I merge. :pray:

Issue #877, opened 2022-10-18
Improve LSP record completion


ebresafegaga:
With #867, we have basic record completion. However, this can be improved in the following areas:

* [x] Record completion for imports (e.g `let r = import "r.ncl" in r[.]`) #993 
    * [x] Record completion for `stdlib` terms (e.g `array[.]` should give correct completion terms) #918
* [x] Nested record completion (e.g `let x = {foo.bar.baz = 1} in x.foo[.]` should get `baz`, `let f : {a:{b:{c:{d:Str}}}} -> Str = fun x => x[.] in f`) #909
  * [x] Quoted fields should be supported too, e.g. `let x = {"fo京o" = {bar = 42}} in x."fo京o".<ctrl space>` #913 
* [x] Use contract metadata to give completions (e.g ` let x | {foo : Num} = null in x[.]`  and `let x | {foo | Num} = null in x[.]`)  #914 
* [x] Recursive record (e.g `{ foo.bar = 1,  baz = foo[.]  }`) #1088
* [x] Record terms should give completion (e.g `({ a = 10})[.]` should get a completion or, more generally, 
      `(<any expr that we know evals to a record>)[.]` )

Note:  `[.]` represents the cursor, after the `.` key has been pressed

PR #878, opened 2022-10-18
Improve type inference for records in `apparent_type`

ebresafegaga:
This PR adds support for record type inference. 

Currently, if we have the term: `let lib : _ = import "lib.ncl" in lib`, and say `lib.ncl` contains: 
`{  name = "lib" }`;  the type-checker infers that `lib` has the type `Dyn`. The type-checker can do more work to infer a more accurate type: `{ name: Str }`, and that is what this change fixes.

This change will be useful to improve LSP record completion (#877)

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -1017,6 +1017,8 @@ impl From<ApparentType> for Types {
 /// - if `bound_exp` is an array, return `Array Dyn`.
 /// - if `bound_exp` is a resolved import, return the apparent type of the imported term. Returns
 ///   `Dyn` if the resolver is not passed as a parameter to the function.
+/// - if `bound_exp` is a record, recursively find the apparent type for each of the record fields
+///   and return it as a Record type with those types as its fields.
 /// - Otherwise, return an approximation of the type (currently `Dyn`, but could be more precise in
 ///   the future, such as `Dyn -> Dyn` for functions, `{ | Dyn}` for records, and so on).
 pub fn apparent_type(
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -1047,6 +1047,21 @@ pub fn apparent_type(
             .and_then(|envs| envs.get(id).cloned())
             .map(ApparentType::FromEnv)
             .unwrap_or(ApparentType::Approximated(Types(AbsType::Dyn()))),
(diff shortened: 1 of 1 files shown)


matthew-healy (commented):
I left a small comment, but apart from that I think this looks good. 

I'll caveat that by saying that's based on the assumption that returning a more accurate type from `apparent_type` won't cause problems anywhere, which _feels_ reasonable, but I didn't completely verify because it's called in a few different places.

This also got me thinking about how we might test behaviours like this, which aren't strictly part of the language surface, but do impact the user experience. I don't think anything needs to change in this PR, but if you have any ideas for, e.g., how to prevent regressions in parts of code the LSP depends on, I'd be interested to hear them.

ebresafegaga:
Looking at it now, I believe this should be done in the LSP too, but currently there is no way to get access to the original term (in the LSP) after the type-checker has checked it. The LSP has access to only the `Linearization` , which is just a linear data structure, respecting the original scope of the term, containing the respective type for the term and each of the sub-terms, gotten from the type-checker. I'm not entirely sure how this can be done from the LSP.

yannham:
Well, you're right that can't do that from the LSP _per se_. One solution is to do it from the typechecking code, but only for the linearization. Concretly, you could have something like:

```rust
// Inside typecheck, or maybe a submodule like typecheck::lsp or whatever
fn determine_record_structure(...) -> RecordStructure {
  // do whatever you need to with access to the whole typing context, term, etc.
}

//inside typecheck
   // The API would accept a closure here, such that for the linearizer stub used in normal execution mode outside of the LSP, the function is not called at all and the compiler will probablyt remove the whole call altogether, letting the behavior and the performance of the normal nickel binary unchanged. A bit like the `or_else` style functions.
   linearizer.set_record_structure(|| determine_record_structure(...));
  // Another solution is to pass the required information to set_record_structure, to move as much logic as possible to the LSP implementation
  linearizer.set_record_structure(state, ctxt, term, etc...)
```

ebresafegaga:
Nice! I'll go with the closure API.

yannham:
@ebresafegaga if we agree to do a separate function that doesn't impact typechecking and type inference, then we can probably close this PR (or repurpose it, if you really want to, closing is really simpler) and reopen one that do this record structure determination?

ebresafegaga:
The change will be needed to close item 1 here #877

PR #882, opened 2022-10-24
Fix `ln -S` mistake in documentation

sir4ur0n:
The option was `-S` (suffix) rather than `-s` (symbolic).

Also, switching to the explicit option name for clarity

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ highlighting and NLS.
 1. *(optional)* make a symbolic link to the executable:
 
   ```console
-  ln -S nickel target/debug/nickel
+  ln --symbolic target/debug/nickel
   ```
 
 ### Tests


sir4ur0n:
I have been informed I can merge this PR myself, but I will take this opportunity to investigate and fix https://github.com/tweag/nickel/issues/883 so keeping open for now

yannham:
@sir4ur0n Shouldn't we do that in a separate PR? This one looks self-contained, is ready to merge and don't look really related to the CI failure.

sir4ur0n:
> Shouldn't we do that in a separate PR? This one looks self-contained, is ready to merge and don't look really related to the CI failure.

Most likely fixing the CI will not require changing anything in the Git-versioned files; instead it will require a change in the Github repo settings.

This PR will not be modified, but it will be used to let Rok relaunch the CI and check that it passes, before we merge it

yannham:
> This PR will not be modified, but it will be used to let Rok relaunch the CI and check that it passes, before we merge it.

Ah, I got that wrong. Gotcha

PR #886, opened 2022-10-25
Update recursive records to accept inherited fields

francois-caddet:
Inherit keyword for nix compatibility.
Because nickel will probably use an similar feature, we introduced it as recursive record field.


yannham (commented):
I'm not sure we need this to be an actual primop. If it's just desugaring , we can do that in an helper function, that we will need anyway for the similar feature in Nickel?

yannham:
This work has diverged too much from the nix branch, let alone the master branch. @francois-caddet has left the Nickel team. I think it'll just be simpler to start from scratch.

PR #896, opened 2022-10-27
Allow type inference for record terms in the LSP

ebresafegaga:
This PR adds support for record type inference in the LSP. 

Currently, if we have the term: `let lib : _ = import "lib.ncl" in lib`, and say `lib.ncl` contains: 
`{  name = "lib" }`;  the type-checker infers that `lib` has the type `Dyn`. The type-checker can do more work to infer a more accurate type: `{ name: Str }`, and that is what this change fixes. 

Note that this type inference only occurs when we are type-checking a term to be used in the LSP.

See also #878

### DIFF
--- a/src/typecheck/lsp.rs
+++ b/src/typecheck/lsp.rs
@@ -1,4 +1,4 @@
-//! Special typing analysis that should be done when we run the type checker in LSP mode
+//! Special typing analysis utils that should be done when we run the type checker in LSP mode
 
 use crate::{
     cache::ImportResolver,
@@ -8,6 +8,7 @@ use crate::{
 
 use super::{replace_wildcards_with_var, ApparentType, Context, Environment, State, TypeWrapper};
 
+/// Similar to [`super::apparent_type`], but it adds support for type inference on record terms.
 pub fn apparent_type(
     t: &Term,
     env: Option<&Environment>,
@@ -60,6 +61,8 @@ pub fn apparent_type(
     }
 }
 
+/// The LSP's version of [`super::binding_type`], but it delegates to a local type inference function
+/// that allows type inference for record terms.
 pub fn binding_type(state: &mut State, t: &Term, ctxt: &Context, strict: bool) -> TypeWrapper {
     let ty_apt = apparent_type(t, Some(&ctxt.type_env), Some(state.resolver));
 
--- a/src/typecheck/lsp.rs
+++ b/src/typecheck/lsp.rs
@@ -1,0 +1,73 @@
+//! Special typing analysis that should be done when we run the type checker in LSP mode
+
+use crate::{
+    cache::ImportResolver,
+    term::{Contract, MetaValue, Term},
+    types::{AbsType, Types},
+};
+
+use super::{replace_wildcards_with_var, ApparentType, Context, Environment, State, TypeWrapper};
+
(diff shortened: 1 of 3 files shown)


src/typecheck/lsp.rs
```diff
@@ -0,0 +1,76 @@
+//! Special typing analysis utils that should be done when we run the type checker in LSP mode
```
sir4ur0n, line 1: General concern with this PR and feature: why is this extra inference only done for LSP?
E.g. the exact same problem happens with the REPL:
```
> cat lib.ncl
{ name = "lib" }
> nickel repl
nickel> :typecheck import "lib.ncl"
Ok: Dyn
```

I am also concerned with the LSP typechecker diverging from Nickel's typechecker for a few reasons:
* double maintenance
* risk of introducing bugs (there is code duplication, so e.g. if someone modifies the `Term::Num(_)` case in `typecheck/mod.rs`, one MUST also do the same change in `typecheck/lsp.rs`
* unclear separation of concerns: the LSP is a different code base, but now we have LSP-specific code in the Nickel codebase

I am holding off on reviewing the PR until this concern is answered

sir4ur0n (changes requested):

src/typecheck/lsp.rs
ebresafegaga: Well, I originally implemented this on the type-checker, but there were issues regarding Nickel's semantics. See [this](https://github.com/tweag/nickel/pull/878#issuecomment-1282592098). We just need this to be able to perform completion for imports and stdlib terms. 

For those concerns, having good tests for the LSP might help, but you're right.

sir4ur0n: All this context - why it's done in the type checking code even though it's LSP-specific; why the code is duplicated rather than enhanced; why we don't want to typecheck imports in regular Nickel; etc. - is currently only present in the Github comment. It is not present in the code or documentation of this PR, meaning that future readers will have no idea why this is done this way.

If we want to stick with this design, then the rationales for this choice have to be documented. We must leave trails for future readers to understand the past decisions.

I don't have a strong opinion on whether it's directly documented in the code, or if the code instead refers to an [Architectural Decision Record](https://github.com/joelparkerhenderson/architecture-decision-record) where this is documented.

yannham (changes requested):
We probably don't want the LSP to infer different types than the typechecker; I fear this may end up being very confusing for the user. The interpreter could throw an error while the LSP would accept this file as valid.

The original suggestion, I think, wasn't to change anything about type inference, but to use the additional information we have at typechecking time to help record completion. A possible approach is to have a dedicated `RecordStructure`, or whatever you call it, stored with declaration items in the Linearization. You would use all the information at hand to build that data:

- the shape of the term
- type annotation
- inferred type
- contract annotation
- etc.

As an aside, it feels like this code should be coupled with the typechecker, not to the LSP. It should only be called/linked to for the LSP, but does some "low-level" manipulation on the types and terms. Usually, all the LSP does is rather to interact directly with the linearization: building the linearization is done on the Nickel side.

ebresafegaga:
Nice, this makes sense. Thanks!

ebresafegaga:
Closing this in favour of a future PR that would add auto completion for import terms.

PR #899, opened 2022-10-28
Add rust-analyzer to devShells

matthew-healy:


flake.nix
```diff
@@ -185,7 +185,7 @@
           buildInputs =
             [ rust ]
             ++ missingSysPkgs
-            ++ (if !isDevShell then [ cargoHome ] else [ ]);
+            ++ (if !isDevShell then [ cargoHome ] else [ pkgs.rust-analyzer ]);
```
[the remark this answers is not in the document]
yannham: Yes, I think what Matthew said is correct. I guess the difference is that cargo, clippy and rustfmt are part of the rust distribution, and provided as such directly by the `rust` package correctly configured, while rust-analyzer is still a separate tool.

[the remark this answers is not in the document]
torhovland: You're right. I tested it.

sir4ur0n, line 188: I know my question is unrelated to your change, but while we're reviewing this line:
Why do we add `cargoHome` only if not in a dev shell? What happens if we are in a dev shell?

No matter the answer to that question, can we please add a comment to explain that to future readers? :pray:

sir4ur0n, line 188: Minor remark: let's avoid double negative:
```suggestion
            ++ (if isDevShell then [ pkgs.rust-analyzer ] else [ cargoHome ]);
```

yannham (approved):

flake.nix
yannham: Let's do this in a separate PR.

PR #904, opened 2022-11-03
Nix/basic tests

francois-caddet:
a basic test suite for nix evaluation.
records and functions tests will be in another PR.

This one is expected to test all the others operators and constructions.


yannham (commented):
I would suggest to have a separate test suite for Nix. In practice, at least a separate test case (nix.rs) and to put the nix files in a different location than pass, which is already a bit overcrowded.

yannham (approved):

PR #905, opened 2022-11-03
fix string concat for Nix

francois-caddet:
fix string concat in Nix.

### DIFF
--- a/stdlib/compat.ncl
+++ b/stdlib/compat.ncl
@@ -1,6 +1,14 @@
 {
 compat | doc "Nix compatibility layer. should not be used by Nickel program."
 = {
+
+  # the adition in Nix can be a string concataination.
+  # this function handle this behaviour.
+  add = fun a b =>
+    if %typeof% a == `Str && %typeof% b == `Str
+    then a ++ b
+    else a + b,
+
   with =
     let AssertFound = fun l val => if val.found 
       then val.value
--- a/stdlib/compat.ncl
+++ b/stdlib/compat.ncl
@@ -2,7 +2,7 @@
 compat | doc "Nix compatibility layer. This library should not be used by Nickel program."
 = {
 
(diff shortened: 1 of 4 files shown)


stdlib/compat.ncl
```diff
@@ -1,6 +1,14 @@
 {
 compat | doc "Nix compatibility layer. should not be used by Nickel program."
```
yannham, line 2:
```suggestion
compat | doc "Nix compatibility layer. This library should not be used by Nickel program."
```

stdlib/compat.ncl
```diff
@@ -1,6 +1,14 @@
...
 = {
+
+  # the adition in Nix can be a string concataination.
```
yannham, line 5:
```suggestion
  # The addition in Nix can be a string concatenation.
```

stdlib/compat.ncl
```diff
@@ -1,6 +1,14 @@
...
+  # this function handle this behaviour.
```
yannham, line 6: This is not a very informative comment. Maybe something in the line of:

> Addition in Nix is overloaded to work both as number addition and string concatenation.
> There is no such operator in Nickel. This function implement the equivalent of the Nix primitive operator by dynamically
> dispatching between addition and concatenation, based on the runtime type of its arguments.

yannham (commented):
Nitpick: the `compat` module of stdlib should maybe mention Nix in its name. It's not obvious what "compatibility" it refers to. Maybe `nix_compat` ? Same for the stdlib module, I think.

francois-caddet:
> 

Yes, it'll better with something like `__nix_compat__".
The issue is that this "compat" lib is already in `nix/nickel` branch. So may we fix it in another PR?

yannham:
> So may we fix it in another PR?

Sounds good.

yannham (approved):

PR #906, opened 2022-11-04
Replace `Scope` with `Environment` in the LSP

ebresafegaga:
This is the final clean up of the work started in #867 to remove the `Scope` data structure, and replace it completely with an `Environment`

### DIFF
--- a/lsp/nls/src/linearization/completed.rs
+++ b/lsp/nls/src/linearization/completed.rs
@@ -1,10 +1,7 @@
 use std::collections::HashMap;
 
 use codespan::ByteIndex;
-use nickel_lang::{
-    term::MetaValue,
-    typecheck::linearization::{LinearizationState, Scope},
-};
+use nickel_lang::{term::MetaValue, typecheck::linearization::LinearizationState};
 
 use super::{
     building::ID,
@@ -15,19 +12,16 @@ use super::{
 #[derive(Debug, Default)]
 pub struct Completed {
     pub linearization: Vec<LinearizationItem<Resolved>>,
-    scope: HashMap<Scope, Vec<usize>>,
     id_to_index: HashMap<ID, usize>,
 }
 
 impl Completed {
     pub fn new(
         linearization: Vec<LinearizationItem<Resolved>>,
-        scope: HashMap<Scope, Vec<usize>>,
         id_to_index: HashMap<ID, usize>,
     ) -> Self {
         Self {
             linearization,
-            scope,
             id_to_index,
         }
     }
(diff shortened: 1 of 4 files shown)


lsp/nls/src/linearization/completed.rs
```diff
@@ -40,14 +34,12 @@ impl Completed {
...
-        let empty = Vec::with_capacity(0);
-        (0..scope.len())
-            .flat_map(|end| self.scope.get(&scope[..=end]).unwrap_or(&empty))
-            .map(|id| self.get_item(*id))
-            .flatten()
-            .collect()
+        env.iter()
+            .map(|(_, id)| self.get_item(*id))
+            .collect::<Option<Vec<_>>>()
+            .unwrap_or(Vec::new())
```
yannham, line 42: FYI, there is a slight variation function for that:

```suggestion
            .unwrap_or_default()
```

yannham (approved):
Nice PR :slightly_smiling_face: the logic is simpler and some obscure code just become useless.

lsp/nls/src/linearization/completed.rs
ebresafegaga: Nice, thanks!

lsp/nls/src/linearization/completed.rs
```diff
@@ -40,14 +34,12 @@ impl Completed {
...
-        let empty = Vec::with_capacity(0);
-        (0..scope.len())
-            .flat_map(|end| self.scope.get(&scope[..=end]).unwrap_or(&empty))
-            .map(|id| self.get_item(*id))
-            .flatten()
-            .collect()
+        env.iter()
+            .map(|(_, id)| self.get_item(*id))
+            .collect::<Option<Vec<_>>>()
+            .unwrap_or_default()
```
matthew-healy, line 42: Just want to make sure I'm understanding this correctly: we iterate through every term in the environment, and then find each element in the linearization, but if even a single one is missing, we return an empty vec?

Am I right in thinking that missing terms would have to be the result of an error in the construction of the linearization?

ebresafegaga: Wow, nice catch. What I want to do is accumulate the `Some(..)` just like `sequenceA` in Haskell. But I just tested it and it seems this one returns `None` if there is any `None` present in the list.

ebresafegaga: Is there a function to do this or do I have to implement it manually?

matthew-healy: Ah, in that case you probably want [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map).

lsp/nls/src/linearization/completed.rs
```diff
@@ -37,9 +37,8 @@ impl Completed {
         LinearizationItem { env, .. }: &LinearizationItem<Resolved>,
     ) -> Vec<&LinearizationItem<Resolved>> {
         env.iter()
-            .map(|(_, id)| self.get_item(*id))
-            .collect::<Option<Vec<_>>>()
-            .unwrap_or_default()
+            .filter_map(|(_, id)| self.get_item(*id))
+            .collect::<Vec<_>>()
```
matthew-healy, line 41: Tiny nitpick: you'll probably find you don't need the `::<Vec<_>>` turbofish anymore. Since we're not calling `unwrap_or_default` after `collect`, the function's return type should be enough to infer what the type of `collect` has to be here.

matthew-healy (approved):

lsp/nls/src/linearization/completed.rs
ebresafegaga: Yeah, thank!

lsp/nls/src/linearization/completed.rs
yannham: If the linearization MUST have an item at an index, I tend to think we should panic. Otherwise, it may let us slip bugs silently.

ebresafegaga: Yeah that’s true, because the environment cannot/shouldn’t have an id that is not in the linearization.

PR #909, opened 2022-11-07
Add support for nested record completion in the LSP

ebresafegaga:
This improves the LSP completion by adding support for nested records. 

This addresses item `2a` in #877

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -313,11 +313,9 @@ pub fn handle_completion(
 
 #[cfg(test)]
 mod tests {
-    use nickel_lang::position::TermPos;
-
-    use crate::linearization::Environment;
-
     use super::*;
+    use crate::linearization::Environment;
+    use nickel_lang::position::TermPos;
     use std::collections::{HashMap, HashSet};
 
     #[test]
@@ -389,12 +387,12 @@ mod tests {
         ));
         let ty = Types(AbsType::RowExtend(
             Ident::from("b"),
-            Some(Box::new(ty)),
+            Some(Box::new(Types(AbsType::Record(Box::new(ty))))),
             Box::new(Types(AbsType::RowEmpty())),
         ));
         let ty = Types(AbsType::RowExtend(
             Ident::from("a"),
-            Some(Box::new(ty)),
+            Some(Box::new(Types(AbsType::Record(Box::new(ty))))),
             Box::new(Types(AbsType::RowEmpty())),
         ));
         let mut path = vec![Ident::from("b"), Ident::from("a")];
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -87,16 +87,13 @@ fn find_fields_from_type(ty: &Box<Types>, path: &mut Vec<Ident>) -> Vec<Ident> {
     let current = path.pop();
     ty.iter_as_rows()
         .flat_map(|item| match (item, current) {
-            (RowIteratorItem::Row(ident, Some(ty)), Some(current)) if current == *ident => {
-                if let Types(AbsType::Record(ty)) = ty {
-                    find_fields_from_type(ty, path)
-                } else {
-                    find_fields_from_type(&Box::new(ty.clone()), path)
-                }
+            (RowIteratorItem::Row(ident, Some(Types(AbsType::Record(ty)))), Some(current))
+                if current == *ident =>
+            {
+                find_fields_from_type(ty, path)
             }
-            (RowIteratorItem::Row(..), Some(_)) => Vec::new(),
             (RowIteratorItem::Row(ident, _), None) => vec![ident.clone()],
-            (RowIteratorItem::Tail(..), _) => Vec::new(),
+            _ => Vec::new(),
         })
         .collect::<Vec<_>>()
 }
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -134,7 +134,7 @@ fn get_identifiers_before_field(text: &str) -> Option<Vec<String>> {
         .rev()
         .collect();
 
-    get_identifier_path(&text[..])
+    get_identifier_path(&text.as_str())
 }
 
 fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -36,7 +36,7 @@ fn find_fields_from_term_kind(
 ) -> Option<Vec<Ident>> {
     let item = linearization.get_item(id)?;
     match item.kind {
-        TermKind::Record(ref fields) if item.id == id => {
+        TermKind::Record(ref fields) => {
             if path.is_empty() {
                 Some(fields.keys().cloned().collect())
             } else {
@@ -50,9 +50,7 @@ fn find_fields_from_term_kind(
             ..
         }
         | TermKind::Declaration(_, _, ValueState::Known(new_id))
-        | TermKind::Usage(UsageState::Resolved(new_id))
-            if item.id == id =>
-        {
+        | TermKind::Usage(UsageState::Resolved(new_id)) => {
             find_fields_from_term_kind(linearization, new_id, path)
         }
         _ => None,
@@ -68,20 +66,19 @@ fn find_fields_from_contract(
 ) -> Option<Vec<Ident>> {
     let item = linearization.get_item(id)?;
     match &item.meta {
-        Some(MetaValue { contracts, .. }) if item.id == id => {
+        Some(MetaValue { contracts, .. }) => {
             contracts.iter().find_map(|contract| match &contract.types {
                 Types(AbsType::Record(row)) => Some(find_fields_from_type(&row, path)),
                 _ => None,
             })
         }
-        _ if item.id == id => match item.kind {
+        None => match item.kind {
             TermKind::Declaration(_, _, ValueState::Known(new_id))
             | TermKind::Usage(UsageState::Resolved(new_id)) => {
                 find_fields_from_contract(&linearization, new_id, path)
             }
             _ => None,
         },
-        _ => None,
     }
 }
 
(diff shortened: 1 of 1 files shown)


lsp/nls/src/requests/completion.rs
```diff
@@ -21,56 +21,85 @@ use crate::{
...
         },
         _ => None,
     }
 }
 
-/// Extract identifiers from a row type which forms a record.
-/// Ensure that ty is really a row type.
-fn extract_ident(ty: &Box<Types>) -> Vec<Ident> {
+/// Extract the fields from a given record type.
+fn find_fields_from_type(ty: &Box<Types>, path: &mut Vec<Ident>) -> Vec<Ident> {
```
matthew-healy, line 89: Rather than taking `&Box<Types>` this could just take `&Types`, since you can pass a `Box` where a reference is expected.

This should remove the need for the clone on line 97.

lsp/nls/src/requests/completion.rs
```diff
@@ -21,56 +21,85 @@ use crate::{
...
     ty.iter_as_rows()
-        .filter_map(|item| match item {
-            RowIteratorItem::Row(ident, _) => Some(ident.clone()),
-            RowIteratorItem::Tail(..) => None,
+        .flat_map(|item| match (item, current) {
+            (RowIteratorItem::Row(ident, Some(ty)), Some(current)) if current == *ident => {
+                if let Types(AbsType::Record(ty)) = ty {
+                    find_fields_from_type(ty, path)
+                } else {
+                    find_fields_from_type(&Box::new(ty.clone()), path)
```
matthew-healy, line 97: I'm not entirely sure I'm following the logic here - if `ty` wasn't a record, then either (1) `iter_as_rows` will return nothing or (2) it's an enum, in which case it doesn't have field names. Do we really need to do the recursive call, or can we just give up?

lsp/nls/src/requests/completion.rs
```diff
@@ -80,22 +109,37 @@ lazy_static! {
...
-/// Get the identifier before the `.`, for record completion.
-fn get_identifier_before_dot(text: &str) -> Option<String> {
+/// Get the string chunks that make up an identifier path.
+fn get_identifier_path(text: &str) -> Option<Vec<String>> {
+    // skip any `.` at the end of the text
     let text: String = text
         .chars()
         .rev()
         .skip_while(|c| *c == '.') // skip . (if any)
+        .collect::<String>()
```
matthew-healy, line 122: I assume you've added the `collect` call here because of [this issue](https://stackoverflow.com/a/69782688). However, we can avoid allocating the string twice by just cloning `text` immediately then repeatedly popping the last character of the string if it's a dot. E.g.

```rust
let text = {
    let mut text = text.to_string;
    while text.ends_with('.') {
         text.pop();
    }
    text
};
```

I've added the extra scope here to limit the mutability of the string to just that expression. I appreciate this feels a lot less FP-ish than what was here before, but I don't think it's too problematic to use a little bit of mutation here to avoid an entire string copy.

lsp/nls/src/requests/completion.rs
```diff
@@ -80,22 +109,37 @@ lazy_static! {
...
+    if result.is_empty() {
+        return None;
+    }
+    let path = result
+        .iter()
+        .rev()
+        .cloned()
+        .collect::<Vec<_>>()
+        .get(0)
+        .cloned()?;
```
matthew-healy, line 137: I think this chain can be simplified to:

```rust
let path = result
    .iter()
    .rev()
    .next()
    .cloned()?;
```

This avoids the need to clone any result other than the first one, and also means we don't need to collect everything into a vec - we can just take the "next" item from the reversed iterator.

lsp/nls/src/requests/completion.rs
```diff
@@ -80,22 +109,37 @@ lazy_static! {
...
+        .chars()
+        .rev()
         .collect();
 
-    let name = RE.find(&text)?;
-    Some(name.as_str().chars().rev().collect())
+    let result: Vec<_> = RE_SPACE.split(text.as_ref()).map(String::from).collect();
+    if result.is_empty() {
+        return None;
+    }
```
matthew-healy, line 130: You can remove these lines, since building an iterator on an empty vec is essentially a no-op, and `get` (or `next` if you follow my suggestion) will return `None`, which means the `?` will result in the whole function returning `None`.

lsp/nls/src/requests/completion.rs
```diff
@@ -106,7 +150,7 @@ fn get_identifier_before_field(text: &str) -> Option<String> {
         .rev()
         .collect();
 
-    get_identifier_before_dot(&text[..])
+    get_identifier_path(&text[..])
```
matthew-healy, line 153: Personally I'd find `text.as_str()` slightly clearer here, but it does essentially the same thing.

lsp/nls/src/requests/completion.rs
```diff
@@ -80,22 +109,37 @@ lazy_static! {
...
+        .cloned()
+        .collect::<Vec<_>>()
+        .get(0)
+        .cloned()?;
+    Some(path.split(".").map(String::from).collect())
 }
 
 /// Get the identifier before `.<text>` for record completion.
-fn get_identifier_before_field(text: &str) -> Option<String> {
+fn get_identifier_before_field(text: &str) -> Option<Vec<String>> {
```
matthew-healy, line 142: Maybe rename this to `get_identifiers_before_field` now that it returns a vec?

lsp/nls/src/requests/completion.rs
```diff
@@ -285,6 +333,68 @@ mod tests {
...
+                r##"let me : _ = { name = "gg", age = 23, a.b.c.d = 10 } in
+                    me.ca.cb"##,
+                vec!["me", "a", "b"],
+            ),
+        ];
+        for (input, expected) in tests {
+            let actual = get_identifier_path(input);
+            let expected: Option<Vec<_>> =
+                Some(expected.iter().cloned().map(String::from).collect());
+            assert_eq!(actual, expected)
```
matthew-healy, line 363: I tend to find it's helpful to name each of the test cases, and include that name in the assertion failure message, e.g.

```rust
("ident chain in let binding", "let person = {} in \n    a.b.c.d", vec!["a", "b", "c", "d"]),
// ...
for (case_name, input, expected) in tests {
// ...
assert_eq!(actual, expected, "test failed: {}", case_name)
// ...
```

This way when a test fails, it's immediately clear which actual case was problematic.

lsp/nls/src/requests/completion.rs
```diff
@@ -285,6 +333,68 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn test_extract_ident_with_path() {
+        let ty = Types(AbsType::RowExtend(
```
matthew-healy, line 369: We could potentially use a macro (similar to `mk_tyw_row`) here to simplify this construction. Totally up to you though, as the duplication only appears in one place currently.

matthew-healy (commented):
Overall I think this looks good! Nice work on the tests too. :grin: 

I left a few comments, let me know if you have any questions. Happy to approve once they're answered.

lsp/nls/src/requests/completion.rs
```diff
@@ -21,56 +21,85 @@ use crate::{
...
+fn find_fields_from_term_kind(
+    linearization: &Completed,
+    id: usize,
+    path: &mut Vec<Ident>,
+) -> Option<Vec<Ident>> {
     let item = linearization.get_item(id)?;
     match item.kind {
-        TermKind::Record(ref fields) if item.id == id => Some(fields.keys().cloned().collect()),
-        TermKind::Declaration(_, _, ValueState::Known(new_id))
+        TermKind::Record(ref fields) if item.id == id => {
```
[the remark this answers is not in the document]
ebresafegaga: Oh that can never happen. I think this is from the way these functions were previously implemented (searching through a vector). Nice catch, thanks!

lsp/nls/src/requests/completion.rs
ebresafegaga: Yeah, we don't need it. I was trying to check if it was possible to get a  case where `ty`  could be a `AbsType::Flat(..)`, representing a record contract.

ebresafegaga: Thanks!

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes, it is because of that issue 😅 This also looks good since the scope limited.

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes, I plan to use macros to abstract a lot of the tests in the future.

ebresafegaga:
All fixed 😁

matthew-healy (approved):

lsp/nls/src/requests/completion.rs
```diff
@@ -21,56 +21,79 @@ use crate::{
...
+    id: usize,
+    path: &mut Vec<Ident>,
+) -> Option<Vec<Ident>> {
     let item = linearization.get_item(id)?;
     match &item.meta {
-        Some(MetaValue { contracts, .. }) if item.id == id => {
+        Some(MetaValue { contracts, .. }) => {
             contracts.iter().find_map(|contract| match &contract.types {
-                Types(AbsType::Record(row)) => Some(extract_ident(&row)),
-                Types(AbsType::Flat(term)) => {
```
yannham, line 46: @ebresafegaga why do we get rid of the code that handles the case where the contract is a record contract (and not a record type)? It looks like we're removing a feature here.

Also, there may be a type annotation inside `meta.types`. In general, if you want **all** the annotations on a value, mixing contracts and the type annotation, you should do something like `meta.types.iter().chain(meta.contracts.iter())`

ebresafegaga: This was on purpose. I removed it completely, to be properly implemented with item `3` in #877. 

It should be able to handle a mixture of types and contracts: 
```nickel
let term | {a | {b : { c | { d : { f : Str } } }} } = { ... } in a. 
```
So, the implementation for item `3` will do this

PR #912, opened 2022-11-08
implementation of the "has fieldpath" operator of nix

francois-caddet:
This PR add support for `{<...>} ? <field.path>` syntax.

It depends on #1062 because it fixes an issue comming from `rnix v0.10` by updating `rnix` to the v0.11


yannham:
@francois-caddet The target branch is master, and the diff looks big. Did you mean to target the nix branch instead?

src/nix.rs
```diff
@@ -3,13 +3,37 @@ use crate::conversion::State;
...
+    match attr {
+        NixAttr::Ident(id) => Term::Str(id.to_string()).into(),
+        NixAttr::Str(s) => s.translate(state),
+        NixAttr::Dynamic(d) => d.expr().unwrap().translate(state),
+    }
+}
+
+fn path_rts_from_nix<T>(n: rnix::ast::Attrpath, state: &State) -> T
+where
+T: FromIterator<RichTerm>
```
yannham, line 33: It's strange to take a parameter `T` that only appears in the output (in fact I didn't know you could do that). I think writing an `impl` constraint is more idiomatic:
```suggestion
fn path_rts_from_nix<T>(n: rnix::ast::Attrpath, state: &State) -> impl FromIterator<RichTerm>
```

yannham (approved):

src/nix.rs
francois-caddet: We can drop the `<T>` in this case

francois-caddet: Actualy, using the `impl ...` syntax, rust complains that we have to know the type at this point.
The form `-> T where T: ...` probably work because the function is private.

yannham: Oh, I missed the `collect()` at the end. Now it all makes sense, and should be a `<T>`. Sorry for the noise

PR #913, opened 2022-11-10
Allow completion for nested quoted fields

ebresafegaga:
This is a tiny PR  to finish up #909. This change will allow completion even when we have quoted fields. 

This closes item `2b` of #877

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -365,7 +365,7 @@ mod tests {
             (
                 "Record binding with unicode string names for fields",
                 r##"let x = {"fo京o" = {bar = 42}} in x."fo京o".foo"##,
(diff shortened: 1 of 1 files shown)


vkleen (commented):
Does completion still do (or has it ever done) the right thing with record paths that need to be quoted? I.e. if completion indicates `fo京o` as a fit and I accept it in the editor, does it get inserted with quotes?

ebresafegaga:
> Does completion still do (or has it ever done) the right thing with record paths that need to be quoted? I.e. if completion indicates `fo京o` as a fit and I accept it in the editor, does it get inserted with quotes?

Yes, the quotes get inserted for you.

vkleen (approved):
Awesome! Then I don't see any problems with this PR 👍

PR #918, opened 2022-11-15
Auto-complete for Nickel's stdlib

ebresafegaga:
This change adds support for auto-completion for the Nickel standard library. 

This fixes item `1b` on #877

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -12,6 +12,7 @@ use nickel_lang::{
 use crate::linearization::interface::{TermKind, UsageState};
 
 use super::{
+    completed::Completed,
     interface::{Unresolved, ValueState},
     Environment, IdGen, ItemID, LinearizationItem,
 };
@@ -22,6 +23,7 @@ use super::{
 #[derive(Default)]
 pub struct Building {
     pub linearization: Vec<LinearizationItem<Unresolved>>,
+    pub lin_cache: HashMap<FileId, Completed>,
 }
 
 impl Building {
@@ -29,30 +31,46 @@ impl Building {
         self.linearization.push(item)
     }
 
-    pub(super) fn add_usage(&mut self, _decl: ItemID, _usage: ItemID) {
+    pub(super) fn add_usage(&mut self, file: FileId, decl: ItemID, usage: ItemID) {
+        let (decl_file, index) = decl;
+        let kind = if file == decl_file {
+            // This usage references an item in the file we're currently linearizing
+            &mut self
+                .linearization
+                .get_mut(index)
+                .expect("Could not find parent")
+                .kind
+        } else {
+            // This usage references an item in another file (that has already been linearized)
+            &mut self
+                .lin_cache
+                .get_mut(&decl_file)
+                .expect("Could not find parent")
+                .get_item_mut(decl)
+                .expect("Could not find parent")
+                .kind
+        };
 
-        // TODO fix this function
-        // match self
-        //     .linearization
-        //     // For now use only the file index
-        //     .get_mut(decl.1)
-        //     .expect("Could not find parent")
-        //     .kind
-        // {
-        //     TermKind::Structure | TermKind::Usage(_) | TermKind::Record(_) => {
-        //         unreachable!()
-        //     }
-        //     TermKind::Declaration(_, ref mut usages, _)
-        //     | TermKind::RecordField { ref mut usages, .. } => usages.push(usage),
-        // };
+        match kind {
+            TermKind::Record(_) | TermKind::Structure | TermKind::Usage(_) => unreachable!(),
+            TermKind::Declaration(_, ref mut usages, _)
+            | TermKind::RecordField { ref mut usages, .. } => usages.push(usage),
+        };
     }
 
-    pub(super) fn inform_declaration(&mut self, declaration: ItemID, value: ItemID) {
-        if let Some(LinearizationItem {
-            kind: TermKind::Declaration(_, _, value_state),
-            ..
-        }) = self.linearization.get_mut(declaration.1)
-        {
+    pub(super) fn inform_declaration(&mut self, file: FileId, declaration: ItemID, value: ItemID) {
+        let (decl_file, index) = declaration;
+        let kind = if file == decl_file {
+            // This usage references an item in the file we're currently linearizing
+            self.linearization.get_mut(index).map(|item| &mut item.kind)
+        } else {
+            // This usage references an item in another file (that has already been linearized)
+            self.lin_cache
+                .get_mut(&decl_file)
+                .and_then(|lin| lin.get_item_mut(declaration).map(|item| &mut item.kind))
+        };
+
+        if let Some(TermKind::Declaration(_, _, value_state)) = kind {
             *value_state = ValueState::Known(value)
         }
     }
@@ -141,7 +159,11 @@ impl Building {
         })
     }
 
-    pub(super) fn resolve_record_references(&mut self, mut defers: Vec<(ItemID, ItemID, Ident)>) {
+    pub(super) fn resolve_record_references(
+        &mut self,
+        file: FileId,
+        mut defers: Vec<(ItemID, ItemID, Ident)>,
+    ) {
         let mut unresolved: Vec<(ItemID, ItemID, Ident)> = Vec::new();
 
         while let Some(deferred) = defers.pop() {
@@ -236,7 +258,7 @@ impl Building {
             }
 
             if let Some(referenced_id) = referenced_id {
-                self.add_usage(referenced_id, *child_item);
+                self.add_usage(file, referenced_id, *child_item);
             }
 
             if defers.is_empty() && !unresolved.is_empty() {
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, mem};
 
+use codespan::FileId;
 use log::debug;
 use nickel_lang::{
     identifier::Ident,
@@ -60,6 +61,7 @@ impl Building {
         record_fields: &HashMap<Ident, RichTerm>,
         record: usize,
         env: &mut Environment,
+        file: FileId,
     ) {
         for (ident, value) in record_fields.iter() {
             let id = self.id_gen().get_and_advance();
@@ -83,7 +85,8 @@ impl Building {
                     _ => None,
                 },
             });
-            env.insert(*ident, id);
+            let key = (*ident, file);
+            env.insert(key, id);
             self.add_record_field(record, (*ident, id))
         }
     }
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -13,7 +13,7 @@ use crate::linearization::interface::{TermKind, UsageState};
 
 use super::{
     interface::{Unresolved, ValueState},
-    Environment, IdGen, LinearizationItem,
+    Environment, IdGen, ItemID, LinearizationItem,
 };
 
 /// A concrete [LinearizationState]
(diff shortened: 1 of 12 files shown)


vkleen (commented):
Referring to stdlib modules by their index in a vector in the implementation of the file cache that is set implicitly by an iteration order feels incredibly brittle. Maybe we could refactor the `stdlib_ids` field in `Cache` to be a hashmap from `StdlibModule`s to `FileId`s? I guess that would also necessitate some changes in how the stdlib modules are loaded before evaluation.

ebresafegaga:
That would be much much better. Thanks for the suggestion!

src/stdlib.rs
```diff
@@ -1,33 +1,102 @@
...
+            "array" => StdlibModule::Array,
+            "record" => StdlibModule::Record,
+            "string" => StdlibModule::String,
+            "num" => StdlibModule::Num,
+            "function" => StdlibModule::Function,
+            "internals" => StdlibModule::Internals,
+            _ => StdlibModule::Array,
+        }
+    }
+}
```
vkleen, line 84: This should probably be an implementation of `TryFrom` instead, unless the use site absolutely can't deal with an `Option`.

src/stdlib.rs
```diff
@@ -1,33 +1,102 @@
...
 
 /// Return the list `(name, source_code)` of all the stdlib modules.
-pub fn modules() -> Vec<(&'static str, &'static str)> {
+pub fn modules() -> Vec<(StdlibModule, &'static str, &'static str)> {
+    // If you change the order of this, please also modify the
+    // `get_module_id` function correspondinly
     vec![
         BUILTIN, CONTRACT, ARRAY, RECORD, STRING, NUM, FUNCTION, INTERNALS,
     ]
 }
```
vkleen, line 55: The comments should be updated after the change to `HashMap`.

lsp/nls/src/server.rs
```diff
@@ -107,8 +111,31 @@ impl Server {
...
+            StdlibModule::Internals,
+        ];
+        for module in modules {
+            let file_id = cache.get_submodule_file_id(module).unwrap();
+            cache
+                .typecheck_with_analysis(file_id, &self.initial_ctxt, &mut self.lin_cache)
+                .unwrap();
+        }
+        Ok(())
+    }
```
vkleen, line 134: This should use the `modules` function from `stdlib.rs` so we have a single source of truth on which modules are in the stdlib.

vkleen (approved):
I've left some cleanup suggestions, other than that this looks good.

Also, I'm wondering why `stdlib_ids` in the file cache is an `Option` instead of having an empty container represent that the stdlib isn't parsed yet. But I don't think a change like that should go into this PR.

src/stdlib.rs
ebresafegaga: At the use site, we're absolutely sure it is a one of those strings. I guess we can just `unwrap` then.

lsp/nls/src/requests/completion.rs
```diff
@@ -167,33 +168,72 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
...
+                if let Some(idents) = find_fields_from_term_kind(&linearization, *value, path) {
+                    vec![(idents, item.ty.clone())]
+                } else {
+                    return Vec::new();
+                }
+            }
+            _ => return Vec::new(),
+        }
+    } else {
+        return Vec::new();
```
matthew-healy, line 213: I don't think any of the `return`s here are needed - you should be able to just remove them and the `;` to get the same result.

lsp/nls/src/requests/completion.rs
```diff
@@ -167,33 +168,72 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
...
+                    value: ValueState::Known(value),
+                    ..
+                },
+                _,
+            ) => {
+                if let Some(idents) = find_fields_from_term_kind(&linearization, *value, path) {
+                    vec![(idents, item.ty.clone())]
+                } else {
+                    return Vec::new();
+                }
```
matthew-healy, line 208: This _might_ be cleaner as something like this:

```suggestion
                find_fields_from_term_kind(&linearization, *value, path)
                    .map(|idents| vec![(idents, item.ty.clone())])
                    .unwrap_or_default()
```

ebresafegaga:
> Also, I'm wondering why `stdlib_ids` in the file cache is an `Option` instead of having an empty container represent that the stdlib isn't parsed yet. But I don't think a change like that should go into this PR.

It is to used to know if the `stdlib` has been loaded or not. Creating an empty `HashMap` (or previously `Vec`) might be a bit misleading.

lsp/nls/src/requests/completion.rs
```diff
@@ -167,33 +168,72 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
...
+                    // Get record fields from contract metadata
+                    Some(fields) => vec![(fields, item.ty.clone())],
+                    // Get record fields from lexical scoping
+                    None => {
+                        if let Some(idents) = find_fields_from_term_kind(&linearization, id, path) {
+                            vec![(idents, item.ty.clone())]
+                        } else {
+                            return Vec::new();
+                        }
                     }
```
matthew-healy, line 194: I think you could probably avoid pattern-matching on the optional here, with something like:

```suggestion
                find_fields_from_contract(&linearization, *body_id, path)
                    .or_else(|| find_fields_from_term_kind(&linearization, id, path))
                    .map(|idents| vec![(idents, item.ty.clone())])
                    .unwrap_or_default()
```

lsp/nls/src/requests/completion.rs
matthew-healy, line 213: As with the examples above, this _might_ be cleaner as a `map` on the result of `get_item` followed by an `unwrap_or_default`.

matthew-healy (commented):
Looks good overall, I just left some small suggestions about restructuring `collect_record_info`.

ebresafegaga:
> I just left some small suggestions about restructuring `collect_record_info`.

Thank you!

yannham (changes requested):
First, I'm very hyped to type `array.` in my editor and see the completion happening :tada: that's a big real life improvement.

With respect to the implementation though, I feel like we do a lot of special casing for the stdlib. Why does it need a specific completion function? Also, we once again do a lilnear search across the whole linearization, which is often a red herring. What prevents us from using the same mechanism as for other records (by looking into the current environment stored in the linearization item)?

ebresafegaga:
> With respect to the implementation though, I feel like we do a lot of special casing for the stdlib. Why does it need a specific completion function? Also, we once again do a linear search across the whole linearization, which is often a red herring. What prevents us from using the same mechanism as for other records (by looking into the current environment stored in the linearization item)?

The `stdlib_completion` function is not really a special case. It just gets the linearization of the `stdlib` module we want completion for from the server's cache, and then performs completion the normal way by calling `collect_record_info` but with that `stdlib` module's linearization, instead of the linearization of the current file.

For the linear search issue: currently, the `stdlib` modules are represented a record with a single field. e.g in `array.ncl`, we have `{ array = {  <definition of the array module> } }`. The current environment doesn't have information about the `array.ncl` file, so there's no way we can get and `id` and perform constant time lookup

lsp/nls/src/requests/completion.rs
```diff
@@ -167,33 +170,61 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
 /// partiular ID, and in the scope of a given linearization item.
 fn collect_record_info(
     linearization: &Completed,
-    item: &LinearizationItem<Types>,
     id: usize,
     path: &mut Vec<Ident>,
 ) -> Vec<(Vec<Ident>, Types)> {
```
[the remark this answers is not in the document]
ebresafegaga: Right. We don't need it. That was from the old implementation, I think

lsp/nls/src/requests/completion.rs
```diff
@@ -167,33 +170,61 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
...
+    let lin = server.lin_cache_get(&file_id).ok()?;
+
+    // Stdlib modules are represented by a record with a single field which is named
+    // after the module name. Here, we try to get the ID of the
+    // `TermKind::RecordField {..}` associated with that field, which defines the module.
+    // Also, because of the way the linearization algorithm works, we can be almost sure that
+    // the ID of this field is `1`, which will give us constant time access to it, but
+    // this might not always be the case (the algorithm can change), so we have to settle
+    // for a linear search here.
+    let id = lin.linearization.iter().find_map(|item| match &item.kind {
```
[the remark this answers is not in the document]
ebresafegaga: I don't think we can, because the environment of the current file doesn't tell us anything about the linearization of the stdlib (which is a different file). That means that we cannot get the `id` of the `array` field (in this case).

Another approach might be to look into the type-checking context, because the `stdlib` modules are already there, but this would be different from the normal completion mechanism  we already have

lsp/nls/src/linearization/mod.rs
```diff
@@ -24,7 +24,7 @@ pub mod building;
 pub mod completed;
 pub mod interface;
 
-pub type Environment = nickel_lang::environment::Environment<Ident, usize>;
+pub type Environment = nickel_lang::environment::Environment<(Ident, FileId), usize>;
```
yannham, line 27: I wonder why you need `FileId` to be an input of your table. My hunch would be that it should be an output, that is, at any point of the program, we have an environment that maps the identifiers of the current scope to a linearization id. This id may be in other files (typically e.g. for the stdlib). Which would give:

```rust
pub type Environment = nickel_lang::environment::Environment<Ident, (FileId, usize)>;
```

In fact, I would even set `pub type ItemID = (FileId, usize)` and use that everywhere instead of `usize`. So that gotodefinition, etc. can cross files.

As it is now, you need to provide a `FileId` before getting information from an identifier. As you can't guess where this identifier comes from, what can you pass beside the current `FileId`?

yannham, line 27: I don't remember the exact type, but I think we have an alias for items ID?

```suggestion
pub type Environment = nickel_lang::environment::Environment<(Ident, FileId), ItemID>;
```

src/cache.rs
```diff
@@ -831,18 +831,37 @@ impl Cache {
         self.terms.get(&file_id).map(|CachedTerm { term, .. }| term)
     }
 
+    /// Retrieve the FileId for a given standard libray module.
+    pub fn get_submodule_file_id(&self, module: StdlibModule) -> Option<FileId> {
```
yannham, line 835: Suggestion, instead of `submodule` (which doesn't really have a definition currently):

```suggestion
    pub fn get_stdlib_module_file_id(&self, module: StdlibModule) -> Option<FileId> {
```

src/cache.rs
```diff
@@ -831,18 +831,37 @@ impl Cache {
...
+        let file = self.stdlib_ids.as_ref()?.get(&module).copied()?;
+        Some(file)
+    }
+
+    /// Retrieve the FileIds for all the stdlib modules
+    pub fn get_all_submodules_file_id(&self) -> Option<Vec<FileId>> {
```
yannham, line 841:
```suggestion
    pub fn get_all_stdlib_modules_file_id(&self) -> Option<Vec<FileId>> {
```

lsp/nls/src/requests/completion.rs
```diff
@@ -202,62 +205,86 @@ fn remove_duplicates(items: &Vec<CompletionItem>) -> Vec<CompletionItem> {
...
+        linearization: &Completed,
+        server: &Server,
+        path: &mut Vec<Ident>,
+        file: FileId,
+    ) -> Option<Vec<Ident>> {
+        item.env
+            .get(&(name, file))
+            .map(|key| (key, linearization))
+            .or_else(|| {
+                // If we can't find in the file's scope, check it from the stdlib's scope.
```
yannham, line 258: Here, if you had `Environment = <Ident, (FileId, usize)>`, you wouldn't have to do this `or_else` special casing.

lsp/nls/src/server.rs
```diff
@@ -70,7 +76,34 @@ impl Server {
...
+            // This module has a different format from the rest of the stdlib items
+            // Also, users are not supposed to use the internal module directly
+            if module == StdlibModule::Internals {
+                continue;
+            }
+            let name: Ident = module.into();
+            let file_id = self.cache.get_submodule_file_id(module)?;
+            let lin = self.lin_cache_get(&file_id).ok()?;
+            // We're using the ID 0 here because the stdlib is represented by a
+            // single record which would have an ID of 0
```
yannham, line 96:
```suggestion
            // We're using the ID 0 here because a stdlib module is currently
            // REQUIRED to be a record literal. The linearization ID of this
            // record must thus be 0
```

lsp/nls/src/linearization/mod.rs
ebresafegaga: I think I misunderstood your initial suggestion

lsp/nls/src/linearization/building.rs
```diff
@@ -11,58 +12,78 @@ use nickel_lang::{
...
-    Environment, IdGen, LinearizationItem,
+    Environment, IdGen, ItemID, LinearizationItem,
 };
 
 /// A concrete [LinearizationState]
 /// Holds any inner datatype that can be used as stable resource
 /// while recording terms.
-#[derive(Default)]
-pub struct Building {
+// #[derive(Default)]
```
yannham, line 23:
```suggestion
```

lsp/nls/src/linearization/building.rs
```diff
@@ -11,58 +12,78 @@ use nickel_lang::{
...
+        let (decl_file, index) = declaration;
+        let kind = if file == decl_file {
+            // This usage references an item in the file we're currently linearizing
+            self.linearization.get_mut(index).map(|item| &mut item.kind)
+        } else {
+            // This usage references an item in another file (that has already been linearized)
+            self.lin_cache
+                .get_mut(&decl_file)
+                .and_then(|lin| lin.get_item_mut(declaration).map(|item| &mut item.kind))
+        };
```
yannham, line 71: It's the second time that this snippet appears: branching on `if file == decl_file` to get a mutable reference to an item. Maybe this could be put in a separate function in building, something like `get_item_mut(&mut self, file: FileId, item: ItemID)`.

lsp/nls/src/linearization/mod.rs
```diff
@@ -24,7 +24,8 @@ pub mod building;
 pub mod completed;
 pub mod interface;
 
-pub type Environment = nickel_lang::environment::Environment<Ident, usize>;
+pub type ItemID = (FileId, usize);
```
yannham, line 27: Nitpick: we could use a struct here, like `struct ItemId { file_id: FileId, index: usize }` (or something like that).

Also, `ItemID` should be `ItemId` in CamelCase, I think.

yannham (approved):
Beside some suggestions, LGTM! I somehow have a hard time triggering completion in the mantis example (`benches/mantis/schemas/nomad/types.ncl`), but it seems my coc-nvim is also failing me with strange errors. The LSP is also slow to update in that case, but this has nothing to do with this PR.

PR #923, opened 2022-11-18
Introduce `%record_diff_fields%` primitive

matthew-healy:
Previously we used a regular Nickel function to diff the fields of the records inside the `$record` contract. However, as we are moving to an "internal" representation of sealed polymorphic record tails, a regular function will no longer work. This is because there's no way on the Nickel side to build a new record which is guaranteed to have the sealed tail of another.

This commit introduces a new primitive, `%record_diff_fields%`, which returns only the (visible) fields of the first argument which are not also present in the second argument, while preserving whatever tail the first argument had.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -1955,6 +1955,42 @@ impl<R: ImportResolver> VirtualMachine<R> {
                     ))
                 }
             },
+            BinaryOp::DiffRecordFields() => match_sharedterm! {t1,
+                with {
+                    Term::Record(r1) => match_sharedterm! {t2,
+                        with {
+                            Term::Record(r2) => {
+                                let mut r1 = r1;
+                                r1.fields.retain(|k, _| !r2.fields.contains_key(k));
+                                Ok(Closure {
+                                    body: RichTerm::new(Term::Record(r1), fst_pos),
+                                    env: env1
+                                })
+                            }
+                        } else {
+                            Err(EvalError::TypeError(
+                                String::from("Record"),
+                                String::from("%record_diff_fields%, 2nd argument"),
+                                snd_pos,
+                                RichTerm {
+                                    term: t2,
+                                    pos: pos2
+                                }
+                            ))
+                        }
+                    }
+                } else {
+                    Err(EvalError::TypeError(
+                        String::from("Record"),
+                        String::from("%record_diff_fields%, 1st argument"),
+                        fst_pos,
+                        RichTerm {
+                            term: t1,
+                            pos: pos1,
+                        }
+                    ))
+                }
+            },
             BinaryOp::HasField() => match_sharedterm! {t1, with {
                     Term::Str(id) => {
                         if let Term::Record(record) = &*t2 {
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -736,6 +736,7 @@ BOpPre: BinaryOp = {
     "str_contains" => BinaryOp::StrContains(),
     "record_insert" => BinaryOp::DynExtend(),
     "record_remove" => BinaryOp::DynRemove(),
+    "record_diff_fields" => BinaryOp::DiffRecordFields(),
 }
 
 NOpPre<ArgRule>: UniTerm = {
@@ -878,6 +879,7 @@ extern {
         "record_map" => Token::Normal(NormalToken::RecordMap),
         "record_insert" => Token::Normal(NormalToken::RecordInsert),
         "record_remove" => Token::Normal(NormalToken::RecordRemove),
+        "record_diff_fields" => Token::Normal(NormalToken::RecordDiffFields),
         "seq" => Token::Normal(NormalToken::Seq),
         "deep_seq" => Token::Normal(NormalToken::DeepSeq),
         "head" => Token::Normal(NormalToken::Head),
--- a/src/parser/lexer.rs
+++ b/src/parser/lexer.rs
@@ -197,6 +197,8 @@ pub enum NormalToken<'input> {
     RecordInsert,
     #[token("%record_remove%")]
     RecordRemove,
+    #[token("%record_diff_fields%")]
+    RecordDiffFields,
     #[token("%seq%")]
     Seq,
     #[token("%deep_seq%")]
--- a/src/term.rs
+++ b/src/term.rs
@@ -1370,6 +1370,11 @@ pub enum BinaryOp {
     DynRemove(),
     /// Access the field of record. The field name is given as an arbitrary Nickel expression.
     DynAccess(),
+    /// Return a new record containing all the fields of the first argument which are
+    /// not featured in the second argument.
+    ///
+    /// Preserves the polymorphic tail of the first argument, if one is set.
+    DiffRecordFields(),
     /// Test if a record has a specific field.
     HasField(),
     /// Concatenate two arrays.
--- a/src/typecheck/operation.rs
+++ b/src/typecheck/operation.rs
@@ -304,6 +304,12 @@ pub fn get_bop_type(
                 mk_typewrapper::dyn_record(res),
             )
         }
+        // Dyn -> Dyn -> Dyn
+        BinaryOp::DiffRecordFields() => (
+            mk_typewrapper::dynamic(),
+            mk_typewrapper::dynamic(),
+            mk_typewrapper::dynamic(),
+        ),
         // forall a. Str -> {_: a} -> Bool
         BinaryOp::HasField() => {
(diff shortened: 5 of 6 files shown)


src/eval/operation.rs
```diff
@@ -1955,6 +1955,42 @@ impl<R: ImportResolver> VirtualMachine<R> {
...
+                        String::from("Record"),
+                        String::from("%record_diff_fields%, 1st argument"),
+                        fst_pos,
+                        RichTerm {
+                            term: t1,
+                            pos: pos1,
+                        }
+                    ))
+                }
+            },
```
sir4ur0n, line 1993: Can we add tests for this case please? :pray: 

I'm thinking at least a few examples (unit tests), and I can think off the top of my head about the following PBTs:
* `for any record r1, diff_record_fields r1 r1 == {}`
* `for any record r1, any list of extra fields (except SPRT) extra_fields, diff_record_fields (r1 // extra_fields) r1 == diff_record_fields r1 (r1 // extra_fields) == extra_fields`
* `for any record r1, for any record r2, diff_record_fields r1 r2 (except SPRT) == diff_record_fields r2 r1 (except SPRT)`

sir4ur0n (changes requested):

src/eval/operation.rs
matthew-healy: I'm not necessarily against adding these tests, but I'll point out that this op is already tested by virtue of its use inside the `$record` contract, so every test for that contract verifies that this is a valid change. 

In general, I'm not sure what the testing strategy is here. On the one hand, this is only ever intended to be used as an implementation detail inside the `$record` contract, so you could argue that it's sufficient to just rely on the tests for that feature. On the other hand, there's nothing (lack of documentation aside) to stop users from calling the `%record_diff_fields%` primitive in their own code, so I can see why we might want this to have its own suite of tests.

@yannham do you have any thoughts on this?

yannham:
> Previously we used a regular Nickel function to diff the fields of the records inside the $record contract. However, as we are moving to an "internal" representation of sealed polymorphic record tails, a regular function will no longer work. This is because there's no way on the Nickel side to build a new record which is guaranteed to have the sealed tail of another.

I haven't review yet, but just as a naive question, if you have operators to fetch and insert the seal, couldn't you just do that to transfer the tail to another record, like `let foo_with_tail_of_bar = %put_in_tail% (%get_tail% bar) prev_foo`?

src/eval/operation.rs
matthew-healy: I guess I'll add: my personal opinion is that we should prioritise testing via the (_intended_) public surface area of the language, so I'd be in favour of not adding tests here, given that any change to this code will break most of the tests for record contracts anyway.

Ultimately I'm happy to do whatever we all agree makes sense, but it'd be great to document whatever decision we reach for future reference.

matthew-healy:
@yannham Huh, I actually hadn't thought of that. :facepalm: I can't think of any reason why it wouldn't work... I'll give it a shot & close this PR once I've confirmed that it works as expected.

matthew-healy:
@yannham Ah, wait, I just realised why this won't work: `$record` needs to be able to handle records that have tails as well as those that don't. Currently, if we pass a tail-less record to the unhide primitive, it raises a blame error because something must have gone wrong. 

On the other hand, `$record` needs to handle records with _any_ sort of tail, which means we're not guaranteed to have a sealed tail at the point we diff the records, so doing what you suggest wouldn't work without at the very least rethinking the design.

yannham:
What about just an `has_tail`?

>  Currently, if we pass a tail-less record to the unhide primitive, it raises a blame error because something must have gone wrong.

This is a tad surprising. IIUC, something must have gone wrong in the implementation of Nickel itself, right? In this case it would be more appropriate to panic or raise a special error, rather than blaming user code, I think. Anyway, couldn't record know in some way if it is operating on a tailless record or not? Even with a simple argument passed from the contract builder? That would eliminate the need for `has_tail`.

> We could definitely consider a %transfer_tail% primitive, but I'm a little hesitant to put anything that mutates private implementation details into the hands of users, so I think a diff primitive might be a better option for now.

Primops are (voluntarily) undocumented and have those ugly scary names. It's a bit like Haskell strict primops or OCaml `Obj.magic`: they are not meant to be used by users, and if they do, it's at their own risk.

Also, I don't think you need to mutate anything (and I don't even think you can, because things are behind `Rc`). You could just create a new empty record from another record, using the tail of the argument. Something like `%empty_with_tail% arg` and insert whatever you need inside the record later. Would that be sufficient? This is not less safe or more powerful than `diff_record_fields`, because it can be implemented from the latter, but it imports less logic inside the primop Rust implementation.

All in all, if those are too many changes, let's review and merge this first solution, and iterate. I'm just trying to be sure of all the possibilities we have here, and that we pick the simplest :slightly_smiling_face:

matthew-healy:
> > Currently, if we pass a tail-less record to the unhide primitive, it raises a blame error because something must have gone wrong.
>
> This is a tad surprising. IIUC, something must have gone wrong in the implementation of Nickel itself, right? In this case it would be more appropriate to panic or raise a special error, rather than blaming user code, I think. 

This is a more-or-less exact replica of the existing behaviour - see [here](https://github.com/tweag/nickel/blob/master/stdlib/internals.ncl#L103).  I had originally initially intended to either (1) do nothing and return `null` or (2) raise an `InternalError`, but decided to keep everything as close as possible to the original implementation , at least for this first pass. As for `panic`king, given that - advisable or otherwise - users can call the seal/unseal primitives with whatever arguments they want, I don't think that's the right solution here.

> Anyway, couldn't $record know in some way if it is operating on a tailless record or not? Even with a simple argument passed from the contract builder? That would eliminate the need for has_tail.

Yes, it probably could. I'm not necessarily advocating for what's here as the best solution, but I think it's the simplest way to make tail sealing an internal detail right now, and I'm more than happy to continue fine-tuning this implementation beyond these first PRs. 

> Also, I don't think you need to mutate anything (and I don't even think you can, because things are behind Rc). You could just create a new empty record from another record, using the tail of the argument. Something like %empty_with_tail% arg and insert whatever you need inside the record later.

Ah, apologies, I was using "mutate" in an informal (& not really accurate) sense, rather than talking about literal mutation. I just think that putting a building-block like that (or like `%empty_with_tail%`) into the wild is going to lead to really weird user programs. I guess ultimately that's not our problem, but that's where I was coming from. 

> All in all, if those are too many changes, let's review and merge this first solution, and iterate.

This would be my preferred way forward. It's not that these are too many changes per-se, but a lot of the work here has been getting my head around how things _currently_ work and trying to replicate that. Now that I've (almost) done that, I think it'll be a lot easier to look at improving things.

yannham:
> This is a more-or-less exact replica of the existing behaviour - see [here](https://github.com/tweag/nickel/blob/master/stdlib/internals.ncl#L103). I had originally initially intended to either (1) do nothing and return null or (2) raise an InternalError, but decided to keep everything as close as possible to the original implementation , at least for this first pass. As for panicking, given that - advisable or otherwise - users can call the seal/unseal primitives with whatever arguments they want, I don't think that's the right solution here.

The current behavior hasn't been thought through that much, I think it's reasonable to change this one already. You are totally right though, we should never panic for something that the user may write, even if primops are undocumented on purpose. That should just be a normal primop error, then, I guess.

> Ah, apologies, I was using "mutate" in an informal (& not really accurate) sense, rather than talking about literal mutation. I just think that putting a building-block like that (or like %empty_with_tail%) into the wild is going to lead to really weird user programs. I guess ultimately that's not our problem, but that's where I was coming from.

But you can implement `%empty_with_tail%` from `%diff_record_fields%`, can't you? As `let empty_with_tail = funr => %diff_record_fields% r (record.fields r |> array.fold (fun acc elt => acc & {"%{elt}" = true}) {})`? I mean, I don't see `empty_with_tail` being more or less "dangereous" than `diff_with_fields`, as the latter is just the former + diffing fields.

> As for `panic` king

Panic King would make a great band name (or developer nickname)

> This would be my preferred way forward. It's not that these are too many changes per-se, but a lot of the work here has been getting my head around how things currently work and trying to replicate that. Now that I've (almost) done that, I think it'll be a lot easier to look at improving things.

Gotcha, sounds reasonable! Feel free to request new reviews whenever this PR is ready again :slightly_smiling_face:

matthew-healy:
> But you can implement %empty_with_tail% from %diff_record_fields%, can't you?

Aha, yep okay fair point. `%empty_with_tail%` it is then. :grin:

matthew-healy:
I'll likely close this PR and open a new one implementing `%empty_with_tail%` instead, but for now I'm just going to leave it as a draft, just in case.

Issue #908, opened 2022-11-07
Recursive overriding is nonassociative


vkleen:
Recursive record overriding is currently nonassociative:
```
let correct_ =
{
    foo = "Overridden",
} & {
  config | {
    output | {
        value | Str,
    }
  },
} & {
  foo | Str
    | default = "blub",

  config.output.value = foo,
} in
let wrong_ =
{
    foo = "Overridden",
} & ({
  config | {
    output | {
        value | Str,
    }
  },
} & {
  foo | Str
    | default = "blub",

  config.output.value = foo,
}) in
{
  correct = correct_,
  wrong = wrong_,
}
```
yields
```
{
  wrong = { foo = "Overridden", config = { output = { value = "blub" } } },
  correct = { foo = "Overridden", config = { output = { value = "Overridden" } } } 
}
```

The semantics of record overriding should always be associative.

**Environment**
 - OS name + version: NixOS unstable
 - Version of the code: c13c51f51b61ce7617cee7796dc4113c3a1567af

yannham:
After some quick investigations, this doesn't look like a consequence of the issues described in RFC005, but rather a plain old bug in the implementation of overriding. It seems that when a field recursively depending on another one is merged with something, the newly allocated thunks is not revertible, nor does it remember the dependencies. Which some set its value at the time, instead of staying overridable.

PR #924, opened 2022-11-18
Expand documentation of revertible thunks

yannham:
First step toward fixing #908. After pairing with @vkleen, we realized that some underlying assumptions and the overall mental model of what are revertible thunks and what they are used for wasn't very clear. This PR tries to fill this gap.

A subsequent PR will bring the conclusion of this documentation to its logical conclusion, by piggy-packing on functions and function application instead of re-implementing that in an ad-hoc way inside `eval::fixpoint` as much as possible.

### DIFF
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -34,7 +34,7 @@ pub struct ThunkData {
 /// - A standard thunk, that is destructively updated once and for all
 /// - A revertible thunk, that can be restored to its original expression. Used to implement
 ///   recursive merging of records and overriding (see the
-///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunks optionally
+///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunk optionally
 ///   stores the set of recusive fields it depends on (see [`crate::transform::free_vars`]).
 ///
 /// # Revertible thunks
@@ -52,7 +52,7 @@ pub struct ThunkData {
 /// is really a builtin representation of `fun self => {foo = self.bar + self.baz + 1, bar = 2, baz
 /// = 1}`. The interpreter computes a fixpoint when evaluating a `RecRecord` away to a `Record`,
 /// equivalent to `let rec fixpoint = repr fixpoint in fixpoint`. We use a different vocabulary and
-/// representations, but this is what we essentially do. We're just hiding the details from the
+/// representation, but this is what we essentially do. We're just hiding the details from the
 /// user and pretend recursive records are data that can be used as a normal record, because it's
 /// both simple and natural (for the user).
 ///
@@ -60,9 +60,9 @@ pub struct ThunkData {
 /// functional representation, and re-compute the fixpoint at each record operation, such as field
 /// access. While correct, this would be wasteful, because the result of the fixpoint computations
 /// is actually the same throughout most record operations (put differently, `self` doesn't
-/// change). But we can't just store the computed `fixpoint` either in a normal thunk and forget
+/// change). But we can't just store the computed `fixpoint` in a normal thunk and forget
 /// about the original function, because upon merge, and more specifically upon recursive
-/// overriding happen, we do need to compute a new fixpoint.
+/// overriding, we do need to compute a new fixpoint.
 ///
 /// Revertible thunks are a simple **memoization device** for the `fun self => ...` functional
 /// representation of a recursive record. First, note that there are several different but
@@ -81,7 +81,7 @@ pub struct ThunkData {
 /// {foo = fun bar baz => bar + baz + 1, bar = 2, bar 1}
 /// ```
 ///
-/// All those representations are isomorphic, in that modulo basic fee variable analysis, we can
+/// All those representations are isomorphic, in that, modulo basic free variable analysis, we can
 /// get from one to another mechanically. Pushing the functions down the fields is better because
 /// it reveals more static information about the record (such as the list of fields) without having
 /// to provide any argument first. Then, splitting `self` and removing the function altogether when
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -35,8 +35,71 @@ pub struct ThunkData {
 /// - A revertible thunk, that can be restored to its original expression. Used to implement
 ///   recursive merging of records and overriding (see the
 ///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunks optionally
-///   stores the set of recusive fields it depends on. See the [`crate::transform::free_vars`] for more
-///   details.
+///   stores the set of recusive fields it depends on (see [`crate::transform::free_vars`]).
+///
+/// # Revertible thunks
+///
+/// Recursive records are better understood as being functions. For example, the recursive record:
+///
+/// ```nickel
+/// {
+///   foo = bar + baz + 1,
+///   bar = 2,
+///   baz = 1,
+/// }
+/// ```
+///
+/// is really a builtin representation of `fun self => {foo = self.bar + self.baz + 1, bar = 2, baz
+/// = 1}`. The interpreter computes a fixpoint when evaluating a `RecRecord` away to a `Record`,
+/// equivalent to `let rec fixpoint = repr fixpoint in fixpoint`. We use a different vocabulary and
+/// representations, but this is what we essentially do. We're just hiding the details from the
+/// user and pretend recursive records are data that can be used as a normal record, because it's
+/// both simple and natural (for the user).
+///
+/// A naive representation of recursive records could be to always keep recursive records in their
+/// functional representation, and re-compute the fixpoint at each record operation, such as field
+/// access. While correct, this would be wasteful, because the result of the fixpoint computations
+/// is actually the same throughout most record operations (put differently, `self` doesn't
+/// change). But we can't just store the computed `fixpoint` either in a normal thunk and forget
+/// about the original function, because upon merge, and more specifically upon recursive
+/// overriding happen, we do need to compute a new fixpoint.
+///
+/// Revertible thunks are a simple **memoization device** for the `fun self => ...` functional
+/// representation of a recursive record. First, note that there are several different but
+/// equivalent ways of encoding the example above as a function:
+///
+/// ```nickel
+/// fun self => { foo = self.bar + self.baz + 1, bar = 2, baz = 1 }
+///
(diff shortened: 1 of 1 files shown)


src/eval/lazy.rs
```diff
@@ -35,8 +35,71 @@ pub struct ThunkData {
...
+/// (this has an important positive memory impact, see [ThunkDeps]).
+///
+/// In practice, Nickel uses the last representation. Fields that don't have any free variable
+/// intersecting with the name of the other fields are allocated as normal thunks that will only be
+/// evaluated at most once. Fields with actual dependencies are stored in revertible thunks, where
+/// the `orig` field stores the body of the function (in our example the expression `bar + baz +
+/// 1`), and the cached version is the application to `self` computed when we evaluated the
+/// recursive record to a normal record. The function arguments (that is, the intersection of the
+/// free variables of the expression with the set of fields) are stored inside `deps`. `deps`
+/// represent the other fields a specific field is depending on, and are guaranteed to appear free
```
vkleen, line 98: My "english senses" think it should be "freely" here as an adverb. But PL theory papers might disagree, so feel free to ignore this suggestion.
```suggestion
/// represent the other fields a specific field is depending on, and are guaranteed to appear freely
```

vkleen (approved):
I think that's a good description of revertible thunks 👍 I've just left some suggestions and found a few typos.

src/eval/lazy.rs
```diff
@@ -35,8 +35,71 @@ pub struct ThunkData {
 /// - A revertible thunk, that can be restored to its original expression. Used to implement
 ///   recursive merging of records and overriding (see the
 ///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunks optionally
```
sir4ur0n, line 37:
```suggestion
///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunk optionally
```

src/eval/lazy.rs
```diff
@@ -35,8 +35,71 @@ pub struct ThunkData {
...
+/// evaluated at most once. Fields with actual dependencies are stored in revertible thunks, where
+/// the `orig` field stores the body of the function (in our example the expression `bar + baz +
+/// 1`), and the cached version is the application to `self` computed when we evaluated the
+/// recursive record to a normal record. The function arguments (that is, the intersection of the
+/// free variables of the expression with the set of fields) are stored inside `deps`. `deps`
+/// represent the other fields a specific field is depending on, and are guaranteed to appear free
+/// inside `orig`.
+///
+/// In fact, unless there is an unbound identifier, `deps` should be exactly the set of free
+/// variables of `orig`.
```
sir4ur0n, line 102:
```suggestion
///   stores the set of recursive fields it depends on (see [`crate::transform::free_vars`]).
///
/// # Revertible thunks
///
/// Recursive records are better understood as being functions. For example, the recursive record:
///
/// ```nickel
/// {
///   foo = bar + baz + 1,
///   bar = 2,
///   baz = 1,
/// }
/// ```
///
/// is really a builtin representation of `fun self => {foo = self.bar + self.baz + 1, bar = 2, baz
/// = 1}`. The interpreter computes a fixpoint when evaluating a `RecRecord` away to a `Record`,
/// equivalent to `let rec fixpoint = repr fixpoint in fixpoint`. We use a different vocabulary and
/// representations, but this is what we essentially do. We're just hiding the details from the
/// user and pretend recursive records are data that can be used as a normal record, because it's
/// both simple and natural (for the user).
///
/// A naive representation of recursive records could be to always keep recursive records in their
/// functional representation, and re-compute the fixpoint at each record operation, such as field
/// access. While correct, this would be wasteful, because the result of the fixpoint computations
/// is actually the same throughout most record operations (put differently, `self` doesn't
/// change). But we can't just store the computed `fixpoint` either in a normal thunk and forget
/// about the original function, because upon merge, and more specifically upon recursive
/// overriding, we do need to compute a new fixpoint.
///
/// Revertible thunks are a **memoization device** for the `fun self => ...` functional
/// representation of a recursive record. First, note that there are several different but
/// equivalent ways of encoding the example above as a function:
///
/// ```nickel
/// fun self => { foo = self.bar + self.baz + 1, bar = 2, baz = 1 }
///
/// # pushing the function down the fields
/// {foo = fun self => self.bar + self.baz + 1, bar = fun self => 2, baz = fun self => 1}
///
/// # removing unused arguments
/// {foo = fun self => self.bar + self.baz + 1, bar = 2, baz = 1}
///
/// # splitting the self argument
/// {foo = fun bar baz => bar + baz + 1, bar = 2, bar 1}
/// ```
///
/// All those representations are isomorphic, in that modulo basic free variable analysis, we can
/// get from one to another mechanically. Pushing the functions down the fields is better because
/// it reveals more static information about the record (such as the list of fields) without having
/// to provide any argument first. Then, splitting `self` and removing the function altogether when
/// the field doesn't depend on other fields is more precise with respect to dependency tracking
/// (this has an important positive memory impact, see [ThunkDeps]).
///
/// In practice, Nickel uses the last representation. Fields that don't have any free variable
/// intersecting with the name of the other fields are allocated as normal thunks that will only be
/// evaluated at most once. Fields with actual dependencies are stored in revertible thunks, where
/// the `orig` field stores the body of the function (in our example the expression `bar + baz +
/// 1`), and the cached version is the application to `self` computed when we evaluated the
/// recursive record to a normal record. The function arguments (that is, the intersection of the
/// free variables of the expression with the set of fields) are stored inside `deps`. `deps`
/// represent the other fields a specific field is depending on, and are guaranteed to appear free
/// inside `orig`.
///
/// In fact, unless there is an unbound identifier, `deps` should be exactly the set of free
/// variables of `orig`.
```

sir4ur0n (approved):
This is a definite improvement, thank you!

src/eval/lazy.rs
```diff
@@ -35,8 +35,71 @@ pub struct ThunkData {
...
+/// functional representation, and re-compute the fixpoint at each record operation, such as field
+/// access. While correct, this would be wasteful, because the result of the fixpoint computations
+/// is actually the same throughout most record operations (put differently, `self` doesn't
+/// change). But we can't just store the computed `fixpoint` either in a normal thunk and forget
+/// about the original function, because upon merge, and more specifically upon recursive
+/// overriding happen, we do need to compute a new fixpoint.
+///
+/// Revertible thunks are a simple **memoization device** for the `fun self => ...` functional
+/// representation of a recursive record. First, note that there are several different but
+/// equivalent ways of encoding the example above as a function:
```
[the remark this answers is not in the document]
sir4ur0n: As a rule of thumb I always recommend to not use "simple" or synonyms in any form of documentation :sweat_smile: 

At best it is unnecessary, at worst it can make the reader feel dumb.

src/eval/lazy.rs
sir4ur0n: Hmmmm, I actually think it's "free" here, as in "free variable" :sweat_smile: But English is not my strong suit

matthew-healy: I'd use "free" in this context, though as usual with English grammar I can't tell you exactly why. I'd read "are guaranteed to appear freely" as "are guaranteed to appear without restriction" which isn't really what's intended here.

src/eval/lazy.rs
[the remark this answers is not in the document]
yannham: Hmm, I meant simple in the fact that it can only memoize on function call, on one argument, basically. As opposed to say memoizing function on `n` different entries. But I agree `simple` is not very specific here.

src/eval/lazy.rs
yannham: Yup, as Julien says, to appear "free" has a precise mathematical meaning, in the sense that it's a free variable of the expression. I think changing it to "freely" changes the meaning, then.

src/eval/lazy.rs
yannham: Mmh, I don't really know how to merge this suggestion, because I'm already batching Viktor's suggestions. that overlap with yours. If that's not too much work, would it be possible to rather make line-per-line suggestions whenever possible :pray: ?

src/eval/lazy.rs
```diff
@@ -34,9 +34,72 @@ pub struct ThunkData {
 /// - A standard thunk, that is destructively updated once and for all
 /// - A revertible thunk, that can be restored to its original expression. Used to implement
 ///   recursive merging of records and overriding (see the
-///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunks optionally
-///   stores the set of recusive fields it depends on. See the [`crate::transform::free_vars`] for more
-///   details.
+///   [RFC overriding](https://github.com/tweag/nickel/pull/330)). A revertible thunk optionally
+///   stores the set of recusive fields it depends on (see [`crate::transform::free_vars`]).
```
sir4ur0n, line 38:
```suggestion
///   stores the set of recursive fields it depends on (see [`crate::transform::free_vars`]).
```

src/eval/lazy.rs
```diff
@@ -34,9 +34,72 @@ pub struct ThunkData {
...
+/// equivalent ways of encoding the example above as a function:
+///
+/// ```nickel
+/// fun self => { foo = self.bar + self.baz + 1, bar = 2, baz = 1 }
+///
+/// # pushing the function down the fields
+/// {foo = fun self => self.bar + 1, bar = fun self => 2, baz = fun self => 1}
+///
+/// # removing unused arguments
+/// {foo = fun self => self.bar + 1, bar = 2, baz = 1}
```
sir4ur0n, line 78:
```suggestion
/// {foo = fun self => self.bar + self.baz + 1, bar = 2, baz = 1}
```

src/eval/lazy.rs
```diff
@@ -34,9 +34,72 @@ pub struct ThunkData {
...
+///
+/// Revertible thunks are a **memoization device** for the `fun self => ...` functional
+/// representation of a recursive record. First, note that there are several different but
+/// equivalent ways of encoding the example above as a function:
+///
+/// ```nickel
+/// fun self => { foo = self.bar + self.baz + 1, bar = 2, baz = 1 }
+///
+/// # pushing the function down the fields
+/// {foo = fun self => self.bar + 1, bar = fun self => 2, baz = fun self => 1}
```
sir4ur0n, line 75:
```suggestion
/// {foo = fun self => self.bar + self.baz + 1, bar = fun self => 2, baz = fun self => 1}
```

Issue #893, opened 2022-10-26
Error when trying to build VS Code extension with Nix


ebresafegaga:
**Describe the bug**
Currently, when we try to build Nickel's VS Code extension with Nix: 
```
 nix build "./#vscodeExtension.vsix"
```

It fails with the following error: 
```
 last 10 log lines:
 > npm WARN old lockfile     at async RegistryFetcher.manifest (/nix/store/sndm8q6yhr6a8yy9zdd1sxni3lc8pcmw-nodejs-18.8.0/lib/node_modules/npm/node_modules/pacote/lib/registry.js:124:23)
 > npm WARN old lockfile     at async Array.<anonymous> (/nix/store/sndm8q6yhr6a8yy9zdd1sxni3lc8pcmw-nodejs-18.8.0/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/build-ideal-tree.js:784:9) {
 > npm WARN old lockfile   code: 'ENOTCACHED'
 > npm WARN old lockfile }
 > npm ERR! code ENOTCACHED
 > npm ERR! request to https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.1.tgz failed: cache mode is 'only-if-cached' but no cached response is available.
 > 
 > npm ERR! A complete log of this run can be found in:
 > npm ERR!     /build/.npm/_logs/2022-10-26T16_35_58_676Z-debug-0.log
 > 
```

**Expected behavior**
It should build correctly without any errors. We should also have a CI job to build the extension automatically, and make the artifact (`.vsix`) available for download.

**Additional context**
This might be caused by npm2nix

jmgilman:
Can confirm I'm also seeing this on an x86_64-linux NixOS system. Unfortunately, I've come across this error before, and I was never able to figure out the root cause (if you Google it, you'll see a ton of open issues around it).

PR #926, opened 2022-11-21
Fix vscode extension vsix build failure

sir4ur0n:
* Fix VS Code extension Nix build
  * Regenerate the 3 `node2nix` files with a more recent `node2nix`
  * Add the missing `vsce` dependency
  * Pass the `--development` flag to `node2nix` to correctly include dev
    dependencies like `typescript` or `vsce`
* Use the verbose form of `tsc` parameters in `package.json`
* Add `node2nix`, `markdownlint` and `nodejs` to the dev shell
* Add a `README.md` to the VS Code extension directory
* Add the VS Code extension to the CI

Close https://github.com/tweag/nickel/issues/893


flake.nix
```diff
@@ -266,7 +266,12 @@
 
       makeDevShell = { rust }: pkgs.mkShell {
         inputsFrom = [ (buildNickel { inherit rust; withCargoHome = false; }) ];
-        buildInputs = [ pkgs.rust-analyzer ];
+        buildInputs = [
+          pkgs.rust-analyzer
+          pkgs.nodejs
+          pkgs.node2nix
```
yannham, line 272: Does it mean that we will pull `nodejs` and `node2nix` whenever we enter a devshell in Nickel, even if we don't care about VSCode at all? Is it for hacking on the VSCode extension? If yes, how feasible would it be that `lsp/client-extension` gets its own `flake.nix`?

flake.nix
```diff
@@ -266,7 +266,12 @@
...
+          pkgs.nodePackages.markdownlint-cli
```
yannham, line 273: Isn't this already provided by pre-commit-hooks in some way or another?

flake.nix
sir4ur0n: > Does it mean that we will pull nodejs and node2nix whenever we enter a devshell in Nickel, even if we don't care about VSCode at all?

No: it will only pull it the first time, then it will be cached in the Nix store, just like `rust-analyzer`. I think the impact for users will be anywhere between "none" and "insignificant", though maybe someone notices a significant negative impact?

> Is it for hacking on the VSCode extension?

Yes, and (probably in a future PR) to check that the `package.json`, `package-lock.json` and the files generated by `node2nix` remain in sync (otherwise it's too easy to change only the `package.json` in a PR and not notice the problem because CI still builds the old dependencies).

> If yes, how feasible would it be that lsp/client-extension gets its own flake.nix?

I think it is feasible, though it will have a negative impact on UX: 
* devs will need to notice there are several `flake.nix` files and that they are meant to nest them
* devs using `direnv` now need to create nested `.envrc` files

Another alternative is to provide another devShell in the top level `flake.nix`, e.g.
```nix
devShells = {
  default = devShells.stable;
  vscodeExtension = devShells.stable.overrideAttrs (old: {
    buildInputs = old.buildInputs ++ [ pkgs.nodejs pkgs.node2nix ];
  });
};
```

Then devs can either type `nix develop .\#vscodeExtension` to land into the non-default shell, or write `use flake .\#vscodeExtension` in their `.envrc` file.

We can even expand a bit, and rename it to `full` or something.

Now that being said, I am generally in favor of following the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), that is in this case: devs expect the default dev shell to provide everything needed to hack on the project.

So if we **really** don't want to "force" those VS Code extensions in a devshell, I would go for the opposite:
* have the **default** devshell provide everything
* have another devshell provide more minimalistic devshells (e.g. `.\#rustOnly`)

But I want to check just to be sure: what would we optimize by providing a shell without those dependencies? Besides a few MBs of disk, a few seconds to download the binary cache the first time for each user. I worry we would unnecessarily optimize a non-problem :sweat_smile:

vkleen: FWIW, I tried a checkout of this PR with direnv setup and didn't notice anything out of the ordinary. So the extra buildInputs must be in cachix already, I guess.

yannham: The closure size of `nodejs` is 283MB. It's not unsustainable but it's not a few MBs either (if, like me, you have a crappy internet albeit living in a capital city :cry: ).

Maybe it just means we should separate the VSCode extension from the main repo, dunno.

> I think it is feasible, though it will have a negative impact on UX:
>
>    - devs will need to notice there are several flake.nix files and that they are meant to nest them
>    - devs using direnv now need to create nested .envrc files

If you're hacking on the VSCode extension, and we put a `flake.nix` inside the vscode extension (that you now need to call to build the extension), I imagine you will know it exists. The question may then be: do you expect it to define a specific shell environment. I don't know the usage very well, but I personally would, since a flake is a generic project interface and may indeed define a devshell output  (in the extreme case, if we had one huge mono repo, you probably aren't gonna provide an environment with ten language toolchains).

But this is apparently not your case:

> that is in this case: devs expect the default dev shell to provide everything needed to hack on the project.

Is this from experience or your personal expectation?

> Yes, and (probably in a future PR) to check that the package.json, package-lock.json and the files generated by node2nix remain in sync (otherwise it's too easy to change only the package.json in a PR and not notice the problem because CI still builds the old dependencies).

Out of curiosity, how is this related to the devshell? This sounds like a CI job, unless you want to check that in the shell as a pre-commit-hook or something?

In any case, I really don't have a strong opinion about this, it's just random questions that popped up while reviewing. Maybe splitting in two flakes also duplicate code between the flakes, if the subflake needs things that aren't exported by the root flake. I'm also not using nix-direnv. If you think it's more trouble that what it's worth, I trust your judgement.

yannham (approved):
The discussion about dev shells is interesting, but let's not block this PR: the important thing is that this fixes the VSCode extension :tada:

dpl0a (approved):

flake.nix
sir4ur0n: I have created https://github.com/tweag/nickel/issues/941 so that those questions don't end up in limbo.

> Out of curiosity, how is this related to the devshell? This sounds like a CI job, unless you want to check that in the shell as a pre-commit-hook or something?

My bad, it's not, I mixed up with the way we used to build in CI in my previous project!

PR #927, opened 2022-11-21
Move comparison table to main README; add JSON, YAML and TOML

sir4ur0n:
I don't know how this PR will be received but I think it's worth having this conversation :smile: 

This PR contains 2 commits but we can remove one of them (I coupled them in the same PR to avoid conflicts).

* Move the comparison table with other languages to the main README
   I have always felt this was needed. When users evaluate a new tech, they want to quickly compare to familiar things, to understand what they would get and lose at first glance. This table should thus be promoted to the main README, and the section renamed
* Add JSON, YAML and TOML in the comparison list
   As silly as it may seem, I think this is legitimate to add those in the comparison.

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -347,16 +347,20 @@ The main difference between Jsonnet and Nickel are types. Jsonnet does not
 feature static types, contracts or enriched values, and thus can't type library
 code and has no principled approach to data validation.
 
-### Summary
+### Comparison with other configuration languages
+<!-- Intentionally duplicated in `README.md`, please update the other one for any change done here -->
 
-| Language | Typing                        | Recursion  | Evaluation | Side-effects                                      |
-|----------|-------------------------------|------------|------------|---------------------------------------------------|
-| Nickel   | Gradual (dynamic + static)    | Yes        | Lazy       | Yes (constrained)                                 |
-| Starlark | Dynamic                       | No         | Strict     | No                                                |
-| Nix      | Dynamic                       | Yes        | Lazy       | Predefined and specialized to package management  |
-| Dhall    | Static (requires annotations) | Restricted | Lazy       | No                                                |
-| CUE      | Static (everything is a type) | No         | Lazy       | No, but allowed in the separated scripting layer  |
-| Jsonnet  | Dynamic                       | Yes        | Lazy       | No                                                |
+| Language | Typing                        | Recursion  | Evaluation | Side-effects                                     |
+|----------|-------------------------------|------------|------------|--------------------------------------------------|
+| Nickel   | Gradual (dynamic + static)    | Yes        | Lazy       | Yes (constrained, planned)                       |
+| Starlark | Dynamic                       | No         | Strict     | No                                               |
+| Nix      | Dynamic                       | Yes        | Lazy       | Predefined and specialized to package management |
+| Dhall    | Static (requires annotations) | Restricted | Lazy       | No                                               |
+| CUE      | Static (everything is a type) | No         | Lazy       | No, but allowed in the separated scripting layer |
+| Jsonnet  | Dynamic                       | Yes        | Lazy       | No                                               |
+| JSON     | None                          | No         | Strict     | No                                               |
+| YAML     | None                          | No         | N/A        | No                                               |
+| TOML     | None                          | No         | N/A        | No                                               |
 
 ## Conclusion
 
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ comparison with these languages.
 
 | Language | Typing                        | Recursion  | Evaluation | Side-effects                                     |
 |----------|-------------------------------|------------|------------|--------------------------------------------------|
-| Nickel   | Gradual (dynamic + static)    | Yes        | Lazy       | Yes (constrained)                                |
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -242,4 +242,18 @@ to be used in production. The next steps we plan to work on are:
...
 
 See [RATIONALE.md](./RATIONALE.md) for the design rationale and a more detailed
-comparison with a selection of these languages.
+comparison with these languages.
+
+### Summary
+
+| Language | Typing                        | Recursion  | Evaluation | Side-effects                                     |
+|----------|-------------------------------|------------|------------|--------------------------------------------------|
+| Nickel   | Gradual (dynamic + static)    | Yes        | Lazy       | Yes (constrained)                                |
```
vkleen, line 251: I think currently, the side-effects are a lie. Maybe we should change that to `Yes (constrained, planned)`?
That doesn't have much to do with this PR, though.

README.md
```diff
@@ -242,4 +242,18 @@ to be used in production. The next steps we plan to work on are:
...
+| Starlark | Dynamic                       | No         | Strict     | No                                               |
+| Nix      | Dynamic                       | Yes        | Lazy       | Predefined and specialized to package management |
+| Dhall    | Static (requires annotations) | Restricted | Lazy       | No                                               |
+| CUE      | Static (everything is a type) | No         | Lazy       | No, but allowed in the separated scripting layer |
+| Jsonnet  | Dynamic                       | Yes        | Lazy       | No                                               |
+| JSON     | None                          | No         | Strict     | No                                               |
```
yannham, line 257: I don't think talking about strict evaluation for data language makes a lot of sense. What about something like `N/A`?

README.md
```diff
@@ -208,7 +208,7 @@ to be used in production. The next steps we plan to work on are:
   [overriding proposal](https://github.com/tweag/nickel/blob/9fd6e436c0db8f101d4eb26cf97c4993357a7c38/rfcs/001-overriding.md))
 - Performance improvements
 
-## Related projects and inspirations
+## Comparison
```
yannham, line 211: Mmh, I don't know about this change. Comparison is shorter but also carries much less information: comparison between what and what? Nix and Bazel? The different way of building Nickel, with Nix and with Cargo? Comparison of performances with and without contracts? Etc.

yannham (commented):
I think it's a fair proposal :+1: but I would also let the original table inside RATIONALE.md. It's duplicating data (or, should we write the README and RATIONALE in Nickel :trollface: ?) but the rationale is quite verbose and having a small and tight recap of the comparison is useful there, IMHO.

README.md
sir4ur0n: "Comparison with alternatives"?

sir4ur0n: Basically I think "comparison" is important because many users when considering a new tech, want a comparison with other, familiar tech. I think "Related projects and inspirations" does not quite capture that.

I'll take any suggestion though, as long as we help readers find what they are looking for

yannham: I see, then Comparison with alternatives sounds better, yes. Comparison with related projects, Comparison with alternative projects, Comparison with other configuration languages?

matthew-healy:
This isn't a comment on this change as such (though I'm fully in-favour of having this in the README), but I think it might also make sense to have this information on nickel-lang.org in some form as well...

README.md
sir4ur0n: Done in 3a283d2f1f85142c605ec3706ed72418cae60240

README.md
sir4ur0n: Done in 13a35021dfab9e7d0436e4245dca41935d1f19c3

README.md
sir4ur0n: I went with "Comparison with other configuration languages":
* "projects" seems weird for users: users don't want to know if JSON or Dhall are projects, they are only interested in the end product, a configuration language
* repeating "configuration languages" is a nice wrap-up to remind what is Nickel, and echoes to the 1st line of the README

Done in 47bbb38b3a5201aaa3a22816339c5e78c767dce0

RATIONALE.md
```diff
@@ -347,16 +347,20 @@ The main difference between Jsonnet and Nickel are types. Jsonnet does not
...
-
-| Language | Typing                        | Recursion  | Evaluation | Side-effects                                      |
-|----------|-------------------------------|------------|------------|---------------------------------------------------|
-| Nickel   | Gradual (dynamic + static)    | Yes        | Lazy       | Yes (constrained)                                 |
-| Starlark | Dynamic                       | No         | Strict     | No                                                |
-| Nix      | Dynamic                       | Yes        | Lazy       | Predefined and specialized to package management  |
-| Dhall    | Static (requires annotations) | Restricted | Lazy       | No                                                |
-| CUE      | Static (everything is a type) | No         | Lazy       | No, but allowed in the separated scripting layer  |
-| Jsonnet  | Dynamic                       | Yes        | Lazy       | No                                                |
+### Comparison with other configuration languages
```
sir4ur0n, line 350: @yannham I duplicated in 495e74280c70ff059f940903488c787e82aa280a

vkleen (approved):
Looks good 👍

yannham (approved):

PR #928, opened 2022-11-22
[Refactor] Renaming `tyw` to `uty`

yannham:
Gets rid of the last references to the old name of `UnifType` (previously `TypeWrapper`) in function, macro or variable names after #875.


matthew-healy (approved):
Yay, consistency! 🙌

PR #940, opened 2022-11-23
Revertible thunks as function II: fix recursive overriding

### LINKED ISSUE #908: Recursive overriding is nonassociative

yannham:
# Revertible thunks are functions

Closes #908.

Follow-up of #924: bring the presentation of revertible thunks as just functions  memoization devices to its logical conclusion by providing helper to explicitly rebuild the represented function and apply it. This explicit application was required to fix recursive overriding in the general case, see #908. Now, the implementation faithfully expresses general merging of recursive fields as `common_field = fun self => (left_field self) & (right_field self)`, when recursive record fields are understood as functions of `self` (representing the ambient record).

## Revertible thunks representation

This PR also reworks the representation of revertible thunks. They used to store the cached data as an `Rc<Closure>`, initialized as soon as the revertible thunks was initialized. This doesn't model revertible thunks faithfully: the cached value should be owned (as it is for normal thunks), and as long as the revertible thunks hasn't be "patched" (or, when seen as a function, applied), the cached value is meaningless and may contain free variables, id est unbound identifiers.

The cached value should never be used before full initialization. This PR sets the cached value to be an `Option<Closure>`, and to panic if it used before initialization. It also ensures that initialization is done only once. The previous implementation wouldn't complain about such violation. Note that we have no choice to perform the construction of a revertible thunks in two steps, and thus observe a transitory state where the cached value is not initialized, because we are potentially building cyclic values. However, if the implementation is correct, the state where the cached value is set to `None` shouldn't really be observable.

### DIFF
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -578,6 +578,26 @@ pub enum ThunkDeps {
     Empty,
 }
 
+impl ThunkDeps {
+    /// Compute the union of two thunk dependencies. [`ThunkDeps::Unknown`] can be see as the top
+    /// element, meaning that if one of the two set of dependencies is [`ThunkDeps::Unknown`], so
+    /// is the result.
+    pub fn union(self, other: Self) -> Self {
+        match (self, other) {
+            (ThunkDeps::Empty, ThunkDeps::Empty) => ThunkDeps::Empty,
+            // If one of the field has unknown dependencies (understand: may depend on all the other
+            // fields), then the resulting fields has unknown dependencies as well
+            (ThunkDeps::Unknown, _) | (_, ThunkDeps::Unknown) => ThunkDeps::Unknown,
+            (ThunkDeps::Empty, ThunkDeps::Known(deps))
+            | (ThunkDeps::Known(deps), ThunkDeps::Empty) => ThunkDeps::Known(deps),
+            (ThunkDeps::Known(deps1), ThunkDeps::Known(deps2)) => {
+                let union: HashSet<Ident> = deps1.union(&*deps2).cloned().collect();
+                ThunkDeps::Known(Rc::new(union))
+            }
+        }
+    }
+}
+
 /// A thunk update frame.
 ///
 /// A thunk update frame is put on the stack whenever a variable is entered, such that once this
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -148,7 +148,7 @@ impl ThunkData {
     ///
     /// This function is similar in spirit to setting the cached value to be the explicit function
     /// application given as built by `saturate`, but applied to arguments taken
-    /// from `rec_env`. The major difference is that `init_cached` avoid the creation of the
+    /// from `rec_env`. The major difference is that `init_cached` avoids the creation of the
     /// intermediate redex `(fun id1 .. id n => orig) %1 .. %n` as well as the intermediate thunks
     /// and terms, because we can compute the result application right away, in-place.
     pub fn init_cached(&mut self, rec_env: &[(Ident, Thunk)]) {
@@ -163,7 +163,7 @@ impl ThunkData {
                 // an invariant that MUST be maintained by the interpreter.
                 //
                 // `cached` set to `None` solely exists because we need to first allocate all the
-                // revertible thunks corresponding to a recursive record, and only then we can
+                // revertible thunks corresponding to a recursive record, and only then can we
                 // patch them (build the cached value) in a second step, but they should be
                 // logically seen as one construction operation.
                 assert!(
@@ -354,7 +354,7 @@ impl ThunkData {
 /// revertible thunks. Most expressions don't need revertible thunks as their evaluation will
 /// always give the same result, but some others, such as the ones containing recursive references
 /// inside a record may be invalidated by merging, and thus need to store the unaltered original
-/// expression. Those aspects are mainly handled and discussed into more details in
+/// expression. Those aspects are handled and discussed in more detail in
 /// [InnerThunkData].
 #[derive(Clone, Debug, PartialEq)]
 pub struct Thunk {
@@ -487,7 +487,7 @@ impl Thunk {
     /// # Standard thunks (non-revertible)
     ///
     /// Non revertible thunks can be seen as a special case of revertible thunks with no
-    /// dependencies. Thus the abstraction and application are zero-ary, and the result is just the
+    /// dependencies. Thus the abstraction and application are nullary, and the result is just the
     /// current thunk closurized in `env` as a fresh variable.
     ///
     /// # Example
--- a/src/eval/lazy.rs
+++ b/src/eval/lazy.rs
@@ -1,5 +1,5 @@
 //! Thunks and associated devices used to implement lazy evaluation.
-use super::{Closure, IdentKind};
+use super::{Closure, Environment, IdentKind};
 use crate::{
     identifier::Ident,
     term::{FieldDeps, RichTerm, Term},
@@ -31,6 +31,13 @@ pub enum ThunkState {
 pub struct ThunkData {
     inner: InnerThunkData,
     state: ThunkState,
+}
+
+/// The two different kind of thunks.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ThunkType {
+    Standard,
+    Revertible,
 }
 
 /// The part of [ThunkData] responsible for storing the closure itself. It can either be:
@@ -124,11 +131,9 @@ impl ThunkData {
 
     /// Create new revertible thunk data.
     pub fn new_rev(orig: Closure, deps: FieldDeps) -> Self {
-        let rc = Rc::new(orig);
-
         ThunkData {
             inner: InnerThunkData::Revertible {
-                orig: rc.clone(),
+                orig: Rc::new(orig),
                 cached: None,
                 deps,
             },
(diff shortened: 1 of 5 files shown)


src/eval/lazy.rs
```diff
@@ -121,41 +131,149 @@ impl ThunkData {
...
     }
 
     /// Consume the data and return the cached closure.
     pub fn into_closure(self) -> Closure {
         match self.inner {
             InnerThunkData::Standard(closure) => closure,
-            InnerThunkData::Revertible { orig, cached, .. } => {
-                std::mem::drop(orig);
-                Rc::try_unwrap(cached).unwrap_or_else(|rc| (*rc).clone())
+            // Nothing should access the cached value of a revertible thunks before the cached
```
yannham, line 266:
```suggestion
            // Nothing should access the cached value of a revertible thunk before the cached
```

src/eval/lazy.rs
```diff
@@ -333,6 +459,95 @@ impl Thunk {
...
 
+    pub fn build_cached(&mut self, rec_env: &[(Ident, Thunk)]) {
+        self.data.borrow_mut().init_cached(rec_env)
+    }
+
+    /// Revert a thunk, abstract over its dependencies to get back a function, and apply the
+    /// function to the given variables. The function part is allocated in a new fresh thunk,
+    /// stored as a generated variable, with the same environment as the original expression.
+    ///
+    /// Recall that revertible thunks are just a memoization mechanism) for function application.
```
yannham, line 470:
```suggestion
    /// Recall that revertible thunks are just a memoization mechanism for function application.
```

src/eval/lazy.rs
```diff
@@ -333,6 +459,95 @@ impl Thunk {
...
+    /// manipulate the body of the function directly (which is what is stored inside the `orig`
+    /// field).
+    ///
+    /// However, in the general case of merging two record fields which may be both recursive (i.e.
+    /// which may contain a revertible thunk), we have to use the explicit function representation
+    /// and apply it to variables, which correspond to the fields of the recursive record being
+    /// built by merging.
+    ///
+    /// `saturate`:
+    /// - abstract the original expression of the underlying revertible thunk, forming a function.
```
yannham, line 482:
```suggestion
    /// - abstracts the original expression of the underlying revertible thunk, forming a function.
```

src/eval/lazy.rs
```diff
@@ -333,6 +459,95 @@ impl Thunk {
...
+    /// - store this function in a fresh standard thunk
```
yannham, line 483:
```suggestion
    /// - stores this function in a fresh standard thunk
```

src/eval/lazy.rs
```diff
@@ -333,6 +459,95 @@ impl Thunk {
...
+    /// - return the application of this function to the provided record field names (as variables)
```
yannham, line 484:
```suggestion
    /// - returns the application of this function to the provided record field names (as variables)
```

src/eval/lazy.rs
```diff
@@ -333,6 +459,95 @@ impl Thunk {
...
+    ///
+    /// Non revertible thunks can be seen as a special case of revertible thunks with no
+    /// dependencies. Thus the abstraction and application are zero-ary, and the result is just the
+    /// current thunk closurized in `env` as a fresh variable.
+    ///
+    /// # Example
+    ///
+    /// If `orig` is `foo + bar + a` where `foo` and `bar` are thunk dependencies (hence are free
+    /// variables) and `a` is bound in the environment. Say the iterator represents the fields
+    /// `bar, b, foo` in that order. Then `rev_thunk_as_explicit_app`:
```
yannham, line 511:
```suggestion
    /// `bar, b, foo` in that order. Then `saturate`:
```

src/eval/lazy.rs
```diff
@@ -121,41 +124,149 @@ impl ThunkData {
...
+            }
+        }
+    }
+
     /// Return a reference to the closure currently cached.
     pub fn closure(&self) -> &Closure {
         match self.inner {
             InnerThunkData::Standard(ref closure) => closure,
-            InnerThunkData::Revertible { ref cached, .. } => cached,
+            // Nothing should peek into a revertible thunks before the cached value has been
```
yannham, line 228:
```suggestion
            // Nothing should peek into a revertible thunk before the cached value has been
```

src/eval/merge.rs
```diff
@@ -500,8 +500,110 @@ fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
     doc1.or(doc2)
 }
 
+/// Take the content of a record field, and saturate the potential revertible thunk with the given
+/// fields. See [crate::eval::lazy::Thunk::saturate].
+///
+/// If the expression is not a variable referring to a thunk (this can happen e.g. for numeric
+/// constant), we just return the term as it is, which falls into the zero dependencies special
```
yannham, line 507:
```suggestion
/// constants), we just return the term as it is, which falls into the zero dependencies special
```

src/eval/merge.rs
```diff
@@ -500,8 +500,110 @@ fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
...
+/// case as well.
```
yannham, line 508:
```suggestion
/// case.
```

src/eval/merge.rs
```diff
@@ -500,8 +500,110 @@ fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
...
+
+/// Take the current environment, two fields with their local environment, and return a term which
+/// is the merge of the two fields, closurized in the provided final environment.
+///
+/// The thunk allocated for the result is revertible if and only if at least one of the original
+/// thunk is (if one of the original value is overridable, then so is the merge of the two). In
+/// this case, the field dependencies are the union of the dependencies of each field.
+///
+/// The fields are saturaed (see [saturate]) to properly propagate recursive dependencies down to
+/// the `t1` and `t2` in the final, merged record.
```
yannham, line 547:
```suggestion
/// `t1` and `t2` in the final, merged record.
```

src/eval/merge.rs
```diff
@@ -500,8 +500,93 @@ fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
...
+    } else {
+        Ok(ThunkDeps::Empty)
+    }
+}
+
+/// Take the current environment, two fields with their local environment, and return a term which
+/// is the merge of the two fields, closurized in the provided final environment.
+///
+/// The thunk allocated for the result is revertible if and only if at least one of the original
+/// thunk is (if one of the original value is overridable, then so is the merge of the two). In
```
vkleen, line 543:
```suggestion
/// thunks is (if one of the original values is overridable, then so is the merge of the two). In
```

src/eval/merge.rs
```diff
@@ -500,8 +500,93 @@ fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
...
+/// this case, the field dependencies are the union of the dependencies of each field.
+///
+/// The fields are saturaed (see [saturate]) to properly propagate recursive dependencies down to
```
vkleen, line 546:
```suggestion
/// The fields are saturated (see [saturate]) to properly propagate recursive dependencies down to
```

vkleen (approved):
I've just found some more typos in the comments. The logic seems good, as far as I can tell 👍

src/eval/merge.rs
```diff
@@ -401,32 +384,41 @@ pub fn merge(
...
             let mut env = Environment::new();
 
-            for (field, t) in left.into_iter() {
-                m.insert(field, t.closurize(&mut env, env1.clone()));
-            }
-
-            for (field, t) in right.into_iter() {
-                m.insert(field, t.closurize(&mut env, env2.clone()));
-            }
+            // Merging recursive record is the one operation that may override recursive fields. To
```
yannham, line 396:
```suggestion
            // Merging recursive records is the one operation that may override recursive fields. To
```

matthew-healy (commented):
As far as I can tell, this looks good, though I have to say my mental model for exactly what's going on & how it all fits together isn't great and as such I don't feel qualified to give an approving review.

I left one or two small comments but just for minor things. I don't think they need to block merging.

src/eval/lazy.rs
```diff
@@ -121,41 +124,149 @@ impl ThunkData {
...
+            // value has been constructed. This is an invariant that MUST be maintained by the
+            // interpreter
+            //
+            // `cached` set to `None` solely exists because we need to first allocate all the
+            // revertible thunks corresponding to a recursive record, and only then can we patch
+            // them (build the cached value) in a second step. But calling to
+            // [`ThunkData::new_rev`] followed by [`ThunkData::build_cached_value`] should be logically
+            // seen as just one construction operation.
+            InnerThunkData::Revertible { cached, .. } => {
+                cached.expect("tried to get data from a revertible thunk without a cached value")
```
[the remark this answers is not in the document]
yannham: Are you talking about the comment or the error message inside `expect` ? I imagine you meant the latter, in which case I agree.

src/eval/lazy.rs
```diff
@@ -333,6 +445,95 @@ impl Thunk {
...
+            data: Rc::new(RefCell::new(
+                inner.revthunk_as_explicit_fun(fields.clone().filter(&mut deps_filter)),
+            )),
+            ident_kind: self.ident_kind,
+        };
+
+        let fresh_var = Ident::fresh();
+        env.insert(fresh_var, thunk_as_function);
+
+        let as_function = RichTerm::from(Term::Var(fresh_var));
```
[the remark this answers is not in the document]
yannham: Mh, that's a hard one. It's as a function as a var :) as the var is the closurized version of the function. Maybe `as_function_closurized`?

src/eval/lazy.rs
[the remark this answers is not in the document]
matthew-healy: I actually meant _both_. I'm suggesting combining them so that the comment _is_ the error message passed to expect. I'm not sure it's a perfect solution, but I can definitely see a future where the comment gets changed in one place but gets missed in others. 

That said, I completely understand if you decide not to do this, as it does mean that devs need to go look up the static string definition to find the info that's currently in the comment...

src/eval/lazy.rs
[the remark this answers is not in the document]
matthew-healy: Yeah, I think that's slightly clearer. It just confused me a little to see `as_function` being assigned to a `Term::Var`. :sweat_smile:

src/eval/lazy.rs
```diff
@@ -121,41 +124,149 @@ impl ThunkData {
...
+                );
+
+                let mut new_cached = Closure::clone(orig);
+
+                match deps {
+                    Some(deps) if deps.is_empty() => (),
+                    Some(deps) => new_cached
+                        .env
+                        .extend(rec_env.iter().filter(|(id, _)| deps.contains(id)).cloned()),
+                    None => new_cached.env.extend(rec_env.iter().cloned()),
```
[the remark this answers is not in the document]
yannham: Tried that, but the diff is bigger than I would like. Will do so in a follow-up PR.

src/eval/lazy.rs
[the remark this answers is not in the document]
yannham: Mmh, I think I like the idea of the comment being in the right place, directly accessible. The text showed in `expect` doesn't have exactly the same purpose. I'll still extract it in a separate variable though

PR #949, opened 2022-11-28
[refactor] Re-implement polymorphic record contracts

### LINKED ISSUE #201: Alternative implementations of record contracts
In [#169][github-169] (related issue [#157][github-157]), contracts for records have been implemented. This gives a first working implementation, but the PR led to some discussion about other possible designs or implementation. This issues gathers ideas about alternatives.

## Context

The non-trivial part about record contracts is polymorphism. Take a function of type `forall a. {field: Num | a} -> {field: Num, field2: Str | a}`. Checking the non-polymorphic part of the type is simple: check that the argument has a field `field` of type `Num` and that the return value have the same field, plus a field `field2` of type `Str`. But the function can violate the polymorphic part of the contract in several ways. Here are some examples:

```
//1: remove a field from the polymorphic part
let wrong = fun r =>
  let r = if hasField "foo" r then r -$ "foo" else r in
  r$["field2" = "some string"]

//2: add a field to the polymorphic part
let wrong = fun r =>
  let r = if !(hasField "foo" r) then r$["foo"=2] else r in
  r$["field2" = "some string"]

//3: rewrite a field from the polymorphic part, possibly changing the type
let wrong = fun r =>
  let r = if hasField "foo" r then r -$ "foo" $["foo"="rewritten"] in
  r$["field2" = "some string"]

//4: rewrite a field from the polymorphic part without changing the type
let wrong = fun r =>
  let r = if hasField "foo" r && isNum r.foo then r -$ "foo" $["foo"=0] in
  r$["field2" = "some string"]

//5: make the result depend on the polymorphic part
let wrong = fun r =>
  let str = if hasField "foo" r then "haz foo!" else "not haz foo" in
  r$["field2" = str]
```

Examples 1, 2 and 3 are clear violations of type soundness, as they change the meaning of `a` between the input and the output. Examples 4 and 5 are less clear-cut: they violate parametricity, which is in general expected from parametric polymorphism, but they don't break type soundness. They are also not of the same severity: 4. changes a value from the polymorphic part, while 5. just inspects it and makes its return result depend on the shape of the polymorphic part.

## Question
The common wisdom is that contracts should ensure parametricity [\[1\]](https://www.ccs.neu.edu/home/amal/papers/blameall.pdf), as is already the case for non-record types in Nickel. A standard technique to achieve this is dynamic sealing: for example, the contract of `forall a. a -> a` seals the argument with a special construct (turning `M` to `Wrapped(a,M)`), which prevents inspection. It can only be passed to other functions or returned. At the end, the contract tries to unseal the result, checking that type variables correspond (in this case this is useless, but in types containing several variables, this forbids ill-typed functions like `let f = Assume(forall a b. a -> b -> a, fun x y => y)`).

Implementing this for records raises the following questions:
- At what granularity do we enforce parametricity ? In particular, while 4. looks like it should not be possible, is 5. really a problem ?
- how do we implement it ?

## Proposals
The criterion are the following:
- **invasiveness**: does it require to change the current abstract machine ? Does it add complexity?
- **performance**: does it inflict a penalty performance on non-polymorphic code ?
- **parametricity**: what level of parametricity does it ensure ?
- **error reporting**

### Coarse sealing through magic field
This is the implementation of #169. Coarse alludes to the fact that the polymorphic tail is hidden as a whole. It is stored as a new record in a special field `_%wrapped`.
- invasiveness: minimal, it uses only existing features.
- performance: it does not affect non-polymorphic code, and the operations used are rather simple.
- parametricity: it enforces full parametricity, as the tail is totally inaccessible.
- error reporting: one questionable consequence is that accessing a field of the tail creates a `missing field error`, instead of a blame error, for example in this case `Assume(forall a. { | a} -> Num, fun r => r.foo)`.

Another consequence is that it add a fields `_%wrapped` which can be observed via `hasField` and `fieldsOf`. Its content is sealed, and it is always present for polymorphic record arguments, so no information leaks, but it may interfere with standard user code (say, one that relies on the precise result of `fieldsOf`).

### Coarse sealing built-in
This is the same solution as above, but where the magic field of record becomes a built-in feature.
- invasiveness: need to add a `Option<Box<Term>>` field to records, and rewrite in consequence all records operations (though in a pretty mechanical way). We need to add primitive operators to set or get the content of this field.
- performance: non-polymorphic contracts have an additional pointer. Other than that, usual operations should not suffer any penalty.
 - parametricity: as above, full parametricity.
- error reporting: this gives the possibility to raise a blame error when accessing a field which exists but is sealed in the polymorphic part.

### Fine-grained sealing
Another approach is to be more granular, and seal each field of the tail independently. A naive implementation would allow a function to inspect at least the shape of the polymorphic row. On the other hand, one can maintain say an access table of the sealed fields to prevent this. In this case, it is easier to control what can be done on sealed fields: we could make them appear in the `fieldsOf` list or not independently of them being accessible or not.
- invasiveness: this approach is more invasive, as all records operation needs to be aware that some fields are different from others, and to adopt a different behavior accordingly. Records must also be extended with an access table.
- performance: same as for invasiveness, this approach require additional checks for all record operations, and probably interact with merging as well.
- parametricity: customizable.
- error reporting: customizable.

### Conclusion
The fine-grained approach is cited for exhaustiveness, but I don't think it's worth the trouble as there is currently no incentive to allow some operations on the sealed part. The first approach is now implemented, and if the drawbacks (presence of an additional magic field in sealed records plus non-blame errors in some cases) are found to be problematic, the second approach mitigates them.

[github-169]: https://github.com/tweag/nickel/pull/169
[github-157]: https://github.com/tweag/nickel/issues/157

matthew-healy:
This PR re-implements polymorphic record contracts to use a pair of seal/unseal primitive operations. This has the benefit of preventing the sealed tail from ever being accessed from user code (whereas in the previous implementation functions like `record.fields` would leak the existence of this sealed field). 

Merging this will close #201.

This change will also make it possible to later implement improved error messaging. For example: correctly assigning blame if a function with a polymorphic record type attempts to access a sealed field (#952), or knowing at the point a function is called which fields a polymorphic record argument _cannot_ contain in the head (#950).

### How to review

Each commit is atomic and passes the whole test suite. Most commits also have a message which explains their contents in a little more detail. As such I would recommend reviewing this commit-by-commit.

### Notes

The error messages for mapping over or merging a record with a sealed tail aren't great right now. One simple way to improve them would be to explicitly set the `tag` of the label before raising the error (see [this comment](https://github.com/tweag/nickel/pull/949/files#r1033797621)), but I couldn't find anywhere else where we did that.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -25,6 +25,8 @@ use crate::{
     types::{TypeF, Types},
 };
 
+use self::blame_error::ExtendWithCallStack;
+
 /// A general error occurring during either parsing or evaluation.
 #[derive(Debug, Clone, PartialEq)]
 pub enum Error {
@@ -92,10 +94,33 @@ pub enum EvalError {
         String,  /* error message */
         TermPos, /* position of the call to deserialize */
     ),
+    /// A polymorphic record contract was broken somewhere.
+    IllegalPolymorphicTailAccess {
+        action: IllegalPolymorphicTailAction,
+        label: label::Label,
+        call_stack: CallStack,
+    },
     /// An unexpected internal error.
     InternalError(String, TermPos),
     /// Errors occurring rarely enough to not deserve a dedicated variant.
     Other(String, TermPos),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum IllegalPolymorphicTailAction {
+    Map,
+    Merge,
+}
+
+impl IllegalPolymorphicTailAction {
+    fn message(&self) -> &'static str {
+        use IllegalPolymorphicTailAction::*;
+
+        match self {
+            Map => "cannot map over a record sealed by a polymorphic contract",
+            Merge => "cannot merge a record sealed by a polymorphic contract",
+        }
+    }
 }
 
 /// An error occurring during the static typechecking phase.
@@ -690,113 +715,6 @@ fn secondary_term(term: &RichTerm, files: &mut Files<String>) -> Label<FileId> {
     secondary_alt(term.pos, term.as_ref().shallow_repr(), files)
 }
 
-/// Generate a codespan label that describes the [type path][crate::label::TyPath] of a (Nickel)
-/// label, and notes to hint at the situation that may have caused the corresponding error.
-fn report_ty_path(l: &label::Label, files: &mut Files<String>) -> (Label<FileId>, Vec<String>) {
-    let end_note = String::from("Note: this is an illustrative example. The actual error may involve deeper nested functions calls.");
-
-    let (msg, notes) = if l.path.is_empty() {
-        (String::from("expected type"), Vec::new())
-    } else if ty_path::has_no_arrow(&l.path) {
-        match l.path.last() {
-            Some(ty_path::Elem::Array) => (String::from("expected array element type"), Vec::new()),
-            Some(ty_path::Elem::Field(_)) => (String::from("expected field type"), Vec::new()),
-            _ => unreachable!(),
-        }
-    }
-    // If the path is only composed of codomains, polarity is necessarily true and the cause of the
-    // blame is the return value of the function
-    else if ty_path::is_only_codom(&l.path) {
-        (
-            String::from("expected return type"),
-            vec![
-                String::from(
-                    "This error may happen in the following situation:
-1. A function `f` is bound by a contract: e.g. `Bool -> Num`.
-2. `f` returns a value of the wrong type: e.g. `f = fun c => \"string\"` while `Num` is expected.",
-                ),
-                String::from(
-                    "Either change the contract accordingly, or change the return value of `f`",
-                ),
-            ],
-        )
-    } else {
-        // We ignore the `Field` and `Array` elements of the path, since they do not impact
-        // polarity, and only consider "higher-order" elements to customize error messages.
-        let last = l
-            .path
-            .iter()
-            .filter(|elt| matches!(*elt, ty_path::Elem::Domain | ty_path::Elem::Codomain))
-            .last()
-            .unwrap();
-        match last {
-            ty_path::Elem::Domain if l.polarity => {
-                (String::from("expected type of an argument of an inner call"),
-                 vec![
-                     String::from("This error may happen in the following situation:
-1. A function `f` is bound by a contract: e.g. `(Str -> Str) -> Str)`.
-2. `f` takes another function `g` as an argument: e.g. `f = fun g => g 0`.
-3. `f` calls `g` with an argument that does not respect the contract: e.g. `g 0` while `Str -> Str` is expected."),
-                     String::from("Either change the contract accordingly, or call `g` with a `Str` argument."),
-                     end_note,
-                 ])
-            }
-            ty_path::Elem::Codomain if l.polarity => {
-                (String::from("expected return type of a sub-function passed as an argument of an inner call"),
-                 vec![
-                     String::from("This error may happen in the following situation:
-1. A function `f` is bound by a contract: e.g. `((Num -> Num) -> Num) -> Num)`.
-2. `f` take another function `g` as an argument: e.g. `f = fun g => g (fun x => true)`.
-3. `g` itself takes a function as an argument.
-4. `f` passes a function that does not respect the contract to `g`: e.g. `g (fun x => true)` (expected to be of type `Num -> Num`)."),
-                     String::from("Either change the contract accordingly, or call `g` with a function that returns a value of type `Num`."),
-                     end_note,
-                 ])
-            }
-            ty_path::Elem::Domain => {
-                (String::from("expected type of the argument provided by the caller"),
-                 vec![
-                     String::from("This error may happen in the following situation:
-1. A function `f` is bound by a contract: e.g. `Num -> Num`.
-2. `f` is called with an argument of the wrong type: e.g. `f false`."),
-                     String::from("Either change the contract accordingly, or call `f` with an argument of the right type."),
-                     end_note,
-                 ])
-            }
-            ty_path::Elem::Codomain => {
-                (String::from("expected return type of a function provided by the caller"),
-                 vec![
-                     String::from("This error may happen in the following situation:
-1. A function `f` is bound by a contract: e.g. `(Num -> Num) -> Num`.
-2. `f` takes another function `g` as an argument: e.g. `f = fun g => g 0`.
-3. `f` is called by with an argument `g` that does not respect the contract: e.g. `f (fun x => false)`."),
-                     String::from("Either change the contract accordingly, or call `f` with a function that returns a value of the right type."),
-                     end_note,
-                 ])
-            }
-            _ => panic!(),
-        }
-    };
-
-    let (start, end) = ty_path::span(l.path.iter().peekable(), &l.types);
-    let label = Label::new(
-        LabelStyle::Secondary,
-        files.add("", format!("{}", l.types)),
-        start..end,
-    )
-    .with_message(msg);
-    (label, notes)
-}
-
-/// Return a note diagnostic showing where a contract was bound.
-fn blame_label_note(l: &label::Label) -> Diagnostic<FileId> {
-    Diagnostic::note().with_labels(vec![Label::primary(
-        l.span.src_id,
-        l.span.start.to_usize()..l.span.end.to_usize(),
-    )
-    .with_message("bound here")])
-}
-
 impl ToDiagnostic<FileId> for Error {
     fn to_diagnostic(
         &self,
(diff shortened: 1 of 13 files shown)


src/eval/operation.rs
```diff
@@ -615,6 +615,19 @@ impl<R: ImportResolver> VirtualMachine<R> {
...
                         Term::Record(record) => {
+                            // While it's certainly possible to allow mapping over
+                            // a record with a sealed tail, it's not entirely obvious
+                            // how that should behave. It's also not clear that this
+                            // is something users will actually need to do, so we've
+                            // decided to prevent this until we have a clearer idea
+                            // of potential use-cases.
+                            if let Some(record::SealedTail { label, .. }) = record.sealed_tail {
+                                return Err(EvalError::BlameError(
+                                    label,
```
matthew-healy, line 626: Currently the error for this is pretty unclear, as the label doesn't have a tag. I considered doing something like this:

```rust
if let Some(record::SealedTail { mut label, .. }) = record.sealed_tail {
    label.tag = String::from("mapping over a polymorphically sealed record is forbidden");
    // ...
```

but I couldn't find anywhere else where we explicitly set the `tag` of a label without using the `%tag%` primitive. Is there a reason for that, or would setting the `tag` here be acceptable?

src/eval/merge.rs
```diff
@@ -367,6 +367,15 @@ pub fn merge(
         // Merge put together the fields of records, and recursively merge
         // fields that are present in both terms
         (Term::Record(mut r1), Term::Record(mut r2)) => {
+            // While it wouldn't be impossible to merge records with sealed tails,
+            // working out how to do so in a "sane" way that preserves parametricity
+            // is non-trivial. It's also not entirely clear that this is something
+            // users will generally have reason to do, so in the meantime we've
+            // decided to just prevent this entirely
+            if let Some(record::SealedTail { label, .. }) = r1.sealed_tail.or(r2.sealed_tail) {
+                return Err(EvalError::BlameError(label, std::mem::take(call_stack)));
```
matthew-healy, line 376: [This comment](https://github.com/tweag/nickel/pull/949/files#r1033797621) also applies here re: explicitly setting the label's `tag`.

src/eval/operation.rs
yannham: I think one reason is that it would erase potentially user-defined tags, at least in the general case (in that, a user could set an error message, and then use `contract.apply` on a polymorphic record contract).

Note that we have had only one kind of blame error until now, but there is no reason for that (well, beside avoiding code duplication, but we could pull out a blame error message out of the `BlameError` match case). If you need a specific variant because the error reporting is different, it makes sense that PolymorhicTailIllegalAccess or whatever shorter and better name you find has its own variant.

yannham: In general the interface or `Lbl` for users is pretty poor: we could reflect better our error reporting infra and give rather access `notes` to allow multiple lines of information without erasing the previous tag. But I digress

matthew-healy: Okay, thanks for clarifying! I actually started hacking in a new variant while I was spiking but then did indeed get put off by the duplication. I'll take another look now that everything else seems to be working correctly at least. :grin:

src/term.rs
```diff
@@ -1310,6 +1310,14 @@ pub enum UnaryOp {
...
     /// doc: TODO
     PushForce(),
+
+    /// Creates an "empty" record with the sealed tail of its [`Term::Record`]
+    /// argument.
+    ///
+    /// Used in the `$record` contract implementation to ensure that we can
+    /// define a `field_diff` function that preserves the sealed polymorphic
+    /// tail of its argument.
+    EmptyRecordWithTail(),
```
vkleen, line 1320: The primitive is called `%record_empty_with_tail%`, so maybe it would be better to call the enum variant `RecordEmptyWithTail` as well?

src/term.rs
```diff
@@ -1488,6 +1488,15 @@ pub enum NAryOp {
...
     MergeContract(),
+    /// Seals one record into the tail of another. Used to ensure that functions
+    /// using polymorphic record contracts do not violate parametricity.
+    ///
+    /// Takes four arguments:
+    ///   - a [sealing key](Term::SealingKey), which must be provided later to unseal the tail,
+    ///   - a [label](Term::Lbl), which will be used to assign blame correctly tail access is attempted,
+    ///   - a [record](Term::Record), which is the record we wish to seal the tail into,
+    ///   - the [term](Term) that we wish to seal.
+    SealRecordTail(),
```
vkleen, line 1499: The same consideration as for `%record_empty_with_tail%`, maybe it would be better to go with `RecordSealTail`?

src/term.rs
```diff
@@ -1497,6 +1497,14 @@ pub enum NAryOp {
...
     ///   - the [term](Term) that we wish to seal.
     SealRecordTail(),
+    /// Unseals a term from the tail of a record and returns it.
+    ///
+    /// Takes three arguments:
+    ///   - the [sealing key](Term::SealingKey), which was used to seal the tail,
+    ///   - a [label](Term::Label) which will be used to assign blame correctly if
+    ///     something goes wrong while unsealing,
+    ///   - the [record](Term::Record) whose tail we wish to unseal.
+    UnsealRecordTail(),
```
vkleen, line 1507: The same consideration as for `%record_empty_with_tail%`, maybe it would be better to go with `RecordUnsealTail`?

vkleen (approved):
I can't really comment on the correctness of this change, but I can bikeshed `enum` variant names! I've left a few comments on the names of the sealing primitive operations.

src/term.rs
matthew-healy: Yeah I wasn't sure what the convention here was. There's a bit of variance in the existing primitives, e.g. `%record_map%` becomes `RecordMap`, but `%record_insert%` and `%record_remove%` become `DynExtend` and `DynRemove`. I decided to just pick something & let the bikeshedding happen in the PR, so I'm more than happy to rename things to whatever folk think is best.

src/eval/operation.rs
```diff
@@ -1201,6 +1215,25 @@ impl<R: ImportResolver> VirtualMachine<R> {
             }
             UnaryOp::PushDefault() => Ok(PushPriority::Bottom.push_into(t, env, pos)),
             UnaryOp::PushForce() => Ok(PushPriority::Top.push_into(t, env, pos)),
+            UnaryOp::EmptyRecordWithTail() => match_sharedterm! { t,
+                with {
+                    Term::Record(r) => {
+                        let mut empty = RecordData::empty();
+                        empty.sealed_tail = r.sealed_tail;
+                        Ok(Closure {
+                            body: RichTerm::new(Term::Record(empty), pos), // TODO: is pos right here?
```
yannham, line 1224: I think we rather use `pos_op.into_inherited()` or something, meaning that the new record will inherit the position from `%empty_record_with_tail%`. Here it seems it will inherit the position of the argument, but they are surely quite different (one is empty, the other may have fields)?

src/eval/operation.rs
```diff
@@ -1916,7 +1949,7 @@ impl<R: ImportResolver> VirtualMachine<R> {
                                         id,
                                         String::from("(-$)"),
                                         RichTerm::new(
-                                            Term::Record(RecordData::new(fields, record.attrs)),
+                                            Term::Record(RecordData::new(fields, record.attrs, record.sealed_tail)),
```
yannham, line 1952: Can't we do something like `RecordData { fields, ..record }` here? I don't remember if that works for something else than `Default`

src/term.rs
```diff
@@ -468,6 +483,41 @@ pub mod record {
...
+    /// lead to parametricity violations.
+    #[derive(Clone, Debug, PartialEq)]
+    pub struct SealedTail {
+        /// The key with which the tail is sealed.
+        sealing_key: SealingKey,
+        /// The label to which blame will be attributed if code tries to
+        /// interact with the sealed tail in any way.
+        pub label: Label,
+        /// The term which is sealed.
+        term: RichTerm,
```
yannham, line 500: Does it make sense to seal something else than a Record? We are evaluating the argument anyway I think (because this is an argument of a primop). The advantage of restricting what can appear here is that we could make better error messages, like instead of this field doesn't exist, we could say "oh, you're trying to access something in the tail, but the type of your function doesn't allow you to do so, you should do that and that".

src/term.rs
yannham: Hmm the whole things (I mean, the current situation) does lack consistency. You can pick whatever you like for now, and we shall make a pass later to make this more uniform.

src/typecheck/operation.rs
```diff
@@ -378,6 +379,25 @@ pub fn get_nop_type(
...
+        ),
+        // Dyn -> Dyn -> Dyn -> Dyn
+        NAryOp::UnsealRecordTail() => (
+            vec![
+                mk_uniftype::dynamic(),
+                mk_uniftype::dynamic(),
+                mk_uniftype::dynamic(),
+            ],
+            mk_uniftype::dynamic(),
+        ),
```
yannham, line 400: I imagine we could have slightly more precise types here, but it's probably quite useless.

src/error.rs
```diff
@@ -92,6 +94,13 @@ pub enum EvalError {
         String,  /* error message */
         TermPos, /* position of the call to deserialize */
     ),
+    /// A polymorphic record contract was broken somewhere.
+    IllegalPolymorphicTailAccess {
+        /// A short description of why the error was raised
+        reason: String,
```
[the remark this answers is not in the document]
matthew-healy: Yeah, this is a reasonable point. I originally defined an enum like this in the `error` module:

```rust
enum IllegalPolymorphicTailAction {
    Map,
    Merge
}

impl IllegalPolymorphicTailAction {
    fn message(&self) -> &'static str {
        // etc.
    }
}
```

but I felt that was slightly too much boilerplate so I put the `String` in instead. I totally get not wanting to leak formatting info to other modules though, so I'll put the enum back in.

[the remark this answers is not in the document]
matthew-healy: Oh, I just saw that you commented almost exactly this below :grin:

src/term.rs
matthew-healy: Yep, completely agree with this. I'd originally planned to make this change when we actually implement the improved error messages, but that's likely to just cause confusion if it's not me that ends up doing it. Will fix this before I merge.

src/eval/operation.rs
matthew-healy: We can! I have a small personal preference for using `new` because it generally allows us to be more explicit about which parts of a struct should be publicly visible (& would, e.g., reduce the diff if we ever added private fields), but given that isn't applied consistently across the codebase I'm happy to use the "splat" notation here.

src/typecheck/operation.rs
matthew-healy: Yeah, the conclusion I came to is that types for these operations aren't worth enforcing as any error would have to be (1) an error within the stdlib, which should be caught by the test suite or (2) a result of users misusing primitive ops, in which case runtime errors are probably sufficient.

That said, out of curiosity, what types do you think would be appropriate here? I tried a few ideas but wasn't happy with any of them.

Taking `%record_unseal_tail%` as an example, my thinking process was: 

- the sealing key and label are typed as `Dyn` everywhere, so we don't have much choice there,
- we can't use row types since we'd have to somehow represent the realization of the row variable into its constituent fields & types,
- we could use dictionary types, but since these operations are designed to work on row types (1) that felt a little misleading and (2) the only valid value type would be `Dyn` which doesn't provide much value.

I may well be missing something though - happy to hear any ideas!

src/term.rs
matthew-healy: Actually, after investigating this, I don't think we should hold on to the `RecordData` here directly. The key issue is that in `SealRecordTail` we need to closurize the entire record's `RichTerm`. We could instead closurize the values of the record, but unless I'm mistaken we'd have to pay the cost of cloning the tail record's env for each entry.

As an alternative, we can take the record argument we receive in `SealRecordTail` and just build a `HashSet` of its keys which we can then check against whenever we would normally raise a missing field error. So, unless I've misunderstood something, I'll leave this as-is & just add a note to #952 that this might be the easiest way to implement this.

src/term.rs
matthew-healy: Okay, in the meantime I'll go with Viktor's suggestions. :+1:

src/term.rs
yannham: Ah, it's true that the closurization is slightly more involved.

> We could instead closurize the values of the record, but unless I'm mistaken we'd have to pay the cost of cloning the tail record's env for each entry.

That being said:

1. Cloning environment is very cheap. Environments are `Rc`-ed, so it's all in all an `Rc::clone`.
2. Record fields must already be closurized. If you re-closurize something already closurized, there is an optimization which avoids adding a useless indirection. So mostly, closurizing would take the Closure out of the environment given as an argument, and put it in the outer environment. We may have to clone an environment to immediately drop it, which is not optimal, but should be pretty insignificant overall.

But the picture is less clear-cut. I think both approaches are valid. Yours may actually be easier. 

Though I think the seal record tail primitive should still fail if it doesn't see a record, even if you keep the current representation. Currently you're taking the tail out and merging it inside the implementation of the record tail contract. If a user (or our faulty implementation) injects something else in the tail, this going to make for bad error messages.

If we do this, you keep your easy closurizing, but we just ensure through sealing that we only actually put a closurized record inside. Doing so, we can very much get back this record from the error reporting code in fact: just get the thunk from the environment and inspect the closure, which must be an evaluated record.

src/typecheck/operation.rs
yannham: I was thinking of `{_ : Dyn}`, mostly.

> we could use dictionary types, but since these operations are designed to work on row types (1) that felt a little misleading and (2) the only valid rhs type would be Dyn which doesn't seem to provide much value.

Mh I wouldn't say those operations "are designed to work on row types". Those primops act on data, and they are pretty dynamic operations moving studd around in the representation of the record. The question is what shape of data they expect and return: we can say at least that most of the arguments and the return type are records, so `{_ : Dyn}`.

I agree it's not very useful.

src/term.rs
matthew-healy: > Though I think the seal record tail primitive should still fail if it doesn't see a record, even if you keep the current representation.

Yep, totally agree. Will fix this now. 

Good point about the clones being cheaper than I was thinking. I'll defer the decision until we actually implement the field access blame error, and then we can see what makes most sense at the time. (Though tbh implementing that error feels pretty trivial whatever we do, so I might just do it quickly this week if I find some time. :grin:)

src/typecheck/operation.rs
matthew-healy: I think you've actually convinced me that `{ _ : Dyn }` makes slightly more sense than just `Dyn`, so I'll update it before I merge. :sweat_smile:

PR #953, opened 2022-11-28
[RFC005 Implementation] Restrict metadata to record field

yannham:
As part of the RFC005 implementation, this PR starts by restricting syntactically metadata to only appear in a record field definition. Fix the tests, benches and examples in consequence.

This PR doesn't modify further the semantics or the internal representation of metadata, which will be done in subsequent PR.

Some let-binding were annotated with a `doc`, which looks like a legitimate usage of a metadata somewhere else than on a record field. Such documentation is not accessible from the outside, but may still be leveraged e.g. by the LSP. For now, those have been commented out, but may reappear in the final version, if we loosen a bit the original proposal.

### DIFF
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -113,7 +113,7 @@ AnnotAtom<TypeRule>: MetaAnnot = {
     },
 };
 
-// A single field metadata annotation, without the pseudo-metata (such as push
+// A single field metadata annotation, without the pseudo-metadata (such as push
 // priorities).
 //
 // The rule to use for type annotations is given as a parameter (cf AnnotAtom
@@ -166,7 +166,7 @@ Annot<TypeRule>: MetaAnnot = AnnotSeries<AnnotAtom<TypeRule>>;
 // field annotation excludes pseudo metadata like push priorities operator.
 SimpleFieldAnnot<TypeRule>: MetaAnnot = AnnotSeries<SimpleFieldAnnotAtom<TypeRule>>;
 
-// An field annotation, with possibly many annotations chained.
+// A field annotation, with possibly many annotations chained.
 FieldAnnot<TypeRule>: FieldAnnot = AnnotSeries<FieldAnnotAtom<TypeRule>>;
 
 // A general term. Wrap the root of the grammar as a `RichTerm`.
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -92,7 +92,7 @@ AsUniTerm<Rule>: UniTerm = <l: @L> <ut: Rule> <r: @R> =>
 AnnotSeries<AnnotAtom>: AnnotAtom = <AnnotAtom+> =>
     <>.into_iter().fold(Default::default(), Annot::combine);
 
-// A single type or meta-data annotation. The `Types` rule forbids the use of
+// A single type or contract annotation. The `Types` rule forbids the use of
 // constructs that can themselves have annotation on the right, such as a `let`.
 // Otherwise, `foo | let x = 1 in x : Num` is ambiguous (the annotation could be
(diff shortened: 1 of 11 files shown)


tests/query.rs
```diff
@@ -3,30 +3,16 @@ use nickel_lang::term::{MetaValue, SharedTerm, Term};
...
-    if let Term::MetaValue(meta) = result {
-        assert_eq!(meta.doc, Some(String::from("Test basic")));
-        assert_eq!(meta.value.unwrap().term, SharedTerm::new(Term::Num(2.0)));
-    } else {
-        panic!();
-    }
-}
-
-#[test]
-pub fn test_query_metadata_from_func() {
```
yannham, line 19: The diff looks strange, but I actually kept the metdata_basic test (just put everything inside a record) and removed the from_func one. The latter tests the propagation of metadata through function application, but RFC005 decides that this won't happen anymore, hence the test isn't relevant

tests/pass/metavalues.ncl
```diff
@@ -1,12 +1,7 @@
 let {check, Assert, ..} = import "testlib.ncl" in
 
 [
-  (10 | default | Num) == 10,
-
-  # Check the correct handling of the update of thunks containing enriched values (see issue
-  # https://github.com/tweag/nickel/issues/123)
-  let x = {a = (fun x => (1 | default)) 1} in
-    builtin.seq (x.a) ((x & {a=2}).a) == 2,
```
yannham, line 9: Same as in the other test: the removed test doesn't make sense for RFC005, which precludes the propagation of metadata.

aspiwack:
> For now, those have been commented out, but may reappear in the final version, if we loosen a bit the original proposal.

If we're confident, you can also parse these `doc` annotations on let and discard them immediately. It would avoid looking for many commended out `doc` things in the future.

vkleen (approved):

yannham:
> If we're confident, you can also parse these doc annotations on let and discard them immediately. It would avoid looking for many commended out doc things in the future.

Yes, that's probably better.

src/parser/grammar.lalrpop
```diff
@@ -102,11 +102,24 @@ AnnotSeries<AnnotAtom>: AnnotAtom = <AnnotAtom+> =>
...
         contracts: vec![Contract {types: ty.clone(), label: mk_label(ty, src_id, l, r)}],
         ..Default::default()
     },
+    ":" <l: @L> <ty: TypeRule> <r: @R> => MetaValue {
+        types: Some(Contract {types: ty.clone(), label: mk_label(ty, src_id, l, r)}),
+        ..Default::default()
+    },
+};
+
+// A single field metadata annotation, without the pseudo-metata (such as push
```
matthew-healy, line 116:
```suggestion
// A single field metadata annotation, without the pseudo-metadata (such as push
```

src/parser/grammar.lalrpop
```diff
@@ -149,8 +162,12 @@ ExtdAnnotAtom<TypeRule>: ExtdAnnot = {
 // An annotation, with possibly many annotations chained.
 Annot<TypeRule>: MetaAnnot = AnnotSeries<AnnotAtom<TypeRule>>;
 
-// An extended annotation, with possibly many annotations chained.
-ExtdAnnot<TypeRule>: ExtdAnnot = AnnotSeries<ExtdAnnotAtom<TypeRule>>;
+// A simple field annotation, with possibly many annotations chained. A simple
+// field annotation excludes pseudo metadata like push priorities operator.
+SimpleFieldAnnot<TypeRule>: MetaAnnot = AnnotSeries<SimpleFieldAnnotAtom<TypeRule>>;
+
+// An field annotation, with possibly many annotations chained.
```
matthew-healy, line 169:
```suggestion
// A field annotation, with possibly many annotations chained.
```

matthew-healy (approved):

dpl0a (approved):

PR #954, opened 2022-11-28
Restructure integration tests into a single crate

matthew-healy:
This PR replaces our multi-crate integration test setup with a single crate that contains all tests, inspired by [this blog](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html).

This has dual benefits:
1. it should reduce the amount of time tests take to compile, as we're only producing a single binary,
2. it should reduce the amount of time tests take to run, as cargo schedules individual test binaries to run sequentially whereas tests within a binary can run in parallel.

I don't have hugely strong feelings about this, and totally understand if people want to keep the existing structure, but at least on my machine this seemed to reduce the time of `cargo test` by about 50%.


yannham (commented):
I don't have any strong opinion about this. This doesn't seem to hurt. Though I would update the CONTRIBUTING.MD document, and maybe add a README.md at the root of /tests, so that clueless people adding tests know what to do (and what not to do)

matthew-healy:
Ooh, the ubuntu CI ran ~10-15 minutes faster than it has on other recent PRs here, and macOS was ~5-10 mins faster. :eyes:

yannham (approved):

PR #959, opened 2022-11-30
Fix LSP server crash when `".` is inserted into the editor

### LINKED ISSUE #958: LSP server crashes on `"` followed by a `.` for completion
**Describe the bug**
When  `".` is typed in the editor, the LSP server crashes. 

**To Reproduce**
Open an editor with the Nickel LSP server attached and running and type `".`.

**Expected behavior**
Nothing, the LSP should not crash.

**Additional context**
The crash error message: 
```
thread 'main' panicked at 'begin <= end (1 <= 0) when slicing `"`', lsp/nls/src/requests/completion.rs:267:27
[Info  - 17:39:21] Connection to server got closed. Server will restart.
[Error - 17:39:21] Request textDocument/completion failed.
```
See [here.](https://github.com/tweag/nickel/blob/master/lsp/nls/src/requests/completion.rs#L169)

ebresafegaga:
Closes #958

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -417,6 +417,7 @@ mod tests {
                     me.ca.cb"##,
                 vec!["me", "ca", "cb"],
             ),
+            ("Single quote", "\"", vec!["\""]),
         ];
         for (case_name, input, expected) in tests {
             let actual = get_identifier_path(input);
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -166,7 +166,8 @@ fn get_identifier_path(text: &str) -> Option<Vec<String>> {
     let path = result.iter().rev().next().cloned()?;
     // Remove quotes from a record name
     fn remove_quotes(name: &str) -> String {
-        if let (Some('"'), Some('"')) = (name.chars().next(), name.chars().last()) {
+        let mut chars = name.chars();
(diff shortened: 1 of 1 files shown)


matthew-healy (approved):
:+1:

matthew-healy (approved):
Okay, I had got a little confused by the `".` in the title. The problem was really if the name was just a single `"` as then the first and last char of the string would be the same, which is why using a single iterator fixes it. Approving with confidence this time. :grin:

lsp/nls/src/requests/completion.rs
```diff
@@ -166,7 +166,8 @@ fn get_identifier_path(text: &str) -> Option<Vec<String>> {
     let path = result.iter().rev().next().cloned()?;
     // Remove quotes from a record name
     fn remove_quotes(name: &str) -> String {
-        if let (Some('"'), Some('"')) = (name.chars().next(), name.chars().last()) {
+        let mut chars = name.chars();
+        if let (Some('"'), Some('"')) = (chars.next(), chars.last()) {
```
sir4ur0n, line 170: Can we add a test as evidence of this fix? :pray: I.e. a test that fails before the fix, and succeeds after the fix

sir4ur0n (changes requested):

lsp/nls/src/requests/completion.rs
ebresafegaga: Done, thanks!

sir4ur0n (approved):

dpl0a (approved):

PR #963, opened 2022-12-02
Migrate Nix build from import-cargo to Crane

sir4ur0n:
Migrate Nickel Nix build from [import-cargo](https://github.com/edolstra/import-cargo) to [crane](https://github.com/ipetkov/crane).

As the main commit description explains, the Nickel WASM build **is not migrated to Crane in this PR**.
This is to keep the PR "small" :asterisk: as changing the Nickel (non-WASM) build is already not trivial.

Below is a list of minor things I had to fix so that the build succeeds, as well as some side refactorings/fixes:
* some documentation about how to build with Nix
* remove Clippy from pre-commit hook: this is as discussed over Slack, because Clippy requires compiling, so the pre-commit can be slow, and the benefit is low as `rust-analyzer` LSP already provides Clippy integration out of the box
* the current CI only ran tests of the main package, it now compiles and runs tests of all packages (workspace)
* `nightly` was removed from the Rust channels
* `nightly` and `beta` were removed from the built and checked packages in CI
* some Rust test code did not compile, I fixed those
* `utilities/Cargo.lock` was removed as it was a relic of the past (as we use Cargo workspace, we have a single `Cargo.lock` file at the project root)

:asterisk: "small" because I fixed and refactored a bit some code in the same PR, sorry about those minor leaks

:bulb: I suggest reviewing commit by commit


flake.nix
```diff
@@ -1,13 +1,23 @@
...
+      url = "github:cachix/pre-commit-hooks.nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.flake-utils.follows = "flake-utils";
+    };
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.flake-utils.follows = "flake-utils";
+    };
+    import-cargo.url = "github:edolstra/import-cargo";
```
sir4ur0n, line 15: I can revert this refactor if you all disagree.

For some reason Flake inputs are often flattened when I think the nesting helps reading what belongs to what, and is clearer regarding organization of code

flake.nix
```diff
@@ -101,179 +111,135 @@
...
+          # Args passed to all `cargo` invocations by Crane.
+          cargoExtraArgs = "--frozen --offline --workspace";
+
+        in
+        rec {
+          # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
+          cargoArtifacts = craneLib.buildDepsOnly {
+            inherit
+              src
+              cargoExtraArgs;
```
sir4ur0n, line 198: I like the declarative aspect of Crane: I find it easier to understand what influences what in the chain of dependencies.

flake.nix
```diff
@@ -370,35 +336,32 @@
     in
     rec {
       packages = {
-        default = packages.build;
-        build = buildNickel { };
-        buildWasm = buildNickelWasm { };
-        dockerImage = buildDocker packages.build; # TODO: docker image should be a passthru
+        nickel = (mkCraneArtifacts { }).nickel;
+        default = packages.nickel;
+        nickelWasm = buildNickelWasm { };
```
sir4ur0n, line 341: I took the liberty to rename the Nix packages into what I think are more sensible names:
* `build` -> `nickel`
* `buildWasm` -> `nickelWasm`

yannham (approved):
In addition to the other benefits, I find the flake structure clearer overall :+1: 

I would maybe add somewhere, either at the top of the flake, on the `inport-from-cargo` inputs, and/or on  `nickelWasm`, a comment that explains that we're currently using crane for the normal build and import from cargo for the WASM REPL, although that situation might be temporary.

You've changed more than just import-from-cargo vs crane (e.g. by removing some channels), but still GH runners should be fairly parallel, so let's compare with a random recent PR (after tests runtime have been improved) to have an idea of the difference:

Random PR:

Capture d’écran du 2022-12-05 16-15-36

This PR:

Capture d’écran du 2022-12-05 16-06-28

Not bad :sunglasses:

matthew-healy (approved):

matthew-healy:
To echo what Yann said: the CI runtime improvement with this PR is excellent. Nice one! :fire:

sir4ur0n:
> I would maybe add somewhere, either at the top of the flake, on the inport-from-cargo inputs, and/or on nickelWasm, a comment that explains that we're currently using crane for the normal build and import from cargo for the WASM REPL, although that situation might be temporary.

@yannham agreed:
* I created issue https://github.com/tweag/nickel/issues/967 to keep track of this
* I added a comment linking to this issue in 982623f6e92c0f78ec8f892036623bebf1396825

PR #939, opened 2022-11-23
Display meta information when providing completion

ebresafegaga:
Closes #920 

Depends on #918

PR #966, opened 2022-12-05
Display meta information when providing completion - Rebase by Julien

sir4ur0n:
:warning: This is (supposed to be) exactly the same PR as https://github.com/tweag/nickel/pull/939 but rebased on `master`.

I did my best to fix the conflicts and squash all commits together, but I may or may not have respected the spirit of the original PR.

Please review with caution :pray:

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -3,7 +3,7 @@ use codespan_lsp::position_to_byte_index;
 use lazy_static::lazy_static;
 use log::debug;
 use lsp_server::{RequestId, Response, ResponseError};
-use lsp_types::{CompletionItem, CompletionParams};
+use lsp_types::{CompletionItem, CompletionItemKind, CompletionParams};
 use nickel_lang::{
     identifier::Ident,
     term::{MetaValue, RichTerm, Term},
@@ -29,18 +29,129 @@ use crate::{
 // We follow the path by traversing a term, type or contract which represents a record
 // and stop when there is nothing else on the path
 
+// Type or contract
+pub enum TorC {
+    T(Types),
+    C(RichTerm),
+}
+
+// #[derive(Eq, PartialEq, Debug, Ord, PartialOrd, Default)]
+struct IdentWithMeta {
+    ident: Ident,
+    info: TorC,
+    item: Option<LinearizationItem<Types>>,
+}
+
+impl From<Ident> for IdentWithMeta {
+    fn from(ident: Ident) -> Self {
+        IdentWithMeta {
+            ident,
+            info: TorC::T(Types(TypeF::Dyn)),
+            item: None,
+        }
+    }
+}
+
+impl From<&str> for IdentWithMeta {
+    fn from(ident: &str) -> Self {
+        IdentWithMeta {
+            ident: Ident::from(ident),
+            info: TorC::T(Types(TypeF::Dyn)),
+            item: None,
+        }
+    }
+}
+
+impl IdentWithMeta {
+    fn compute_detail(&self) -> String {
+        let info = || match &self.info {
+            TorC::T(ty) => ty.to_string(),
+            TorC::C(term) => term.to_string(),
+        };
+        let Some(item) = &self.item else {
+            return info()
+        };
+        item.meta
+            .as_ref()
+            .map(|meta| {
+                meta.types
+                    .as_ref()
+                    .map(|ty| ty.types.to_string())
+                    .or_else(|| {
+                        let result = meta
+                            .contracts
+                            .iter()
+                            .map(|contract| format!("{}", contract.label.types,))
+                            .collect::<Vec<_>>()
+                            .join(",");
+                        if result.is_empty() {
+                            None
+                        } else {
+                            Some(result)
+                        }
+                    })
+                    .unwrap_or_else(info)
+            })
+            .unwrap_or_else(info)
+    }
+
+    fn compute_completion_item_kind(&self) -> CompletionItemKind {
+        match &self.info {
+            TorC::T(ty) if ty.is_function_type() => CompletionItemKind::Function,
+            TorC::T(_) => CompletionItemKind::Property,
+            TorC::C(term) => match term.as_ref() {
+                Term::Fun(..) => CompletionItemKind::Function,
+                _ => CompletionItemKind::Property,
+            },
+        }
+    }
+
+    fn to_lsp_completion_item(&self) -> CompletionItem {
+        /// Attach quotes to a non-ASCII string
+        fn adjust_name(name: &str) -> String {
+            if name.is_ascii() {
+                String::from(name)
+            } else {
+                format!("\"{}\"", name)
+            }
+        }
+        CompletionItem {
+            label: adjust_name(self.ident.label()),
+            detail: Some(self.compute_detail()),
+            kind: Some(self.compute_completion_item_kind()),
+            ..Default::default()
+        }
+    }
+}
+
 /// Find the record field associated with a particular ID in the linearization
 /// using lexical scoping rules.
 fn find_fields_from_term_kind(
     linearization: &Completed,
     id: ItemId,
     path: &mut Vec<Ident>,
-) -> Option<Vec<Ident>> {
+) -> Option<Vec<IdentWithMeta>> {
     let item = linearization.get_item(id)?;
     match item.kind {
         TermKind::Record(ref fields) => {
             if path.is_empty() {
-                Some(fields.keys().cloned().collect())
+                Some(
+                    fields
+                        .keys()
+                        .cloned()
+                        .map(|ident| {
+                            // These unwraps are safe
+                            let id = fields.get(&ident).unwrap();
+                            let item = linearization.get_item(*id).unwrap();
+                            let (ty, _) = linearization.resolve_item_type_meta(item);
+                            IdentWithMeta {
+                                ident,
+                                info: TorC::T(ty),
+                                item: Some(item.clone()),
+                            }
+                        })
+                        .collect(),
+                )
             } else {
                 let name = path.pop()?;
                 let new_id = fields.get(&name)?;
@@ -65,7 +176,7 @@ fn find_fields_from_contract(
     linearization: &Completed,
     id: ItemId,
     path: &mut Vec<Ident>,
-) -> Option<Vec<Ident>> {
+) -> Option<Vec<IdentWithMeta>> {
     let item = linearization.get_item(id)?;
     match &item.meta {
         Some(meta_value) => Some(find_fields_from_meta_value(meta_value, path)),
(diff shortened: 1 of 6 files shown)


matthew-healy:
One thing to note here is that because the entire PR history has been squashed down, and this PR only contains a single commit, the commit message of this will appear in the project history as "Rename `from` to `new`" which isn't accurate. If this branch ends up being merged then I'd recommend renaming the commit to better reflect its overall contents.

sir4ur0n:
Yeah I squashed and kept commit messages that seemed to be useful (you can see them in the commit description), but otherwise I did not rework it

yannham:
Thanks for doing that, @sir4ur0n. Unfortunately I have non trivial questions which need to be addressed, so let's wait for @ebresafegaga's return and an up-to-date rebase.

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,129 @@ use crate::{
...
+        }
+    }
+}
+
+impl IdentWithMeta {
+    fn compute_detail(&self) -> String {
+        let info = || match &self.info {
+            TorC::T(ty) => ty.to_string(),
+            TorC::C(term) => term.to_string(),
+        };
```
yannham, line 70: Maybe `TOrC` should implement `to_string()`? But once again, if you just use `Types`, this part isn't even needed.

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,129 @@ use crate::{
...
+        let Some(item) = &self.item else {
+            return info()
+        };
```
yannham, line 73:
```suggestion
        self.item.unwrap_or_else(info)
```

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,129 @@ use crate::{
 // We follow the path by traversing a term, type or contract which represents a record
 // and stop when there is nothing else on the path
 
+// Type or contract
+pub enum TorC {
+    T(Types),
+    C(RichTerm),
+}
+
+// #[derive(Eq, PartialEq, Debug, Ord, PartialOrd, Default)]
```
yannham, line 38:
```suggestion
```

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,129 @@ use crate::{
...
+            .unwrap_or_else(info)
+    }
+
+    fn compute_completion_item_kind(&self) -> CompletionItemKind {
+        match &self.info {
+            TorC::T(ty) if ty.is_function_type() => CompletionItemKind::Function,
+            TorC::T(_) => CompletionItemKind::Property,
+            TorC::C(term) => match term.as_ref() {
+                Term::Fun(..) => CompletionItemKind::Function,
+                _ => CompletionItemKind::Property,
```
yannham, line 104: I'm not sure about this one. Does it mean that if a term has a contract attached, and the contract is a function, then the term is considered to be documented as a function?

A contract is actually most of the time a function, but this doesn't tell anything about the value to which the contract is applied (and most of the time this function is hidden behind a variable, so this match case will never be taken anyway). I think there's not meaningful, simple, computable property able to check "is this contract checking for a function type". Function contracts are mostly always written as `TypesF::Arrow(.., ..)`, which is taken care of by the first case.

Here, it seems we should always return `CompletionItemKind::Property`.

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,129 @@ use crate::{
...
+    T(Types),
+    C(RichTerm),
+}
+
+// #[derive(Eq, PartialEq, Debug, Ord, PartialOrd, Default)]
+struct IdentWithMeta {
+    ident: Ident,
+    info: TorC,
+    item: Option<LinearizationItem<Types>>,
+}
```
yannham, line 43: I think we might also want to print the documentation, the merge priorities, etc. In fact, everything that we print on `hover`, which are mostly what's inside a `Metadata`. Have you considered just storing that here, instead of just `TorC`?

```suggestion
struct IdentWithMeta {
    ident: Ident,
    meta: MetaValue,
    item: Option<LinearizationItem<Types>>,
}
```

yannham (commented):
Much wanted feature :slightly_smiling_face: !

The definition of `TorC` and the like feels a bit ad-hoc and limited, though. We are able to find (and represent in Rust) the metadata and type associated to any item already. We're using that to display information on hover, including documentation and merge priority. I think we should do exactly the same thing here: just carry `MetaValue` around and show the full information instead of just the type/contract.

lsp/nls/src/requests/completion.rs
ebresafegaga: We already have a `MetaValue` present in `item`; the `info` field is used to get extra info when that is not available. Also, we can just use a `Types` to replace `TorC`, like you suggested.

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,109 @@ use crate::{
...
+            } else {
+                format!("\"{}\"", name)
+            }
+        }
+        let doc = || {
+            let item = self.item.as_ref()?;
+            let meta = item.meta.as_ref()?;
+            let doc = meta.doc.as_ref()?;
+            let doc =
+                Documentation::String(format!("{}\nMerge Priority {}", doc.clone(), meta.priority));
```
yannham, line 94: I think you have a way to pretty-print metadata somewhere in the code, for the REPL and `nickel query` (including a version without markdown support). I'm not saying this should be done in this PR, but maybe we could reuse the same code, in order to have a consistent way of showing metadata about values.

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +29,109 @@ use crate::{
...
     match item.kind {
         TermKind::Record(ref fields) => {
             if path.is_empty() {
-                Some(fields.keys().cloned().collect())
+                Some(
+                    fields
+                        .keys()
+                        .cloned()
+                        .map(|ident| {
+                            // These unwraps are safe
```
yannham, line 123: The important information to add here is _why_ they are safe :slightly_smiling_face: . Because we ensured somewhere else in the code that all those indices are indeed valid, I guess. But if I refactor the code I have no idea what I have to do to preserve this property.

yannham (commented):
Beside my comments, I think the PR is good to go. However, I couldn't get it to work under neovim + coc.nvim. When I show the log, I do see the completion command correctly, and the field `detail` is indeed set with the type, but it doesn't show in my menu.

Capture d’écran du 2022-12-15 11-21-09

For comparison, here is what happens with rust-analyzer:

Capture d’écran du 2022-12-15 11-29-34

I wonder if `coc-rust-analyzer` is doing something special here? Also, even in the log, I still don't see the documentation. Note that we can set `documentationFormat` to `markdown` in the completion capabilities and the editors would probably renders nicely our documentation.

Does it work correctly on VSCode?

ebresafegaga:
It works correctly on VS Code: 
image

The only difference is that it only shows the type for the item currently selected, unlike `rust-analyzer`: 
image

yannham:
Interesting. I tried on the previous version of `lsp/metainfo`, before your rebase, and it works there. The difference in the logs seem to be that `kind` is set to `3` everywhere. I agree it shouldn't impact the completion, so maybe it's a coc bug.

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes, I agree. I've already done a little bit in this PR: when we display metadata on LSP hover, there is a  separate pretty-printing for that, so I moved some of that to be methods of `MetaValue`

yannham:
Ok, after diffing the logs between the two versions, it seems that the one that works *DOESN'T HAVE* the `labelDetailsSupport` enabled. By default completion seems to take `detail` into account in any case, and I don't think the `labelDetailsSupport` capability needs to be activated for that. This option says that we may fill the `labelDetails` field, with is a more structured way of providing details. I suspect that `coc.nvim`, when this option is enabled, expect the detail to be only precised using this field, and ignore the `label` field altogether.

I suggest to not enable this option when sending server capabilities.

yannham:
@ebresafegaga Could we also add documentation in the `description` field of the item completion?

ebresafegaga:
There is a `documentation` field, which already populated with the doc gotten from the item's `MetaValue`. 

And the documentation seems to work fine in VS Code: 
image

Does this work for you?

yannham:
It doesn't, but I really can't figure out why.

lsp/nls/src/server.rs
```diff
@@ -59,6 +59,7 @@ impl Server {
             references_provider: Some(OneOf::Left(true)),
             completion_provider: Some(CompletionOptions {
                 trigger_characters: Some(vec![String::from(DOT_COMPL_TRIGGER)]),
+                resolve_provider: Some(true),
```
yannham, line 62: Ok, the culprit was this line. We don't really provide a resolver I think, so if that's fine by you, let's remove it.
```suggestion
```

lsp/nls/src/requests/completion.rs
```diff
@@ -29,18 +31,117 @@ use crate::{
...
+                            // we're currently analyzing. We're sure that the linearization
+                            // phase doesn't produce wrong or invalid ids.
+                            let item = linearization.get_item(*id).unwrap();
+                            let (ty, _) = linearization.resolve_item_type_meta(item);
+                            IdentWithType {
+                                ident,
+                                ty,
+                                item: Some(item.clone()),
+                            }
+                        })
```
yannham, line 142: I think you could just iterate over pairs of `(key, value)` using `keys.iter()` instead of having to use `field.get(&ident).unwrap()`

lsp/nls/src/requests/completion.rs
```diff
@@ -115,20 +219,33 @@ fn find_fields_from_type(rrows: &RecordRows, path: &mut Vec<Ident>) -> Vec<Ident
...
+fn find_fields_from_term(term: &RichTerm, path: &mut Vec<Ident>) -> Option<Vec<IdentWithType>> {
     let current = path.pop();
     match (term.as_ref(), current) {
-        (Term::Record(data) | Term::RecRecord(data, ..), None) => {
-            Some(data.fields.keys().cloned().collect())
-        }
+        (Term::Record(data) | Term::RecRecord(data, ..), None) => Some(
+            data.fields
+                .keys()
+                .cloned()
```
yannham, line 241: Do we need `cloned()` here? `Ident` is `Copy` I think

lsp/nls/src/requests/completion.rs
ebresafegaga: Yes, thanks!

aspiwack:
For the record, I think that at this point, it is better to merge this and fix any remaining issues as separate PRs (probably it's been the case for a while).

yannham (approved):

PR #969, opened 2022-12-05
[fix] Re-enable integration tests

matthew-healy:
I made a mistake when implementing #954 and gave the `integration` crate a `lib.rs` instead of  a `main.rs` which means none of the tests were actually running.

This PR fixes that, and then fixes the failures in the test case. Almost all of them were due to failures when loading Nickel files, due to the paths now being wrong. One additional failure was due to a test using an illegal metadata annotation, which I fixed by inserting the bindings into a parent record. 

Apologies for this. :pray:

### DIFF
--- a/tests/integration/free_vars.rs
+++ b/tests/integration/free_vars.rs
@@ -25,7 +25,7 @@ fn stat_free_vars_incl(
         .all(|(id, set)| free_vars_eq(set, expected.remove(id.as_ref()).unwrap()))
 }
 
-fn dyn_free_vars_incl(dyn_fields: &Vec<FieldDeps>, mut expected: Vec<Vec<&str>>) -> bool {
+fn dyn_free_vars_incl(dyn_fields: &[FieldDeps], mut expected: Vec<Vec<&str>>) -> bool {
     dyn_fields
         .iter()
         .enumerate()
--- a/tests/integration/free_vars.rs
+++ b/tests/integration/free_vars.rs
@@ -1,7 +1,9 @@
+use nickel_lang::term::FieldDeps;
 use nickel_lang::{identifier::Ident, term::Term, transform::free_vars};
 
 use std::collections::{HashMap, HashSet};
 use std::iter::IntoIterator;
+use std::rc::Rc;
 
 use nickel_lang_utilities::parse;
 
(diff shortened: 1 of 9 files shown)


matthew-healy:
I'm going to merge this as soon as the tests pass even without review, because I think it's more important that these run again. If there's any feedback after that I'll submit follow-up PRs.

src/eval/tests.rs
```diff
@@ -22,7 +22,6 @@ fn parse(s: &str) -> Option<RichTerm> {
     grammar::TermParser::new()
         .parse_term(id, lexer::Lexer::new(s))
         .map(RichTerm::without_pos)
-        .map_err(|err| println!("{:?}", err))
```
yannham, line 25: Is there a reason to remove that? Printing only happen when the test fail, IIRC. Those print statements are probably on purpose, to see the error when the test fail

src/types.rs
```diff
@@ -1021,7 +1021,6 @@ mod test {
 
         // Wrap the type in a contract to have it accepted by the parser.
         let wrapper = format!("null | {}", s);
-        println!("{}", wrapper);
```
yannham, line 1024: Might be the same thing here.

tests/integration/free_vars.rs
```diff
@@ -50,10 +52,6 @@ fn check_stat_vars(expr: &str, expected: HashMap<&str, Vec<&str>>) -> bool {
     match rt.term.into_owned() {
         Term::RecRecord(record, _, deps) => {
             let deps = deps.unwrap();
-            println!(
-                "-- comparing {:#?} **AND* {:#?}",
-                deps.stat_fields, expected
-            );
```
yannham, line 56: same thing here

tests/integration/pretty.rs
```diff
@@ -13,14 +13,8 @@ fn diff(s1: &str, s2: &str) {
 
     for change in diff.iter_all_changes() {
         match change.tag() {
-            ChangeTag::Delete => {
-                nb_diff += 1;
-                println!("- {}", change);
-            }
-            ChangeTag::Insert => {
-                nb_diff += 1;
-                println!("+ {}", change);
```
yannham, line 22: Those ones are probably leftovers, though.

yannham (approved):
LGTM, but some println calls may actually not be leftovers, but were made on purpose to help diagnose failing tests.

src/eval/tests.rs
matthew-healy: Ah okay! I ran the tests with `--nocapture` at one point to see some debug info and was surprised by how noisy it was so I just did a sweep to get rid of all the `println`s :sweat_smile: 

I didn't actually know that they get printed on test failure! That's very cool. I'll put them back in in that case.

Issue #477, opened 2021-11-30
Could the last argument of `switch` be curried?


silverraven691:
**Is your feature request related to a problem? Please describe.**

Because of how large the expression can get, I end up moving `switch` statements to their own functions, as such:
```
let see = fun color => switch {
  Red => "I see Red",
  Green => "I see Green",
  Blue => "I see Blue",
} color
in
see `Red
```
As you can see, the dance to pass the `color` parameter feels very unnecessary, but the parser fails without it.

**Describe the solution you'd like**
It would be nice to be able to write `switch { A => .. , B => .. }` and get a function of type `<A , B> -> Dyn`
The default block should be able to fit as well: `switch { A => .. , _ => .. }` would be of type `forall e. <A | e> -> Dyn`

yannham:
The fact that switch takes the argument last is a bit unfortunate. I don't know yet if we will keep the syntax this way. However, if we do, I guess the currification doesn't hurt.

If we reverse it to something like `case foo of {...}` or `match foo with {...}`, then there are other possible solutions to this problem, such as Haskell's [lambda case](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html) or OCaml syntax `let f = function Foo -> 1 | Bar -> 2`. The idea is fundamentally to have a special syntax that squashes a function definition of one argument witch pattern matching.

matthew-healy:
As of 0.3.0, this is how `match` works, so I think we can close this issue.

PR #970, opened 2022-12-05
Change switch to match

### LINKED ISSUE #817: The Syntax Bikeshedding Dojo, round 8: switch
# Switch

We have an unloved construct in Nickel, the switch. It's currently not as useful as it could be, since it's just a C-like switch that works only on enum tags. However, it could become a nice tool in the future, if we allow to match against patterns, literals, etc.

## Issues

However, there are a few quirks:

1. The current syntax is odd: we first write the cases, and only then the expression that is matched. This structure makes the flow hard to follow, in my humble opinion:

      ```nickel
      switch {
        `foo => 1,
        `bar => 2,
         _ => 3
      } some_exp
      ```
2. `switch` is reminiscent of C or JavaScript (to be fair, it is currently close to those operators in semantics), but isn't common in functional languages with pattern matching. If we are to implement pattern matching, we should probably use a different keyword.

## Proposals

Note that because of the ML-y syntax of Nickel, we can't just swap the order of the cases and of the argument. Otherwise, we can't easily disambiguate between a record that is an argument to a function application and the cases of a switch (well we can right now, but probably won't be able in the future using an LR(1) parser without hacky backtracking, eg if we allow to match on string literal or to bind simple variables like "x") :

```nickel
# should we parse this as `switch (f {...}) { cases })` or `(switch f { cases...}) {...}`? 
switch f {...} {...}
```

One common way to disambiguate is to add a keyword between the expression and the cases. For example:

- OCaml has `with`: `match ... with ...`
- Haskell, Elm, Idris, PureScript, SML, etc. have `of`: `cases ... of ... `

I think using either of the two `match ... with ...` or `case ... of ...` is reasonable. C-like languages seems to have stuck with the switch (Java, TypeScript for the compatibility with JavaScript, C#, etc.), while elsewhere the `match` seems to be popular (Python, Rust, Racket, etc.). `case ... of ...` looks quite specific to Haskell-y or ML-y languages.

Another solution for disambiguation is to use a keyword for cases that makes a clear distinction with record literal, for example:

```nickel
match expr {
  case `foo => ...,
  case `bar => ...,
}
```

However we pay one additional keyword for each case, while the previous solution is just one word for the whole pattern match.

yannham:
Closes #817 #477

This PR changes the keyword of the switch branching construct to a match, to make room for future more advanced features like pattern matching.

A match expression is also now curried, meaning it doesn't have to be applied to an argument and can be used directly as a function:

```nickel
let match_foo = match {`foo => true, _ => false} in
match_foo `bar == false

function.id `bar |> match {`bar => true, `foo => false}
```

## Note

Since we've merged #916, this is going to be hard to cherry-pick this commit for the 0.3 release later if I rebase onto master now. Hence I target the 0.3.0-release branch, and once reviewed and approved, I will merge it there and rebase separately on master to adapt for the eval cache changes.

### DIFF
--- a/src/eval/mod.rs
+++ b/src/eval/mod.rs
@@ -380,40 +380,7 @@ impl<R: ImportResolver> VirtualMachine<R> {
                         env,
                     }
                 }
-                Term::Switch(exp, cases, default) => {
-                    self.set_mode(EvalMode::UnwrapMeta);
 
-                    let has_default = default.is_some();
-
-                    if let Some(t) = default {
-                        self.stack.push_arg(
-                            Closure {
-                                body: t.clone(),
-                                env: env.clone(),
-                            },
-                            pos,
-                        );
-                    }
-
-                    self.stack.push_arg(
-                        Closure {
-                            body: RichTerm::new(
-                                Term::Record(RecordData::with_fields(cases.clone())),
-                                pos,
-                            ),
-                            env: env.clone(),
-                        },
-                        pos,
-                    );
-
-                    Closure {
-                        body: RichTerm::new(
-                            Term::Op1(UnaryOp::Switch(has_default), exp.clone()),
-                            pos,
-                        ),
-                        env,
-                    }
-                }
                 Term::Op1(op, t) => {
                     self.set_mode(op.eval_mode());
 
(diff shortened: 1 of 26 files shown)


vkleen (approved):

src/eval/mod.rs
```diff
@@ -677,6 +644,64 @@ impl<R: ImportResolver> VirtualMachine<R> {
...
+                                ),
+                                env: env.clone(),
+                            },
+                            pos,
+                        );
+
+                        // Now evaluating `%match% arg`, the left-most part of the application `%match%
+                        // arg cases default`, which is in fact a primop application.
+                        self.stack.push_op_cont(
+                            OperationCont::Op1(UnaryOp::Match { has_default }, pos_app),
```
matthew-healy, line 686: I think `has_default` here could just be `true`, right? As we've already destructed `default` above.

matthew-healy (approved):

src/eval/mod.rs
yannham: Unless I'm misreading something, the `if let Some(t) = default` block ends line 670, so right here we may or may not have a default value.

PR #977, opened 2022-12-06
Add manual section on recursive priorities

yannham:


doc/manual/merging.md
```diff
@@ -313,6 +313,61 @@ Dually, values with the `force` annotation are given the highest priority. Such
...
+is to simply annotate each value with a `default`:
+
+```nickel
+{server = {ip | default = "192.168.1.1", port | default = 80}}
+```
+
+However, doing so can quickly become cumbersome for larger sets of default
+values, where we end up duplicating `default` annotations lot. We can do better
+thanks to *recursive priorities*, written `rec default` and `rec force`. During
+evaluation, recursive priorities are dynamically pushed down until values which
```
vkleen, line 343:
```suggestion
evaluation, recursive priorities are dynamically pushed down until they reach values which
```

doc/manual/merging.md
```diff
@@ -313,6 +313,61 @@ Dually, values with the `force` annotation are given the highest priority. Such
...
+overriding other values individually and not as a block.
+
+Another usage of `rec default` and `rec force` is to make a whole record
+overridable or overriding dynamically, when you don't necessarily have the
+option to change its definition (if it's coming from an external library).
+
+If a value already has a merge priority set, the rules are the following:
+
+- `rec default` doesn't apply on values with explicit priorities set. That is,
+  `{foo | rec default {bar | priority 10 = 1, baz = 2}}` will evaluate to
```
vkleen, line 362:
```suggestion
  `{foo | rec default = {bar | priority 10 = 1, baz = 2}}` will evaluate to
```

doc/manual/merging.md
```diff
@@ -313,6 +313,61 @@ Dually, values with the `force` annotation are given the highest priority. Such
...
+  `{foo = {bar | priority 10 = 1, baz | default = 2}}`.
+- `rec force` always overrides existing priorities. That is,
+  `{foo | rec forcde {bar | priority 10 = 1, baz | default = 2}}` will evaluate
```
vkleen, line 365:
```suggestion
  `{foo | rec force = {bar | priority 10 = 1, baz | default = 2}}` will evaluate
```

doc/manual/merging.md
```diff
@@ -313,6 +313,61 @@ Dually, values with the `force` annotation are given the highest priority. Such
...
+are not records anymore, that is to the leafs of the record, so to speak:
+
+```text
+nickel> {server | rec default = {ip = "192.168.1.1", port = 80}} & {server.ip =
+"42.42.42.42"}
+{ server = { ip = "42.42.42.42", port = 80 } }
+```
+
+`rec force` is the same idea for `force`, which makes the value of a record
+overriding other values individually and not as a block.
```
vkleen, line 353: I can't really parse this sentence, but I think my suggestion is what you mean. 😅
```suggestion
`rec force` implements the same idea for `force`. It makes overriding happen for individual fields of a record and not as a block.
```

vkleen (approved):
I've added some suggestions, feel free to pick and choose.

doc/manual/merging.md
```diff
@@ -313,6 +313,60 @@ Dually, values with the `force` annotation are given the highest priority. Such
...
+nickel> {server | rec default = {ip = "192.168.1.1", port = 80}} & {server.ip =
+"42.42.42.42"}
+{ server = { ip = "42.42.42.42", port = 80 } }
+```
+
+`rec force` implements the same idea for `force`. It makes overriding happen for individual fields of a record and not as a block.
+
+Another use of `rec default` and `rec force` is to make a whole record
+overridable or overriding dynamically, when you don't necessarily have the
+option to change its definition (if it's coming from an external library).
```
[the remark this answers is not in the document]
yannham: Yup, this reads way better, thanks!

yannham:
I ended up removing the section on the second use-case entirely. With the syntax on `master` we can't use `rec default` annotations on a random value, so the example was non obvious to write. In general, with RFC005, it's not totally obvious how to use recursive priorities somewhere else than on record fields. For now, I propose not to talk about the usage yet, and see after RFC005 how recursive priorities will look like.

PR #988, opened 2022-12-12
Return the shape string.find expects from %str_find%

### LINKED ISSUE #987: `string.find`, as of nickel 0.3.0, always has a contract error
**Describe the bug**

As of nickel 0.3.0 (and on master), `string.find` always has a contract error now.

**To Reproduce**

```
$ nickel repl

nickel> string.find "a" "a"
error: contract broken by a function: missing field matched
   ┌─ :1:15
   │
 1 │ Str -> Str -> {matched: Str, index: Num, groups: Array Str}
   │               --------------------------------------------- expected return type

nickel> %str_find% "a" "a"
{ match = "a", index = 0, groups = [  ] }
```

The mismatch of "match" vs "matched" for `%str_find%` vs the `string.find` contract is the culprit.

**Expected behavior**

`string.find` should work

**Environment**
- `NixOS 23.05 (Stoat)`
 - Version of the code: 59eec80c

euank:
The included integration test fails without the change.

Fixes #987 

Based on "match" becoming a keyword, I think it's intentional that the rename happens on the `%str_find%` side rather than reverting the `string.find` contract change to still be `match`, so I took that approach for this PR.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -1131,7 +1131,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                             .collect();
 
                         mk_record!(
-                            ("match", Term::Str(String::from(first_match.as_str()))),
+                            ("matched", Term::Str(String::from(first_match.as_str()))),
                             ("index", Term::Num(first_match.start() as f64)),
                             (
                                 "groups",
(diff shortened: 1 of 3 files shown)


euank:
I think we should have much more extensive string.find tests too, but I think it'll be nicer to keep this change small, and add those as a separate PR.

PR #990, opened 2022-12-12
[RFC005] First phase of RFC005: field metadata

### LINKED ISSUE #900: Missing field error message blames the wrong field
**Describe the bug**
This error message doesn't make any sense:
```
nickel> { foo | Bool, bar = if foo then 2 else null }
error: missing definition for `bar`
  ┌─ repl-input-0:1:1
  │
1 │ { foo | Bool, bar = if foo then 2 else null }
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ in this record

note:
  ┌─ repl-input-0:1:9
  │
1 │ { foo | Bool, bar = if foo then 2 else null }
  │         ^^^^ bound here
```
The error presumably should say "missing definition for `foo`".

**Environment**
 - Version of the code: 741552b065108a1564eaffcf4c28a7a96bb8c047

yannham:
Closes #900 #1056

First part of RFC005, which is to restrict metadata to only appear at the level of record fields, basically. Reviewing by commit is probably a sane approach (although there's a lot of purely new code, so this might actually be ok to review as a block, I don't know). I tagged the commits that I think deserve reviews by `[REVIEW]`.

The starting point is to introduce a new `Field` and `FieldMetadata` types, and store them inside `RecordData`. Then, fix everything :slightly_smiling_face: 

One difficulty has been that before, metadata were first class, in that you could put them in a term, closurize them, etc. It made a bunch of things quite simpler, in particular merging, dynamic field extension, and querying. I won't enter the details here: I tried to give a meaningful explanation whenever a commit deserved it in the commit message.

Note that this PR doesn't implement lazy propagation, which is due in a coming one.

## TODO

- [ ] fix and restore the failing test for recursive priorities (future work)
- [x] Update the LSP to work with new definitions (this PR)


src/typecheck/mod.rs
```diff
@@ -1546,15 +1698,36 @@ impl From<ApparentType> for Types {
     }
 }
 
+//TODO: doc
```
ebresafegaga, line 1701: I guess you forgot to add some doc here

lsp/nls/src/linearization/mod.rs
```diff
@@ -456,12 +450,20 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
                     ty,
                     kind: TermKind::Structure,
-                    meta: self.meta.take(),
+                    metadata: self.meta.take(),
                 })
             }
         }
     }
 
+    fn add_field_metadata(&mut self, lin: &mut Linearization<Building>, field: &Field) {
```
ebresafegaga, line 459:
```suggestion
    fn add_field_metadata(&mut self, _lin: &mut Linearization<Building>, field: &Field) {
```

ebresafegaga (approved):
LGTM

src/term/record.rs
```diff
@@ -73,23 +73,140 @@ pub struct RecordDeps {
     pub dyn_fields: Vec<FieldDeps>,
 }
 
+/// A type or a contract together with its corresponding label.
+// #[derive(Debug, PartialEq, Clone)]
+// pub struct LabeledType {
+//     pub types: Types,
+//     pub label: Label,
+// }
```
vkleen, line 84: This type has been moved to `term/mod.rs`.

src/program.rs
```diff
@@ -54,6 +57,108 @@ impl std::str::FromStr for ColorOpt {
...
 
+#[derive(Clone, Default)]
+pub struct QueryPath(pub Vec<Ident>);
+
+impl QueryPath {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn parse(cache: &mut Cache, input: String) -> Result<Self, ParseError> {
```
vkleen, line 72: This parser seems to be a special case of something we already do in the general Nickel parser. Namely, we need to understand field paths already for record definitions like
```
{ foo.bar.baz = ... }
```
It's probably not easy or reasonable to try to call into the generic parser from here. But is there some mechanism to keep these parsers in sync? For example, what if we start adding wildcard fields?

src/error.rs
```diff
@@ -849,6 +849,7 @@ impl ToDiagnostic<FileId> for EvalError {
                 diags
             }
             EvalError::TypeError(expd, msg, orig_pos_opt, t) => {
+                println!("type error term: {} / {:#?}", t, t);
```
vkleen, line 852: That `println!` is probably left over from debugging and should be removed.

vkleen:
```suggestion
```

src/eval/merge.rs
```diff
@@ -496,70 +347,281 @@ pub fn merge<C: Cache>(
...
 }
 
-/// Apply a series of contract to term and closurize the result, and apply the necessary
-/// intermediate closurization.
+/// Apply a series of contract to the value of a field as well as the necessary saturation (see
+/// [saturate]) and closurize operations.
+///
+/// Both the contract and the field are saturated, and the result is closurized. The result is
+/// revertible if either one of the contract or the record is, with the union of their
+/// dependencies.
```
vkleen, line 355:
```suggestion
/// Apply a series of contracts to the value of a field as well as the necessary saturation (see
/// [saturate]) and closurize operations.
///
/// Both the contract and the field are saturated, and the result is closurized. The result is
/// revertible if either the contract or the record is, with the union of their
/// dependencies.
```

src/eval/merge.rs
```diff
@@ -496,70 +347,281 @@ pub fn merge<C: Cache>(
...
+    );
+
+    Ok(Closure {
+        body: RichTerm::from(Term::Var(fresh_var)),
+        env: final_env,
+    })
 }
 
-/// Merge the two optional documentations of a metavalue.
+/// Take two record fields in their respective environment and combine both their metadat and
```
vkleen, line 413:
```suggestion
/// Take two record fields in their respective environment and combine both their metadata and
```

src/transform/share_normal_form.rs
```diff
@@ -149,23 +110,139 @@ pub fn transform_one(rt: RichTerm) -> RichTerm {
...
+/// evaluation of recursive records. Only constant are not required to be
+/// closurized.
+///
+/// In theory, the variable case is one exception: if the field is already a bare
+/// variable, it seems useless to add one more indirection through a generated
+/// variable. However, it is currently fundamental for recursive record merging that
+/// the sare normal form transformation ensure the following post-condition: the
+/// fields of recursive records contain either a constant or a *generated* variable,
+/// but never a user-supplied variable directly (the former starts with a special
+/// marker). See comments inside [`crate::RichTerm::closurize`] for more details.
```
vkleen, line 139:
```suggestion
/// the share normal form transformation ensures the following post-condition: the
/// fields of recursive records contain either a constant or a *generated* variable,
/// but never a user-supplied variable directly (the former starts with a special
/// marker). See comments inside [`crate::RichTerm::closurize`] for more details.
```

vkleen (approved):
I've found a few documentation typos and leftover code from debugging. Functionally, it passes CI and works better than master for `tf-ncl`. So 👍

src/program.rs
yannham: In fact it's not too hard to call the parser: we just have to make the rule public. I just found it a bit too heavy then, but maybe it's a good idea. Would that be ok to let it that way for the time being, and to consider this option in a subsequent PR?

vkleen: I think leaving it the way it is would be perfectly reasonable. This is mostly a maintainability concern if we ever end up changing the accepted syntax for field paths. Which we shouldn't be doing often anyway.

PR #991, opened 2022-12-13
Fix LSP server not giving completion when a non-contract meta-information is present in a declaration

### LINKED ISSUE #989: LSP server doesn't give completion when only non-contract meta-information are present
**Describe the bug**
When we have the following program: 
```
let foo | doc "foo bar" = {bar = 10 } in 
foo[.]
```
and we want a completion for `foo`, the LSP  server doesn't give a completion

**To Reproduce**
Try the example above with the latest LSP server on the `master` branch

**Expected behavior**
It should give completion, as usual. 

**Additional context**
See this [file](https://github.com/tweag/nickel/blob/master/lsp/nls/src/requests/completion.rs)

This happens because in the `collect_record_info` function, we first check for completion data using the available `MetaValue`, with the function `find_fields_from_contract`, and in this case (example above), this returns a `Some(vec![])`, but the `collect_record_info` doesn't bother to check other ways to collect completion items since it sees that it is a `Some(..)`, without checking the contents of the `Some(..)`.

ebresafegaga:
Closes #989

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -333,8 +333,8 @@ fn collect_record_info(
                     let mut p = path.clone();
                     let mut fst = find_fields_from_contract(linearization, *body_id, path)
                         .unwrap_or_default();
-                    let snd =
-                        find_fields_from_term_kind(linearization, id, &mut p).unwrap_or_default();
+                    let snd = find_fields_from_term_kind(linearization, *body_id, &mut p)
+                        .unwrap_or_default();
                     fst.extend(snd);
                     fst
                 }
@@ -478,8 +478,33 @@ mod tests {
     use super::*;
     use crate::linearization::Environment;
     use codespan::Files;
-    use nickel_lang::position::TermPos;
+    use nickel_lang::{position::TermPos, term::MergePriority};
     use std::collections::{HashMap, HashSet};
+
+    fn make_lin_item(
+        id: ItemId,
+        kind: TermKind,
+        meta: Option<MetaValue>,
+    ) -> LinearizationItem<Types> {
+        LinearizationItem {
+            env: Environment::new(),
+            id,
+            pos: TermPos::None,
+            ty: Types(TypeF::Dyn),
+            kind,
+            meta,
+        }
+    }
+
+    fn make_completed(linearization: Vec<LinearizationItem<Types>>) -> Completed {
+        let id_to_index: HashMap<_, _> = linearization
+            .iter()
+            .map(|item| item.id)
+            .enumerate()
+            .map(|(index, id)| (id, index))
+            .collect();
+        Completed::new(linearization, id_to_index)
+    }
 
     #[test]
     fn test_get_identifier_path() {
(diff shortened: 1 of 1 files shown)


matthew-healy (commented):
The fix itself looks good. Does it make sense to add a regression test for this somewhere?

lsp/nls/src/requests/completion.rs
```diff
@@ -68,7 +68,10 @@ fn find_fields_from_contract(
 ) -> Option<Vec<Ident>> {
     let item = linearization.get_item(id)?;
     match &item.meta {
-        Some(meta_value) => Some(find_fields_from_meta_value(meta_value, path)),
+        Some(meta_value) => {
+            let result = find_fields_from_meta_value(meta_value, path);
+            (!result.is_empty()).then_some(result)
+        }
```
yannham, line 74: I'm not sure to understand why this fixes the issue, but I imagine you tested the result. Is it because `find_fields_from_contracts` is called somewhere in a chain, and if it returns something, we keep the result, otherwise we try to find a completion by other means?

If this is case, why don't we aggregate all source of information (types, contracts, literal definition)? For example, if the contract is `{foo, ..}` but we know that the record is equal to `{foo = 1, bar = 2}`, does this mean that even after this PR, completion won't show `bar`?

ebresafegaga: > Is it because find_fields_from_contracts is called somewhere in a chain, and if it returns something, we keep the result, otherwise we try to find a completion by other means?

Yes. 

I think aggregation is better. Thanks!

yannham (approved):
You probably want a hashet when aggregating. Or is there something else preventing duplicate elsewhere ? Otherwise, there is a high change than a record literal definition and the associated contracts will have the same fields defined :slightly_smiling_face:

lsp/nls/src/requests/completion.rs
```diff
@@ -331,9 +328,15 @@ fn collect_record_info(
...
-                    find_fields_from_contract(linearization, *body_id, path)
-                        .or_else(|| find_fields_from_term_kind(linearization, id, path))
-                        .unwrap_or_default()
+                    // The path is mutable, so the first case would consume the path
+                    // so we have to clone it so that it can be correctly used for the second case.
+                    let mut p = path.clone();
+                    let mut fst = find_fields_from_contract(linearization, *body_id, path)
+                        .unwrap_or_default();
+                    let snd =
+                        find_fields_from_term_kind(linearization, id, &mut p).unwrap_or_default();
```
yannham: Then, does it make sense anymore to have an option returned by `find_fields_from` if we're aggregating? Would an empty collection be enough and make sense? (getting wrap of the `unwrap_or_default`)

ebresafegaga: Yes, that would be much better. Thank you!

ebresafegaga: But on a second thought, most of these functions use the monadic `?` operator on `Options` to make error propagation or early return much cleaner

yannham: But do they need that at all? Couldn't you just return an empty collection instead?

yannham: In any case, returning an option should be decided first and foremost by the semantics of the function and the callers though

yannham:
The CI is failing because the order of completion suggestion is not the same. I guess you want to sort your vectors inside your tests (or in the LSP directly)?

PR #992, opened 2022-12-14
Automate adding `.ncl` file test cases

matthew-healy:
There are currently two integration test files where we run checks on a bunch of `.ncl` files in a given directory. Until now, these were manually maintained, meaning developers had to add the `.ncl` file and then remember to actually write the test case.

This PR automates the second step by using the [test-generator](https://github.com/frehberg/test-generator) crate. This crate provides a declarative macro which can build test cases from all file paths matching a given glob, which means we can very easily run the same test for every Nickel source file in a given directory.

This is then applied to the `pass.rs` suite of Nickel files we expect to run without error.


vkleen (approved):
That's a great idea! I'm wondering if we could encode the result annotations in `examples/` in Nickel instead of in comments. But that shouldn't be a blocker.

yannham (commented):
Automating the passing tests looks great! 

For the examples, the solution is clean currently but can feel a bit ad-hoc (why is there `blame` but not `missing field error`, for example). After all, failing examples are not different from other failing tests. I wonder if we should rather design a more generic framework (and, why not, describing what we expect in Nickel itself :sunglasses: as @vkleen  ). Or if in the end Rust is more adapted to describe the pattern of errors we expect.

On the other hand, examples aren't expected to change a lot, so maybe this doesn't hurt for now. Would that be easy to split this PR in two (automating passing tests vs examples)? If not, nevermind.

matthew-healy:
> why is there blame but not missing field error, for example

Mostly because there are currently no examples which would use a missing field error right now, and I'm a big proponent of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). :grin: 

As I said in the PR description, this could definitely be expanded on in the future, but I think what's here brings benefit now and can serve as a basis for further iteration (which I'm very much in favour of - I love the idea of a generic framework for annotating files with expected errors).

> Would that be easy to split this PR in two (automating passing tests vs examples)?

Very easy - I can just drop the last commit from here & open a new PR with just that. I'll do it just now, then we can merge the first two commits & continue the discussion about the examples on that PR.

matthew-healy:
I dropped the `examples/` commit from this PR & opened #997 to further discuss that change.

yannham (approved):

PR #993, opened 2022-12-14
LSP completion for import terms

ebresafegaga:
Closes item 1 in #877

### DIFF
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -40,16 +40,19 @@ impl CacheExt for Cache {
         initial_env: &Environment,
         lin_cache: &mut HashMap<FileId, Completed>,
     ) -> Result<CacheOp<()>, CacheError<TypecheckError>> {
-        if !self.terms_mut().contains_key(&file_id) {
+        if !self.terms().contains_key(&file_id) {
             return Err(CacheError::NotParsed);
         }
 
-        if let CacheOp::Done(ids) = self.resolve_imports(file_id).unwrap() {
+        if let Ok(CacheOp::Done(ids)) = self.resolve_imports(file_id) {
             for id in ids {
-                self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache).unwrap();
+                // Linearize all imports in this file
+                // NOTE: This only goes down one level
+                self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache)
+                    .unwrap();
             }
         }
-        
+
         // After self.parse(), the cache must be populated
         let CachedTerm { term, state, .. } = self.terms().get(&file_id).unwrap();
 
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -60,9 +60,11 @@ impl CacheExt for Cache {
             Ok(CacheOp::Cached(()))
         } else if *state >= EntryState::Parsed {
             let host = AnalysisHost::new(file_id, initial_env.clone());
+            let terms = self.terms();
             let building = Linearization::new(Building {
                 lin_cache,
                 linearization: Vec::new(),
+                terms
             });
             let (_, linearized) =
                 typecheck::type_check_linearize(term, initial_ctxt.clone(), self, host, building)?;
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -47,6 +47,9 @@ impl CacheExt for Cache {
         if let Ok(CacheOp::Done(ids)) = self.resolve_imports(file_id) {
             // TODO: we need a way to stop infite recursion with imports
             for id in ids {
+                if id == file_id {
+                    continue;
+                }
                 // Linearize all imports in this file
                 // NOTE: This only goes down one level
                 self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache)
@@ -65,7 +68,7 @@ impl CacheExt for Cache {
             let building = Linearization::new(Building {
                 lin_cache,
                 linearization: Vec::new(),
-                terms
+                terms,
             });
             let (_, linearized) =
                 typecheck::type_check_linearize(term, initial_ctxt.clone(), self, host, building)?;
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -46,10 +46,6 @@ impl CacheExt for Cache {
 
         if let Ok(CacheOp::Done(ids)) = self.resolve_imports(file_id) {
             for id in ids {
-                if id == file_id {
-                    // Avoid infinite recursion
-                    continue;
-                }
                 self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache)
                     .unwrap();
             }
@@ -62,11 +58,10 @@ impl CacheExt for Cache {
             Ok(CacheOp::Cached(()))
         } else if *state >= EntryState::Parsed {
             let host = AnalysisHost::new(file_id, initial_env.clone());
-            let terms = self.terms();
             let building = Linearization::new(Building {
                 lin_cache,
                 linearization: Vec::new(),
-                terms,
+                cache: self,
             });
             let (_, linearized) =
                 typecheck::type_check_linearize(term, initial_ctxt.clone(), self, host, building)?;
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -44,11 +44,11 @@ impl CacheExt for Cache {
             return Err(CacheError::NotParsed);
         }
 
-        // if let CacheOp::Done(ids) = self.resolve_imports(file_id).unwrap() {
-        //     for id in ids {
-        //         self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache).unwrap();
-        //     }
-        // }
+        if let CacheOp::Done(ids) = self.resolve_imports(file_id).unwrap() {
+            for id in ids {
+                self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache).unwrap();
+            }
+        }
         
         // After self.parse(), the cache must be populated
         let CachedTerm { term, state, .. } = self.terms().get(&file_id).unwrap();
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -45,6 +45,7 @@ impl CacheExt for Cache {
         }
 
         if let Ok(CacheOp::Done(ids)) = self.resolve_imports(file_id) {
+            // TODO: we need a way to stop infite recursion with imports
             for id in ids {
                 // Linearize all imports in this file
                 // NOTE: This only goes down one level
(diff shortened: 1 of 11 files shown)


lsp/nls/src/linearization/mod.rs
```diff
@@ -24,6 +25,24 @@ pub mod building;
...
+            }
+            None => {
+                let mut table = HashMap::new();
+                table.insert(file, value);
+                LIN_CACHE = Some(table);
+            }
+        }
+    }
+}
+
```
vkleen, line 45: Does the linearization cache really need to be `static mut`? As far as I can tell, it's only accessed in `impl`-methods on `AnalysisHost`. Could it instead be a field there? That would also save us from these `unsafe` blocks.

yannham: I agree. `unsafe` seems like a code smell in this situation, which doesn't look like anything special requiring something as powerful.

ebresafegaga: Yes, my initial solution was to have a `&HashMap<FileId, Completed>`  in the `Completed` data structure, but then I ran into lifetime scoping issues. I then decided to `clone` the `lin_cache`, but the `lin_cache` gets very large (contains `stdlib` and other opened files), so this significantly slowed down the LSP.  We don't need it to be accessible from `AnalysisHost` methods, we actually want it to be accessible from `impl`-methods on `Completed`

ebresafegaga: See [here](https://github.com/tweag/nickel/pull/993/commits/a8dc7c6649968efd789da812b9189f161087d1cc)

vkleen: If the problem is just lifetime scoping and `Completed` should actually hold a reference to the linearization cache, could that be solved with `Rc`'s?

ebresafegaga: After using an `Rc` we just have this unsafe code: 
```rust
  pub fn new(
        linearization: Vec<LinearizationItem<Resolved>>,
        id_to_index: HashMap<ItemId, usize>,
        lin_cache: &HashMap<FileId, Completed>,
    ) -> Self {
        Self {
            linearization,
            id_to_index,
            lin_cache: unsafe { Rc::from_raw(lin_cache) },
        }
    }

```

lsp/nls/src/cache.rs
```diff
@@ -40,20 +40,33 @@ impl CacheExt for Cache {
...
         lin_cache: &mut HashMap<FileId, Completed>,
     ) -> Result<CacheOp<()>, CacheError<TypecheckError>> {
-        if !self.terms_mut().contains_key(&file_id) {
+        if !self.terms().contains_key(&file_id) {
             return Err(CacheError::NotParsed);
         }
 
+        if let Ok(CacheOp::Done(ids)) = self.resolve_imports(file_id) {
+            for id in ids {
+                if id == file_id {
```
yannham, line 49: To be checked, but I'm believe `resolve_imports` doesn't return the `file_id` given as a parameter, but just the ones of the import encountered inside `file_id`?

ebresafegaga: Yes, it goes arbitrarily deep and returns all the file ids of the imports it can reach.

src/cache.rs
```diff
@@ -692,7 +709,7 @@ impl Cache {
                 "cache::prepare(): expected source to be parsed before imports resolutions",
             )
         })?;
-        if import_res == CacheOp::Done(()) {
+        if matches!(import_res, CacheOp::Done(..)) {
```
yannham, line 712: I think you can use a `if let CacheOp::Done(..) = import_res {`, which doesn't resort to a macro.

src/cache.rs
```diff
@@ -637,15 +637,23 @@ impl Cache {
...
+            Some(state) if state >= EntryState::ImportsResolved => Ok(CacheOp::Cached(Vec::new())),
             Some(state) if state >= EntryState::Parsed => {
-                if state < EntryState::ImportsResolving {
+                let pending = if state < EntryState::ImportsResolving {
                     let CachedTerm {
                         term, parse_errs, ..
-                    } = self.terms.remove(&file_id).unwrap();
+                    } = self.terms.get(&file_id).unwrap();
+                    // okay, for now these clones are here becuase the function call below
+                    // short circuts, and then we don't put back the item in cache, so the
```
yannham, line 649: I'm not sure to fully understand this comment. What do you mean by "short-circuits"?

yannham: And what is the "function below"? Probably better to name it explicitly

src/cache.rs
```diff
@@ -655,19 +663,28 @@ impl Cache {
...
-                        self.resolve_imports(id)?;
-                    }
+                    pending
+                        .iter()
+                        .flat_map(|id| {
+                            if let Ok(CacheOp::Done(mut ps)) = self.resolve_imports(*id) {
+                                ps.push(*id);
+                                ps
+                            } else {
+                                vec![*id]
```
yannham, line 673: Do we really want to include `ids` where parse errors happened? I think that for the LSP we are in error-tolerant mode anyway, so this case probably doesn't happen.

yannham (commented):
The ownership structure looks better. It makes sense that the server owns the map of linearization.

Beside some specific comments, I also think we should avoid leaking implementation details like `HashMap<FileId, RichTerm>` into the LSP. We can rather access the cache interface directly, probably.

lsp/nls/src/cache.rs
yannham: yes, but `transform::resolve_import` or whatever is called only return the import that has not been resolved before I think. It's not very important but the cache should already be resilient to cyclic imports (I think there are even tests for that), so this is most probably not needed, in that `file_id` shouldn't be in the return value of `resolve_imports(file_id)`

lsp/nls/src/cache.rs
```diff
@@ -40,20 +40,33 @@ impl CacheExt for Cache {
...
+        }
+
         // After self.parse(), the cache must be populated
         let CachedTerm { term, state, .. } = self.terms().get(&file_id).unwrap();
 
         if *state > EntryState::Typechecked && lin_cache.contains_key(&file_id) {
             Ok(CacheOp::Cached(()))
         } else if *state >= EntryState::Parsed {
             let host = AnalysisHost::new(file_id, initial_env.clone());
+            let terms = self.terms();
```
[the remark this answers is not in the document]
ebresafegaga: Thank you!

src/cache.rs
ebresafegaga: So previously, before we resolve an import term in a file, we (1) remove the file's `RichTerm` from the cache, (2) resolve the import, (3) update the file's `RichTerm` and (4) put it back in the cache. However, the "resolve the import" phase can fail for whatever reason (e.g wrong file path, maybe parse errors, etc), and if it fails, we no longer put the `RichTerm` back in the cache, because of the monadic operator `?` (this is what I mean by short circuits), and the file cannot be linearized if the `RichTerm` is not present, so the LSP crashes.

yannham: What about putting it back in the case `resolve_imports` return an error? I'm just curious though, since cloning a term isn't very expensive, I guess it's fine and less error-prone this way.

ebresafegaga: The initial term is moved by the call to `import_resolution::resolve_imports(term, self)`, and if it fails, there's nothing to put back in the cache

yannham: Ah, right. Then cloning it is :shrug:

lsp/nls/src/linearization/mod.rs
```diff
@@ -393,6 +395,37 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
+                }
+
+                let Some(linearization) = lin.lin_cache.get(file) else {
+                    return
+                };
+                // This is safe because the import file is resolved, before we linearize
+                // the containing file, therefore the cache MUST have the term stored.
+                let term = lin.cache.get(*file).unwrap();
+                let position = final_term_pos(&term);
+                let locator = (*file, position.unwrap().start);
```
yannham, line 417: is it the case that `position` is always defined? If yes, maybe add a comment to explain why as well. If not, maybe unwrapping isn't the best idea, and we should just return as well.

src/cache.rs
```diff
@@ -634,18 +634,27 @@ impl Cache {
     /// nothing if the imports of the entry have already been resolved. Require that the
     /// corresponding source has been parsed.
     /// If resolved imports contain imports themselves, resolve them recursively.
+    /// It returns a `Vec<FileId>`, which is a vector of all the imports that were resolved.
```
yannham, line 637:
```suggestion
    /// It returns a `Vec<FileId>`, which is a vector of all the imports that were transitively resolved.
```

yannham (approved):
Yay, goto definition through files! Left some nitpick comments, feel free to do whatever you want with it. Otherwise, good to go.

lsp/nls/src/linearization/mod.rs
```diff
@@ -393,6 +395,37 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                     })
                 }
             }
+            Term::ResolvedImport(file) => {
+                fn final_term_pos(term: &RichTerm) -> &TermPos {
```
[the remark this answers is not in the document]
ebresafegaga: Yes, exactly. It returns the position of the "result" of an expression

lsp/nls/src/linearization/mod.rs
```diff
@@ -393,6 +395,37 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
+                            final_term_pos(body)
+                        }
+                        Term::Op1(UnaryOp::StaticAccess(field), _) => &field.pos,
+                        _ => pos,
+                    }
+                }
+
+                let Some(linearization) = lin.lin_cache.get(file) else {
+                    return
+                };
```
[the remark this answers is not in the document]
ebresafegaga: Yes

lsp/nls/src/linearization/mod.rs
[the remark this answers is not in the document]
ebresafegaga: If it's a `Let` it returns the final inner expression

lsp/nls/src/linearization/mod.rs
```diff
@@ -403,6 +404,7 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                             final_term_pos(body)
                         }
                         Term::Op1(UnaryOp::StaticAccess(field), _) => &field.pos,
+                        Term::Fun(ident, _) | Term::FunPattern(Some(ident), ..) => &ident.pos,
```
yannham, line 407: Are we coping with the bug by searching for the indent position, because it's mistakenly stored instead of the function body? No, please, let's not add a new strange behavior on top of a bug. We can do that in a subsequent PR but the right thing to do is to fix the behavior that registers the position of the function as the position of its argument.

yannham (changes requested):

lsp/nls/src/linearization/mod.rs
ebresafegaga: Okay

yannham (approved):

Issue #948, opened 2022-11-28
Symbolic strings (Nix string contexts-like)


yannham:
**Is your feature request related to a problem? Please describe.**

Working on [Nickel-nix](https://github.com/nickel-lang/nickel-nix) and in general Nix integration (#693), we've been needing something like [Nix string context](https://shealevy.com/blog/2018/08/05/understanding-nixs-string-context/).

String context is a way of implicitly and automatically attaching and combining metadata to string values (in the case of Nix, the dependencies that must be built before the paths present inside the string become valid). When interpolating strings with context inside another string, all the dependencies (the contexts) are combined. This feature is really useful to avoid specifying obvious dependencies explicitly (e.g. source files).

However, we don't want to implement Nix string contexts as it is, because it's pretty ad-hoc and Nix specific. We would rather like to have a more general mechanism, of which string context would just be an instance, that may be used for other domains (Terraform, Kubenertes, etc.), or different use-cases within Nix (IFD/recursive Nix-like).

Fundamentally, Nix string context are an overloading of string interpolation (and other string operations) to work on richer values than just string. Very schematically, Nix strings are rather `{ctxt : Array Deps, value: Str}`.

We've discussed the possibilities many times. Having a general ad-hoc overloading mechanism would be possible but pretty heavy (think trait/typeclasses, or even a very restricted form just for strings), with the usual problems of coherence, complexity for new users, etc.

In some way, Nix string context might be implemented armed with effects (#85), e.g. if we allow to perform effects at string interpolation. However such an effect system is still to be properly designed for Nickel, and effects handler would be implemented in Rust, as interpreter plugins, which make them rather heavy to implement and distribute. For something like Nix string context, that could be ok, as we would have to do it once and for all per target tools. It's still a long way to get there.
 
This issue makes a simple and lighter proposal that could achieve the same effect, but relies only on one language feature (very small) and otherwise pure Nickel library code. It also seems to be forward-compatible with performing effects at string interpolation.

**Describe the solution you'd like**

We propose to introduce a new form of strings, let's call them symbolic strings, and write them using the delimiters `s%"` and `"%s`. Normal strings with interpolation are parsed as a list of chunks, where one chunk is either a string literal or an interpolated expression. For example, `"foo %{bar} baz"` is represented as (something like) `[Chunk::Literal("foo "), Chunk::Expr(..), Chunk::Literal(" baz")]`. String chunks are then evaluated at runtime, when first encountered, and turned into an actual string.

[...]

PR #994, opened 2022-12-14
Symbolic strings

matthew-healy:
This PR introduces a simple form of "symbolic strings" (#948), whereby the syntax `s%"..."%` is just de-sugared to an array.

### DIFF
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -474,7 +474,9 @@ Bool: bool = {
     "false" => false,
 };
 
-// Strings that support interpolation.
+// String-like syntax which supports interpolation.
+// Depending on the opening brace, these either parse as strings, or as "symbolic strings",
+// which get desugared here to an array of terms.
 StrChunks: RichTerm = {
   <start: StringStart> <fst: ChunkLiteral?> <chunks: (ChunkExpr+ChunkLiteral)*> <lasts:ChunkExpr*> <end: StringEnd> => {
         debug_assert!(
@@ -493,20 +495,30 @@ StrChunks: RichTerm = {
             .chain(lasts.into_iter())
             .collect();
 
-        let mut chunks = if start.needs_strip_indent() {
+        let chunks = if start.needs_strip_indent() {
             strip_indent(chunks)
         } else {
             chunks
         };
-        chunks.reverse();
 
-        RichTerm::from(Term::StrChunks(chunks))
+        if start == StringStartDelimiter::Symbolic {
+            let terms = chunks.into_iter().map(|chunk| match chunk {
+                StrChunk::Literal(l) => Term::Str(l).into(),
+                StrChunk::Expr(e, _) => e,
+            }).collect();
+            RichTerm::from(Term::Array(terms, Default::default()))
+        } else {
+            let mut chunks = chunks;
+            chunks.reverse();
+            RichTerm::from(Term::StrChunks(chunks))
+        }
     },
 };
 
 StringStart : StringStartDelimiter = {
     "\"" => StringStartDelimiter::Standard,
     "m%\"" => StringStartDelimiter::Multiline,
+    "s%\"" => StringStartDelimiter::Symbolic,
 };
 
 StringEnd : StringEndDelimiter = {
@@ -886,6 +898,7 @@ extern {
         "\"" => Token::Normal(NormalToken::DoubleQuote),
         "\"%" => Token::MultiStr(MultiStringToken::End),
         "m%\"" => Token::Normal(NormalToken::MultiStringStart(<usize>)),
+        "s%\"" => Token::Normal(NormalToken::SymbolicStringStart(<usize>)),
 
         "Num" => Token::Normal(NormalToken::Num),
         "Dyn" => Token::Normal(NormalToken::Dyn),
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -523,7 +523,7 @@ StringStart : StringStartDelimiter = {
 
 StringEnd : StringEndDelimiter = {
     "\"" => StringEndDelimiter::Standard,
-    "\"%" => StringEndDelimiter::Multiline,
+    "\"%" => StringEndDelimiter::Special,
 };
 
 ChunkLiteral : String =
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -477,7 +477,10 @@ Bool: bool = {
 // Strings that support interpolation.
 StrChunks: RichTerm = {
   <start: StringStart> <fst: ChunkLiteral?> <chunks: (ChunkExpr+ChunkLiteral)*> <lasts:ChunkExpr*> <end: StringEnd> => {
-        debug_assert_eq!(start, end);
+        debug_assert!(
+            start.is_closed_by(&end),
+            "Fatal parser error: a string starting with {start:?} should never be closed by {end:?}"
+        );
 
         let chunks: Vec<StrChunk<RichTerm>> = fst.into_iter()
             .map(StrChunk::Literal)
@@ -490,10 +493,9 @@ StrChunks: RichTerm = {
             .chain(lasts.into_iter())
             .collect();
 
-        let mut chunks = if start == StringKind::Multiline {
+        let mut chunks = if start.needs_strip_indent() {
             strip_indent(chunks)
-        }
-        else {
+        } else {
             chunks
         };
         chunks.reverse();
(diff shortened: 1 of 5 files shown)


src/parser/grammar.lalrpop
```diff
@@ -490,26 +495,35 @@ StrChunks: RichTerm = {
             .chain(lasts.into_iter())
             .collect();
 
-        let mut chunks = if start == StringKind::Multiline {
+        let chunks = if start.needs_strip_indent() {
```
yannham, line 498: I had this in mind but forget to ask. Should we behave like multiline string with respect to indentation, or not? It seems like you answer "no", which is very reasonable. That being said, I wonder if we then would like to combine the semantics of symbolic strings and multiline strings, e.g. in Nix often we currently use multiline strings to write builder commands, where symbolic strings would be used. This may not be trivial with respect to indenting interpolated values, though. I guess we don't have to answer this question right now.

src/parser/utils.rs
```diff
@@ -23,10 +23,43 @@ use crate::{
     types::{TypeF, Types},
 };
 
-/// Distinguish between the standard string separators `"`/`"` and the multi-line string separators
-/// `m%"`/`"%` in the parser.
+/// Distinguish between the standard string opening delimiter `"`, the multi-line string
+/// opening delimter `m%"`, and the symbolic string opening delimiter `s%"`.
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
-pub enum StringKind {
```
yannham, line 29: Yes, one XXXKind less in the codebase :tada:

src/parser/utils.rs
```diff
@@ -23,10 +23,43 @@ use crate::{
...
+}
+
+impl StringStartDelimiter {
+    pub fn is_closed_by(&self, close: &StringEndDelimiter) -> bool {
+        matches!(
+            (self, close),
+            (StringStartDelimiter::Standard, StringEndDelimiter::Standard)
+                | (
+                    StringStartDelimiter::Multiline,
+                    StringEndDelimiter::Multiline
```
yannham, line 42: Nitpick: should we call this `Multiline` if it also works for `Symbolic`? It looks inconsistent with `StringStartDelimiter`. Maybe `Special`, if we want to call all those `xxx%"` strings "special strings".

src/parser/utils.rs
```diff
@@ -23,10 +23,43 @@ use crate::{
...
+                    StringStartDelimiter::Symbolic,
+                    StringEndDelimiter::Multiline
+                )
+        )
+    }
+
+    pub fn needs_strip_indent(&self) -> bool {
+        match self {
+            StringStartDelimiter::Standard => false,
+            StringStartDelimiter::Multiline | StringStartDelimiter::Symbolic => true,
```
yannham, line 54: Ah, so what I wrote before is wrong, multiline strings and symbolic strings actually behave the same. I'm not sure how this works with interpolated values though. There is some logic at runtime such that:

```
let bar = "alpha\nbeta" in
m%"
  foo
     %{bar}
  baz
"%
```

evaluates to 

```
foo
  alpha
  beta
baz
```

I don't think we can easily keep this logic for symbolic strings, unless we delegate this to custom parsing functions (but that start to be cumbersome).

yannham (commented):
Mostly LGTM. I just wonder what we should do about the "multiline" semantics of symbolic strings. The first question is: should we have symbolic strings be multiline by default, or offer a separate version for multiline strings? 

Another solution would be to have `s"foo"` strings as well. I think C++ has those. Then there is a clear delimitation between the multiline axis and the "special" axis.

The second question, which is independent, is what to do with interpolated blocks inside mutliline strings. Currently, we carry around inside string chunks an extra `ident` arguments to handle those in the natural way (see https://github.com/tweag/nickel/pull/994#discussion_r1049733871). One possibility would be to pass this argument to the parsing function, but it can quickly become cumbersome to pass this information around.

Another solution is just to abandon this semantics for symbolic strings (and do not interpolate other strings as "blocks"). It would be inconsistent though, that the behavior differs between multiline strings and those symbolic string.

src/parser/utils.rs
matthew-healy: I'm not so sure about "special" as a term of art here. I'm not sure it really means anything beyond "these are different somehow".

What about just being very literal and using `StringEndDelimiter::Quote` and `StringEndDelimiter::QuotePercent`?

yannham: > I'm not so sure about "special" as a term of art here. I'm not sure it really means anything beyond "these are different somehow".

Yes.... in this case I can't think of something better than just saying "it's not the normal string delimiter".

> What about just being very literal and using StringEndDelimiter::Quote and StringEndDelimiter::QuotePercent?

It's a possible solution as well. It has the advantage of being descriptive. Though it's tied to the syntax instead of the semantics, so if we change `%` for `#`, we'll have to change that.

Both are reasonable!

matthew-healy:
@yannham

> The first question is: should we have symbolic strings be multiline by default, or offer a separate version for multiline strings?

I think starting off by only supporting multiline symbolic strings makes sense. Especially as it's possible to just write mutliline symbolic strings on a single line. 

I can't currently think of a use case for a syntax like `s"..."` that isn't already supported by the multiline version, but maybe I'm missing something?

>The second question, which is independent, is what to do with interpolated blocks inside mutliline strings.

In the current iteration I just ignore the `indent` parameter. My thinking on this was that symbolic strings will always go through post-processing before they ever end up as regular strings, so making assumptions about how the interpolated values should be indented feels like it might be unnecessary, as Nickel has absolutely no control over the ultimate string value (assuming one is ever created).

To look at your example from the comment above:

```nickel
let bar = "alpha\nbeta" in
m%"
  foo
     %{bar}
  baz
"%
```

currently evaluates to

```
[ "
foo
  ", 
"alpha
beta",
"
baz"
]
```

This _feels_ like how I'd expect this to behave (though obviously that's completely unscientific :sweat_smile:) - it seems to make sense to give the exact value that was interpolated here. Though I suppose one place it could matter is if the symbolic string is intended to be evaluated into a language with meaningful whitespace... :thinking: 

Ultimately, I don't have particularly strong feelings & am happy with whatever you think makes sense. I do think what we have now is probably "good enough" for a first pass - what do you think about opening a separate issue for how to handle multiline interpolated values?

src/parser/utils.rs
matthew-healy: Fair point about coupling the syntax and semantics. I'll go with `Special` for now because I can't think of anything better. :sweat_smile:

vkleen:
I think @matthew-healy's example evaluation is the most reasonable. I wouldn't expect Nickel to process a string that's interpolated into a symbolic string at all. That's also in line with how normal string interpolation in indented multiline strings behaves.

yannham:
>  That's also in line with how normal string interpolation in indented multiline strings behaves.

Precisely, I don't think this is the case:

```
nickel> let bar = "alpha\nbeta" in
m%"
  foo
     %{bar}
  baz
"%
"foo
   alpha
   beta
baz"

nickel> let bar = "alpha\nbeta" in
s%"
  foo
     %{bar}
  baz
"% |> array.foldl (++) ""
"foo
   alpha
beta
baz"

```

We carry explicitly the whole indentation of an interpolation (here e.g. 4 spaces for the `%{bar}` block) and re-apply that at runtime, once we've evalauted bar, to the beginning of each line. We can't do that, at least not trivially, for symbolic string, because the parsing function is the one reconstructing the string (if ever done) at the end. The best we can do is to pass this indentation. Maybe we don't care :man_shrugging: but I just wanted to raise the difference in behavior.  

> Ultimately, I don't have particularly strong feelings & am happy with whatever you think makes sense. I do think what we have now is probably "good enough" for a first pass - what do you think about opening a separate issue for how to handle multiline interpolated values?

Sure, this is the way to go.

yannham (approved):
:tada:

vkleen:
@yannham That's interesting! I tested with the following in `test.ncl`:
```nickel
let s = "a\nb" in
  m%"foo
  in %{s}
  bar"%
```
and `nickel -f test.ncl` results in
```
"foo
  in a
b
  bar"
```
I think the indent behaviour only kicks in if the multiline string starts with whitespace?

matthew-healy:
@vkleen I think it kicks in if the line the interpolation is on starts with whitespace, so e.g.

```nickel
let s = "a\nb" in
m%"hello
    %{s}
test
"%
```

results in

```
hello
    a
    b
test
```

But because your example has the text `in` on the same line we don't count that as an indent.

yannham:
> I think the indent behaviour only kicks in if the multiline string starts with whitespace?

Yes :slightly_smiling_face: you can find more context in #197. Maybe this information should be somewhere else, in the documentation?

PR #997, opened 2022-12-15
Automate adding tests for examples

matthew-healy:
This PR updates the  `examples.rs` test file to use the `test-generator` crate to automatically generate tests from programs in the `examples/` directory. 

Since not every example file is (1) actually tested or (2) expected to pass, this PR also introduces a limited form of annotation in these files. This is a testing approach I first noticed in the Swift project (see, e.g., [here](https://github.com/apple/swift/blob/main/test/expr/expressions.swift#L50)), where source files are annotated with expectations about their behaviour in comments, which are then parsed during testing. 

Currently the annotations here aren't particularly complex - they just expect a file to pass, fail with a blame error, or ignore the file entirely. This is something which could - & should! - be expanded in future (which would mean we no longer need to write Nickel code inside Rust strings when we want to test failure cases).

### DIFF
--- a/examples/arrays/arrays.ncl
+++ b/examples/arrays/arrays.ncl
@@ -1,3 +1,5 @@
+# test: pass
+
 # Example array functions. This code is illustrative: prefer using the array 
 # stdlib functions `array.map` and `array.fold` instead.
 let my_array_lib = {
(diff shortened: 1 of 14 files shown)


yannham (approved):
As a side note, that we could already put the failing tests in separate files that we load from Rust instead of embedding them inside failing tests. But indeed, we still have to specify somewhere else how we expect the test to fail. I wonder if a custom DSL would that much butter than just writing Rust patterns, I really don't know.

However, I guess examples are a good niche, because we probably won't have to handle much more complex failing mode, and the `# test : xxx` syntax is really trivial to understand.

Issue #1003, opened 2022-12-16
`string.length` returns the number of bytes, not length of a string


euank:
**Describe the bug**

Most string functions, like `string.split`, `string.chars`, etc, operate on actual unicode characters.

`string.length` seems to operate on number of bytes.

**To Reproduce**

```
nickel> string.length "👨‍❤️‍💋‍👨"
27

nickel> string.length "四字熟語"
12
```

**Expected behavior**

I believe those should be 1 and 4 respectively.

**Environment**
 - NixOS something or other
 - current master, 33caa639

**Additional context**

I'm happy to implement a fix if the expected behavior above is what we want! I think it is, but it's possible I'm missing some reason here.

euank:
A similar issue occurs for substring.
I'm happy to file a separate issue for this one if you'd like!

Basically:

```
nickel> string.substring 3 6 "子猫"
"猫"

nickel> string.substring 0 1 "子猫"
error: substring: index out of bounds. Expected the 3rd argument (end) to be between 1 and 6, got 1

nickel> string.substring 1 2 "子猫"
error: substring: index out of bounds. Expected the 2nd argument (start) to be between 0 and 6, got 1
```

Kinda confusing error there too.

aspiwack:
Thanks for raising this.

Length of Unicode strings is always a thorny subject. I don't know the encoding for CJK characters, but I imagine that it's mostly 1 code point per glyph, so it's alright. But I don't know that there are good ways to count glyphs in general, is there? Then you have to decide on a normalisation and count code points in said normalised form, probably. Or maybe just count code points in the unnormalised string that you get as input?

At any rate, I agree that counting bytes is probably not what we want.

matthew-healy:
If we want to count code points then we can just replace [this line](https://github.com/tweag/nickel/blob/master/src/eval/operation.rs#L988) with 

```rust
Term::Num(s.chars().count() as f64)
```

In the second example that gives a length of `4` as expected, though in the first example it gives a length of `8`. I know Swift does some magick while calculating string sizes & does actually return `1` for the first example, but I haven't looked into how in any detail. 

Personally I think it'd be fine to just count code points for now & go to the extra effort later if it turns out knowing the "visual" size of emoji is critical to some kind of configuration wrangilng we want to support. :grin:

aspiwack:
[...]

PR #1004, opened 2022-12-16
Add integration tests for string stdlib funcs

euank:
This adds happy-path coverage for the string stdlib functions.

A couple cases were left out due to bugs discovered over here: https://github.com/tweag/nickel/issues/1003

Those can be added once that issue's closed. In the meanwhile, this gives a pretty significant increase in coverage, and should hopefully avoid breaking the string stdlib by accident.

### DIFF
--- a/tests/integration/pass/strings.ncl
+++ b/tests/integration/pass/strings.ncl
@@ -11,6 +11,123 @@ let {check, ..} = import "lib/assert.ncl" in
     "Hello, %{x}! Welcome in %{let y = "universe" in "the %{x}-%{y}"}")
     == "Hello, world! Welcome in the world-universe",
 
+  # happy path for all stdlib functions
+  # string.BoolLiteral
+  ("True" | string.BoolLiteral) == "true",
+  ("False" | string.BoolLiteral) == "false",
+  ("true" | string.BoolLiteral) == "true",
+  ("false" | string.BoolLiteral) == "false",
+  # string.NumLiteral
+  ("+1.2" | string.NumLiteral) == "+1.2",
+  ("-1.2" | string.NumLiteral) == "-1.2",
+  ("9001" | string.NumLiteral) == "9001",
+  # string.CharLiteral
+  ("e" | string.CharLiteral) == "e",
+  # string.EnumTag
+  (`Foo | string.EnumTag) == `Foo,
+  # string.Stringable
+  (`Foo | string.Stringable) == `Foo,
+  (true | string.Stringable) == true,
+  (1 | string.Stringable) == 1,
+  ("_" | string.NonEmpty) == "_",
+  # string.join
+  string.join " " ["Hello", "世界"] == "Hello 世界",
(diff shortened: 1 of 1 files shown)


tests/integration/pass/strings.ncl
```diff
@@ -11,6 +11,123 @@ let {check, ..} = import "lib/assert.ncl" in
...
+  (1 | string.Stringable) == 1,
+  ("_" | string.NonEmpty) == "_",
+  # string.join
+  string.join " " ["Hello", "世界"] == "Hello 世界",
+  string.join "anything" [] == "",
+  string.join "a" ["b", "n", "n", ""] == "banana",
+  # string.split
+  string.split "," "1,2,3" == ["1", "2", "3"],
+  string.split "" "123" == ["", "1", "2", "3", ""],
+  string.split "❤" "👩‍❤️‍💋‍👨" == [ "👩‍", "️‍💋‍👨" ],
```
euank, line 40: I'm not sure if this one is correct or not. It seems arguable either way. Same for the one below this one.

tests/integration/pass/strings.ncl
```diff
@@ -11,6 +11,123 @@ let {check, ..} = import "lib/assert.ncl" in
...
+  # string.trim
+  string.trim "  ひげ  " == "ひげ",
+  string.trim "   " == "",
+  string.trim "" == "",
+  string.trim "​ ​ ​ a ​ ​ ​" == "​ ​ ​ a ​ ​ ​", # zero width spaces aren't whitespace
+  string.trim "   \n   new\nline\n" == "new\nline",
+  string.trim "　   	  " == "",
+  # string.chars
+  string.chars "漢字" == ["漢", "字"],
+  string.chars "👨‍❤️‍💋‍👨" == ["👨", "‍", "❤", "️", "‍", "💋", "‍", "👨"], # arguably wrong, splits the ZWJ combined emoji
```
euank, line 52: This one's also arguable either way in my opinion.

matthew-healy (approved):
This is a really great suite of tests, thanks so much for adding them!

For the cases you've called out as being unusual, I think it's still reasonable to have tests for the _current_ behaviour, accompanied with a comment (& ideally an issue) pointing out that this should probably change. :+1: 

If I was nitpicking, I'd maybe say that this could be a separate file - `stdlib_string_tests.ncl` - but I don't think that needs to block the PR from being merged. We'll probably change the structure as we add tests for the rest of the stdlib anyhow.

yannham (approved):
We've been talking about testing the stdlib in a systematic way not later than yesterday :slightly_smiling_face: including some form of property-based testing.

In the meantime, those tests should prevent the kind of error we had on 0.3 for the string library. Thanks for writing them. I agree with @matthew-healy, let's have the controversial tests be adapted to the current behavior. We'll change them in the future if needed (we will have to).

I agree as well with splitting those tests in a separate file, if possible.

PR #1005, opened 2022-12-16
Return grapheme cluster count from `%str_length%` primitive

### LINKED ISSUE #1003: `string.length` returns the number of bytes, not length of a string

matthew-healy:
Previously `%str_length%` was returning the number of bytes in the Rust `String`, which lead to unexpected results.

This commit uses the `unicode-segmentation` crate to instead count the number of extended grapheme clusters in the string. This means that even complex emoji comprised of multiple characters, but visually rendering as one, get counted as just one character.

One potential downside to this is that this isn't consistent with how other standard library functions (such as splitting a string) currently work, however it also wasn't consistent before, and this change at least moves things in the right direction.

Merging this will close #1003. (cc @euank)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,6 +1216,7 @@ dependencies = [
  "test-generator",
  "toml",
  "typed-arena 2.0.1",
+ "unicode-segmentation",
  "void",
  "wasm-bindgen",
 ]
@@ -2388,9 +2389,9 @@ dependencies = [
 
 [[package]]
 name = "unicode-segmentation"
-version = "1.9.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e8820f5d777f6224dc4be3632222971ac30164d4a258d595640799554ebfd99"
+checksum = "0fdbf052a0783de01e944a6ce7a8cb939e295b1e7be835a1112c3b9a7f047a5a"
 
 [[package]]
 name = "unicode-width"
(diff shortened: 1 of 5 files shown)


vkleen (approved):
As a minimal change for `string.length` this looks good. Should we change the behaviour of the other string functions to suit? Specifically, should `string.substring` work with grapheme cluster counts as well?

matthew-healy:
@vkleen Yep, I think we should update them too. I'm opening an issue about that right now!

ebresafegaga (approved):
@matthew-healy  Is there another `length` function that returns the number of bytes? I think this might still be useful in some cases.

matthew-healy:
@ebresafegaga Not currently. To be honest I'm not really sure in what use cases a user would need to know the number of bytes a string is taking up - it seems like almost all Nickel programs should be "higher level" than that, but I'm also not against adding a `%str_bytes_count%` primitive either.

yannham (approved):
LGTM. As @matthew-healy said, I think we can ignore having a low level byte counting replacement for the old `string.length` for now. We can always add it later if there is a demand for that.

PR #1016, opened 2022-12-22
Symbolic strings with prefixes

vkleen:
This PR changes the syntax and semantics of symbolic strings. Instead of `s%"Hello, world!"%` a symbolic string now takes a prefix: `foo-s%"Hello, world!"%`. The intention is that consumers in Nickel library code can enforce different namespacing for symbolic strings.

Any valid identifier can precede `-s%` and the parser will interpret the result as a symbolic string. The identifier is converted to a string and, e.g., `foo-s%"Hello %{1}"` evaluates to the record
```
{
  tag = `SymbolicString,
  prefix = "foo",
  fragments = ["Hello ", 1]
}
```

### DIFF
--- a/src/parser/lexer.rs
+++ b/src/parser/lexer.rs
@@ -42,7 +42,7 @@ fn symbolic_string_prefix_and_length<'input>(
         .expect("The logos regexp ensures this succeeds");
     SymbolicStringStart {
         prefix,
-        percent_count: postfix.len(),
+        length: postfix.len(),
     }
 }
 
@@ -342,7 +342,7 @@ pub const KEYWORDS: &[&str] = &[
 #[derive(Debug, Clone, PartialEq)]
 pub struct SymbolicStringStart<'input> {
     pub prefix: &'input str,
-    pub percent_count: usize,
+    pub length: usize,
 }
 
 /// The tokens in string mode.
@@ -630,8 +630,7 @@ impl<'input> Iterator for Lexer<'input> {
             Some(Normal(
                 NormalToken::MultiStringStart(delim_size)
                 | NormalToken::SymbolicStringStart(SymbolicStringStart {
-                    percent_count: delim_size,
-                    ..
+                    length: delim_size, ..
                 }),
             )) => {
                 // for interpolation & closing delimeters we only care about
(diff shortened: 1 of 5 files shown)


src/parser/lexer.rs
```diff
@@ -160,8 +169,8 @@ pub enum NormalToken<'input> {
     Underscore,
     #[regex("m(%+)\"", |lex| lex.slice().len())]
     MultiStringStart(usize),
-    #[regex("s(%+)\"", |lex| lex.slice().len())]
-    SymbolicStringStart(usize),
+    #[regex("[a-zA-Z][_a-zA-Z0-9-']*-s(%+)\"", symbolic_string_prefix_and_length)]
+    SymbolicStringStart((&'input str, usize)),
```
vkleen, line 173: This is a bit weird, but logos does not support enum variants with more than one argument. And I didn't feel like making a new record for this single usecase.

yannham: It's a detail, but suing a struct for a tuple doesn't sound too bad. If only for the explicit naming of arguments, which is self-documenting.

src/parser/lexer.rs
```diff
@@ -342,7 +342,7 @@ pub const KEYWORDS: &[&str] = &[
 #[derive(Debug, Clone, PartialEq)]
 pub struct SymbolicStringStart<'input> {
     pub prefix: &'input str,
-    pub percent_count: usize,
+    pub length: usize,
```
yannham, line 345: I think we had bug earlier because it's not clear what this quantity really is. Might be useful to either document it, or to name it a bit more explicitly. It seems it's the length of the opening delimiter in total, minux the prefix?

yannham (approved):

src/parser/lexer.rs
vkleen: That's a good point and yes, that's what it is. I struggle to come up with a better name than `length`, since `percent_count` is definitely wrong. A doc comment is probably the best here.

yannham: It's well documented now. I think it's ok. Indeed `percent_count` is worst, being deceptive

PR #1017, opened 2022-12-23
Fix nix run

rapenne-s:
With the current flake, `nix run` commands aren't working. The command found on the homepage isn't working either:

```
> nix run --experimental-features "flakes nix-command" github:tweag/nickel -- repl
error: unable to execute '/nix/store/hggp49h0q7kw77i4mrj4ccz3dyxmg815-nickel-lang-0.3.1/bin/nickel-lang': No such file or directory
```

This PR instructs the flake that the default app is `bin/nickel` for package nickel. You may want to backport it to stable branch to fix the homepage example (I need to make a PR to change the command to use the stable branch there)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -350,6 +350,13 @@
         inherit stdlibDoc;
       };
 
+      apps = {
+        default = {
+          type = "app";
+          program = "${packages.nickel}/bin/nickel";
+        };
+      };
+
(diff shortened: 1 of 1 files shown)


yannham (approved):
Ah, this must have changed since the `crane` refactoring. It probably worked before because the derivation had the same name as the binary.

You can ignore the failing deploy job and merge.

PR #1024, opened 2022-12-30
Use static dictionary types for `record.fields` and `record.values`

### LINKED ISSUE #936: Improve typing of record.fields and record.values to take a dictionary type
The current typing of `record.fields` and `record.values` is the following:

```nickel
nickel> :query record.fields
• contract: { ; Dyn} -> Array Str

nickel> :query record.values
• contract: { ; Dyn} -> Array Dyn
```

The type for `record.values` is losing information, because we know that if all fields of the input record are the same (when used e.g. as a dictionary), then the output values have this type as well. But right now, everything is `Dyn`.

What's more, this type is useless in statically typed code, because there is no way to construct values of type `{; Dyn}`. Dynamically, this contract is essentially the same as `{..}`, and we should probably get rid of it. Or, a second approach could be to allow the type of the tail to be anything, and define `{_ : T} := {; T}`.

But right now, this type feels redundant, and useless statically. We propose to rather use dictionary types `{_ : T}`, which  accept some values.

The aforementioned functions to have the following signature:

- `fields : forall a. {_ : a} -> Array Str`
- `values: forall a. {_ : a} -> Array a`

matthew-healy:
This closes #936.

### DIFF
--- a/src/typecheck/operation.rs
+++ b/src/typecheck/operation.rs
@@ -129,15 +129,24 @@ pub fn get_uop_type(
         }
         // This should not happen, as ChunksConcat() is only produced during evaluation.
         UnaryOp::ChunksConcat() => panic!("cannot type ChunksConcat()"),
-        // BEFORE: forall rows. { rows } -> Array
-        // Dyn -> Array Str
-        UnaryOp::FieldsOf() => (
-            mk_uniftype::dynamic(),
-            //mk_uty_record!(; TypeWrapper::Ptr(state.table.fresh_type_var_id())),
-            mk_uniftype::array(TypeF::Str),
-        ),
-        // Dyn -> Array Dyn
-        UnaryOp::ValuesOf() => (mk_uniftype::dynamic(), mk_uniftype::array(TypeF::Dyn)),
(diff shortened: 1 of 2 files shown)


vkleen (approved):

src/typecheck/operation.rs
```diff
@@ -129,15 +129,24 @@ pub fn get_uop_type(
...
-        UnaryOp::FieldsOf() => (
-            mk_uniftype::dynamic(),
-            //mk_uty_record!(; TypeWrapper::Ptr(state.table.fresh_type_var_id())),
-            mk_uniftype::array(TypeF::Str),
-        ),
-        // Dyn -> Array Dyn
-        UnaryOp::ValuesOf() => (mk_uniftype::dynamic(), mk_uniftype::array(TypeF::Dyn)),
+        // forall a. { _: a } -> Array Str
+        UnaryOp::FieldsOf() => {
+            let ty_a = UnifType::UnifVar(state.table.fresh_type_var_id());
```
yannham, line 134: You can use directly `state.table.fresh_type_uvar()` instead here

PR #1025, opened 2022-12-30
Assign the correct position to a linearized function term 

ebresafegaga:
Currently, the linearization algorithm assumes that the position of a function is the position of the its argument. For example, if we have the term `fun x => x`, the algorithm assigns the position of this function to be the position of the argument `x`  in the linearization data structure.

This is *obviously* not right: the position of a function should span the whole range of the function, not just its arguments. 

This causes problems (e.g LSP server crash because of wrong assumptions or wrong results to the LSP client)  with some of the LSP features (e.g hover, cross-file auto-completion, etc)

This change fixes this issue.

### DIFF
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -210,11 +210,16 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                         index: id_gen.get_and_advance(),
                     };
                     self.env.insert(ident.to_owned(), id);
+                    let pos = match term {
+                        Term::LetPattern(..) => ident.pos,
+                        Term::FunPattern(..) => pos,
+                        _ => unreachable!(),
+                    };
                     lin.push(LinearizationItem {
                         env: self.env.clone(),
                         id,
                         ty,
-                        pos: ident.pos,
+                        pos,
                         kind: TermKind::Declaration(ident.to_owned(), Vec::new(), value_ptr),
                         meta: None,
                     });
@@ -283,6 +288,11 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                         index: id_gen.get(),
                     },
                 );
+                let pos = match term {
+                    Term::Let(..) => ident.pos,
+                    Term::Fun(..) => pos,
+                    _ => unreachable!(),
+                };
                 lin.push(LinearizationItem {
                     env: self.env.clone(),
                     id: ItemId {
@@ -290,7 +300,7 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                         index: id_gen.get(),
                     },
                     ty,
-                    pos: ident.pos,
+                    pos,
                     kind: TermKind::Declaration(ident.to_owned(), Vec::new(), value_ptr),
                     meta: self.meta.take(),
                 });
@@ -423,6 +433,7 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                 let locator = (*file, start);
 
                 let Some(term_id) = linearization.item_at(&locator) else {
+                    panic!("hey");
                     return
                 };
                 let term_id = term_id.id;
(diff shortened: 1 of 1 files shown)


vkleen (approved):

lsp/nls/src/linearization/mod.rs
```diff
@@ -210,12 +210,27 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
+                            ),
+                        ),
+                        _ => unreachable!(),
+                    };
                     lin.push(LinearizationItem {
                         env: self.env.clone(),
                         id,
                         ty,
-                        pos: ident.pos,
-                        kind: TermKind::Declaration(ident.to_owned(), Vec::new(), value_ptr),
```
yannham, line 218: Are you sure this code was handling the function item itself? It seems it added a `TermKind::Declaration` item in the linearization. Something used e.g. for the goto definition feature, which I believe is different from the linearized term itself. In particular, in the case of the let-binding, you still use the position of the indent, which is weird.

Wouldn't the function be linearized when the body is walked/typechecked?

ebresafegaga: Yes it was. It represents the function argument as a `TermKind::Declaration(..)`, so it can be used in the body. 

> In particular, in the case of the let-binding, you still use the position of the indent, which is weird.

Maybe, but I guess this is done because when you hover on the `ident` you get the type and other information, and  we don't want this information if we hover on *any* other part of the let binding. I think this was the rationale for using the ident's position for the function's position. For the function's case, I'm thinking we can create a `TermKind` for the `ident` , and another for the function itself.

ebresafegaga: >So the previous code was kinda correct, isn't it? What are the cases that are problematic? When you try to hover before the body, like on the `=>` ?

Yes, hovering on the `=>` would be the problematic case. Or maybe the `fun` keyword.

>But then, this is the same thing with the function? You don't want to get the declaration info when hovering on the body, do you ? I might be misunderstanding something, but from what I understand, shouldn't we rather keep the declaration item, and add before that an item for the whole function, if that's what's missing?

Yes, we actually keep the declaration item for both cases, the only thing we change is the value the declaration item points to for the function case. Previously, for the function case, we create a declaration item that points back to itself; I changed it to have an unknown value.

yannham: I see, thanks for the clarification. I'm still not sure to understand what's the answer to this question though:

> shouldn't we rather keep the declaration item, and add before that an item for the whole function, if that's what's missing?

ebresafegaga: Yes, exactly. I was even considering creating a separate `TermKind` for functions.

yannham: So should we do that instead? And keep the term declaration position to be the one of the variable? Because it seems a bit strange to affect the position of the whole function for the declaration. The declaration usually correspond to the introduction of an identifier

yannham: For example, currently, when you hover on the parameter of a function within a typed block, you get the type of this argument, which is arguably useful. Is this behavior preserved by this PR?

ebresafegaga: > So should we do that instead? And keep the term declaration position to be the one of the variable? Because it seems a bit strange to affect the position of the whole function for the declaration. The declaration usually correspond to the introduction of an identifier

Yes, we should do that. But instead of a new term kind for functions, we can reuse `TermKind::Structure` , which represents an arbitrary term.

ebresafegaga: > For example, currently, when you hover on the parameter of a function within a typed block, you get the type of this argument, which is arguably useful. Is this behavior preserved by this PR?

Currently, it doesn't preserve this. But with the suggestion above, it preserves it.

yannham: Yes, this sounds good. Do you want to fix that here or in a new PR?

ebresafegaga: I'll fix it here

yannham (approved):
Now the changes looks more natural logically speaking and we can hover either on function arguments, or on the full function, giving the associated types :+1:

PR #1028, opened 2023-01-03
[Fix] Do not panic on type path mismatch

### LINKED ISSUE #1021: panic label::span when doing something wrong
**Describe the bug**

nickel is crashing instead of handling an error

**To Reproduce**

Create `lib.ncl` with
```
let system-users = Array Num in
import "users.ncl" | system-users
```

Create `users.ncl` with
```
["a"]
```

run
```
> RUST_BACKTRACE=1 nickel -f lib.ncl
thread 'main' panicked at 'label::span: unexpected type system-users with path element Some(Array)', src/label.rs:195:27
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: nickel_lang::label::ty_path::span
   3: nickel_lang::error::blame_error::report_ty_path
   4: <nickel_lang::error::EvalError as nickel_lang::error::ToDiagnostic<codespan::file::FileId>>::to_diagnostic
   5: <nickel_lang::error::Error as nickel_lang::error::ToDiagnostic<codespan::file::FileId>>::to_diagnostic
   6: nickel_lang::program::Program<EC>::report
   7: nickel::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**Expected behavior**

nickel should tell me something is wrong

**Environment**
 - NixOS 22.11
 - Nickel 0.3.1

yannham:
Closes #1021 

Before, the type path information encoded in a label and the original type/contract annotation had to match, by virtue of the Nickel implementation. We panicked if not the case.

Typically, if the implementation of builtin contracts called to `%go_dom%`, `%go_codom%` and `%go_array%` in this order, giving a type path `[Dom, Codom, Array]`, we expected that the original contract annotation must have the shape `(_ -> Array _) -> _`. 

Since the addition of `contract.apply` and the possibility of mixing types and terms, this is not true anymore. 
For example, if we set `let Alias = Num -> Array Num -> Num` and use `Alias` in place of the original type: the type path will be unchanged, but the original annotation is now just `Alias`.

This PR relaxes the invariant expected by the error reporting code to avoid panicking in valid cases, and adapt the error reporting messages accordingly.

## TODO

The error reporting may be slightly off now when using aliases:

```nickel
nickel> let Foo = Array Num in ["a"] | Foo
  ┌─ :1:1
  │
1 │ Foo
  │ --- expected type
  │
  ┌─ repl-input-0:1:25
  │
1 │ let Foo = Array Num in ["a"] | Foo
  │                         ^^^ applied to this expression
  │
  │
1 │ "a"
  │ --- evaluated to this value

note: 
  ┌─ repl-input-0:1:32
  │
1 │ let Foo = Array Num in ["a"] | Foo
  │                                ^^^ bound here
```

The error reporting points at `Foo`, but points to the specific value inside the array which is failing the contracts. The latter is more precise, and probably what we want, but the overall error might be deceiving. We're saying that there is a problem with a `Foo = Array Num` contract and points to a string value, which can be interpreted as we're using a string in place of an array, while we're really using a string in place of a number.

Compare with the version without alias:

```nickel
["a"] | Array Num
  ┌─ :1:7
  │
1 │ Array Num
  │       --- expected array element type
  │
  ┌─ repl-input-1:1:2
  │
1 │ ["a"] | Array Num
  │  ^^^ applied to this expression
  │
  │
1 │ "a"
  │ --- evaluated to this value

note: 
  ┌─ repl-input-1:1:9
  │
1 │ ["a"] | Array Num
  │         ^^^^^^^^^ bound here
```

I don't have a great solution right now. In the meantime, this PR still strictly improves on the previous situation (it just handles more cases instead of panicking, and the cases already supported on the current master should be identical).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1216,7 +1216,7 @@ mod blame_error {
         } = ty_path::span(l.path.iter().peekable(), &l.types);
 
         let (msg, notes) = match (last, last_arrow_elem) {
-            // The type path contract doesn't contain any arrow, and the failing subcontract is the
+            // The type path doesn't contain any arrow, and the failing subcontract is the
             // contract for the elements of an array
             (Some(ty_path::Elem::Array), None) => (String::from("expected array element type"), Vec::new()),
             // The type path doesn't contain any arrow, and the failing subcontract is the contract
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,9 +11,12 @@ use lalrpop_util::ErrorRecovery;
 use crate::{
     eval::callstack::CallStack,
     identifier::Ident,
-    label::{self, ty_path},
-    parser,
+    label::{
+        self,
+        ty_path::{self, PathSpan},
+    },
     parser::{
+        self,
         error::{LexicalError, ParseError as InternalParseError},
         lexer::Token,
         utils::mk_span,
@@ -778,7 +781,7 @@ impl ToDiagnostic<FileId> for EvalError {
 
                 diagnostics.push(blame_error::note(l));
 
-                if ty_path::is_only_codom(&l.path) {
+                if ty_path::has_no_dom(&l.path) {
                 } else if let Some(id) = contract_id {
                     diagnostics.extend_with_call_stack(id, call_stack);
                 }
@@ -1055,7 +1058,7 @@ impl ToDiagnostic<FileId> for EvalError {
 
                 diagnostics.push(blame_error::note(l));
 
-                if ty_path::is_only_codom(&l.path) {
+                if ty_path::has_no_dom(&l.path) {
                 } else if let Some(id) = contract_id {
                     diagnostics.extend_with_call_stack(id, call_stack);
                 }
@@ -1073,7 +1076,10 @@ mod blame_error {
 
     use crate::{
         eval::callstack::CallStack,
-        label::{self, ty_path},
+        label::{
+            self,
+            ty_path::{self, PathSpan},
+        },
         position::TermPos,
     };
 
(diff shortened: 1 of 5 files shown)


matthew-healy (approved):
This seems like a sensible fix, (and it's broadly similar to what I did when I spiked the solution myself though, I have to say, significantly cleaner :sweat_smile:).

Ideally I think we'd have some tests for this, Currently, unless I'm mistaken, that would involve writing quite granular unit tests against the error formatting code, since our integration tests never actually try to render the errors they return, but I'm not sure actually writing those tests is a great use of time... I wonder if we could use something like snapshot testing to capture the current error for a selection of programs and then test against that. :thinking: 

Either way, not something I think should block this PR, as it's definitely better to just not have this panic even without tests. :grin:

src/label.rs
```diff
@@ -58,28 +58,82 @@ pub mod ty_path {
...
 
-    /// Determine if the path has only `Codomain` components.
-    pub fn is_only_codom(p: &Path) -> bool {
-        p.iter().all(|elt| *elt == Elem::Codomain)
+    /// Determine if the path has no `Domain` arrow component in it.
+    pub fn has_no_dom(p: &Path) -> bool {
+        !p.iter().any(|elt| matches!(*elt, Elem::Domain))
+    }
+
+    /// Return the last element of the past which is either `Domain` or `Codomain`.
```
vkleen, line 66:
```suggestion
    /// Return the last element of the path which is either `Domain` or `Codomain`.
```

src/label.rs
```diff
@@ -58,28 +58,82 @@ pub mod ty_path {
...
+    #[derive(Clone, Copy, Debug)]
+    /// Represent the span of a type path in the string representation of the corresponding type,
+    /// together with additional data useful to error reporting.
+    ///
+    /// Produced by [`span`]. The last element of the path that could be matched with the type is
+    /// returned, as well as the last element of the part of the path that could be matched which
+    /// is an arrow element (`Domain` or `Codomain`).
+    ///
+    /// In the general case, this should be respectively the last element of the path and the last
+    /// element of the path filtered by keeping only `Domain` and `Codmain`. But sometimes the
```
vkleen, line 89:
```suggestion
    /// element of the path filtered by keeping only `Domain` and `Codomain`. But sometimes the
```

vkleen (approved):
This looks like a sensible fix. I've just found a few typos 👍

src/error.rs
```diff
@@ -1202,56 +1208,61 @@ mod blame_error {
...
+            // The original contract contains an arrow, and the path is only composed of codomains.
+            // Then polarity is necessarily true and the cause of the blame is the return value of
+            // the function
+            (Some(_), Some(ty_path::Elem::Codomain)) if ty_path::has_no_dom(&l.path) => {
+                (
+                    String::from("expected return type"),
+                    vec![
+                        String::from(
+                            "This error may happen in the following situation:
+    1. A fun    ction `f` is bound by a contract: e.g. `Bool -> Num`.
```
yannham, line 1234:
```suggestion
    1. A function `f` is bound by a contract: e.g. `Bool -> Num`.
```

src/error.rs
```diff
@@ -1202,56 +1208,61 @@ mod blame_error {
...
+    2. `f` r    eturns a value of the wrong type: e.g. `f = fun c => \"string\"` while `Num` is expected.",
```
yannham, line 1235:
```suggestion
    2. `f` returns a value of the wrong type: e.g. `f = fun c => \"string\"` while `Num` is expected.",
```

ebresafegaga (approved):

yannham:
> Ideally I think we'd have some tests for this, Currently, unless I'm mistaken, that would involve writing quite granular unit tests against the error formatting code, since our integration tests never actually try to render the errors they return, but I'm not sure actually writing those tests is a great use of time... I wonder if we could use something like snapshot testing to capture the current error for a selection of programs and then test against that. 

Ah, yes, I forgot to add at least some regression tests that the problematic cases don't panic anymore. As of testing the error message themselves, it's probably a bit early to do so as it's tedious and very sensitive to changes (we will most probably rework them for 1.0). But it's definitely something that we could do at some point.

src/error.rs
```diff
@@ -1202,56 +1208,61 @@ mod blame_error {
...
-                        "This error may happen in the following situation:
+        let PathSpan {
+            start,
+            end,
+            last,
+            last_arrow_elem,
+        } = ty_path::span(l.path.iter().peekable(), &l.types);
+
+        let (msg, notes) = match (last, last_arrow_elem) {
+            // The type path contract doesn't contain any arrow, and the failing subcontract is the
```
yannham, line 1219:
```suggestion
            // The type path doesn't contain any arrow, and the failing subcontract is the
```

matthew-healy (commented):
I left a comment on the test commit but for some reason it isn't rendering on the PR... it's [here](https://github.com/tweag/nickel/commit/aa065a4d269439ec01c7c3aba2f293f303d2d3c7#r94900170).

matthew-healy:
> As of testing the error message themselves, it's probably a bit early to do so as it's tedious and very sensitive to changes (we will most probably rework them for 1.0). But it's definitely something that we could do at some point.

It's less tedious with proper snapshot testing capabilities. By which I mean: we have two modes we can run a given test in, `Capture` or `Verify`. When we run the test in `Capture` mode, it updates an expectation file with the output of the Nickel binary given some input, and when we run them in `Verify` mode, it checks the returned value against the saved expectation.

That way, if we make a change that we know has updated the error output, we simply run the relevant test(s) in `Capture` mode and commit the results.

This is a pretty common way to write tests against UI in the mobile & web development worlds, where the snapshots are generally images or HTML dumps. UI-intensive apps are even more sensitive to minor changes than PL error output, but it still works pretty well as a check that (1) things don't crash and (2) you didn't accidentally change something you didn't intend to.

PR #1029, opened 2023-01-03
Improve the "goto definition" feature of the LSP by making it work across multiple files

### LINKED ISSUE #932: Make goto reference work across files

ebresafegaga:
Closes #932 

This change adds support for cross file navigation using the "goto reference" feature of the LSP.

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -32,8 +32,12 @@ impl<'b> Building<'b> {
         self.linearization.push(item)
     }
 
-    fn get_item_kind_with_id(&self, file: FileId, id: ItemId) -> Option<(&ItemId, &TermKind)> {
-        if file == id.file_id {
+    fn get_item_kind_with_id(
+        &self,
+        current_file: FileId,
+        id: ItemId,
+    ) -> Option<(&ItemId, &TermKind)> {
+        if current_file == id.file_id {
             // This usage references an item in the file we're currently linearizing
             let item = self.linearization.get(id.index)?;
             Some((&item.id, &item.kind))
@@ -47,13 +51,13 @@ impl<'b> Building<'b> {
         }
     }
 
-    fn get_item_kind(&self, file: FileId, id: ItemId) -> Option<&TermKind> {
-        let (_, kind) = self.get_item_kind_with_id(file, id)?;
+    fn get_item_kind(&self, current_file: FileId, id: ItemId) -> Option<&TermKind> {
+        let (_, kind) = self.get_item_kind_with_id(current_file, id)?;
         Some(kind)
     }
 
-    fn get_item_kind_mut(&mut self, file: FileId, id: ItemId) -> Option<&mut TermKind> {
-        if file == id.file_id {
+    fn get_item_kind_mut(&mut self, current_file: FileId, id: ItemId) -> Option<&mut TermKind> {
+        if current_file == id.file_id {
             // This usage references an item in the file we're currently linearizing
             let item = self.linearization.get_mut(id.index)?;
             Some(&mut item.kind)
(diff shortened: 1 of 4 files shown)


lsp/nls/src/linearization/building.rs
```diff
@@ -120,40 +139,37 @@ impl<'b> Building<'b> {
...
+            _ => Some(item),
+        }
     }
 
     pub(super) fn resolve_record_references(
         &mut self,
         file: FileId,
         mut defers: Vec<(ItemId, ItemId, Ident)>,
-    ) {
+    ) -> Vec<(ItemId, ItemId, Ident)> {
```
yannham, line 172: What is this new return value doing?

lsp/nls/src/requests/goto.rs
```diff
@@ -44,20 +44,22 @@ pub fn handle_to_definition(
...
+    };
 
     debug!("found referencing item: {:?}", item);
 
     let location = match item.kind {
         TermKind::Usage(UsageState::Resolved(usage_id)) => {
             let definition = linearization.get_item(usage_id, &server.lin_cache).unwrap();
+            if server.cache.is_stdlib_module(definition.id.file_id) {
+                // The standard library files are embeded in the executable,
+                // so we can't possibly go their definition on disk.
```
yannham, line 59:
```suggestion
                // so we can't possibly go to their definition on disk.
```

lsp/nls/src/linearization/building.rs
```diff
@@ -32,6 +32,26 @@ impl<'b> Building<'b> {
...
+                .get_item(id, self.lin_cache)?;
+            Some((&item.id, &item.kind))
+        }
+    }
+
+    fn get_item_kind(&self, file: FileId, id: ItemId) -> Option<&TermKind> {
+        let (_, kind) = self.get_item_kind_with_id(file, id)?;
+        Some(kind)
+    }
+
```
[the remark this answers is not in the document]
ebresafegaga: >Would it make sense to rather have a `get_item` and `get_item_mut` functions, as in `Completed`, rather than those specialized versions? 

That would have been better, but the problem with it is that, with `Building`, we're not yet done linearizing the current file, so the linearization items in `Building` have the type `LinearizationItem<UnifType>`, while the linearization items in the other files would have the type `LinearizationItem<Types>`, so that will lead to a return type mismatch from the function. I noticed that we only need the `TermKind` (and in some cases, the `ItemId`), so this was a nice way to avoid using the whole struct regardless of the generic parameter.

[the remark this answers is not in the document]
ebresafegaga: >Then the caller could just get the item and decide what to do with it. Here it feels a bit like we are just instantiating a get_item function with some arbitrary post-processing.

Right. It's just a convenience, because we know the caller does just one thing it. But we can change it to avoid the "post-processing"

lsp/nls/src/linearization/building.rs
```diff
@@ -32,6 +32,26 @@ impl<'b> Building<'b> {
         self.linearization.push(item)
     }
 
+    fn get_item_kind_with_id(&self, file: FileId, id: ItemId) -> Option<(&ItemId, &TermKind)> {
```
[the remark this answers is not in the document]
ebresafegaga: Yes, thank you!

lsp/nls/src/linearization/building.rs
ebresafegaga: It returns a list of items that could not be resolved.

yannham: I see. Then it might be valuable to document this function, and in particular what the return value is.

yannham (approved):
Fair enough for `get_item_kind_XXX`, if it's used a lot. Do as you see fit, but here is a suggestion though: split the part that fetches the items in separate functions `get_item` and `get_item_mut`, and use them from within `get_item_kind_xxx`. Doing so, we've not added much code (juste moved some parts in dedicated functions), function definitions are smaller, and if we need to get an item for another reason, the methods to do so are already in place.

Issue #820, opened 2022-09-08
Python bindings


GuillaumeDesforges:
**Is your feature request related to a problem? Please describe.**
I want to read Nickel files as configuration to my Python program.

**Describe the solution you'd like**
I would like a very classical `nickel.load(file)` or `nickel.loads(text)` Python function which evaluates the Nickel file as Python primitives.

**Describe alternatives you've considered**
Calling the Nickel CLI and parsing the output, but it's not the best performance wise and its harder to integrate the possible failures.

PR #1036, opened 2023-01-09
Python bindings

GuillaumeDesforges:
This PR adds very simple Python bindings. Fixes https://github.com/tweag/nickel/issues/820.

It builds a package `nickel-lang` with a module `nickel_lang` containing a single function `run` which takes in a `str` of a Nickel expression and returns the result as `str`.

The Python module is built using [maturin](https://github.com/PyO3/maturin).
Note that building the Python modules part is Rust is optional, the binary shouldn't be bigger.

To test this PR:
- go to `pyckel` folder
- create a Python virtual environment and activate it
  ```shell
  python -m venv .venv
  source .venv/bin/activate
  ```
- install maturin
  ```shell
  pip install "maturin>=0.14,<0.15"
  ```
- build and install the `nickel_lang` Python module to the local venv
  ```shell
  maturin develop
  ```
- test with Python
  ```shell
  $ python
  >>> import pyckel
  >>> pyckel.run("let x = 1 in x + 3")
  '4'
  ```

This PR does not
- implement proper Python exceptions: all Nickel errors are just dumped as their Debug formatted string to a NickelException

Todo:
- [x] `run` should return JSON
- [x] `run` should evaluate fully


Cargo.toml
```diff
@@ -17,13 +17,16 @@ bench = false
 
 [lib]
 bench = false
+name = "nickel_lang"
+crate-type = ["cdylib", "lib"]
```
yannham, line 21: I don't know if we want to make the main Nickel crate a dynamic C library by default. My current understanding is that it has implications (e.g. if you're trying to use Nickel from another Rust crate, see e.g. [this random issue][https://github.com/EmbarkStudios/rust-gpu/issues/367) or [this post](https://users.rust-lang.org/t/what-is-the-difference-between-dylib-and-cdylib/28847)). This is also a problem when compiling the same crate to both a normal Rust library and to WASM, see https://github.com/rust-lang/cargo/issues/4881.

One simple possibility is to make a very thin wrapper around `nickel-lang`, which is essentially just a crate re-exporting Nickel as it is but with a different library type. This is exactly what we do for the WASM REPL: see the `nickel-wasm-repl` directory. It could make sense to have a `nickel-lang-python` (or whatever better name you have) as well, that could be a `cdylib` without issues. You wouldn't need to have a `python-module` feature either.

src/python.rs
```diff
@@ -0,0 +1,31 @@
...
+            // TODO better exceptions
+            error => NickelException::new_err(format!("unexpected error: {error:?}")),
+        }
+    }
+}
+
+// Some test function
+#[pyfunction]
+pub fn run(s: String) -> PyResult<String> {
+    let mut program: Program<CBNCache> =
```
yannham, line 21: FYI, if you want to serialize to JSON instead, here is the code of the main binary that does that: https://github.com/tweag/nickel/blob/bff8451f7e8e860859070e6f4f10221f79bef1f2/src/bin/nickel.rs#L222.

There is not much more: use `eval_full()` instead of `eval` (which substitutes the strange `%123` variables for their actual value), valide the result (check if there isn't any function or other non-serializable expression remaining), and then use `serialize::to_xxx`. In particular there is a `serialize::to_string` function.

GuillaumeDesforges: Thanks a lot! I could implement it in a454f9a1 💯

Cargo.toml
GuillaumeDesforges: Addressed in 1ecfdf1c

yannham (commented):
LGMT overall :+1:

The CI is failing, presumably because now building the dependencies of the workspace implies to build `pyo3` which requires python. I guess it's fine to do so in the checks, but not when just building e.g. Nickel.

But that may not be totally straightforward to achieve right now, not because of any Nix or Crane limitations, but because of how the flake is organized. It seems we always builds the dependencies of the whole workspace before anything else.

pyckel/README.md
```diff
@@ -0,0 +1,21 @@
+# pyckel
+
+Python bindings to use Nickel.
+
+## Install
+
+```shell
+pip install pyckel
```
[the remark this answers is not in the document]
GuillaumeDesforges: Not yet indeed.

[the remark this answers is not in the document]
GuillaumeDesforges:
```suggestion
pip install .
```

GuillaumeDesforges:
Trying to fix the error from the CI here: https://github.com/tweag/nickel/actions/runs/3876173138/jobs/6609682802

```
nickel-lang-deps> error: failed to run custom build command for `pyo3-build-config v0.17.3`
nickel-lang-deps> Caused by:
nickel-lang-deps>   process didn't exit successfully: `/build/source/target/release/build/pyo3-build-config-bd0e0341d2a36725/build-script-build` (exit status: 1)
nickel-lang-deps>   --- stdout
...
nickel-lang-deps>   --- stderr
nickel-lang-deps>   error: no Python 3.x interpreter found
```

I can fix this issue with a patch on `flake.nix`, adding Python when building the Rust deps:

```diff
diff --git a/flake.nix b/flake.nix
index 1c6fd331..580368cc 100644
--- a/flake.nix
+++ b/flake.nix
@@ -196,6 +196,7 @@
             inherit
               src
               cargoExtraArgs;
+            buildInputs = [ pkgs.python3 ];
           };

           nickel = craneLib.buildPackage {
```

but then building he `nickel` package with Nix fails with the error

```
$ nix flake checks
error: builder for '/nix/store/hz4gvqkjf22xlhd612amzic1a07jm3n6-nickel-lang-0.3.1.drv' failed with exit code 101;
       last 10 log lines:
       > ++ command cargo build --profile release --message-format json-render-diagnostics --frozen --offline --workspace
       >    Compiling pyo3-build-config v0.17.3
       > error: failed to run custom build command for `pyo3-build-config v0.17.3`
       >
       > Caused by:
       >   could not execute process `/build/source/target/release/build/pyo3-build-config-bd0e0341d2a36725/build-script-build` (never executed)
       >
       > Caused by:
       >   Permission denied (os error 13)
```

GuillaumeDesforges:
I think I fixed the Nix build. I removed `--workspace` from the cargo build extra arguments, and created a derivation per package.

Note that there is no derivation for the Python bindings (pyckel) since I have not yet managed to make it work. I suggest to tackle that in another PR.

GuillaumeDesforges:
I've requested reviews from @sir4ur0n and @matthew-healy since they have worked on the Nix flake not too long ago.

flake.nix
```diff
@@ -187,29 +187,39 @@
...
-          nickel = craneLib.buildPackage {
-            inherit
-              src
-              cargoExtraArgs
-              cargoArtifacts;
-          };
+          buildPackage = packageName:
+            let
+            in
+            craneLib.buildPackage {
```
sir4ur0n, line 203:
```suggestion
            craneLib.buildPackage {
```

flake.nix
```diff
@@ -187,29 +187,39 @@
...
+          cargoBuildExtraArgs = "--frozen --offline";
+
           # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
           cargoArtifacts = craneLib.buildDepsOnly {
-            inherit
-              src
-              cargoExtraArgs;
+            inherit src;
+            cargoExtraArgs = "${cargoBuildExtraArgs} --workspace";
+            buildInputs = [ pkgs.python3 ];
```
sir4ur0n, line 197: Please add a comment explaining why we need to add `python3`

lsp/client-extension/default.nix
```diff
@@ -1,8 +1,11 @@
 # This file has been generated by node2nix 1.11.1. Do not edit!
 
-{pkgs ? import <nixpkgs> {
+{ pkgs ? import <nixpkgs> {
```
sir4ur0n, line 3: This comment applies to this file and the 2 other generated by `node2nix`: please don't format them. It's annoying because every time someone regenerates them, it leads to huge diffs, and forces to re-format those files.

See https://github.com/tweag/nickel/blob/master/flake.nix#L134-L138

sir4ur0n (commented):
I haven't reviewed actual `pyckel` part, only the Nix files

matthew-healy:
Thanks for tagging me, but I'm gonna remove myself from this PR since my experience editing the flake was pretty limited. I think Julien has a lot more context than I do on how it actually works.

flake.nix
```diff
@@ -187,29 +187,39 @@
...
+
+              cargoExtraArgs = "${cargoBuildExtraArgs} --package ${packageName}";
+            };
+
+
+        in
+        rec {
+          nickel = buildPackage "nickel-lang";
+          lsp-nls = buildPackage "nickel-lang-lsp";
+          utilities = buildPackage "nickel-lang-utilities";
```
yannham, line 216: I think we don't need to build utilities explicitly. It's just a build dependency for benchmarks and tests (that has to be taken out of the nickel crate because of how cargo, testing and crates work).

```suggestion
```

flake.nix
```diff
@@ -187,29 +187,39 @@
...
+          nickel-wasm-repl = buildPackage "nickel-repl";
 
           rustfmt = craneLib.cargoFmt {
             # Notice that unlike other Crane derivations, we do not pass `cargoArtifacts` to `cargoFmt`, because it does not need access to dependencies to format the code.
             inherit src;
 
-            # We don't reuse the `cargoExtraArgs` in scope because `cargo fmt` does not accept nor need any of `--frozen`, `--offline` or `--workspace`
```
yannham, line 212: Is this deletion on purpose?

flake.nix
```diff
@@ -391,6 +401,9 @@
       checks = {
         inherit (mkCraneArtifacts { })
           nickel
+          lsp-nls
+          utilities
```
yannham, line 405:
```suggestion
```

yannham (approved):
Mostly LGTM. The CI running time is suspiciously high, but you've changed the `nickel-lang-deps` derivation, which is the cornerstone of the build, so it might just be that this PR invalidated all previous cached artifacts. I took another look at the `crane` doc, and the current PR, and I don't see why the artifacts wouldn't be cached again in the future.

Let's see in subsequent PRs if something looks wrong with respect to CI time.

Bringing python in is surprising for people who don't care about python bindings, suggesting that said bindings could also be in their own repo. But I guess Python is a well-cached, not-too-big derivation, and it's fine as it is for now.

flake.nix
GuillaumeDesforges: Yes, I've changed how `cargoExtraArgs`s are handled in the file

flake.nix
GuillaumeDesforges:
```suggestion
            # pyo3 needs a Python interpreter in the build environment
            # https://pyo3.rs/v0.17.3/building_and_distribution#configuring-the-python-version
            buildInputs = [ pkgs.python3 ];
```

sir4ur0n (approved):
Approving as in "Nothing looks off in the Nix code". I haven't tried it, nor looked at the Python/Rust part

yannham:
Ok, let's keep an eye on CI times, but otherwise LGTM.

PR #1037, opened 2023-01-11
Make the "find references" feature of the LSP work across multiple files

ebresafegaga:
This is a continuation of the work started in #1029 

Depends on #1029


lsp/nls/src/linearization/mod.rs
```diff
@@ -121,11 +121,13 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
-                    pos.end = ByteIndex(0);
-                    pos
-                });
+                if pos == TermPos::None {
+                    pos = field_pos.map(|mut x| {
+                        x.start = ByteIndex(u32::MAX);
+                        x.end = ByteIndex(u32::MAX);
+                        x
+                    });
+                }
```
yannham, line 130: Why does this change fix the issue? It's not very clear to me. May I suggest to add something in the PR's description?

Side question, why do we need to fill a dummy position? I think we can now support terms without position. Maybe this code dates from before such support. How does it work if we just keep the `TermPos::None` value?

lsp/nls/src/linearization/mod.rs
```diff
@@ -141,7 +143,6 @@ impl<'a> Linearizer for AnalysisHost<'a> {
                     };
                     match field.kind {
                         TermKind::RecordField { ref mut value, .. } => {
-                            pos = field.pos;
```
yannham, line 144: Will this work as before with goto references for record fields?

yannham (commented):
I tested this on the mantis benchmark, and it didn't work as expected. Maybe it's something else (as the mantis benchmark is not trivial), but if you go to `benchmarks/mantis/jobs/mantis.ncl` L44 and goto definition for `job` in `types.stanza.job`, this correctly goes to `schemas/nomad/types.ncl` L684. Now, if I do goto references on the same job, I only get the references inside the current file, and not the one I came from.

ebresafegaga:
> I tested this on the mantis benchmark, and it didn't work as expected. Maybe it's something else (as the mantis benchmark is not trivial), but if you go to `benchmarks/mantis/jobs/mantis.ncl` L44 and goto definition for `job` in `types.stanza.job`, this correctly goes to `schemas/nomad/types.ncl` L684. Now, if I do goto references on the same job, I only get the references inside the current file, and not the one I came from.

Right. I hadn't noticed this. It turns out that if you open `schemas/nomad/types.ncl` first, and then open `benchmarks/mantis/jobs/mantis.ncl`, and then go back to `schemas/nomad/types.ncl`, the find reference works. Does this work on your side? I think this is a very subtle bug, I'll start working on this now.

lsp/nls/src/linearization/mod.rs
ebresafegaga: First of all, we only want to add dummy position when the current position is a `TermPos::None`; the former code assigns this dummy position to every term's position (regardless of if it is a `TermPos::None` or not). This assigns the wrong position to nested record field's value. To avoid this, this was added on line 144: `pos = field.pos` (It assigns the position of a nested record field's value to be the same as the nested record field). However, we won't need to do this if we only give a dummy position to terms which have a position of `TermPos::None`.

Now, the problem with giving a record field (i.e the identifier) and it's value the same position is that when we want to find references at the position of a record field, we do not get the linearization item for the record field, but we get the linearization item for the record's field's value; so they should not be the same.

And finally, the reason why the dummy value was changed from `(0,0)` to `(max, max)` is that if we have another item at say, `(0, 9)`, and we want to get the item at `(0,9)` with `linearization.item_at((0,9))`, somehow, we get the item at `(0,0)`, so I think, `(max, max)` would be safer.

ebresafegaga: >Side question, why do we need to fill a dummy position? I think we now support terms without position. Maybe this code dates back from before such support. How does it work if we just keep the TermPos::None value?

There's [this](https://github.com/tweag/nickel/blob/master/lsp/nls/src/linearization/mod.rs#L170), which doens't allow terms with `TermPos::None` to be linearized. Apart from that, [this](https://github.com/tweag/nickel/blob/master/lsp/nls/src/requests/symbols.rs#L30) expects terms to have a non-`TermPos::None` position.

yannham:
> Right. I hadn't noticed this. It turns out that if you open schemas/nomad/types.ncl first, and then open benchmarks/mantis/jobs/mantis.ncl, and then go back to schemas/nomad/types.ncl, the find reference works. Does this work on your side? I think this is a very subtle bug, I'll start working on this now.

Indeed, this works :+1: can be fixed in subsequent PR then.

lsp/nls/src/linearization/mod.rs
yannham: Hmm. Setting a position offset that is out of the file sounds wrong, I wonder if it could backfire (e.g. when doing error reporting or so). Maybe not, I guess the position inside the linearization is just used to find an item given a position. But I was convinced that the linearizer was able to handle stuff without position, by augmenting the ordering relation to `TermPos::None`. I'm surprised to see that in fact it can't.

Could we set the position of elaborated records correctly, as discussed during the daily? Could we also make the code work even if some position is set to `None`.

yannham:
@ebresafegaga what do you think are the next actions for this PR? Merge as it is, and fix the missing position directly in the parser? Or fix the position first in a separate PR, and then we could even ignore stuff without position at this point?

ebresafegaga:
@yannham I suggest we merge as it is, and then fix the the position in a separate pull request.

yannham (approved):
Let's see if this `MAX` value cause any harm, or if we can remove it late. For now it seems to strictly improve on the current situation.

ebresafegaga:
@yannham Now that [this](https://github.com/tweag/nickel/pull/1046) is merged, should we rebase on master, and then remove the dummy positions?

yannham:
@ebresafegaga sounds good :+1:

PR #1038, opened 2023-01-11
DumbCache experiments: Fixed `Thunk` leakage

dpl0a:


src/eval/merge.rs
```diff
@@ -623,10 +647,15 @@ fn merge_closurize<C: Cache>(
...
 /// Revert the thunk inside the provided field (if any), and closurize the result inside `env`.
-fn revert_closurize(rt: RichTerm, env: &mut Environment, local_env: &Environment) -> RichTerm {
+fn revert_closurize<C: Cache>(
+    cache: &mut C,
+    rt: RichTerm,
+    env: &mut Environment,
+    local_env: &Environment,
+) -> RichTerm {
     if let Term::Var(id) = rt.as_ref() {
         // This create a fresh variable which is bound to a reverted copy of the original thunk
```
ebresafegaga, line 657:
```suggestion
        // This creates a fresh variable which is bound to a reverted copy of the original thunk
```

src/eval/cache.rs
```diff
@@ -1,10 +1,11 @@
...
+    Closure, Environment, IdentKind,
+};
+use crate::{
+    identifier::Ident,
+    term::{record::FieldDeps, BindingType, RichTerm},
 };
-use crate::{eval::Closure, identifier::Ident, term::BindingType};
 
 pub trait Cache: Clone {
     type UpdateIndex; // Temporary: we won't really need that once an alternative caching mechanism gets implemented
```
ebresafegaga, line 11: Is this alternative caching mechanism implemented in this pull request?

ebresafegaga (approved):
I left some comment, other than that, LGTM.

src/eval/cache.rs
dpl0a: A test version was implemented in this branch but wasn't included in the PR

src/error.rs
```diff
@@ -46,7 +46,7 @@ pub enum Error {
 #[derive(Debug, Clone, PartialEq)]
 pub enum EvalError {
     /// A blame occurred: a contract have been broken somewhere.
-    BlameError(label::Label, CallStack),
+    BlameError(Option<RichTerm>, label::Label, CallStack),
```
yannham, line 49: The structured version is not strictly necessary (we can do that later for all error variant in a pass, as discussed internally), we should at least document what is this `Option<RichTerm>`.

```suggestion
    BlameError {
        /// The argument failing the contract. If the argument has been forced by the contract, `tracked_arg` provides the final value. 
        evaluated_arg: Option<RichTerm>,
        /// The label of the corresponding contract.
        label: label::Label,
        /// The callstack when the blame error was raised.
        call_stack: CallStack,
      },
```

src/eval/cache.rs
```diff
@@ -19,13 +20,25 @@ pub trait Cache: Clone {
...
-    // TODO: Needs a better name
     fn build_cached(&mut self, idx: &mut CacheIndex, rec_env: &[(Ident, CacheIndex)]);
+    fn ident_kind(&self, idx: CacheIndex) -> IdentKind;
+    fn saturate<'a, I: DoubleEndedIterator<Item = &'a Ident> + Clone>(
+        &mut self,
+        idx: CacheIndex,
+        env: &mut Environment,
+        fields: I,
+    ) -> RichTerm;
+    fn revert(&mut self, idx: CacheIndex) -> CacheIndex;
```
yannham, line 34: `revert` is taking a immutable reference in `Thunk`. Is there any reason why it needs an owned variant here? I know in a future ideal world `CacheIndex` will be `Copy`, probably. But right now we take an immutable reference e.g. for `map_at_index`. It think it would make sense to take an immutable reference as well for `ident_kind`, `revert`, and `deps`, which don't mutate anything. That would get rid of some thunk cloning.

dpl0a: I left these from the `DumbCache` interface, which used `usize` as indices for the cache and resorted to cloning thunks around just to make it respect the interface since we talked about having something easily copy-able as indices in the future, but I'm not really attached to this choice, I'll change it

src/error.rs
```diff
@@ -46,7 +46,14 @@ pub enum Error {
 #[derive(Debug, Clone, PartialEq)]
 pub enum EvalError {
     /// A blame occurred: a contract have been broken somewhere.
-    BlameError(label::Label, CallStack),
+    BlameError {
+        /// The argument failing the contract. If the argument has been forced by the contract, `tracked_arg` provides the final value.
```
yannham, line 50:
```suggestion
        /// The argument failing the contract. If the argument has been forced by the contract, `evaluated_arg` provides the final value.
```

yannham (approved):

PR #1046, opened 2023-01-17
Assign the correct position to the value of nested record fields

ebresafegaga:
Currently, when we have a nested record, defined like this: 
```
{ foo.bar.baz.boo = 5 }
```
Nickel desugars it to: 
```
{ foo = {bar = {baz = {boo = 5}}}}
```
But the position of the value of `foo`, `bar`, and `baz` are set to `TermPos::None`

This pull request changes this by setting the position of a nested record field's value (that is defined inline), to span from the identifier of the field up until the value (represented by an expression)

For example:  
```
{ foo = <foos value> }
```
`<foos value>` would have a position which spans  `foo.bar.baz.boo = 5`

and 

```
{  ... bar = <bars value> }
```
<bars value> would have a position which spans `baz.boo = 5`

### DIFF
--- a/src/parser/utils.rs
+++ b/src/parser/utils.rs
@@ -278,50 +278,51 @@ pub fn elaborate_field_path(
     let mut it = path.into_iter();
     let fst = it.next().unwrap();
 
-    let content = it.rev().fold(content, |acc, path_elem| match path_elem {
-        FieldPathElem::Ident(id) => {
-            // unwrap is safe here because every id should have a non-`TermPos::None` position
-            let id_span = id.pos.unwrap();
-            // unwrap is safe here becuase the initial content has a position,
-            // and we make sure we assign a position for the next field.
-            let acc_span = acc.pos.unwrap();
-            // `RawSpan::fuse` only returns `None` when the two spans are in different files.
-            // A record field and its value *must* be in the same file, so this is safe.
-            let pos = TermPos::Original(RawSpan::fuse(id_span, acc_span).unwrap());
+    let content = it.rev().fold(content, |acc, path_elem| {
+        // unwrap is safe here becuase the initial content has a position,
+        // and we make sure we assign a position for the next field.
+        let acc_span = acc.pos.unwrap();
+        match path_elem {
+            FieldPathElem::Ident(id) => {
+                // unwrap is safe here because every id should have a non-`TermPos::None` position
+                let id_span = id.pos.unwrap();
+                // `RawSpan::fuse` only returns `None` when the two spans are in different files.
+                // A record field and its value *must* be in the same file, so this is safe.
+                let pos = TermPos::Original(RawSpan::fuse(id_span, acc_span).unwrap());
 
-            let mut fields = HashMap::new();
-            fields.insert(id, acc);
-
-            RichTerm::new(Term::Record(RecordData::with_fields(fields)), pos)
-        }
-        FieldPathElem::Expr(exp) => {
-            let static_access = match exp.term.as_ref() {
-                Term::StrChunks(chunks) => {
-                    chunks
-                        .iter()
-                        .fold(Some(String::new()), |acc, next| match (acc, next) {
-                            (Some(mut acc), StrChunk::Literal(lit)) => {
-                                acc.push_str(lit);
-                                Some(acc)
-                            }
-                            _ => None,
-                        })
-                }
-                _ => None,
-            };
-
-            let id_span = exp.pos.unwrap();
-            let acc_span = acc.pos.unwrap();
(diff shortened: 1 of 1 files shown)


src/parser/utils.rs
```diff
@@ -280,9 +280,19 @@ pub fn elaborate_field_path(
 
     let content = it.rev().fold(content, |acc, path_elem| match path_elem {
         FieldPathElem::Ident(id) => {
+            // unwrap is safe here because every id should have a non-`TermPos::None` position
+            let id_span = id.pos.unwrap();
+            // unwrap is safe here becuase the initial content has a position,
+            // and we make sure we assign a position for the next field.
+            let acc_span = acc.pos.unwrap();
+            // `RawSpan::fuse` only returns `None` when the two spans are in different files.
+            // A record field and its value *must* to be in the same file, so this is safe.
```
yannham, line 289:
```suggestion
            // A record field and its value *must* be in the same file, so this is safe.
```

src/parser/utils.rs
```diff
@@ -280,9 +280,19 @@ pub fn elaborate_field_path(
...
+            let pos = TermPos::Original(RawSpan::fuse(id_span, acc_span).unwrap());
+
             let mut fields = HashMap::new();
             fields.insert(id, acc);
-            Term::Record(RecordData::with_fields(fields)).into()
+
+            RichTerm::with_pos(Term::Record(RecordData::with_fields(fields)).into(), pos)
```
yannham, line 295:
```suggestion
            RichTerm::new(Term::Record(RecordData::with_fields(fields)), pos)
```

src/parser/utils.rs
```diff
@@ -300,14 +310,22 @@ pub fn elaborate_field_path(
...
-                Term::Record(RecordData::with_fields(fields)).into()
+
+                RichTerm::with_pos(Term::Record(RecordData::with_fields(fields)).into(), pos)
             } else {
                 let empty = Term::Record(RecordData::empty());
-                mk_app!(mk_term::op2(BinaryOp::DynExtend(), exp, empty), acc)
+                RichTerm::with_pos(
+                    mk_app!(mk_term::op2(BinaryOp::DynExtend(), exp, empty), acc),
+                    pos,
+                )
```
yannham, line 328:
```suggestion
                mk_app!(mk_term::op2(BinaryOp::DynExtend(), exp, empty), acc).with_pos(pos)
```

src/parser/utils.rs
```diff
@@ -300,14 +310,22 @@ pub fn elaborate_field_path(
...
+            let acc_span = acc.pos.unwrap();
+            let pos = TermPos::Original(RawSpan::fuse(id_span, acc_span).unwrap());
+
             if let Some(static_access) = static_access {
                 let id = Ident::new_with_pos(static_access, exp.pos);
                 let mut fields = HashMap::new();
                 fields.insert(id, acc);
-                Term::Record(RecordData::with_fields(fields)).into()
+
+                RichTerm::with_pos(Term::Record(RecordData::with_fields(fields)).into(), pos)
```
yannham, line 322:
```suggestion
                RichTerm::new(Term::Record(RecordData::with_fields(fields)), pos)
```

src/parser/utils.rs
```diff
@@ -300,14 +310,22 @@ pub fn elaborate_field_path(
                 _ => None,
             };
 
+            let id_span = exp.pos.unwrap();
+            let acc_span = acc.pos.unwrap();
+            let pos = TermPos::Original(RawSpan::fuse(id_span, acc_span).unwrap());
+
```
yannham, line 316: This code looks to be strictly similar to the one above. Can it be taken out of the pattern matching to avoid duplication?

yannham (approved):

src/parser/utils.rs
ebresafegaga: Hm. We need to get the `id` from the `FieldPathElem::Ident(..)`, and the `exp` from `FieldPathElem::Expr(..)` to get their positions. To do this before pattern matching , we would have to add a method to `FieldPathElem`, that returns the position or just do smaller pattern matching at the top to extract the position.

vkleen: Without a new method on `FieldPathElem`, we could at least take `acc_span` out of the pattern match.

vkleen (approved):

src/parser/utils.rs
ebresafegaga: Ah, yes

yannham: Ah, indeed, I missed that.

yannham: Duplicating the pattern matching is still slightly less code in general, but in this case the duplicated part is so tiny that it's not very interesting indeed.

PR #1048, opened 2023-01-17
Add multiline doc logic to the pretty printer

### LINKED ISSUE #1044: Pretty printer produces spurious newlines
The pretty printer inserts a spurious newline for `doc` metadata. For example,
```
❯ nickel pprint-ast <<<'{ field | doc "Hello, world" = 1 }'
{
  field | doc "
    Hello, world"
     = 1,
}
```

vkleen:
This PR makes the pretty printer emit multiline strings if it encounters documentation metadata containing newlines. Previously, as I complained about in #1044, the pretty printer would unconditionally insert a hard newline after the opening double quote. Additionally, it would indent the documentation string and thereby change the semantics of the code.

With this PR, multiline strings are used when necessary and single line strings are used whenever the docstring does not contain newlines. Additionally, there's a regression test using the new snapshot testing infrastructure.

Closes #1044

### DIFF
--- a/src/pretty.rs
+++ b/src/pretty.rs
@@ -61,13 +61,9 @@ where
     }
 
     fn multiline_string(&'a self, s: &str) -> DocBuilder<'a, Self, A> {
-        let percent_count = s.split(|c| c != '%').map(|v| v.len()).max().unwrap();
-        let delimiter = "%".repeat(percent_count + 1);
+        let delimiter = "%".repeat(min_interpolate_sign(s));
         self.hardline()
-            .append(self.intersperse(
-                s.lines().map(|d| self.text(d.to_owned())),
-                self.hardline().clone(),
-            ))
+            .append(self.intersperse(s.lines().map(|d| self.text(d.to_owned())), self.hardline()))
             .enclose(format!("m{delimiter}\""), format!("\"{delimiter}"))
     }
 
--- a/src/pretty.rs
+++ b/src/pretty.rs
@@ -60,6 +60,17 @@ where
         self.text(s)
     }
 
+    fn multiline_string(&'a self, s: &str) -> DocBuilder<'a, Self, A> {
+        let percent_count = s.split(|c| c != '%').map(|v| v.len()).max().unwrap();
+        let delimiter = "%".repeat(percent_count + 1);
+        self.hardline()
+            .append(self.intersperse(
+                s.lines().map(|d| self.text(d.to_owned())),
+                self.hardline().clone(),
+            ))
+            .enclose(format!("m{delimiter}\""), format!("\"{delimiter}"))
+    }
+
     fn metadata(&'a self, mv: &MetaValue, with_doc: bool) -> DocBuilder<'a, Self, A> {
         if let Some(types) = &mv.types {
             self.text(":")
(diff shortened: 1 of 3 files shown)


src/pretty.rs
```diff
@@ -60,6 +60,17 @@ where
         self.text(s)
     }
 
+    fn multiline_string(&'a self, s: &str) -> DocBuilder<'a, Self, A> {
+        let percent_count = s.split(|c| c != '%').map(|v| v.len()).max().unwrap();
+        let delimiter = "%".repeat(percent_count + 1);
```
vkleen, line 65: This might be too clever, but I like it 🙂

yannham: FWIW, I think it's fine :slightly_smiling_face: 

It's not the best we can do (here we count any percent sign, while for example a simple modulo alone doesn't require double delimiters). A step further would be to match on sequences we know the lexer would interpret, i.e `m%"`, `"%`, and `%{`. But at least it's correct, and it's probably ok for now. Maybe add a comment to say so? (that the counts is over-approximated)

src/pretty.rs
```diff
@@ -60,6 +60,17 @@ where
...
+        self.hardline()
+            .append(self.intersperse(
+                s.lines().map(|d| self.text(d.to_owned())),
+                self.hardline().clone(),
```
yannham, line 69: Why is this clone required? It looks odd

src/pretty.rs
```diff
@@ -60,6 +60,17 @@ where
...
+            ))
+            .enclose(format!("m{delimiter}\""), format!("\"{delimiter}"))
```
yannham, line 71: I didn't even know you could do that in Rust. Has this landed recently?

tests/snapshot/snapshots/snapshot__pretty_multiline_doc.ncl.snap
```diff
@@ -0,0 +1,21 @@
...
+    ```nickel
+      ("True" | BoolLiteral) =>
+        "true"
+      ("hello" | BoolLiteral) =>
+        error
+      (true | BoolLiteral) =>
+        error
+    ```"%%
+     = 1,
+}
```
yannham, line 21: Currently, we've rather written such things as:

```
{
  field | doc m%%"
    Contract to enforce the value is a string that represents a boolean literal. Additionally casts "True" to "true"
    and "False" to "false". This shouldn't interpolate: %{null}

    For example:
    ```nickel
      ("True" | BoolLiteral) =>
        "true"
      ("hello" | BoolLiteral) =>
        error
      (true | BoolLiteral) =>
        error
    ```
    "%%
     = 1,
}
```

That being said, nothing is set in stone, as we haven't decided on clear formatting rules yet.

yannham (approved):

src/pretty.rs
vkleen: Good catch, I don't think it's required. Seems to be a copy&paste artefact from the old code.

src/pretty.rs
vkleen: I think it's https://github.com/rust-lang/rust/pull/90473. I guess, it was stabilized in 1.58, can't really tell easily.

src/pretty.rs
vkleen: Actually, that was a terrible idea. We have a function `min_interpolate_sign` in the same file to compute the precise count needed 🤦

tests/snapshot/snapshots/snapshot__pretty_multiline_doc.ncl.snap
vkleen: I've changed to your suggested formatting 👍

ebresafegaga (approved):

PR #1050, opened 2023-01-17
Allow special characters in quoted field identifiers 

### LINKED ISSUE #888: Special characters in quoted field identifiers should be allowed
**Describe the bug**
It is currently impossible to use a special character, e.g. `京`, in quoted type field identifiers, while it should be possible.

**To Reproduce**
```
nickel> {"fo京o" = "bar"}
{ fo京o = "bar" }
nickel> {"fo京o" = "bar"} : {"fo京o" : Str}
error: incompatible types
  ┌─ repl-input-11:1:1
  │
1 │ {"fo京o" = "bar"} : {"fo京o" : Str}
  │ ^^^^^^^^^^^^^^^^^ this expression
  │
  = The type of the expression was expected to be `{ "fo京o" : Str  , }` (a contract)
  = The type of the expression was inferred to be `{fo京o: Str}`
  = Static types and contracts are not compatible
```

**Expected behavior**
This should correctly parse and be supported.

**Environment**
 - Version of the code: `master`

**Additional context**
* Special characters are accepted in expressions, e.g. `{ "fo京o" = "bar" }`
* Special characters are accepted in contracts, e.g. `{"fo京o" = "bar"} | {"fo京o" : Str}`
* Special characters **are not accepted** in types input by users, e.g. `{"fo京o" = "bar"} : {"fo京o" : Str}`:
```
nickel> {"fo京o" = "bar"} : {"fo京o" : Str}
error: incompatible types
  ┌─ repl-input-8:1:1
  │
1 │ {"fo京o" = "bar"} : {"fo京o" : Str}
  │ ^^^^^^^^^^^^^^^^^ this expression
  │
  = The type of the expression was expected to be `{ "fo京o" : Str  , }` (a contract)
  = The type of the expression was inferred to be `{fo京o: Str}`
  = Static types and contracts are not compatible
```
* Special characters **are accepted** in types inferred/generated by Nickel, e.g.:
```
nickel> :typecheck {"fo京o" = "bar"} : _
Ok: {fo京o: Str}
```

The documentation should probably also be updated.

ebresafegaga:
Closes #888

### DIFF
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -261,7 +261,7 @@ impl UniRecord {
                                     }) else {
                                         return Err(InvalidRecordTypeError(rt.pos))
                                     };
-                                let id = Ident::new_with_pos(id, rt.pos);
+                                let id = Ident::new_with_pos(id, expr.pos);
                                 term_to_record_rows(id, rt, acc)
                             }
                         }
--- a/src/parser/uniterm.rs
+++ b/src/parser/uniterm.rs
@@ -6,7 +6,10 @@ use utils::{build_record, elaborate_field_path, FieldPath, FieldPathElem};
 use crate::{
     environment::Environment,
     position::{RawSpan, TermPos},
-    term::{record::RecordAttrs, Contract, MergePriority, MetaValue, RichTerm, SharedTerm, Term},
+    term::{
+        record::RecordAttrs, Contract, MergePriority, MetaValue, RichTerm, SharedTerm, StrChunk,
+        Term,
+    },
     types::{
         EnumRows, EnumRowsIteratorItem, RecordRow, RecordRows, RecordRowsF, TypeF, Types,
         UnboundTypeVariableError, VarKind,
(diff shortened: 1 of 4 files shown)


src/parser/uniterm.rs
```diff
@@ -171,6 +174,40 @@ impl UniRecord {
         // `pos_ellipsis` should be set iff `attrs.open` is true.
         debug_assert!((self.pos_ellipsis == TermPos::None) != self.attrs.open);
 
+        fn term_to_record_rows(
```
yannham, line 177: nitpick: I would put this function at the beginning of `into_type_strict`. It's slightly odd to have some code (the debug assert), then a local function definition, then code again.

src/parser/uniterm.rs
```diff
@@ -207,38 +244,26 @@ impl UniRecord {
...
-                                        )))
-                                    }
-                                }
+                            FieldPathElem::Ident(id) => term_to_record_rows(id, rt, acc),
+                            FieldPathElem::Expr(expr) => {
+                                let Term::StrChunks(chunks) = expr.term.as_ref() else {
+                                        return Err(InvalidRecordTypeError(rt.pos))
+                                    };
+                                let Some(id) =
+                                    chunks.iter().fold(Some(String::new()), |acc, next| {
```
yannham, line 253: Looks like it has been copied verbatim from https://github.com/tweag/nickel/blob/5449b4c63f247cc804221f0f1cbdc865e8496784/src/parser/utils.rs#L246. I think it's worth making it a function, such as `StrChunks::try_as_static_str(&self) -> Option<String>` and replace the use sites by it.

yannham (changes requested):
Ok overall, but it's missing tests, at least a regression test of the example originally provided in the issue.

src/term/mod.rs
```diff
@@ -792,6 +792,25 @@ impl Term {
             | Term::ParseError(_) => false,
         }
     }
+
+    /// Extract the static literal from string chunk. It only returns a `Some(..)`
+    /// when the term is a `Term::StrChunk` and all the chunks are `StrChunk::Literal(..)`
+    pub fn try_str_chunk_as_static_str(&self) -> Option<String> {
```
yannham, line 798: Any reason not to put it inside `StrChunks` directly, and get rid of the `str_chunk` in the name?

ebresafegaga: `StrChunks` is not a special type:  it is a term containing a `Vec<StrChunk>`

yannham: Ah, it's a good reason, then :sweat_smile:

src/parser/uniterm.rs
```diff
@@ -167,6 +167,40 @@ impl UniRecord {
...
     /// `Err`.
     pub fn into_type_strict(self) -> Result<Types, InvalidRecordTypeError> {
+        fn term_to_record_rows(
+            id: Ident,
+            rt: RichTerm,
+            tail: RecordRows,
+        ) -> Result<RecordRows, InvalidRecordTypeError> {
+            // At parsing stage, all `Rc`s must be 1-counted. We can thus call
+            // `into_owned()` without risking to actually clone anything.
+            match rt.term.into_owned() {
```
vkleen, line 177: Not a blocker, but is there some way we could make this call fail instead of silently cloning if the invariant in the comment ends up broken by someone? In the spirit of failing early.

vkleen (approved):

yannham (approved):

PR #1051, opened 2023-01-17
Improve flake source filtering

yannham:
Depends on #1049

Trying to avoid undue rebuilds, this PR makes source filtering of the Rust outputs more precise: for example, we don't include `/lsp/*` inside the source of the `nickel` derivation.


yannham:
It seems that `cargo` is unhappy to not find one of the crate of the workspace. Will investigate rapidly, but if that turns out to be too complex, let's just forget about this.

PR #1053, opened 2023-01-18
Extending the standard library

vkleen:
These are standard library functions we wrote in Nickel hour 15 together with @ebresafegaga and @yannham.

### DIFF
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -296,8 +296,8 @@
 
     flat_map : forall a b. (a -> Array b) -> Array a -> Array b
       | doc m%"
-        The flatmap operation for arrays. First maps the given function over the
-        array and then concatenates the result.
+        First maps the given function over the array and then flattens the
+        result.
 
         For example:
         ```nickel
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -306,5 +306,24 @@
         ```
       "%
       = fun f xs => map f xs |> flatten,
+
+    intersperse : forall a. a -> Array a -> Array a
+      | doc m%"
+        Intersperse the passed value between the elements of an array.
+
+        For example:
+        ```nickel
+        intersperse ", " [ "Hello", "wonderful", "world!" ]
+          => [ "Hello", ", ", "wonderful", ", ", "world!" ]
+        intersperse ", " [ "Hello" ]
+          => [ "Hello" ]
+        intersperse ", " []
+          => []
+        ```
+      "%
+      = fun v arr =>
+        if length arr <= 1
+        then arr
+        else [ head arr ] @ (tail arr |> flat_map (fun a => [v, a])),
   }
 }
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -280,5 +280,13 @@
           l
         else
           (sort cmp (parts.right)) @ [first] @ (sort cmp (parts.wrong)),
+
+    optionals : forall a. Bool -> Array a -> Array a
+      | doc "TODO"
+      = fun condition values => if condition then values else [],
+
+    flat_map : forall a b. (a -> Array b) -> Array a -> Array b
+      | doc "TODO"
+      = fun f xs => map f xs |> flatten,
   }
 }
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -281,19 +281,6 @@
         else
           (sort cmp (parts.right)) @ [first] @ (sort cmp (parts.wrong)),
 
-    optionals : forall a. Bool -> Array a -> Array a
-      | doc m%"
-        Takes a Bool and an Array and returns the array back if it's true,
-        otherwise, it returns an empty array.
-
-        For example:
-        ```nickel
-          optionals true [1, 2, 3] => [1, 2, 3]
-          optionals false [`a, `b, `c] => []
-        ```
-      "%
-      = fun condition values => if condition then values else [],
-
     flat_map : forall a b. (a -> Array b) -> Array a -> Array b
       | doc m%"
         First maps the given function over the array and then flattens the
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -282,11 +282,29 @@
           (sort cmp (parts.right)) @ [first] @ (sort cmp (parts.wrong)),
 
     optionals : forall a. Bool -> Array a -> Array a
-      | doc "TODO"
+      | doc m%"
+        Takes a Bool and an Array and returns the array back if it's true,
+        otherwise, it returns an empty array.
+
+        For example:
+        ```nickel
+          optionals true [1, 2, 3] => [1, 2, 3]
+          optionals false [`a, `b, `c] => []
+        ```
+      "%
       = fun condition values => if condition then values else [],
 
     flat_map : forall a b. (a -> Array b) -> Array a -> Array b
-      | doc "TODO"
+      | doc m%"
+        The flatmap operation for arrays. First maps the given function over the
+        array and then concatenates the result.
+
+        For example:
+        ```nickel
+        flat_map (fun x => [x, x]) [1, 2, 3]
+          => [1, 1, 2, 2, 3, 3]
+        ```
+      "%
       = fun f xs => map f xs |> flatten,
   }
 }
--- a/stdlib/builtin.ncl
+++ b/stdlib/builtin.ncl
@@ -226,11 +226,37 @@
     = fun msg x => %trace% msg x,
 
     FailWith
-    | doc "TODO"
+    | doc m%"
+      A contract that always fails with a given message.
+
+      For example:
+      ```nickel
+      1 | FailWith "message" =>
+        error: contract broken by a value: message
+          ┌─ :1:1
+          │
+        1 │ FailWith "message"
+          │ ---------------------------- expected type
+          │
+          ┌─ repl-input-0:1:1
+          │
+        1 │ 1 | FailWith "message"
+          │ ^ applied to this expression
+      ```
+    "%
     = fun msg label value => contract.blame_with msg label,
 
-    fail_with 
-    | doc "TODO"
+    fail_with | Str -> Dyn
+    | doc m%"
+      Unconditionally abort evaluation with the given message.
+
+      For example:
+      ```nickel
+      fail_with "message" =>
+        error: contract broken by a value: message
+      ```
+    "%
     = fun msg => null | FailWith msg,
+
   },
 }
--- a/stdlib/builtin.ncl
+++ b/stdlib/builtin.ncl
@@ -224,5 +224,13 @@
       ```
       "%
     = fun msg x => %trace% msg x,
+
+    FailWith
+    | doc "TODO"
+    = fun msg label value => contract.blame_with msg label,
+
+    fail_with 
+    | doc "TODO"
+    = fun msg => null | FailWith msg,
   },
 }
--- a/stdlib/function.ncl
+++ b/stdlib/function.ncl
@@ -16,7 +16,7 @@
     | doc "TODO"
     = fun f x y => f y x,
 
-    fst
+    fst : forall a b. a -> b -> a
     | doc "TODO"
     = const,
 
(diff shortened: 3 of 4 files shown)


vkleen:
To add from the Nixel meeting:
- [ ] `array.intersperse : forall a. a -> Array a -> Array a`
- [ ] Something like `string.join` for symbolic strings

vkleen:
I'll mark this as ready for review the way it is. We have doc comments on all new functions by now. The only thing that might be missing are integration tests for the new functions.

Extensions that we have forgotten for now can go into a separate PR.

stdlib/array.ncl
```diff
@@ -280,5 +280,50 @@
...
+        ```nickel
+          optionals true [1, 2, 3] => [1, 2, 3]
+          optionals false [`a, `b, `c] => []
+        ```
+      "%
+      = fun condition values => if condition then values else [],
+
+    flat_map : forall a b. (a -> Array b) -> Array a -> Array b
+      | doc m%"
+        The flatmap operation for arrays. First maps the given function over the
```
yannham, line 299: Nitpick: is it really useful to start the documentation of `flat_map` by saying "it's the flatmap operation"?

```suggestion
        First maps the given function over the
```

stdlib/array.ncl
```diff
@@ -280,5 +280,50 @@
...
+        array and then concatenates the result.
```
yannham, line 300: Or can also keep concatenate only, but then I think it needs to be expanded. Concatenate the result sounds a bit strange, because in general you concatenate many pieces together, not "one thing". 

```suggestion
        array and then flatten the result.
```

stdlib/function.ncl
```diff
@@ -7,5 +7,35 @@
     const : forall a b. a -> b -> a
     | doc "The const function, results in the first argument."
```
yannham, line 8:
```suggestion
    | doc "Results in the first argument."
```

stdlib/function.ncl
```diff
@@ -7,5 +7,35 @@
...
+
+    fst : forall a b. a -> b -> a
+    | doc "Always returns the first argument."
+    = const,
+
+    snd : forall a b. a -> b -> b
+    | doc "Always returns the second argument."
+    = flip const,
+
+    pipe | Dyn -> Array (Dyn -> Dyn) -> Dyn
```
yannham, line 27: I wonder if `apply_all` wouldn't be more explicit, as a name. But this is just a remark, we can bikeshed later.

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+      ```nickel
+        map_values (fun x => x + 1) { hi = 2 } =>
+          { hi = 3 }
+        map (fun x => x + 1) { hello = 1, world = 2 } =>
+          { hello = 2, world = 3 }
+      ```
+    "%
+    = fun f => map (fun _field value => f value),
+
+    as_array : forall a. {_: a} -> Array {field: Str, value: a}
```
yannham, line 125: `to` is more in line with the naming elsewhere in Nickel? Maybe?

```suggestion
    to_array : forall a. {_: a} -> Array {field: Str, value: a}
```

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+        as_array { hello = "world", foo = "bar" } =>
+          [ { field = "hello", value = "world" },
+            { field = "foo", value = "bar" }
+          ]
+      ```
+    "%
+    = fun record =>
+      record
+      |> fields
+      |> array.map (fun field_ => {field = field_, value = record."%{field}"}),
```
yannham, line 140: Just a suggestion, but I always forget that we can use `'` as indentifiers. I find it prettier, but that's unimportant, and entirely subjective, so feel free to ignore :slightly_smiling_face: 

Beside, using `field'` instead of `field` makes the record non-recursive (technically, the thunk is non-revertible), which is probably better in general for performances.

```suggestion
      |> array.map (fun field' => {field = field', value = record."%{field'}"}),
```

stdlib/function.ncl
yannham: Also, is there a reason to use a contract instead of a type annotation here?

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+      Check whether a record is empty.
+      
+      ```nickel
+      is_empty {} => true
+      is_empty { foo = 1 } => false
+      ```
+    "%
+    = (==) {},
+
+    merge_array | forall a. Array {_: a} -> {_: a}
```
yannham, line 170: Simple suggestion. Not sure about that.

```suggestion
    merge_all | forall a. Array {_: a} -> {_: a}
```

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+    | doc m%"
+      Merge an array of records.
+
+      For example:
+      ```nickel
+      merge_array [ { foo = 1 }, { bar = 2 } ]
+        => { foo = 1, bar = 2 }
+      ```
+    "%
+    = array.foldl (fun acc r => acc & r) {},
```
yannham, line 180: Don't we have `(&)` as well?

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+
+    filter | (Str -> Dyn -> Bool) -> {_: Dyn} -> {_: Dyn}
```
yannham, line 182: Ditto: can't this be statically typed?

stdlib/record.ncl
```diff
@@ -107,5 +107,93 @@
...
+            { field = "foo", value = "bar" }
+          ]
+      ```
+    "%
+    = fun record =>
+      record
+      |> fields
+      |> array.map (fun field_ => {field = field_, value = record."%{field}"}),
+
+    from_array | forall a. Array { field: Str, value: a} -> {_: a}
```
yannham, line 142: Ditto: can't this be statically typed?

stdlib/function.ncl
```diff
@@ -7,5 +7,35 @@
     const : forall a b. a -> b -> a
     | doc "The const function, results in the first argument."
     = fun x y => x,
+
+    compose : forall a b c. (b -> c) -> (a -> b) -> (a -> c)
+    | doc "Function composition, right to left."
+    = fun g f x => x |> f |> g,
+
+    flip : forall a b c. (a -> b -> c) -> b -> a -> c
+    | doc "Flip the argument order for a curried function."
```
yannham, line 16: I find this explanation not extremely clear for someone who doesn't already know what `flip` does.

stdlib/function.ncl
```diff
@@ -7,5 +7,35 @@
     const : forall a b. a -> b -> a
     | doc "The const function, results in the first argument."
     = fun x y => x,
+
+    compose : forall a b c. (b -> c) -> (a -> b) -> (a -> c)
+    | doc "Function composition, right to left."
```
yannham, line 12: Could be useful to write down how it acts on an element: `compose f g x = f (g x)` in the documentation

yannham (commented):
The additions to `function` should probably have examples, even if they look simple. Let's keep in mind that our potential user might not be functional programmers, and don't necessarily understand what `flip` does, and why it even exists.

stdlib/function.ncl
vkleen: I seem to remember that we were concerned about changing types through different stages of the pipeline. But that doesn't seem to actually be a problem, `pipe : forall a. a -> Array (a -> a) -> a` works just fine for the examples.

stdlib/record.ncl
vkleen: `to_array` also feels more complementary to `from_array` 👍

stdlib/record.ncl
vkleen: The implementation used destructuring and needed @matthew-healy's #1099 to typecheck. But since that's merged now, it was just a rebase away 🎉

stdlib/record.ncl
vkleen: I like `merge_all` better 👍

stdlib/record.ncl
vkleen: This can indeed be statically typed.

stdlib/record.ncl
yannham: Ah, I see :slightly_smiling_face:

stdlib/array.ncl
```diff
@@ -280,5 +280,50 @@
...
+        Takes a Bool and an Array and returns the array back if it's true,
+        otherwise, it returns an empty array.
+
+        For example:
+        ```nickel
+          optionals true [1, 2, 3] => [1, 2, 3]
+          optionals false [`a, `b, `c] => []
+        ```
+      "%
+      = fun condition values => if condition then values else [],
```
matthew-healy, line 295: Are we certain we need a standard library function for this? I'm not sure `array.optionals some_bool some_array` is really any more descriptive than just using the `if` statement - I personally think it might be _less_ clear.

stdlib/builtin.ncl
```diff
@@ -224,5 +224,39 @@
       ```
       "%
     = fun msg x => %trace% msg x,
+
+    FailWith
```
matthew-healy, line 228: I'm not a fan of `With` in the name here, as it doesn't tell us _what_ we're failing with. I think `FailWithMessage` is significantly clearer.

stdlib/array.ncl
vkleen: The idea of this function is really stolen from the nixpkgs libraries. They seem to find value in writing, e.g.
```
buildInputs = [ foo bar ]
  ++ lib.optionals some_condition [ baz another_package ]
  ++ lib.optionals some_other_condition [ yet_another_thing im_running_out_of_names ];
```
Since Nix could just as easily use `if-then-else` as we could, I feel that's empirical evidence that the function is useful. Now, of course we probably would like to structure our dependency specifications differently, so maybe that's a moot point.

matthew-healy: Fair enough! I'm still not convinced this warrants inclusion in the standard library, but I at least understand the context now.

To expand on my thoughts a little more: in general I'm not sure "does this exist in nixpkgs?" is the best way to gauge whether a function warrants inclusion in the standard library. It's easy enough for users to write their own ports of nixpkgs functions they miss in Nickel (with the caveat that we don't yet have a good package management story). I think the more important question is: is this function likely to be useful to a large number of people writing Nickel? And in this case, I'm not convinced that it is.

Additionally, without looking at the definition of the `optional` function, I'd have a hard time telling you what it does based on the code snippet above, whereas:

```
buildInputs = [ foo bar ]
 ++ (if some_condition then [ baz another_package ] else [])
 ++ (if other_condition then [ yet_another_thing im_running_out_of_names ] else [])
```

is likely to be immediately clear even to newcomers to the language.

stdlib/builtin.ncl
vkleen: The `With` probably comes from analogy with `contract.blame_with`. I'm not opposed to renaming `FailWith` to `FailWithMessage`, it just raises the question whether `contract.blame_with` (and `fail_with` right here) should also have their names revised.

matthew-healy: Ah good point! I guess you could infer this from what I've said above, but to be extra clear: I'm in favour of renaming both for clarity & consistency. :grinning:

stdlib/array.ncl
yannham: While I'm not sure it's a full justification, the fact that `if-then-else` introduces one additional level of nesting and the trailing `else []` which is always the same is enough, I fear, to make people write their own version of `lib.optionals` anyway (well, that's what they did in Nix, as it's not in the stdlib). But there may be a case for putting it in, say, the Nixel library, and not in the standard library.

It's also a good point that in Nickel we might use mostly records instead of arrays. I'm fine with not including this right away in the stdlib. I think we've used it here and there but I don't know if there's evidence that this is required in the 1.0 stdlib.

stdlib/builtin.ncl
vkleen: Let's do that bikeshedding later and merge it the way it is, just in the interest of getting this PR over with.

yannham (approved):
The documentation and code formatting aren't always uniform, but we'll make a pass about that before 1.0 anyway. Good to go, as far as I'm concerned :+1:

PR #1054, opened 2023-01-18
pyckel: diagnostics in exception

GuillaumeDesforges:
Makes `pyckel` render beautiful diagnostics on evaluation error.

```shell
$ python -c 'import pyckel; print(pyckel.run("let x = 1 in { y = x / 0 }"))'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
pyckel.NickelException: error: division by zero
  ┌─ python:1:20
  │
1 │ let x = 1 in { y = x / 0 }
  │                    ^^^^^ here
```

### DIFF
--- a/pyckel/src/lib.rs
+++ b/pyckel/src/lib.rs
@@ -1,10 +1,20 @@
 use std::io::Cursor;
 
-use nickel_lang::{eval::cache::CBNCache, program::Program, serialize};
+use nickel_lang::{
+    error::Error,
+    eval::cache::{CBNCache, Cache},
+    program::Program,
+    serialize,
+};
 
 use pyo3::{create_exception, exceptions::PyException, prelude::*};
 
 create_exception!(pyckel, NickelException, PyException);
+
+/// Turn a Nickel (Rust) error into a NickelException (Python)
+fn error_to_exception<E: Into<Error>, EC: Cache>(error: E, program: &mut Program<EC>) -> PyErr {
+    NickelException::new_err(program.report_as_str(error.into()))
+}
 
 /// Evaluate from a Python str of a Nickel expression to a Python str of the resulting JSON.
 #[pyfunction]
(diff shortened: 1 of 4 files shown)


src/program.rs
```diff
@@ -163,6 +163,32 @@ impl<EC: EvalCache> Program<EC> {
         report(self.vm.import_resolver_mut(), error, self.color_opt)
     }
 
+    /// Wrapper for [`report`].
+    pub fn report_as_str<E>(&mut self, error: E) -> String
+    where
+        E: ToDiagnostic<FileId>,
+    {
+        let contracts_id = self.vm.import_resolver().id_of("<stdlib/contract.ncl>");
```
yannham, line 171: `self.vm.import_resolver()` is used 3 times inside the function. Might it be a tad more readable to store it in a variable?

src/program.rs
```diff
@@ -163,6 +163,32 @@ impl<EC: EvalCache> Program<EC> {
...
+                codespan_reporting::term::emit(
+                    &mut buffer,
+                    &config,
+                    self.vm.import_resolver_mut().files_mut(),
+                    d,
+                )
+            })
+            // safe because writing to a cursor in memory
+            .unwrap();
+        String::from_utf8(buffer.into_inner().into_inner()).unwrap()
```
yannham, line 189:
```suggestion
        // unwrap(): emit() should only print valid utf8 to the the buffer
        String::from_utf8(buffer.into_inner().into_inner()).unwrap()
```

pyckel/src/lib.rs
```diff
@@ -1,62 +1,26 @@
...
 /// Evaluate from a Python str of a Nickel expression to a Python str of the resulting JSON.
 #[pyfunction]
 pub fn run(s: String) -> PyResult<String> {
     let mut program: Program<CBNCache> =
         Program::new_from_source(Cursor::new(s.to_string()), "python")?;
 
-    let term = program.eval_full().map_err(NickelError)?;
+    let term = program
+        .eval_full()
+        .map_err(|err| NickelException::new_err(program.report_as_str(err)))?;
```
yannham, line 17: That's probably my own proposal to inline this expression :sweat_smile: , but I didn't realize we use it 3 times. Maybe it's better to factor it out in a dedicated function, in the end?

src/program.rs
GuillaumeDesforges: Fixed in 20145575

src/program.rs
GuillaumeDesforges: Fixed in 20145575

src/program.rs
```diff
@@ -163,6 +163,32 @@ impl<EC: EvalCache> Program<EC> {
         report(self.vm.import_resolver_mut(), error, self.color_opt)
     }
 
+    /// Wrapper for [`report`].
```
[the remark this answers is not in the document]
GuillaumeDesforges: Fixed in 20145575

pyckel/src/lib.rs
GuillaumeDesforges: Fixed in 54e7cad3

Funny thing is, it did not make it shorter. 😅

yannham (approved):

PR #1055, opened 2023-01-19
Add a %trace% primop

vkleen:
Add a `%trace%` builtin which prints a message to standard output when encountered by the evaluator, together with a standard libary function `builtin.trace`.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -1258,6 +1258,29 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     ))
                 }
             },
+            UnaryOp::Trace() => {
+                if let Term::Str(s) = &*t {
+                    println!("builtin.trace: {s}");
+                    Ok(())
+                } else {
+                    Err(EvalError::TypeError(
+                        String::from("Str"),
+                        String::from("trace"),
+                        arg_pos,
+                        RichTerm { term: t, pos },
+                    ))
+                }?;
+
+                if self.stack.count_args() >= 1 {
+                    let (next, ..) = self
+                        .stack
+                        .pop_arg(&self.cache)
+                        .expect("Condition already checked.");
+                    Ok(next)
+                } else {
+                    Err(EvalError::NotEnoughArgs(2, String::from("trace"), pos_op))
+                }
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/eval/operation.rs
```diff
@@ -1258,6 +1258,29 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     ))
                 }
             },
+            UnaryOp::Trace() => {
+                if let Term::Str(s) = &*t {
+                    println!("builtin.trace: {s}");
+                    Ok(())
+                } else {
```
vkleen, line 1265: Just using `println!` might not be the best option. E.g. I don't know how that interacts with the WASM REPL.

vkleen: After trying it locally in the WASM playground, it doesn't seem to break anything. But the REPL in the playground just doesn't print out any messages.

src/term/mod.rs
```diff
@@ -1047,6 +1047,10 @@ pub enum UnaryOp {
     /// define a `field_diff` function that preserves the sealed polymorphic
     /// tail of its argument.
     RecordEmptyWithTail(),
+
+    /// Print a message when encountered during evaluation. Operationally the same as the identity
```
yannham, line 1051:
```suggestion
    /// Print a message when encountered during evaluation and proceed with the evaluation of the argument
    /// on the top of the stack. Operationally the same as the identity
```

yannham (approved):

src/eval/operation.rs
vkleen: I think I'll make an issue to track that. For now, I'll just change `println!` to `eprintln!` to make this more compatible with invocations like `nickel export -f test.ncl > test.json`.

src/eval/operation.rs
```diff
@@ -1258,6 +1258,29 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        arg_pos,
+                        RichTerm { term: t, pos },
+                    ))
+                }?;
+
+                if self.stack.count_args() >= 1 {
+                    let (next, ..) = self
+                        .stack
+                        .pop_arg(&self.cache)
+                        .expect("Condition already checked.");
```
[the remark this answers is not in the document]
vkleen: I did indeed paste it from `%seq%` 😅 Your non-panicking version looks much nicer; I'll change both places.

PR #1057, opened 2023-01-19
Add support for LSP completion using the surrounding context

### LINKED ISSUE #995: LSP server should provide completion based on contract/type context
Say we have the following program, with the cursor represented as `[.]`: 
```
{
[.]
} | SchemaFromNix
```
We should be able to get completion based on the field names from the `SchemaFromNix` contract (or maybe type)

### LINKED ISSUE #998: Support for contracts represented by variables (or record field names)
Currently the LSP only supports completion when we have inline contracts. For example, if we have: 
```
let baz | {bar |Str, } = {bar = "bar"} in baz[.] 
```

completion happens as usual, but if we assign the contract to a variable like this: 
```
let foo = {
  bar | Str,
} in
let baz | foo = {bar = "bar", } in baz[.] 
```
no completion happens.

Or even a more complex example (involving multiple files): 
```
# file: lib.ncl
let contracts_ = {
  Foo = {foo, bar},
  # etc.
} in
let functions_ = {..} in
{
  functions,
  contracts,
}

# file: confing.ncl
let lib = import "lib.ncl" in
{
  [.]
} | lib.contracts.Foo
```

ebresafegaga:
This pull request adds support for auto-completion based on the context (from a record, type, or contract.)

### Case 1
Say we have the following: 
```
{
[.]
} | SchemaFromNix
```
With this change, if the contract `SchemaFromNix`, represents a record or dictionary contract, we will get completion corresponding to its fields. 

### Case 2
This will for work not only for basic records, but also for nested records: 
```
let Tf = import "./contract.ncl" in
{
  config = {
    data.dhall.my_dhall_resource = {
      [.]
    }
  }
} | Tf.Config
```
Here, we would get a completion based on the fields of  `Tf.Config.config.data.dhall.my_dhall_resource`, if it is a record dictionary contract.

### Case 3
This also includes support for nested record completion based on the context contract.
```
let Tf = import "./contract.ncl" in
{
  config[.]
} | Tf.Config
```
In this case, we would get completion based on the fields of `Tf.Config.config`, if it is a record or dictionary contract.

#### 2. Adds support for completion when we have a dictionary contract: say we have the following:
  ``` 
let foo | {_: {a: Str, b: Str}} = <record definition> in 
 foo[.] 
```
This doesn't get a completion, becuase the dictionary has dynamic fields, but when we do this: 
```
foo.random_dynamic_field[.]
```
We will get a completion
#### 3. Adds support for contracts represented by variables or static record access (See #998)

<hr />
Closes #995, Closes #998

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -400,8 +400,9 @@ fn get_completion_identifiers(
     }
 
     let in_scope = match trigger {
-        // Record Completion
         Some(server::DOT_COMPL_TRIGGER) => {
+            // Record completion
+            // TODO: Context completion can also happen here
             get_identifier_path(source)
                 .and_then(|path| {
                     let mut path: Vec<_> = path.iter().rev().cloned().map(Ident::from).collect();
@@ -425,8 +426,43 @@ fn get_completion_identifiers(
             } else {
                 if let TermKind::RecordField { record, .. } = item.kind {
                     // Context completion
-                    // This unwrap is safe: we know a `RecordField` must have a containing `Record`.
-                    let parent = linearization.get_item(record, &server.lin_cache).unwrap();
+
+                    /// Find the topmost record that contains a particular record field.
+                    /// It returns an an item (which is the containing record), and the
+                    /// field path from the field to the topmost record (not including the
+                    /// record field's ident)
+                    fn find_topmost_record<'a>(
+                        linearization: &'a Completed,
+                        server: &'a Server,
+                        record: ItemId,
+                        mut path: Vec<Ident>,
+                    ) -> (&'a LinearizationItem<Types>, Vec<Ident>) {
+                        // This unwrap is safe: we know a `RecordField` must have a containing `Record`.
+                        let parent = linearization.get_item(record, &server.lin_cache).unwrap();
+                        let next =
+                            // Oops.. linear search
+                            linearization
+                                .linearization
+                                .iter()
+                                .find_map(|item| match item.kind {
+                                    TermKind::RecordField {
+                                        ident,
+                                        record,
+                                        value: ValueState::Known(value),
+                                        ..
+                                    } if value == parent.id => Some((ident, record)),
+                                    _ => None,
+                                });
+                        match next {
+                            None => (parent, path),
+                            Some((name, record)) => {
+                                path.push(name);
+                                find_topmost_record(linearization, server, record, path)
+                            }
+                        }
+                    }
+                    let (parent, mut path) =
+                        find_topmost_record(linearization, server, record, Vec::new());
                     let Some(meta) = parent.meta.as_ref().map(|meta| &meta.contracts) else {
                         // I'm not *entirely* sure we want to do this here.
                         // Alternative: if we cannot find a meta on any record,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by vkleen — APPROVED** [review]

File: lsp/nls/src/requests/completion.rs [hunk 1]
```diff
@@ -379,6 +424,43 @@ fn collect_record_info(
...
+/// accumulate all the record metadata and corresponding path.
+fn accumulate_record_meta_data<'a>(
+    record: ItemId,
+    info @ ComplInfo {
+        linearization,
+        lin_cache,
+    }: &'a ComplInfo<'a>,
+    mut path: Vec<Ident>,
+    mut result: Vec<(&'a LinearizationItem<Types>, Vec<Ident>)>,
+) -> Vec<(&'a LinearizationItem<Types>, Vec<Ident>)> {
```
yannham, line 437: Is it useful to have a return value? Aren't we mutating the argument directly?

src/parser/grammar.lalrpop
```diff
@@ -311,7 +311,7 @@ RecordOperand: UniTerm = {
 
 // A record operation chain, such as `{foo = data}.bar.baz`.
 RecordOperationChain: RichTerm = {
-    <t: AsTerm<RecordOperand>> "." <id: Ident> => mk_term::op1(UnaryOp::StaticAccess(id), t),
+    <t: AsTerm<RecordOperand>> "." <id: Ident> => mk_term::op1_with_pos(UnaryOp::StaticAccess(id), t, id.pos),
```
yannham, line 314: You don't need to add a new function, there is `RichTerm::with_pos`.

```suggestion
    <t: AsTerm<RecordOperand>> "." <id: Ident> => mk_term::op1_with_pos(UnaryOp::StaticAccess(id), t).with_pos( id.pos),
```

src/term/mod.rs
```diff
@@ -1792,6 +1792,13 @@ pub mod make {
         Term::Op1(op, t.into()).into()
     }
 
+    pub fn op1_with_pos<T>(op: UnaryOp, t: T, pos: TermPos) -> RichTerm
+    where
+        T: Into<RichTerm>,
+    {
+        RichTerm::new(Term::Op1(op, t.into()), pos)
+    }
+
```
yannham, line 1801:
```suggestion
```

src/parser/grammar.lalrpop
```diff
@@ -396,7 +396,19 @@ RecordField: (FieldPath, RichTerm) = {
...
+        errors.push(t.clone());
+
+        let expr = RichTerm::new(Term::ParseError(
+            crate::error::ParseError::from_lalrpop(t.error, src_id)), pos);
+
+        let elem = FieldPathElem::Expr(expr);
+        let path = vec![elem];
+        let term = RichTerm::new(Term::MetaValue(MetaValue::new()), pos);
+        (path, term)
+    },
```
yannham, line 411: I think this code has been copy paster from another rule, isn't it? It might be interesting to factor this out in a separate rule:

```
Error : RichTerm =  <l: @L> <t: !> <r: @R> => {
        let pos = mk_pos(src_id, l, r);
        errors.push(t.clone());

        RichTerm::new(Term::ParseError(
            crate::error::ParseError::from_lalrpop(t.error, src_id)), pos)
};
```

And then use this instead:
```suggestion
     <Error> => {
        let path = vec![FieldPathElem::Expr(<>)];
        let term = RichTerm::new(Term::MetaValue(MetaValue::new()), pos);
        (path, term)
    },
```

And use error as well in the Uniterm Rule

yannham (approved):
LGTM. Nice trick to get the parent node by just looking on the left, giving a logarithmic time algo at worst.

lsp/nls/src/requests/completion.rs
ebresafegaga: We're mutating it locally. I'm mostly using it as accumulator here. But we can also pass in it as a mutable reference, and not return anything.

yannham: I mean, it seems like you never actually use the return value, so I think it makes more sense to return `unit`, yeah.

PR #1059, opened 2023-01-20
Make export snapshot tests capture stderr

vkleen:
Add standard error capture in the snapshot tests for `nickel export`. This is done by calling `assert_snapshot!` twice, once for `stdout` and once for `stderr`, so that the snapshots are reasonably reviewable in `cargo insta review`.

Combining both into one call would be feasible, but only at the cost of presenting the output as raw string literals during snapshot review, I think. If someone knows of a better way, I'd love to see it!

Depends on #1055

### DIFF
--- a/tests/snapshot/README.md
+++ b/tests/snapshot/README.md
@@ -18,6 +18,11 @@ nix dev shell, `cargo-insta` is already on your path. Most of this guide will
 assume you have `cargo-insta` installed. If you'd prefer not to install it,
 please read [this section](#without-cargo-insta).
 
+The snapshot tests for `nickel export` capture both `stdout` and `stderr`. For
+easier reviewability in `cargo insta review` this is done with two separate
+assertions. As an unfortunate side effect, adding a new test input requires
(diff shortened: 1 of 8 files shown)


matthew-healy (approved):
Glad to see the snapshot tests are useful! :grin: 

I'm not against merging this as-is, but I do wonder if it might be simpler to have the `stderr` checks run as an entirely separate test. i.e., we split `check_export_snapshots` into `check_export_stdout_snapshots` and `check_export_stderr_snapshots`, both of which consume the same input files. This way, we avoid the annoyance of having to re-run the tests at the relatively low cost of a small amount of test code duplication.

matthew-healy:
Very true! Personally I don't think test runtime is a bottleneck right now so it's probably fine to run it twice, but happy for you to go with whatever you think is best. :smile:

vkleen:
I'll go with running the `stdout` and `stderr` capturing snapshot tests separately for now. If the test runtime becomes an issue we can revisit this, of course. But for now, I feel it's more important to have a smooth experience when adding tests. That way we have fewer barriers to building up a good library of them.

PR #1062, opened 2023-01-22
Nix/rnix update

francois-caddet:
To try to fix some strange bugs and because the API is a lot cleaner, this branch update `rnix` from 0.10 to 0.11. with a lot of breakage in the API but with a good improvement.

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -8,36 +8,45 @@ use crate::term::make::{self, if_then_else};
 use crate::term::{record::RecordData, BinaryOp, UnaryOp};
 use crate::term::{MergePriority, MetaValue, RichTerm, Term};
 use codespan::FileId;
-use rnix;
-use rnix::types::{BinOp, EntryHolder, TokenWrapper, TypedNode, UnaryOp as UniOp};
+use rnix::ast::{BinOp as NixBinOp, Path as NixPath, Str as NixStr, UnaryOp as NixUniOp};
 use std::collections::HashMap;
 
-impl ToNickel for UniOp {
+impl ToNickel for NixStr {
     fn translate(self, state: &State) -> RichTerm {
-        use rnix::types::UnaryOpKind::*;
-        let value = self.value().unwrap().translate(state);
-        match self.operator() {
+        use rnix::ast::InterpolPart;
+        Term::StrChunks(
+            self.parts()
+                .enumerate()
+                .map(|(i, c)| match c {
+                    InterpolPart::Literal(s) => crate::term::StrChunk::Literal(s.to_string()),
+                    InterpolPart::Interpolation(interp) => {
+                        crate::term::StrChunk::Expr(interp.expr().unwrap().translate(state), i)
+                    }
+                })
+                .collect(),
+        )
+        .into()
+    }
+}
+
+impl ToNickel for NixUniOp {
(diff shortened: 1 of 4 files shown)


src/nix.rs
```diff
@@ -250,53 +238,49 @@ impl ToNickel for rnix::SyntaxNode {
...
-            // The following nodes aren't `RichTerm`s but sub-parts of other nodes higher in the syntax tree.
-            // They can't be parsed independently of a term and should never occur at this level of the translation.
-            ParsedType::Pattern(_)
-            | ParsedType::PatEntry(_)
-            | ParsedType::Inherit(..)
-            | ParsedType::InheritFrom(..)
-            | ParsedType::Key(..)
-            | ParsedType::KeyValue(..) => unreachable!(),
+            Expr::Path(_) | Expr::HasAttr(_) => unimplemented!(), // TODO: What is the diff with
+                                                                  // Select / operator HasField?
```
yannham, line 275: I didn't understand this comment.

src/nix.rs
```diff
@@ -250,53 +238,49 @@ impl ToNickel for rnix::SyntaxNode {
...
         .with_pos(crate::position::TermPos::Original(span))
     }
 }
 
 pub fn parse(cache: &Cache, file_id: FileId) -> Result<RichTerm, rnix::parser::ParseError> {
     let source = cache.files().source(file_id);
-    let nixast = rnix::parse(source).as_result()?;
-    Ok(nixast.node().to_nickel(file_id))
+    let root = rnix::Root::parse(source).ok()?; // TODO: we could return a list of errors calling
+                                                // `errors()` could improve error management.
```
yannham, line 284:
```suggestion
    let root = rnix::Root::parse(source).ok()?; // TODO: we could return a list of errors calling
                                                // `errors()` to improve error management.
```

src/nix.rs
```diff
@@ -60,80 +69,59 @@ impl ToNickel for BinOp {
...
-        let node: ParsedType = self.try_into().unwrap();
-        match node {
-            ParsedType::Error(_) => {
+        match self {
+            Expr::Error(_) => {
                 Term::ParseError(crate::error::ParseError::NixParseError(file_id)).into()
+                // TODO:
+                // Improve
+                // error
+                // management
```
yannham, line 86:
```suggestion
                // TODO: Improve error management
```

yannham (approved):
Looks good overall. As long as the tests pass, you can merge this now and worry about clippy on the nix/nickel branch directly.

src/nix.rs
francois-caddet: Sorry:

- For the `HasAttr` I discovered later that it's not an operator anymore in `rnix 0.11`. It was an operator before but now it's a specific language construct.
- But for `Expr::Path` I don't understand what it is for. we already have `Select` which is the `.` operator. So what is a `Expr::Path`? 

`rnix` is a very good crate but sadly without any documentation...

yannham: @francois-caddet I suppose it's a file path, like `/foo/bar/baz`, not something related to attribute paths?

francois-caddet: Yes but so what is `Literal::Uri(_)`? 😢 
I will merge as it is and see what it is adding tests in the main branch `nix/nickel`

PR #1066, opened 2023-01-23
Feature/incremental cache

dpl0a:
This branch introduces `IncCache` a first draft of a real implementation of the `Cache` interface.
`IncCache` will be the basis of a new incremental re-computation back-end. While this is still not a real incremental cache, and it won't be used in the main branch it's probably better to merge it in ASAP. I'll be opening other PRs once I'll finish experimenting.

TO-DO (possibly in following PRs):
- Bump allocation for the `IncCache` `store`: This would make caching faster at the cost of pre-allocating some memory at the start of evaluation.
- **(IMPORTANT)** Incremental node updating and backlinks: The content of an `IncNode` is modified it should be marked as "to be re-evaluated" along with all the other nodes that depend on it. To this end, each node should have a backlinks field, storing all the indices of the nodes that depend on it (suggestions are welcome if you have a better way of reversing the dependency graph/propagating the "to be re-evaluated" status).


src/eval/cache/mod.rs
```diff
@@ -0,0 +1,56 @@
+use super::{Closure, Environment, IdentKind};
```
yannham, line 1: Could deserve a bit of top-level documentation to describe what is the evaluation cache.

src/eval/cache/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// The current Cache implementation
+pub type CacheImpl = lazy::CBNCache;
+//pub type CacheImpl = incremental::IncCache;
+
+/// A black-holed thunk was accessed, which would lead to infinite recursion.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct BlackholedError;
+
+pub trait Cache: Clone {
```
yannham, line 22: I think we need to document those public trait items. Could take inspiration from the corresponding function in `Thunk`, or be more concise for the time being, but otherwise I fear we're going to forget why we needed them in the first place and what they are for.

src/eval/mod.rs
```diff
@@ -94,7 +94,7 @@ use crate::{
 pub mod cache;
 pub mod callstack;
 pub mod fixpoint;
-pub mod lazy;
+//pub mod lazy;
```
yannham, line 97:
```suggestion
```

src/eval/mod.rs
```diff
@@ -147,7 +147,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
     pub fn reset(&mut self) {
         self.call_stack.0.clear();
         self.stack.reset(&mut self.cache);
-        self.cache = Cache::new();
+        //self.cache = Cache::new();
```
yannham, line 150:
```suggestion
```

yannham (commented):
LGTM as far as the implementation is concerned (well incremental caching is WIP, but what's there now), and doesn't change the behavior of the interpreter.

However, I think we should take the occasion to document things a bit more, because we're adding quite a few new notions to the codebase.

src/eval/cache/mod.rs
dpl0a: Right. I hadn't documented them yet because I wasn't sure whether the interface was still not 100% definitive. Now I'm pretty confident there aren't going to be lots of major changes so I'll document the functions.

yannham (approved):

src/eval/tests.rs
```diff
@@ -10,7 +10,7 @@ use crate::transform::import_resolution::strict::resolve_imports;
 use crate::{mk_app, mk_fun};
 use codespan::Files;
 
-type EC = CBNCache;
+type EC = CacheImpl;
```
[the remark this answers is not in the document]
dpl0a: Right, I think I missed that when changing stuff around

PR #1074, opened 2023-01-25
add a tutorial in the documentation

rapenne-s:
this adds a tutorial page that will be shown on nickel-lang.org

### DIFF
--- a/doc/manual/tuto.md
+++ b/doc/manual/tuto.md
@@ -9,29 +9,48 @@ and then export it as a YAML file for a program consuming it.
 
 ## Step 1: Install nickel
 
+The first step is to obtain nickel on your system. There are different
+methods to install Nickel, they are all covered on the [Getting
+started](https://nickel-lang.org/getting-started/#getting-started)
+page of Nickel project website.
+
 ## Step 2: Think about the schema
 
-Our current YAML has different fields, such as `name`, `ssh-keys`,
-`is-admin` or `extra-groups`. To create a Nickel contract that correctly
-specifies the shape of this data, we first need to think about the
-types of each field
+The YAML we want to produce as different fiels with different types
+(lists, booleans, strings), here is a sample of what we need:
 
-Nickel provides a lot of types, but also structures. In our case,
-we want `ssh-keys` to allow multiple keys, so this will be an array,
-`name` is a sequence of characters, so a type `Str` (for String),
-the field `is-admin` can be either true or false, so this is a
-`Bool`, finally, `extra-groups` is a list of group names,
-which translates as an array of string.
+```yaml
+users:
+  - name: Aisha
+    is-admin: true
+    ssh-keys:
+      - AAAAApqCA8oKAB5S/47f...... aisha@work
+  - name: Violet
+    extra-groups:
+      - accounting
+```
 
-We can also mark fields as optional, so you don't have
-to fill them if they don't have any value, `extra-groups` and `ssh-keys`
-can be empty in our case, we will make them optional by applying `|
-optional`. 
+We can spot attributes such as `name`, `ssh-keys`, `is-admin` or
+`extra-groups`. To create a Nickel contract that correctly specifies
+the shape of this data, we need to think about the type needed for
+each attribute.
 
-The field `is-admin` must be present in the YAML we output, but
-for most users it will be set to `false`. We can give the field
-a default value of `false`, which means it only needs to be 
-present in Nickel code when the user is actually an admin.
+For example, the attribute `name` is a string, which translates as `Str`
+in Nickel. Meanwhile, `ssh-keys` must allow multiple keys, so this is
+an list of strings, written as `Array Str` in Nickel. `is-admin` is a
+boolean, written as `Bool`. Finally, `extra-groups` is a list of group
+names, which is the same type used for `ssh-keys`, we need `Array Str`.
+
+We can also mark fields as optional so you won't have to explicitly write
+about them if they don't have any value, `extra-groups` and `ssh-keys`
+can be empty in the example, we will make them optional by using the
+`optional` keyword.
+
+The field `is-admin` must always be present in the YAML file, but
+for most of our users it will be set to `false`. Fortunately, we can
+assign the default value `false` to this field , which means you only
+need write `is-admin = true` in Nickel code when the user is actually
+an administrator.
 
 ## Step 3: Write a contract
 
--- a/doc/manual/tuto.md
+++ b/doc/manual/tuto.md
@@ -9,8 +9,9 @@ users, and then export it as a YAML file.
 
 ## Step 1: Install nickel
 
-The first step is to make Nickel available on your system. Please refer to the different installation methods documented in [Getting
-started](https://nickel-lang.org/getting-started/#getting-started).
+The first step is to make Nickel available on your system. Please refer
+to the different installation methods documented in
+[Getting started](https://nickel-lang.org/getting-started/#getting-started).
 
 ## Step 2: Think about the schema
 
@@ -57,9 +58,8 @@ defining fields and associated constraints such as a type, marking
 the attribute optional, and a default value if there is one.
 
 Please note that using a contract isn't mandatory per-se to use Nickel,
-but it will allow you to validate your input data (See Step 6 for an
-example).
-
+but it will allow you to validate your input data
+(See [Step 6](#step-6-try-to-make-a-mistake) for an example).
 
 ```nickel
 {
--- a/doc/manual/tuto.md
+++ b/doc/manual/tuto.md
@@ -4,8 +4,8 @@ slug: tutorial
 
 # Managing users
 
-In this tutorial, you will learn how to use Nickel to manage a list of users,
-and then export it as a YAML file for a program consuming it.
+In this tutorial, you will learn how to use Nickel to manage a list of
+users, and then export it as a YAML file for a program consuming it.
 
 ## Step 1: Install nickel
 
@@ -36,27 +36,31 @@ the shape of this data, we need to think about the type needed for
 each attribute.
 
 For example, the attribute `name` is a string, which translates to `Str`
-in Nickel. Meanwhile, `ssh-keys` must allow multiple keys, so this is
-a list of strings, written as `Array Str` in Nickel. The field `is-admin` is a
-boolean, written as `Bool`. Finally, `extra-groups` is a list of group
-names, so we need `Array Str`, the same type used for `ssh-keys`.
+in Nickel. Meanwhile, `ssh-keys` must allow multiple keys, so this is a
+list of strings, written as `Array Str` in Nickel. The field `is-admin`
+is a boolean, written as `Bool`. Finally, `extra-groups` is a list of
+group names, so we need `Array Str`, the same type used for `ssh-keys`.
 
-We can also mark fields as `optional` so you won't have to explicitly write
-them if they don't have any value. In the example,`extra-groups` and `ssh-keys`
-can be empty, so we will mark them as optional using the
+We can also mark fields as `optional` so you won't have to explicitly
+write them if they don't have any value. In the example,`extra-groups`
+and `ssh-keys` can be empty, so we will mark them as optional using the
 `optional` keyword.
 
-The field `is-admin` must always be present in the YAML file, but
-for most of our users it will be set to `false`. Fortunately, we can
-assign the default value `false` to this field , which means you only
-need to write `is-admin = true` in Nickel code when the user is actually
+The field `is-admin` must always be present in the YAML file, but for
+most of our users it will be set to `false`. Fortunately, we can assign
+the default value `false` to this field , which means you only need
+to write `is-admin = true` in Nickel code when the user is actually
 an administrator.
 
 ## Step 3: Write a contract
 
-Create a text file named `users-schema.ncl`. We will
-write our contract defining attributes and associated constraints such
-as a type, marking the attribute optional, and a default value if there is one.
+Create a text file named `users-schema.ncl`. We will write our contract
+defining attributes and associated constraints such as a type, marking
+the attribute optional, and a default value if there is one.
+
+Please note that using a contract isn't mandatory per-se to use Nickel,
+but it will allow you to validate your input data (See Step 6 for an
+example).
 
 Please note that using a contract isn't mandatory per-se to use Nickel,
 but it will allow you to validate your input data (See Step 6 for an
@@ -115,8 +119,8 @@ let {UserSchema, ..} = import "users-schema.ncl" in
 
 ## Step 5: Export as YAML
 
-By default, nickel exports data as JSON. But we can change that using the extra
-parameter `--format yaml` to export as YAML.
+By default, nickel exports data as JSON. But we can change that using
+the extra parameter `--format yaml` to export as YAML.
 
 ```shell
 nickel -f users.ncl export --export yaml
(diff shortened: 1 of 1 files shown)


djacu:
looks pretty good so far!

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+and then export it as a YAML file for a program consuming it.
+
+## Step 1: Install nickel
+
+## Step 2: Think about the schema
+
+Our current YAML has different fields, such as `name`, `ssh-keys`,
+`is-admin` or `extra-groups`. To create a Nickel contract that produce
+what we want, and validates input, we need to think about the type of
+each field.
```
matthew-healy, line 17:
```suggestion
`is-admin` or `extra-groups`. To create a Nickel contract that correctly
specifies the shape of this data, we first need to think about the
types of each field
```

I'm not sure it's entirely correct to say the contract "produces" the data we want - it really just does the specification & validation.

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+
+Nickel provides a lot of types, but also structures. In our case,
```
matthew-healy, line 19: > Nickel provides a lot of types, but also structures.

I'm not entirely sure what you mean by "structures" here. I think this sentence could possibly be cut.

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+we want `ssh-keys` to allow multiple keys, so this will be an array,
+`name` is a sequence of characters, so a type `Str` (for String),
+the field `is-admin` can be either true or false, so this is a
+`Bool`, finally, `extra-groups` is a list of group names,
+which translates as an array of string.
+
+We can also define some fields optionals, so you don't have
```
matthew-healy, line 26:
```suggestion
We can also mark fields as optional, so you don't have
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+to fill them if they don't have any value, `extra-groups` and `ssh-keys`
+can be empty in our case, we will make them optional by applying `|
+optional`. The field `is-admin` must be provided to our program consuming
+the YAML, but we don't want to fill it when it's false, we can apply
+a default value of `false`, so we will only define it in our file when
+the user is actually an admin, this makes the file maintenance easier.
```
matthew-healy, line 32:
```suggestion
optional`. 

The field `is-admin` must be present in the YAML we output, but
for most users it will be set to `false`. We can give the field
a default value of `false`, which means it only needs to be 
present in Nickel code when the user is actually an admin.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Step 3: Write a contract
+
```
matthew-healy, line 35: We should possibly mention here that we want to create a `users-schema.ncl` & add this code there, otherwise there's a chance the reader puts this in a different file & gets confused at the next step.

doc/manual/tuto.md
```diff
@@ -0,0 +1,112 @@
...
+    {
+      name = "Grace",
+      is-admin = true,
+      extra-groups = [ "administrative" ],
+    },
+ ]
+}
+```
+
```
matthew-healy, line 82: It could make sense to add a step here where we deliberately enter some incorrect data in order to see the resulting error. (Though, on the other hand, that might also be best left to a different tutorial).

matthew-healy (commented):
This is great, thank you so much for doing this!

I left a couple of comments, mostly just with grammar & phrasing suggestions. None of them need to block merging though.

I do think providing an example of the YAML schema before jumping into the Nickel code will be helpful, as it provides clear context for what we're actually trying to achieve.

doc/manual/tuto.md
matthew-healy: Yeah, I see what you mean. I think that since the "structure" terminology isn't one that's widely used or well-defined it could potentially create additional confusion, but ultimately it's up to you how you want to phrase this. :smile:

rapenne-s:
I think this first tutorial is finished, it's ready to be reviewed.

The thing that need the most care in my opinion is using the correct words for contract attributes, records attributes etc... I'm not sure I'm using the right words, and nickel-lang.org didn't help me much to figure the correct names, maybe a glossary would be very handy

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
+---
+slug: tutorial
+---
+
+# Managing users
+
+In this tutorial, you will learn how use Nickel to manage a list of users,
```
vkleen, line 7:
```suggestion
In this tutorial, you will learn how to use Nickel to manage a list of users,
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+
+The first step is to obtain nickel on your system. There are different
+methods to install Nickel, they are all covered on the [Getting
+started](https://nickel-lang.org/getting-started/#getting-started)
+page of Nickel project website.
+
+## Step 2: Think about the schema
+
+The YAML we want to produce as different fiels with different types
+(lists, booleans, strings), here is a sample of what we need:
```
vkleen, line 20:
```suggestion
The YAML we want to produce has different fields with different types
(lists, booleans, strings). Here is a sample of what we need:
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+and then export it as a YAML file for a program consuming it.
+
+## Step 1: Install nickel
+
+The first step is to obtain nickel on your system. There are different
+methods to install Nickel, they are all covered on the [Getting
+started](https://nickel-lang.org/getting-started/#getting-started)
+page of Nickel project website.
+
+## Step 2: Think about the schema
```
vkleen, line 17: Maybe it's worth adding a paragraph explaining that writing a contract is optional. That is, you could just write out the users configuration as a list of records and export that. The reason to spend the time thinking about a contract is to catch mistakes like in step 6

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+We can spot attributes such as `name`, `ssh-keys`, `is-admin` or
+`extra-groups`. To create a Nickel contract that correctly specifies
+the shape of this data, we need to think about the type needed for
+each attribute.
+
+For example, the attribute `name` is a string, which translates as `Str`
+in Nickel. Meanwhile, `ssh-keys` must allow multiple keys, so this is
+an list of strings, written as `Array Str` in Nickel. `is-admin` is a
+boolean, written as `Bool`. Finally, `extra-groups` is a list of group
+names, which is the same type used for `ssh-keys`, we need `Array Str`.
```
vkleen, line 42:
```suggestion
For example, the attribute `name` is a string, which translates to `Str`
in Nickel. Meanwhile, `ssh-keys` must allow multiple keys, so this is
a list of strings, written as `Array Str` in Nickel. The field `is-admin` is a
boolean, written as `Bool`. Finally, `extra-groups` is a list of group
names, so we need `Array Str`, the same type used for `ssh-keys`.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+
+We can also mark fields as optional so you won't have to explicitly write
+about them if they don't have any value, `extra-groups` and `ssh-keys`
+can be empty in the example, we will make them optional by using the
+`optional` keyword.
```
vkleen, line 47:
```suggestion
We can also mark fields as `optional` so you won't have to explicitly write
them if they don't have any value. In the example,`extra-groups` and `ssh-keys`
can be empty, so we will mark them as optional using the
`optional` keyword.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+
+The field `is-admin` must always be present in the YAML file, but
+for most of our users it will be set to `false`. Fortunately, we can
+assign the default value `false` to this field , which means you only
+need write `is-admin = true` in Nickel code when the user is actually
+an administrator.
```
vkleen, line 53:
```suggestion
need to write `is-admin = true` in Nickel code when the user is actually
an administrator.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Step 3: Write a contract
+
+Create a text file named `users-schema.ncl`, we will
+write our contract defining attributes and associated constraints such
+as a type, marking the attribute optional, and a default value if any.
```
vkleen, line 59:
```suggestion
Create a text file named `users-schema.ncl`. We will
write our contract defining attributes and associated constraints such
as a type, marking the attribute optional, and a default value if there is one.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+      extra-groups = [ "administrative" ],
+    },
+ ]
+}
+```
+
+## Step 5: Export as YAML
+
+By default, nickel exports as JSON type, so we need to use the extra
+parameter `--format yaml` to export as YAML.
```
vkleen, line 115:
```suggestion
By default, nickel exports data as JSON. But we can change that using the extra
parameter `--format yaml` to export as YAML.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,179 @@
...
+      - administrative
+    is-admin: true
+    name: Grace
+```
+
+## Step 6: Try to make a mistake
+
+In this extra step, we will make a mistake on purpose in the file
+`users.ncl` and try to export the data as YAML. We will see what happens
+when Nickel detects that a contract use is incorrect.
```
vkleen, line 147:
```suggestion
In this extra step, we will make a mistake on purpose in the file
`users.ncl` and try to export the data as YAML. We will see what happens
when Nickel detects that a value doesn't satisfy a contract.

**Review by vkleen — APPROVED** [review]
Awesome! I've found a few typos and left some suggestions for formulations. Feel free to mix and match, or even ignore them if you think something else would be better.

Also, I think we should explicitly say that writing the contract is not necessary for exporting the users data from Nickel. Rather, the contract gives you the extra power of validation in Nickel.

File: doc/manual/tuto.md (hunk 9 above)
Comment (rapenne-s) [reply doc/manual/tuto.md:17]: Good

File: doc/manual/tuto.md [hunk 16]
```diff
@@ -0,0 +1,187 @@
+---
+slug: tutorial
+---
+
+# Managing users
+
+In this tutorial, you will learn how to use Nickel to manage a list of
+users, and then export it as a YAML file for a program consuming it.
```
yannham, line 8: The "for a program consuming it" sounds a little strange to the ear, although I don't think it's incorrect per se. Maybe "for another program to consume"? Or just get rid of the end of the sentence?

```suggestion
users, and then export it as a YAML file.
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Step 1: Install nickel
+
+The first step is to obtain nickel on your system. There are different
+methods to install Nickel, they are all covered on the [Getting
+started](https://nickel-lang.org/getting-started/#getting-started)
+page of Nickel project website.
```
yannham, line 15:
```suggestion
The first step is to make Nickel available on your system. Please refer to the different installation methods documented in [Getting
started](https://nickel-lang.org/getting-started/#getting-started).
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+
+We can also mark fields as `optional` so you won't have to explicitly
+write them if they don't have any value. In the example,`extra-groups`
+and `ssh-keys` can be empty, so we will mark them as optional using the
+`optional` keyword.
+
+The field `is-admin` must always be present in the YAML file, but for
+most of our users it will be set to `false`. Fortunately, we can assign
+the default value `false` to this field , which means you only need
+to write `is-admin = true` in Nickel code when the user is actually
```
yannham, line 52:
```suggestion
to write `is-admin = true` when the user is actually
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+an administrator.
+
+## Step 3: Write a contract
+
+Create a text file named `users-schema.ncl`. We will write our contract
+defining attributes and associated constraints such as a type, marking
```
yannham, line 58:
```suggestion
defining fields and associated constraints such as a type, marking
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+    extra-groups:
+      - accounting
+```
+
+We can spot attributes such as `name`, `ssh-keys`, `is-admin` or
+`extra-groups`. To create a Nickel contract that correctly specifies
+the shape of this data, we need to think about the type needed for
+each attribute.
+
+For example, the attribute `name` is a string, which translates to `Str`
```
yannham, line 38:
```suggestion
For example, the field `name` is a string, which translates to `Str`
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+  - name: Aisha
+    is-admin: true
+    ssh-keys:
+      - AAAAApqCA8oKAB5S/47f...... aisha@work
+  - name: Violet
+    extra-groups:
+      - accounting
+```
+
+We can spot attributes such as `name`, `ssh-keys`, `is-admin` or
```
yannham, line 33: Just to be consistent with the documentation and the naming in Nickel in general (attributes is Nix terminology). I think we even use `attribute` sometimes for metadata like `optional` or `default`.

```suggestion
We can spot fields such as `name`, `ssh-keys`, `is-admin` or
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+an administrator.
+
+## Step 3: Write a contract
+
+Create a text file named `users-schema.ncl`. We will write our contract
+defining attributes and associated constraints such as a type, marking
+the attribute optional, and a default value if there is one.
+
+Please note that using a contract isn't mandatory per-se to use Nickel,
+but it will allow you to validate your input data (See Step 6 for an
```
yannham, line 62: We can make a link to `Step 6`. I don't remember the exact syntax, but it's probably something like `[Step 6](#step- 6-try-to-make-a-mistake]`. You might find examples of such anchors in other pages of the manual.

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+example).
+
+Please note that using a contract isn't mandatory per-se to use Nickel,
+but it will allow you to validate your input data (See Step 6 for an
+example).
```
yannham, line 67:
```suggestion
```

Duplication

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+
+```nickel
+{
+  UserSchema =
+  {
+    name | Str,
+    ssh-keys | Array Str | optional,
```
yannham, line 74: We don't have clear formatting rules yet (but soonish maybe, with topiary :slightly_smiling_face:  ) but I think we should abide by the rule of making a multiline block if there is more than one metadata; otherwise, I find it hard to read a bit overwhelming.

```suggestion
    ssh-keys
        | Array Str
        | optional,
```

doc/manual/tuto.md
```diff
@@ -0,0 +1,187 @@
...
+    is-admin | Bool | default = false,
+    extra-groups | Array Str | optional,
```
yannham, line 76:
```suggestion
    is-admin
      | Bool
      | default = false,
    extra-groups
      | Array Str
      | optional,
```

yannham (approved):

doc/manual/tuto.md
rapenne-s: I wanted to give some kind of background about the "why we need to generate a YAML file", but I suppose we can drop the end of the sentence

doc/manual/tuto.md
rapenne-s: okay!

doc/manual/tuto.md
rapenne-s: done

PR #1075, opened 2023-01-25
Improve the performance of array.foldl and array.fold

vkleen:
This change improves the performance of the standard library fold functions from "Eats all my RAM before finishing" to
```
❯ time nickel <<<'array.generate function.id 16384 |> array.foldl (+) 0'
134209536
nickel <<< 'array.generate function.id 16384 |> array.foldl (+) 0'  0.49s user 0.02s system 99% cpu 0.512 total
```

There are still performance problems with using the standard library array functions, for example
```
❯ time nickel <<<'array.generate (function.const []) 16384 |> array.flatten'
[  ]
nickel <<< 'array.generate (function.const []) 16384 |> array.flatten'  6.66s user 0.04s system 99% cpu 6.694 total
```
while inlining `array.flatten` without a contract annotation runs in about 0.4s. But this change already improves the situation from not terminating in a reasonable timeframe to just being slow.

### DIFF
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -108,12 +108,17 @@
             6
         ```
         "%
-      = fun f acc l =>
-        if %length% l == 0 then
-          acc
+    = fun f acc l =>
+        let length = %length% l in
+        if length == 0 then acc
         else
-          let newAcc = f acc (%head% l) in
-          %seq% newAcc (foldl f newAcc (%tail% l)),
+          let rec loop = fun acc n => 
+            if n == length then acc
+            else 
+              let next_acc = %elem_at% l n |> f acc in
+              %seq% next_acc (loop next_acc (n+1))
+          in
+          loop acc 0,
 
     fold : forall a b. (a -> b -> b) -> b -> Array a -> b
       | doc m%"
@@ -128,10 +133,12 @@
         ```
         "%
       = fun f fst l =>
-        if %length% l == 0 then
-          fst
-        else
-          f (%head% l) (fold f fst (%tail% l)),
+        let rec go = fun n =>
+          if n == %length% l then
+            fst 
+          else
+            go (n+1) |> f (%elem_at% l n)
+        in go 0,
 
     cons : forall a. a -> Array a -> Array a
       | doc m%"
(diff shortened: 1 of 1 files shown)


stdlib/array.ncl
```diff
@@ -128,10 +133,12 @@
         ```
         "%
       = fun f fst l =>
-        if %length% l == 0 then
-          fst
-        else
-          f (%head% l) (fold f fst (%tail% l)),
+        let rec go = fun n =>
+          if n == %length% l then
```
matthew-healy, line 137: I don't think it'll necessarily make a huge difference overall but I believe adding a `let length = %length% l in` before defining `go` here will be marginally more performant than repeatedly calling the `%length%` primop in the loop.

matthew-healy (approved):
Nice! :ok_hand:

ebresafegaga (approved):

stdlib/array.ncl
vkleen: Turns out the difference isn't really measurable within 1% or so. But I like having it outside on principle 👍

stdlib/array.ncl
```diff
@@ -108,12 +108,17 @@
...
         else
-          let newAcc = f acc (%head% l) in
-          %seq% newAcc (foldl f newAcc (%tail% l)),
+          let rec loop = fun acc n => 
+            if n == length then acc
+            else 
+              let next_acc = %elem_at% l n |> f acc in
+              %seq% next_acc (loop next_acc (n+1))
+          in
+          loop acc 0,
```
yannham, line 121: I don't really have a preference for either `go` or `loop` but we should probably use the same naming style for both functions.

```suggestion
          let rec go = fun acc n => 
            if n == length then acc
            else 
              let next_acc = %elem_at% l n |> f acc in
              %seq% next_acc (go next_acc (n+1))
          in
          loop acc 0,
```

stdlib/array.ncl
```diff
@@ -128,10 +133,13 @@
...
-          fst
-        else
-          f (%head% l) (fold f fst (%tail% l)),
+        let length = %length% l in
+        let rec go = fun n =>
+          if n == length then
+            fst 
+          else
+            go (n+1) |> f (%elem_at% l n)
+        in go 0,
```
yannham, line 142:
```suggestion
            go (n+1)
            |> f (%elem_at% l n)
        in go 0,
```

yannham (approved):

PR #1080, opened 2023-01-27
Treat variables bound by record deconstruction like record static indexing

### LINKED ISSUE #1011: LSP server doesn't give completion on pattern variables
**Describe the bug**
Say we have this program: 
```
let {a, ..} = {a.b = 10 } in 
a[.]
```
We should get `b` as a completion, but in this case, we don't.

**Additional context**
This actually happens because the `RichTerm` associated with each pattern binding is not linearized.

ebresafegaga:
Currently, we can't get completion on variables bound by pattern matching, e.g:
```nickel
let {a, ..} = {a.b = 10 } in 
a[.]
```

With this change, we support completion for all variables (apart from variables bound by "rest" patterns) bound by a pattern deconstruction. This will work for both "context completion" and record completion. e.g:
```nickel 
let {Config, ..} = import "lib.ncl" in 
{
   data | Config = {
      [.]
   }
} 
```

Closes #1011

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -198,7 +198,7 @@ impl<'b> Building<'b> {
                         _ => break,
                     }
                 }
-                // return the `prev_item` becuase that was the record field we want to 
+                // return the `prev_item` becuase that was the record field we want to
                 // resolve, and `curr_item` points to the record field's value.
                 Some(prev_item)
             }
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -162,7 +162,7 @@ impl<'b> Building<'b> {
                     .as_option()
                     .and_then(|value_index| self.get_item_kind(current_file, value_index))
             }
-            TermKind::Declaration(ident, _, ValueState::Known(value), true) => {
+            TermKind::Declaration(_, _, ValueState::Known(value), Some(ident)) => {
                 let item = self.get_item_kind(current_file, *value)?;
                 let item = self.resolve_reference(current_file, item)?;
                 match item {
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -175,9 +175,9 @@ impl<'b> Building<'b> {
 
                 let mut ids = idents.clone();
                 // (previous value, current value)
-                let mut result = (item, item);
+                let (mut prev_item, mut curr_item) = (item, item);
                 while let Some(id) = ids.pop() {
-                    match result.1 {
+                    match curr_item {
                         TermKind::Record(ref fields) => {
                             let item = fields.get(&id)?;
                             let item = self.get_item_kind(current_file, *item)?;
@@ -186,9 +186,9 @@ impl<'b> Building<'b> {
                                     value: ValueState::Known(next_item),
                                     ..
                                 } => {
-                                    result.0 = item;
+                                    prev_item = item;
                                     let item = self.get_item_kind(current_file, *next_item)?;
-                                    result.1 = item;
+                                    curr_item = item;
                                     continue;
                                 }
                                 // the value from a record is always a record field
@@ -198,9 +198,9 @@ impl<'b> Building<'b> {
                         _ => break,
                     }
                 }
-                // return the previous value becuase that was the record field we were
-                // looking for
-                Some(result.0)
+                // return the `prev_item` becuase that was the record field we want to 
+                // resolve, and `curr_item` points to the record field's value.
+                Some(prev_item)
             }
             // if declaration is a let binding resolve its value
             TermKind::Declaration {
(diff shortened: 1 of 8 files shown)


lsp/nls/src/linearization/building.rs
```diff
@@ -162,8 +162,41 @@ impl<'b> Building<'b> {
                     .as_option()
                     .and_then(|value_index| self.get_item_kind(current_file, value_index))
             }
+            TermKind::Declaration(_, _, ValueState::Known(value), Some(idents)) => {
+                let item = self.get_item_kind(current_file, *value)?;
+                let item = self.resolve_reference(current_file, item)?;
+
+                let mut ids = idents.clone();
+                // (previous value, current value)
+                let mut result = (item, item);
```
yannham, line 171: IMHO it would be clearer to have two separate variables `prev_value` and `curr_value` (or whatever you call them) rather than using an anonymous tuple. Unless there is a good reason to do so (like calling a stdlib functions that expect a tuple or something).

lsp/nls/src/linearization/interface.rs
```diff
@@ -24,7 +24,12 @@ impl ResolutionState for Resolved {}
 /// Can be extended later to represent Contracts, Records, etc.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum TermKind {
-    Declaration(Ident, Vec<ItemId>, ValueState),
+    Declaration(
+        Ident,
+        Vec<ItemId>,
+        ValueState,
+        /* pattern binding? */ Option<Vec<Ident>>,
```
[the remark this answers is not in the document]
ebresafegaga: This is the path to a bound variable. E.g if we have `let { a = {b = {c, ..}, ..}, ..} = ... in ...`, the path will be `vec![a, b]`, and the ident will be `c`

lsp/nls/src/linearization/interface.rs
```diff
@@ -24,7 +24,12 @@ impl ResolutionState for Resolved {}
 /// Can be extended later to represent Contracts, Records, etc.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum TermKind {
-    Declaration(Ident, Vec<ItemId>, ValueState),
+    Declaration {
+        id: Ident,
+        usages: Vec<ItemId>,
+        value: ValueState,
+        pattern_bindings: Option<Vec<Ident>>,
```
yannham, line 31: Then, I think you should add what you just explained in an answer to my comment directly in the code, so that I'm not the only one to enjoy it :slightly_smiling_face: 

As a side note, `pattern_bindings` doesn't reflect what this variables really represent IMHO. There is a notion of `path`, so `path` should probably be in the name. There aren't bindings, because none but potentially the last element of the path is actually bound in the environment.

ebresafegaga: Agreed. Thanks! 🙂

lsp/nls/src/linearization/building.rs
```diff
@@ -162,10 +165,48 @@ impl<'b> Building<'b> {
...
+            TermKind::Declaration {
+                value: ValueState::Known(value),
+                path: Some(idents),
+                ..
+            } => {
+                let item = self.get_item_kind(current_file, *value)?;
+                let item = self.resolve_reference(current_file, item)?;
+
+                let mut ids = idents.clone();
+                // (previous value, current value)
```
yannham, line 177:
```suggestion
```

lsp/nls/src/linearization/building.rs
```diff
@@ -162,10 +165,48 @@ impl<'b> Building<'b> {
...
+                let (mut prev_item, mut curr_item) = (item, item);
+                while let Some(id) = ids.pop() {
+                    match curr_item {
+                        TermKind::Record(ref fields) => {
+                            let item = fields.get(&id)?;
+                            let item = self.get_item_kind(current_file, *item)?;
```
yannham, line 183: Nitpick, but you're calling two things with different types with the same name (which is ok, sometimes, granted).

```suggestion
                            let item_kind = self.get_item_kind(current_file, *item)?;
```

lsp/nls/src/linearization/building.rs
```diff
@@ -162,10 +165,48 @@ impl<'b> Building<'b> {
...
+                                    continue;
+                                }
+                                // the value from a record is always a record field
+                                _ => unreachable!(),
+                            }
+                        }
+                        _ => break,
+                    }
+                }
+                // return the `prev_item` becuase that was the record field we want to
```
yannham, line 201:
```suggestion
                // return the `prev_item` because that was the record field we wanted to
```

lsp/nls/src/linearization/building.rs
```diff
@@ -162,10 +165,48 @@ impl<'b> Building<'b> {
...
+                // resolve, and `curr_item` points to the record field's value.
+                Some(prev_item)
             }
+            // if declaration is a let binding resolve its value
```
yannham, line 205:
```suggestion
            // if declaration is a let binding, resolve its value
```

yannham (approved):

PR #1081, opened 2023-01-27
Make the Nickel LSP available in packages

vkleen:
Add `lsp-nls` as a field in the `packages` output of the flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -201,13 +201,14 @@
             buildInputs = [ pkgs.python3 ];
           };
 
-          buildPackage = packageName:
+          buildPackage = pname:
             craneLib.buildPackage {
               inherit
+                pname
                 src
                 cargoArtifacts;
 
-              cargoExtraArgs = "${cargoBuildExtraArgs} --package ${packageName}";
+              cargoExtraArgs = "${cargoBuildExtraArgs} --package ${pname}";
             };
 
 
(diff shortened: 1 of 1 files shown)


yannham (changes requested):
The change is fine but the documentation inside `README.md`, and more specifically inside `lsp/README.md` and maybe `lsp/nls/README.md` about the installation of the LSP might need to be updated.

vkleen:
Giving it a little more thought, I made the default package in `flake.nix` a `buildEnv` of `nickel` together with `lsp-nls`. That way, users don't have to think about installing both and the READMEs stay correct.

yannham (approved):
In the end your initial approach is the right one, I think: right now we build and install the LSP when getting `nickel` from a flake for e.g. `nickel-nix`, which sounds wrong. But I understand you don't wann do that now :slightly_smiling_face: for the time being, this PR is a strict improvement on the previous situation.

vkleen:
I'm not so sure; `nickel-nix` can just use the `nickel` output to only get Nickel without the LSP. It's only the default output that builds both and links them together.

yannham:
> I'm not so sure; nickel-nix can just use the nickel output to only get Nickel without the LSP. It's only the default output that builds both and links them together.

Ah, you're right. Sounds reasonable, then!

PR #1083, opened 2023-01-27
Implementation of the update operator for nix/nickel

francois-caddet:
Now nickel can evaluate `{...} // {...}` from Nix.

### DIFF
--- a/stdlib/compat.ncl
+++ b/stdlib/compat.ncl
@@ -11,7 +11,7 @@ compat | doc "Nix compatibility layer. This library should not be used by Nickel
 
   # The update operator of Nix `//`. It's a "general form" of the `record.update` of Nickel.
   # TODO: May be intresting to be addapted and integrated the the actual Nickel stdlib.
-  update_from_rec = fun r1 r2 =>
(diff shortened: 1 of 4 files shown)


stdlib/compat.ncl
```diff
@@ -9,6 +9,14 @@ compat | doc "Nix compatibility layer. This library should not be used by Nickel
     then a ++ b
     else a + b,
 
+  # The update operator of Nix `//`. It's a "general form" of the `record.update` of Nickel.
+  # TODO: May be intresting to be addapted and integrated the the actual Nickel stdlib.
+  update_from_rec = fun r1 r2 =>
```
yannham, line 14: Just a suggestion, but for the naming: what about `update_all`? Or at least `update_from_record`? `rec` is a bit ambiguous in Nickel, as it also stands for _recursive_.

yannham (approved):

stdlib/compat.ncl
francois-caddet: I like the `update_all`

PR #1087, opened 2023-01-31
Add tutorial on stable branch

rapenne-s:
this is a backport of the tutorial to the stable branch, in order to display it on nickel-lang.org

### DIFF
--- a/doc/manual/tuto.md
+++ b/doc/manual/tuto.md
@@ -9,8 +9,9 @@ users, and then export it as a YAML file.
 
 ## Step 1: Install nickel
 
-The first step is to make Nickel available on your system. Please refer to the different installation methods documented in [Getting
-started](https://nickel-lang.org/getting-started/#getting-started).
+The first step is to make Nickel available on your system. Please refer
+to the different installation methods documented in
+[Getting started](https://nickel-lang.org/getting-started/#getting-started).
 
 ## Step 2: Think about the schema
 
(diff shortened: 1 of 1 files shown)


yannham (approved):
Now that I see it again, I think it could be nice to add documentation next to each field of the schema with `doc`. Probably some won't be very useful (like `name`), but if this tutorial is to appear on the website, it has to set the example, and in an ideal world users should have the - almost Pavlovian - reflex to document contracts and contract fields.

rapenne-s:
> Now that I see it again, I think it could be nice to add documentation next to each field of the schema with `doc`. Probably some won't be very useful (like `name`), but if this tutorial is to appear on the website, it has to set the example, and in an ideal world users should have the - almost Pavlovian - reflex to document contracts and contract fields.

I thought about it, but I was afraid to clutter the tutorial. Do you think it should be directly in the contract, or as an extra step to not overwhelm the reader?

yannham:
> I thought about it, but I was afraid to clutter the tutorial. Do you think it should be directly in the contract, or as an extra step to not overwhelm the reader?

Mmh, I see what you mean. Then yes, maybe a last step like "Bonus step", "be nice with your future you, colleagues or open source collaborators and document your contract" could help have smaller, more digestible chunks :+1:

yannham:
@rapenne-s how do you want to proceed? We can merge this first, then add the bonus step. You'll probably have to cherry pick the commit on master as well.

rapenne-s:
that looks fine, I'll also prepare a PR for the website

PR #1088, opened 2023-01-31
Support completion for field names inside recursive records.

ebresafegaga:
Implements item 4 in #877 

As a consequence of this change, we get completion on stdlib names (such as `string`, `array`, `contracts`, etc), because they are record field names which are in scope.

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -561,7 +561,8 @@ fn get_completion_identifiers(
                     .get_in_scope(item, &server.lin_cache)
                     .iter()
                     .filter_map(|i| match i.kind {
-                        TermKind::Declaration(ident, _, _) => Some(IdentWithType {
+                        TermKind::Declaration(ident, _, _)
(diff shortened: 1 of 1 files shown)


yannham:
> As a consequence of this change, we get completion on stdlib names (such as string, array, contracts, etc), because they are record field names which are in scope.

:sunglasses:

yannham (approved):
Code looks good, but I tried this inside the mantis benchmark and it didn't work. That being said, I don't know if it didn't because it was too slow, or something related to coc.nvim not liking the LSP. Does it work for you @ebresafegaga ? If you go inside `types.ncl`, pick a random record, and add a new filed that refers to one of its sibling (which is a record), do you get completion?

ebresafegaga:
I'm not sure what is going on, but I know in some cases, if the program in incomplete, the parser will not be able to correctly parse the record with good error recovery. 

I get completion if I do this (note the comma at the end):
<br />
image

yannham:
Got it. In any case incomplete programs are a more general problem, and orthogonal to this PR. It's good to go for me :+1:

PR #1092, opened 2023-02-01
Rename stdlib functions

vkleen:


stdlib/array.ncl
```diff
@@ -120,14 +139,14 @@
           in
           go acc 0,
 
-    fold : forall a b. (a -> b -> b) -> b -> Array a -> b
+    fold_right : forall a b. (a -> b -> b) -> b -> Array a -> b
       | doc m%"
         Fold a function over a array.
```
rapenne-s, line 144: I have actually no idea what "fold" means, and this sentence doesn't help me understanding what it's doing. Is it simply common knowledge I'm missing?

vkleen: If I remember right, we had a discussion during the Nickel hour that made these changes about that. We couldn't come up with a good, concise explanation on the spot and left it this way.

While I think most functional programmers will know what "folds" are, we probably should include some more explanation in this documentation. I'm just not sure what to write exactly 😅

aspiwack: For the record, here is Ocaml's documentation for `fold_right`

> f`old_right f [a1; ...; an] init` is `f a1 (f a2 (... (f an init) ...))`. Not tail-recursive.

yannham: Actually, we've used something similar from the beginning, because I also find the documentation of Ocaml's folds to be quite good :slightly_smiling_face: . If you scroll down this file, the full documentation in Nickel is in fact:

>Fold a function over a array.
>`fold f init [x1, x2, ..., xn]` results in `f x1 (f x2 (... (f xn init) ...))`.
>
>For example:
>```
>  fold (fun e acc => acc @ [e]) [] [ 1, 2, 3 ] =>
>    ((([] @ [3]) @ [2]) @ [1]) =>                
>    [ 3, 2, 1 ]
>```

stdlib/array.ncl
```diff
@@ -70,13 +89,13 @@
         "%
       = fun f l => %map% l f,
 
-    elem_at : forall a. Num -> Array a -> a
+    at : forall a. Num -> Array a -> a
```
matthew-healy, line 92: Maybe just personal preference, but I think I prefer `elem_at` to just `at` here.

matthew-healy (commented):
The renamings themselves mostly look good, though ideally I don't think we should merge this with commented-out functions present in the stdlib. Happy to approve if those are removed though.

stdlib/array.ncl
```diff
@@ -45,6 +49,21 @@
         "%
       = fun l => %tail% l,
 
+    tail_but_reverse_bikeshed_me : forall a. Array a -> Array a
+      | doc "TODO"
+      # TODO: we need a slice primop
+      = fun l => l
+        |> reverse
+        |> tail
+        |> reverse,
```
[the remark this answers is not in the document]
yannham: I proposed `cotail`, but people didn't like it :/

`pop` would be confusing because you don't know in an immutable language if it returns the element or the array.

Note that we can in fact make it `O(1)` in Nickel, but we would need a primop.

Another possibility is to have more explicit names like `without_last` and `without_first`. Less common than `tail` for the latter, but at least symmetric and might be more intuitive for non functional programmers.

stdlib/array.ncl
yannham: Another, similar possibility: `get`.

stdlib/array.ncl
yannham: Revisiting this, I think we can still expand a bit for non functional programmers. Maybe add a line or two explaining that fold are more or less like `for` loops, with examples. @rapenne-s has a point that `fold` might not be something very intuitive to DevOps people.

stdlib/array.ncl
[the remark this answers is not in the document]
vkleen: `pop_front` and `pop_back` would also be reasonable candidates.

stdlib/array.ncl
vkleen: I like leaving it just `at`, since almost all uses will be a fully qualified `array.at` or maybe later `std.array.at`. So we're not polluting the global namespace with extremely short names and I don't think making it longer gives any more clarity.

stdlib/array.ncl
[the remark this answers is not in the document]
vkleen: The newest option: `drop_first` and `drop_last`. Performance shouldn't really play into this decision since we're planning to add an array slice primop anyway. That will make both functions take constant time. Namingwise, `drop` makes it clear that the array is returned instead of the element and naming between `drop_first` and `drop_last` is uniform an unambiguous.

stdlib/string.ncl
```diff
@@ -202,9 +202,9 @@
       "%
     = fun s => %str_chars% s,
 
-    code | CharLiteral -> Num
+    codepoint | Character -> Num
     | doc m%%"
-      Results in the ascii code of the given character.
+      Results in the unicode codepoint of the given character if it fits into a single codepoint.
```
vkleen, line 207: I'm not entirely sure of the correctness of this function and `from_codepoint`. But I do think that this is the interface we should be aiming for and if it doesn't work quite right at the moment that should be a bug in the evaluator.

vkleen:
We (me and @yannham) cleaned up this branch yesterday and believe it's now in a mergable state, in principle.

yannham (approved):
I'm partial, but I approve :stuck_out_tongue: 

I wonder about the `Char`/`Character` and `chars`/`characters`. As for `bool`, I feel like the reduced form is so common that it might actually be ok. Anyway, it's not tremendously important, either way.

stdlib/array.ncl
```diff
@@ -21,30 +21,50 @@
        else
          %blame% (%tag% "not a array" label),
 
-    head : forall a. Array a -> a
+    first : forall a. Array a -> a
       | doc m%"
-        Results in the head (first element) of the given array.
+        Results in the first element of the given array.
```
matthew-healy, line 26: Nitpick: I think "Returns..." is a little more direct than "Results in..." though both are fine.

matthew-healy (approved):
Looks great!

stdlib/array.ncl
matthew-healy: Yep, totally fair.

PR #1093, opened 2023-02-01
Fixes typos

deining:
This PR fixes  typos I spotted in various files. It also converts two links from `http` to `https` protocol.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -542,7 +542,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                             let mut shared_env = Environment::new();
                             let f_as_var = f.body.closurize(&mut self.cache, &mut env, f.env);
 
-                            // Array elements are closurized to preserve lazyness of data structures. It
+                            // Array elements are closurized to preserve laziness of data structures. It
                             // maintains the invariant that any data structure only contain thunks (that is,
                             // currently, variables).
                             let ts = ts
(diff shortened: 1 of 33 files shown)


src/eval/operation.rs
```diff
@@ -594,7 +594,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                         let mut shared_env = Environment::new();
                         let f_as_var = f.body.closurize(&mut self.cache, &mut env, f.env);
 
-                        // Array elements are closurized to preserve lazyness of data structures. It
+                        // Array elements are closurized to preserve laziness of data structures. It
```
yannham, line 597: Ah, I've always thought lazyness was a valid US spelling. But it seems it's actually not the case.

src/types.rs
```diff
@@ -1050,7 +1050,7 @@ mod test {
     /// Take a string representation of a type, parse it, and assert that formatting it gives the
     /// same string as the original argument.
     ///
-    /// Note that their are infintely many string representations of the same type since, for
+    /// Note that their are infinitely many string representations of the same type since, for
```
yannham, line 1053:
```suggestion
    /// Note that there are infinitely many string representations of the same type since, for
```

yannham (approved):
Thanks for the clean up! As a side note, If you've read the in-code documentation yourself, be ware that while individual items (attached to methods, types, etc.) are mostly up-to-date, the top-level documentation of some modules is a bit outdated.

src/eval/operation.rs
deining: So no change on my side here ...

PR #1096, opened 2023-02-02
Remove redundant check for "indirect" nested record usage.

ebresafegaga:
This fixes an issue where the LSP crashes if you try to reference a field which has a value of `Import(..)`, and the import refers to a record.  

i.e, if you have this: 
```nickel
let foo = {
  S3Backend = import "./aws-s3-backend.ncl",
}
foo.S3Backend
```
And `import "./aws-s3-backend.ncl"`, has a record term as its value.  What causes the crash exactly is  resolving the  `foo.S3Backend` field reference.

See [here](https://github.com/tweag/tf-ncl-examples/blob/lsp-crash/aws/main.tf.ncl) for the example that caused this crash.

### DIFF
--- a/lsp/nls/src/linearization/building.rs
+++ b/lsp/nls/src/linearization/building.rs
@@ -233,36 +233,6 @@ impl<'b> Building<'b> {
                     _ => None,
                 });
 
-            let referenced_declaration =
-                referenced_declaration.and_then(|(id, referenced)| match &referenced {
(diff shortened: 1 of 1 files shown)


yannham (approved):
I'm not sure to fully understand to implication of removing that. But if you're confident that this doesn't break recursive field resolving, and doesn't cause a regression on goto definition/goto references for recursive fields, I'm always keen to remove code.

PR #1099, opened 2023-02-03
Correctly type identifiers introduced in destructuring patterns

matthew-healy:
Previously identifiers introduced via record destructuring were always typed as `Dyn`. This PR improves the situation by correctly inferring the type of these identifiers. The expressions in `tests/integration/destructuring/pass/preserves_types.ncl` should give reasonable guidance as to how this behaves.

One notable omission is that type annotations within the pattern are not enforced during typechecking (though they _are_ converted to contracts & applied during evaluation). I decided to hold off on implementing that until after the RFC005 PR gets merged to avoid dealing with whatever merge conflicts might arise.

I also took the opportunity to remove the `Destruct::Array` variant, which was never used in the codebase.


ebresafegaga (approved):
Looks good

src/typecheck/mod.rs
```diff
@@ -1000,23 +1003,192 @@ fn walk_rrows<L: Linearizer>(
...
+    /// This allows us to be resilient to ordering differences between the
+    /// pattern and its type. As well as keeping track of which identifiers
+    /// have already been "used" in the pattern, to ensure that we can
+    /// correctly construct the type of a `..rest` match, if it exists.
+    enum RecordTypes {
+        Rows {
+            known_types: HashMap<Ident, UnifType>,
+            tail: UnifRecordRows,
+        },
+        Dyn,
```
yannham, line 1117: I'm not sure to see why we need a `Dyn`, and in which case does it arise. Shouldn't we always infer a record type from a pattern? If yes, this state shouldn't be representable, and `RecordType` should probably implement `try_from`, or `From<RecordRows>`.

tests/integration/destructuring/pass/preserves_types.ncl
```diff
@@ -0,0 +1,42 @@
...
+  # Note: we need to annotate `a` on the right-hand side of the binding
+  #       because we don't currently have a subtyping rule like:
+  #         `{ f_1: T, ..., f_N: T } <: {_ : T}`
+  #       If you're reading this and we _do_ have such a subtyping rule
+  #       (e.g. after RFC003 has been implemented) then it should be
+  #       safe to remove that type annotation from this test case.
+  "destructuring with explicit types" =
+    let { a : { _ : Num } } = { a: { _ : Num } = { b = 1 } } in
+    a : { _ : Num },
+
```
yannham, line 36: This files could have additional tests for:

- type inference (make patterns and terms without any annotation)
- type inference in the presence of functions (pattern match on function arguments without type annotation), because function argument are the simplest way to have "unconstrained" values (typed with unification variables)

src/typecheck/mod.rs
```diff
@@ -1000,23 +1003,192 @@ fn walk_rrows<L: Linearizer>(
...
+
+                matches.iter().for_each(|m| match m {
+                    Match::Simple(id, ..) => {
+                        let ty = type_map.get_type(id);
+                        env.insert(*id, ty);
+                    }
+                    Match::Assign(id, _, (bind_id, pat)) => {
+                        let ty = type_map.get_type(id);
+                        let id = bind_id.as_ref().unwrap_or(id);
+                        env.insert(*id, ty.clone());
```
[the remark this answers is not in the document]
matthew-healy: Oh, this is a really good point. I was a bit confused as to how `bind_id` could ever be `None`, but this makes sense now. I'll take another look at this.

src/typecheck/mod.rs
matthew-healy: We need `Dyn` here in order to introduce the pattern variables when `walk`ing. In that case, we pass `GenericUnifType::Concrete(Dyn)` as the pattern type, and so we create a `RecordType::Dyn` which then always returns `Dyn` from `get_type` and `{ _ : Dyn }` from `rest`.

I'm totally open to better ways of structuring this if you have suggestions.

tests/integration/destructuring/pass/preserves_types.ncl
matthew-healy: I've added some extra tests - I _think_ they cover what you've asked for, but if there's anything still missing please let me know!

src/typecheck/mod.rs
```diff
@@ -1000,23 +1003,192 @@ fn walk_rrows<L: Linearizer>(
...
+                    Match::Assign(id, _, (_, r_pat @ Destruct::Record { .. })) => {
+                        let ty = build_pattern_type(state, r_pat);
+                        RecordRowF {
+                            id: *id,
+                            types: Box::new(ty),
+                        }
+                    }
+                });
+
+                let rrows = rows.fold(tail, |tail, row| {
```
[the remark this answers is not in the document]
matthew-healy: Yes! I was looking for something like this, but ended up just `fold`ing multiple times...

src/typecheck/mod.rs
yannham: I think you want the elaborate the pattern type even when walking. For example, during walking:

```
nickel> let r = {x = 1} in
let x | Num = r.x in
(x + 2 : Num)
3

nickel>
```

`x` is still affected the type `Num`, because it has a contract annotation. However, if we use destructuring:

```
nickel> let r = {x = 1} in
let {x | Num} = r in
(x + 2 : Num)
error: incompatible types
  ┌─ repl-input-1:3:2
  │
3 │ (x + 2 : Num)
  │  ^ this expression
  │
  = The type of the expression was expected to be `Num`
```

However, the second example is desugared to the same thing as the first one, and contains as much information. The type is not really inference per se, but mostly building a map from identifiers to types based on annotations alone, which you do even during the walk phase.

yannham (commented):
`Destruct::Empty` can never be produced by a full pattern (that is, the parser currently always produces something of the form `Destruct::Record`). Thus, `build_pattern_type` can in fact only return rows right now.

It's not utterly important, but the fact that `From<UnifType> for RecordTypes` has the case 
`_ => RecordTypes::Dyn` bothers me, as a pattern type is in fact much more constrained, and it wouldn't make sense for this function to be called on say `Num`.

One possibility would be to make the `Destruct` type stricter (only allow `Record` at the top-leve), make `build_pattern_type` returns a map (that is `known_types` directly), typing everything more strictly. `inject_pattern_variables` would then be passed this map directly. Then you convert this map to a `UnifType` if you are in typecheck mode and need to unify stuff. As per my comment somewhere, I think you always want to use the record map, even in walking mode.

The suggestion shouldn't be implemented in this PR. Let's just fix the issue when the walking phase, merge, and iterate later :slightly_smiling_face:

src/typecheck/mod.rs
matthew-healy: Ah yeah, this is a great point. Thank you!

yannham (approved):
LGTM so far! An embarrassing gap filled :slightly_smiling_face:

PR #1102, opened 2023-02-03
Fix completion on terms annotated with more than one contract

ebresafegaga:
Depends on #990 

Currently, if we have the Nickel program below and we want to perform "context completion" inside the record, we only get correct completion identifiers from one of the contract.  
```nickel
let Client = import "client.ncl" in 
let Server = import "server.ncl" in 
{

   [.]

} | Client | Server

```

With this change, we get correct completion identifiers for all the contracts present.

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -131,7 +131,9 @@ fn find_fields_from_term_kind(
     let Some(item) = linearization.get_item(id, lin_cache) else {
         return Vec::new()
     };
-    match item.kind {
+    let mut path_clone = path.clone();
+    let contract_result = find_fields_from_contract(item.id, &mut path_clone, info);
+    let result = match item.kind {
         TermKind::Record(ref fields) => {
             if path.is_empty() {
                 fields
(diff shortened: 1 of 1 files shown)


lsp/nls/src/requests/completion.rs
```diff
@@ -217,17 +217,22 @@ fn find_fields_from_contracts(
...
+            let mut path_copy = path.clone();
+            match &contract.types {
+                Types(TypeF::Record(row)) => find_fields_from_type(row, &mut path_copy, info),
+                Types(TypeF::Dict(ty)) => match (&**ty, path_copy.pop()) {
+                    (Types(TypeF::Flat(term)), Some(_)) => {
+                        find_fields_from_term(term, &mut path_copy, info)
+                    }
+                    (Types(TypeF::Record(rrows)), Some(_)) => {
+                        find_fields_from_type(rrows, &mut path_copy, info)
+                    }
```
yannham, line 230: It seems this code is more or less similar to the outer code, but just nested (check if something is flat or a record). Should we put that in a separate subfunction `find_field_from_type` (and rename `find_field_from_type` to `find_field_from_rrows`? And should we just call it recursively, so that a nested dictionary type `{_: {_: {_: {bar}}}}` gives completion for `baz.rlog.blirg[.]` ?

ebresafegaga: Yesss. Thank you!

src/error.rs
```diff
@@ -849,6 +849,7 @@ impl ToDiagnostic<FileId> for EvalError {
                 diags
             }
             EvalError::TypeError(expd, msg, orig_pos_opt, t) => {
+                println!("type error term: {} / {:#?}", t, t);
```
yannham, line 852: Probably just a rebase artifact.

```suggestion
```

yannham (approved):

PR #1104, opened 2023-02-03
Compile and typecheck benchmarks in main ci workflow

### LINKED ISSUE #1042: Test that benchmarks parse correctly
Benchmarks aren't part of the default CI, and for good reasons (they take a lot of time to complete, and aren't relevant for all PRs).

However, it has happened several times that we change the syntax and the benchmark aren't updated. We only discover much later, with possibility several incompatibilities stacked up, that they are broken.

As a first step toward avoiding this in the future, we can check in the CI that the benchmarks correctly parse and typecheck.

asymmetric:
@yannham is this what you had in mind, or are there any additional steps to run?

Closes #1042.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -56,4 +56,6 @@ jobs:
       run: |
         nix develop --accept-flake-config
         cargo bench --no-run
-        find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nix run --accept-flake-config . -- -f file typecheck
+
+        nix shell --accept-flake-config
+        find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nickel -f file typecheck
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -51,3 +51,9 @@ jobs:
     - name: Run all checks
       run: |
         nix flake check --print-build-logs
+
+    - name: Compile and typecheck benchmarks
+      run: |
+        nix develop --accept-flake-config
+        cargo bench --no-run
+        find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nix run --accept-flake-config . -- -f file typecheck
```
asymmetric, line 59: A neater way to do this would be adding a `devShells.ci` to the flake with `nickel` in its `buildInputs`, so that we could then just pass `nickel` to `xargs`.

yannham (approved):
LGTM. Probably better to do that in-flake long term, but for now it's a strict improvement. Thanks!

PR #1105, opened 2023-02-03
CI: configure cachix via provided nix flag instead of git action

asymmetric:
Since this action never pushes, there's no need for the secret, and therefore the action.


yannham:
Unless I'm mistaken, this workflow does push via the Cachix Post-Setup step. Otherwise cachix would never be filled, but it seems to be?

asymmetric:
You're [right](https://github.com/tweag/nickel/actions/runs/4116442691/jobs/7106500527#step:10:19), sorry for the spam.

PR #1106, opened 2023-02-03
flake: filter .github and .vscode dirs

asymmetric:
Prevents them from being copied to the store, and triggering rebuilds.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -166,7 +166,15 @@
           snapFilter = mkFilter ".*snap$";
         in
         pkgs.lib.cleanSourceWith {
-          src = pkgs.lib.cleanSource ./.;
+          # Exclude paths that would trigger unnecessary rebuilds
+          src = pkgs.lib.cleanSourceWith {
+            src = pkgs.lib.cleanSource ./.;
+
+            filter = path: _type:
+              let basename = builtins.baseNameOf path; in
+
+              basename != ".github" && basename != ".vscode";
+          };
 
           # Combine our custom filters with the default one from Crane
           # See https://github.com/ipetkov/crane/blob/master/docs/API.md#libfiltercargosources


yannham (commented):
Hey again :) thanks for contributing! 

In fact those filters say what should be included in the source, additionally to the default filter of `crane` (including e.g. Rust files), not what is excluded. For example `.ncl` files are required to build `nickel` and for flake checks. If this is the case, I think we can close this PR.

asymmetric:
You're indeed right. The fact temains that `.github` and `.vscode` dirs currently end up in the store, causing what are probably unnecessary hash mismatches and therefore rebuilds.

```
❯ ll /nix/store/dznrgfwg8swfj53q5k48kyn2qpxjxqfc-source -a
Permissions Size User Date Modified Name
dr-xr-xr-x     - root  1 Jan  1970  .github
dr-xr-xr-x     - root  1 Jan  1970  .vscode
dr-xr-xr-x     - root  1 Jan  1970  benches
.r--r--r--   156 root  1 Jan  1970  build.rs
.r--r--r--   72k root  1 Jan  1970  Cargo.lock
.r--r--r--  2.6k root  1 Jan  1970  Cargo.toml
dr-xr-xr-x     - root  1 Jan  1970  doc
dr-xr-xr-x     - root  1 Jan  1970  examples
dr-xr-xr-x     - root  1 Jan  1970  lsp
dr-xr-xr-x     - root  1 Jan  1970  nickel-wasm-repl
dr-xr-xr-x     - root  1 Jan  1970  notes
dr-xr-xr-x     - root  1 Jan  1970  pyckel
dr-xr-xr-x     - root  1 Jan  1970  rfcs
dr-xr-xr-x     - root  1 Jan  1970  spec
dr-xr-xr-x     - root  1 Jan  1970  src
dr-xr-xr-x     - root  1 Jan  1970  stdlib
dr-xr-xr-x     - root  1 Jan  1970  tests
dr-xr-xr-x     - root  1 Jan  1970  utilities
```

Are you interested in a fix? If so, I can adapt this PR.

yannham:
@asymmetric you're right. I tried to do something like that for selecting only subcombinations of crates (e.g. building `nickel` doesn't need `lsp`). It didn't work right away because of how cargo workspaces work, but it could probably be easily adapted to exclude directories that don't aren't needed at all: https://github.com/tweag/nickel/pull/1051

Note that I don't think `"^\\.vscode"`, IIRC, my `builtin.trace` showed that Nix passes the full Nix store path, which thus always starts with a prefix like `foobar-iwuhziUIHIZUdiq/.vscode/`.

asymmetric:
@yannham I went for a simpler solution, let me know what you think!

flake.nix
```diff
@@ -166,7 +166,15 @@
...
-          src = pkgs.lib.cleanSource ./.;
+          # Exclude paths that would trigger unnecessary rebuilds
+          src = pkgs.lib.cleanSourceWith {
+            src = pkgs.lib.cleanSource ./.;
+
+            filter = path: _type:
+              let basename = builtins.baseNameOf path; in
+
+              basename != ".github" && basename != ".vscode";
+          };
```
vkleen, line 177: Can this be merged into the `filter` function below somehow? It feels kind of weird to have two layers of `cleanSourceWith` for this.

asymmetric:
@vkleen thanks for taking a step back here, I think you're right, there's no spurious builds, and as you say those directories are empty.

This can be permanently closed.

PR #1107, opened 2023-02-04
ci: cancel previously running jobs

asymmetric:
I'm not sure if this is the desired behaviour on master, but it definitely seems like the right thing to do on feature branches.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -5,6 +5,9 @@ on:
       - master
   pull_request:
 
+concurrency:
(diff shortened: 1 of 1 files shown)


yannham:
Hey, thanks for contributing! Do you know if this have any implication on concurrency, like forcing the builds to be sequential while they could be parallel before?

yannham (approved):

asymmetric:
Note that there's only one job, the rest are steps. Jobs are parallel by default, steps sequential. :)

But yeah there shouldn't be any change here.

PR #1109, opened 2023-02-06
Make value bindings point to their actual value instead of their annotated contract. 

ebresafegaga:
Depends on #1102 

This pull request fixes a bug introduced in #990, where let declarations or record fields declarations which have values that are annotated with contracts, point to the contract (contained in the annotation) as their value, instead of the actual value.


lsp/nls/src/requests/completion.rs
```diff
@@ -152,20 +152,18 @@ fn find_fields_from_term_kind(
...
-                };
-                find_fields_from_term_kind(
-                    *new_id,
-                    path,
-                    &ComplCtx {
-                        linearization,
-                        lin_cache,
-                    },
-                )
+                let id = path.pop().and_then(|name| fields.get(&name));
```
yannham, line 155: I also like it better this way :stuck_out_tongue:

yannham (approved):
Good catch! LGTM

PR #1124, opened 2023-02-14
[LSP] Perform import resolution independently and provide diagnostics when a resolution fails

### LINKED ISSUE #1108: [LSP] Unresolved imports when one import resolution fails
**Describe the bug**
When we have more than one import term in a Nickel file, and one of those import fails to be resolved correctly, the other imports are automatically unresolved.

**Expected behavior**
The LSP *should* resolve all imports that have a correct Nickel file path, regardless of whether other imports resolve correctly.

ebresafegaga:
Currently, when we have multiple import terms in a file, and one of the imports fails to resolve without any errors, all the other imports automatically fail; this is not ideal for a good user experience. With this change, a single import resolution failure doesn't affect other import resolution, and the imports that fail are flagged out using the LSP's diagnostics feature.

Closes #1108

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -658,7 +658,10 @@ impl Cache {
     /// nothing if the imports of the entry have already been resolved. Require that the
     /// corresponding source has been parsed.
     /// If resolved imports contain imports themselves, resolve them recursively.
-    /// It returns a `Vec<FileId>`, which is a vector of all the imports that were transitively resolved.
+    /// It returns a two vectors which contain all the imports that were transitively resolved, and
+    /// all the errors it encountered while resolving imports in `file_id`, respectively. It only
+    ///  accumulates errors if the `Cache` is in error tolerant mode, otherwise it returns an `Err(..)`
+    /// containing  a `CacheError`
     #[allow(clippy::type_complexity)]
     pub fn resolve_imports(
         &mut self,
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -663,11 +663,13 @@ impl Cache {
         &mut self,
         file_id: FileId,
         resillient: bool,
-    ) -> Result<CacheOp<Vec<FileId>>, CacheError<ImportError>> {
+    ) -> Result<CacheOp<(Vec<FileId>, Vec<ImportError>)>, CacheError<ImportError>> {
         match self.entry_state(file_id) {
-            Some(state) if state >= EntryState::ImportsResolved => Ok(CacheOp::Cached(Vec::new())),
+            Some(state) if state >= EntryState::ImportsResolved => {
+                Ok(CacheOp::Cached((Vec::new(), Vec::new())))
+            }
             Some(state) if state >= EntryState::Parsed => {
-                let pending = if state < EntryState::ImportsResolving {
+                let (pending, errors) = if state < EntryState::ImportsResolving {
                     let CachedTerm {
                         term, parse_errs, ..
                     } = self.terms.get(&file_id).unwrap();
@@ -678,12 +680,11 @@ impl Cache {
                     // put it back when done, but to get a reference and clone it.
                     let term = term.clone();
                     let parse_errs = parse_errs.clone();
-                    let (term, pending) = if resillient {
-                        let (term, pending, _errors) =
-                            import_resolution::resolve_imports_resillient(term, self);
-                        (term, pending)
+                    let (term, pending, errors) = if resillient {
+                        import_resolution::resolve_imports_resillient(term, self)
                     } else {
-                        import_resolution::resolve_imports(term, self)?
+                        let (term, pending) = import_resolution::resolve_imports(term, self)?;
+                        (term, pending, Vec::new())
                     };
 
                     self.terms.insert(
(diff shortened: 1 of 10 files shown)


src/cache.rs
```diff
@@ -659,14 +659,18 @@ impl Cache {
     /// corresponding source has been parsed.
     /// If resolved imports contain imports themselves, resolve them recursively.
     /// It returns a `Vec<FileId>`, which is a vector of all the imports that were transitively resolved.
+    #[allow(clippy::type_complexity)]
     pub fn resolve_imports(
         &mut self,
         file_id: FileId,
-    ) -> Result<CacheOp<Vec<FileId>>, CacheError<ImportError>> {
+        resillient: bool,
```
yannham, line 666: Please, no :cry:  we've had a lot of bugs (and still have, in fact) because of those pesky error-tolerance booleans that were present in each and every function of the cache, and it's really hard to keep track on what should be error tolerant or not. Now, there is a data `error_tolerance` or something in the cache that dictates the behavior.  There is an `ErrorTolerance` type. You should probably use it instead and not take any additional argument

src/transform/import_resolution.rs
```diff
@@ -26,10 +27,50 @@ pub fn resolve_imports<R>(
     rt: RichTerm,
     resolver: &mut R,
 ) -> Result<(RichTerm, Vec<FileId>), ImportError>
+where
+    R: ImportResolver,
+{
+    let (term, resolved, errors) = resolve_imports_resillient(rt, resolver);
```
yannham, line 33:
```suggestion
    let (term, resolved, errors) = resolve_imports_resilient(rt, resolver);
```

src/transform/import_resolution.rs
```diff
@@ -26,10 +27,50 @@ pub fn resolve_imports<R>(
...
+pub fn transform_one<R>(
+    rt: RichTerm,
+    resolver: &mut R,
+    parent: &Option<PathBuf>,
+) -> Result<RichTerm, ImportError>
+where
+    R: ImportResolver,
+{
+    let term = rt.as_ref();
+    match term {
```
yannham, line 51: Would that make sense to reuse `transform_one_resilient` here, as you did in `resolve_imports` using `resolve_imports_resilient`?

src/transform/import_resolution.rs
```diff
@@ -40,41 +81,52 @@ where
...
-            Ok(RichTerm::new(Term::ResolvedImport(file_id), rt.pos))
-        }
-        _ => Ok(rt),
+        Term::Import(path) => match resolver.resolve(path, parent.clone(), &rt.pos) {
+            Ok((_, file_id)) => (RichTerm::new(Term::ResolvedImport(file_id), rt.pos), None),
+            Err(err) => (rt, Some(err)),
+        },
+        _ => (rt, None),
     }
 }
```
yannham, line 132: Suggestion: we have several function with similar names, the XXX, and the XXX_resilient. One way to organize this a bit more would be to have submodules, say `strict` and `tolerant` (just because it's the terminology we are already using in the cache; for what it's worth, resilient is totally fine too), each with their own `resolve_imports` and `transform_one` function.

src/cache.rs
ebresafegaga: Right. This is much better. Thank you! 🙂

src/transform/import_resolution.rs
ebresafegaga: Ah yes 👍

src/cache.rs
```diff
@@ -682,10 +682,11 @@ impl Cache {
                     let parse_errs = parse_errs.clone();
```
yannham, line 682: It think the documentation of `resolve_imports` deserve to be updated, as you changed the return value, it's not in sync anymore.

src/transform/import_resolution.rs
```diff
@@ -17,113 +16,133 @@ struct ImportsResolutionState<'a, R> {
...
+    /// * A `Vec<ImportError>`, telling the imports that couldn't be resolved.
+    pub fn resolve_imports<R>(
+        rt: RichTerm,
+        resolver: &mut R,
+    ) -> (RichTerm, Vec<FileId>, Vec<ImportError>)
+    where
+        R: ImportResolver,
+    {
+        let mut stack = Vec::new();
+        let mut unresolved = Vec::new();
```
yannham, line 88: Small nitpick: `unresolved` makes it sound this is a vector of `FileId` for unresolved imports. I think it's rather a vector of `parse_errors`?

src/transform/import_resolution.rs
```diff
@@ -17,113 +16,133 @@ struct ImportsResolutionState<'a, R> {
...
-    });
+    /// Performs imports resolution for each import term independent from others.
+    /// It returns back a triple, containing:
+    /// * A `RichTerm` which may still contain unresolved imports.
+    /// * A `Vec<FileId>`, telling the imports which were resolved.
+    /// * A `Vec<ImportError>`, telling the imports that couldn't be resolved.
+    pub fn resolve_imports<R>(
+        rt: RichTerm,
+        resolver: &mut R,
+    ) -> (RichTerm, Vec<FileId>, Vec<ImportError>)
```
yannham, line 83: Nitpick: could use a proper struct for the return type.

yannham (approved):

yannham:
The benchmarks are broken, because they referred to `import_resolution::some_function_that_doesnt_exist_anymore`, I reckon.

PR #1125, opened 2023-02-15
Fix new clippy warnings

yannham:
Fix new clippy warnings coming with 1.67. Originally done to fix #1118, but the latter is still failing because of an unrelated Nix segfault in the CI. In the meantime, this PR backports those fixes to `master`.


yannham:
Those fixes have been done automatically with `cargo clippy --fix`, and consist uniquely in rewriting `println("foo{}bar", val)` to `println!("foo{val}bar")`. Since I already reviewed myself, I'm merging without review.

PR #1126, opened 2023-02-15
Simplify typed destructuring implementation

matthew-healy:
This PR simplifies the implementation of typed destructuring in a few ways:

- Previously we always parsed functions as `FunPattern`s, and now we only parse functions as `FunPattern`s if they actually contain patterns.
- It removes the `RecordTypes::Dyn` variant by making the api of the `typechecking::destructuring` module more explicitly focused on typing destructured records.
- It removes `Destruct::Empty` in favour of using `Option<Destruct>` anywhere we're not certain we have a destructuring pattern.
- It renames `Destruct::Record` to `DestructuredRecord`
- It renames the `destruct` module to `destructuring`

I'd recommend reviewing commit-by-commit. The first three commits contain detailed messages which provide extra context, and the last two are just renaming things so probably don't require an in-depth look.


src/destructuring.rs
```diff
@@ -35,22 +35,16 @@ pub enum LastMatch {
...
-    Record {
-        matches: Vec<Match>,
-        open: bool,
-        rest: Option<Ident>,
-        span: RawSpan,
-    },
-    /// An empty destructuring. In this case, the pattern is a classical `let var = something in
-    /// body` form.
-    Empty,
+pub struct DestructuredRecord {
```
yannham, line 40: Nitpick (naming): `RecordPattern`. Destructuring is the feature but I think pattern is more commonly understood for the thing on the right hand side of the equal sign. That being said, it's a suggestion that you are free to ignore :slightly_smiling_face:

src/parser/utils.rs
```diff
@@ -583,6 +583,21 @@ pub fn mk_let(
...
 
+/// Generate a `Fun` or a `FunPattern` (depending on `pat` being empty or not) from the
+/// parsing of a function definition. This function panics if the definition somehow
+/// has neither an `Ident` nor a non-`Empty` `Destruct` pattern.
+pub fn mk_fun(id: Option<Ident>, pat: Option<DestructuredRecord>, body: RichTerm) -> Term {
+    match pat {
+        Some(d) => Term::FunPattern(id, d, body),
+        None => {
+            let Some(id) = id else {
+                unreachable!("functions always have either a non-Empty pattern or an ident")
```
yannham, line 594: Could we make an enum type to encode this information and avoid the `unreachable!` ? (and the `panic!` in `mk_let`). Something like

```
enum FunOrLetLeftHandSideButObviouslyPleaseFindAShorterName {
  Simple(id),
  Pattern(DestructureRecord),
}
```

src/transform/desugar_destructuring.rs
```diff
@@ -54,23 +55,26 @@ pub fn transform_one(rt: RichTerm) -> RichTerm {
...
+                    Term::Fun(
+                        x,
+                        RichTerm::new(Term::LetPattern(None, pat, Term::Var(x).into(), t_), t_pos /* TODO: should we use rt.pos? */),
+                    ),
+                    rt.pos,
+                )
+            },
+            t@Term::FunPattern(..) => panic!(
+                "A function can not have empty pattern without name in {t:?}"
+            ),
```
yannham, line 73: Isn't this dead code now?
```suggestion
```

src/typecheck/destructuring.rs
```diff
@@ -1,56 +1,47 @@
...
-
-            let rrows = rows.fold(tail, |tail, row| {
-                UnifRecordRows::Concrete(RecordRowsF::Extend {
-                    row,
-                    tail: Box::new(tail),
-                })
-            });
-
-            UnifType::Concrete(TypeF::Record(rrows))
+pub fn build_pattern_type(state: &mut State, pat: &DestructuredRecord) -> UnifRecordRows {
```
yannham, line 13: I think `build_pattern_type` has to act differently, depending on if we are in strict mode or not. Otherwise, your pattern will infer "too much" type in walking mode:

```
fun {x} => (x : Num)
```

Right now this typechecks, while it shouldn't (for example `fun x => (x : Num)` doesn't). `x` is introduced outside of any typed block, so it should be given type `Dyn`. Note that this also happens with intermediate pattern:

```
fun {x = foo @ {bar | Num}} => ...
```

So, `build_pattern_type` should probably does as `binding_type`: depending on being in strict mode or not, it should affect either `Dyn` or a unification variable to the leaf of the pattern.

src/parser/utils.rs
matthew-healy: Yeah, I had this thought myself but I was being lazy & figured I could come back to it later. You're probably right that it makes sense to do it now though.

src/transform/desugar_destructuring.rs
matthew-healy: Ah, yeah, good point - I guess the compiler doesn't notice because it's inside the `match_sharedterm!` macro. I would have thought dead code analysis happens after macro expansion, but I guess not... :thinking:

src/typecheck/destructuring.rs
matthew-healy: Ah yeah, good point. In the first version this worked because we used the `Dyn` variant. I'll have a look at how `binding_type` does this & try to copy the pattern used there.

src/transform/desugar_destructuring.rs
yannham: Yeah, I wondered why it wasn't flagged as well.

src/typecheck/destructuring.rs
yannham: It might be dead simple: take an additional `strict: bool` argument (well, ideally, `mode: TypecheckMode`, but this is not the only place where this change needs to be made, so we can do that in a pass in a later PR). If strict, when reaching the "leaf" of a pattern, produces a fresh uvar, as it is now. In non-strict mode, assign `Dyn` instead.

matthew-healy: Yeah, my only concern there is that passing a flag like that is a form of [control coupling](https://thoughtbot.com/blog/types-of-coupling#control-coupling), which is maybe fine, but equally we could just provide another function like `walk_pattern_type` & let the caller decide which is most appropriate.

yannham: I guess you still want to have only one implementation though, isn't it? Otherwise we'll have to duplicate code, which is bad  (I didn't know about control coupling! That being said, it seems to be about OOP, where "the object should decide what to do based on their own state", but at first sight it doesn't seem to fully translate to the functional style).

In the end I find `f("hello")/f_foo("hello", Mode::Foo)` and `f("hello")/f_bar("hello", Mode::Bar)` mostly isomorphic, but if your point is that we should wrap the one with the flag in two different functions for strict and non strict mode, I'm totally fine with that! That's probably cleaner, yes. Just have to find a good name, which is the easy part :roll_eyes: 

As a side note, I'm afraid `walk_pattern_type`  is a bit misleading, as we don't actually walk the pattern. It's probably more something like `build_pattern_in_walk_mode` and `build_pattern_in_check_mode` or in strict/nonstrict mode. The latter distinction might sound a bit odd at first, but I think it might be more precise, because with bidirectional typing, it would be fair to say we actually have 3 modes: walk, check, and infer. So "checking" is ambiguous. Maybe walk/typechecking is also fine, instead of nonstrict/strict.

matthew-healy: Yeah in my mind we'd maintain a single implementation, but it'd be internal to the `typecheck::destructuring` module, and we'd expose two functions, (with significantly better names than) `build_pattern_with_unif_var_leaves` and `build_pattern_with_dyn_leaves`.

Another possibility which is _maybe_ a little less coupled is to take a `() -> UnifType` argument which decides what to do at each pattern leaf. That way the `destructuring` module owns the algorithm, but the caller can control the behaviour as required. 

I'll play around with this & hopefully push something later today, though I'm a little busy with client work right now so can't make any promises.

yannham: As a datapoint, I think the `() -> UnifType` sounds a bit overkill. In practice it seems we only have to handle two behaviors, so in the end we will just end up having to make a closure inside typechecking. On the other hand just calling a specialized function makes the intent clearer IMHO.

For the private function though, both approach (boolean, enum, or closure) are reasonable, I guess.

matthew-healy:
I've pushed changed for all the comments which are resolved. I still need to look at having a better type for "either it's an ID or a pattern or both but not neither", so no need to re-review just yet.

yannham:
Hey @matthew-healy, this PR is in good shape, and fixes a big hole in typechecking code. There's already a new conflict, and I'm trying to minimize ongoing work: would that make sense to merge it as it is (after fixing the conflict), and improve on using a dedicated enum types later on?

matthew-healy:
@yannham Sure, if that's easiest! Sorry, I was hoping to get to fixing the last comment sooner, but I've been busy with client work. I'll fix the conflict now & merge when CI passes.

PR #1132, opened 2023-02-20
Non-exported record fields

vkleen:
This PR adds a `not_exported` field to record field metadata along with syntax to set it to true. The field is set to `false` by default and a metadata annotation in a record like
```nickel
{
  foo = 1,
  bar | not_exported = 2,
}
```
will set it to `true`. Upon merging it infects everything, i.e. when a field marked as `not_exported` is merged with any other field, the result will be `not_exported`.

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -432,6 +432,8 @@ fn merge_fields<'a, C: Cache, I: DoubleEndedIterator<Item = &'a Ident> + Clone>(
         // If one of the record requires this field, then it musn't be optional. The
         // resulting field is optional iff both are.
         opt: metadata1.opt && metadata2.opt,
+        // The resulting field will be suppressed from serialization if either of the fields to be merged is.
+        not_exported: metadata1.not_exported || metadata2.not_exported,
         priority,
     };
 
--- a/src/term/record.rs
+++ b/src/term/record.rs
@@ -362,22 +362,24 @@ impl RecordData {
             })
     }
 
-    /// Return an iterator over the fields' values, ignoring optional fields without
-    /// definition. Fields that aren't optional but yet don't have a definition are mapped to the
-    /// error `MissingFieldDefError`.
-    pub fn iter_without_opts(
+    /// Return an iterator over the fields' values, ignoring optional fields
+    /// without definition and fields marked as not_exported. Fields that
+    /// aren't optional but yet don't have a definition are mapped to the error
+    /// `MissingFieldDefError`.
+    pub fn iter_serializable(
         &self,
     ) -> impl Iterator<Item = Result<(&Ident, &RichTerm), MissingFieldDefError>> {
-        self.fields
-            .iter()
-            .filter_map(|(id, field)| match field.value {
-                Some(ref v) => Some(Ok((id, v))),
+        self.fields.iter().filter_map(|(id, field)| {
+            debug_assert!(field.pending_contracts.is_empty());
+            match field.value {
+                Some(ref v) if !field.metadata.not_exported => Some(Ok((id, v))),
                 None if !field.metadata.opt => Some(Err(MissingFieldDefError {
                     id: *id,
                     metadata: field.metadata.clone(),
                 })),
-                None => None,
-            })
+                _ => None,
+            }
+        })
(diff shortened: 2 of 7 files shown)


src/eval/merge.rs
```diff
@@ -432,6 +432,8 @@ fn merge_fields<'a, C: Cache, I: DoubleEndedIterator<Item = &'a Ident> + Clone>(
         // If one of the record requires this field, then it musn't be optional. The
         // resulting field is optional iff both are.
         opt: metadata1.opt && metadata2.opt,
+        // The resulting field will be suppressed from serialization if either of the fields to be merged is.
+        not_exported: metadata1.not_exported || metadata2.not_exported,
```
vkleen, line 436: This seems to be the correct semantics, especially when considering a program like
```nickel
{
  a = 1, b = 2
} | {
  a | Num,
  b | not_exported
    | Num,
}
```

src/term/record.rs
```diff
@@ -359,22 +362,24 @@ impl RecordData {
...
+                Some(ref v) if !field.metadata.not_exported => Some(Ok((id, v))),
                 None if !field.metadata.opt => Some(Err(MissingFieldDefError {
                     id: *id,
                     metadata: field.metadata.clone(),
                 })),
-                None => None,
-            })
+                _ => None,
+            }
+        })
```
vkleen, line 382: This function is only used in `src/serialize.rs` to decide which fields are serialized. I renamed it to make this more explicit and I added a `debug_assert!` to ensure that no pending contracts are encountered. This seems to be an important invariant in the codebase: "On accessing record fields, pending contracts are applied". The `into_` counterpart `into_iter_without_opts` of this function is able to apply pending contracts since it takes ownership of the field values. But `iter_serializable` only works on immutable references, and is called only after fully evaluating a term. Therefore, just asserting that no pending contracts are left seemed to be the most reasonable course of action.

src/eval/merge.rs
yannham: Interesting, because it means it actually gives a way to "remove" fields via merging, at least as far as serialization is concerned. I don't know if it should be used that way, or if it is a problem at all, but just wanted to point it out. It's probably also morally better than plain removal of a field, because you can still query it, see it when not exporting, etc.

src/term/record.rs
```diff
@@ -359,22 +362,24 @@ impl RecordData {
...
 
-    /// Return an iterator over the fields' values, ignoring optional fields without
-    /// definition. Fields that aren't optional but yet don't have a definition are mapped to the
-    /// error `MissingFieldDefError`.
-    pub fn iter_without_opts(
+    /// Return an iterator over the fields' values, ignoring optional fields
+    /// without definition and fields marked as not_exported. Fields that
+    /// aren't optional but yet don't have a definition are mapped to the error
+    /// `MissingFieldDefError`.
+    pub fn iter_serializable(
```
yannham, line 369: Very nitpicky nitpick, but `iter_serializable` could make it sound like it only iter through serializable fields (avoiding the one that aren't). I don't have great suggestions unfortunately. It's probably `iter_to_serialize` but that doesn't sound great either. Feel free to ignore :upside_down_face:

yannham (approved):

src/term/record.rs
vkleen: Funnily enough, I had `iter_to_serialize` in a previous iteration, but it did indeed sound too bad for my tastes. I'm open to suggestions 😄

ebresafegaga: How about `iter_fields_with_value` ? 😅

yannham: > How about iter_fields_with_value ? sweat_smile

I fear it might be misleading, but it does iter on fields that don't have values (where value is `None`). What it does is that it excludes fields without value and with the `optional` attribute set, together with fields with the `not_exported` attribute.

vkleen: I'll leave it the way it is (the path of least resistance :laughing:). The next time someone touches that code, they can see if another name fits better.

bew: (late to the party) Naming idea: `iter_exported_fields`

PR #1135, opened 2023-02-21
[Fix] Simplify typechecking and import resolution phase, fixing  a bug

### LINKED ISSUE #1090: Internal error when processing nested import
Put `import "error2.ncl"` into `error.ncl` and `import "<nonexistent>"` into `error2.ncl`. Running `nickel export -f error.ncl` yields
```
error: internal error: Unresolved import (<nonexistent>)
  ┌─ /home/vkleen/work/tweag/nickel/tf-ncl-examples/oci/error2.ncl:1:1
  │
1 │ import "<nonexistent>"
  │ ^^^^^^^^^^^^^^^^^^^^^^ here
  │
  = This error should not happen. This is likely a bug in the Nickel interpreter. Please considerreporting it at https://github.com/tweag/nickel/issues with the above error message.
```
Also, the error message is missing a space.

Tested on Nickel master, b930544978e657dc2f214a9357bcbb06979095d7

yannham:
Closes #1090.

This PR started as a stab at #1090. Before understanding what went wrong, I found the code typechecking and imports to be a bit obscure: I started by reworking it a bit, making clearer the assumptions and special cases. Those two functions are special in that they have to be re-entrant: they might be recursively called by themselves.

After the rework, the initial bug just went away: it turns out we were not propagating error on transitive imports properly inside a `flat_map` call. This PR adds a regression test for #1090, and also strengthen the test of cyclic imports (with several cyclic imports), as this kind of cases would probably have broken the original implementation, more than a simple, singular cyclic import.


src/cache.rs
```diff
@@ -799,10 +802,15 @@ impl Cache {
         initial_ctxt: &typecheck::Context,
     ) -> Result<(RichTerm, Vec<FileId>), Error> {
         let (term, errs) = self.parse_nocache(file_id)?;
-        if errs.no_errors() {
+        if !errs.no_errors() {
```
yannham, line 805: Not related to this PR, but this conditions seemed wrong.

src/cache.rs
```diff
@@ -673,65 +680,61 @@ impl Cache {
...
-                                    Vec::new()
-                                }
-                            })
-                            .collect(),
-                        errors,
-                    )
+                    self.update_state(file_id, EntryState::ImportsResolved);
+                    (done, import_errors)
                 } else {
-                    let pending = self.imports.get(&file_id).cloned().unwrap_or_default();
```
yannham, line 726: This part didn't make a lot of sense to me. If we are in this branch, we are recursively resolving imports. Which means:

1. We shouldn't set the state to `ImportsResolved` yet at the end of the call
2. Resolving imports of entries inside `self.imports` doesn't make a lot of sense (those imports were resolved already, meaning they should have been treated already - or be on in pending queue. It seems like the code might have been copy pasted from e.g. the typechecking code, where it makes sense (the imports indicate transitive imports that are yet to be typechecked).

src/cache.rs
```diff
@@ -486,15 +486,17 @@ impl Cache {
...
+                            self.typecheck(f, initial_ctxt)?;
+                        }
                     }
-                }
 
-                self.update_state(file_id, EntryState::Typechecked);
+                    self.update_state(file_id, EntryState::Typechecked);
+                }
+                // The else case correponds to `EntryState::Typechecking`. There is nothing to do:
+                // cf (grep for) [transitory_entry_state]
```
vkleen, line 499: This might just be my ignorance of the typechecking algorithm, but how does Nickel typecheck a file before having typecheckt its imports?

vkleen:
Both with this PR and on master, an immediately cycling import yields a stack overflow. E.g. put `import "test.ncl"` in `test.ncl` and try `nickel -f test.ncl`. Maybe we want to fix this as part of this cleanup?

yannham:
> Maybe we want to fix this as part of this cleanup?

If the bug is already on master, I would say no. Let's keep PR small and focused. I'll take a stab in a subsequent PR.

PR #1138, opened 2023-02-23
Add not_exported to the list of KEYWORDS

vkleen:
I forgot to add `not_exported` to the list `KEYWORDS`, that I introduced for the pretty printer 🤦.


yannham (approved):
It shows that this is easy to overlook :upside_down_face: unfortunately I remember we don't have a better solution that is simple as well.

PR #1141, opened 2023-02-23
Fix dictionary contract semantics

### LINKED ISSUE #892: Recursive overriding for dictionary contracts
## Status Quo
Currently dictionary contracts behave differently from record contracts when recursively overriding fields:
```
{
    b = {
      foo = c.foo,
    },
    c = {
    },
} | {
  _: {
    foo | Str
        | optional
        | default = "",
  }
}
```
complains about `c.foo` missing, while
```
{
    b = {
      foo = c.foo,
    },
    c = {
    },
} | {
  b | {
    foo | Str
        | optional
        | default = "",
  },
  c | {
    foo | Str
        | optional
        | default = "",
  }
}
```
yields the expected result `{ b = { foo = "" }, c = { foo = "" } }` upon evaluation.
## Proposal
We should ensure that dictionary contracts behave like record contracts with only the top-level field names being unknown. Here, @yannham suggests that the dictionary contract `{ _: C }` upon being applied generate a "scaffold" of the same shape as the record it is being applied to. I.e. `{ foo = ..., bar = ... } | { _: C }` should be equivalent to `{ foo = ..., bar = ... } | { foo | C, bar | C, .. }`.

The overall goal should be to ensure that applying a dictionary contract `{ _: C }` to a record is the same as applying `C` to each of the fields in the record.

vkleen:
This PR changes the semantics of dictionary contracts, hopefully for the better. Previously, dictionary contracts were implemented entirely in Nickel. The interpreter rewrites a contract assertion like `{foo = 1} | {_: Str}` into an application of `$dyn_record Str l {foo = 1}` in `stdlib/internals.ncl`. Before this change, `$dyn_record` would simply map e.g. `fun t => %assume% Str l t` over all fields of the record it was applied to.

Now, `$dyn_record` instead simply applies the `%dictionary_assume%` primop after checking that was actually passed a record. This new primop inspects the record it is applied to and synthesizes a magic open record contract. This contract contains precisely the fields of the record under inspection and each field is assigned the contract originally passed in the dictionary, i.e. for a dictionary contract `{_: Str}` each field is assigned the contract `Str`.

This has several consequences for the semantics:
- It ensures that applying a dictionary contract doesn't remove the ability to recursively override fields:
  ```
  ({ foo = 1, bar = foo } | {_: Num}) & { foo = 2}
    OLD ==> { foo = 2, bar = 1 }
    NEW ==> { foo = 2, bar = 2 }
  ```
- It makes recursive references to fields whose only value is a default in the contract possible. This was the gist of #892.
- It makes dictionary contracts work precisely like abbreviations for record contracts, if the names of the field were known:
  ```
  ({ foo } | {_: Num}) & { foo = "a string" }
    OLD ==> { foo = "a string" }
    NEW ==> blame error, "a string" doesn't satisfy Num
  ```

Closes #892


src/term/mod.rs
```diff
@@ -1305,6 +1305,9 @@ pub enum BinaryOp {
     /// Lazily apply a contract to an Array.
     /// This simply inserts a contract into the array attributes.
     ArrayLazyAssume(),
+
+    /// TODO
```
matthew-healy, line 1313: Does it make sense to document this before merging?

matthew-healy (approved):
This is a great change. I think it looks good overall, though I do think ideally we should have at least _some_ documentation of the primop before merging.

src/term/mod.rs
vkleen: Good catch, that should definitely get a proper got-docmment.

src/eval/operation.rs
```diff
@@ -2513,6 +2513,69 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     ))
                 }
             }
+            BinaryOp::DictionaryAssume() => {
```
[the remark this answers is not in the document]
vkleen: That's a very good point. I like `contract_term` for `rt3` and `contract_env` for `env3`; might as well change this right now and not make the problem worse 😅

src/eval/operation.rs
```diff
@@ -453,6 +453,24 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        l.path.push(ty_path::Elem::Dict);
+                        Ok(Closure::atomic_closure(RichTerm::new(
+                            Term::Lbl(l),
+                            pos_op_inh,
+                        )))
+                    }
+                } else {
+                    Err(EvalError::TypeError(
+                        String::from("Label"),
+                        String::from("go_array"),
```
yannham, line 468:
```suggestion
                        String::from("go_dict"),
```

src/eval/operation.rs
```diff
@@ -2513,6 +2531,71 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        }
+                    ))
+                };
+
+                match_sharedterm! {t2,
+                    with {
+                        Term::Record(record_data) => {
+                            let contract = contract_term.closurize(&mut self.cache, &mut env2, contract_env);
+                            let record_contract = Term::Record(RecordData{
+                                fields: record_data.fields.iter().map(|(key, value)| (*key, Field {
```
yannham, line 2563: The formatting is probably a bit off here. It's pretty annoying that `matched_sharedterm!` makes `cargo fmt` ignore this part. I don't know if that can be fixed in any way; in the meantime, a laborious but functional way is to replace the `matched_sharedterm` with a normal pattern matching, format, and turn it back to the macro.

src/eval/operation.rs
```diff
@@ -2513,6 +2531,71 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                                    NAryOp::MergeContract(),
+                                    Term::Lbl(lbl),
+                                    Term::Record(record_data),
+                                    record_contract
+                                ),
+                                env: env2,
+                            };
+                            Ok(merge_term)
+                        }
+                    } else Err(EvalError::TypeError(
```
yannham, line 2588: I didn't even know you could omit the `{ ... }` around the else block :exploding_head:

src/eval/operation.rs
```diff
@@ -2513,6 +2531,71 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+
+                match_sharedterm! {t2,
+                    with {
+                        Term::Record(record_data) => {
+                            let contract = contract_term.closurize(&mut self.cache, &mut env2, contract_env);
+                            let record_contract = Term::Record(RecordData{
+                                fields: record_data.fields.iter().map(|(key, value)| (*key, Field {
+                                    value: None,
+                                    pending_contracts: vec![PendingContract::new(contract.clone(), lbl.clone())],
+                                    metadata: FieldMetadata {
```
yannham, line 2566: Ah, this is a bit tricky. We have to found a "neutral" element for the original field's metadata. I think it might deserve a comment, so that if we extend metadata (for example with custom merge functions or whatnot), we know what we have to do here.

yannham (approved):
LGTM. However, it could be useful to include our private conversations somewhere, saying that creating a mirror record contract might not be free when applying a dictionary contract on a huge record. We did that for simplicity, and to keep merging logic inside, well, merge, but that if this becomes a problem, we could directly map the contracts onto each field and revert them adequately.

src/eval/operation.rs
vkleen: 🤯 I don't think I did that conciously

src/eval/operation.rs
vkleen: In that case it might be better to not use `..Default::default()`. That way we'll be forced by the compiler to consider that question when we change `FieldMetadata` in the future.

ebresafegaga (approved):

src/eval/operation.rs
vkleen: Oh, I think that's a quirk in the `match_sharedterm!` macro. It expects the else clause to be an `expr`. So it can be either a single expression, like it is here, or a compound block with `{ ... }`. This doesn't work with if-else in rust.

yannham: Yep, I saw that this occurs in a `matched_sharedterm`, but I still thought that one needed braces. It just turned out this was an habit.

src/eval/operation.rs
```diff
@@ -2559,18 +2568,40 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                                                    lbl.clone(),
+                                                )],
+                                                // This is intentionally not written as
+                                                // `{ opt: value.metadata.opt, ..Default::default() }`
+                                                //
+                                                // When `FieldMetadata` is changed
+                                                // in the future, we will need to consider
+                                                // which values should be inherited from
+                                                // `value.metadata` and which values can be
+                                                // left at their defaults.
```
yannham, line 2591: I think it's still missing the important information: `metadata` must be choosen such that, when merged with the original fields' metadata, all but `annotation.contracts` must be left unchanged.

vkleen: In fact, the metadata should be left entirely unchanged upon merging. The contract is directly inserted into the `pending_contracts` field because the primop is getting a closure, not just a term.

yannham: Hmm. I wonder if we should still insert it, so that `nickel query` shows all contracts attached (otherwise could be strange that you get a contract failure but can't see the contract in the list given by `query`). We can use the type contained in the label, which should be untransformed. That being said, let's think about that separately, as this PR is already approved and reviewed.

PR #1142, opened 2023-02-23
[Fix] Direct import loop causing stack overflow

yannham:
Fix https://github.com/tweag/nickel/pull/1135#issuecomment-1440005721.

The issue was indeed during typechecking (the evaluation still loops, but as expected, and without overflowing). When peeking at the apparent type of the imports, we would unwrap direct import chains until we find a different term. In the case of a trivial import loop, we don't. This PR adds a hashset to remember seen imports, as well as a regression test.

### DIFF
--- a/src/typecheck/mod.rs
+++ b/src/typecheck/mod.rs
@@ -1746,7 +1746,7 @@ pub fn apparent_type(
     use codespan::FileId;
 
     // Check the apparent type while avoiding cycling through direct imports loops. Indeed,
-    // `apparent_type` tries to pick through imported terms. But doing so can lead to an infinite
+    // `apparent_type` tries to see through imported terms. But doing so can lead to an infinite
     // loop, for example with the trivial program which imports itself:
     //
     // ```nickel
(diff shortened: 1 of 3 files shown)


vkleen (approved):

src/typecheck/mod.rs
```diff
@@ -1743,34 +1743,57 @@ pub fn apparent_type(
...
-                let t = r
-                    .get(*f)
-                    .expect("Internal error: resolved import not found during typechecking.");
-                apparent_type(&t.term, env, Some(r))
-            } else {
-                ApparentType::Approximated(Types(TypeF::Dyn))
+    use codespan::FileId;
+
+    // Check the apparent type while avoiding cycling through direct imports loops. Indeed,
+    // `apparent_type` tries to pick through imported terms. But doing so can lead to an infinite
```
yannham, line 1749:
```suggestion
    // `apparent_type` tries to see through imported terms. But doing so can lead to an infinite
```

PR #1143, opened 2023-02-24
[Fix] Fix query command in the REPL

### LINKED ISSUE #1115: REPL query doesn't find metadata
In the REPL `:query` is currently broken, at least for the standard library:
```
nickel> :query string.uppercase
```
This is probably related to #990.

yannham:
Closes #1115.

After the implementation of RFC005, the structure of metadata queries have changed. Before, one could simply query an arbitrary value. Now that metadata are attached to record fields only, the user must provide a base record (or expression) as well as a path pointing to a field. This commit fix the previously broken metadata queries for the REPL by updating to this new interface.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -410,6 +410,12 @@ pub enum ReplError {
         cmd: repl::command::CommandType,
         msg_opt: Option<String>,
     },
+    InvalidArg {
+        cmd: repl::command::CommandType,
+        argument: String,
+        expected_format: String,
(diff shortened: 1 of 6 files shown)


yannham:
The WASM build is broken, I'm gonna fix that, but it can still be reviewed in the meantime.

vkleen (approved):

ebresafegaga (approved):

PR #1144, opened 2023-02-24
[Refactor] Only represent valid match pattern states

matthew-healy:
Previously we stored match patterns as
`(Option<Ident>, Option<Pattern>)`, but this
allowed for the invalid state `(None, None)`.

This commit replaces this with an enum which
only allows the three valid states to be
represented, which removes some `panic`s.

### DIFF
--- a/src/destructuring.rs
+++ b/src/destructuring.rs
@@ -12,6 +12,19 @@ use crate::{
     types::{TypeF, Types},
 };
 
+#[derive(Debug, PartialEq, Clone)]
+pub enum FieldPattern {
+    /// An assignment match like `{ ..., a = b, ... }`
+    Ident(Ident),
+    /// A nested record pattern like `{ ..., a = { b, c }, ... }`
+    RecordPattern(RecordPattern),
+    /// An aliased nested record pattern like `{ ..., a = b @ { c, d }, ... }`
+    AliasedRecordPattern {
+        alias: Ident,
+        pattern: RecordPattern,
+    },
+}
+
 /// A match field in a `Destruct` pattern. Every field can be annotated with a type, with contracts
 /// or with a default value.
 #[derive(Debug, PartialEq, Clone)]
@@ -19,7 +32,7 @@ pub enum Match {
     /// `{..., a=b, ...}` will bind the field `a` of the record to variable `b`. Here, `a` is the
     /// first field of this variant and `b` the optional one. The last field can actualy be a
     /// nested destruct pattern.
-    Assign(Ident, Field, (Option<Ident>, Option<RecordPattern>)),
+    Assign(Ident, Field, FieldPattern),
     /// Simple binding. the `Ident` is bind to a variable with the same name.
     Simple(Ident, Field),
 }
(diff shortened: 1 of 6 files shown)


src/destructuring.rs
```diff
@@ -12,14 +12,24 @@ use crate::{
     types::{TypeF, Types},
 };
 
+#[derive(Debug, PartialEq, Clone)]
+pub enum MatchAssignment {
```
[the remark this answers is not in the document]
vkleen: What about `FieldPattern`?

src/destructuring.rs
```diff
@@ -12,14 +12,24 @@ use crate::{
...
+    /// An assignment match like `{ ..., a = b, ... }`
+    Ident(Ident),
+    /// A nested record pattern like `{ ..., a = { b, c }, ... }`
+    RecordPattern(RecordPattern),
+    /// An identified nested record pattern like `{ ..., a = b @ { c, d }, ... }`
+    IdentPattern(Ident, RecordPattern),
```
yannham, line 22: IMHO, the name `IdentPattern` is slighty confusing/inconsistent with the previous variants (it looks like it's a pattern destructuring an ident). In the end it's still a record pattern, but with something additional. I don't know the general name of such a top-level capture on the top of my head, unfortunately. Still, for the record, the OCaml manual refers to them as "alias patterns". Maybe `AliasRecordPattern`? `AliasedRecordPattern`?

yannham (commented):
Implies a tad of code duplication, but for the better, IMHO, and reasonably :+1: getting rid of panics and unreachable, as well as reducing the possibility of illegal state.

PS: the PR is still draft, is it on purpose?

matthew-healy:
@yannham Apologies, meant to move it out of review once CI passed but I forgot before the weekend.

src/destructuring.rs
matthew-healy: I'd originally gone with `IdentifiedPattern` but felt it was a little too long. I like `AliasedRecordPattern` a lot, so let's go with that. :+1:

src/destructuring.rs
```diff
@@ -12,14 +12,24 @@ use crate::{
...
 };
 
+#[derive(Debug, PartialEq, Clone)]
+pub enum FieldPattern {
+    /// An assignment match like `{ ..., a = b, ... }`
+    Ident(Ident),
+    /// A nested record pattern like `{ ..., a = { b, c }, ... }`
+    RecordPattern(RecordPattern),
+    /// An aliased nested record pattern like `{ ..., a = b @ { c, d }, ... }`
+    AliasedRecordPattern(Ident, RecordPattern),
```
yannham, line 22: Nitpick: any reason not to use a struct here? `AliasedRecordPattern { alias: Ident, pattern: RecordPattern }`

src/parser/utils.rs
```diff
@@ -558,53 +558,39 @@ pub fn mk_label(types: Types, src_id: FileId, l: usize, r: usize) -> Label {
...
-                }
-            } else {
-                panic!("unexpected let-binding without pattern or identifier")
-            }
+    match assgn {
+        FieldPattern::Ident(id) => Ok(if rec {
+            mk_term::let_rec_in(id, t1, t2)
+        } else {
+            mk_term::let_in(id, t1, t2)
+        }),
```
yannham, line 576: Leave the decision to you, but IMHO, I find guarded patterns to be slightly more readable:

```rust
        FieldPattern::Ident(id) if rec => Ok(mk_term::let_rec_in(id, t1, t2)),
        FieldPattern::Ident(id) => Ok(mk_term::let_in(id, t1, t2)),
```

yannham (approved):

PR #1146, opened 2023-02-26
Fix printing of higher-order funcs with type vars

### LINKED ISSUE #943: Higher-order function contracts with type variables are printed incorrectly in blame errors
**Describe the bug**

Higher-order functions which take an argument whose type has a quantified variable are printed in a potentially confusing way when outputting blame errors. In particular, the function argument is not parenthesised. 

**To Reproduce**

Consider the following program:

```
let f | Num -> (forall a. a -> Str) -> Str = fun n g => g n in f "" (fun a => %to_str% a)
```

This raises a blame error as we try to pass a sealed term to `%to_str%` which is not valid. However the first line of the error is:

```
error: contract broken by the caller
   ┌─ :1:18
   │
 1 │ Num -> forall a. a -> Str -> Str
   │                  - expected type of the argument provided by the caller
   │
```

**Expected behavior**

It would be sufficient to just parenthesise the function type:

```
error: contract broken by the caller
   ┌─ :1:18
   │
 1 │ Num -> (forall a. a -> Str) -> Str
   │                  - expected type of the argument provided by the caller
   │
```

This is how function types without quantified variables are currently printed.

asymmetric:
Very new to this codebase so let me know if I've misunderstood something :)

Closes #943

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1025,7 +1025,7 @@ impl Display for Types {
             TypeF::Record(row) => write!(f, "{{{row}}}"),
             TypeF::Dict(ty) => write!(f, "{{_: {ty}}}"),
             TypeF::Arrow(dom, codom) => match dom.0 {
-                TypeF::Arrow(_, _) => write!(f, "({dom}) -> {codom}"),
+                TypeF::Arrow(_, _) | TypeF::Forall { .. } => write!(f, "({dom}) -> {codom}"),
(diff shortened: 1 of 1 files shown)


yannham (approved):
LGTM! At some point we'll probably use the pretty-printer (in the `pretty` module) for everything, including types, instead of duplicating the formatting code. In the meantime, this change is small, correct, and improves on the previous situation :+1:

Issue #1052, opened 2023-01-18
Improve interface of labels (contracts)


yannham:
**Is your feature request related to a problem? Please describe.**
Currently, the only way to propagate custom information when a contract blames a value is to set the `tag` field of the label, using `contract.tag` or `contract.blame_with`.

The first issue is that it's not clear what to do on multiple calls to `contract.tag`, which happens e.g. when you compose contracts together:

```
MyFunContract Domain Codomain = fun label value input =>
  # we don't check that value is a function for conciseness
  contract.apply Domain (contract.tag label "argument is wrong") input
  |> contract.apply Codomain (contract.tag label label "return value is wrong")
```

Here, `Domain` or `Codomain` may also call to `contract.tag` themselves. Currently, Nickel just erases the previous tag, which might lose information: in this case, the tag of `MyFunContract` would be erased.

The second issue, explaining maybe why does Nickel just keep the last tag, is that the tag is printed in the main label ("label" with respect to error reporting diagnostics this time, not the contracts' labels). This main label ought to stay small, hence we avoid stacking many tags and printing them all here. But we may want to give a bit of context for some contract errors.

**Describe the solution you'd like**
In the end, from the point of view of a user, a label is an interface to blame error reporting. Labels could be more reflective of the current error reporting capabilities of Nickel, such as offering a way to add different type of tags (such as additional notes, and/or additional diagnostics with the `note` severity for example). The label API could propose different way of tagging as well: one cumulative, which doesn't erase the previous labels, and one that does, if useful (e.g. when you're confident that the data of the failing subcontract are enough and that the previous context only adds noise).

While the label API shouldn't be tied to the current error reporting infrastructure too much (`codespan`), it can still expose generic abstraction that should be present in any error reporting infrastructure.

**Describe alternatives you've considered**
Even without augmenting the API, we can at least provide an erasing and a cumulative version of `tag`.

The other obvious solution is to do nothing, and keep in mind that the label API is limited and that tagging erases the previous tag.

PR #1150, opened 2023-02-28
[Refactor] ToDiagnostic -> IntoDiagnostics

yannham:
Initially done as part of undertaking #1052, then separated as a standalone PR.

Errors use the trait (previously) `ToDiagnostic`, which provides a method `to_diagnostic(&self, ..) -> Vec<Diagnostic>`, but:

- Although the method takes `self` by reference, it's always used on `self` in practice (that is, we never use an error again after converting it to a diagnostic as of today). On the other hand, `&self` forces a bunch of needless clones in the implementations of `to_dagnostic`, for example having to clone `String` error messages stored in the error variant.
- The method is called `to_diagnostic` but returns a vector of diagnostics.

This PR replaces `ToDiagnostic` and `to_diagnostic` with `IntoDiagnostics` and `into_diagnostics`, taking `self` instead of `&self`, saving a few clones and reflecting more accurately the signature of the function.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -302,15 +302,15 @@ impl From<Vec<ParseError>> for ParseErrors {
     }
 }
 
-impl ToDiagnostic<FileId> for ParseErrors {
-    fn to_diagnostic(
-        &self,
+impl IntoDiagnostics<FileId> for ParseErrors {
+    fn into_diagnostics(
+        self,
         files: &mut Files<String>,
         contract_id: Option<FileId>,
     ) -> Vec<Diagnostic<FileId>> {
         self.errors
-            .iter()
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -782,16 +782,16 @@ impl ToDiagnostic<FileId> for EvalError {
                 let mut msg = String::new();
 
                 // Writing in a string should not raise an error, hence the fearless `unwrap()`
-                write!(&mut msg, "{}", blame_error::title(label)).unwrap();
+                write!(&mut msg, "{}", blame_error::title(&label)).unwrap();
```
vkleen, line 785: While we're here, we could make that comment first class and use `expect`.

vkleen (approved):

ebresafegaga (approved):

PR #1152, opened 2023-03-01
Enrich label custom reporting data

yannham:
First step of #1052. Enrich the label interface for custom error reporting, with the additional `notes` field. Put this field, together with the previous `tag` (now `message`) to a dedicated data structure, `ContractDiagnostic`. Finally, `Label` now stores a stack of such diagnostic, to help fix one issue of #1052 which is that subcontract might erase the message previously set by parent contract, although it could be relevant.

This PR doesn't add any primop to access this new interface yet: it mostly update the error reporting code to handle it. Adding primops is postponed to a subsequent PR.

I'm not totally sure yet what should go in the `ContractDiagnostic`s (should we also include `type_path` and `arg_idx`, to store the type path and argument to parent contracts as well, and enrich the corresponding "parent" diagnostics) ?. Should we add more fields than `notes` to the `label` interface?.

However, those questions need not be answered right away. We can improve later in a backward compatible manner: for now, the plan is to add at least `notes` to make it accessible to the user (in a subsequent PR), with a backward-compatible API. We can always:

- add more fields later (hints/suggestions, for example, or the label of the contract argument)
- move things form `Label` to `ContractDiagnostic`, allowing to store a stack of value instead of just the latest one, to improve the reporting of blame errors that have several contract diagnostics. This is invisible to user code, and thus can't break backward compatibility.

### DIFF
--- a/lsp/nls/src/linearization/completed.rs
+++ b/lsp/nls/src/linearization/completed.rs
@@ -168,14 +168,7 @@ impl Completed {
             if let Some(doc) = doc {
                 extra.push(doc.to_owned());
             }
-            if let Some(custom_contract_msg) = annotation
-                .types
-                .as_ref()
-                .and_then(|ty| ty.label.current_diagnostic())
-                .and_then(|diag| diag.message.as_ref())
-            {
-                extra.push(custom_contract_msg.clone());
-            }
+
             if let Some(contracts) = annotation.contracts_to_string() {
                 extra.push(contracts);
             }
--- a/lsp/nls/src/linearization/completed.rs
+++ b/lsp/nls/src/linearization/completed.rs
@@ -168,8 +168,13 @@ impl Completed {
             if let Some(doc) = doc {
                 extra.push(doc.to_owned());
             }
-            if let Some(types) = &annotation.types {
-                extra.push(types.label.tag.to_string());
+            if let Some(custom_contract_msg) = annotation
+                .types
+                .as_ref()
+                .and_then(|ty| ty.label.current_diagnostic())
+                .and_then(|diag| diag.message.as_ref())
+            {
+                extra.push(custom_contract_msg.clone());
             }
             if let Some(contracts) = annotation.contracts_to_string() {
                 extra.push(contracts);
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,8 +2,6 @@
 //!
 //! Define error types for different phases of the execution, together with functions to generate a
 //! [codespan](https://crates.io/crates/codespan-reporting) diagnostic from them.
-use std::fmt::Write;
-
 use codespan::{FileId, Files};
 use codespan_reporting::diagnostic::{Diagnostic, Label, LabelStyle};
 use lalrpop_util::ErrorRecovery;
@@ -27,8 +25,6 @@ use crate::{
     term::{record::FieldMetadata, RichTerm},
     types::{TypeF, Types},
 };
-
-use self::blame_error::ExtendWithCallStack;
 
 /// A general error occurring during either parsing or evaluation.
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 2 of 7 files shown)


src/error.rs
```diff
@@ -1337,13 +1284,122 @@ is None but last_arrow_elem is Some"),
...
+
+        // diagnostics.extend(label.diagnostics.into_iter().map(Diagnostic::from));
+        // // Contract diagnostics are pushed in-order: usually, the last one is the most
+        // // precise/relevant one. Thus, we want to show them in reverse order.
+        // diagnostics.reverse();
+
+        // We are going to work on the first diagnostic, so we ensure there is always one.
+        // if diagnostics.is_empty() {
+        //     diagnostics.push(Diagnostic::error());
+        // }
```
vkleen, line 1332: Is this left-over code from a previous iteration?

src/error.rs
```diff
@@ -1337,13 +1284,122 @@ is None but last_arrow_elem is Some"),
...
+                Diagnostic::error()
+                    .with_message(msg)
+                    .with_labels(labels)
+                    .with_notes(contract_notes),
+            );
+
+            if !notes_higher_order.is_empty() {
+                diagnostics.push(
+                    Diagnostic::note()
+                        .with_message("this is a function fonctract violiation")
```
vkleen, line 1364:
```suggestion
                        .with_message("this is a function contract violation")
```

src/error.rs
```diff
@@ -1337,13 +1284,122 @@ is None but last_arrow_elem is Some"),
...
+                    .with_message(msg)
+                    .with_labels(labels)
+                    .with_notes(notes_higher_order),
+            );
+        }
+
+        diagnostics.push(contract_bind_loc(&label));
+
+        for ctr_diag in contract_diagnostics {
+            let mut msg = String::from("from a parent contract violiation");
```
vkleen, line 1380:
```suggestion
            let mut msg = String::from("from a parent contract violation");
```

src/eval/merge.rs
```diff
@@ -246,7 +246,15 @@ pub fn merge<C: Cache>(
                     let fields: Vec<String> =
                         left.keys().map(|field| format!("`{field}`")).collect();
                     let plural = if fields.len() == 1 { "" } else { "s" };
-                    lbl.tag = format!("extra field{} {}", plural, fields.join(","));
+                    let fields_list = fields.join(",");
+
+                    lbl.set_diagnostic_message(format!("extra field{plural} {fields_list}"));
+                    lbl.set_diagnostic_notes(vec![
+                        format!("Have you mispelled a field?"),
```
vkleen, line 253:
```suggestion
                        format!("Have you misspelled a field?"),
```

src/label.rs
```diff
@@ -362,43 +366,127 @@ pub struct Label {
     pub path: ty_path::Path,
 }
 
+/// Custom reporting diagnostic that can be set by user-code through the `label` API. Used to
+/// customized contract error messages, and provide more context than "a contract has failed".
```
vkleen, line 370:
```suggestion
/// customize contract error messages, and provide more context than "a contract has failed".
```

lsp/nls/src/linearization/completed.rs
```diff
@@ -168,8 +168,13 @@ impl Completed {
...
             }
-            if let Some(types) = &annotation.types {
-                extra.push(types.label.tag.to_string());
+            if let Some(custom_contract_msg) = annotation
+                .types
+                .as_ref()
+                .and_then(|ty| ty.label.current_diagnostic())
+                .and_then(|diag| diag.message.as_ref())
+            {
+                extra.push(custom_contract_msg.clone());
```
ebresafegaga, line 177: While we're here, is it really useful to display this in the LSP? 🤔

ebresafegaga (approved):
LGTM

lsp/nls/src/linearization/completed.rs
yannham: I don't know? But at first sight, there is absolutely no reason (for now) that a label can have any custom diagnostic, as nothing is ever evaluated in the LSP. So I imagine we might get rid of that altogether

yannham:
Oops, [d77577b](https://github.com/tweag/nickel/pull/1152/commits/d77577b26a3b7f8f5468e408a6fb886ffccdcdba) was pushed on the wrong branch. Will drop it and force push.

PR #1165, opened 2023-03-08
Use a custom debug instance for Closure

vkleen:
This instance hides the `env` field in `Closure` when printing. Usually the environment is too large to be helpful on a quick inspection and in fact causes stack overflows because of its deeply recursive structure in some cases.


yannham (approved):
I imagine one could want to see the environment as well, but I agree that there's no sane way to print a common environment on a terminal, and the previous situation was worse :+1:

Issue #1161, opened 2023-03-07
#1141 breaks record.map with recursive dependencies


vkleen:
The new dictionary contract semantics from #1141 seem to break `record.map` on records with recursive dependencies. Specifically, #1141 is the first commit to produce the following:
```
nickel> record.map (function.const function.id) { a = 1 }
{ a = 1 }
nickel> record.map (function.const function.id) { a = 1, b = a }
error: contract broken by a function
  ┌─ :1:36
  │
1 │ forall a b. (Str -> a -> b) -> {_: a} -> {_: b}
  │                                    - expected type of an argument of an inner call
  │
  ┌─ repl-input-5:1:47
  │
1 │ record.map (function.const function.id) { a = 1, b = a }
  │                                               ^ applied to this expression
  │
  = This error may happen in the following situation:
        1. A function `f` is bound by a contract: e.g. `(Str -> Str) -> Str)`.
        2. `f` takes another function `g` as an argument: e.g. `f = fun g => g 0`.
        3. `f` calls `g` with an argument that does not respect the contract: e.g. `g 0` while `Str -> Str` is expected.
  = Either change the contract accordingly, or call `g` with a `Str` argument.
  = Note: this is an illustrative example. The actual error may involve deeper nested functions calls.

note:
  ┌─ <stdlib/record.ncl>:3:11
  │
3 │     map : forall a b. (Str -> a -> b) -> {_: a} -> {_: b}
  │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ bound here

note:
  ┌─ repl-input-5:1:47
  │
1 │ record.map (function.const function.id) { a = 1, b = a }
  │                                               - (1) calling <func>
```

Interestingly, all other ways I've tried to narrow down the issue evaluate without a hitch:
```
nickel> { a = 1, b = a } | { _: Num }
{ a = 1, b = 1 }

nickel> let f : forall a. {_: a} -> {_: a} = fun r => r
nickel> f { a = 1, b = a }
{ b = 1, a = 1 }
```

yannham:
Reporting here our ideas/findings during the weekly meeting: the bulk of the issue is that there are somehow two different - and natural, at least each one in some situations - interpretation of what could `record.map` mean on a recursive record.

## First interpretation: "outer" mapping

[...]

PR #1175, opened 2023-03-09
Note on polymorphic contracts and recursive records

yannham:
cf https://github.com/tweag/nickel/issues/1161#issuecomment-1462397141

### DIFF
--- a/notes/fixing-sealing-and-recursive-records.md
+++ b/notes/fixing-sealing-and-recursive-records.md
@@ -1,0 +1,174 @@
+# Int
+
+WIP. Notes on trying to get our head around an issue involving lazy record
+contracts, recursive record, and polymorphic sealing ().
+
+## Playgroud
+
+```
+{ foo = 1, bar = foo + 1} |> record.map (fun _key => (+) 1)
+```
+
+- type: forall a b. (String -> a -> b) -> {_: a} -> {_: b}
+
+Evaluation of the type to a contract:
+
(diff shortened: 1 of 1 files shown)


yannham:
cc @vkleen

notes/fixing-sealing-and-recursive-records.md
```diff
@@ -0,0 +1,227 @@
...
+up being paired one to one in program being evaluated. However, `%seal%` and
+`%unseal%` are generated in different places.
+
+If a contract containing an as yet unpaired `%seal%` or `%unseal%` ends up being
+attached to a record field, any reference to this field will have to apply the
+contract and essentially duplicates the `%seal%` or `%unseal%` operation. This
+means that if a type containing a free type variable is turned into a contract,
+and this contract gets attached to a record field for some reason, any reference
+to the field has the potential to produce a mismatch between `%seal%` and
+`%unseal%` primops in the program.
```
yannham, line 25: Might be useful to reproduce the original example here, just for the sake of (relative) self containment.

yannham: The example is detailed below, though. Still, it's very drafty, and it's probably better if one can understand what this is about without having to dive into the second part if not needed.

yannham (commented):
Thanks for taking care of this. I can't approve my own PR, but I'm fine with merging it in this state (though I left a review comment to be looked at)

vkleen (approved):

PR #1181, opened 2023-03-13
Feature/IncCache propagation

dpl0a:
A small part of the propagation rework for `IncCache`

### DIFF
--- a/src/eval/cache/incremental.rs
+++ b/src/eval/cache/incremental.rs
@@ -16,9 +16,10 @@ pub enum IncNodeState {
     Evaluated,
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct DependencyLink {
     id: Ident,
-    dependency: CacheIndex,
+    idx: CacheIndex,
 }
 
 /// A node in the dependent computation graph stored in [IncCache].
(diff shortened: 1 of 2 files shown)


src/eval/cache/incremental.rs
```diff
@@ -1,5 +1,6 @@
-/// A [Cache] implementation with incremental computation features.
+use std::collections::{HashMap, HashSet};
 
+/// A [Cache] implementation with incremental computation features.
```
yannham, line 3:
```suggestion
//! A [Cache] implementation with incremental computation features.
```

I think top-level documentation should be (but might not NEED be?) at the start of the file, and use the `//!` syntax.

src/eval/cache/incremental.rs
```diff
@@ -26,8 +33,10 @@ pub struct IncNode {
     bty: BindingType,
     // The state of the node.
     state: IncNodeState,
+    // Forward links to dependencies.
```
yannham, line 36: The documentation for links would be clearer by stating what "backward" and "forward" means in practice. Something like "Nodes that depends on this node" or "Dependencies of this node"

yannham (approved):
LGTM, but could do with a bit more documentation - even succinct - on what each function is actually doing, if only for your future self :slightly_smiling_face:

PR #1182, opened 2023-03-13
Use arbitrary precision rationals as the underlying representation of numbers

### LINKED ISSUE #1163: Use arbitrary precision numbers instead of floats as the default representation
The current `Num` (soon to be `Number`) primitive type is implemented as a float64. The rationale is that JSON doesn't have distinct number types (but representations don't really make sense for JSON, as no operations are actually carried on numbers, so they don't have to choose - mostly JSON has to fix an encoding, which is decimal number literals), and that in a configuration language, you probably want to use both integers and decimals numbers (say a load balancing factor, for example). You don't really care much about nor rely on the exact memory representation

As floats can represent a lot of integers exactly (up to `2^53` for floats64), and we don't want users to bother with two distinct number types, we started with _everything is a float64_. This is for example what Jsonnet does as well.

Floats have well known shortcomings, namely rounding errors. Once of the explicit goals of Nickel is to make configuration more correct, thanks to typing and contracts. Moreover, we expect most values to actually be integers, and a bit of decimals. Using floats silently as the _default_ representation bears a non-zero risk of leading to a non trivial misconfiguration happening, because of a rounding error.

Nickel isn't expected to be used to crush numbers. It's a high-level configuration language, and most execution time will be probably spent running functions, accessing recursive records or performing validation. Raw numbers crushing performance doesn't sound like a major criterion, design-wise.

We thus propose to make the default representation of numbers to be an arbitrary precision representation, using e.g. the [scientific crate](https://crates.io/crates/scientific). Doing so, we keep only one number type, accommodate for decimals without risking any loss of precision for arithmetic operations on rationals.

When we have to choose a precision (e.g. when performing a square root operation), we would default to the same precision as a float64: the idea being that when you evaluate a square root, there's no magical solution and you have to think about precision anyway, as a user. What we want to avoid is to get rounding errors like `0.1 + 0.1 + 0.1 != 3.0`.

The only drawback is that number operations may become more costly, but as stated before, we don't consider it being relevant in the overall performance picture.

Another solution is to introduce both floats and integers. This is what most languages do usually (e.g. CUE, Nix, Dhall, etc.). This has a number of drawbacks in a statically typed language without overloading: for example, we have to choose a type for `+`. It would probably be `Int -> Int -> Int` if integers are the most used. Then, we need a different operation for floats, such as Ocaml's `+.`. We have to define conversion rules (if any), how literals are interpreted, make users think about the representation, etc., for no clear benefit in our context.

yannham:
Closes #1163.

This pull request introduces arbitrary precision rationals as the underlying number representation for Nickel, instead of the current 64-bits float. The motivation and the rationale of the specific design choices are detailed in #1163.

## Choice of the arithmetic library

There are a few Rust crates for doing arbitrary-precision arithmetic:

- [rust-gmp](https://crates.io/crates/rust-gmp) is a direct binding to the well established GMP C library. It's however quite dry and low-level. Doesn't support a WASM build
- [rug](https://docs.rs/rug/latest/rug/#) is a high-level interface to GMP, MPFR and MPC. Because of its C binding, Rug doesn't support a WASM build.
- [ibig](https://crates.io/crates/ibig) is pure Rust, but only support big integers.
- [malachite](https://github.com/mhogrefe/malachite) is suggested in one of the previous's crate documentation if you need WASM. It seems quite complete and reasonably performant, also it doesn't have the same order of magnitude of usages/downloads.
- [scientific](https://github.com/alexkazik/scientific) could work as well, but looks small (2 stars, no issues, no PRs on the repo).

Among the crates that are both WASM-capable and handle arbitrary precision rationals, I chose to go with `malachite` which seemed to be the more solid. It's not out of the question to migrate to a different library anyway, if needed.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -1575,17 +1575,20 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                         // power is computed in an exact way.
                         //
                         // If the conversion fails, we fallback to converting both the exponent and
-                        // the value to an `f64`, perform the exponentiation, and convert the
-                        // result back to rationals, with a possible loss of precision.
-                        //
-                        // If one of the conversion fails, we error out.
+                        // the value to the nearest `f64`, perform the exponentiation, and convert
+                        // the result back to rationals, with a possible loss of precision.
                         let result = if let Ok(n2_as_i64) = i64::try_from(n2) {
                             n1.pow(n2_as_i64)
                         } else {
-                            Rational::try_from_float_simplest(
-                                f64::try_from(n1).unwrap().powf(f64::try_from(n2).unwrap()),
-                            )
-                            .unwrap()
+                            let result_as_f64 = f64::rounding_from(n1, RoundingMode::Nearest)
+                                .powf(f64::rounding_from(n2, RoundingMode::Nearest));
+                            // The following conversion fails if the result is NaN or +/-infinity
+                            Rational::try_from_float_simplest(result_as_f64).map_err(|_|
+                              EvalError::Other(
+                                  format!("invalid arithmetic operation: {n1}^{n2} returned {result_as_f64}, but {result_as_f64} isn't representable in Nickel"),
+                                  pos_op
+                              )
+                            )?
                         };
 
                         Ok(Closure::atomic_closure(RichTerm::new(
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -25,7 +25,7 @@ use crate::{
     stdlib::internals,
     term::{
         array::{Array, ArrayAttrs},
-        make as mk_term, number_approx_to_string,
+        make as mk_term,
         record::{self, Field, FieldMetadata, RecordAttrs, RecordData},
         BinaryOp, MergePriority, NAryOp, Number, PendingContract, RecordExtKind, RichTerm,
         SharedTerm, StrChunk, Term, UnaryOp,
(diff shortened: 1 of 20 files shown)


src/eval/operation.rs
```diff
@@ -604,48 +614,56 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
                         String::from("generate, 1st argument"),
                         arg_pos,
                         RichTerm { term: t, pos },
                     ))
+                };
+
+                if n < &Number::ZERO {
+                    return Err(EvalError::Other(
+                        format!(
+                            "generate: expected the 1st argument to be a positive number, got {n}"
```
vkleen, line 629: For the trace primop, I used the `builtin.trace` as the output prefix. Maybe this here should use `array.generate` as well? We're generally assuming that the functions in the stdlib are addressed by their full path (without trickery like `let`-rebinding).

vkleen (approved):

src/parser/utils.rs
```diff
@@ -19,11 +19,20 @@ use crate::{
...
-        BinaryOp, LabeledType, LetMetadata, RichTerm, StrChunk, Term, TypeAnnotation, UnaryOp,
+        BinaryOp, LabeledType, LetMetadata, Rational, RichTerm, StrChunk, Term, TypeAnnotation,
+        UnaryOp,
     },
     types::{TypeF, Types},
 };
 
+use malachite::num::conversion::traits::FromSciString;
+
+pub struct ParseNumberError;
```
ebresafegaga, line 30: Would it be useful if `ParseNumberError` stored the string that failed to parse?

ebresafegaga (approved):

src/parser/utils.rs
yannham: I think it is customary to not store anything inside parsing errors in Rust, because the caller to the `parse()` function surely still has access to the original `&str`. So unless we add more precise information, like an error message, or the incriminated position inside the string, just returning back the original argument is not really useful.

PR #1183, opened 2023-03-14
Rename arguments in `stdlib/internals.ncl`

vkleen:
Move away from single letter or abbreviated argument names in `stdlib/internals.ncl` for better readability.

### DIFF
--- a/stdlib/internals.ncl
+++ b/stdlib/internals.ncl
@@ -3,50 +3,50 @@
   # valid starting character for an identifier.
 
   # Contract implementations
-  "$dyn" = fun l t => t,
+  "$dyn" = fun _label value => value,
 
-  "$num" = fun l t => if %typeof% t == `Number then t else %blame% l,
+  "$num" = fun label value => if %typeof% value == `Number then value else %blame% label,
 
-  "$bool" = fun l t => if %typeof% t == `Bool then t else %blame% l,
+  "$bool" = fun label value => if %typeof% value == `Bool then value else %blame% label,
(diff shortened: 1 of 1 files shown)


stdlib/internals.ncl
```diff
@@ -3,50 +3,50 @@
...
+          (fun x => %assume% codomain (%go_codom% label) (value (%assume% domain (%chng_pol% (%go_dom% label)) x)))
       else
-          %blame% l,
+          %blame% label,
 
-  "$forall_var" = fun sy pol l t =>
-      let lPol = %polarity% l in
-      if pol == lPol then
-          %unseal% sy t (%blame% l)
+  "$forall_var" = fun sealing-key polarity label value =>
```
ebresafegaga, line 28: I didn't know Nickel variable names could have a `-` in them 😅

ebresafegaga (approved):
LGTM

stdlib/internals.ncl
vkleen: Not sure if that fits the coding style of stdlib right now, but I felt like kebab 😛

yannham: @ebresafegaga Yup :slightly_smiling_face: it was made to accommodate for potential target configurations that use  options with this style. So it's typically rather for exported record fields (whose name is not chosen by you), but all identifiers have the same syntax, so you can indeed use a dash inside a function variable or a let-binding.

yannham (approved):

PR #1184, opened 2023-03-14
[Refactor] Make label polarity an explicit enum

vkleen:
This refactor makes label polarities an explicit enum `label::Polarity` instead of a confusion-prone `bool`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1018,6 +1018,7 @@ mod blame_error {
         label::{
             self,
             ty_path::{self, PathSpan},
+            Polarity,
         },
         position::TermPos,
         term::RichTerm,
@@ -1031,9 +1032,9 @@ mod blame_error {
         if ty_path::has_no_arrow(&l.path) {
             // An empty path or a path that contains only fields necessarily corresponds to
             // a positive blame
-            assert!(l.polarity);
+            assert_eq!(l.polarity, Polarity::Positive);
             "contract broken by a value"
-        } else if l.polarity {
+        } else if l.polarity == Polarity::Positive {
             "contract broken by a function"
         } else {
             "contract broken by the caller"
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -1031,9 +1032,9 @@ mod blame_error {
         if ty_path::has_no_arrow(&l.path) {
             // An empty path or a path that contains only fields necessarily corresponds to
             // a positive blame
-            assert!(l.polarity);
+            assert!(matches!(l.polarity, label::Polarity::Positive));
             "contract broken by a value"
-        } else if l.polarity {
+        } else if let label::Polarity::Positive = l.polarity {
```
yannham, line 1037: It's a detail, but sometimes we use matching (as here), and sometimes we use `Eq` (as in the pattern match below). I'm always hesitating between the two: matching is more general (it doesn't require `Eq`) but it's also more verbose (require `matches!` in pattern guards, for example). Still, whatever you think best, I wonder if we should be more consistent (use `l.polarity == ...` here, or keep this one that way and use `matches!` below)

src/eval/operation.rs
```diff
@@ -387,7 +387,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
             UnaryOp::Pol() => {
                 if let Term::Lbl(l) = &*t {
                     Ok(Closure::atomic_closure(RichTerm::new(
-                        Term::Bool(l.polarity),
+                        l.polarity.into(),
```
yannham, line 390: That can be addressed into another PR, but I wonder if now that wouldn't make more sense to have polarity be `` [| `Positive, `Negative |] `` instead of a `Bool` in usercode.

yannham (approved):

src/error.rs
vkleen: Using `==` in a pattern guard seemed more sensible to me. I think I'll change this function to using `==` as well for consistency.

src/eval/operation.rs
vkleen: That's an excellent idea. It really shouldn't be a user-facing change, the label polarity is well hidden behind non-exposed primops.

vkleen: This turns out to be pretty easy to do and it doesn't even require changes to the Nickel standard library. It only ever compares polarities for equality and doesn't observe that they were `Bool`s.

PR #1186, opened 2023-03-15
[Refactor] Use type variable kinds when generating contracts

vkleen:
This refactor gets rid of confusing (presumably leftover) naming in the code to generate contracts from static types. Additionally, it makes use of type variable kinds for deciding whether a variable contract or a tail contract should be generated, instead of redetermining the kind during the type traversal.


src/types.rs
```diff
@@ -679,16 +679,10 @@ impl<'a> Iterator for EnumRowsIterator<'a, EnumRows> {
 /// Helper used by the `subcontract` functions. `pol` is the polarity of the variable occurrence
```
yannham, line 679: This comment might need to be fixed/updated.

yannham (approved):
Looks definitely better :+1:

PR #1189, opened 2023-03-16
Fix pretty printer with new multiline string syntax

vkleen:
The code determining how many `%` signs are needed for the delimiter in order to avoid spurious attempts at interpolation was not updated when the closing delimiters for multiline strings were changed.


yannham (approved):
Always count me in for getting rid of branching :+1:

PR #1191, opened 2023-03-20
Fix LSP crash when a resolved import term fails to typecheck

ebresafegaga:
Currently, when we open a Nickel file and some of content of one or more imports don't typecheck, the LSP crashes. 

This pull request fixes that.

### DIFF
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -3,7 +3,7 @@ use std::{collections::HashMap, ffi::OsString, io};
 use codespan::FileId;
 use nickel_lang::{
     cache::{Cache, CacheError, CacheOp, CachedTerm, EntryState},
-    error::Error,
+    error::{Error, ImportError},
     typecheck::{self, linearization::Linearization},
 };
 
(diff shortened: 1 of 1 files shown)


lsp/nls/src/cache.rs
```diff
@@ -48,11 +49,59 @@ impl CacheExt for Cache {
...
 
+        // After self.parse(), the cache must be populated
+        let CachedTerm {
+            term,
+            state,
+            parse_errs,
+        } = self.terms_mut().remove(&file_id).unwrap();
+
+        // We do this to get a list of the imports that have been resolved
+        // but don't typecheck, and then, we give an appriopiate diagnostics
```
yannham, line 65:
```suggestion
        // but don't typecheck, and then, we give an appropriate diagnostic
```

lsp/nls/src/cache.rs
```diff
@@ -48,11 +49,59 @@ impl CacheExt for Cache {
...
+        let mut errors = Vec::new();
+        let term = term.traverse::<_, _, ()>(
```
yannham, line 67: I don't think traversing the whole term again _unconditionally_ (even if no import failed) is reasonable. Some terms might be huge (@vkleen's auto-generated contracts).

If you need this information, would that be possible to build it up during linearization? Keeping e.g. a list of tuples `(location, import_id)`, or even a map  `HashMap<FileId, Vec<TermPos>>`. I won't include the imports that are in a file which failed to parse or typecheck, but I think it's fine for now: at least, a well-typed file will show if there is an issue with its imports.

Secondly, we might want to know if the previous `for` succeeded, in case we can avoid to do this search in the first place.

yannham (commented):
I think there are two things in one: not crashing on typechecking failure, and showing a nice error message. I have reservation with respect to the second point; however, the first one is valuable in itself. So if you want to separate the two (and remove the error-reporting part), I'm fine merging the first part already.

ebresafegaga:
Sure, I've made a separate pull request for the diagnostics part here #1214

yannham (approved):

lsp/nls/src/cache.rs
yannham: Because it seems harder to do. If you build the import during linearization, but the current file fails to typecheck, you might not have visited all the terms and the import node.

Also, the need is maybe less clear cut (at least to me): if the current file doesn't even typecheck, the user has already some local red squiggles to work on anyway. Then, once the current file typechecks, we can show imports that has failed, in a second time.

So, put differently, I believe ignoring the import statements in a file which doesn't typecheck seems easier to do, and the consequence doesn't seem too bad.

PR #1192, opened 2023-03-20
More generic contract generation

vkleen:
This PR introduces a type variable environment into labels and makes the tracking type variable polarities for contract generation dynamic. Additionally, it introduces two new primops to manipulate this environment.

This is preparatory work for #1177.

### DIFF
--- a/src/eval/operation.rs
+++ b/src/eval/operation.rs
@@ -16,7 +16,7 @@ use super::{
 use crate::{
     error::{EvalError, IllegalPolymorphicTailAction},
     identifier::Ident,
-    label::ty_path,
+    label::{ty_path, Polarity, TypeVarData},
     match_sharedterm, mk_app, mk_fun, mk_opn, mk_record,
     position::TermPos,
     serialize,
@@ -2746,6 +2746,39 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     pos2.into_inherited(),
                 )))
             }
+            BinaryOp::LookupTypeVar() => {
+                let t1 = t1.into_owned();
+                let t2 = t2.into_owned();
+
+                let Term::SealingKey(key) = t1 else {
+                    return Err(EvalError::TypeError(
+                        String::from("Sym"),
+                        String::from("lookup_type_variable, 1st argument"),
+                        fst_pos,
+                        RichTerm {
+                            term: t1.into(),
+                            pos: pos1,
+                        }
+                    ));
+                };
+
+                let Term::Lbl(label) = t2 else {
+                    return Err(EvalError::TypeError(
+                        String::from("Lbl"),
+                        String::from("lookup_type_variable, 2nd argument"),
+                        snd_pos,
+                        RichTerm {
+                            term: t2.into(),
+                            pos: pos2,
+                        }
+                    ));
+                };
+
+                Ok(Closure::atomic_closure(RichTerm::new(
+                    label.type_environment.get(&key).unwrap().into(),
+                    pos_op_inh,
+                )))
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/eval/operation.rs
```diff
@@ -3106,6 +3139,92 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        RichTerm {
+                            term: polarity,
+                            pos: pos2,
+                        },
+                    ))
+                };
+
+                let Term::Lbl(label) = &*label else {
+                    return Err(EvalError::TypeError(
+                        String::from("Lbl"),
```
yannham, line 3208:
```suggestion
                        String::from("Label"),
```

src/label.rs
```diff
@@ -366,8 +371,42 @@ pub struct Label {
...
     pub path: ty_path::Path,
+    /// An environment mapping type variables to [`TypeVarData`]. Used by
+    /// polymorphic contracts to decide which actions to take when encountering a `forall`.
+    pub type_environment: HashMap<SealingKey, TypeVarData>,
 }
 
+/// Data about type variables that is needed for polymorphic contracts to decide which actions to take.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TypeVarData {
+    pub polarity: Polarity,
```
yannham, line 382: I guess this field will take on more values once you do the "one-pass" version? Morally, `positive`, `negative`, and the composite partial identities `(positive, negative)` and `(negative, positive)` ?

src/label.rs
```diff
@@ -366,8 +371,42 @@ pub struct Label {
...
+                Field::from(RichTerm::from(Term::from(value.polarity))),
+            )]
+            .into(),
+            attrs: Default::default(),
+            sealed_tail: None,
+        })
+    }
+}
+
+impl HasUnifType for TypeVarData {
```
yannham, line 399: I find the name a tad misleading, I would expect a trait named `HasUnifType` to define a method `has_unif_type(&self) -> bool`. I don't know of a good name scheme for those kind of `* -> *` functions from a type to a type. `AssociatedUnifType`? Looks a bit long. `ReifyAsUnifType`? `ReifyAsNickelType`? I like reify, it sounds professional :sweat_smile:

stdlib/internals.ncl
```diff
@@ -25,9 +25,10 @@
       else
           %blame% label,
 
-  "$forall_var" = fun sealing_key polarity label value =>
-      let current-polarity = %polarity% label in
-      if polarity == current-polarity then
+  "$forall_var" = fun sealing_key label value =>
+      let current_polarity = %polarity% label in
+      let {polarity, ..} = %lookup_type_variable% sealing_key label in
```
yannham, line 30: I don't know if it makes a big difference, but destructuring involves a contract application and thus merging. It's a bit less visible, but I would use a more down-to-earth `let var_data = ...` and `var_data.polarity` in such low-level code, which is not very pretty anyway, to be avoid the codepath of merging to be taken.

But honestly it couldn't also not make any noticeable difference at all, I'm not sure.

```suggestion
      let {polarity, ..} = %lookup_type_variable% sealing_key label in
```

yannham (approved):
It's not urgent, but looking at code each time we add a new primop, it really feels cumbersome to do (add stuff to the lexer, to the parser, implement boiler plate code for dynamically typing the arguments, extract them, oh should I take ownership with `match_shared!` or not :man_shrugging: ). It's not urgent but there must be a better way of organizing (and splitting, as well) `eval::operation`.

src/label.rs
vkleen: I was expecting to need more data for type variables, indeed. Originally I was also tracking the variable kind, but that ended up being unnecessary for now.

src/label.rs
vkleen: +1 for reify 👑 Another possibility would be `AsUnifType`. Really, the problem seems to be that `TypeVarData::unif_type()` should actually be an associated constant; and that the trait can't be named `UnifType` to conform to the usual naming format.

yannham: Right, I would also have called it `UnifType` naturally.

vkleen: I've renamed it to `ReifyAsUnifType`. Yell if that's a bad idea 😅

PR #1193, opened 2023-03-20
Follow more closely the bidrectional typechecking discipline

yannham:
First part of #584.

Adds an `infer` and a  `subsumption` functions (corresponding to the matching rules of bidirectional typechecking), and in general follow more closely the bidirectional typechecking discipline, in particular for elimination rules (function application).

This PR shouldn't change much the behavior of typechecking, beside different error messages. This is preparatory work for closing #584, where the next step is to do proper type instantiation with variable levels, instead of the current ad-hoc instantiation happening at some random places (random = where it has been needed to make things work most of the time).

The error message when a record field hasn't the right type has lost a bit of context: while it's still ok, we plan to address this issue and bring back the previous error style in a subsequent PR.


src/typecheck/mod.rs
```diff
@@ -1,38 +1,46 @@
...
-//! statically typed block (an expression annotated with a type), and is switched back to
-//! non-strict mode when entering an expression annotated with a contract. Type and contract
-//! annotations thus serve both another purpose beside enforcing a type or a contract, which is to
-//! switch the typechecking mode.
+//! Corresponding to dynamic typing static typing, the typechecking algorithm runs in two separate
+//! modes:
+//!
+//! - **enforce** corresponds to traditional typechecking in a statically typed language. This
+//!   happens inside a statically typed block. Such blocks are introduced by the type ascription
+//!   operator `:`, as in `1 + 1 : Num` or `let f : Num -> Num = fun x => x + 1 in ..`. Enforce
```
matthew-healy, line 14:
```suggestion
//!   operator `:`, as in `1 + 1 : Number` or `let f : Number -> Number = fun x => x + 1 in ..`. Enforce
```

src/typecheck/mod.rs
```diff
@@ -1,38 +1,46 @@
...
+//!
+//! Note that the static typing part is based on the bidirectional typing framework, which defines
+//! two different modes. Thus, the enforce mode is itself divided again into **checking** mode and
+//! **inference** mode.
 //!
 //! # Type inference
 //!
 //! Type inference is done via a form of bidirectional typechecking coupled with unification, in
-//! the same spirit as GHC (Haskell), albeit the type system of Nickel is much simpler. The type of
+//! the same spirit as GHC (Haskell), albeit the type system of Nickel is simpler. The type of
```
matthew-healy, line 31:
```suggestion
//! the same spirit as GHC (Haskell), although the type system of Nickel is simpler. The type of
```

src/typecheck/mod.rs
```diff
@@ -1077,9 +1085,41 @@ fn walk_with_annot<L: Linearizer>(
...
+/// Elimination rules (such as function application or primitive operator application) are only in
+/// inference mode (still following Pfenning's recipe). `check` follows the inference mode here
+/// (typically on function application, where we first call to `infer` on the function part, and
+/// then check the argument).
+///
+/// Still, `check` is supposed to be implementing checking mode from the outside. We thus also
+/// apply the rule which switches from inference to checking mode. This is in particular where
+/// subtyping happens, for example, when there is subtyping. Currently, subtyping isn't supported
+/// yet in Nickel, so the corresponding switching rule doesn't do much. But the implementation of
+/// RFC004 will bring subtyping, and [`subsumption`] will be the place to apply it.
```
vkleen, line 1117: This sentence confused me for a second. It looked like a leftover from editing "for example, when there is subtyping", but it makes standalone sense as well. My suggestion hopefully makes this clearer.
```suggestion
/// apply the rule which switches from inference to checking mode. This occurs for example where
/// subtyping happens, at least when there is subtyping. Currently, subtyping isn't supported
/// in Nickel, yet, so the corresponding switching rule doesn't do much. But the implementation of
/// RFC004 will bring subtyping, and [`subsumption`] will be the place to apply it.
```

src/typecheck/mod.rs
```diff
@@ -1077,9 +1085,41 @@ fn walk_with_annot<L: Linearizer>(
     }
 }
 
-/// Typecheck a term against a specific type.
+/// Check a term against a given type. Although this method mostly corresponds to checking mode in
+/// the classical bidirectional framework, it combines both checking and inference mode in
```
matthew-healy, line 1089:
```suggestion
/// the classical bidirectional framework, it combines both checking and inference modes in
```

src/typecheck/mod.rs
```diff
@@ -1077,9 +1085,41 @@ fn walk_with_annot<L: Linearizer>(
...
-/// # Arguments
+/// Following Pfenning's recipe (see [Bidirectional Typing][bidirectional-typing]),
+/// introduction rules (e.g. typechecking a record) are checking. `check` follows the same logic
+/// here: it uses unification to "match" on the expected type (in the case of record, a record type
+/// or a dictionary type) and pushes typechecking down the record fields.
+///
+/// # Elimination rules
+///
+/// Elimination rules (such as function application or primitive operator application) are only in
+/// inference mode (still following Pfenning's recipe). `check` follows the inference mode here
```
matthew-healy, line 1109:
```suggestion
/// Elimination rules (such as function application or primitive operator application) only exist in
/// inference mode (still following Pfenning's recipe). `check` follows the inference mode here
```

src/typecheck/mod.rs
```diff
@@ -1497,16 +1600,16 @@ fn type_check_annotated<L: Linearizer>(
     rt: &RichTerm,
     ty: UnifType,
 ) -> Result<(), TypecheckError> {
-    type_check_with_annot(state, ctxt, lin, linearizer, annot, Some(rt), ty, rt.pos)
+    check_with_annot(state, ctxt, lin, linearizer, annot, Some(rt), ty, rt.pos)
 }
 
-/// Subfunction handling the common part of typechecking term with type or contract annotation, with
+/// Function handling the common part of typechecking term with type or contract annotation, with
```
vkleen, line 1606:
```suggestion
/// Function handling the common part of typechecking terms with type or contract annotation, with
```

src/typecheck/mod.rs
```diff
@@ -1077,9 +1085,41 @@ fn walk_with_annot<L: Linearizer>(
...
+/// # Elimination rules
+///
+/// Elimination rules (such as function application or primitive operator application) are only in
+/// inference mode (still following Pfenning's recipe). `check` follows the inference mode here
+/// (typically on function application, where we first call to `infer` on the function part, and
+/// then check the argument).
+///
+/// Still, `check` is supposed to be implementing checking mode from the outside. We thus also
+/// apply the rule which switches from inference to checking mode. This is in particular where
+/// subtyping happens, for example, when there is subtyping. Currently, subtyping isn't supported
```
matthew-healy, line 1115:
```
This is in particular where subtyping happens, for example, when there is subtyping.
```

I'm not entirely sure what this sentence is trying to say. :sweat_smile:

vkleen (approved):
Looking good! I've found a few typos; the logic seems okay to my lay eyes.

src/typecheck/mod.rs
```diff
@@ -1381,51 +1435,88 @@ fn type_check_<L: Linearizer>(
...
+                    .map(|id| (*id, state.table.fresh_type_uvar()))
+                    .collect();
+
+                let rows_skeleton =
+                    rows_expected_type
+                        .iter()
+                        .fold(mk_uty_row!(), |acc, (id, row_ty)| {
+                            // if let Term::RecRecord(..) = t.as_ref() {
+                            //     ctxt.type_env.get(id).cloned().unwrap()
+                            // }
```
matthew-healy, line 1451: Is there a reason to leave this commented-out code here?

src/typecheck/mod.rs
```diff
@@ -1381,51 +1435,88 @@ fn type_check_<L: Linearizer>(
...
+                    check_field(
+                        state,
+                        ctxt.clone(),
+                        lin,
+                        linearizer.scope(),
+                        *id,
+                        field,
+                        // unwrap(): we've built `rows_expected_type` in this very function
+                        // from record.fields.keys(), so it must contain `id`
+                        rows_expected_type.remove(id).unwrap(),
```
matthew-healy, line 1484: Nitpick: we could use `expect` here & put the comment into the string argument.

matthew-healy (approved):
This looks great! I left a few comments on the docs, but the code itself looks good-to-go. :+1:

src/typecheck/mod.rs
yannham: Out of curiosity, what is the rational here? Does albeit require a be+ing form after? Is there a nuance in the meaning? Or does it just sound better to you?

src/typecheck/mod.rs
yannham: :sweat_smile: the switch between inference and checking mode is where we apply subtyping in the general framework. Currently, it's a degenerate case, because there is no subtyping yet (or, viewed differently, subtyping is trivially the equality relation). But it's indeed quite confusing. I think @vkleen suggestion below makes it better.

src/typecheck/mod.rs
yannham: I agree the original wording is bad. However, my pet-peeve with this suggestion is that it kinda reverse the relation between the switch rule and subtyping: it makes it sound like subtyping (when there is) happens at some places, depending on the precise type system design, and those places are where we should also switch from inference mode to checking mode.

In fact, it is the converse: the switching rule is pre-existing and makes sense alone, and if you have subtyping and follows the bidirectional approach, the only place you should/can apply subytping is when switching.

I'll try to reword this part anyway.

src/typecheck/mod.rs
yannham: Ah, no, it's a left-over. Thanks.

src/typecheck/mod.rs
matthew-healy: Initially it was just because it intuitively sounded better, but I just looked it up & apparently "albeit" should only be used if the clause it introduces is *not* an independent clause. (See [here](https://www.scribbr.com/definitions/albeit/)). In this case "the type system of Nickel is simpler" is an independent clause - i.e. it would be a valid sentence on its own - so "although" should be used instead of "albeit" here.

But I'm not gonna pretend I knew any of that when I wrote the comment. :joy:

src/typecheck/mod.rs
yannham: Hmm, I always wonder about this dilemma. My usual line of thought is that I don't see a big advantage of doing so:

1. The error message is useless without the code around it: you're not going to extract any useful information from it
2. I don't need it to find the location of the panic, because Rust tracks that for me.

On the other hand, there isn't a big disadvantage of using `expect` either (beside adding a few bytes to the binary). What's your personal rationale for using `expect` ?

matthew-healy: I tend to view comments as a last-resort method for expressing the purpose of code. My reasoning being that they require manual labour to ensure they're kept in sync with the code they describe. As such, in places where the comment can be merged into the code in some form, my preference is always to do so. 

Here, by using `expect` rather than `unwrap`, we make the justification for `panic`king part of the code that causes the `panic`, which means if that code changes, the justification is more likely to change with it. 

But I marked it as a "nitpick" for a reason: my feelings on this aren't massively strong, and while I'd do it in my own code, I wouldn't insist on it in code written by others. :smile:

matthew-healy: We could always try using `flux` instead https://github.com/flux-rs/flux :trollface:

src/typecheck/mod.rs
yannham: Tentative improvement: https://github.com/tweag/nickel/pull/1193/commits/3b674c88b6c2187cdce6eca9e79f74153d789e1b

vkleen: I like it 👍

Issue #935, opened 2022-11-23
Tracking issue: stdlib audit 


matthew-healy:
We should work out what the desired standard library interface is. This includes (but is not limited to):

- naming (e.g. `fold` vs `foldLeft`)
- typing (e.g. `record.values` currently has contract `{ ; Dyn } -> Array Str`, but perhaps could have `forall r. { ; r } -> Array Str`)
- filling in any gaps (e.g. `record.map' : forall a b. (Str -> a -> { key: Str, value: b }) -> { _: a } -> { _: b }` from today's Nickel Hour)
- removing any functions we don't wish to support long-term

matthew-healy:
An idea that came up today: there are certain functions involving polymorphic record contratcs which respect parametricity but due to the contracts involved are not supported in Nickel today. An example would be:

```
removeFoo | forall t r. { foo: t; r } -> { ; r } = fun r => record.remove "foo" r,
```

This _seems_ valid, but currently fails because `record.remove` has type `forall a. Str -> {_: a} -> {_: a}`, and that type gets converted into a contract whose implementation calls `%record_map%`, which _can_ violate parametricity, so is not allowed.

A workaround is to use the primitive operation `%record_remove%`, for which we do not generate a contract.

One way of supporting the above program _without_ encouraging users to use primitive ops would be to expose an `unsafe` module from the standard library, which contained untyped versions of various functions. The above function could then be written as:

```
removeFoo | forall t r. { foo: t; r } -> { ; r } = fun r => unsafe.record.remove "foo" r,
```

matthew-healy:
IMO it's a little weird that `join : Str -> Array Str -> Str` is accessed via `string.join` rather than `array.join`. I get that it's not a "proper" join as we can't express it generically, but I was still surprised to find it there.

matthew-healy:
We should probably consider #1007 part of this issue.

yannham:
As discussed before, let's gather here proposals for addition to the stdlib as well. Taking those of #321 here (originally proposed by @silverraven691):

- [ ] `nums.clamp : Num -> Num -> Num -> Num`
- [ ] `functions.flip : forall a b c. (a -> b -> c) -> b -> a -> c`
- [ ] `lists.optional : Bool -> List -> List`
- [ ] `records.optional : Bool -> { _: Dyn } -> { _: Dyn }`
- [ ] `records.filter : (Str -> Dyn -> Bool) -> { _: Dyn } -> { _: Dyn }`
[...]

Issue #420, opened 2021-10-12
Base types for contracts (or contracts seen as refinement types)


yannham:
**Is your feature request related to a problem? Please describe.**
Very often, user-defined contracts start along these lines:

```
let Port = fun label value =>
  if builtins.isNum value then
    ...
  else
    builtins.blame (builtins.tag "not a number" label) in

let Url = fun label value =>
  if builtins.isStr value then
    ...
  else
    builtins.blame (builtins.tag "not a string" label) in

...
```

That is, most contracts accept values that are in a subset of an existing builtin type (they are (dynamic) refinement types). This boilerplate feels like it could be added automatically by declaring explicitly a base type for a contract. Moreover, this information can then be leveraged by the type system:

1. Conservatively in one direction, uniquely allowing safe implicit upcast from, say, a `#Port` to a `Num`, but not the other way around (this sets `#Port` as a subtype of `Num`).
1. Conservatively in both directions, allowing implicit casts from `#Port` to `Num` as in 1., as well as the reverse `Num` to `#Port` but protecting the unsafe conversion by inserting automatically a contract check. This corresponds more to the philosophy of traditional gradual typing which is to statically allow casts and insert checks automatically between "compatible types" (casts that have a chance of success), but would depart from Nickel's current design to require any statically unsafe cast to be written explicitly.
1. Forgetting the refinement represented by the contract altogether and just considering all `#Port`s as `Num`s statically

**Describe the solution you'd like**

A syntax to declare that a contract is a refinement of an existing builtin static type, as it is the case for example in the system of [Well-typed program can't be blamed](https://homepages.inf.ed.ac.uk/wadler/papers/blame/blame.pdf) (although they don't do it exactly for the same reasons). Syntax suggestion (`<:` is not taken and fit well with the subtyping interpretation):

```
let Port <: Num = fun label value => ...
```
Dynamically, this would wrap the implementation of `Port` with a code that checks if value is a string in the first place. Something like:

```
let refine = fun typeName hasType contract label value =>
 if hasType value then
    contract label value
  else
    builtins.blame (builtins.tag "not a #{typeName}" label) in

let Port <: Num = ...
// rewrites to
let Port_ = ...
let Port = refine "number" builtins.isNum Port_
```
Statically, it can be interpreted as any of the alternatives mentioned above, but this can be decided later. We could probably actually support having any contract (including all static types that are convertible to contracts) as a base by defining `refine` as:

```
let refine = fun baseName baseContract contract label value =>
[...]

PR #1195, opened 2023-03-23
Stdlib array extension

yannham:
Following #935, add a batch of missing functions to the array stdlib. This PR adds a `slice` primitive operation as well, required to implement the corresponding function of the stdlib. This PR adds a new `enum` stdlib module as well, moving a few functions/contracts from the `string` module.

Doing so, we also expanded the documentation of folds, by adding a more informal description as well as a paragraph on `fold_right` vs `fold_left`. This is useful as a reference for the documentation of `reduce_right` and `reduce_left` too.

## Follow-up

We can now get rid of the `%tail%` primop, which is subsumed by `%slice% 1 (%length% <.>)`. In the same spirit, we should probably get rid of `%head%` in favor of `%elem_at% 0`. This is planned in a follow-up PR.

### DIFF
--- a/src/term/array.rs
+++ b/src/term/array.rs
@@ -78,7 +78,7 @@ impl Array {
     /// Resize the view to be a a sub-view of the current one, by considering a slice `start`
     /// (included) to `end` (excluded).
     ///
-    /// The parameters must verify `0 <= start <= end <= self.end - self.start`. Otherwise,
+    /// The parameters must satisfy `0 <= start <= end <= self.end - self.start`. Otherwise,
     /// `Err(..)` is returned.
     pub fn slice(&mut self, start: usize, end: usize) -> Result<(), OutOfBoundError> {
         if start > end || end > self.len() {
--- a/src/term/array.rs
+++ b/src/term/array.rs
@@ -53,12 +53,18 @@ impl ArrayAttrs {
     }
 }
 
+/// A Nickel array, represented as a view (slice) into a shared backing array. The view is
+/// delimited by `start` (included) and `end` (excluded). This allows to take the tail of an array,
+/// or an arbitrary slice, in constant time, providing actual linear time iteration when
+/// imlementing recursive functions, such as folds, for example.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Array {
     inner: Rc<[RichTerm]>,
     start: usize,
     end: usize,
 }
+
+pub struct OutOfBoundError;
 
 impl Array {
     /// Creates a Nickel array from reference-counted slice.
@@ -67,6 +73,23 @@ impl Array {
         let end = inner.len();
 
         Self { inner, start, end }
+    }
+
+    /// Resize the view to be a a sub-view of the current one, by considering a slice `start`
+    /// (included) to `end` (excluded).
+    ///
+    /// The parameters must verify `0 <= start <= end <= self.end - self.start`. Otherwise,
+    /// `Err(..)` is returned.
+    pub fn slice(&mut self, start: usize, end: usize) -> Result<(), OutOfBoundError> {
+        if start > end || end > self.len() {
+            return Err(OutOfBoundError);
+        }
+
+        let prev_start = self.start;
+        self.start = prev_start + start;
+        self.end = prev_start + end;
+
+        Ok(())
     }
 
     /// Returns the effective length of the array.
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -390,7 +390,7 @@
         "%
       = fun start end step => %generate%
           ((end - start) / step |> number.floor)
-          (fun i => start + i*step), 
+          (fun i => start + i*step),
 
     range
       : Number -> Number -> Array Number
@@ -402,7 +402,7 @@
       = fun start end => range_step start end 1,
 
     reduce_right
-      : forall a. (a -> a -> a) -> Array a -> a 
+      : forall a. (a -> a -> a) -> Array a -> a
       | doc m%"
            Fold a function over a array.
         `fold_right f init [x1, x2, ..., xn]` results in `f x1 (f x2 (... (f xn init) ...))`.
@@ -414,9 +414,9 @@
             [ 3, 2, 1 ]
         ```
         "%
-      = fun f array => 
-          let last = last array in 
-          let rest = drop_last array in 
+      = fun f array =>
+          let last = last array in
+          let rest = drop_last array in
           fold_right f last rest,
   }
 }
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -346,5 +346,77 @@
         if %length% arr <= 1
         then arr
         else [ %head% arr ] @ (%tail% arr |> flat_map (fun a => [v, a])),
+
+    slice
+      : forall a. Number -> Number -> Array a -> Array a
+      | doc m%"
+          `slice start end array` the sub-array of `array`, starting at `start`
+(include  d) and ending at `end` (excluded).
+
+          TODO: errors (start < 0 , etc.)
+        "%
+      # TODO: actually implement the operation using a slice primop
+      = fun start end value => %slice% start end value,
+
+    split_at
+      : forall a. Number -> Array a -> {left: Array a, right: Array a}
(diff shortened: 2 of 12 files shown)


src/term/array.rs
```diff
@@ -69,6 +75,23 @@ impl Array {
         Self { inner, start, end }
     }
 
+    /// Resize the view to be a a sub-view of the current one, by considering a slice `start`
+    /// (included) to `end` (excluded).
+    ///
+    /// The parameters must verify `0 <= start <= end <= self.end - self.start`. Otherwise,
```
vkleen, line 81:
```suggestion
    /// The parameters must satisfy `0 <= start <= end <= self.end - self.start`. Otherwise,
```

stdlib/array.ncl
```diff
@@ -346,5 +408,200 @@
         if %length% arr <= 1
         then arr
         else [ %head% arr ] @ (%tail% arr |> flat_map (fun a => [v, a])),
+
+    slice
+      : forall a. Number -> Number -> Array a -> Array a
+      | doc m%"
+          Results in the slice of the given array between `start` (included) and
+          `end` (excluded).
```
vkleen, line 416: I'm not a big fan of the "Results in" formulation. It sounds unidiomatic to me. But I'm not a native English speaker, so take that with a grain of salt. While we don't have a `return` construct in Nickel, I feel it would be better to use `Returns` in the documentation.

A separate issue is that we don't have a way of naming parameters in type signatures (outside of using records). So when reading this first sentence, it's not actually clear which parameters `start` and `end` refer to. Similar situations occur a lot in the stdlib documentation and spelling out a call like in this suggestions seems to be the best solution.
```suggestion
          `slice start end array` returns the slice of `array` between `start` (included) and
          `end` (excluded).
```

stdlib/array.ncl
```diff
@@ -346,5 +408,200 @@
...
+      | doc m%"
+          `range_step start end step` generates the array of numbers
+          `[start, start + step, start + 2*step, ..]` up to the first element
+          (excluded) larger than or equal to `end`
+
+          # Preconditions
+
+          In `range_step start end step`, `start` and `end` must verifies `start
+          <= end`.`step` must be strictly greater than `0` (or the function will
+          loop).
```
vkleen, line 495: I don't think the function will actually loop if `step` is 0. Instead it will probably throw a division by 0 error when trying to calculate the length of the result.
```suggestion
          In `range_step start end step`, `start` and `end` must satisfy `start
          <= end`. `step` must be strictly greater than `0`.
```

stdlib/array.ncl
```diff
@@ -346,5 +408,200 @@
...
+          `range start end` generates the array of numbers
+          `[start, start + 1, start + 2, ..]` up to the first element
+          (excluded) larger than or equal to `end`.
+
+          `range start end` is equivalent to `range_step start end 1`.
+
+          # Preconditions
+
+          In `range_step start end`, `start` and `end` must verify
+          `start <= end`.
```
vkleen, line 520:
```suggestion
          In `range_step start end`, `start` and `end` must satisfy
          `start <= end`.
```

stdlib/array.ncl
```diff
@@ -346,5 +408,200 @@
...
+      : forall a. (a -> a -> a) -> Array a -> a
+      | doc m%"
+        Reduces the elements to a single one, by repeatedly applying a reducing
+        operation. If the array is empty, returns an empty array.
+
+        `reduce_left` associates to the left, that is
+        `reduce_left op [x1, x2, ..., xn]` results in `op (... (op (op x1 x2) x3) ...) xn`.
+
+        `reduce_left` is the same as `fold_left`, but uses the first element as
+        the initial element.
```
vkleen, line 541:
```suggestion
        the initial accumulator.
```

stdlib/array.ncl
```diff
@@ -346,5 +408,200 @@
...
+            => -8
+        ```
+        "%
+      = fun f array =>
+        if array == [] then
+          array
+        else
+          let first = %elem_at% array 0 in
+          let rest = %array_slice% 1 (%length% array) array in
+          fold_left f first rest,
```
vkleen, line 566: I don't think returning an empty array when given one makes sense here (How does this even typecheck?). It would probably be best to somehow fail on an empty array; maybe by applying the `array.NonEmpty` contract somewhere?

stdlib/enum.ncl
```diff
@@ -0,0 +1,62 @@
...
+          "%
+        = contract.from_predicate builtin.is_enum,
+
+      TagOrString
+        | doc m%%"
+            Accepts both enum tags and strings. Strings are automatically
+            converted to an enum tag.
+
+            `TagOrString` is typically used in conjunction with an enum type, to
+            accept tags represented as string (e.g. coming from a JSON
```
vkleen, line 26:
```suggestion
            accept tags represented as strings (e.g. coming from a JSON
```

vkleen (approved):

stdlib/array.ncl
```diff
@@ -116,12 +117,62 @@
...
+          array.fold_right (-) 0 [1, 2, 3, 4]
+           => -2
+          array.fold_left (-) 0 [1, 2, 3, 4]
+           => -10
+          ```
+        - If the folded function is associative, both `fold_right` and
+          `fold_left` return the same result. In that case, **`fold_left` is
+          generally preferred**, because it forces the evaluation of the
+          intermediate results resulting in less memory consumption and overall
+          better performances (outside of pathological cases).
```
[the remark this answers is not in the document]
yannham: I have no strong opinion on this one, but isn't "performances" as well? In that overall, it should lead to better performances, all considered?

stdlib/array.ncl
yannham: It doesn't make any sense, I think I thought about returning `None` and may brain then failed. We should fail on this, probably.

The issue I had with more specific contracts like `NonEmpty` is that they make those standard functions a pain to use in statically typed code. On one hand it's a shame, because we have those nice contracts with argument tracking and whatnot which could provide really good error messages. On the other hand, making folds, reduce and map a pain to use in typed code is also kind of a show stopper. One solution could be to separate them in submodules (having a more loosely typed version, and a tight contract-protected version), but that isn't satisfying either.

Maybe one solution at some point would be the ideas exposed in #420.

stdlib/array.ncl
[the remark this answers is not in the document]
vkleen: I don't think "performance" is a countable noun in English. At least I can't think of a situation where "performances" doesn't sound weird to me.

stdlib/array.ncl
yannham: Would that work? Now `array` has type `NonEmpty`, and you can't do anything with it in typed code anymore, I fear.

yannham: Or maybe something like `%seq% (array | NonEmpty) ...restofthefunction...` but this starts to be really convoluted :sweat_smile:

yannham: Maybe this calls for a construct to check a contract without influencing typechecking

stdlib/array.ncl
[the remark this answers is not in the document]
yannham: It seems performance can be countable when used with this meaning (https://english.stackexchange.com/questions/367488/plural-of-performance). Still, it looks less common and natural than using the uncountable version. It might very well be my own deformation, as it's more common to use the pural "les performances" than the singular "la performance" in French, in this situation.

stdlib/array.ncl
vkleen: So, we'd write
```
reduce_left
  : forall a. (a -> a -> a) -> Array a -> a
  | forall a. (a -> a -> a) -> NonEmpty -> a
  = ...
```
I guess? That's a little bit of duplication, but I think it's a price worth paying here 👍

yannham: We can also just add the "differential" contract, so to speak:

```
reduce_left
  : forall a. (a -> a -> a) -> Array a -> a
  | Dyn -> NonEmpty -> Dyn
```

That might make for a slightly worse error message, though. We could also duplicate the type fully as you propose, and rely on contract elision (yet to be added) to not do too much work.

PR #1196, opened 2023-03-23
Audit stdlib documentation

vkleen:
For now, a quick pass over the stdlib documentation for consistency and typos.

### DIFF
--- a/stdlib/array.ncl
+++ b/stdlib/array.ncl
@@ -2,15 +2,16 @@
   array = {
     NonEmpty
       | doc m%"
-        Contract to ensure the given array is not empty.
+          Enforces that an array is not empty.
 
-        For example:
-        ```nickel
+          # Examples
+
+          ```nickel
           ([] | NonEmpty) =>
             error
           ([ 1 ] | NonEmpty) =>
             [ 1 ]
-        ```
+          ```
         "%
       = fun label value =>
         if %typeof% value == `Array then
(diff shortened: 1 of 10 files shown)


stdlib/array.ncl
```diff
@@ -98,9 +105,10 @@
 
     at : forall a. Number -> Array a -> a
       | doc m%"
-        Retrieves the n'th element from a array (0-indexed).
+        Retrieve the n'th element from an array, with indices starting at 0.
```
yannham, line 108: I think we used "Returns", so it's fair to keep the `s` everywhere.
```suggestion
        Retrieves the n-th element from an array, with indices starting at 0.
```

stdlib/array.ncl
```diff
@@ -231,9 +240,10 @@
 
     prepend : forall a. a -> Array a -> Array a
       | doc m%"
-        Construct an array given the first element and the rest of the array.
+        Build an array given the first element and the rest of the array.
```
yannham, line 243:
```suggestion
        Builds an array given the first element and the rest of the array.
```

stdlib/array.ncl
```diff
@@ -256,9 +267,10 @@
 
     reverse : forall a. Array a -> Array a
       | doc m%"
-        Reverses the order of a array.
+        Reverse an array.
```
yannham, line 270:
```suggestion
        Reverses an array.
```

stdlib/array.ncl
```diff
@@ -280,9 +293,10 @@
 
     flatten : forall a. Array (Array a) -> Array a
       | doc m%"
-        Flatten a array of arrays to a single array, essentially concatenating all arrays in the original array.
+        Concatenate all elements of a given array of arrays.
```
yannham, line 296:
```suggestion
        Concatenates all elements of a given array of arrays.
```

stdlib/array.ncl
```diff
@@ -386,29 +409,31 @@
...
+        First `map` the given function over the array and then `flatten` the
         result.
 
-        For example:
+        # Examples
+        
         ```nickel
-        flat_map (fun x => [x, x]) [1, 2, 3]
-          => [1, 1, 2, 2, 3, 3]
+          flat_map (fun x => [x, x]) [1, 2, 3]
```
yannham, line 418: Any reason for the additional indentation, or it's just an artifact ?

stdlib/builtin.ncl
```diff
@@ -44,9 +47,10 @@
 
     is_enum : Dyn -> Bool
     | doc m%"
-      Checks if the given value is an enum tag.
+      Check if a value is an enum tag.
```
yannham, line 50:
```suggestion
      Checks if a value is an enum tag.
```

stdlib/builtin.ncl
```diff
@@ -169,9 +180,10 @@
 
     serialize : [| `Json, `Toml, `Yaml |] -> Dyn -> String
     | doc m%"
-      Serializes the given value to the desired representation.
+      Serialize a value into the desired representation.
```
yannham, line 183: Is serialize into more idiomatic than serializing to a format ? The latter actually sounds more natural to me, but I can't say for sure.

stdlib/function.ncl
```diff
@@ -2,72 +2,83 @@
...
+        id x = x
+      ```
+      for any value `x`.
     "%
     = fun x => x,
 
     const : forall a b. a -> b -> a
     | doc m%"
-    Results in the first argument.
+      Always return the first argument, ignoring the second.
```
yannham, line 15:
```suggestion
      Always returns the first argument, ignoring the second.
```

stdlib/function.ncl
```diff
@@ -2,72 +2,83 @@
...
     "%%
     = fun f x y => f y x,
 
     first : forall a b. a -> b -> a
     | doc m%"
-    Always returns the first argument:
-    ```nickel
-    first 5 7 => 5
-    ```
+      Always return the first argument, ignoring the second.
```
yannham, line 49:
```suggestion
      Always returns the first argument, ignoring the second.
```

stdlib/function.ncl
```diff
@@ -2,72 +2,83 @@
...
     "%
     = const,
 
     second : forall a b. a -> b -> b
     | doc m%"
-    Always returns the second argument:
-    ```nickel
-    second 5 7 => 7
-    ```
+      Always return the second argument, ignoring the first.
```
yannham, line 61:
```suggestion
      Always returns the second argument, ignoring the first.
```

yannham (commented):
So I had to stop at some point with suggestions, there were just too many lines :sweat: but I think we should be consistent over:

- verbal forms. If we start some documentation with `Returns`, then all verbs describing the action of a function should use the `s` form.
- indentation. I believe there are cases where the code is indented with respect to the block fences, but not everywhere. Also, I don't think it's standard practice, it feels like it should start the the same level as the block, as I don't see much usefulness in having the whole code block indented by one level from the leftmost boundary.

vkleen:
> So I had to stop at some point with suggestions, there were just too many lines

No worries :laughing: I'll find all the lines.

>     * verbal forms. If we start some documentation with `Returns`, then all verbs describing the action of a function should use the `s` form.
👍

>     * indentation. I believe there are cases where the code is indented with respect to the block fences, but not everywhere. Also, I don't think it's standard practice, it feels like it should start the the same level as the block, as I don't see much usefulness in having the whole code block indented by one level from the leftmost boundary.
It always felt to me like a fenced in block acts like a new scope, and as such deserves to be indentend. But that may indeed just be me.

I'll make another pass over all of it, remove indents from code block and pay attention to the verbal forms.

stdlib/builtin.ncl
vkleen: To be honest, I'm not sure. I think, I'm using it similarly to "convert into". FWIW the Wikipedia article on serialization agrees with me in its first sentence 😅

yannham:
Sorry if my suggestions sounded a bit too affirmative :sweat_smile: I'm not necessarily pushing for the trailing `s`, and maybe there's a case to be made for indenting code inside a fenced block? Let met quickly check if I can find style advises online. But the hard constraint is that we're consistent.

yannham:
So with respect to indenting code, it seems like [GitHub's documentation](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) and [W3C's documentation](https://www.w3schools.io/file/markdown-code-fence-blocks/) show examples where the code isn't indented indeed (starts at the same indentation as the fence).

yannham:
Oh my, I've merged #1198, which might make this rebase quite difficult. Might be simpler in the end to revert, merge this one, and reformat afterward? Or do you think it'll be manageable?

stdlib/function.ncl
```diff
@@ -2,74 +2,85 @@
...
-    ```nickel
-    id x == x
-    ```
-    for any value `x`.
-    "%
+        The identity function, that is
+        ```nickel
+        id x = x
+        ```
+        for any value `x`.
```
yannham, line 9: I think a new line before (but the after, I'm not sure) a fenced code block is advised by markdownlint. I wonder if we could even run markdownlint on the in-code documentation as part of the CI, via nickel-doc

```suggestion
        The identity function, that is
        
        ```nickel
        id x = x
        ```

        for any value `x`.
```

stdlib/function.ncl
```diff
@@ -2,74 +2,85 @@
...
     | doc m%"
-    Function composition, right to left. That is,
-    ```nickel
-    compose f g x == f (g x)
-    ```
-    "%
+        Function composition, from right to left. That is,
+        ```nickel
+        compose f g x == f (g x)
+        ```
```
yannham, line 30:
```suggestion
        Function composition, from right to left. That is,

        ```nickel
        compose f g x == f (g x)
        ```
```

stdlib/record.ncl
```diff
@@ -2,105 +2,121 @@
   record = {
     map : forall a b. (String -> a -> b) -> {_: a} -> {_: b}
     | doc m%"
-      Maps a function on every field of a record. The string argument of the function argument is the name of the
-      field.
+        Maps a function over every field of a record. The function is passed the
```
yannham, line 5: Well in this particular case the gain in clarity really shows :sweat_smile:

stdlib/string.ncl
```diff
@@ -156,209 +164,231 @@
...
+
+        **WARNING**: this function will currently return incorrect indices if
+        `str` contains Unicode extended grapheme clusters. The indices returned
+        will be the location of the matching codepoints, rather than the
+        matching grapheme clusters.
+
+        # Performance
+        
+        Note that this function may perform better by sharing its partial application between multiple calls,
+        because in this case the underlying regular expression will only be compiled once (See the documentation
```
yannham, line 377:
```suggestion
        because in this case the underlying regular expression will only be compiled once (see the documentation
```

yannham (approved):

PR #1198, opened 2023-03-24
Reformat the .ncl files of the repo with Topiary

Xophmeister:
This is the result of running Topiary against the Nickel standard library and the Mantis benchmark Nickel files. This PR is mostly to drive discussion, rather than to be merged...although, I guess it ultimately will be. Nonetheless, I've marked it as draft for now.

There are some niggles (see https://github.com/tweag/topiary/issues/389), but I believe everything in https://github.com/tweag/topiary/issues/331 has been addressed. Please have a look through and let me know if there's anything you have strong opinions about.

**Note** Multi-line strings are not affected by formatting, so their indentation may need to be manually changed to correct their alignment.

### DIFF
--- a/benches/mantis/deploy.ncl
+++ b/benches/mantis/deploy.ncl
@@ -1,83 +1,99 @@
 let lib = import "lib.ncl" in
-
 fun vars =>
-  let defaultLoggers = {
-    "io.netty" = "WARN",
-    "io.iohk.scalanet" = "INFO",
-    "io.iohk.ethereum.blockchain.sync.SyncController" = "INFO",
-    "io.iohk.ethereum.network.PeerActor" = "${LOGSLEVEL}",
-    "io.iohk.ethereum.network.rlpx.RLPxConnectionHandler" = "${LOGSLEVEL}",
-    "io.iohk.ethereum.vm.VM" = "OFF",
-    "org.jupnp.QueueingThreadPoolExecutor" = "WARN",
-    "org.jupnp.util.SpecificationViolationReporter" = "ERROR",
-    "org.jupnp.protocol.RetrieveRemoteDescriptors" = "ERROR",
-    "io.iohk.scalanet.discovery.ethereum.v4.DiscoveryService" = "WARN",
-  } in
-
-  let bootstrapNodes = {
-    "mantis-testnet" = [
-      "enode://f92aa66337ab1993cc7269d4295d296aefe6199b34e900eac08c514c947ec7340d46a5648ffc2da10325dbaba16bdf92aa9c0b5e51d97a7818c3f495d478ddad@mantis-testnet-0.mantis.ws:9001?discport=9501",
-      "enode://d8a010f019db37dcaf2e1fb98d4fcbf1f57dbd7e2a7f065e92fbe77dca8b9120d6e79f1617e98fa6134e6af8858ac8f3735b1e70a5708eb14f228080356eb0a7@mantis-testnet-1.mantis.ws:9002?discport=9502",
-      "enode://442e2bd50eece65f90dee0d5c6075da4e1b4bc62e36b261a52e7f393dae6a68241e4dbad868c7ecc14fed277ed72e99a289a811b6172f35fb18bdca0b7a5602c@mantis-testnet-2.mantis.ws:9003?discport=9503",
-      "enode://ff86741b7b35087b2b53f44a612b233336490d5fae10b1434619b7714fe2d5346c71427a5e126cd27b9422a4d4376c1534ef66e88c5e62d6441d2541f63de0cf@mantis-testnet-3.mantis.ws:9004?discport=9504",
-      "enode://af97643f364b805d5b0e32b5356578a16afcc4fb9d1b6622998e9441eeb7795e8daf8e6b0ff3330da9879034112be56954f9269164513ece0f7394b805be3633@mantis-testnet-4.mantis.ws:9005?discport=9505",
-    ],
-    "mantis-e2e" = [
-      "enode://c6ea9b9a96cdd6ab6b8b98bb1dc6db163d1d2006629c5cb2b6e9cb64e44b398741bfe4787963e75d6a01618af1f92330882a873d3c9198de04ba028b5abe4485@mantis-e2e-0.mantis.ws:35000?discport=35500",
-      "enode://c93b354a4d0782cc61f3e55132540ad8f83678ca1cd1d5fdc301eda60a770e793149e3166ba95065f0ae8cda054988a34906b3c906d4c713c373c001b90e40c1@mantis-e2e-1.mantis.ws:35001?discport=35501",
-      "enode://763eba5440b57f008316801cb4e5cc6428c3073ebc72563432b8210e0fb378fc84ae0f639d31f60b01c19e989a2103453eaee449fd469a511d8ed98e97f653d6@mantis-e2e-2.mantis.ws:35002?discport=35502",
-      "enode://8536966bc1e115d0628b7d582e4bc6b538b043135eb5297e6fe2eff2ea27ac4d31b23d0f7dfcc44cef4fef0a820c56d3cb3bd61a6267416c6541cbc78bcf88ac@mantis-e2e-3.mantis.ws:35003?discport=35503",
-      "enode://0088ef3a87c1883699fdd6cc36391338b3f162695b94112ffd2c71e1f2c63570dbb2f17a0e0c63d3123c56835af7b48dc2e5a22fc71396c53d59eb7772495a10@mantis-e2e-4.mantis.ws:35004?discport=35504",
-    ],
-    "mantis-staging" = [
-      "enode://39b925ba0beffdb80859a0ab34895c98bb61bd20d686ccd27f8c5a04dddc82b712081fd11bfd43f3bc08b00423a5ff8fee70b8a22dcc95e85537b2084dc6816a@mantis-staging-0.mantis.ws:33000?discport=33500",
-      "enode://cbd80c7f72a889101b7f23d51be2de7e3f1f46ad3b25c438e959e24e08f03bd9fe833460e84b60174d4eb120af3b127389c4606f81c842943c4922cab384a234@mantis-staging-1.mantis.ws:33001?discport=33501",
-      "enode://0e63642be49c5a092569aa01663fcda1505362cd0ac41e24ff9296ab80c97af135fb6fb247273631a3a11257774f39ed882d72a20fd45131e53e9015adf6b9e5@mantis-staging-2.mantis.ws:33002?discport=33502",
-      "enode://3ee3641a25cfc611ba54a898260af7768ecf0643f06aefedf853864ed433d5ad6265eeb24abcc4d6f6ee90a1eac6c1fbf157fc05fd8e28e194dfc864cb56058e@mantis-staging-3.mantis.ws:33003?discport=33503",
-      "enode://907842e336fc757bbfde70368aef329714aa627e72e5da687f31b097fa71a59f36404aebbc83885c9b515270042e025a6788b700c314ee8bc68099dcff32afcd@mantis-staging-4.mantis.ws:33004?discport=33504",
-    ],
-  } in
-
-  let params = {
-    namespace = vars.namespace,
-    datacenters | Array (lib.contracts.OneOf [
-        "eu-central-1",
-        "us-east-2",
-        "eu-west-2"])
-                | default = ["eu-central-1", "us-east-2"],
-    fqdn = "mantis.ws",
-    networkConfig | String
-                  | default = m%"
+  let defaultLoggers =
+    {
+      "io.netty" = "WARN",
+      "io.iohk.scalanet" = "INFO",
+      "io.iohk.ethereum.blockchain.sync.SyncController" = "INFO",
+      "io.iohk.ethereum.network.PeerActor" = "${LOGSLEVEL}",
+      "io.iohk.ethereum.network.rlpx.RLPxConnectionHandler" = "${LOGSLEVEL}",
+      "io.iohk.ethereum.vm.VM" = "OFF",
+      "org.jupnp.QueueingThreadPoolExecutor" = "WARN",
+      "org.jupnp.util.SpecificationViolationReporter" = "ERROR",
+      "org.jupnp.protocol.RetrieveRemoteDescriptors" = "ERROR",
+      "io.iohk.scalanet.discovery.ethereum.v4.DiscoveryService" = "WARN",
+    }
+  in
+  let bootstrapNodes =
+    {
+      "mantis-testnet" = [
+        "enode://f92aa66337ab1993cc7269d4295d296aefe6199b34e900eac08c514c947ec7340d46a5648ffc2da10325dbaba16bdf92aa9c0b5e51d97a7818c3f495d478ddad@mantis-testnet-0.mantis.ws:9001?discport=9501",
+        "enode://d8a010f019db37dcaf2e1fb98d4fcbf1f57dbd7e2a7f065e92fbe77dca8b9120d6e79f1617e98fa6134e6af8858ac8f3735b1e70a5708eb14f228080356eb0a7@mantis-testnet-1.mantis.ws:9002?discport=9502",
+        "enode://442e2bd50eece65f90dee0d5c6075da4e1b4bc62e36b261a52e7f393dae6a68241e4dbad868c7ecc14fed277ed72e99a289a811b6172f35fb18bdca0b7a5602c@mantis-testnet-2.mantis.ws:9003?discport=9503",
+        "enode://ff86741b7b35087b2b53f44a612b233336490d5fae10b1434619b7714fe2d5346c71427a5e126cd27b9422a4d4376c1534ef66e88c5e62d6441d2541f63de0cf@mantis-testnet-3.mantis.ws:9004?discport=9504",
+        "enode://af97643f364b805d5b0e32b5356578a16afcc4fb9d1b6622998e9441eeb7795e8daf8e6b0ff3330da9879034112be56954f9269164513ece0f7394b805be3633@mantis-testnet-4.mantis.ws:9005?discport=9505",
+      ],
+      "mantis-e2e" = [
+        "enode://c6ea9b9a96cdd6ab6b8b98bb1dc6db163d1d2006629c5cb2b6e9cb64e44b398741bfe4787963e75d6a01618af1f92330882a873d3c9198de04ba028b5abe4485@mantis-e2e-0.mantis.ws:35000?discport=35500",
+        "enode://c93b354a4d0782cc61f3e55132540ad8f83678ca1cd1d5fdc301eda60a770e793149e3166ba95065f0ae8cda054988a34906b3c906d4c713c373c001b90e40c1@mantis-e2e-1.mantis.ws:35001?discport=35501",
+        "enode://763eba5440b57f008316801cb4e5cc6428c3073ebc72563432b8210e0fb378fc84ae0f639d31f60b01c19e989a2103453eaee449fd469a511d8ed98e97f653d6@mantis-e2e-2.mantis.ws:35002?discport=35502",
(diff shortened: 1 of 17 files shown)


yannham:
Will take a deeper look, but at first sight, this looks pretty good!

benches/mantis/deploy.ncl
```diff
@@ -1,83 +1,95 @@
...
+                "us-east-2",
+                "eu-west-2"
+              ])
+              | default
+              = [ "eu-central-1", "us-east-2" ],
+            fqdn = "mantis.ws",
+            networkConfig
+              | String
+              | default
+              =m%"
```
vkleen, line 57: We'd probably want a space before `m%"` here

benches/mantis/jobs/mantis.ncl
```diff
@@ -1,260 +1,316 @@
 let types = import "../schemas/nomad/types.ncl" in
-let promtail = import "../tasks/promtail.ncl" in
-let Telegraf = import "../tasks/telegraf.ncl" in
-let lib = import "../lib.ncl" in
+  let promtail = import "../tasks/promtail.ncl" in
+    let Telegraf = import "../tasks/telegraf.ncl" in
+      let lib = import "../lib.ncl" in
```
vkleen, line 4: This might end up being a bit of a pain point. Nickel doesn't allow multiple bindings in a single let (yet), so many files will start with a number of `let` statements. While it makes sense for a let to increase the indentation elsewehere, maybe at top level that shouldn't be the case? I have no idea if that's possible with topiary at the moment.

benches/mantis/deploy.ncl
```diff
@@ -1,83 +1,95 @@
...
-  } in
-
-  namespaces."%{params.namespace}".jobs."%{params.job}"
+            logLevel | default = "INFO",
+            loggers = defaultLoggers,
+            job = vars.job,
+            role | [| `passive, `miner, `backup |] = vars.role,
+          }
+        in
+          let jobDefs =
```
yannham, line 68: As for the OCaml grammar (I suppose!), we don't want to indent one level further the left hand side of a let-binding. The common usage is to have a bunch of let-bindings as variable declarations before an expression. This would correspond e.g. in Rust to:

```rust
  let foo = 1;
  let bar = "foo";
  let baz = vec![];
  // content
```

stdlib/internals.ncl
```diff
@@ -41,54 +41,45 @@
...
+          let missing_fields = %fields% contracts_not_in_value in
+            if %length% missing_fields == 0 then
+              let tail_fields = field_diff value field_contracts in
+                let fields_with_contracts =
+                  array.fold_left (fun acc field =>
+                    if %has_field% field field_contracts then
+                      let contract = field_contracts. "%{field}" in
+                        let label = %go_field% field label in
+                          let val = value. "%{field}" in
+                            %record_insert% field acc (%assume% contract label val)
```
yannham, line 75: Here is a very nice diagonal that demonstrates that we don't want to indent the body of a let-binding :sweat_smile:

benches/mantis/jobs/mantis.ncl
yannham: I think OCaml and SML don't increase indentation with a let. It's really similar to a variable binding in other languages, and that doesn't increase indentation either. I vouch for never indenting further the body of a let (you have the same style of block of let-definitions at the level of functions as well, and in fact potentially pretty much anywhere, arbitrarily deep)

Xophmeister: Removed the indentation after `let...in...` in https://github.com/tweag/topiary/commit/a26a8708eb5524abbf36468594cfb85a66633429. I'll run Topiary against the Nickel files again once I've figured out [this problem](https://github.com/tweag/nickel/pull/1198#discussion_r1147862341).

benches/mantis/deploy.ncl
Xophmeister: This was surprising hard to fix! Done in https://github.com/tweag/topiary/commit/5463fbf2e4a4b1d222860c590a3391871d4310d5, with an accompanying issue in https://github.com/tweag/topiary/issues/395

Xophmeister:
13cba36 fixes:
* The `let` expression indentation
* The errant spacing before dynamic record field accessors
* The lack of spacing before interpolated strings that contain record field accessors

Compare:
* From master: e70f4ba...13cba36
* From last attempt: af0d5eec5f5d...13cba36

benches/mantis/jobs/mantis.ncl
```diff
@@ -121,47 +139,60 @@ fun params =>
...
+        }
+        & (if params.network != "etc" then
+          {
+            check_restart = {
+              limit = 5,
+              grace = "10m",
+            }
+          }
+        else
+          {}),
```
vkleen, line 182: Nitpick: Maybe we should (under some conditions?) make sure that `if` starts on a new line? This formatting isn't bad, but I wonder if
```
& (
  if params.network != "etc" then
    {
      check_restart = {
        limit = 5,
        grace = "10m",
      }
    }
  else
    {}),
```
might be better.

vkleen (commented):
This is looking really good, in my opinion 🎉 I think right now we're left with only picking the smallest of nits.

benches/mantis/jobs/mantis.ncl
Xophmeister: This one stood out to me, too; but more specifically at the closing paren, which looks weird to me. Incidentally, in this case, are the parens even necessary?

vkleen: Yes, the parens are necessary for parsing. Maybe the closing paren should also go on a new line? I'm really not sure...

Xophmeister: Rather than targetting the conditional expression, I've added a rule that will create a new indentation block between parens when the contents cover multiple lines.

I haven't applied it here, yet, but you can see the [effect on the stdlib here](https://github.com/tweag/topiary/commit/d7c038bbcabaa5de7add34bbddce489291dfe7f8#diff-6986669f19c1caad9af03e15df1e512864cc90f72071dee180866ea9dc1210e0). In the above example, the output looks like this:

```nickel
  # etc.
  service."%{params.namespace}-mantis-%{string.from_enum params.role}-rpc" = {
    check.rpc = {
      address_mode = `host,
      interval = "10s",
      port = "rpc",
      timeout = "3s",
      type = `http,
      path = "/healthcheck",
    }
    & (
      if params.network != "etc" then
        {
          check_restart = {
            limit = 5,
            grace = "10m",
          }
        }
      else
        {}
    ),
  }
  # etc.
```

How does that look, to you? :slightly_smiling_face:

Xophmeister: Yeah, I think it's an improvement too :tada:

I'll run it against this repo again...

benches/mantis/deploy.ncl
```diff
@@ -1,83 +1,95 @@
...
+    & {
+      role = `miner,
+      mantisRev = revisions.mantisRev
+    }
+  in
+  let namespaces =
+    {
+      "mantis-staging".jobs = {
+        miner = miner_,
+        passive | (passive & { count = 1 })
```
yannham, line 93: It's not that important, but I wonder if there are many language where lists and/or record have the additional spacing after `{`/`[` and before `}`/`]`? It looks a tad alien to me, especially for lists/arrays, but I can't say for sure. Is there any inspiration here?

benches/mantis/jobs/mantis.ncl
```diff
@@ -5,69 +5,83 @@ let lib = import "../lib.ncl" in
...
+    role | [| `passive, `miner, `backup |],
+    datacenters | Dyn,
+    job | Dyn,
+    namespace | String,
+    logLevel | String,
+    mantisRev | String,
+    fqdn | String,
+    loggers | { _ : String },
+    network
+      | lib.contracts.OneOf [ "testnet-internal-nomad", "etc" ]
```
yannham, line 23: For example (continuing with the spacing), why does this array get spacing around braces, but `baseTags` above doesn't?

benches/mantis/jobs/mantis.ncl
```diff
@@ -5,69 +5,83 @@ let lib = import "../lib.ncl" in
...
-               #       ]
-               # },
-} in
-
+    fastSync
+      | Bool
+      | default
+      = false,
+    reschedule | { .. }
+    #{_ : lib.contracts.PseudoOr [
```
yannham, line 33: Is it expected to re-indent comments like that? (I don't necessarily say it's a good or bad idea, but it's just surprising, so I'm genuinely wondering)

stdlib/enum.ncl
```diff
@@ -1,7 +1,7 @@
 {
-    enum = {
-      Tag
-        | doc m%"
+  enum = {
```
yannham, line 2: It seems sometimes there is a newline inserted before the opening `{`, and sometimes not (as here). Is there any rule around this  behavior?

stdlib/internals.ncl
```diff
@@ -15,80 +15,71 @@
...
-            %record_insert% field acc (left."%{field}"))
-        (%record_empty_with_tail% left)
-        (%fields% left)
+      let field_diff =
+        fun left right =>
+          array.fold_left (fun acc field =>
+            if %has_field% field right then
+              acc
+            else
+              %record_insert% field acc (left."%{field}")) (%record_empty_with_tail% left) (%fields% left)
```
yannham: Is there a character limit or any trigger for removing newlines like these and putting a mult-iline application like that back on one line?

Xophmeister:
* Merged changes from `master`
* Formatted multi-line parens with an indentation block internally

Compare:
* From `master`: 80b4a730ef64...843ac0e93a092
* From last attempt: 13cba36c4dd78...843ac0e93a092 (modulo merge cruft)

stdlib/string.ncl
```diff
@@ -15,19 +15,20 @@
...
-      if %typeof% s == `String then
-        if s == "true" || s == "True" then
-          "true"
-        else if s == "false" || s == "False" then
-          "false"
+      = fun l s =>
+        if %typeof% s == `String then
+          if s == "true" || s == "True" then
+            "true"
+          else
```
yannham, line 22: I think this is the last substantial pet-peeve of mine with the current formatting: it adds one level of nesting for a chain of `if..else..if..else..if...else`. Would it be doable in topiary to have a rule that if an else has an if-then-else as a children, we remove the first newline (between the else and the inner if) and remove one level of indentation, so that we get the more idiomatic:

```nickel
if ... then
  ...
else if .. then
 ...
else if .. then
 ...
else
  ...
```

yannham (commented):
Beside minor comment, the last obvious thing I see is the comment about chained `if-then-else`. But otherwise the rest looks great to me!

benches/mantis/jobs/mantis.ncl
Xophmeister: (See above comment about single-element lists)

benches/mantis/jobs/mantis.ncl
Xophmeister: Yes, comments will be affected by the indentation level. In the original, the source was being aligned to the annotation -- that seems to be common in the Mantis Nickel code -- but Topiary can't do that. As such, it treats it the same as any other code.

stdlib/enum.ncl
Xophmeister: It depends on the context. Usually a new line isn't inserted before a record, but there can be cases where a new line is inserted when the following node is multi-line (regardless of it being a record). This happens in `let...in...` expressions, for example:

```nickel
# Single-line
let x = { foo = 123 } in x

# Multi-line
let x =
  {
    foo = 123
  }
in x
```

The fact that the multi-line node is a record, in this case, is circumstantial. It _may_ be possible to add more nuance to these rules...

stdlib/internals.ncl
Xophmeister: No, I'm afraid there's no character counting (by design, IIRC). It's either multi-line, or not.

In the case of a conditional expression, it would be possible to at least split this distinction over the true and false branches, independently. So, for example, if the true branch was originally written on a single-line, then it would stay that way.

stdlib/string.ncl
Xophmeister: I'm going to say "maybe"! Adding exceptions seem to fall into two categories: really easy, or almost impossible :sweat_smile:

I'll see what I can do :muscle:

stdlib/internals.ncl
yannham: Sorry, I might have not been totally clear: I'm not talking about the if-then-felse (which is fine!), but about the whole enclosing application of `fold_left`:

I think the original code was:

```diff
-     let field_diff = fun left right => array.fold_left
-       (fun acc field =>
-        ...FUN DEF...)
-       (%record_empty_with_tail% left)
-       (%fields% left)
+     let field_diff =
+       fun left right =>
+        array.fold_left (
+           ...FUN DEF...
+        ) (%record_empty_with_tail% left) (%fields% left)
```

In particular, the initial application was split such that each argument was on its own line, and everything indented. In particular the application is thus arguably multi-line. But in the formatted version, it got back on one line, which is in my opinion a bit less readable for long application, but I also wonder if there is a limit to this behavior. If not, we're at risk of squashing multi-line applications on one very very long line.

stdlib/enum.ncl
yannham: But here, `enum` is multiline too, isn't it? In particular, I'm still wondering why

```
let enum = {

}
in

```

and

```
{
  enum = {
  ...
  },
}
```

would be treated differently, as both bind a name to a record, and both are multi-line. It seems, for some reason, that the first one get a newline before the `{`, but not the second one. Unless the record field isn't multiline, because what is taken into account is the declaration (left hand side of the `=`), and the content of the field (right hand side of `=`) ?

stdlib/internals.ncl
Xophmeister: Right, I see what you mean :+1: In the grammar, these are applicative chains. At the moment, the only formatting rule for them is to insert a space between the operands. Like with infix operator chains, I think it would be possible to do as you say, based on the multi-line-ness of the top-most applicative node in the tree.

Issue: https://github.com/tweag/topiary/issues/398

stdlib/string.ncl
Xophmeister: Issue: https://github.com/tweag/topiary/issues/399

Xophmeister: Fixed by https://github.com/tweag/topiary/pull/402

stdlib/enum.ncl
yannham: Sorry, I'm not sure to follow. I think I much prefer the former example, not the latter. I think the way let-bindings are formatted is totally fine, and is the right way (consider the multi-line property differently for both the RHS and the body of the let).

My issue is rather with record fields, I think. Why:

```
{
   enum = {
     foo,
     bar,
   },
}
```

isn't reformated to

```nickel
{
   enum = 
   {
     foo,
     bar,
   },
}
```

? It seems everywhere else, mult-iline records get their first opening braces put on a fresh line, but not for record fields. At first sight, it might seem that a record field is considered single line because its definition up to `=` is on single line (for example, I imagine that if it were `enum | Foo | Bar | Baz =`, we would keep the metadata on one line). But I'm just guessing there. Is that right?

Xophmeister: Hmm... This proposal is way too coarse. You end up with things on the same line that have no place being there :disappointed:

As a compromise, I have come up with [quite complicated] rules that add the line break and indentation block in every case _except_ for when followed by a record, list or parens. Now we have this:

```nickel
# single/normal single/normal
let x = f x in f x

# single/normal single/exceptional
let x = f x in { x = x }

# single/normal multi/normal
let x = f x in
f x

# single/normal multi/exceptional
let x = f x in {
  x = x
}

# single/exceptional single/normal
let x = { x = 1 } in f x

# single/exceptional single/exceptional
let x = { x = 1 } in { x = x }

# single/exceptional multi/normal
let x = { x = 1 } in
f x

# single/exceptional multi/exceptional
let x = { x = 1 } in {
  x = x
}

# multi/normal single/normal
let x =
  f x
in f x

# multi/normal single/exceptional
let x =
  f x
in { x = x }

# multi/normal multi/normal
let x =
  f x
in
f x

# multi/normal multi/exceptional
let x =
  f x
in {
  x = x
}

# multi/exceptional single/normal
let x = {
  x = 1
} in f x

# multi/exceptional single/exceptional
let x = {
  x = 1
} in { x = x }

# multi/exceptional multi/normal
let x = {
  x = 1
} in
f x

# multi/exceptional multi/exceptional
let x = {
  x = 1
} in {
  x = x
}
```

The rules aren't perfect yet, but I think this is closer to what you're looking for... Please let me know :pray:

Xophmeister: Sorry, @yannham; I didn't see your comment while I was writing mine.

I'm not sure this answers your question, but...

So, IIUC, your proposal is to normalise the record field value formatting, when the value is a multi-line record, rather than normalising the `let` expression when multi-line records are involved? (And by "records", I mean "records, lists, parens and enums".)

I think the reason why I didn't go in that direction is because:

1. I was adding an extra level of indentation, so you'd get:

   ```nickel
   {
      enum = 
        {
          foo,
          bar,
        },
   }
   ```

2. While the indentation isn't necessary, opinionatedly, it seems common to see the opening brace at the end of a line, rather than on a new line on its own.

yannham: I see, thanks. The indentation argument is indeed a good point.

Then what bothers me is the difference of treatment between record fields and LHS of a let-binding. I don't even think it's related to records, I guess the same thing happens with the opening bracket of an array, right? Record fields would be formatted like

```
{
  foo = [
    1,
    2
  ]
}
```

but let-bindings would be:

```
let foo =
  [
    1,
    2,
  ]
in
...
```

Here is my main constraint:

**if you take a let binding, erase the heading `let ` and the trailing `\nin`, that gives you a record field definition (well, you have to add enclosing `{`, and a trailing comma, but you get what I mean). I think the rest of this operation should be formatted the same as if it were a record field from the beginning.**

Otherwise, everything else is fine (including the `in` of let-bindings introducing a newline or not), but this property is not verified:

```nickel
let enum =
  {
    foo,
    bar
  }
in enum

# do the erase operation
enum =
  {
    foo,
    bar
  }

# but a record field would formatted as
enum = {
  foo,
  bar
}
```

You say we don't want to introduce an additional indent level for records, so we should keep the `enum = {` with the brace on the same line. Fair point! Then, would that be possible to that to the LHS of a let-binding to (not introducing an additional level even if it's multi-line), giving something similar to one of your suggestion, but not totally equivalent (the difference is that in the following, I didn't change the `in ...` introducing a newline before the RHS or not):

```
{
  single-single = let x = [ 1, 2 ] in x,

  single-multi = let x = [ 1, 2 ] in fun a =>
    x,

  multi-single = let x = [
    1,
    2
  ]
  in x,

  multi-multi = let x = [
    1,
    2
  ]
  in fun a =>
    x
}
```

But, in those examples, I would expect to the `in` to be on a line alone though. I don't know if it's related. So something like:

```
{
  single-single = let x = [ 1, 2 ] in x,

  single-multi = let x = [ 1, 2 ] in fun a =>
    x,

  multi-single = let x = [
    1,
    2
  ]
  in
  x,

  multi-multi = let x = [
    1,
    2
  ]
  in
  fun a =>
     x
}
```

Technically, I imagine it would amount to say that the only property affected by a let-binding being multi-line is to put the `in` on its own single line?

yannham: In fact, I'm not sure anymore. There are also other situations when not putting the LHS of the let-binding makes it less nice. I think the current approach is already fine, and the path forward isn't totally clearcut. I will open an issue to discuss this on the Topiary repo directly, and gather more input.

In the meantime the current version is totally reasonable, so let's not block on this. Thanks for bearing with me and explaining this through!

Xophmeister:
Thanks, @yannham :slightly_smiling_face: Apologies for the delayed response (OOO, today): I've rebased and rerun Topiary with the `else..if` and applicative chain formatting. I'll open the PR up for review :ok_hand: 

(I nuked the previous commits because it was easier, with the rebase.)

yannham (approved):

yannham:
LGMT, but there's a last rebasing to do @Xophmeister , sorry for that. In the meantime we'll try to not merge anything with Nickel code in it.

yannham:
Oh, but he's OOO. I'm taking care of this PR, then.

Issue #1187, opened 2023-03-16
Merge is not idempotent


yannham:
**Describe the bug**

The merge operation is advertised as being idempotent, but it is currently not. In particular, non-empty arrays are not mergeable, even if they are the same. It's probably fine not to be idempotent on functions (or is it?), but it's a real problem on data, as detailed through actual examples below.

**To Reproduce**

Here is a typical example of non idempotency:

```nickel
nickel>let path = ["/bin/bash", "/bin/sh"] in
path & path
error: non mergeable terms
  ┌─ repl-input-0:1:12
  │
1 │ let path = ["/bin/bash", "/bin/sh"] in
  │            ^^^^^^^^^^^^^^^^^^^^^^^^
  │            │
  │            cannot merge this expression
  │            with this expression
2 │ path & path
  │ ----------- merged here
  │
  = Both values have the same merge priority but they can't be combined
```

It actually happens in real life, in particular with default values. If a contract sets a default value which is a non-empty array, and the contract happened to be applied several times, then we get the following. For example, in [Nixel](https://github.com/nickel-lang/nickel-nix):

- The base builder `BashShell` defines an `env` field, that automatically compute the environment from the `structured_env` field
- Most higher-level builders like `RustShell` or `CShell` "derives" from this base builder, that is `RustShell = BashShell & {...}`.
- Now, if you want to define a development shell with both a Rust and a C environment, you get an error, because Nickel doesn't succeed in merging the two auto-generated env, while they're exactly the same. The issue is that writing `MyShell = {...} & RustShell & CShell` evaluates to `MyShell = {...} & BashShell & { ...def of Rust shell.. } & BashShell & { ... def of CShell ... }`, leading to two occurrences of `BashShell`.

So, "contract inheritance" can lead to the same contract being applied several time. Different parts of the code (a framework consuming a configuration versus the configuration itself) may also have to ensure that some contract is applied, and can't rely on the fact that another part will do it.

**Expected behavior**

Merging pure data (excluding functions, at least as a first step) should be idempotent.

**Possible solutions**

Until now, we've refrained from merging arrays, because it might not correspond to what the user has in mind (typically, concatenation makes more sense that pointwise merging in a lot of concrete applications). But this is becoming a problem. Here are possible solutions:

- "artificially" makes merging idempotent, by testing arrays for equality. If they are equals, return the first operand, otherwise fail. This leads to no unexpected merges: if two arrays are different, merge will fail.
[...]

PR #1199, opened 2023-03-24
Implement array merging, part I

yannham:
In order to make merge idempotent, we've decided to implement a very restricted form of merging on arrays as well (see #1187). This PR makes a first step toward closing #1187, implementing point-wise merging for arrays (and requiring their length to be the same).

In a second time, we plan to be even more restrictive, and refuse to merge array elements that aren't the equal (in particular, records should have exactly the same set of fields). The logic and error reporting for this second step are delayed to a follow-up PR.

### DIFF
--- a/src/eval/merge.rs
+++ b/src/eval/merge.rs
@@ -56,6 +56,7 @@ use crate::error::{EvalError, IllegalPolymorphicTailAction};
 use crate::label::Label;
 use crate::position::TermPos;
 use crate::term::{
+    array,
     record::{self, Field, FieldDeps, FieldMetadata, RecordAttrs, RecordData},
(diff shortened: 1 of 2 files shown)


vkleen:
I was wondering, could we, instead of walking the array, generate a contract application `%assume% label (contract.Equals array1) array2` and defer the equality check this way?

yannham:
Ah, I think I god ahead of myself a bit. `contract.Equal` would probably be just implemented as `fun base => contract.from_predicate ((==) base)`? In that case the contract is strict, which we don't want to do. Otherwise, we have to make `contract.Equal` lazy. Which is maybe something to aim for.

yannham:
Superseded by #1229.

PR #1200, opened 2023-03-25
Update stdlib string functions to work on Unicode extended grapheme clusters

matthew-healy:
This PR updates the standard library `string` module to function predominantly on Unicode [extended grapheme clusters](https://unicode.org/reports/tr29/).

In particular:
- `string.split`, `string.characters`, `string.replace`, `string.substring` and `string.contains` all avoid ever breaking up extended grapheme clusters,
- `string.codepoint` and `string.from_codepoint` have been removed,
- a lot of string functionality has been pulled out of `eval/operation.rs` and moved into a new `term/string.rs` module.

## What's left?

This PR doesn't touch any of the functions which take a regex argument. I'll make a follow-up PR when I get some time to work on it.

## Review tips

I'd recommend reviewing c72f490 and e197dab individually, but then skipping to 60132c0, as the commits in-between mostly reimplement the string operations in-place, whereas that commit refactors everything into the `string` module. It's likely simpler to just review the (heavily-commented) `string` module than each of the intermediate states.

Please let me know if you have any questions!


src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+            ),
+            NonIntEnd(end) => write!(
+                f,
+                "expected 3rd argument (end) to be a positive integer smaller than {}, got {end}",
+                usize::MAX
+            ),
+            StartOutOfBounds { start, str_len } => write!(
+                f,
+                "index out of bounds. 
+                 Expected 2nd argument (start) to be between 0 and {str_len}, got {start}"
```
yannham, line 255: I fear indentation is included in the string, which would give strange result on the terminal (unfortunately Rust doesn't have Nickel's multiline strings :p). If you want to split things a bit, maybe `{}\n{}` followed by your two lines? Or your can also use a trailing `\` as well. This will strip all the indentation, so it's not as smart as Nickel, but as long you only use one indentation level for the whole string, this should be fine:

```suggestion
                "index out of bounds.\n\
                 Expected 2nd argument (start) to be between 0 and {str_len}, got {start}"
```

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+        last_boundary: usize,
+    }
+
+    impl<'a> Iterator for SearchIter<'a> {
+        type Item = SearchEvent<'a>;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            let unwrap_explanation = "
+                None of the GraphemeIncomplete errors are possible here:
+                    - PreContext and PrevChunk only happen if chunk_start is nonzero.
```
yannham, line 328: same thing here, I don't think this will give the indentation you expect. However with indented text inside indented text you're out of luck: either you don't use multiline string but you need to indent everything on the left since the very first column, or...I don't now :man_shrugging: use a macro like `indent_doc!` provided by the eponymous crate, I believe

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+}
+
+/// Types and functions designed to make it as easy as possible not to accidentally
+/// break up Unicode extended grapheme clusters in string operations.
+mod grapheme_cluster_preservation {
+    use unicode_segmentation::GraphemeCursor;
+
+    /// Returns a [`SearchIter`], which is an `Iterator` for iterating through
+    /// `haystack` in order to find `needle`.
```
yannham, line 276:
```suggestion
    /// `src` in order to find `needle`.
```

yannham (approved):

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+        if separator.is_empty() {
+            // If there's no separator, we split the string between
+            // each grapheme cluster.
+            self.grapheme_clusters()
+        } else {
+            use grapheme_cluster_preservation::SearchEvent;
+
+            let mut result = Vec::new();
+
+            // We bulid our result vec by searching through the string for the separator.
```
vkleen, line 108:
```suggestion
            // We build our result vec by searching through the string for the separator.
```

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+
+                        if self.haystack[self.last_boundary..].starts_with(self.needle)
+                            && !does_match_intersect_cluster()
+                        {
+                            // Yay, we found a match! We start by grabbing the slice of
+                            // the string we traversed while looking for the match.
+                            let since_last_match =
+                                &self.haystack[self.current_split_start..self.last_boundary];
+                            // Now we move the cursor to after the match, to avoid needlessly
+                            // traversing that part of the string again.
```
vkleen, line 364: If you're looking for, say, "aa" in "aaaa", this will produce only two matches, one at offset 0 and the other at offset 2, won't it?

On reflection, this doesn't matter in our case, since all our string primitives either look for the first match or look to do a search-and-replace, where you want this skipping behaviour explicitly. But maybe it's worth explicitly documenting that?

vkleen (approved):

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+use serde::{Deserialize, Serialize};
+use unicode_segmentation::UnicodeSegmentation;
+
+use super::{array::Array, Term};
+
+/// A Nickel string is really just a Rust `String`, overlayed with some
+/// methods implementing custom logic (in particular, functions which
+/// avoid ever breaking up Unicode extended grapheme clusters.)
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct NclString(String);
```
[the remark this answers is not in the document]
matthew-healy: Yeah, I went back-and-forth on this, but more than happy to change it.

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// This method has `O(self.len())` time complexity.
+    pub fn length(&self) -> usize {
+        self.graphemes(true).count()
+    }
+
+    /// Returns an [`Array`] of Nickel strings, each one
+    /// containing a single Unicode extended grapheme cluster.
+    ///
+    /// This method has `O(self.len())` time complexity.
+    pub fn characters(&self) -> Array {
```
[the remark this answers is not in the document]
matthew-healy: My justification was that outside of the `string` module - and specifically within `operation.rs` - we don't necessarily want to have to think about what an extended grapheme cluster is. However, inside `string.rs` we likely do want to know that we're breaking a string into grapheme clusters, e.g. in the `separator.is_empty()` clause of `split`. But I could also be overthinking it a little. :shrug: 

I guess I could at least mark `grapheme_clusters` as `#[inline(always)]`.

src/term/string.rs
matthew-healy: I actually meant to put these on the same line but just split up the text so it's easier to read. I just forgot to escape the newline. :sweat_smile:

src/term/string.rs
```diff
@@ -0,0 +1,399 @@
...
+        };
+
+        let Ok(end_usize) = usize::try_from(end) else {
+            return Err(SubstringError::NonIntEnd(end.clone()));
+        };
+
+        // We need to know the length of the grapheme cluster iterator,
+        // and we have to walk the graphemes to do that, so we might
+        // as well build a Vec to make slicing out the substring easier.
+        let graphemes: Vec<_> = self.graphemes(true).collect();
```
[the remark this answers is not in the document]
matthew-healy: I would never have thought to do this! Great suggestion.

Two small caveats to this though. One is that we've replaced the allocation with an extra iteration of the length of the substring at the end, because we need to walk its grapheme clusters to figure out whether it was actually the length we wanted. The other is that we need to walk the entire length of the string in order to produce error messages. 

I think optimising for the happy path, and for shorter substring lengths, makes sense though, so I've implemented your suggestion.

PR #1203, opened 2023-03-27
Add Equal contract to the stdlib

### LINKED ISSUE #1187: Merge is not idempotent

yannham:
Following #935, and https://github.com/tweag/nickel/pull/1199#issuecomment-1484696929 (meaning this PR is required before tackling #1187), this PR introduces an `Equal` contract to the stdlib, which simply checks that the tested value equals to some constant provided as a parameter to the contract. The `Equal` contract should preserve laziness over arrays and records.

As per https://github.com/tweag/nickel/pull/1199#issuecomment-1484696929 and the following comment, this is probably required to fix #1187. This PR doesn't deal with records lazily yet, as it might require a new primop to properly preserve recursiveness (if we use record.map, then we're going to kill recursive references, which is not expected from contracts). This is planned as an immediate follow-up, but this chunk is already reviewable and functional.

### DIFF
--- a/stdlib/internals.ncl
+++ b/stdlib/internals.ncl
@@ -146,4 +146,11 @@
 
   "$rec_force" = fun value => %rec_force% (%force% value),
   "$rec_default" = fun value => %rec_default% (%force% value),
+
+  # Provide an access to contract.Equal within the initial environement. Merging
+  # makes use of `contract.Equal`, but it can't blindly substitute such an
+  # expression, because `contract` might have been redefined locally. Putting it
+  # in an internal value prefixed with `$` makes it accessible from the initial
+  # environment and prevent it from being shadowed.
+  "$stdlib_contract_equal" = contract.Equal,
 }
--- a/stdlib/internals.ncl
+++ b/stdlib/internals.ncl
@@ -147,10 +147,10 @@
   "$rec_force" = fun value => %rec_force% (%force% value),
   "$rec_default" = fun value => %rec_default% (%force% value),
 
-  # Provide an access to contract.Equal within the initial environement. Merging
+  # Provide access to contract.Equal within the initial environement. Merging
   # makes use of `contract.Equal`, but it can't blindly substitute such an
   # expression, because `contract` might have been redefined locally. Putting it
   # in an internal value prefixed with `$` makes it accessible from the initial
-  # environment and prevent it from being shadowed.
+  # environment and prevents it from being shadowed.
   "$stdlib_contract_equal" = contract.Equal,
 }
--- a/stdlib/record.ncl
+++ b/stdlib/record.ncl
@@ -199,7 +199,7 @@
     length
       : forall a. {_: a} -> Number
       | doc m%"
-          Returns the number of fields in a record. This cound doesn't include
+          Returns the number of fields in a record. This count doesn't include
           fields both marked as optional and without a definition.
 
           Because of the need to filter empty optional fields, the cost of
--- a/stdlib/record.ncl
+++ b/stdlib/record.ncl
@@ -195,5 +195,19 @@
       |> to_array
       |> array.filter (fun { field, value } => f field value)
       |> from_array,
+
+    length
+      : forall a. {_: a} -> Number
+      | doc m%"
+          Returns the number of fields in a record. This cound doesn't include
+          fields both marked as optional and without a definition.
+
+          Because of the need to filter empty optional fields, the cost of
+          `length` is linear in the size of the record.
+        "%
+      = fun record =>
+        record
+        |> fields
+        |> array.length,
   }
 }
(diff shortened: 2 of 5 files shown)


stdlib/internals.ncl
```diff
@@ -146,4 +146,11 @@
 
   "$rec_force" = fun value => %rec_force% (%force% value),
   "$rec_default" = fun value => %rec_default% (%force% value),
+
+  # Provide an access to contract.Equal within the initial environement. Merging
+  # makes use of `contract.Equal`, but it can't blindly substitute such an
+  # expression, because `contract` might have been redefined locally. Putting it
+  # in an internal value prefixed with `$` makes it accessible from the initial
+  # environment and prevent it from being shadowed.
```
vkleen, line 154:
```suggestion
  # Provide access to contract.Equal within the initial environement. Merging
  # makes use of `contract.Equal`, but it can't blindly substitute such an
  # expression, because `contract` might have been redefined locally. Putting it
  # in an internal value prefixed with `$` makes it accessible from the initial
  # environment and prevents it from being shadowed.
```

stdlib/record.ncl
```diff
@@ -195,5 +195,19 @@
       |> to_array
       |> array.filter (fun { field, value } => f field value)
       |> from_array,
+
+    length
+      : forall a. {_: a} -> Number
+      | doc m%"
+          Returns the number of fields in a record. This cound doesn't include
```
vkleen, line 202:
```suggestion
          Returns the number of fields in a record. This count doesn't include
```

vkleen (approved):
👍 I just found some typos.

PR #1209, opened 2023-03-29
Add a JSON documentation export option

vkleen:
This is to facilitate better standard library documentation formatting on the website.


src/program.rs
```diff
@@ -401,70 +405,178 @@ impl From<ColorOpt> for ColorChoice {
...
+                    })
+                    .collect();
+                Some(ExtractedDocumentation { fields })
+            }
+            _ => None,
+        }
+    }
+
+    /// Create a markdown or json file with documentation for the specified FileId.
     pub fn output_doc(
```
yannham, line 525: I wonder if we can reorganize those free-standing functions as proper, smaller and composable methods, now that you've introduced a bunch of types. Something like `ExtractedDocumentation::extract_from_term(t: RichTerm) -> Self`, then `ExtractedDocumentation::extract_from_file(...) -> Self`, and two functions `to_markdown` and `to_json` (doing so we can also probably make all those functions independent of a free-standing `format` parameter).

src/program.rs
```diff
@@ -401,70 +405,178 @@ impl From<ColorOpt> for ColorChoice {
...
-
-                    if let Some(ref doc) = field.metadata.doc {
-                        document.append(parse_documentation(header_level, arena, doc, options));
-                    }
-
-                    if let Some(ref value) = field.value {
-                        to_markdown(value, header_level + 1, arena, document, options)?;
-                    }
-                }
+        let mut entries: Vec<(_, _)> = docs.fields.iter().collect();
```
yannham, line 569: It's a matter of taste (so just a comment, not an actual review suggestion), but I found the original version with newlines to be more readable.

src/bin/nickel.rs
```diff
@@ -81,11 +81,15 @@ enum Command {
...
     Doc {
-        /// Specify the path of the generated documentation file. Default to
+        /// The path of the generated documentation file. Default to
         /// `~/.nickel/doc/<input-file>.md` for input `<input-file>.ncl`, or to
         /// `~/.nickel/doc/out.md` if the input is read from stdin.
         #[structopt(short = "o", long, parse(from_os_str))]
         output: Option<PathBuf>,
+        /// The output format for the generated documentation. Defaults
+        /// to rendered markdown. Possible values: markdown, json
+        #[structopt(long)]
```
yannham, line 91: Setting a default value with structopt will print it automatically (see the help on any subcommand, and the documentation of --color for an example).

```suggestion
        /// The output format for the generated documentation. Possible values: markdown, json
        #[structopt(long, default_value = "markdown")]
```

flake.nix
```diff
@@ -404,7 +411,8 @@
         dockerImage = buildDocker packages.nickel; # TODO: docker image should be a passthru
         inherit vscodeExtension;
         inherit userManual;
-        inherit stdlibDoc;
+        stdlibDoc = stdlibDoc' "markdown";
```
yannham, line 414: What about renaming `stdlibDoc'` to just `stdlibDoc`, and this output to `stdlibMarkdown`? It just feels a bit more uniform. But just a suggestion.

yannham (approved):

flake.nix
vkleen: That was just for keeping the packages that the flake exposes consistent. If that's not important here, I'd like `stdlibMarkdown` better, too.

src/program.rs
vkleen: I've tried to do that refactoring. It does seem to make the code a bit more composable 👍

flake.nix
yannham: I personally think we can do with such a breakage for now (I don't expect anyone to actually use this output outside of us, but I really don't know).

src/program.rs
```diff
@@ -401,77 +405,209 @@ impl From<ColorOpt> for ColorChoice {
...
     use comrak::nodes::{Ast, AstNode, NodeCode, NodeHeading, NodeValue};
     use comrak::{format_commonmark, parse_document, Arena, ComrakOptions};
+    use serde::{Deserialize, Serialize};
+    use std::collections::HashMap;
+    use std::fmt;
     use std::io::Write;
 
-    /// Create a markdown file with documentation for the specified FileId.
+    /// Create a markdown or json file with documentation for the specified FileId.
     pub fn output_doc(
```
yannham, line 426: any reason not to make it be an `output()` method of `ExtractedDoc`, rather than a freestanding function? Also, it's a matter a taste, but I have a slight preference toward `output_json` and `output_markdown` as an interface (or even `write_json` or `write_markdown`, as it is customary with serde modules), rather than one function parametrized by (something isomorphic to) a boolean.

vkleen: The `to_json` and `to_markdown` should probably really be called `write_json` and `write_markdown`. I think those would be the `output_xxx` functions you're thinking of. The `output_doc` function might actually make more sense as a function on `DocFormat`, since it takes care of extracting the documentation from a file.

src/program.rs
```diff
@@ -486,107 +505,109 @@ mod doc {
...
-                                documentation: field.metadata.doc.clone(),
-                            },
-                        )
-                    })
-                    .collect();
-                Some(ExtractedDocumentation { fields })
+    impl ExtractedDocumentation {
+        /// Extract documentation from a [`RichTerm`], descending into records.
+        /// Returns `None` if the top-level term is not a record.
+        fn try_from_term(rt: &RichTerm) -> Option<Self> {
```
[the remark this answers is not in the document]
vkleen: Maybe it makes sense to get rid of `try_from_file`, implement `TryFrom<&RichTerm>` and shift the file extraction logic to what is now `output_doc`? `try_from_file` feels a bit weird to me, right now; and it doesn't fit with the `TryFrom` idea.

src/program.rs
yannham: To me, `DocFormat` is really just an basic enum, so that would be weird to implement the logic on this one. My initial idea was rather to have a function `ExtractedDocumentation::extract_from_file` or whatever you call it, that would take the same arguments (excepted format), and produce an `ExtractedDocumentation` (because that's what you do in the middle, right). Then, separate functions to write this documentation in whatever format you want to a stream, which would be `write_json` and `write_markdown` (or, if you really fancy it, one `write` function taking a `Format` argument). IMHO it makes more sense that a function traversing `ExtractedDocumentation` to write to a stream is attached to `ExtractedDocumentation`. Here it seems to be an artifact from the fact that `output_doc` is actually the composition of morally smaller functions.

yannham: Agreed for `to_json` vs `write_json`, I was quoting the name from serde, but `to_xxx` generally produces a `String`, and `write_xxx` write to a stream.

yannham: I agree with the second option as well. This is also more in line with how we already implement the `export` subcommand, I think.

> Separately, I feel like renaming SerializationError to ExportError and adding an enum case for MissingDocumentation to it. At the moment, Nickel just doesn't write anything to the specified file if it can't find any documentation to extract.

Sounds sensible, yes :+1:

src/program.rs
```diff
@@ -304,10 +304,20 @@ impl<EC: EvalCache> Program<EC> {
         String::from_utf8(buffer.into_inner().into_inner()).unwrap()
     }
 
-    /// Create a markdown file with documentation for the specified program in `.nickel/doc/program_main_file_name.md`
+    /// Extract documentation from the program
     #[cfg(feature = "doc")]
-    pub fn output_doc(&mut self, out: &mut dyn std::io::Write) -> Result<(), Error> {
-        doc::output_doc(self.vm.import_resolver_mut(), self.main_id, out)
+    pub fn extract_docs(&mut self) -> Result<doc::ExtractedDocumentation, Error> {
```
yannham, line 309:
```suggestion
    pub fn extract_doc(&mut self) -> Result<doc::ExtractedDocumentation, Error> {
```

Nitpick, but I think we use the singular form otherwise as in `ExtractedDocumentation`.

tests/snapshot/main.rs
```diff
@@ -58,6 +58,40 @@ fn check_error_snapshots(file: &str) {
...
+fn check_doc_stderr_snapshots(file: &str) {
+    let file = TestFile::from_project_path(file);
+
+    let snapshot = NickelInvocation::new()
+        .subcommand("doc")
+        .file(&file)
+        .args(["--stdout"])
+        .snapshot_stderr();
+
+    insta::with_settings!({filters => vec![
```
yannham, line 84: Don't we do something like this already elsewhere in the test framework? Just to know if we could factor this out (e.g. in `utilities`)

yannham (approved):
I think it looks really good like this :+1:

PR #1210, opened 2023-03-29
Format type and contract information in documentation output

vkleen:
Type and contract information is rendered as a list of the form
```
# ident
- `ident : forall a. a`
- `ident | Contract`
```
before the documentation for `ident`.


yannham (approved):
LGMT. Wonder if we should add the stdlib documentation to the snapshot tests.

Issue #1007, opened 2022-12-16
stdlib string functions should work on extended grapheme clusters


matthew-healy:
Once #1005 is merged, `string.length` will count the number of extended grapheme clusters in a (UTF-8) string. This means that characters comprised of multiple code points, but still representing a "single character" are counted as one character for the purpose of this function.

However, other standard library string functions (or more accurately, the string primops) still operate on code points, e.g. `string.substring`, `string.split`, etc. 

In general, this issue proposes that we should consider the extended grapheme cluster to be the smallest unit of Nickel strings, and all string manipulation functionality should be updated accordingly.

matthew-healy:
Noting here that we'll remove `string.codepoint` and `string.from_codepoint` as part of this work as:

1. they break the extended grapheme cluster abstraction, and
2. there are no similar functions for strings in Nix
3. we don't have any concrete use cases for them.

matthew-healy:
I just merged #1200 which partly solves this. The remaining work is around the behaviour of regex search functions, which currently *do* break up grapheme clusters.

yannham:
@matthew-healy I don't know if we will have time to finish that before 1.0. Anyway, I think the most important is already here: Unicode-aware regexps are hopefully more of a edge use-case. In the meantime, we probably want to update the documentation of regex-related functions with a big warning saying that those aren't (yet) unicode aware.

matthew-healy:
Yes, very fair. I'll try to spend some time updating the docs later today, as I'm away from tomorrow until Wednesday.

yannham:
Removing from 1.0 milestone because we agreed that #1200 + #1212 are enough for 1.0, while not strictly closing this issue either.

PR #1212, opened 2023-03-29
Add grapheme cluster warnings to stdlib.string regex function docs

matthew-healy:
See the discussion in #1007 for context.

Feel free to merge if/when this is approved (or make any required changes), as I'll be away from my laptop for the next week.

### DIFF
--- a/stdlib/string.ncl
+++ b/stdlib/string.ncl
@@ -270,7 +270,7 @@
 
     replace_regex: String -> String -> String -> String
     | doc m%"
-      `replace_regex regex repl String` replaces every match of `regex` in `String` with `repl`.
+      `replace_regex regex repl str` replaces every match of `regex` in `str` with `repl`.
 
       For example:
       ```nickel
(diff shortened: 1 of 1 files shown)


yannham (approved):
Suggestions:

- put warning in bold (`**WARNING**`)
- add something on the fact that this will be solved in a coming version of Nickel

stdlib/string.ncl
```diff
@@ -334,6 +342,11 @@
       Note that this function may perform better by sharing its partial application between multiple calls,
       because in this case the underlying regular expression will only be compiled once (See the documentation
       of `string.is_match` for more details).
+
+      WARNING: this function will currently return incorrect indices if `str`
```
yannham, line 346: Ah, this one was also missing. Adding it as part of the reformatting

vkleen: Yeah, I noticed that one when rebasing #1196. It's fixed there as well.

PR #1214, opened 2023-03-30
Show diagnostics when import terms content don't typecheck correctly

ebresafegaga:
Successor to #1191

### DIFF
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -4,6 +4,8 @@ use codespan::FileId;
 use nickel_lang::{
     cache::{Cache, CacheError, CacheOp, CachedTerm, EntryState},
     error::{Error, ImportError},
+    position::TermPos,
+    term::{RichTerm, Term, Traverse, TraverseOrder},
     typecheck::{self, linearization::Linearization},
 };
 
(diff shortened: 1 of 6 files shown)


yannham:
As mentioned in https://github.com/tweag/nickel/pull/1191#discussion_r1151715820, I don't think re-traversing the whole AST in full just to find import is a viable approach. I would rather have this information stored when we traverse the AST the first time when linearizing, if possible.

lsp/nls/src/cache.rs
```diff
@@ -47,12 +47,35 @@ impl CacheExt for Cache {
         let mut import_errors = Vec::new();
         if let Ok(CacheOp::Done((ids, errors))) = self.resolve_imports(file_id) {
             import_errors = errors;
+            let mut errors = Vec::new();
             for id in ids {
-                self.typecheck_with_analysis(id, initial_ctxt, initial_env, lin_cache)
-                    .unwrap();
+                // Ignore the results, and check for errors after resolution
```
yannham, line 52:
```suggestion
```

lsp/nls/src/cache.rs
```diff
@@ -47,12 +47,35 @@ impl CacheExt for Cache {
...
             }
+            let message = "This import could not be resolved because its content have either failed to parse or typecheck correctly.";
+            let errors = errors
+                .into_iter()
+                .map(|(name, pos)| ImportError::IOError(name, String::from(message), pos));
+
+            import_errors.extend(errors);
         }
 
+        // Two things
```
yannham, line 75: which code does this comment refer to?

yannham (approved):

lsp/nls/src/cache.rs
ebresafegaga: Ah, this is some noise from the draft implementation.  PTAL

lsp/nls/src/cache.rs
```diff
@@ -72,11 +84,26 @@ impl CacheExt for Cache {
             lin_cache.insert(file_id, linearized);
             Ok(CacheOp::Done(()))
         } else {
-            panic!()
+            unreachable!()
```
yannham, line 87: I suspect this wasn't related to your PR, but while we're at it, could we add an explicit justification of why this is unreachable? I've had to debug unreachable recently and the fact that there wasn't any justification although it was'not trivial was a hinder. I imagine it's because if we could get the term from `self.terms`, then it must have been parsed before, and thus the invariant `*state > EntryState::Typechecked` must hold.

src/cache.rs
```diff
@@ -918,6 +918,11 @@ impl Cache {
         Some(file)
     }
 
+    pub fn get_imports(&self, file: &FileId) -> Option<Vec<FileId>> {
+        let s = self.imports.get(file)?;
```
yannham, line 922: Nitpick: what does `s` mean? A longer name might more explicit :slightly_smiling_face:

yannham (approved):

src/cache.rs
ebresafegaga: Forgot about that while cleaning up. Thanks 🙂

lsp/nls/src/cache.rs
ebresafegaga: Sure 👍. I think it is because `EntryState::Parsed` is the first item of the enum, and we have the case `if *state >= EntryState::Parsed ` before the `else` branch which contains the `uncreachable!()`, so it it actually unreachable unless the order of the enum is changed 😅

PR #1218, opened 2023-04-03
Append a trailing newline to JSON exports

vkleen:
This appends a trailing newline to the output of `nickel export` if and only if the export format is JSON. The other export formats include a trailing newline by default.

I ran into this at https://github.com/tweag/tf-ncl/pull/35. Since we're exporting a formatted JSON instead of a compacted representation anyway, this seems like a sensible thing to do.


yannham (approved):
Wondered if this was possible to do by an option in `serde_json`, but it's deliberately not possible (https://github.com/serde-rs/json/issues/716).

PR #1223, opened 2023-04-04
Fix the type of `array.elem` and the mantis benchmark

yannham:
When trying to run the benchmarks on #1220, they were broken.

Beside updating to the latest stdlib changes, I had to fix the type of `array.elem` which lies, because it's polymorphic but equality isn't actually parametric. This means that sadly, currently some statically typed code can actually fails its contract, because the equality is typed with `_a -> _a -> Bool`. Maybe it shouldn't. In the meantime, let fix `elem`, which currently fails on any reasonable usage with a spurious blame error.

### DIFF
--- a/benches/arrays/fold.ncl
+++ b/benches/arrays/fold.ncl
@@ -1,6 +1,7 @@
-let letter | Number -> string.CharLiteral = fun n =>
-  string.code "a" + (n % 26)
-  |> string.from_code in
+let letter | Number -> string.Character = fun n =>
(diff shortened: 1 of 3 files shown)


vkleen (approved):
That's pretty unfortunate, but this seems like a good interim fix. It might be useful to document the failing example in a comment in the stdlib, so people like me don't go around changing everything to `forall`'s in a few months 😅

yannham:
Addressed in [f07ee8](https://github.com/tweag/nickel/pull/1223/commits/af07ee802014d2f93ade909297816a4f96a500b5).

PR #1227, opened 2023-04-04
Export a parser for `Types` and refactor a bit `parser`

yannham:
This PR has been done as part of fixing #1187, and split as a stand-alone change. For some reason I needed to parse types. This PR exposes the rule for parsing type in the LALRPOP grammar.

I also noticed that the code already present had duplication, and that I was going to add one more copy of the `parse_term_tolerant` and `parse_term` methods. Those methods are now implemented once and for all for any parser generated by Lalrpop, and have been renamed to be more consistent with the rest of the codebase.


src/parser/mod.rs
vkleen, line 1: This feels like it's coupling us quite tightly to larlpop implementation details. I'm not if that's a problem but it's worth pondering.

yannham: I wouldn't say it does, because `LalrpopParser` ist in fact mostly a facade for Lalrpop-generated parsers. But everything else in the Nickel codebase should use the `ErrorTolerantParser` trait, which is mostly independent from Lalrpop (that's not entirely true, because of error type and so forth, but I think it's hardly avoidable). Maybe I should make this trait private? I wonder if it's possible to depend on private trait for a public trait. Let me try.

yannham: I think [1c31f53](https://github.com/tweag/nickel/pull/1227/commits/1c31f53cb4afddfe9e83f5fa88c4d05051a1cb37) decouples the trait implements a bit, in a cleaner way, although it's a tad more verbose (we had to repeat the interface of `parse_tolerant` and `parse_strict`).

vkleen (approved):

Issue #1226, opened 2023-04-04
Undefined TermPos when processing typed block


vkleen:
The following code yields a `TermPos::unwrap` panic at `src/label.rs:248:35`.
```nickel
{
  Contract | {},
  Config = {
    x | Contract = {},
    y = lib.f x,
  },
  lib = {
    f | {_: Dyn} -> {_: Dyn} = fun x => x,
  }
} :_
```

vkleen:
But it happens only sometimes, when run with `nickel -f repro.ncl`. Presumably, that's the randomized evaluation order caused by differen `HashMap` seeds.

yannham:
Are you using an up-to-date master branch? I don't see an unwrap at line `248`, but I see one at line `245`, which I'm hopefully fixing right now as part of the merging arrays PR. At least some of them.

yannham:
By the way, did you mean `Contract = {}` instead of `Contract | {}` ?

yannham:
The issue is that some static type errors, most notably `ArrowTypeMismatch`, have been using the same infrastructure as contract's type paths to report nice underlining of subtype in a more complex type. When the type was pretty-printed and the offset computed, everything was fine. But then we added position to types and said oh we want to show the subtype _in situ_, where the contract has been defined.

This works well for contract, but the static types are (most of them) generated by type inference, so they don't have a position, and everything is blowing up. My current fix will be to make the function fallible, and if we found that the subtype has no position, we pretty-print the type in a temporary source (a bit like what we do for some terms, the infamous `<unknown> (generated by evaluation)`, but for types), and do everything one more time. Now the type and all of its subtype have a position, and hence should be fine

vkleen:
Oh sorry about that, it should be the one at 245. I made some edits in that file while trying to trace this. In practice, I would mean `Contract = {}`; but the panic happens either way.

vkleen:
This was fixed incidentally in #1229.

PR #1229, opened 2023-04-05
Implement array merging, take II

### LINKED ISSUE #1187: Merge is not idempotent

yannham:
Closes #1187 #1226.

Implement array merging by generating an `contract.Equal` application, ie two arrays are properly merged only if they are equal.

This required to slightly rework the way contract reporting work, see https://github.com/tweag/nickel/issues/1226#issuecomment-1496068308. This PR doesn't panic if a type doesn't have a position anymore when reporting an error, but pretty-print this type, parse it again with positions, and use this type for error reporting.

Doing so, this PR also fixes an unreported bug, which was that reporting contract failure on deeply (more than one level) type constructor was off. On master:

```console
$ nickel <<< '[[5]] | Array (Array String)'
  ┌─ <stdin>:1:3
  │
1 │ [[5]] | Array (Array String)
  │   ^            ^^^^^^^^^^^^ expected array element type
  │   │             
  │   applied to this expression

```

This underlines `Array String` instead of `String`. This manifested through dubious code in the `ty_path::span` method. This issue is fixed by this PR and a snapshot test was added.

## Error reporting when merging fails

The current error reporting for a failure of merging two arrays isn't optimal. We have to choose between reporting the generated contract (`contract.Equal some_array`), which is indeed the contract that failed, but is not related (position-wise) to the original merge.

We can instead report the original merge location as being the position of the type (which is a bit hacky, but whatever), but then the error reporting is underlining the merge with the message "expected type". For now we went with the latter. All the information required for good error reporting is propagated, but doing so requires to rethink a bit the structure of contract error reporting. For now, we use this imperfect reporting, and plan on improving on it later. It's arguably at least viable.


src/error.rs
```diff
@@ -1176,17 +1175,41 @@ mod blame_error {
         }
     }
 
+    /// Calls [`crate::label::ty_path::span`], but if the call returns `None` (the position of the
+    /// subtype isn't defined), [path_span] pretty-print the type inside a new source, parse it,
```
vkleen, line 1179:
```suggestion
    /// Calls [`crate::label::ty_path::span`], but if the call returns `None` (the position of the
    /// subtype isn't defined), [path_span] pretty-prints the type inside a new source, parses it,
```

vkleen (approved):

src/error.rs
```diff
@@ -1349,11 +1364,10 @@ is None but last_arrow_elem is Some"),
...
             .map(|diag| diag.notes)
             .unwrap_or_default();
-        let (path_label, notes_higher_order) = report_ty_path(&label);
+        let (path_label, notes_higher_order) = report_ty_path(files, &label);
 
-        // Report contract error with the surrounding source program
-        // for more context.
-        let path_label = contract_error_with_context(&label, &path_label.message, path_label.range);
+        // Report contract error with the surrounding source program for more context.
+        // let path_label = contract_error_with_context(&label, &path_label.message, path_label.range);
```
yannham, line 1370:
```suggestion
```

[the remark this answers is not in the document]
yannham: Right, this can go, thanks.

PR #1231, opened 2023-04-06
Put the stdlib under the std namespace

### LINKED ISSUE #1131: Namespacing stdlib modules

Since the beginning of Nickel, the stdlib has grown: in particular, quite a few modules were added, like `function`, `record`, `array` or `contract`. They are currently available directly in the environment of each program at the start. This has a number of disadvantages:

1. It's not discoverable. There's no root module/record/value on which we can ask Nickel (through `nickel query` for example) what are the current stdlib modules. We get completion when we start to type them within an LSP-enabled editor, but we need to know - at least partially - what we are looking for first.
2. It's clashing with local names. For example, `array`, `record` and `contract` aren't **un**common to use when writing generic library functions. We already had to use `array_` and `record'` instead. Unwilling shadowing a stdlib module by a local variable happens silently and can be sneaky.

One solution is to put the whole stdlib in a specific namespace. In practice, it just means putting the modules inside a top-level `std` or `stdlib` record. `array` would become `std.array`, for example. `nickel query <<< 'std'` or typing `std` in a code editor with the Nickel LSP would provide the list of available modules.

- Do nothing, putting up with the inconveniences listed above
- Have a proper notion of namespacing, that doesn't rely on records. Modules, for example. However, this solution is complex and hasn't obvious advantages over using a simple record, as proposed in this issue.

yannham:
Closes #1131. 

This PR introduces a top-level value `std` in which lives the stdlib modules, to make them more discoverable, help completion, and avoid polluting the initial environment.

## Design

There was a surprising number of small but non trivial dilemma to solve. The first is where this overarching `std` module should lie:

- It sounds appealing to simply generate a record programmatically from Nickel, such that we don't even have to change anything to the current stdlib code. However, this unfortunately doesn't play well with the LSP, which can't handle terms without positions in a file. This would break completion and type information without a serious rethink of how the LSP currently works.
- If we need a source (can be generated), then we can write a `std.ncl` file which simply imports the modules:

  ```nickel
  {
    std = {
      array = import "array.ncl",
      string = import "string.ncl",
      ..
    }
  }
   ```
  The issue is that imports don't work for the stdlib, because it's included at compile-time in the binary.
- Then we can generate this file by doing compile-time concatenation/include_str/substitution. We still have a source code to provide to the LSP (albeit being generated), but we don't have to use imports. Now, the issue is that say a contract failure on `std.array.at null null` will point to code that exists nowhere (has been generated and concatenated from existing fragments), and the user has no way to find this code anywhere, and thus can't even know what function is failing, depending on the formatting. Hence, we want the `std.ncl` file to be accessible
- In the end, I went with squashing the modules together. Later on we might add special imports like `import <std/array.ncl>` which would allow to split the stdlib into separate files again, but this requires a bit of thinking. In the meantime, this decision put slightly more burden on us maintainers, but don't impact users (LSP is working, the file defining the stdlib is accessible as text).

Typechecking mutually recursive modules with polymorphic functions also proved challenging: thus, the enclosing `std = { }` is omitted (as opposed to the previous style) and inter-modules reference all goes through the `std` interface present in the initial typing environment, instead of being directly mutually recursive.

## Reviewing

The important commits to review are:

- Introduce the std namespace ([62e7b09](https://github.com/tweag/nickel/pull/1231/commits/62e7b09762f998d3cecd688f440f56f40ca7391b))
- Update the LSP to work with the new std namespace ([59d02a6](https://github.com/tweag/nickel/pull/1231/commits/59d02a63fcf927f40c64804b93b3944ca7eedf12))
- Update RELEASES.md (https://github.com/tweag/nickel/pull/1231/commits/0a3c1e68fb3047c9eb5d40412db08df0ef0ce85a)

The rest is mostly a painful search and replace to update all the existing Nickel code. We've tested the benchmarks as well, to make sure they are up to date. The benchmarks showed a noticeable regression on some examples. Showing the flamegraphs didn't really help, as although the proportion varied a bit, nothing really stood out. Running a bigger example for longer doesn't seem to make that much of a difference, so it might be that loading the stdlib or accessing it has a penalty which shows on micro-benchmarks.

## TODO

- [x] Fix panic in the LSP, because we're adding usages for `std`, but `std` doesn't get a `Declaration` or `RecordField` item. Should we just add an usage field to any node?
- [x] Documentation: **will be done in a subsequent PR**
- [x] Have a tentative fix for the LSP not completing on `std`


RELEASES.md
```diff
@@ -15,6 +15,16 @@ Breaking changes
   renamed to `is_number`, `is_string` and `to_string`
 - The `string.to_num` and `string.from_num` functions have been renamed to
   `to_number` and `from_number`
+- All the stdlib module, `array`, `string`, `record`, etc. have no been put
```
vkleen, line 18:
```suggestion
- All the stdlib module, `array`, `string`, `record`, etc. have now been put
```

src/typecheck/mod.rs
```diff
@@ -1932,18 +1949,45 @@ pub fn apparent_type(
...
                 RecordRowsF::Empty,
                 |r, (id, field)| {
-                    let uty = UnifType::from_apparent_type(
-                        field_apparent_type(field, None, None),
-                        term_env,
-                    );
+                    let uty = match field_apparent_type(field, None, None) {
+                        ApparentType::Annotated(ty) => UnifType::from_type(ty, term_env),
+                        ApparentType::FromEnv(uty) => uty,
+                        // If we haven't reached max_depth yet, and the time is only approximated,
```
vkleen, line 1975:
```suggestion
                        // If we haven't reached max_depth yet, and the type is only approximated,
```

stdlib/std.ncl
```diff
@@ -0,0 +1,2549 @@
...
+# a possible avenue.
+#
+# In the meantime, we choose to have this one big monolithic `std` module. This
+# choice doesn't negatively impact Nickel users (the file can be browsed looking
+# for error locations, typechecking and the LSP are happy because no code is
+# progammatically generated). It's just harder to maintain.
+#
+# Typing
+#
+# Stdlib module shouldn't recursively refer to other modules directly, they
```
vkleen, line 25:
```suggestion
# Stdlib modules shouldn't recursively refer to other modules directly, they
```

stdlib/std.ncl
```diff
@@ -0,0 +1,2549 @@
...
+# should use the stdlib provided in the environment instead. This means using
+# `std.array` from within the `record` module instead of just `array`. While the
+# latter could work in principle, typechecking mutually recursive records is
+# just harder, and the chances are this won't typecheck.
+#
+# On the other hand, the `std` module in the environment is properly typed, as
+# the interface is elaborated from individual notations.
```
vkleen, line 32:
```suggestion
# the interface is elaborated from individual annotations.
```

vkleen (approved):
Looks good! Just a few typos in comments 👍

yannham:
Ah, completion for the stdlib is broken. I think I understood why, will probably fix in a coming PR, but I prefer to have a solution lined up before merging this one.

yannham:
With the help of @ebresafegaga, we could restore the stdlib completion with a one-liner in [6186118](https://github.com/tweag/nickel/pull/1231/commits/6186118f953e1514f47866195cda4f3ec9b9c051).

PR #1232, opened 2023-04-06
Don't touch `not_exported` fields during export

### LINKED ISSUE #1230: Not_exported fields get evaluated during export anyways
# Situation
A `nickel export` evaluates a program fully before continuing on the `export` codepath. Consequently, when exporting a toplevel record, all record fields will be evaluated, even if they will never appear in the result. On the one hand, this can be a performance problem. On the other hand, it can also be quite puzzling when building a configuration piecemeal using merging. Consider:
```nickel
let Module = {
  configs
    | not_exported
    | Array {..},
  module
    | not_exported
    | {
      config | {..},
      output | {..},
    },
  output = configs |> array.map (fun config' => (module & { config = config' }).output),
} in
{
  configs = [
    { os = "a" },
    { os = "b" },
  ],
  module = {
    config,
    output = { on = config.os },
  },
} & Module
```
Before starting the export, Nickel will evaluate this configuration fully and rightfully complain that the `config` field in `module` doesn't have a definition. But a user may expect (and we *did* expect during a recent Nickel hour) that with `nickel export` the `module` field never gets evaluated standalone.

# Possible approaches

- We could not do anything, except documenting that behaviour. The current semantics are reasonably clear.
- We could introduce a variant of `UnaryOp::Force` which doesn't descend into fields marked as `not_exported`.

vkleen:
Add a `for_export` boolean parameter to `UnaryOp::Force` which isn't exposed to the user. When forcing the evaluation of a term for export, this parameter is set and the evaluator will ignore record fields marked as `not_exported`.

Incidentally, refactor `map_values_closurize` in `src/eval/operation.rs`. The new version operates on iterators with item type `(Ident, Field)` instead of `RecordData`. It expects the caller to apply any necessary filtering, in particular for optional fields. For example, what used to be
```rust
record.map_values_closurize(...)
```
should now correspond to
```rust
let fields = record
  .fields
  .into_iter()
  .filter(|(_, f)| !f.is_empty_optional())
  .map_values_closurize(...);
RecordData { fields, ..record }
```
The extra verbosity is worth it in my opinion, in exchange for being able to customize the filter step.

### A note for reviewers
This PR ended up having some conflicts with the recent change from `HashMap` to `IndexMap`. Please watch out for places where I might have overlooked a stray `HashMap`.

Closes #1230


src/eval/operation.rs
```diff
@@ -1119,17 +1124,33 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+
+                            let terms = fields
+                                .values()
+                                .cloned()
+                                .map(|field| {
+                                    field
+                                        .value
+                                        .expect("fields without definition would have thrown MissingFieldDefError before")
+                                })
+                                .collect::<Vec<_>>();
```
vkleen, line 1145: This `collect` should actually be superfluous.
```suggestion
                                });
```

src/eval/operation.rs
```diff
@@ -1119,17 +1124,33 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
-                            let terms = record.fields.clone().into_values().map(|field| field.value.unwrap());
-                            let cont = RichTerm::new(Term::Record(record), pos.into_inherited());
+                            let cont = RichTerm::new(
+                                Term::Record(RecordData { fields, ..record }),
+                                pos.into_inherited(),
+                            );
 
                             Ok(Closure {
-                                body: seq_terms(terms, pos_op, cont),
+                                body: seq_terms(terms.into_iter(), pos_op, cont),
```
vkleen, line 1153:
```suggestion
                                body: seq_terms(terms, pos_op, cont),
```

src/term/record.rs
```diff
@@ -367,13 +367,15 @@ impl RecordData {
         &self,
     ) -> impl Iterator<Item = Result<(&Ident, &RichTerm), MissingFieldDefError>> {
         self.fields.iter().filter_map(|(id, field)| {
-            debug_assert!(field.pending_contracts.is_empty());
+            // debug_assert!(field.pending_contracts.is_empty());
```
vkleen, line 370: Oops...
```suggestion
            debug_assert!(field.pending_contracts.is_empty());
```

src/term/mod.rs
```diff
@@ -1081,7 +1081,17 @@ pub enum UnaryOp {
...
+    /// # About `for_export`
+    ///
+    /// When exporting a Nickel term, we first apply `Force` to the term to
+    /// evaluate it. If there are record fields that have been marked `not_exported`,
+    /// they would still be evaluated ordinarily, see [#1230](https://github.com/tweag/nickel/issues/1230).
+    /// To stop this from happening, we introduce the `for_export` parameter
+    /// here. When `for_export` is `true`, the evaluation of `Force` will skip
+    /// fields that are marked as `not_exported`. When `for_export` is `false`,
+    /// these fields are evaluated.
+    Force { for_export: bool },
```
yannham, line 1094: Nitpick: although the documentation is nice and explains it well, I tend to think that the naming could reflect the _behavior_ of the two different variant for `Force`, instead of the motivation. That is, `for_export` doesn't tell much to a newcomer on the codebase (and the meaning could even evolve). On the other hand, something like `ignore_not_exported` is more specific, IMHO.

src/eval/operation.rs
```diff
@@ -1119,14 +1124,29 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                            let terms = fields
+                                .clone()
+                                .into_values()
+                                .map(|field| {
+                                    field
+                                        .value
+                                        .expect("fields without definition would have thrown MissingFieldDefError before")
+                                });
 
-                            // unwrap: the call to map_fields_without_optionals must ensure that
```
yannham, line 1126: Nitpick: I think we lose a bit of information here by removing the comment. The expect message doesn't talk about `map_fields_without_optionals` at all, for example. Although I guess it's now called `map_values_closurize` ? Anyway, I would suggest to either keep both (comment + message), or to extend the message of `expect` to provide at least as much context as before.

yannham (approved):

src/eval/operation.rs
vkleen: I've changed the `expect` message to mention that `map_values_closurize` enforces the invariant. The comment had indeed gone stale, `map_fields_without_optionals` got refactored to `map_values_closurize` at some point.

yannham: Sounds good

PR #1236, opened 2023-04-12
[CI] Avoid segfaults on MacOS

yannham:
Various flake-centric Nix commands seem to randomly segfault when evaluating the Nickel flake on the CI, in some situations. It seems to be GC-related (after discussing with more Nix people internally at Tweag), and disabling garbage collection should hopefully get rid of those errors in the meantime.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Run all checks
       run: |
-        nix flake check --print-build-logs
+        export GC_DONT_GC=1; nix flake check --print-build-logs
 
     - name: Typecheck benchmarks
       run: find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nix run . -- -f file typecheck


thufschmitt:
Nix 2.15 (released yesterday) is supposed to fix that issue. So hopefully this won't be needed

yannham:
@thufschmitt so we just have to wait for the update of `cachix/install-nix-action`? Or does it even just download the latest Nix without having to update the action?

vkleen:
As far as I can tell `cachix/install-nix-action` essentially uses `https://nixos.org/nix/install`. So I think it should pick up the latest released nix version?

thufschmitt:
Yeah, it's automatically pulling in the latest Nix release (you can see it [here](https://github.com/tweag/nickel/actions/runs/4681032612/jobs/8293115581#step:3:128) for instance). So it should have fixed itself

yannham:
Great news, then :slightly_smiling_face: closing this for the time being. Let's see if this happens again.

PR #1237, opened 2023-04-12
Update documentation to reflect RFC002 and RFC005

### LINKED ISSUE #1171: Update documentation on metadata post RFC005
RFC005 (#819) changed the semantics of metadata. In particular, there's now a difference between a free-standing contract application (like `5 | Num`) and a field metadata (`{foo | Num}`). We should update the documentation accordingly.

yannham:
Closes #1171.

Update the syntax and the contracts section of the user manual to reflect the changes made to the syntax by RFC002 (mixing type and term syntax) and RFC005 (restricting general metavalues to appear only at the level of fields as well as the change in semantics).

### DIFF
--- a/doc/manual/contracts.md
+++ b/doc/manual/contracts.md
@@ -351,7 +351,7 @@ nickel>{foo = "a", bar = 1} | Contract
 #### Giving values for fields
 
 While most record contracts don't have field definitions, they can. In fact,
-records contracts are nothing special: any record literal can be used both as a
+record contracts are nothing special: any record literal can be used both as a
 contract or a normal value. If a field is defined both in the contract and the
 checked value, the two definitions will be merged in the final result. For
 example:
--- a/doc/manual/contracts.md
+++ b/doc/manual/contracts.md
@@ -26,6 +26,11 @@ let x | Number = 1 + 1 in x
 {x | Number = 1 + 1}
 ```
 
+**Important**: contracts attached to a record field, such as `{x | Number = 1}`,
+and a free standing contract annotation such as `{x = (1 | Number)}` behave
+*differently* with respect to record merging. However, as far as runtime checks
+are concerned (which is the focus of this document), they are equivalent.
+
 Here, `x` is bound to a `Number` contract. When evaluating `x`, the following steps
 are performed:
 
@@ -346,7 +351,7 @@ nickel>{foo = "a", bar = 1} | Contract
 #### Giving values for fields
 
 While most record contracts don't have field definitions, they can. In fact,
-records contracts are not special: any record literal can be used both as a
+records contracts are nothing special: any record literal can be used both as a
 contract or a normal value. If a field is defined both in the contract and the
 checked value, the two definitions will be merged in the final result. For
 example:
@@ -385,9 +390,9 @@ At first sight, `=` also fits the bill. However, there are a number of subtle
 but potentially surprising differences.
 
 One concerns open contracts. Merging never requires the presence of specific
-fields: thus, the contract `{bar | String}` attached to `foo` will actually behave
-as an open contract, even if you didn't use `..`. This is usually not what you
-want:
+fields: thus, the contract `{bar | String}` attached to `foo` will actually
+behave as an open contract, even if you didn't use `..`. This might not be what
+you want:
 
 ```text
 nickel>let ContractPipe = {
@@ -538,11 +543,11 @@ needed. For example, record fields are not evaluated until they are accessed,
 printed (in the REPL or as a result of `nickel`), or until the whole record is
 serialized/exported. Thanks to laziness, you can for example query specific
 information on a large configuration without having to actually evaluate
-everything. We will use a dummy failing expression `1 + "a"` to observe where
-evaluation takes place:
+everything. We will use a failing expression `builtin.fail_with "ooch!"` to observe
+where evaluation takes place:
 
 ```text
-nickel>let config = {fail = 1 + "a", data | doc "Some information" = 42}
+nickel>let config = {fail = builtin.fail_with "ooch", data | doc "Some information" = 42}
 nickel>:query config.data
 * value: 42
 * documentation: Some information
@@ -560,8 +565,8 @@ corresponding to its schema. If this contract checked everything eagerly,
 forcing the evaluation of the most part of the configuration, we would lose the
 benefits of laziness. Thus, *we want contracts to be lazy as well*. In
 particular, a subcontract attached to a field should only fire when the field is
-requested. This is the case with record contracts, and in general all native
-contracts combinators are lazy too.
+requested. This is the case with record contracts, and in general with all native
+contracts combinators.
 
 ### Writing lazy contracts
 
--- a/doc/manual/contracts.md
+++ b/doc/manual/contracts.md
@@ -27,8 +27,8 @@ let x | Number = 1 + 1 in x
 ```
 
 **Important**: contracts attached to a record field, such as `{x | Number = 1}`,
-and a free standing contract annotation such as `{x = (1 | Number)}` behave
-*differently* with respect to record merging. However, as far as runtime checks
+and a freestanding contract annotation such as `{x = (1 | Number)}` *behave
+differently with respect to record merging*. However, as far as runtime checks
 are concerned (which is the focus of this document), they are equivalent.
 
 Here, `x` is bound to a `Number` contract. When evaluating `x`, the following steps
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -574,11 +574,11 @@ including being dropped in favor of another value, the final value for `foo` has
 to respect the contract as well or the evaluation will fail accordingly.
 
 This is only true for *contracts attached directly to record fields* (either
-directly, or coming from an enclosing record contract). In particular, `{foo |
-Num = 1} & {foo | force = "bar"}` will fail, but `{foo = (1 | Num)} & {foo |
-force = "bar"}` will succeed. In the latter case, the contract is not considered
-to be field metadata, but a local contract check, which is not propagated by
-merging.
+directly, or coming from an enclosing record contract). In particular,
+`{foo | Number = 1} & {foo | force = "bar"}` will fail, but
+`{foo = (1 | Number)} & {foo | force = "bar"}` will succeed. In the latter case,
+the contract is not considered to be field metadata, but a local contract check,
+which is not propagated by merging.
 
 #### Specification
 
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -20,9 +20,9 @@ this means that order doesn't matter, and `left & right` is the same thing as
 default values for example, the idea is to use metadata to do so (annotations),
 rather than relying on the left or right position.
 
-**Warning**: At the time of writing, Nickel's version is 0.3.1. Custom merge
-functions are planned for a future version. They are not detailed here yet. For
-more details, see the associated technical document [RFC001][rfc001].
+**Warning**: At the time of writing, Nickel's version is 1.0. Custom merge
+functions are planned for a future version. They are not detailed here yet. See
+the associated technical document [RFC001][rfc001].
 
 The section describes the behavior and use-cases of merge, by considering the
 following situations:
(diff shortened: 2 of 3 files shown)


doc/manual/contracts.md
```diff
@@ -346,7 +351,7 @@ nickel>{foo = "a", bar = 1} | Contract
 #### Giving values for fields
 
 While most record contracts don't have field definitions, they can. In fact,
-records contracts are not special: any record literal can be used both as a
+records contracts are nothing special: any record literal can be used both as a
```
vkleen, line 354:
```suggestion
record contracts are nothing special: any record literal can be used both as a
```

doc/manual/contracts.md
```diff
@@ -538,11 +543,11 @@ needed. For example, record fields are not evaluated until they are accessed,
 printed (in the REPL or as a result of `nickel`), or until the whole record is
 serialized/exported. Thanks to laziness, you can for example query specific
 information on a large configuration without having to actually evaluate
-everything. We will use a dummy failing expression `1 + "a"` to observe where
-evaluation takes place:
+everything. We will use a failing expression `builtin.fail_with "ooch!"` to observe
```
vkleen, line 546:
```suggestion
everything. We will use a failing expression `std.fail_with "ooch!"` to observe
```

doc/manual/contracts.md
```diff
@@ -538,11 +543,11 @@ needed. For example, record fields are not evaluated until they are accessed,
...
+where evaluation takes place:
 
 ```text
-nickel>let config = {fail = 1 + "a", data | doc "Some information" = 42}
+nickel>let config = {fail = builtin.fail_with "ooch", data | doc "Some information" = 42}
```
vkleen, line 550:
```suggestion
nickel>let config = {fail = std.fail_with "ooch", data | doc "Some information" = 42}
```

doc/manual/merging.md
```diff
@@ -163,6 +162,8 @@ unless one of the following condition hold:
 
 - They are both of a primitive data type `Number`, `Bool`, `Enum`, `String` and
   they are equal
+- They are both arrays and they are equal (checked by generating an application
+  of the lazy contract `contract.Equal`)
```
vkleen, line 166:
```suggestion
  of the lazy contract `std.contract.Equal`)
```

doc/manual/merging.md
```diff
@@ -256,9 +261,17 @@ final record:
 ## Merging records with metadata
 
 Metadata can be attached to values thanks to the `|` operator. Metadata
-currently includes contract annotations, default value, merge priority, and
+currently include contract annotation, default value, merge priority, and
```
vkleen, line 264: I actually find the previous formulation "Metadata currently includes contract annotations" more natural 😕 I don't have a good rationale, though.

doc/manual/merging.md
```diff
@@ -256,9 +261,17 @@ final record:
...
 documentation. We describe in this section how metadata interacts with merging.
 
+Note that metadata can only be syntactically attached to record fields, at the
```
vkleen, line 267:
```suggestion
Note that metadata can only be syntactically attached to record fields, with the
```

doc/manual/merging.md
```diff
@@ -256,9 +261,17 @@ final record:
...
+exception of type and contract annotations, which can appear anywhere in a
+freestanding expression such as `(x | Num) + 1`. However, a contract annotation
+outside of a record field isn't considered a metadata (it's a mere contract
```
vkleen, line 270:
```suggestion
outside of a record field isn't considered metadata (it's a mere contract
```

doc/manual/merging.md
```diff
@@ -453,13 +466,12 @@ priority between the recursive priority and the existing one.
...
-We can consider the merging system to feature priorities. To each field
-definition `foo = val` is associated a priority `p(val)`. When merging two
-common fields `value_left` and `value_right`, then the results is either the one
-with the highest priority (that overrides the other), or the two are tentatively
-recursively merged, if the priorities are the same. Without loss of generality,
-we consider the simple case of two records with only one field, which is the
-same on both side:
+To each field definition `foo = val` is associated a priority `p(val)`. When
+merging two common fields `value_left` and `value_right`, the results is either
+the one with the highest priority (that overrides the other), or the two are
```
vkleen, line 471:
```suggestion
Each field definition `foo = val` is assigned a priority `p(val)`. When
merging two common fields `value_left` and `value_right`, the result is either
the one with the highest priority (that overrides the other), or the two are
```

doc/manual/merging.md
```diff
@@ -561,10 +573,17 @@ to a field `foo`, merging ensures that whatever is this field merged with,
 including being dropped in favor of another value, the final value for `foo` has
 to respect the contract as well or the evaluation will fail accordingly.
 
+This is only true for *contracts attached directly to record fields* (either
+directly, on coming from an enclosing record contract). In particular, `{foo |
```
vkleen, line 577:
```suggestion
directly, or coming from an enclosing record contract). In particular, `{foo |
```

doc/manual/merging.md
```diff
@@ -561,10 +573,17 @@ to a field `foo`, merging ensures that whatever is this field merged with,
...
+Num = 1} & {foo | force = "bar"}` will fail, but `{foo = (1 | Num)} & {foo |
+force = "bar"}` will succeed. In the later case, the contract is not considered
+to be a field metadata, but a local contract check, which is not propagated by
+merging.
```
vkleen, line 581:
```suggestion
force = "bar"}` will succeed. In the latter case, the contract is not considered
to be field metadata, but a local contract check, which is not propagated by
merging.
```

doc/manual/merging.md
```diff
@@ -589,6 +608,64 @@ Leftn, Right1, .., Rightk`. Here, we ignore the case of type annotations such as
...
+{ foo | FooContract}
+& { foo.required_field1 = "here" }
+& { foo.required_field2 = "here" }
+```
+
+Running the above program succeeds, although the intermediate value
+`{foo.required_field1 = "here"}` doesn't respect `FooContract` (it misses the
+field `required_field2`).
+
+If we try to observe the intermediate result (`deep_seq` recrusively forces the
```
vkleen, line 630:
```suggestion
If we try to observe the intermediate result (`deep_seq` recursively forces the
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+### Types
+
+The Nickel syntax mixes both terms and types in the same namespace. The
+following program is perfectly legal: `let value = Number -> (fun value label =>
+value) in ((fun x => x + 1) : value)`. See the
+[RFC002](https://github.com/tweag/nickel/blob/d723a3721c6b0fe9c4b856e889bb7211d6136665/rfcs/002-merge-types-terms-syntax.md)
+for a detailed account of this design.
+
+The documentation still makes a distinction between *types* and other
+expressions, the former being constructs which are handled specifically by the
```
vkleen, line 660:
```suggestion
expressions, the former being constructs which are handled specially by the
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+typechecker and are listed below. However, any expression can be considered a
+type (in the generic case, it's considered as an opaque type), and types
```
vkleen, line 662:
```suggestion
type (in the generic case, it's considered as an opaque type), and type
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+Nickel features the following builtin types and type constructors:
+
+- Primitive types: `Number`, `String`, `Bool`, and `Dyn` (the dynamic type, wich
+represents any value)
+- Array: `Array <type>` is an array whose elements are of type `<type>`.
+- Dictionary: `{_ : <type>}` is a record whose fields are of type `<type>`.
+- Enum: ``[| `tag1, .., `tagn |]`` is an enumeration comprised of alternatives
+  `` `tag1 ``, .., `` `tagn``. Tags have the same syntax as identifiers and must
+  be prefixed with a backtick `` ` ``. As for record fields, they can be however
+  espaced with double quotes if they contain special characters:
```
vkleen, line 680:
```suggestion
  be prefixed with a backtick `` ` ``. As for record fields, they can however be
  enclosed in double quotes if they contain special characters:
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+  `` `"tag with space" ``.
+- Arrow: `<source> -> <target>` is a function taking an argument of type
+  `<source>` and returns values of type `<target>`.
+- Forall: `forall var1 .. varn. <type>` is a polymorphic type quantifying over type
+variables `var1`, .., `varn`.
+- Record: see the next section [Record types](#record-types).
+
+Type variables bound by a `forall` are only visible inside types (any of the
+constructor listed above). As soon as a term expression arise under a `forall`
```
vkleen, line 689:
```suggestion
constructor listed above). As soon as a term expression arises under a `forall`
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+  │
+  = The type of the expression was expected to be `Number`
+  = The type of the expression was inferred to be `String`
+  = These types are not compatible
+```
+
+#### Record types
+
+Record types are syntactically a restricted subset of records literal. They are
+handled differently from normal record literals with respect to typechecking.
```
vkleen, line 745:
```suggestion
Record types are syntactically a restricted subset of record literals. They are
handled differently than normal record literals with respect to typechecking.
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+> # while MyDyn isn't a proper type, the right record respect all the
+  # requirements of a record type
+let MyDyn = fun label value => value in
+{foo = 1, bar | MyDyn = "foo"} : {foo : Number, bar : MyDyn}
+{ bar = "foo", foo = 1 }
+```
+
+## Metadata
+
+Metadata are used to attach type and contract annotations, documentation, a
```
vkleen, line 798:
```suggestion
Metadata is used to attach type and contract annotations, documentation, a
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
...
+merge priority or other decorations to record fields (and record fields only).
+Multiple metadata can be chained. A metadata is introduced with the syntax
```
vkleen, line 800:
```suggestion
Multiple metadata annotations can be chained. Metadata is introduced with the syntax
```

vkleen (approved):
Very nice 👍 I found some typos and I added some suggestions for bringing stdlib functions in line with the new `std` namespace. Feel free to ignore the latter if you would rather take care of all similar references in the user manual at once later.

doc/manual/merging.md
yannham: I don't have a good one either :sweat_smile: I think data is technically plural (the plural of _datum_) but is both plural and singular in practice. Maybe should I summon a native speaker such as @matthew-healy

doc/manual/merging.md
yannham: Ah, good catch of this horrible Frenchism :stuck_out_tongue:

doc/manual/syntax.md
yannham: Waiting for the conclusion of the "plural vs singular" metadata for this one.

yannham:
> Feel free to ignore the latter if you would rather take care of all similar references in the user manual at once later.

Yes, this is my plan indeed.

doc/manual/merging.md
```diff
@@ -561,10 +573,17 @@ to a field `foo`, merging ensures that whatever is this field merged with,
 including being dropped in favor of another value, the final value for `foo` has
 to respect the contract as well or the evaluation will fail accordingly.
 
+This is only true for *contracts attached directly to record fields* (either
+directly, or coming from an enclosing record contract). In particular, `{foo |
+Num = 1} & {foo | force = "bar"}` will fail, but `{foo = (1 | Num)} & {foo |
```
ebresafegaga, line 578:
```suggestion
This is only true for *contracts attached directly to record fields* (either
directly, or coming from an enclosing record contract). In particular, `{foo |
Number = 1} & {foo | force = "bar"}` will fail, but `{foo = (1 | Number)} & {foo |
```

doc/manual/syntax.md
```diff
@@ -619,19 +648,182 @@ error: contract broken by a value.
 3
 ```
 
-Adding documentation can be done with `| doc < string >`. Examples:
+### Types
+
+The Nickel syntax mixes both terms and types in the same namespace. The
+following program is perfectly legal: `let value = Number -> (fun value label =>
+value) in ((fun x => x + 1) : value)`. See the
+[RFC002](https://github.com/tweag/nickel/blob/d723a3721c6b0fe9c4b856e889bb7211d6136665/rfcs/002-merge-types-terms-syntax.md)
```
ebresafegaga, line 656: Is there any reason you are referencing the `RFC002` file from a particular commit instead of, say, `master`?

ebresafegaga (approved):
Looks good 👍 I made a comment about the source of a link and added suggestion to rename `Num` to `Number`

doc/manual/syntax.md
yannham: But in general the RFC should be pretty much set in stone (if you need to evolve, better write a new one), so the benefit of being renaming/move/deletion resilient is probably higher than the risk of getting out of date?

PR #1244, opened 2023-04-13
Fix position bug in context completion 

### LINKED ISSUE #1221: No auto-completion in tf-ncl
**Describe the bug**
When initializing a flake from the template of https://github.com/tweag/tf-ncl (say commit dcdbc5a09778336cfdfa558642a56fe668e192fb), and using some provider (say "google"), we don't get any completion nor documentation in the LSP when hacking on the example. Given that the contract is defined as a succession of records and dictionary contracts, I don't see an obvious reason why we wouldn't have completion (another thing to consider is that the generated contract is 300K lines and 16MB, so the LSP might choke on it - to be investigated)

**To Reproduce**

1. Initialize a flake from tf-ncl on commit `dcdbc5a09778336cfdfa558642a56fe668e192fb`: `nix flake init --template github:tweag/tf-ncl/dcdbc5a09778336cfdfa558642a56fe668e192fb`
2. Replace the `null` provider by `google` L.26
3. Start the development shell (`nix develop`)
4. Hack on `hello-tf.ncl`, and try to add new fields to the `config` record, waiting for completion to help. Nothing happens.

**Expected behavior**

That context auto-completion works.

**Additional context**

The issue seems to be that `Tf.Config` define a schema like this

```nickel
    Config = {
        config | Schema,
        renderable_config = TfNcl.mkConfig config,
        ..
    },
  Schema = {
   data | ...
  }
```

Because of the indirection, the LSP doesn't seem to pick the information that `config` have fields defined by `Schema`.

ebresafegaga:
Say we have this program: 
```nickel
let Contract = {
    A = {
        value | String, 
        other | Number,
    }, 

    value1 | String, 
    other1 | Number,
} in 

{ 
   [.]
} | Contract.A
```

and we want we want context completion at the `[.]` point, we get the identifiers `value1` and `other1` instead of `value` and `other`.  The problem is that for some reason `Contract.A` point to `Contract` instead of the field `A`. 

This pull request fixes this (as a side note: this was not the case when context completion was implemented so I don't know how this sneaked it)

Closes #1221

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -348,8 +348,8 @@ fn find_fields_from_term(
         Term::Annotated(annot, term) => {
             find_fields_from_term_with_annot(annot, Some(term), path, info)
         }
-        Term::Var(..) | Term::Op1(UnaryOp::StaticAccess(..), _) => {
-            let pos = term.pos;
+        Term::Var(ident) | Term::Op1(UnaryOp::StaticAccess(ident), _) => {
+            let pos = ident.pos;
             let span = pos.unwrap();
             let locator = (span.src_id, span.start);
             // This unwrap is safe because we're getting an expression from


yannham:
@ebresafegaga this PR is a draft one, is it on purpose?

ebresafegaga:
> @ebresafegaga this PR is a draft one, is it on purpose?

I was still doing some testing. Should be ready for review now 👍

vkleen (approved):

PR #1246, opened 2023-04-13
Simplify formatting instructions in LSP

yannham:
The current instructions for formatting in the LSP are a bit confusing to me:

- The first step says to install topiary, and the second says to clone the topiary repository. However cloning the topiary repository is, to my knowledge, required to install topiary in the first place. The steps aren't in the chronological order.
- We are adding specific instructions about `TOPIARY_LANGUAGE_DIR` that made me think that NLS had specific requirements. However, it doesn't: we just need to have a topiary executable available in the path, and able to be run from any location. It turns out the latter requirement isn't trivial for Topiary, but IMHO this is the Topiary README's responsibility (cf https://github.com/tweag/topiary/pull/424). We shouldn't duplicate Topiary installation instructions here, which could get out of sync, etc.
- There is an "Alternatives" section, but this section doesn't speak about currently working alternatives. This section actually enumerates what NLS _could do_ differently, design-wise, to interface with topiary: while this information is valuable, the NLS README is probably not the place where to put it. A Nickel user doesn't really care about what we could do differently in the future, but cares about what they can do as of today, and how. We should probably move this information to an issue though, to avoid losing it.

This PR tries to simplify the setup guide, by removing the "Alternatives" section and by pointing to Topiary's README instead of duplicating (or adding our own variation of) the installation instructions.

### DIFF
--- a/lsp/nls/README.md
+++ b/lsp/nls/README.md
@@ -12,45 +12,11 @@ it in VSCode, (Neo)Vim and Emacs.
 
 ## Formatting Capabilities
 
-Formatting in `nls` is currently based on [topiary](https://github.com/tweag/topiary).
+Formatting in `nls` is currently based on
+[topiary](https://github.com/tweag/topiary).
 
-To use it successfully, you need ensure you follow these steps:
-
-1. Have the `topiary` binary installed in your `PATH`. See [here](https://github.com/tweag/topiary#installing).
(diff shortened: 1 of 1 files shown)


vkleen (approved):
I like the new description and the rationale 👍

ebresafegaga (approved):
Thanks. This is much better 👍

PR #1254, opened 2023-04-17
Pass over `doc/manual/syntax.md`

vkleen:


ebresafegaga (approved):

doc/manual/syntax.md
```diff
@@ -302,33 +304,33 @@ The following examples show how symbolic strings are desugared:
...
 
 An enum tag `` `foo `` is serialized as the string `"foo"`:
 
 ```nickel
-let as_yaml_string = std.serialize `Yaml {foo = `bar}
-# gives "---
-#       foo: bar
-#       "
-
+> std.serialize `Yaml {foo = `bar}
```
yannham, line 327: Nitpick: maybe JSON would actually do a better job at showing that it is a string

doc/manual/syntax.md
```diff
@@ -385,17 +387,17 @@ Arrays can be concatenated with the operator `@`:
...
+### Records
 
-Records are key-value storage, or in Nickel terms, field-value storage. They are
-delimited by `{` and `}`, and elements are separated with `,`. Field-value
-elements are noted as `field = value`. The fields are strings, but can be
-written without quotes `"` if they respect identifiers syntax. Values can be of
+Records are key-value storage, or in Nickel terms, field-value storage. They
+are delimited by `{` and `}`, and elements are separated with `,`. A field
+definition is written as `field = value`. The fields are strings, but can be
+written without quotes `"` if they would be valid identifiers. Values can be of
```
yannham, line 395:
```suggestion
written without quotes `"` if they are valid identifiers. Values can be of
```

doc/manual/syntax.md
```diff
@@ -238,10 +240,10 @@ That is precisely the use-case for symbolic strings:
 ```
 
 This example is an excerpt of a Nix configuration written in Nickel, emulating
-Nix string contexts. Lines 4 to 8 define a symbolic string. Values `inputs.gcc`,
-`inputs.hello`, etc. aren't actually strings, but arbitrary records, because
-they carry additional context. Yet, they can be interpolated as if they were
-strings.
+Nix string contexts. Lines 4 to 8 define a symbolic string. The values
```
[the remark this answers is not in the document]
vkleen: That's a very good point. It won't have line numbers on the website.

doc/manual/syntax.md
```diff
@@ -467,12 +469,13 @@ Examples:
 
 ### Let-In
 
-Let-in allows the binding of an expression. It is used as
-`let <rec?> <ident> = <expr> in <expr>`. The `rec` keyword in Let-in constructs
-allows the let binding to become recursive, enabling the use of the `<ident>`
-within the first `<expr>`.
+Let-in allows binding an expression to a variable. It is used like `let <rec?>
```
yannham, line 472:
```suggestion
A `let` binds an expression to a variable. It is used like `let <rec?>
```

doc/manual/syntax.md
```diff
@@ -467,12 +469,13 @@ Examples:
...
+<ident> = <expr> in <expr>`. The `rec` keyword in Let-in constructs allows the
+let binding to become recursive, enabling the use of `<ident>` within the bound
```
yannham, line 474:
```suggestion
<ident> = <expr> in <expr>`. The `rec` keyword makes the binding recursive,
enabling the use of `<ident>` within the bound
```

doc/manual/syntax.md
```diff
@@ -491,26 +494,26 @@ true
...
 ## Functions
 
 A function is declared using the `fun` keyword, then arguments separated with
 spaces, and finally an arrow `=>` to add the body of the function. To call a
-function, just add the arguments after it separated with spaces. Functions in
-Nickel are curried, meaning that a function taking multiple arguments is
-actually a function that takes a single argument and returns a function taking
-the rest of the arguments, until it is applied.
+function, just write the arguments after it separated with spaces. Functions in
+Nickel are curried: A function taking multiple arguments is actually a function
```
yannham, line 506:
```suggestion
Nickel are curried: a function taking multiple arguments is actually a function
```

doc/manual/syntax.md
```diff
@@ -567,25 +567,25 @@ Examples:
...
+`{<field> : <type> = <value>}` as well.
 
-A type wildcard `_` indicates that part of a type is unknown to the user (or not
-worth spelling out). The typechecker will infer it. Slapping an `: _` on top of
-an existing expression is particulary useful for debugging, as it's the simplest
-way to have the typechecker run on an expression without having to come up with
-a type.
+A type wildcard `_` indicates that part of a type is unknown to the user (or
+is not worth spelling out). The typechecker will attempt to infer it. Adding a
+wildcard type annotaiton `: _` to an existing expression is particulary useful
```
yannham, line 584:
```suggestion
wildcard type annotation `: _` to an existing expression is particularly useful
```

doc/manual/syntax.md
```diff
@@ -746,20 +746,22 @@ handled differently than normal record literals with respect to typechecking.
...
 - Each field doesn't have any other metadata attached (see [Metadata](#metadata))
-- The record literal might have a record tail, written as
-  `{ <fields> ; <tail> }`. The tail appears at the end of the field declarations
-  and is preceded by `;`. `<tail>` must be a valid identifier.
 
-If the above properties are satisfied, a record literal is considered to be a
+If these properties are satisfied, a record literal is considered to be a
 record type by the typechecker.
 
+A record literal which is to be interpreted as a record type may have a *record
```
yannham, line 756:
```suggestion
A record literal which is interpreted as a record type may have a *record
```

doc/manual/syntax.md
```diff
@@ -768,9 +770,9 @@ Examples:
 > {foo.bar = 1, baz = 2} : {foo: {bar : Number}, baz : Number}
 { baz = 2, foo = { bar = 1 } }
 
-># the right-hand side is missing a type annotation for baz, so it isn't a
- # record type
- {foo = 1, bar = "foo" } : {foo : Number, bar : String, baz}
+The right-hand side is missing a type annotation for baz, so it isn't a
```
yannham, line 773: It's a bit strange to mix plain text english inside a seemingly REPL session

yannham (approved):
Sorry, I initially meant to approve, although I left a bunch of comments.

yannham:
Ah, I broke the CI because of markdown-lint. Let me spin a PR for that.

doc/manual/syntax.md
vkleen: So, you'd prefer the comments? Or should we try to reformulate the entire example?

doc/manual/syntax.md
```diff
@@ -188,21 +189,22 @@ Examples:
 "Hello World"
 ```
 
-Multiline strings are "indentation-aware". This means that one could use an
-indented string interpolation and the indentation would behave as expected:
+Multiline string interpolation is "indentation-aware". This means that you can
+interpolate a string containing multiple lines into another string with prefixed
+indentation and the result will be as expected:
```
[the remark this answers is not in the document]
vkleen: I've made the sentence a bit terser to rely more on the example afterwards 👍

doc/manual/syntax.md
yannham: In fact, I don't know, maybe it's fine. Or maybe as you did (comment before the example), but still prefixing them with `#`? I know it doesn't correspond to anything anymore (it's not really an REPL prompt nor a valid Nickel file nor plain English), but it might at least make the distinction between code, comment and result of the evaluation of the code clearer?

vkleen: I was a bit worried about breaking up the flow of the examples by interspersing prose outside of the code block, but maybe that's not such a bad thing anyway. In fact, I like the rendering on the website with smaller code blocks more.

yannham: In any case, I just wanted to share my feeling, but I have no strong opinion either and I trust you to keep what you think is best, including the original presentation.

PR #1262, opened 2023-04-18
Replace Term::deep_repr by the pretty printer

### LINKED ISSUE #1257: Control characters and quotes in strings are not escaped when printed
Nickel doens't escape quotes in strings output from the interpreter:
```
nickel> "a\"bcd\"d"
"a"bcd"d"
```
Other special characters are also not escaped, leading to this confusing output:
```
nickel> "a\r"
"a
```
The interpreter should escape control characters and quotes when displaying a string value.

vkleen:
This PR gets rid of `Term::deep_repr` and replaces all calls to it by invocations of the pretty printer. Incidentally, it ensures that quotes and newlines in strings are correctly escaped when printing evaluation results.

As a bonus, the pretty printer formats records with many fields on multiple lines in the REPL if they don't fit into 80 columns.

Closes #1257

### DIFF
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -22,7 +22,6 @@ pub mod record;
 pub mod string;
 
 use array::{Array, ArrayAttrs};
-use pretty::BoxAllocator;
 use record::{Field, FieldDeps, FieldMetadata, RecordData, RecordDeps};
 use string::NickelString;
 
@@ -1702,7 +1701,7 @@ impl std::fmt::Display for RichTerm {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         use crate::pretty::*;
 
-        let allocator = BoxAllocator;
+        let allocator = pretty::BoxAllocator;
 
         let doc: DocBuilder<_, ()> = self.clone().pretty(&allocator);
         doc.render_fmt(80, f)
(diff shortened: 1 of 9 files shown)


src/term/mod.rs
```diff
@@ -1734,7 +1701,6 @@ impl From<Term> for RichTerm {
 impl std::fmt::Display for RichTerm {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         use crate::pretty::*;
-        use pretty::BoxAllocator;
```
yannham, line 1737: Any reason to move this specific `use` from the local function to the top-level? We don't really have consistent rules right now about where to place `use` statements, but I vaguely intuit that it makes more sense to keep it local if it's very specific to the current function, and doesn't have much to do with the rest of the module, which seems to be the case here.

yannham (approved):
Sounds better, but I think we had a discussion back then about what to do about e.g. functions. Where you're defining a function in the REPL, it's customary (as it, in other languages) to not re-print again the whole code of the function. Might be the same for documentation and metadata in general (I guess in the REPL you really want to print values, not the fuss around).

IIRC, one proposal was to implement `Pretty` on a more elaborated type, basically isomoprhic to `(Term, PrettyConfig)`, with `PrettyConfig` controls the options of the pretty printer. That being said, we can consider that in a second step.

src/term/mod.rs
vkleen: That's a very good question. It might have happened during a previous iteration, but right now there's no good reason. I think I'll revert this part.

vkleen:
I've made a tracking issue for making the pretty printer configurable at #1349.

PR #1263, opened 2023-04-18
Pass on the documentation of `std.array`

yannham:
Update the documentation of `std.array`. Make the example executable by fully qualifying the name of the stdlib functions.

PR #1264, opened 2023-04-18
Pass on the documentation of std.contract

yannham:
Depends on #1263.

### DIFF
--- a/stdlib/std.ncl
+++ b/stdlib/std.ncl
@@ -909,8 +909,8 @@
 
     label
       | doc m%"
-          The label submodule provides functions that manipulate the label of a
-          contract.
+          The label submodule provides functions that manipulate the label
+          associated with a contract application.
 
           A label is a special opaque value automatically passed by the Nickel
           interpreter to contracts when performing a contract check.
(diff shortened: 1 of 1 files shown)


stdlib/std.ncl
```diff
@@ -909,8 +909,8 @@
 
     label
       | doc m%"
-          The label submodule gathers functions that manipulate the label of
-          a contract.
+          The label submodule provides functions that manipulate the label of a
+          contract.
```
vkleen, line 913: I'm not sure what else to write here, but I feel like a label doesn't really belong to a contract, but to a contract application.

vkleen (approved):

stdlib/std.ncl
yannham: Ah, you're right indeed. I guess saying exactly that is fine.

yannham, line 913:
```suggestion
          The label submodule provides functions that manipulate the label
          associated to a contract application.
```

yannham: @vkleen do you think that'd be ok?

vkleen: I'd say "associated with" (because otherwise my thesis adviser will come haunt me) and "manipulates". Otherwise, looks good 👍

PR #1265, opened 2023-04-18
A pass over `user-manual/contracts.md`

vkleen:
Mostly minor tweaks.

### DIFF
--- a/doc/manual/contracts.md
+++ b/doc/manual/contracts.md
@@ -7,10 +7,10 @@ slug: contracts
 **Note**: Most of the examples of this document are run in the Nickel REPL. To
 start a session, run `nickel repl` in your terminal.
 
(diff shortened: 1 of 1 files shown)


doc/manual/contracts.md
```diff
@@ -444,7 +444,7 @@ but potentially surprising differences.
 
 One concerns open contracts. Merging never requires the presence of specific
 fields: thus, the contract `{bar | String}` attached to `foo` will actually
-behave as an open contract, even if you didn't use `..`. This might or might not
+behave like an open contract, even if you didn't use `..`. This might or might not
```
yannham, line 447: I would tend to think both "behave as" and "behave like" are valid here. Out of curiosity, what is the nuance making "behave like" more adapted here?

yannham (approved):

PR #1266, opened 2023-04-19
Pass on documentation of std.enum and std.function

yannham:


stdlib/std.ncl
```diff
@@ -1213,24 +1215,42 @@
...
         # Examples
 
         ```nickel
-        first 5 7 => 5
+        std.function.first 5 7
+          => 5
         ```
       "%
-      = const,
+      = fun x y => x,
```
yannham, line 1253: The definition is very small, and overall it's simpler to avoid recursive definition in the stdlib

stdlib/std.ncl
```diff
@@ -1240,10 +1260,11 @@
         # Examples
 
         ```nickel
-        second 5 7 => 7
+        std.function.second 5 7
+          => 7
         ```
       "%
-      = flip const,
+      = fun x y => y,
```
yannham, line 1267: Same rationale here

stdlib/std.ncl
```diff
@@ -1129,20 +1129,25 @@
...
         `TagOrString` is typically used in conjunction with an enum type, to
-        accept tags represented as strings (e.g. coming from a JSON
-        serialization) as well.
+        accept both actual enum tags and tags represented as strings (e.g. coming from a
+        JSON serialization).
+
+        **Warning**: contract are applied in-order. The pattern described here
+        requires that `TagOrString` is applied before the corresponding enum
+        contract. Thus, `TagOrString` must appears before the enum  contract in
+        the annotations, as in the example below.
```
vkleen, line 1138:
```suggestion
        **Warning**: contracts are applied in-order. The pattern described here
        requires that `TagOrString` is applied before the corresponding enum
        contract. Thus, `TagOrString` must appear before the enum  contract in
        the annotations, as in the example below.
```

stdlib/std.ncl
```diff
@@ -1171,36 +1176,33 @@
...
+      = fun x => x,
 
     compose
       : forall a b c. (b -> c) -> (a -> b) -> (a -> c)
       | doc m%"
-        Function composition, from right to left. That is,
+        Function composition, from right to left. That is, `compose f g x` is
+        `f (g x)`.
+
+        # Example
```
vkleen, line 1198:
```suggestion
        # Examples
```
We're using the plural everywhere else, if I remember right. Even if there is only a single example given.

vkleen (approved):

PR #1269, opened 2023-04-19
Pass on the documentation of std.string

yannham:
This PR also makes the string to bool conversion stricter. It used to accept an uppercase letter as the first letter like "True" or "False", but that's kinda arbitrary. The user can still use `lowercase` and other string functions to achieve the same behavior if needed.

### DIFF
--- a/stdlib/std.ncl
+++ b/stdlib/std.ncl
@@ -2009,7 +2009,7 @@
       : String -> String
       | doc m%"
         Returns the uppercase version of a string. Unicode extended grapheme
-        clusters without an uppercase version are left the same.
+        clusters without an uppercase version are left untouched.
 
         # Examples
 
@@ -2028,7 +2028,7 @@
       : String -> String
       | doc m%"
         Returns the lowercase version of a string. Unicode extended grapheme
-        clusters without an uppercase version are left the same.
+        clusters without a lowercase version are left untouched.
 
 
         # Examples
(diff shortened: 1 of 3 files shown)


stdlib/std.ncl
```diff
@@ -1995,46 +1999,47 @@
...
       "%
       = fun s => %str_chars% s,
 
     uppercase
       : String -> String
       | doc m%"
-        Returns the uppercase version of the given character (including non-ascii characters) if it exists, the same
-        character if not.
+        Returns the uppercase version of a string. Unicode extended grapheme
+        clusters without an uppercase version are left the same.
```
vkleen, line 2012:
```suggestion
        clusters without an uppercase version are left untouched.
```
"left the same" sounds a bit weird to me. But I don't have a good justification as usual 😅

stdlib/std.ncl
```diff
@@ -1995,46 +1999,47 @@
...
       "%
       = fun s => %str_uppercase% s,
 
     lowercase
       : String -> String
       | doc m%"
-        Returns the lowercase version of the given character (including non-ascii characters) if it exists, the same
-        character if not.
+        Returns the lowercase version of a string. Unicode extended grapheme
+        clusters without an uppercase version are left the same.
```
vkleen, line 2031:
```suggestion
        clusters without a lowercase version are left untouched.
```

vkleen (approved):

stdlib/std.ncl
yannham: Right, I think I initially used "remains the same" and then wanted to write "left untouched", but ended up with a mix of the two :sweat_smile:

PR #1270, opened 2023-04-19
Pass on the top-level symbols of `std`

yannham:


stdlib/std.ncl
```diff
@@ -2482,33 +2481,48 @@
...
+      is reached, which is either:
+
+       - an array
+       - a record
+       - a function (or a match statement, which is function in disguise)
+       - a constant (a number, a boolean, a string literal or an enum tag)
+       - a contract label
+
+      Note that `seq` won't evaluate _the content_ of arrays and records
+      further. Please see `std.deep_seq` or `std.force` to recursively force the
```
yannham, line 2494:
```suggestion
      further. Please see `std.deep_seq` to recursively force the
```

yannham: There's in fact no such thing as a `std.force`

stdlib/std.ncl
```diff
@@ -2469,10 +2468,10 @@
...
 
       ```nickel
-      typeof [ 1, 2 ] =>
-        `Array
-      typeof (fun x => x) =>
-        `Function
+      std.typeof [ 1, 2 ]
+        => `Array
+      std.typeof (fun x => x)
+       => `Function
```
vkleen, line 2474:
```suggestion
        => `Function
```

Issue #1228, opened 2023-04-04
Polymorphic contracts and contract propagation seem fundamentally incompatible


yannham:
#1141 made dictionary contracts (or, should I say, dictionary type, as we will see that the difference matters) behave more lazily, and be overriding friendly. The gist is that `{foo = ..., bar = ..., baz = ...} | {_ : T}` is now strictly equivalent to write `{foo | T = ..., bar | T = ..., baz | T = ...}`. This unveiled a subtle but important question about the interaction between polymorphic contracts and the fact merging propagates contracts.

## Propagation

You can read more into RFC005, but _propagation_ in Nickel means that if I merge `{foo | Number | default = 5}` with `{foo = "a"}`, the contract will propagate to `"a"` and blame:

```
nickel> {foo | Number | default = 5} & {foo = "a"}
error: contract broken by a value
  ┌─ repl-input-3:1:39
  │
1 │ {foo | Number | default = 5} & {foo = "a"}
  │        ------                         ^^^ applied to this expression
  │        │                               
  │        expected type
  │
  ┌─ <unknown> (generated by evaluation):1:1
  │
1 │ "a"
  │ --- evaluated to this value
``` 

This is inspired from CUE and NixOS, in the spirit that merging can only refine the values and strengthen the contraints (in practice, the contracts).

## Polymorphic contracts

Nickel implements polymorphic contracts which enforce parametricity. This is done thanks to a technique called sealing. Anything which enters a polymorphic function as a generic value is sealed, and this value is only unsealed when exiting the function, ensuring that a polymorphic function can't inspect a generic argument: it can only pass it around to other functions (with a matching type) or return it.

## The issue

Here is the problematic case which appeared after #1141. Recall the type signature of `record.insert`:

```nickel
insert : forall a. String -> a -> {_: a} -> {_: a}
```

Now, let's run the following example on current master:

```nickel
nickel> let singleton_foo = fun value => record.insert "foo" value {}
in (singleton_foo {sub_left = 1}) & {foo.sub_right = 2} 
error: contract broken by a function
   ┌─ <stdlib/record.ncl>:65:29
   │
65 │       : forall a. String -> a -> { _ : a } -> { _ : a }
   │                             - expected type of an argument of an inner call
   │
   ┌─ <unknown> (generated by evaluation):1:1
   │
 1 │ x
   │ - evaluated to this value
   │
   = This error may happen in the following situation:
   [...]
```

[...]

PR #1271, opened 2023-04-19
Make type annotations to not propagate through merging

yannham:
First step toward fixing #1228. Make type annotations behave differently from contract annotations, and in particular, make them not propagate through merging.

## Content

This PR makes a distinction between the contracts coming from a type annotation and the one derived from a contract annotation. In particular, the former aren't stored lazily inside the pending contracts of record fields, nor are they propagated through merging anymore. The contract application for a type annotation is generated statically during program transformations once and for all, making such contract checks local.

### Conversion from a record type literal

Doing so, we encountered an interesting issue, which was that the following program was accepted:

```nickel
let C = {foo : Number} in
{foo = "a"} | C
```

Here, `{foo : Number}` was parsed (on current `master`) as a record literal with one field and without a definition. But because record contract application is mostly merging, the type annotation `{foo : Number}` isn't propagated anymore.

But if we write

```nickel
{foo = "a"} | {foo : Number}
```

Then it do report a contract violation, because `{foo : Number}` is now considered first as a record type, then converted to a proper contract (not a record contract, but a builtin special function), and applied correctly. 

It turns out this is a discrepancy already present on master, and `{foo : Number}` qualifies as a record type but is only considered so in _a type position_. It's just that on master, the difference between the interpretation as a record type and as a record literal is much less noticeable (but still here: for example, record type contracts currently kill metadata, it's a bug, but this is nonetheless an observable distinction between the first and the second example already on master).

**This PR makes a record literal qualifying as a record type to be always interpreted as such, even in a term position**. As an illustration:

```nickel
nickel> {foo : Num}
<func>

```

We get an opaque function instead of a record. I think it's mostly fine, because it doesn't make much sense to apply record operations such as merging on such an object anyway. It can be considered the same as `Number -> {_ : String}`, that is some opaque value that can only be applied as a contract.

### The typed field without definition restriction

After the change, something like `{foo : Number, bar | String}` can be misleading. It's not a record type, so we have the same behavior as the initial problematic example, which is that when used as a contract this expression will never enforce that `foo` is a number (but it will enforce that `bar` is a `String`). Because a freestanding type annotation on a field without a definition doesn't make much sense (after the current change, they are useless, and can be removed to exactly the same effect, because they don't apply to any expression), we simply make them illegal at the parsing stage. That is, you must either write a record type, or you can't write a field with a type annotation but no definition.

It might also help people not mixing and confusing type and contract annotations, mostly by forcing them to write contract annotations everywhere inside record contracts, outside record type.

## TODOs

- [x] Fix the issue that right now, doing `{foo = 5} | {foo : String}` fails as expected, but `let Foo = {foo : String} in {foo = 5} | Foo` doesn't. The reason is that the latter isn't inside a type annotation, and is thus parsed as a record literal which is applied through merging, making the type annotation inoperative (which is the point of this PR). I think the solution is to make `{foo : String}` to become the contract associated to the corresponding record type all the time, not only inside type annotations. Additionally, we should forbid a type annotation to be attached to a field without a definition inside a normal record, because it's useless (that is, forbid `{foo : String, bar | Number}` or `{foo : String, bar = 2}` because the type annotation is useless and it might lead to surprising result; it can become a warning later on).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1590,7 +1590,7 @@ impl IntoDiagnostics<FileId> for ParseError {
                 ]),
             ParseError::TypedFieldWithoutDefinition { field_span, annot_span } => {
                 Diagnostic::error()
-                    .with_message("statically typed field without definition")
+                    .with_message("statically typed field without a definition")
                     .with_labels(vec![
                         primary(&field_span).with_message("this field doesn't have a definition"),
                         secondary(&annot_span).with_message("but it has a type annotation"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -385,6 +385,26 @@ pub enum ParseError {
     /// - a variable is used as both a record and enum row variable, e.g. in the
     ///   signature `forall r. [| ; r |] -> { ; r }`.
     TypeVariableKindMismatch { ty_var: Ident, span: RawSpan },
+    /// A record literal, which isn't a record type, has a field with a type annotation but without
+    /// a definition. While we could technically handle this situation, this is most probably an
+    /// error from the user, because this type annotation is useless and, maybe non-intuitively,
+    /// won't have any effect as part of a larger contract:
+    ///
+    /// ```nickel
+    /// let MixedContract = {foo : String, bar | Number} in
+    /// { foo = 1, bar = 2} | MixedContract
+    /// ```
+    ///
+    /// This example works, because the `foo : String` annotation doesn't propagate, and contract
+    /// application is mostly merging, which is probably not the intent. It might become a warning
+    /// in a future version, but we don't have warnings for now, so we rather forbid such
+    /// constructions.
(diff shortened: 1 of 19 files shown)


src/error.rs
```diff
@@ -1561,6 +1588,23 @@ impl IntoDiagnostics<FileId> for ParseError {
                     String::from("Type variables may be used either as types, polymorphic record tails, or polymorphic enum tails."),
                     String::from("Using the same variable as more than one of these is not permitted.")
                 ]),
+            ParseError::TypedFieldWithoutDefinition { field_span, annot_span } => {
+                Diagnostic::error()
+                    .with_message("statically typed field without definition")
```
vkleen, line 1593:
```suggestion
                    .with_message("statically typed field without a definition")
```

vkleen (approved):

PR #1272, opened 2023-04-20
Introduce dictionary contracts

### LINKED ISSUE #1228: Polymorphic contracts and contract propagation seem fundamentally incompatible

yannham:
Depends on #1271
Closes #1228

Follow-up of #1271. Second part of #1228. Introduce a new construct `{_ | T}` for dictionary contracts, which act like record contracts and propagate. Restore the previous implementation for the contract of `{_ : T}`, which amounts to simply map the contract once over the elements the record. The contract for `{_ | T}` is implemented as the previous (before this PR) contract of `{_ : T}`, the one which is lazy and propagates.

## AST

How to represent the `{_ | T}` node in the AST? I initially thought it should simply be syntactic sugar for a builtin contract, that is `{_ | T}` could be desugared at parsing time to `$dict_contract contract_of_T`. However, that would mean `{_ | T}` would not be understood by the LSP for completion, and wouldn't benefit from the type path calculation inside contract error reporting which helps underline the precise sub contract that fails. Even without thinking about a deadline for 1.0, restoring those capacities is non trivial, because matching on specific primop operation is a bit fragile, and `contract_of_T` would also obfuscate what `T` initially was.

A second solution would be to have a proper AST node for `{_ | T}`, which could be then treated as required by the LSP and the contract error reporting error, and could be evaluated away to the `$dict_contract` application. However, this incurs a non trivial amount of changes, and until now we refrained until from having too many parsing-specific AST nodes (this node is irrelevant and evaluated away at runtime, or could even be at transformation time). This might be the right solution in the long term, but I think we would be better off first splitting the AST in two types, and properly redesign them, with proper nodes for types embedded inside terms, etc.

In the end, the LSP and the error reporting code should treat dictionary contracts the same as a dictionary type: the simplest change was just to use `TypeF::Dict` for both, just with an additional parameter that only impacts contract generation. This means that `{_ | T}` is currently considered a static type equivalent to `{_ : T}`, but I don't think that's a problem. It still can't capture type variables anymore, and its contract should still be enough for type/blame safety (its contract used to be the one of `{ _ : T}` before this PR, anyway).

## TODO

- [x] fix `forall a. {_ | a}` being accepted. Hint: something to do with fixing type variables.

### DIFF
--- a/stdlib/internals.ncl
+++ b/stdlib/internals.ncl
@@ -104,7 +104,7 @@
     else
       %blame% (%label_with_message% "not a record" label),
 
-   # Lazy dictionary contract for `{_ | T}`
+  # Lazy dictionary contract for `{_ | T}`
   "$dict_contract" = fun contract label value =>
     if %typeof% value == `Record then
       %record_lazy_assume% (%go_dict% label) value (fun _field => contract)
@@ -114,7 +114,12 @@
   # Eager dictionary contract for `{_ : T}`
   "$dict_type" = fun contract label value =>
     if %typeof% value == `Record then
-      %record_map% value (fun _field field_value => %assume% contract label field_value)
+      %record_map%
+        value
+        (
+          fun _field field_value =>
+            %assume% contract (%go_dict% label) field_value
+        )
     else
       %blame% (%label_with_message% "not a record" label),
 
--- a/stdlib/internals.ncl
+++ b/stdlib/internals.ncl
@@ -104,9 +104,17 @@
     else
       %blame% (%label_with_message% "not a record" label),
 
-  "$dyn_record" = fun contract label value =>
+   # Lazy dictionary contract for `{_ | T}`
+  "$dict_contract" = fun contract label value =>
     if %typeof% value == `Record then
       %record_lazy_assume% (%go_dict% label) value (fun _field => contract)
+    else
+      %blame% (%label_with_message% "not a record" label),
+
+  # Eager dictionary contract for `{_ : T}`
+  "$dict_type" = fun contract label value =>
+    if %typeof% value == `Record then
+      %record_map% value (fun _field field_value => %assume% contract label field_value)
     else
       %blame% (%label_with_message% "not a record" label),
 
@@ -152,8 +160,7 @@
 
   # Recursive priorities operators
 
-  "$rec_force" = fun value => %rec_force% (%force% value),
-  "$rec_default" = fun value => %rec_default% (%force% value),
+  "$rec_force" = fun value => %rec_force% (%force% value),"$rec_default" = fun value => %rec_default% (%force% value),
 
   # Provide access to std.contract.Equal within the initial environement. Merging
   # makes use of `std.contract.Equal`, but it can't blindly substitute such an
@@ -162,3 +169,4 @@
   # environment and prevents it from being shadowed.
   "$stdlib_contract_equal" = std.contract.Equal,
 }
+
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -248,8 +248,8 @@ fn find_fields_from_type(
 ) -> Vec<IdentWithType> {
     match &ty.types {
         TypeF::Record(row) => find_fields_from_rrows(row, path, info),
-        TypeF::Dict(ty) => match path.pop() {
-            Some(..) => find_fields_from_type(ty, path, info),
+        TypeF::Dict { type_fields, .. } => match path.pop() {
+            Some(..) => find_fields_from_type(type_fields, path, info),
             _ => Vec::new(),
         },
         TypeF::Flat(term) => find_fields_from_term(term, path, info),
--- a/src/label.rs
+++ b/src/label.rs
@@ -212,8 +212,8 @@ but this field doesn't exist in {}",
                     ..path_span
                 })
             }
-            (TypeF::Dict(ty), next @ Some(Elem::Dict)) => {
-                let path_span = span(path_it, ty)?;
+            (TypeF::Dict { type_fields, .. }, next @ Some(Elem::Dict)) => {
+                let path_span = span(path_it, type_fields)?;
 
                 Some(PathSpan {
                     last: path_span.last.or_else(|| next.copied()),
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -62,10 +62,7 @@ use crate::{
         array::Array,
         make as mk_term,
     },
-    types::{
-        Types, TypeF, EnumRows, EnumRowsF, RecordRows, RecordRowsF,
-        VarKind
-    },
+    types::*,
     position::{TermPos, RawSpan},
     label::Label,
 };
@@ -912,7 +909,19 @@ TypeAtom: Types = {
             Types::from(TypeF::Enum(ty))
     },
     "{" "_" ":" <t: WithPos<Types>> "}" => {
-        Types::from(TypeF::Dict(Box::new(t)))
+        Types::from(TypeF::Dict {
+            type_fields: Box::new(t),
+            attrs: DictAttrs::Eager
+        })
+    },
+    // Although dictionary contracts aren't really type, we treat them like
+    // types for the time being, at least syntactically, as they are represented
+    // using a `TypeF::Dict` constructor as well.
+    "{" "_" "|" <t: WithPos<Types>> "}" => {
+        Types::from(TypeF::Dict {
+            type_fields: Box::new(t),
+            attrs: DictAttrs::Lazy
+        })
     },
     "_" => {
         let id = *next_wildcard_id;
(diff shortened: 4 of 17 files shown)


stdlib/internals.ncl
```diff
@@ -152,8 +165,7 @@
 
   # Recursive priorities operators
 
-  "$rec_force" = fun value => %rec_force% (%force% value),
-  "$rec_default" = fun value => %rec_default% (%force% value),
+  "$rec_force" = fun value => %rec_force% (%force% value),"$rec_default" = fun value => %rec_default% (%force% value),
```
vkleen, line 168: Some weird happened here with the formatting.
```suggestion
  "$rec_force" = fun value => %rec_force% (%force% value),
  "$rec_default" = fun value => %rec_default% (%force% value),
```

vkleen (approved):
I'm wondering whether this means that we can rip out the preliminary fix for polymorphism in record merging. It seems like we're preventing sealed terms from entering recursive envionments now, since the only way of making that happen (if I recall correctly) was to put a type variable into a dictionary contract.

yannham:
>  I'm wondering whether this means that we can rip out the preliminary fix for polymorphism in record merging. It seems like we're preventing sealed terms from entering recursive envionments now, since the only way of making that happen (if I recall correctly) was to put a type variable into a dictionary contract.

Oh, this didn't cross my mind, but it might be the case indeed. Let's keep this somewhere in an issue and check a bit more thoroughly later on, but that would be a good news (and a good sign that, in the end, this PR is maybe the right approach).

PR #1273, opened 2023-04-21
Use `record.insert` for `std.record.from_array` instead of merging

vkleen:
Remove (hopefully) all recursive references from the standrad library. Instead, recursive use of stdandard library functions goes through the `std.` namespace. Incidentally, this generalizes `std.record.merge_all` to `std.merge_all : Array Dyn -> Dyn` and makes `std.record.from_array` use `record.insert` instead of merging.

### DIFF
--- a/stdlib/std.ncl
+++ b/stdlib/std.ncl
@@ -237,16 +237,17 @@
         if length == 0 then
           acc
         else
-          let rec go = fun acc n =>
-            if n == length then
-              acc
-            else
-              let next_acc =
-                %elem_at% array n
-                |> f acc
-              in
-              go next_acc (n + 1)
-              |> %seq% next_acc
+          let rec go =
+            fun acc n =>
+              if n == length then
+                acc
+              else
+                let next_acc =
+                  %elem_at% array n
+                  |> f acc
+                in
+                go next_acc (n + 1)
+                |> %seq% next_acc
           in
           go acc 0,
 
@@ -276,12 +277,13 @@
         "%
       = fun f fst l =>
         let length = %length% l in
-        let rec go = fun n =>
-          if n == length then
-            fst
-          else
-            go (n + 1)
-            |> f (%elem_at% l n)
+        let rec go =
+          fun n =>
+            if n == length then
+              fst
+            else
+              go (n + 1)
+              |> f (%elem_at% l n)
         in go 0,
 
     prepend
@@ -429,11 +431,12 @@
           ```
         "%
       = fun pred l =>
-        let aux = fun acc x =>
-          if (pred x) then
-            { right = acc.right @ [x], wrong = acc.wrong }
-          else
-            { right = acc.right, wrong = acc.wrong @ [x] }
+        let aux =
+          fun acc x =>
+            if (pred x) then
+              { right = acc.right @ [x], wrong = acc.wrong }
+            else
+              { right = acc.right, wrong = acc.wrong @ [x] }
         in
         fold_left aux { right = [], wrong = [] } l,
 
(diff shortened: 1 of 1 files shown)


stdlib/std.ncl
```diff
@@ -237,16 +237,17 @@
...
+            fun acc n =>
+              if n == length then
+                acc
+              else
+                let next_acc =
+                  %elem_at% array n
+                  |> f acc
+                in
+                go next_acc (n + 1)
+                |> %seq% next_acc
```
yannham, line 250: This is strange, it seems this reverts back to an old formatting style.

stdlib/std.ncl
```diff
@@ -276,12 +277,13 @@
...
-          else
-            go (n + 1)
-            |> f (%elem_at% l n)
+        let rec go =
+          fun n =>
+            if n == length then
+              fst
+            else
+              go (n + 1)
+              |> f (%elem_at% l n)
```
yannham, line 286: Same, shouldn't the `fun n =>` appears on the same line as `go`?

yannham (commented):
I wonder what is the motivation for removing recursive references. Is it for consistency? On one hand it incurs two more record accesses than doing a direct reference. It also remove the ability to do monkey patching (I don't know if that's a good idea, but currently, you could technically override `fold_left` with a version which `trace` for example, and the other library functions relying on `fold_left` would use the new version).

I think the original rationale I followed was that reference should't be recursive for inter-module references, or top-level `std` elements. But then maybe it's just simpler to do everything without being recursive, I honestly don't really know.

stdlib/std.ncl
vkleen: The problem is that the Nickel flake devshell still has an old version of topiary 😅

vkleen:
Removing the recursive references was mostly for consistency. But maybe it would be better to split that change out of this PR anyway and make this one just fix `merge_all`.

vkleen:
I've limited this PR to just dealing with `std.record.merge_all`. It's now moved to `std.merge_all : Array Dyn -> Dyn` and can handle more than just records and `std.record.from_array` is again using `%record_insert%` instead of merging.

stdlib/std.ncl
```diff
@@ -2656,4 +2641,18 @@
       ```
     "%
     = fun msg => null | FailWith msg,
+
+  merge_all
+    : Array Dyn -> Dyn
+    | doc m%"
+      Merges an array of records.
```
vkleen, line 2648:
```suggestion
      Merges an array of values.
```

yannham: Would that make sense to restrict the type further to `{_ : Dyn}`? On one hand, there's not many value that are interesting to merge beside record (otherwise it's going to fail unless the operands are equals). On the other hand merge does accept to merge more stuff than records, although it is _very_ restricted.

stdlib/std.ncl
```diff
@@ -815,8 +815,7 @@
                     (
                       fun field =>
                         contract_map."%{field}"
-                    ),
-            `Array =>
+                    ),`Array =>
```
yannham, line 818: This formatting change is fishy.

yannham (approved):

stdlib/std.ncl
vkleen: There seems to be some edge case with topiary? There was a similar formatting issue in another PR today.

stdlib/std.ncl
vkleen: That's sort of what we had before. I felt that it doesn't make much sense to restrict `merge_all` if we don't restrict `(&)` to records. Especially if we end up pointwise merging of arrays or something like that in the future.

stdlib/std.ncl
vkleen: https://github.com/tweag/topiary/issues/439

yannham (approved):

PR #1275, opened 2023-04-21
Fix clippy failing on master

yannham:


yannham:
For posterity, I've been confused by the GH interface: you need to tick the box, but then proceed to list which checks are required. The list was empty by default.

vkleen (approved):

PR #1276, opened 2023-04-21
Fix query path parsing

### LINKED ISSUE #1249: Query doesn't properly handle quoted identifiers
**Describe the bug**
Trying to query a quoted identifier fails.

**To Reproduce**

```nickel
nickel> let config = {foo = 1}
nickel> :q config "foo"
error: query-path parse error: unexpected token while parsing a quoted identifier
  ┌─ query-path:1:5
  │
1 │ "foo"
  │     ^
```

**Expected behavior**

Query should handle quoted identifiers just fine.

yannham:
Closes #1249.

The parsing of a field path when querying metadata was broken (when using quotes). Instead of duplicating the parsing logic, this commit makes the query path code to piggy back on the actual Nickel parser to parse a field path.

### DIFF
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -58,7 +58,7 @@ impl Ident {
     pub fn label_quoted(&self) -> String {
         let reg = Regex::new("^_?[a-zA-Z][_a-zA-Z0-9-]*$").unwrap();
         let label = self.label();
-        if reg.is_match(label) && !KEYWORDS.contains(&&label) {
+        if reg.is_match(label) && !KEYWORDS.contains(&label) {
             String::from(label)
         } else {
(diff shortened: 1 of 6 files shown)


src/identifier.rs
```diff
@@ -39,14 +40,31 @@ impl Ident {
         Self::new_with_pos(label, TermPos::None)
     }
 
+    /// Create a new fresh identifier. This identifier is unique and is guaranteed not to collide
+    /// with any identifier defined before. Generated identifier starts with a special prefix that
```
vkleen, line 44:
```suggestion
    /// with any identifier defined before. Generated identifiers start with a special prefix that
```

vkleen (approved):

PR #1278, opened 2023-04-24
Use `transform` for the stdlib and `:load` in the REPL

### LINKED ISSUE #1248: `std.fail_with` fails (but not in the expected way)
**Describe the bug**
`std.fail_with` fails with an unexpected identifier over the use of `FailWith`. Well, it's expected to fail, but not in this way.

**To Reproduce**

```text
nickel> std.fail_with "foo"
error: unbound identifier `FailWith`
     ┌─ <stdlib/std.ncl>:2555:26
     │
2555 │     = fun msg => (null | FailWith msg),
     │                          ^^^^^^^^ this identifier is unbound
```

**Expected behavior**
`std.fail_with msg` should raise a contract error which includes the given message `msg`.

**Additional context**
The identifier isn't actually unbound (at least syntactically). I suspect that since we put everything under the std namespace, something was overlooked with respect to transformation, and the top-level record `std` doesn't get transformed, which breaks recursive records. Indeed, `FailWith` seems to be one of the rare occurrence of recursive cross-references at the top-level `std` module itself.

vkleen:
Restrict the use of `transform_inner` to only the internals module in the standard library. This enables recursive references and top-level let bindings in files loaded in the REPL by `:load` and in all other stdlib modules.

Closes #1248.

### DIFF
--- a/src/repl/mod.rs
+++ b/src/repl/mod.rs
@@ -428,7 +428,7 @@ pub fn print_help(out: &mut impl Write, arg: Option<&str>) -> std::io::Result<()
             Ok(c @ CommandType::Load) => {
                 writeln!(out, ":{c} <file>")?;
                 print_aliases(out, c)?;
-                writeln!(out,"Evaluate the content of <file> to a record and add its fields to the environment.")?;
+                writeln!(out, "Evaluate the content of <file> to a record and add its fields to the environment.")?;
                 writeln!(
                     out,
                     "Fail if the content of <file> doesn't evaluate to a record."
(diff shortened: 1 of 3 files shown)


src/repl/mod.rs
```diff
@@ -443,10 +428,10 @@ pub fn print_help(out: &mut impl Write, arg: Option<&str>) -> std::io::Result<()
             Ok(c @ CommandType::Load) => {
                 writeln!(out, ":{c} <file>")?;
                 print_aliases(out, c)?;
-                write!(out,"Evaluate the content of <file> to a record and load its attributes in the environment.")?;
+                writeln!(out,"Evaluate the content of <file> to a record and add its fields to the environment.")?;
```
yannham, line 431:
```suggestion
                writeln!(out, "Evaluate the content of <file> to a record and add its fields to the environment.")?;
```

yannham (approved):

PR #1285, opened 2023-04-25
Pass on error messages

yannham:
Pass on error messages. This PR attempts to:

- add missing information from error messages when required, or make important information to appear as high as possible (for example, repeating the name of the unbound identifier in the very first tagline `error: unbound identifier`, so that it is immediately visible without having to read the rest of the error message)
- removing useless/confusing information
- make error messages lighter (avoid heavy, passive formulations)
- use consistent jargon and formatting

This PR also gets rid of an error variant that wasn't used anymore since the the row types refactoring, and factor some error messages into small helpers to avoid repeating the same string literals over and over again.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1003,7 +1003,7 @@ impl IntoDiagnostics<FileId> for EvalError {
                         "Primitive values (Number, String, and Bool) or arrays can be merged \
                         only if they are equal."
                             .into(),
-                        "Functions can never be merged together.".into(),
+                        "Functions can never be merged.".into(),
                     ])]
             }
             EvalError::UnboundIdentifier(ident, span_opt) => vec![Diagnostic::error()
@@ -1737,7 +1737,7 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                 )
                 .with_message("this type variable is unbound")])
                 .with_notes(vec![format!(
-                    "Did you forget to put a `forall {ident}.` somewhere in the enclosing type ?"
+                    "Did you forget to put a `forall {ident}.` somewhere in the enclosing type?"
                 )])],
             TypecheckError::TypeMismatch(expd, actual, span_opt) => {
                 fn addendum(ty: &Types) -> &str {
@@ -1875,7 +1875,7 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                         diags.extend(err.into_diagnostics(files, stdlib_ids).into_iter().map(
                             |mut diag| {
                                 diag.message =
-                                    format!("While matching function types: {}", diag.message);
+                                    format!("while matching function types: {}", diag.message);
                                 diag
                             },
                         ));
@@ -1962,7 +1962,7 @@ impl IntoDiagnostics<FileId> for ExportError {
                 .with_message("non serializable term")
                 .with_labels(vec![primary_term(&rt, files)])
                 .with_notes(vec![
-                    "Nickel only support serlializing to and from strings, booleans, numbers, \
+                    "Nickel only supports serlializing to and from strings, booleans, numbers, \
                     enum tags, `null` (depending on the format), as well as records and arrays \
                     of serializable values."
                         .into(),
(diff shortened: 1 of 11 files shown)


src/error.rs
```diff
@@ -1041,8 +1052,7 @@ impl IntoDiagnostics<FileId> for EvalError {
 
                 vec![Diagnostic::error()
                     .with_message(format!("{format} parse error: {msg}"))
-                    .with_labels(labels)
-                    .with_notes(vec![String::from(INTERNAL_ERROR_MSG)])]
+                    .with_labels(labels)]
```
yannham, line 1055: This can very much happen, for example `std.deserialize `Json "{]"`, and is not an internal error. I wonder if the INTERAL_ERROR_MSG weren't just a leftover from copy pasting code around.

src/error.rs
```diff
@@ -991,12 +996,18 @@ impl IntoDiagnostics<FileId> for EvalError {
...
-                        "Both values have the same merge priority but they can't be combined",
-                    )])]
+                    .with_notes(vec![
+                        "Both values have the same merge priority but they can't \
+                        be combined."
+                            .into(),
+                        "Primitive values (Number, String, and Bool) or arrays can be merged \
+                        only if they are equal."
+                            .into(),
+                        "Functions can never be merged together.".into(),
```
vkleen, line 1006:
```suggestion
                        "Functions can never be merged.".into(),
```

src/error.rs
```diff
@@ -1640,59 +1662,83 @@ impl IntoDiagnostics<FileId> for TypecheckError {
...
+            TypecheckError::UnboundTypeVariable(ident) => vec![Diagnostic::error()
+                .with_message(format!("unbound type variable `{ident}`"))
+                .with_labels(vec![primary_alt(
+                    ident.pos.into_opt(),
+                    ident.to_string(),
+                    files,
+                )
+                .with_message("this type variable is unbound")])
+                .with_notes(vec![format!(
+                    "Did you forget to put a `forall {ident}.` somewhere in the enclosing type ?"
```
vkleen, line 1740:
```suggestion
                    "Did you forget to put a `forall {ident}.` somewhere in the enclosing type?"
```

src/error.rs
```diff
@@ -1833,32 +1872,33 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                     // information, so we just ignore it.
                     TypecheckError::TypeMismatch(_, _, _) => (),
                     err => {
-                        diags.extend(err.into_diagnostics(files, stdlib_ids).into_iter()
-                            .map(|mut diag| {
-                                diag.message = format!("While matching function types: {}", diag.message);
+                        diags.extend(err.into_diagnostics(files, stdlib_ids).into_iter().map(
+                            |mut diag| {
+                                diag.message =
+                                    format!("While matching function types: {}", diag.message);
```
vkleen, line 1878: I'm not sure how exactly this message is formatted, but maybe the "While" should be lowercase here as well, lik on line 1826.

src/error.rs
```diff
@@ -1909,26 +1949,38 @@ impl IntoDiagnostics<FileId> for ExportError {
...
             ExportError::UnsupportedNull(format, rt) => vec![Diagnostic::error()
-                .with_message(format!("{format} doesn't support null values"))
+                .with_message(format!("{format} format doesn't support null values"))
                 .with_labels(vec![primary_term(&rt, files)])],
             ExportError::NonSerializable(rt) => vec![Diagnostic::error()
                 .with_message("non serializable term")
-                .with_labels(vec![primary_term(&rt, files)])],
+                .with_labels(vec![primary_term(&rt, files)])
+                .with_notes(vec![
+                    "Nickel only support serlializing to and from strings, booleans, numbers, \
```
vkleen, line 1965:
```suggestion
                    "Nickel only supports serlializing to and from strings, booleans, numbers, \
```

vkleen (approved):

src/error.rs
yannham, line 1878:
```suggestion
                                    format!("while matching function types: {}", diag.message);
```

PR #1287, opened 2023-04-27
Keep track of field names in labels

vkleen:
Add a way of tracking field names for contracts that are attached to specific record fields. For example, this enables some blame errors to give the user a hint for which field was involved when a contract definition is too far away to see this in the codespan diagnostics.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1149,17 +1149,17 @@ mod blame_error {
             // a positive blame
             assert_eq!(l.polarity, Polarity::Positive);
             match l.field_name {
-                Some(ident) => format!("contract broken by the value of {ident}"),
+                Some(ident) => format!("contract broken by the value of `{ident}`"),
                 None => "contract broken by a value".to_owned(),
             }
         } else if l.polarity == Polarity::Positive {
             match l.field_name {
-                Some(ident) => format!("contract broken by the function {ident}"),
+                Some(ident) => format!("contract broken by the function `{ident}`"),
                 None => "contract broken by a function".to_owned(),
             }
         } else {
             match l.field_name {
-                Some(ident) => format!("contract broken by the caller of {ident}"),
+                Some(ident) => format!("contract broken by the caller of `{ident}`"),
                 None => "contract broken by the caller".to_owned(),
             }
         }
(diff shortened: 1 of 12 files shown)


src/error.rs
```diff
@@ -1143,16 +1143,25 @@ mod blame_error {
...
     /// of the label.
-    pub fn title(l: &label::Label) -> &str {
+    pub fn title(l: &label::Label) -> String {
         if ty_path::has_no_arrow(&l.path) {
             // An empty path or a path that contains only fields necessarily corresponds to
             // a positive blame
             assert_eq!(l.polarity, Polarity::Positive);
-            "contract broken by a value"
+            match l.field_name {
+                Some(ident) => format!("contract broken by the value of {ident}"),
```
yannham, line 1152: For consistency, other message errors tend to quote identifiers coming from Nickel code, to distinguish them from a word of the error message.

```suggestion
                Some(ident) => format!("contract broken by the value of `{ident}`"),
```

src/error.rs
```diff
@@ -1143,16 +1143,25 @@ mod blame_error {
...
+                None => "contract broken by a value".to_owned(),
+            }
         } else if l.polarity == Polarity::Positive {
-            "contract broken by a function"
+            match l.field_name {
+                Some(ident) => format!("contract broken by the function {ident}"),
```
yannham, line 1157:
```suggestion
                Some(ident) => format!("contract broken by the function `{ident}`"),
```

src/error.rs
```diff
@@ -1143,16 +1143,25 @@ mod blame_error {
...
+                None => "contract broken by a function".to_owned(),
+            }
         } else {
-            "contract broken by the caller"
+            match l.field_name {
+                Some(ident) => format!("contract broken by the caller of {ident}"),
```
yannham, line 1162:
```suggestion
                Some(ident) => format!("contract broken by the caller of `{ident}`"),
```

yannham (approved):
LGTM, but any reason to set this separately during the gen_pending__contracts phase? This looks like a static information that could (and maybe should) be set during parsing, such that generating contract doesn't have to care about this somehow separate issue.

vkleen:
I was blindly looking for the first place that `RuntimeContract`s get constructed, I think. But I agree that setting the field name information during parsing does make more sense.

Issue #329, opened 2021-03-18
Handle non-local dependencies


thufschmitt:
**Is your feature request related to a problem? Please describe.**

The only way in Nickel to load some external code is by hardcoding its path (relative or absolute) in the nickel file. This is fine to refer to files inside of a project, but doesn't scale as soon as someone wants to refer to an external nickel “library”

**Describe the solution you'd like**

Have a mechanism to reference “external files” in nickel.

I guess designing the final shape will require a bit of work, but right on top of my head I can think of several approaches

1. [Dhall-like semantics](https://github.com/dhall-lang/dhall-lang/blob/master/standard/imports.md) where some urls can be directly imported
2. Unix-style (also Jsonnet's) semantics, with a list of search paths (either provided on the cli with a `-I` argument, or via a `NICKEL_PATH` environment variable)
3. lockfile semantics, with a `nickel.lock` file (or whatever) that would specify how to map specific inputs to actual files. This lockfile could either be generated by nickel itself (but that seems quite out-of-scope for the language), or could have a public schema so that it could be generated by other tools (Nix, looking at you) or manually.

regnat:
> About the first one, do you mean to have sugaring for the imports? something like: import "./path/to/some.ncl" would be equivalent to ./path/to/some.ncl ?

I had in mind something a bit more involved, allowing for example `import "https://some.nickel.file.on/the/internet.ncl"`. But I’m actually not fond of it as it makes the language notably more complex and creates a whole class of issues.

> I suppose that the 3rd is not in the scope for a MVP

It might be as it can be reasonably simple (not really more complex than the 2d approach) and it’s the kind of things that might be worth setting early-enough to make it standard.

aspiwack:
I think that only option 4 makes sense today. We don't really have an identified need for anything more complex today. When we know more about how people use Nickle, we can revisit the question.

Ericson2314:
The Dhall experience seems much worse than NIx today, when it's really unclear what code needs network access. I do not recommend that.

Profpatsch:
Maybe we want to ping @Gabriel439 on this, since they have the most experience with non-local dependencies in a config language.

Profpatsch:
From what i gathered you really want a semantics where file `A` can import file `H1` on Host `H`, which in turn wants to import file `H2` on Host `H` or even File `I1` on Host `I`.

So you need both a more complex notion of what “relative” import means, as well as a concept of cross origin security (CORS).

[...]

PR #1293, opened 2023-05-03
Convert `import` to statement instead of pseudo-fun

YorikSar:
Currently `import` is treated as a very special function that only accepts literal string as its first argument. It has following downsides:

* Special handling of `import` is hidden. User can assume that its just a function while it's a special keyword. User might expect to be able to pass variables to it while it is only handled before typechecking and evaluation.
* We can't extend `import` functionality without introducing new keywords which is not backward-compatible.

This change makes `import` into another statement like `let` or `fun`, which means it cannot be confused with a function anymore. It also means that expressions like `import "foo.ncl" bar` and `import "foo.ncl" & {..}` are invalid, and `import` statement need to be put in parethesis: `(import "foo.ncl") bar`.

For more context, see discussion in https://github.com/tweag/nickel/issues/329#issuecomment-1531333491

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -40,6 +40,7 @@ Breaking changes
 - Stdlib `string.Stringingable` -> `string.Stringable` by @vkleen in https://github.com/tweag/nickel/pull/1180
 - Fix the type of `array.elem` by @yannham in https://github.com/tweag/nickel/pull/1223
 - Change the enum tag start delimiter from backtick to single-quote by @vkleen in https://github.com/tweag/nickel/pull/1279
+- `import` is now a statement, `import "foo.ncl" arg1 arg2` requires parenthesis now: `(import "foo.ncl") arg1 arg2`, see https://github.com/tweag/nickel/pull/1293
 
 Language features
 -----------------
--- a/benches/mantis/deploy-example.ncl
+++ b/benches/mantis/deploy-example.ncl
@@ -1,4 +1,4 @@
-import "deploy.ncl"
+(import "deploy.ncl")
   {
     namespace = "mantis-staging",
     job = "miner",
--- a/benches/mantis/deploy.ncl
+++ b/benches/mantis/deploy.ncl
@@ -67,7 +67,7 @@ fun vars =>
   }
   in
   let jobDefs = {
-    Mantis = import "jobs/mantis.ncl" params,
+    Mantis = (import "jobs/mantis.ncl") params,
   }
   in
   let revisions = {
--- a/doc/manual/merging.md
+++ b/doc/manual/merging.md
@@ -695,7 +695,7 @@ configuration, from which other fields are derived:
 `greeter` can then be customized without interfering with the final output:
 
 ```console
-$ nickel export <<< 'import "hello-service.ncl" & {greeter = "country"}'
+$ nickel export <<< '(import "hello-service.ncl") & {greeter = "country"}'
 {
   "systemd": {
     "services": {
--- a/src/parser/grammar.lalrpop
+++ b/src/parser/grammar.lalrpop
@@ -244,6 +244,7 @@ UniTerm: UniTerm = {
     <err: Error> => {
         UniTerm::from(err)
     },
+    "import" <s: StandardStaticString> => UniTerm::from(Term::Import(OsString::from(s))),
 };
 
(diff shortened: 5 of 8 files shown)


YorikSar:
Will do. There's also one spot I've missed in tests.

aspiwack:
I don't see why the added parentheses are needed, can you elaborate on this part?

YorikSar:
> I don't see why the added parentheses are needed, can you elaborate on this part?

Currently `import` is syntactically a function, which means every term after it is treated as another argument for function application. So for example `import "foo.ncl" from "somewhere"` that I’ve tried to implement in the linked issue is treated as passing 2 arguments to the result of the import. We need to somehow distinguish import functionality from just arguments and variables. We can do this by either adding more keywords to the grammar, but that’s not backward compatible. Another option - pull `import` up to the statement level, in that case it will need to be separated from the rest of expression, just like `let` or `fun`.

aspiwack:
Ah, I see, I misunderstood the context, but reading your original comment again, I see that this is entirely a parsing issue.  Specifically, you want to create more parsing rules around `import`, like `import "foo" from "bar"` and you're hitting a bit of a snag because it interferes with `import foo` binding exactly as tight as a function. So `import "foo" from "bar"` is parsed as passing three argument to `import` including `from`.

I don't know which is most valuable: to be able to use `import "foo" bar` as a function, or to have a somewhat elegant `import "foo" from "package"` (compared to something like `import-from "foo" "package"`). This is worth pondering. And now is kind of the right time, as we are reaching 1.0.

In a sense your proposal has the advantage of forward compatibility: if we go for your more restrictive parsing rule, then we can always change our minds later, and make `import "foo" bar` a thing again. But if we keep the current parsing rule, then implementing your change will be a breaking change.

YorikSar:
> I don't know which is most valuable: to be able to use `import "foo" bar` as a function, or to have a somewhat elegant `import "foo" from "package"` (compared to something like `import-from "foo" "package"`). This is worth pondering. And now is kind of the right time, as we are reaching 1.0.

Another benefit of this approach is that we don't pretend that import is a function even though in Nickel it's a language construct.

> In a sense your proposal has the advantage of forward compatibility: if we go for your more restrictive parsing rule, then we can always change our minds later, and make `import "foo" bar` a thing again. But if we keep the current parsing rule, then implementing your change will be a breaking change.

Exactly. We can work around this in nickel-nix right now, for example, but without this  change there would be no hope to add more functionality to `import` without adding keywords and breaking backward compatibility.

vkleen:
I'm not sure what the original reason was for not having the benchmark typechecking be configured in the flake checks. If there even is a reason :laughing:

YorikSar:
> I'm not sure what the original reason was for not having the benchmark typechecking be configured in the flake checks. If there even is a reason 😆

I guess, it was just easier. I see that there is `benchmark` check that runs `cargo bench --no-run`, but it doesn't include type checking them. Maybe we could hook type check into the benchmark building process...

aspiwack:
> Another benefit of this approach is that we don't pretend that import is a function even though in Nickel it's a language construct.

I really wouldn't use this terminology. “Parsing at the same level as functions” and “pretending that it's a function” are completely different things. And it's actually what caused me to be confused at first.

RELEASES.md
```diff
@@ -40,6 +40,7 @@ Breaking changes
 - Stdlib `string.Stringingable` -> `string.Stringable` by @vkleen in https://github.com/tweag/nickel/pull/1180
 - Fix the type of `array.elem` by @yannham in https://github.com/tweag/nickel/pull/1223
 - Change the enum tag start delimiter from backtick to single-quote by @vkleen in https://github.com/tweag/nickel/pull/1279
+- `import` is now a statement, `import "foo.ncl" arg1 arg2` requires parenthesis now: `(import "foo.ncl") arg1 arg2`, see https://github.com/tweag/nickel/pull/1293
```
yannham, line 43: It's a detail, but we should refrain from using statement I think. Everything is an expression in Nickel, including imports, which do have a value (as opposed to an if statement in e.g. C). It's not too bad in RELEASES.md, but just as a general note.

YorikSar: I agree. We should probably rephrase it in RELEASES as well, although I'm not sure how to properly put it to words.

PR #1296, opened 2023-05-05
Raise blame error when trying to access fields in the sealed tail of a polymorphic record

### LINKED ISSUE #952: Correctly assign blame when accessing a field in the tail of a polymorphically-sealed record
Currently accessing a field which is sealed into the tail of a polymorphically-sealed record raises a missing field error. However, given that this is essentially an attempt to violate the polymorphic record contract that sealed the record in the first place, it would be more appropriate to raise a blame error.

With the previous implementation this was not feasible, however once #949 is merged it should get a lot easier.

In practical terms, the following programs should fail with blame errors:

```
let f | forall a. { ; a} -> { ; a} = fun x => %seq% x.a x in f
```

```
let remove_x | forall r. { ; r } -> { ; r } = fun r => %record_remove% "x" r in
remove_x { x = 1 }
```

matthew-healy:
This updates `SealedTail` to store a `HashSet` of the identifiers which the tail contains. Then, when we hit a point where we'd otherwise raise a `FieldMissing` error, we check whether the field is in the sealed tail. If so, we instead raise a blame error.

This closes #952.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -143,15 +143,19 @@ pub enum EvalError {
 pub enum IllegalPolymorphicTailAction {
     Map,
     Merge,
+    StaticAccess { field: Ident },
 }
 
 impl IllegalPolymorphicTailAction {
-    fn message(&self) -> &'static str {
+    fn message(&self) -> String {
         use IllegalPolymorphicTailAction::*;
 
         match self {
-            Map => "cannot map over a record sealed by a polymorphic contract",
-            Merge => "cannot merge a record sealed by a polymorphic contract",
+            Map => "cannot map over a record sealed by a polymorphic contract".to_owned(),
+            Merge => "cannot merge a record sealed by a polymorphic contract".to_owned(),
+            StaticAccess { field } => {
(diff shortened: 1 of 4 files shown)


matthew-healy:
@vkleen Yeah, iirc when I discussed this with @yannham last year we decided this was likely to be a reasonable compromise initially. 

Though, to be honest, I was happier with this solution before it involved allocating a new `String` for every `Ident` in the tail. I can't really see any way to avoid doing that and still get `O(1)` containment checking, but maybe taking linear time to check if an element is in the tail would be okay since we only hit that codepath when we already know we're going to raise _some_ sort of error... :thinking: 

If so, then we could just store a `Vec<Ident>`, which avoids the extra string allocations.

vkleen:
Storing a `Vec<Ident>` actually sounds pretty reasonable. Generally, we don't care about the performance of the error path (within limits). Being able to avoid the `String` allocation should be worth that cost, I think.

matthew-healy:
Alright, sounds good. Thanks!

PR #1298, opened 2023-05-10
Add test infra to perform arbitrary checks on annotated ncl source files

matthew-healy:
This PR adds some testing infrastructure to make it easier to perform checks against Nickel source annotated with a toml-formatted header. It then uses this infra to rewrite the existing `integration`, `examples` and `snapshot` tests.

The overall machinery isn't too complex. There's a function, `read_annotated_test_case` which takes a path to a file in the repo, parses the annotation (in the form of an arbitrary deserializable type), and returns a `TestCase` struct containing the parsed annotation and the program source.

This means that to implement a new type of test, you just need to define a type to parse the annotation into, and then decide how to compare that to the result of the program. 

So, for example, in the integration tests, we define an `Expectation` type which has cases for success and errors. Each error annotation is parsed as an `ErrorExpectation`, which is intended to be broadly 1:1 with `nickel::error::Error`, but without the things that make that type really hard to serialize (e.g., call stacks). We then implement `PartialEq<Error>` for `ErrorExpectation`, which means we can use `assert_eq!` to compare the annotation to the actual result of the program.

The example and snapshot tests are similar, though both are slightly simpler, so it's not worth going into too much detail.

*Note*: I haven't yet finished converting all of the existing integration tests to this new format. However, I think it makes sense to have this reviewed & merged sooner rather than later to avoid conflicts.


vkleen:
I love that idea :) Could we maybe simplify the mess we created over time in the snapshot tests using the same approach?

matthew-healy:
@vkleen I haven't looked at the snapshot tests in a little while, but it's definitely possible! Are you thinking of annotating the file with the type of output we want to capture? (e.g. stdout vs stderr, etc?)

vkleen:
Yes, I'd like to be able to annotate with the captured output and with the nickel CLI command to use. For example, I briefly wanted to add snapshot testing for `nickel query`, but that also needs a specified query path in addition to the file.

matthew-healy:
It depends what exactly you mean by "captured output" here. It'd be very easy to have an annotation like...

```
# capture = 'stderr'
# command = 'query other-args etc'
```

But putting the actual output into the annotation would likely be harder without essentially rewriting part of `cargo-insta`. It's not a bad idea, but might not be worth the effort right now.

vkleen:
I was thinking precisely of your example 👍 The actual output will go into the snapshot files anyway and won't be written by hand.

matthew-healy:
@vkleen I added a quick PoC of how this would work for snapshot tests. I just did the `export` ones for now, and I still need to work out some details (specifically around test naming), but essentially we should be able to have a single test function for any snapshot with a very small amount of extra work (& sharing the parsing code with the integration tests).

Let me know what you think.

vkleen:
I think that looks pretty good 👍

vkleen (approved):

matthew-healy:
@yannham What do you think about merging this in its current state? Appreciate you might not have time to review the whole thing, but I'd feel more confident about merging knowing you'd at least given a thumbs up to the general approach. :grin:

yannham (commented):
I haven't reviewed things in detail, but I think the overall result is a net positive, and the format doesn't look too bloated or complex. I wonder how it will fare with more complex test though, like if you need to match errors more precisely or if you expect a specific thing to appear in the message. I would be wary of having our expectation language growing to become an kind of generic `matches!` command, which would be a mess. Maybe we can keep more complex tests separate and written in Rust? Those are things to think about. Anyway, for now, the first step look reasonable :+1:

matthew-healy:
@yannham Thanks for taking a look! 

Yeah, I think that could be reasonable. A quick check of our tests so far suggests that we almost always just match against the enum variant. The toml format can expand to do more complex matching, but I'm not against having special cases stay in Rust. I just want to make the "happy path" for testing as easy as possible.

PR #1299, opened 2023-05-11
Update the manual introduction

vkleen:
This fixes a typo in the introduction section and updates the contents section.

I'm not entirely sure about the formatting change for the design motto. As a blockquote it doesn't get rendered distinctly enough to convey the meaning, namely that the next paragraph is intended to explain the motto.


yannham (approved):

yannham:
I think the change is fine, as long as the motto itself is emphasized by any mean.

PR #1301, opened 2023-05-17
Convert tests to new annotated format, pt. 1

matthew-healy:
This PR converts a bunch of existing tests to the new annotated format introduced in #1298. 

In particular, it rewrites the tests from `contracts_fail`, and then updates every test that already existed as a `.ncl` file to use the new annotations.

The annotation format also changes a little bit, to accommodate additional top-level test configuration (in this case, a `repeat` field to allow tests to be rerun arbitrary numbers of times, to better catch non-deterministic behaviour).


vkleen (approved):

yannham (changes requested):
Nothing wrong with this PR, but we're potentially releasing 1.0 today, so let's avoid merging anything except a critical fix to minimize the risk of introducing issues

matthew-healy:
@yannham This doesn't touch anything outside of the `tests` directory, but it's totally your call. :+1:

yannham:
Yeah, I don't really think this PR could have any such impact, it's rather than I was in the middle of the release process (branching, running `nix flake check`), and I didn't want to risk a race condition. We can merge that soon hopefully :)

matthew-healy:
@yannham Could you approve this when you have a sec? I don't think I can merge while there are still changes requested.

yannham (approved):
Sorry, I wanted to take a deeper look first, but won't have enough time. So I haven't actually reviewed the code in detail, but let's not block this PR.

PR #1309, opened 2023-05-19
Migrate to clap

### LINKED ISSUE #1162: Migrate to clap
`structopt`, that we use for parsing the argument line, has been integrated into clap since version 3 and is in maintenance mode. We should simply use a recent version of clap directly.

kbknapp:
This PR migrates both `nls` and `nickel` from structopt to clap. There should be no semantic changes.

Closes #1162

### DIFF
--- a/src/bin/nickel.rs
+++ b/src/bin/nickel.rs
@@ -15,69 +15,68 @@ use std::{
     process,
 };
 // use std::ffi::OsStr;
+use clap::{Parser, Subcommand, ValueEnum};
 use directories::BaseDirs;
-use structopt::StructOpt;
 
 /// Command-line options and subcommands.
-#[derive(StructOpt, Debug)]
+#[derive(Parser, Debug)]
 /// The interpreter of the Nickel language.
 struct Opt {
     /// The input file. Standard input by default
-    #[structopt(short = "f", long, global = true, parse(from_os_str))]
+    #[arg(short, long, global = true)]
     file: Option<PathBuf>,
 
     #[cfg(debug_assertions)]
     /// Skips the standard library import. For debugging only. This does not affect REPL
-    #[structopt(long)]
+    #[arg(long)]
     nostdlib: bool,
 
     /// Coloring: auto, always, never.
-    #[structopt(long, global = true, case_insensitive = true, default_value = "auto")]
+    #[arg(long, global = true, value_enum, default_value_t = ColorOpt::Auto)]
     color: ColorOpt,
 
-    #[structopt(subcommand)]
+    #[command(subcommand)]
     command: Option<Command>,
 }
 
 /// Available subcommands.
-#[derive(StructOpt, Debug)]
+#[derive(Subcommand, Debug)]
 enum Command {
     /// Converts the parsed representation (AST) back to Nickel source code and prints it. Used for
     /// debugging purpose
     PprintAst {
         /// Performs code transformations before printing
-        #[structopt(long)]
+        #[arg(long)]
         transform: bool,
     },
     /// Exports the result to a different format
     Export {
-        /// Available formats: `raw, json, yaml, toml`. Default format: `json`.
-        #[structopt(long)]
-        format: Option<ExportFormat>,
+        #[arg(long, default_value_t = ExportFormat::default(), value_enum)]
+        format: ExportFormat,
+
         /// Output file. Standard output by default
-        #[structopt(short = "o", long)]
-        #[structopt(parse(from_os_str))]
+        #[arg(short, long)]
         output: Option<PathBuf>,
     },
     /// Prints the metadata attached to an attribute, given as a path
     Query {
         path: Option<String>,
-        #[structopt(long)]
+        #[arg(long)]
         doc: bool,
-        #[structopt(long)]
+        #[arg(long)]
         contract: bool,
-        #[structopt(long = "type")]
+        #[arg(long = "type")]
         types: bool,
-        #[structopt(long)]
+        #[arg(long)]
         default: bool,
-        #[structopt(long)]
+        #[arg(long)]
         value: bool,
     },
     /// Typechecks the program but do not run it
     Typecheck,
     /// Starts an REPL session
     Repl {
-        #[structopt(long)]
+        #[arg(long)]
         history_file: Option<PathBuf>,
     },
     /// Generates the documentation files for the specified nickel file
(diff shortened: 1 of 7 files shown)


src/program.rs
```diff
@@ -27,13 +27,14 @@ use crate::eval::cache::Cache as EvalCache;
...
 use crate::term::{record::Field, RichTerm};
+use clap::ValueEnum;
 use codespan::FileId;
 use codespan_reporting::term::termcolor::{Ansi, ColorChoice, StandardStream};
 use std::ffi::OsString;
 use std::io::{self, Cursor, Read};
 use std::result::Result;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum)]
```
matthew-healy, line 37: clap exposes a [`ColorChoice`](https://docs.rs/clap/latest/clap/enum.ColorChoice.html) type which does the same job as `ColorOpt`. we could possibly just use `ColorChoice` everywhere, though we need to provide mappings into `rustyline::config::ColorMode` and `termcolor::ColorChoice`, which we currently do with `From` impls, so it might also make sense to newtype it.

If nothing else, I'm pretty sure we can delete the `FromStr` impl below as it was only there for `structopt` compatibility.

matthew-healy (approved):
Thanks for doing this! I've left a few comments, but they don't necessarily need to block merging.

Let me know if you want to make the updates here, otherwise I'll merge this early next week.

src/bin/nickel.rs
```diff
@@ -15,69 +15,68 @@ use std::{
     process,
 };
 // use std::ffi::OsStr;
+use clap::{Parser, Subcommand, ValueEnum};
```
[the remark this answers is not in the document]
kbknapp: Absolutely! Done in fe49dbdb

src/program.rs
kbknapp: Done in 9af6fa7a

asymmetric:
Could you explain the tradeoff here? How does clap improve things compared to structopt? :pray:

src/bin/nickel.rs
```diff
@@ -16,68 +16,66 @@ use std::{
...
     #[cfg(debug_assertions)]
     /// Skips the standard library import. For debugging only. This does not affect REPL
-    #[structopt(long)]
+    #[arg(long)]
     nostdlib: bool,
 
     /// Coloring: auto, always, never.
-    #[structopt(long, global = true, case_insensitive = true, default_value = "auto")]
-    color: ColorOpt,
+    #[arg(long, global = true, value_enum, default_value_t = clap::ColorChoice::default())]
```
yannham, line 34: According to [the Clap documentation](https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes), you should be able to omit the value when it is equal to `OptionType::default()`

```suggestion
    #[arg(long, global = true, value_enum, default_value_t)]
```

src/bin/nickel.rs
```diff
@@ -16,68 +16,66 @@ use std::{
...
-        #[structopt(long)]
+        #[arg(long)]
         transform: bool,
     },
     /// Exports the result to a different format
     Export {
-        /// Available formats: `raw, json, yaml, toml`. Default format: `json`.
-        #[structopt(long)]
-        format: Option<ExportFormat>,
+        #[arg(long, default_value_t = ExportFormat::default(), value_enum)]
```
yannham, line 53:
```suggestion
        #[arg(long, default_value_t, value_enum)]
```

src/serialize.rs
```diff
@@ -21,7 +21,7 @@ use std::{fmt, io, rc::Rc, str::FromStr};
 
 /// Available export formats.
 // If you add or remove variants, remember to update the CLI docs in `src/bin/nickel.rs'
-#[derive(Copy, Clone, Eq, PartialEq, Debug, Default)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, clap::ValueEnum)]
```
yannham, line 24: I wonder, do we still need the `impl FromStr for ExportFormat`, if we now delegate parsing to clap? Or maybe we are parsing `ExportFormat` values outside of clap?

src/program.rs
```diff
@@ -34,22 +34,11 @@ use std::io::{self, Cursor, Read};
...
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum ColorOpt {
-    Auto,
-    Always,
-    Never,
-}
-
-impl std::str::FromStr for ColorOpt {
-    type Err = &'static str;
+pub struct ColorOpt(pub(crate) clap::ColorChoice);
```
yannham, line 37: I wonder if we should just get rid of `ColorOpt` entirely. Right now it's just wrapping `clap::ColorChoice`, and the inner value is public anyway (so we are tied to `clap::ColorChoice` for the public API already). If we don't implement any trait on it, I don't think there's any advantage in keeping the wrapper.

yannham (commented):
Thanks for tackling this! It looks great overall, I just left a bunch of comments wondering if we could simplify things even further. Once they're answered, happy to approve.

yannham:
@asymmetric cf https://github.com/tweag/nickel/issues/1162

src/program.rs
matthew-healy: Unfortunately we can't. Like I said [above](https://github.com/tweag/nickel/pull/1309#discussion_r1199591236), we need to implement `From<ColorOpt>` for two other libs' own color types, and we can't implement `From<Clap::ColorChoice>` instead due to orphan rules.

kbknapp:
> Could you explain the tradeoff here? How does clap improve things compared to structopt?

structopt used clap internally, providing the derive macros on top of clap. As of clap 3.x, it took the structopt codebase internally behind a `derive` feature. The goal was to have a more cohesive codebase and API. [Since that point, structopt has gone into maintainance mode.](https://github.com/TeXitoi/structopt#maintenance) clap continues to be actively developed.

> Thanks for tackling this! It looks great overall, I just left a bunch of comments wondering if we could simplify things even further. Once they're answered, happy to approve.

Sounds great, I'll address the individual comments and any associated commits/changes in those threads. My primary goal at the start of this PR was to change as little possible so as to not make too many waves on an initial contribution; however if y'all are happy to make changes, so am I 😄

src/bin/nickel.rs
kbknapp: Great catch! Fixed in 5b8919a1

src/bin/nickel.rs
kbknapp: Fixed in 5b8919a1

src/serialize.rs
kbknapp: Deriving `ValueEnum` implements `FromStr` for the type, so the manual implementation isn't needed even if you are using `ExportFormat`s outside of clap. The only exception would be if you want to do something more complex inside the manual `FromStr` implementation. However, the code does not so removing the manual implementation is perfectly fine.

Fixed in ded632fb

src/program.rs
yannham: Ah, I missed that. Thanks!

yannham (approved):

src/bin/nickel.rs
```diff
@@ -16,68 +16,66 @@ use std::{
...
-        #[structopt(long)]
+        #[arg(long)]
         transform: bool,
     },
     /// Exports the result to a different format
     Export {
-        /// Available formats: `raw, json, yaml, toml`. Default format: `json`.
-        #[structopt(long)]
-        format: Option<ExportFormat>,
+        #[arg(long, value_enum)]
```
vkleen, line 53:
```suggestion
        #[arg(long, value_enum, default_value_t)]
```

vkleen (approved):

PR #1320, opened 2023-05-25
Improve error messages for polymorphic tail parametricity violations

### LINKED ISSUE #1316: Poor error message in polymorphic tail type error
**Describe the bug**

The following program:

```
let f : forall r. (forall p. [| 'blo, 'ble ; r |] -> [| 'bla, 'bli; p |]) =
  match { 'blo => 'bla, 'ble => 'bli, _ => 'blo } 
in f 'bli
```

raises a type error:

```
error: incompatible types
  ┌─ repl-input-0:2:44
  │
2 │   match { 'blo => 'bla, 'ble => 'bli, _ => 'blo } 
  │                                            ^^^^ this expression
  │
  = Expected an expression of type `a`
  = Found an expression of type `[| 'blo ; _erows_a |]`
```

The type `a` here is pulled out of thin air. What the error is really trying to say is that this function body should be valid for _any_ tail `p`, including those that don't include the variant `'blo`. We should improve the error message in this specific case to make that clearer.

**Additional context**

Note that record tails exhibit similar behaviour. E.g. in the program:

```
let f : forall r p { blo: String, ble: Number ; r } -> { bla: String, bli: Number ; p } =
  fun { blo, ble, .. } =>
    { bla = blo, bli = ble }
in f { blo = "", ble = 0 }
```

where the error message is:

```
error: incompatible types
  ┌─ repl-input-3:3:5
  │
3 │     { bla = blo, bli = ble }
  │     ^^^^^^^^^^^^^^^^^^^^^^^^ this expression
  │
  = Expected an expression of type `a`
  = Found an expression of type `{  }`
```

matthew-healy:
Previously, if we had a type whose return value contained a polymorphic tail and we tried to return a value which was not statically known to be in the return type, we would see a type mismatch error with an "invented" type name (e.g. `a`) to represent "an arbitrary type in the tail".

This made it unclear where the problem was, and didn't give any real direction on how to fix the problem.

This commit raises a specific error in this case, which informs users that the type they've tried to return is not guaranteed to exist in the polymorphic tail. We also now pass around a var kind with type errors related to constant mismatches, so we can print a representation of the tail type (e.g. `[| ; x |]`) instead of inventing phantom type names for errors.

Fixes #1316.


vkleen (approved):
The general approach looks like a net improvement to me. I'm just not happy with the error message 😅 Values don't exist in a type, they have types or maybe they're compatible with types if subtyping is allowed. Maybe
```
"values of type {violating_type} are not guaranteed to be compatible with a polymorphic {tail_king} `{tail}`"
```

The record tail error message referencing the type `{ }` also feels a bit weird to me, but I don't think that's introduced in this PR.

matthew-healy:
@vkleen Totally happy to change the error message. I wasn't happy with it either, but figured it was better to just do _something_ & then bikeshed it in the PR. :grin: Your suggestion is much better.

PR #1322, opened 2023-05-25
Add flake output for static binaries on linux

vkleen:
This seems to work and produces statically linked Linux binaries. I don't know enough about Darwin to figure out how to build a redistributable executable there, unfortunately.

As a fun side effect, using the static binary in the Docker image and removing `bash` shrinks the image from 18M to 4.3M; just removing `bash` shrinks it to only 15M 😎

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -105,6 +105,10 @@
           [ ];
 
       mkRust =
+        let
+          inherit (pkgs.stdenv) hostPlatform;
+          inherit (pkgs.rust) toRustTarget;
+        in
         { rustProfile ? "minimal"
         , rustExtensions ? [
             "rust-src"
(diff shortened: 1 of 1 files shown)


jneem:
I don't feel particularly qualified to review this, but I can confirm that it builds a working static executable on linux...

flake.nix
```diff
@@ -357,10 +372,9 @@
         tag = version;
         contents = [
           nickel
-          pkgs.bashInteractive
         ];
         config = {
-          Cmd = "bash";
+          Cmd = "${nickel}/bin/nickel";
```
yannham, line 377: Should we use `Entrypoint` instead, for this kind of run-and-return CLI like Nickel?

yannham (approved):
Great to see worked on, so that we can ship static binaries of 1.0 soon. Approved with the caveat that I'm not really proficient in cross-compilation/static binaries building via Nix.

flake.nix
vkleen: It looks like `Entrypoint` is indeed the right choice, cf. https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact.

Issue #950, opened 2022-11-28
Polymorphic record contracts should know which fields their row variables cannot contain


matthew-healy:
Consider the following program:

```
let f | forall r. { ; r } -> { x: Num; r } = fun r => %record_insert% "x" r 1 in f { x = 0 }
```

We know statically that the `r` row variable cannot contain the field `x`, and we return an error to that effect if we typecheck this code (by changing the end to `(f { x = 0 } : _)`). 

However, after merging #949, the dynamic error for this is:

```
error: non mergeable terms
  ┌─ repl-input-1:1:78
  │
1 │ let f | forall r. { ; r } -> { x : Num; r } = fun r => %record_insert% "x" r 1 in (f { x = 0 })
  │                                                                              ^             ^ with this expression
  │                                                                              │              
  │                                                                              cannot merge this expression
  │
  = Both values have the same merge priority but they can't be combined
```

it should be possible to give a more detailed error message (e.g. "arguments to f cannot contain the field x"), by providing this information to the polymorphic record contract.

PR #1323, opened 2023-05-25
Better error message when contract makes certain rows illegal

Radvendii:
addresses #950

worked on with @jneem @matthew-healy 

After we finished working on it, I changed two things
1. We need to recurse into the type declarations inside a record e.g. `forall r. { x : { y : Number ; r } ; r }` makes `y` in `r` illegal.
2. I rewrote it to be imperative. This has three advantages I see:
   1. Stack size can't explode
   2. We only ever need two `HashSet`s
   3. Don't need to pass arguments around when they're right there and never change
   
   One disadvantage is that we *always* need two `HashSet`s, as well as `to_be_checked: Vec`. But maybe we could get rid of any allocations by moving things around a bit.
   The original recursive implementation is there if you just look at the first commit.

TODO for this PR:
- [x] decide between recursive and imperative implementation (or a secret third thing?)
- [x] write a test?
- [ ] make sure the error message is what we want
- [x] update function documentation where needed
- [x] decide where this code should live. It feels like this makes the `subcontract()` function rather long. Maybe we should move it to its own function. Or maybe not! I don't know what the norms are for this code base.


stdlib/internals.ncl
```diff
@@ -123,24 +123,36 @@
...
           (
             %label_with_message%
-              "extra field%{plural} `%{std.string.join ", " extra_fields}`"
+              "extra field%{plural extra_fields} `%{std.string.join ", " extra_fields}`"
               label
           )
     else
+      let conflicts = std.array.filter
+        (fun field => std.array.elem field constr)
+        (%fields% value)
```
vkleen, line 146: Using standard library functions in the internals module makes me a bit scared, because this can relatively easily lead to infinite recursions. Because most standard library functions carry type annotations, these types get converted to contract checks when they get called, which causes `subcontract` in the interpreter to get called. And if the constellation of types is slightly unlucky that ends up producing a recursion. I think in this case it would be fine, but `internals.ncl` is special in this way.

Even if it's ugly, we should inline the definitions of `filter` and `elem` here.

matthew-healy: This is a good point - I hadn't really considered this. There are 9 references to `std.` in this file. Do you think it's worth rewriting all of them?

vkleen: Except for the reference to `std.contract.Equal`, I think that would be a good idea.

matthew-healy: As discussed on Matrix ([thread](https://matrix.to/#/!CTCVXeJTucJbViJokz:matrix.org/$Cx3VUTcBEzOUIJQk2CQNddxsuYrcyya2wlm1xxkprQE?via=matrix.org&via=kleen.org&via=matrix.dapp.org.uk)), this is extra painful because `internals.ncl` _must_ currently be a syntactic record - i.e., it's not enough that it just evaluates to a record.

I proposed an alternative solution whereby we split standard library function implementations out into a module `std.unsafe`, and then re-export them in the existing modules with type & contract annotations provided on top. This means we can (safely) call `std.unsafe` functions from within `internals` without worrying about infinite recursion.

matthew-healy: I spiked the `std.unsafe` idea, and it's gonna be a bit more complex than I'd hoped (because, e.g., `std.unsafe.array.length` gets inferred as `Dyn` and then we can't unify that with `std.array.length: forall a. Array a -> Number`). It might be possible to get the sort of API we want using codegen to avoid duplicating definitions, but that's probably not something we should rush into.

As such, we have two options:
- Leave these calls as-is, and tackle every call to `std` in this file at the same time, or
- Fix these two callsites, and come back for the rest of the file once we've worked out a better solution.

I'm maybe leaning towards the first option, but I don't think I have strong feelings either way.

yannham: Another solution to this problem could be to have a special type annotation that doesn't generate a corresponding contract at runtime. I guess those would need to be forbidden outside of the stdlib (enforced by the parser), but that would allow to:

- write a single implementation that is statically typechecked
- export it either without any contract nor type (`std.unsafe`) or wrapped inside a type annotation (the current  `std` interface)

I don't know if that would work so well with mutually recursive functions, but at least self-contained ones such as `length` could be handled that way.

src/types.rs
```diff
@@ -956,12 +957,79 @@ impl Types {
...
+                        }
+                        _ => (),
+                    }
+                }
+
+                let constr_ncl: RichTerm = Term::Array(
+                    Array::from_iter(
+                        constr
+                            .into_iter()
+                            .map(|id| Term::Str(id.label().to_owned().into()).into()),
```
yannham, line 1021: I think there is a `From<Label>` instance for `NickelString`

```suggestion
                            .map(|id| Term::Str(id.into()).into()),
```

yannham (commented):
We need to compute those constraints as well during typechecking. This is probably not super costly, but it's a bit sad that we now duplicate the work in both `contract()` and `type_check()`. See `typecheck::ConstrainFreshRRowsVar` and its implementation.

Those so-called "lack predicate" constraints are purely syntactical. If we use them in many places, I wonder if we shouldn't compute them earlier. Parsing time is a good candidate (so that we don't have to first produce an incomplete AST with `Option`s that is then latter filled, but rather generate right away a complete structure). A `forall` variable would have a name and a kind as before, together with an additional `constraint` field or better named (`must_not_contain` or `must_lack` or `lacks` or whatever) field which is a `HashSet` or simply a `Vec`, as you suggest, because I expect the most common lengths for those constraints to be between 1 and 5.

src/types.rs
```diff
@@ -956,12 +957,79 @@ impl Types {
...
                 var_kind,
             } => {
-                use VarKind::*;
+                let mut constr: HashSet<Ident> = HashSet::new();
+                // TODO: optimization: might be faster if maybe_constr is a Vec, and since it only
+                // ever gets merged into `constr`, this won't affect semantics.
+                let mut maybe_constr: HashSet<Ident> = HashSet::new();
+                let mut to_be_checked: Vec<&Types> = vec![body];
+                while let Some(tys) = to_be_checked.pop() {
+                    match &tys.types {
```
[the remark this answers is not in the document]
Radvendii: This is a good thought. Hmm... I'm not sure if you meant something else by visitor pattern, but I tried implementing it as an iterator, and in order to allow the shadowed variable check, it had to look something like this:

```rust
pub struct TypesIterator<'a> {
    pub(crate) todo: Vec<&'a Types>,
    pub(crate) should_recurse: Option<Box<dyn Fn(&'a Types) -> bool>>,
}

impl<'a> Iterator for TypesIterator<'a> {
    type Item = &'a Types;

    fn next(&mut self) -> Option<Self::Item> {
        self.todo.pop().map(|tys| {
            let should_recurse = match &self.should_recurse {
                None => true,
                Some(f) => f(tys),
            };
            if should_recurse {
                match &tys.types {
                    TypeF::Arrow(s, t) => {
                        self.todo.push(s);
                        self.todo.push(t);
                    }
                    TypeF::Forall { body, .. } => {
                        self.todo.push(body);
                    }
                    TypeF::Record(rrows) => {
                        for ritem in rrows.iter() {
                            if let RecordRowsIteratorItem::Row(row) = ritem {
                                self.todo.push(row.types);
                            }
                        }
                    }
                    TypeF::Dict {
                        type_fields,
                        flavour,
                    } => {
                        // XXX: is this the right semantics? I'm not sure whether we should recurse
                        // into contracts or not
                        if flavour == &DictTypeFlavour::Type {
                            self.todo.push(type_fields);
                        }
                    }
                    TypeF::Array(t) => {
                        self.todo.push(t);
                    }
                    _ => (),
                }
            }
            tys
        })
    }
}

impl Types {
    pub fn iter(&self, should_recurse: Option<impl Fn(&Types) -> bool + 'static>) -> TypesIterator {
        TypesIterator {
            todo: vec![self],
            should_recurse: should_recurse.map(|x| Box::new(x) as Box<dyn Fn(&Types) -> bool>),
        }
    }
}
```

And then you call it like
```rust
                let var_ = var.clone();
                let ignore_shadowed = move |tys: &Types| match &tys.types {
                    TypeF::Forall { var, .. } if var == &var_ => false,
                    _ => true,
                };
                for tys in body.iter(Some(ignore_shadowed)) {
                    ...
                }
```

Advantages:
- clear separation of concerns

Problems:
- It's not clear that this is actually general enough. What if someone wants to only recurse into the domain of a function type for some reason?
- Dealing with these closure types seems really thorny. It took me a while to get the types right, and even then I'm not sure how to get it to work if you pass in `None`, because it needs a concrete type there.

One alternative is a much more specialized function, like `iter_forall(body: &Types, var: Ident)` which takes into account shadowing. But that feels really specific...

stdlib/internals.ncl
Radvendii: I made [this issue](https://github.com/tweag/nickel/issues/1334) for discussing this issue further. Can we resolve the thread on this PR and leave in `std.` for now?

vkleen: Since in this case we're not actually producing infinite recursions, I agree we should keep this as is in the interest of moving this PR forward 👍

Radvendii:
> We need to compute those constraints as well during typechecking. This is probably not super costly, but it's a bit sad that we now duplicate the work in both contract() and type_check(). See typecheck::ConstrainFreshRRowsVar and its implementation.
> Those so-called "lack predicate" constraints are purely syntactical. If we use them in many places, I wonder if we shouldn't compute them earlier. 

Yeah, we based this implementation on that one. It would be nice if they could be unified.

> Parsing time is a good candidate (so that we don't have to first produce an incomplete AST with Options that is then latter filled, but rather generate right away a complete structure). 

I think we'll have to wait until `fix_type_vars` phase, since we need to know the `var_kind` before this code runs. Maybe we can roll it into that phase though? It seems like a related action.

> A forall variable would have a name and a kind as before, together with an additional constraint field or better named (must_not_contain or must_lack or lacks or whatever) field

i.e. `TypeF::Forall { var, var_kind, body, lacks }`?

stdlib/internals.ncl
yannham: By the way, it's not only about infinite recursion, but for performances as well. Contract do have a non-trivial impact on performance, and this is also why we try to refrain from firing contracts from within contracts in a non obvious way. For example, I believe we're also using `%length%` instead of `array.length` although there's no risk of recursion. This can be reworked later, so let's not block this PR on this, but just to give a bit more context. `internals` are really low-level functions that are called implicitly, potentially quite a lot, by the interpreter, so we've been trying to keep them low level as well.

yannham:
> I think we'll have to wait until fix_type_vars phase, since we need to know the var_kind before this code runs. Maybe we can roll it into that phase though? It seems like a related action.

Yes, `fix_type_vars` is interleaved with parsing so I didn't think of it as a separate phase, but I was precisely thinking about this one.

> i.e. TypeF::Forall { var, var_kind, body, lacks }?

Either that, or putting the lacks predicate inside `var_kind`, because I believe `lacks` doesn't make sense for normal type variable, only for record and enum rows. Maybe even only for record rows.

Radvendii:
> Maybe even only for record rows.

Oh, you're right. We implemented it for enums but that doesn't really apply. Type checking verifies that 
```
let f : forall e. [| ; e |] -> [| 'x ; e |] = fun a => 'x in f 'x
```
works fine.

---

As an aside, I was poking around while testing that and found
```
let f : forall e. Bool -> [| ; e |] -> [| 'x; e |] = fun b a => if b then a else 'x in f 'x
```
Fails to type check, but will work as a contract (`s/:/|/`). Is that intended behaviour? It seems like`[| 'x |]` won't unify either with `[| 'x ; e |]` or with `[| ; e |]`, but it seems to me that both should work? Maybe this has to do with the polymorphic tail sealing?

yannham:
> Fails to type check, but will work as a contract (s/:/|/). Is that intended behaviour? It seems like[| 'x |] won't unify either with [| 'x ; e |] or with [| ; e |], but it seems to me that both should work? Maybe this has to do with the polymorphic tail sealing?

In general, when you introduce a polymorphic type variables (be it a normal type variable, or a row type variable), the variable can unify with _nothing_ (but itself) in the body of the function. For example:

```nickel
let f : forall a . Number -> a = fun x => x
```

will rightly fail, because the `forall a.` means "as a caller of `f`, I can pick absolutely any `a`, so you can't assume anything. But you're assuming `a = Number`". Technically, we introduce a so-called rigid type variable (or skolem variable, or type constant, etc.), which can only unifies with itself or a free unification variable.

Under this interpretation, it makes sense I think that `[| 'x |]` doesn't unify with `[| 'x ; e |]`, because you would arbitrarily choose that `e` is the empty row, but `e` is a universally quantified variable. That being said, there's this strange things with `enums` that because this is a sum type (coproduct), the tails has somehow a different meaning that for records. Records are products, so `{foo : T, bar : U; r}` means a value that must have field `foo : T` AND `bar : U` AND the tail `r`. An enum `[| 'x, 'y; e|]` is a enum tag which is either `'x` OR `'y` OR something in `e`. This means that morally, `[| 'x |]` is indeed a subtype of `[|'x ; e|]` for any tail `e`. Which is, I think, your intuition.

Subtyping is a whole can of worm, and the point of row polymorphism is to get the same kind of flexibility while staying in a purely classical polymorphic approach. We really don't want to introduce subtyping.

I believe there is an obscure primop which can widen a row type, called `embed_id` (for a fixed `id`, which is an enum tag like `'x`), with type `forall rows. [| ; rows |] -> [| 'id ; rows |]`. Unfortunately I think widening is sufficient to type your example, because we would need to widen the type of `[| 'x |]` with a full row `e`, and not just one `'id`.

I wonder if we can't simply infer open types for enum tags. That is, infer `'x` to have type `forall e. [| 'x; e|]` (in practice, `[| 'x; _a |]` with `_a` a unification variable). That might be sufficient to handle those cases.

src/types.rs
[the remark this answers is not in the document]
yannham: Yup, I was thinking of something like this. Even if it's not reused, I feel it separates concerns and makes the code more readable (the call site to this method is only concerned with the logic of the processing; if you want to see how the iteration work, you can then go to the definition of `iter()`).

> It's not clear that this is actually general enough. What if someone wants to only recurse into the domain of a function type for some reason?

This is a good question, which is hard to answer now. But, what we can do is to give a specific name to this function, make the documentation clear enough. If one day we need something more general, it shouldn't be too hard to do, and to replace the function you wrote by a specialized call to the general one.

> Dealing with these closure types seems really thorny. It took me a while to get the types right, and even then I'm not sure how to get it to work if you pass in None, because it needs a concrete type there.

I'm not sure to see why we need `'static` here, by the way. And we can probably do with `FnMut` instead of `Fn`. But, as far as the `Option` is concerned, I would just get rid of it. `None` is semantically equivalent to `|_| true`, which is really not that longer to write for the caller, while as you mentioned, it is annoying with generics (and require an additional pattern matching). I'm relatively confident that the optimized code will be as efficient with `|_| true` as with `None`, thanks to monorphization.

vkleen:
> I wonder if we can't simply infer open types for enum tags. That is, infer `'x` to have type `forall e. [| 'x; e|]` (in practice, `[| 'x; _a |]` with `_a` a unification variable). That might be sufficient to handle those cases.

When typechecking, I think we're already doing that. In `src/typecheck/mod.rs:1401` in the `check` function we have:

```rust
        Term::Enum(id) => {
            let row = state.table.fresh_erows_uvar();
            unify(state, &ctxt, ty, mk_uty_enum!(*id; row))
        }
```

yannham:
So @vkleen is right, my bad. We already do this. We looked at the example and deduced that the issue was actually the converse : the type of `a` is not general enough, because it doesn't include `'x`. The typechecker enforces that both branches of the `if-then-else` have the same type, one is `[| ; e |]` (`a`) and the other is `[| 'x; _t0 |]` for a free unification variable `_t0`. The unifier complains that `[| ; e |]` doesn't contain the row `'x`.

We can thus actually fix the example with the widening operator (this is an undocumented, prototype legacy - the syntax in particular is strange, as it takes an identifier as a parameter, but interpret it as a symbol, a bit like quoting in Lisp):

```nickel
> let f : forall e. Bool -> [| ; e |] -> [| 'x; e |] = fun b a => if b then (%embed% x a) else 'x in f true 'y
'y

```

`%embed% x` is the identity, but has type `forall a. [| ; a|] -> [| 'x ; a |]`. I honestly don't know if we can do much better currently. Starting to try to automatically apply widening sounds much like having subtyping. It's just a first impression though, we might want to take a second look at the literature on row types.

One possibility would be to fix the odd syntax (and name, probably) of `%embed%` and wrap it in the stdlib as a proper function, and document it. It's not ideal but would at least permit to write some functions that are actually well-typed but currently fails typechecking, like your example.

Radvendii:
In rebasing, I had to contend with https://github.com/tweag/nickel/commit/f8c8a02375289a1309d1fce7067a2cef40920055 which actually relies on `VarKind` just being a simple enum with no fields.

At first I was going to use `Discriminant<VarKind>` to get that with the new `VarKind` but there's no way to match on values of that type, so I made a new type `VarKindDiscriminant`. Not the prettiest code, but it works.

I think ultimately we may want to store the `excluded` information in a separate `HashSet` like we do at typechecking time, but create that earlier at parse time. That way (a) we don't need to store the `HashSet` pointer for all the other `VarKind`s, and (b) we don't have to create the `HashSet` later in typechecking phase.

src/typecheck/mod.rs
```diff
@@ -2237,6 +2228,7 @@ pub fn unify(
                     body: body2,
                 },
             ) if var_kind1 == var_kind2 => {
+                // XXX: ^^ do we want to compare std::mem::discriminant() or also check equality of `excluded`s above?
```
yannham, line 2231: At first sight, I would say that the discriminant is sufficient, because equality of the substituted types as checked below probably implies that the discriminant are the same.

On the other hand, we fail early and avoid any traversal or substitution if we can determine that the discriminants aren't the same directly :man_shrugging: . In the end it depends on what is the "happy path" and what we want to optimize for. I would say we want to optimize for when the type are equal, because otherwise we fail anyway, and traversing an additional type won't change much (you'll get your error message 0.010s later). On the other hand, if for some reason a lot of polymorphic types are unified, then you'll pay the price of checking the additional discriminants for each such unification.

Honestly, in any case, this check is dumbly slow (as per the comment), and shouldn't fire a lot (ideally, most of the unifications involving polymorphic types happen when types are already instantiated). It's probably fine either way.

src/typecheck/mod.rs
```diff
@@ -2898,43 +2702,8 @@ pub fn constr_unify_rrows(
...
-                }
-
-                Ok(())
-            }
-            _ => Ok(()),
-        }
-    } else {
-        Ok(())
-    }
-}
```
yannham, line 2915: Nothing to do with your PR, but this is a good example of when the git diff isn't very smart. I was trying to understand why you got rid of that part, but it's just an artifact of the next function being very similar and removed as well, makes it like your removed half this function and half the next one, instead of just the next one

yannham (approved):
Left a bunch of comments, but LGTM otherwise.

src/error.rs
```diff
@@ -1930,7 +1930,7 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                         .with_message("multiple rows declaration")
                         .with_labels(mk_expr_label(&span_opt))
                         .with_notes(vec![
-                            format!("Found an expression of a record type with the row `{}: {}`", ident, conflict.as_ref().cloned().unwrap()),
+                            format!("Found an expression of a record type `{}` with the row `{}`", conflict.as_ref().cloned().unwrap(), ident),
```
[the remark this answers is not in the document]
Radvendii: With the change it shows up like this:
```
error: multiple rows declaration
  ┌─ repl-input-4:1:89
  │
1 │ let f | forall r. { ; r } -> { x : Number; r } = fun r => %record_insert% "x" r 1 in (f { x = 0 }: _)
  │                                                                                         ^^^^^^^^^ this expression
  │
  = Found an expression of a record type `{ x: _a }` with the row `x`
  = But this type appears inside another row type, which already has a declaration for the field `x`
  = A type cannot have two conflicting declarations for the same row
```

src/typecheck/mod.rs
Radvendii: Oh, yes, thank you. I discussed this with Joe and we came to basically the same conclusion. I'd meant to take out the comment but forgot. Good to get confirmation anyways.

src/typecheck/mod.rs
Radvendii: Yup. Very annoying. I actually got confused by exactly the same part of the diff at some point.

src/parser/uniterm.rs
```diff
@@ -557,39 +541,48 @@ pub(super) struct VarKindMismatch;
...
+            s @ None => {
+                *s = Some(var_kind);
                 Ok(())
             }
-            VarKindCellData {
-                var_kind: data,
-                state: VarKindCellState::Set,
-            } if *data == var_kind => Ok(()),
-            _ => Err(VarKindMismatch),
+
```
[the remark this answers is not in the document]
Radvendii: Thanks! I'm always up for small nitpicks for the sake of consistency =)

src/error.rs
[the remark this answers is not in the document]
yannham: Ok, I'm not 100% sure of what really is `conflict` right now. It might actually be the tail of the original record type, but fixing an error message later on if we find edge cases is not very costly, so let's proceed with that :+1:

PR #1326, opened 2023-05-26
Give a better error message when trying to query a non-record

### LINKED ISSUE #1292: Trying to query something that isn't a record gives suboptimal error message
**Describe the bug**
When trying to query a value that isn't a record, either at top-level, or because you specify a field path that isn't valid, Nickel gives a dynamic type error that might point to some internals, in particular if the faulty value has a contract attached.

**To Reproduce**

```text
nickel>  let config | String -> {..} = fun value_param => { value = value_param }
nickel> :query config value
error: dynamic type error
   ┌─ <stdlib/internals.ncl>:24:7
   │
24 │       (fun x => %assume% codomain (%go_codom% label) (value (%assume% domain (%chng_pol% (%go_dom% label)) x)))
   │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this expression has type Function, but Record was expected
   │
   = query
```

**Expected behavior**
The query evaluation code should check by itself that the result has the right type before proceeding further, in order to show a tailored error message.

jneem:
The first commit is a simple bug fix: the error position should point to the un-evaluated term. This is consistent with what happens for later elements in the path.

The second commit adds a special error variant for querying fields of non-records.

Fixes #1292

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -133,6 +133,15 @@ pub enum EvalError {
     },
     /// A non-equatable term was compared for equality.
     EqError { eq_pos: TermPos, term: RichTerm },
+    /// Tried to query a field of something that wasn't a record.
+    QueryNonRecord {
+        /// Position of the original unevaluated expression.
+        pos: TermPos,
+        /// The identifier that we tried to query.
+        id: Ident,
+        /// Evaluated expression
+        value: RichTerm,
+    },
     /// An unexpected internal error.
     InternalError(String, TermPos),
     /// Errors occurring rarely enough to not deserve a dedicated variant.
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -1153,6 +1162,26 @@ impl IntoDiagnostics<FileId> for EvalError {
...
+                let label = format!(
+                    "tried to query field `{}`, but the expression has type {}",
+                    id,
+                    value
+                        .term
+                        .type_of()
+                        .unwrap_or_else(|| String::from("<unevaluated>")),
+                );
+
+                let labels = if let Some(span) = pos.into_opt() {
```
yannham, line 1175: (very) minor nitpick: if `vec![..]` is applied in both branches, you might as well factor it out, defining `let label = ...` as a single label and using `vec![label]` in `with_labels` below. It'll save a staggering number of 7 characters!

yannham (approved):
LGTM!

PR #1335, opened 2023-06-01
Documentation fixes

mkatychev:
* Moved `docs/tuto.md` to `docs/tutorial.md` hopefully this will fix the borken _Content_ footer in 
https://nickel-lang.org/user-manual/introduction to hopefully mirror the TOC here: https://github.com/tweag/nickel-lang.org/blob/1f309ec8868a83d15bb084037ea266f81c8ab2ea/gatsby-config.js#L25
* Fixed invalid command


doc/manual/introduction.md
```diff
@@ -93,7 +93,7 @@ The next steps we plan to work on are:
 This document is a detailed documentation on the main aspects of the language.
 It is composed of the following sections:
 
-1. [Tutorial](./tuto.md)
+1. [Tutorial](./tutorial.md)
```
mkatychev, line 96: this generates https://nickel-lang.org/user-manual/tuto which is an invalid link.

doc/manual/contracts.md
```diff
@@ -38,7 +38,7 @@ Here, `x` is bound to a `Number` contract. When evaluating `x`, the following st
 are performed:
 
 1. Evaluate `1 + 1`.
-2. Check that the result is a number.
+2. Check that the result is a `Number`.
```
yannham, line 41: Hmm, while your suggestion is equally valid, I don't think the original text is incorrect (or less correct) either. I think it's supposed to be a sentence in plain English, using the common word "number" on purpose.

While the difference is small, IMHO it's a bit less of a circular definition to say that

> `Number` check that something is a number

rather than

> `Number` checks that something is a `Number`

Typically, the contract was `Num` before 1.0, in which case the second wording would a bit more obscure.

```suggestion
2. Check that the result is a number.
```

yannham (changes requested):
Thanks for contributing! I'm not sure about the documentation change, but the tutorial renaming is surely better. I've changed the target branch to `stable`, such that it's picked up by the website. Note for maintainer: once merged, we'll have to cherry-pick this on master as well [edit: couldn't do that because this commit starts from current master which is already quite ahead of stable, so let's do the converse: cherry-pick this commit on `stable` after merging]

doc/manual/contracts.md
mkatychev: makes sense

yannham (approved):

doc/manual/tutorial.md
```diff
@@ -121,7 +121,7 @@ By default, nickel exports data as JSON. But we can change that using
 the extra parameter `--format yaml` to export as YAML.
 
 ```shell
-nickel -f users.ncl export --export yaml
+nickel -f users.ncl export --format yaml
 ```
```
mkatychev, line 125: @yannham  One change, I forgot to push, `--export` is not a valid flag.

yannham: Ah, good catch! Thanks

PR #1337, opened 2023-06-01
Refactor record type errors, and detect duplicate fields

### LINKED ISSUE #1284: Nickel happily accepts a record type literal with several declarations for the same row
**Describe the bug**
Nickel accepts ill-formed record types, where the same field is defined several time, while it should reject them.

**To Reproduce**

```
nickel> ({foo.bar.baz = "a"} : {foo : String, foo : Number})
error: type error: missing row `foo`
  ┌─ repl-input-2:1:2
  │
1 │ ({foo.bar.baz = "a"} : {foo : String, foo : Number})
  │  ^^^^^^^^^^^^^^^^^^^ this expression
  │
  = The type of the expression was expected to be { foo: String, foo: Number }, which contains the field `foo`
  = The type of the expression was inferred to be { foo: String }, which does not contain the field `foo`
```

**Expected behavior**
I expect the previous example to either fail with a specific parsing error, or to at least not consider this as a record type, but as a general record literal.

**Additional context**
This probably comes from the way `UniRecord` are converted to record types. Unirecords are just a list of declaration, with potential duplicates, but the check for record types just look for field with type annotation and without definitions. Is suspect it doesn't check for duplicate fields. This is wild guessing, and I haven't actually investigated the issue.

jneem:
This refactors record type literal errors to be more structured, and also adds a variant for checking duplicate field names.

 Fixes #1284.

### DIFF
--- a/src/parser/error.rs
+++ b/src/parser/error.rs
@@ -18,7 +18,7 @@ pub enum LexicalError {
 /// Error indicating that a construct is not allowed when trying to interpret an `UniRecord` as a
 /// record type in a strict way.
 ///
-/// See [`UniRecord::into_type_strict`].
+/// See [`UniRecord::into_type_strict`](crate::parser::uniterm::UniRecord::into_type_strict).
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum InvalidRecordTypeError {
     /// The record type had an invalid field, for example because it had a contract,
--- a/src/parser/error.rs
+++ b/src/parser/error.rs
@@ -1,3 +1,6 @@
+use codespan::FileId;
+use codespan_reporting::diagnostic::Label;
+
 use crate::{identifier::Ident, position::RawSpan};
 
 #[derive(Clone, PartialEq, Eq, Debug)]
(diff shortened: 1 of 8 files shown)


src/parser/error.rs
```diff
@@ -12,31 +15,79 @@ pub enum LexicalError {
...
+
+    pub fn notes(&self) -> Option<String> {
+        match self {
+            InvalidRecordTypeError::InvalidField(_) => Some(
+                "Value assignments such as `<field> = <expr>`, and metadata \
+                    annotation (annotation, documentation, etc.) are forbidden."
+                    .into(),
+            ),
+            InvalidRecordTypeError::InterpolatedField(_) => {
+                Some("String interpolation in field names is forbidden in record types".into())
```
yannham, line 67: As a side note, this one is a bit subtle to classify. In RFC002, we say something like "if something is a record type (in the old syntax), then it is parsed as a static type, otherwise it's a record literal". Given that interpolation is forbidden inside record types, taking this literally would mean that we shouldn't complain that "a record type has interpolation", but rather see this record as a record literal. However, we fail in this case too, because we forbid (since 1.0) type annotation on fields without definition.

So the semantics (as far as failing is considered) is the same, but not the error message. In the end I wonder if we shouldn't mention both cause, because we can't really know if the user wanted to write a record type with interpolation or a record literal with a field with a static type annotation but without a definition. Maybe adding a note corresponding to the latter could be sufficient? Something in the spirit of "and a field with a type annotation and without a definition is forbidden in record literals".

src/parser/uniterm.rs
```diff
@@ -508,12 +500,10 @@ impl TryFrom<UniRecord> for Types {
...
         if let Some((_, tail_pos)) = ur.tail {
             ur.into_type_strict()
-                .map_err(|InvalidRecordTypeError(illegal_pos)| {
-                    ParseError::InvalidUniRecord(
-                        illegal_pos.unwrap(),
-                        tail_pos.unwrap(),
-                        pos.unwrap(),
-                    )
+                .map_err(|cause| ParseError::InvalidRecordType {
+                    tail_span: Some(tail_pos.unwrap()),
```
yannham, line 504:
```suggestion
                    tail_span: tail_pos.into_opt(),
```

yannham (approved):
Besides comments, LGTM. It departs a bit from the status quo of centering all error-rendering code inside the `error` module, but it's still in an error module (`parser::error`), so it's fine IMHO (as long as we don't entangle e.g. parsing logic with error rendering).

src/parser/error.rs
jneem: I've updated the error message to explain why we interpreted the literal as a type (one of two reasons: presence of a tail, or all fields have type annotations only). This is also relevant for the duplicated field name error.

PR #1339, opened 2023-06-05
Add a GitHub action for automatically uploading release artifacts

vkleen:
The action is triggered on publishing a GitHub release and automaticlly builds the `dockerImage` and `nickel-static` flake outputs for x86_64 Linux. It then uploads the results as release artifacts on GitHub.

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -1,4 +1,4 @@
-name: Uplaod release artifacts
+name: Upload release artifacts
 on:
   release:
     types: [published]
@@ -42,7 +42,7 @@ jobs:
         with:
           extra_nix_config: "experimental-features = nix-command flakes"
           nix_path: "nixpkgs=channel:nixos-unstable"
-      - name: "Build docker image"
+      - name: "Build x86_64 static binary"
         run: |
           nix build --print-build-logs .#nickel-static
           cp ./result nickel-x86_64-linux
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -1,0 +1,54 @@
+name: Uplaod release artifacts
+on:
+  release:
+    types: [published]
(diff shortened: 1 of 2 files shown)


.github/workflows/release-artifacts.yaml
```diff
@@ -0,0 +1,54 @@
+name: Uplaod release artifacts
```
yannham, line 1:
```suggestion
name: Upload release artifacts
```

.github/workflows/release-artifacts.yaml
```diff
@@ -0,0 +1,54 @@
...
+    name: "Build Nickel release binary"
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: cachix/install-nix-action@v21
+        name: "Installing Nix"
+        with:
+          extra_nix_config: "experimental-features = nix-command flakes"
+          nix_path: "nixpkgs=channel:nixos-unstable"
+      - name: "Build docker image"
```
yannham, line 45:
```suggestion
      - name: "Build x86_64 static binary"
```

yannham (changes requested):
LGTM, but I think we have to fix the RELEASING steps. The GH action will use `master`, but given how the releasing steps are organized currently, the latest change (such as bumping version numbers, for example) aren't yet cherry-picked, so `master` isn't up-to-date. I think we have to move the very last section

> ### Port changes to master

and put it as the last step (6) of Prepare, such that at the end of Prepare, master features the new release.

vkleen:
I think the action should run on whatever commit got tagged by the release, so that should be the stable branch. But let me check that first.

yannham (approved):
Ah, my bad. In any case I'll approve this, and let you do the patch to `RELEASING.md` if you find out it's needed.

vkleen:
I've verified that the action runs against the tagged commit and I've added a few words to `RELEASING.md` on which branch to target when making a release.

PR #1201, opened 2023-03-26
flake.lock: Update

yannham:
Automated changes by the [update-flake-lock](https://github.com/DeterminateSystems/update-flake-lock) GitHub Action.

```
Flake lock file updates:

• Updated input 'crane':
    'github:ipetkov/crane/5291dd0aa7a52d607fc952763ef60714e4c881d4' (2023-03-07)
  → 'github:ipetkov/crane/35110cccf28823320f4fd697fcafcb5038683982' (2023-05-25)
• Updated input 'crane/flake-utils':
    'github:numtide/flake-utils/3db36a8b464d0c4532ba1c7dda728f4576d6d073' (2023-02-13)
  → 'github:numtide/flake-utils/cfacdce06f30d2b68473a46042957675eebb3401' (2023-04-11)
• Added input 'crane/flake-utils/systems':
    'github:nix-systems/default/da67096a3b9bf56a91d16901293e51ba5b49a27e' (2023-04-09)
• Updated input 'crane/rust-overlay':
    'github:oxalica/rust-overlay/e53e8853aa7b0688bc270e9e6a681d22e01cf299' (2023-03-03)
  → 'github:oxalica/rust-overlay/d59c3fa0cba8336e115b376c2d9e91053aa59e56' (2023-05-03)
• Updated input 'flake-utils':
    'github:numtide/flake-utils/93a2b84fc4b70d9e089d029deacc3583435c2ed6' (2023-03-15)
  → 'github:numtide/flake-utils/a1720a10a6cfe8234c0e93907ffe81be440f4cef' (2023-05-31)
• Added input 'flake-utils/systems':
    'github:nix-systems/default/da67096a3b9bf56a91d16901293e51ba5b49a27e' (2023-04-09)
• Updated input 'nixpkgs':
    'github:NixOS/nixpkgs/ac718d02867a84b42522a0ece52d841188208f2c' (2023-03-15)
  → 'github:NixOS/nixpkgs/e635192892f5abbc2289eaac3a73cdb249abaefd' (2023-06-01)
• Updated input 'pre-commit-hooks':
    'github:cachix/pre-commit-hooks.nix/32b1dbedfd77892a6e375737ef04d8efba634e9e' (2023-03-16)
  → 'github:cachix/pre-commit-hooks.nix/ca2fdbf3edda2a38140184da6381d49f8206eaf4' (2023-05-29)
• Updated input 'rust-overlay':
[...]

PR #1343, opened 2023-06-06
flake: Explicitly set wasm-pack cache dir

YorikSar:
It tries to create it in home directory that doesn't exist in sandbox and build fails with this:

```
nickel-lang> ++ command cargo --version
nickel-lang> cargo 1.70.0 (ec8a8a0ca 2023-04-25)
nickel-lang> Error: Permission denied (os error 13)
nickel-lang> Caused by: Permission denied (os error 13)
```

Fix failure in #1201.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -348,7 +348,7 @@
           inherit cargoArtifacts src;
 
           buildPhaseCargoCommand = ''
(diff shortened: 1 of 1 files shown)


yannham (approved):
Nice catch! Thanks again for investigating this issue

PR #1346, opened 2023-06-07
Add instructions for building ARM64 artifacts to RELEASING.md

vkleen:
Since we don't have an ARM64 GitHub Actions runner handy, this PR adds some instructions for manually building ARM64 release artifacts to `RELEASING.md`.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -81,12 +81,23 @@ crates and dependent repositories (such as the website) in a consistent state.
 
 ### Release on GitHub
 
-1. Do the [release on
+1. Build a Docker image and a static binary for ARM64 manually, using
+
+   ```console
+   nix build --out-link nickel-arm64-docker-image.tar.gz .#packages.aarch64-linux.dockerImage
+   nix build --out-link nickel-arm64-linux .#packages.aarch64-linux.dockerImage
+   ```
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -81,12 +81,23 @@ crates and dependent repositories (such as the website) in a consistent state.
...
 
-1. Do the [release on
+1. Build a Docker image and a static binary for ARM64 manually, using
+
+   ```console
+   nix build --out-link nickel-arm64-docker-image.tar.gz .#packages.aarch64-linux.dockerImage
+   nix build --out-link nickel-arm64-linux .#packages.aarch64-linux.dockerImage
+   ```
+
+   on an ARM64 Linux machine.
```
yannham, line 91: Nitpick: I don't know if it's the right suggestion, but it felt a bit dry to have this new very specific step without much context (why ARM64 only?), although it's explained later.

```suggestion
   on an ARM64 Linux machine (x86_64 assets are automatically handled by a workflow, see 3).
```

yannham (approved):

RELEASING.md
vkleen: I like this suggestion 👍

PR #1347, opened 2023-06-07
Upload docker image to GHCR in release workflow

YorikSar:
Also add labels recognised by GHCR to the image.


YorikSar:
@yannham @vkleen I couldn't make it work while running from a branch, I'm getting `denied: installation not allowed to Write organization package` error (latest one [here](https://github.com/tweag/nickel/actions/runs/5355624784/jobs/9714126496)). I'm out of ideas now, the only one that I would like to try is running this from `master` branch that might affect these permissions. If you're comfortable with this, we could merge this and trigger it from the `master` to see if it works, if it does not, we can revert the part that builds and pushes the image later.

YorikSar:
I think I've tried this already. Let me try it again...

YorikSar:
@yannham Oh, wow. I feed stupid now. I was sure I tried that one already. Now it worked fine! https://github.com/tweag/nickel/pkgs/container/nickel has 1.0.0 now :)
The static build failed here https://github.com/tweag/nickel/actions/runs/5356809612/jobs/9716798931 because 1.0.0 doesn't have necessary output, but that's expected.

YorikSar:
Posted a PR with update to docs: https://github.com/tweag/nickel-lang.org/pull/497, rendered: https://6495aa2ce15b4e3e7be55bc1--nickel-lang.netlify.app/getting-started#build-using-docker

yannham (approved):
LGTM. Thanks!

PR #1351, opened 2023-06-09
Move the CLI into its own crate

vkleen:
Move the Nickel CLI into its own crate `nickel-lang-cli`. This will make it easier to integrate external tools such as [Topiary](https://github.com/tweag/topiary/) without incurring a ton of dependencies in the `nickel-lang` crate.

In the future, we may want to split out the REPL handling from `nickel-lang` into the CLI, and the WASM interface. But for now, this PR is just a straightforward move.


cli/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+authors = ["The Nickel Team <nickel-lang@protonmail.com>"]
+description = "Common helper functions for the tests and benchmarks of the nickel-lang crate."
+edition = "2021"
+
+[[bin]]
+name = "nickel"
+path = "bin/nickel.rs"
+bench = false
+
+[lib]
```
yannham, line 13: Do we need a lib here, if it's only the binary?

flake.nix
```diff
@@ -222,15 +222,16 @@
             } // extraArgs);
         in
         rec {
-          nickel = buildPackage { pname = "nickel-lang"; };
+          nickel-lang = buildPackage { pname = "nickel-lang"; };
+          nickel = buildPackage { pname = "nickel-lang-cli"; };
```
yannham, line 226: Just to avoid confusion

```suggestion
          nickel-lang-cli = buildPackage { pname = "nickel-lang-cli"; };
```

yannham (commented):
I think there several things to consider. How that will play with the next update? For example, people updating the `nickel-lang` package via `cargo`, thinking they will just have the latest version of the binary, will find an error saying something like "there's no binary available" I guess? I wonder how we could smooth the transition.

Second, It's not horrible but it's a bit sad to split the test in two (mainly because the source code controlling error messages are not in the CLI crate currently, but thinking about it, maybe they should just be? I don't really know the answer at this point)

cli/Cargo.toml
vkleen: That was mostly for future topiary integration where I didn't want to continue cluttering `bin/nickel.rs`.

vkleen:
I like the idea of `nickel-lang` living in its own subdirectory as well. The last commit does that and fixes the associated test failures. I'll have a look at the READMEs next.

yannham:
@vkleen is it ready for review? Last time I checked there were still some mixed up stuff with `nickel-lang`, `nickel-lang-cli` and `nickel-lang-lib`

vkleen:
I think it's ready for review now. Hopefully I've found all the mixups 😅

utilities/src/project_root.rs
```diff
@@ -0,0 +1,11 @@
+use std::path::PathBuf;
+
+/// NOTE: this is a workaround for the lack of a CARGO_WORKSPACE_DIR
+///       environment variable, as suggested [here](https://github.com/rust-lang/cargo/issues/3946#issuecomment-1433384192).
+///       A better workaround might be to set this in the `[env]` section
+///       of `.cargo/config.toml`, as suggested[here](https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993)
+///       but we currently don't check that file in, and I'm not sure
+///       whether doing so is a good idea.
```
yannham, line 8: Nitpick: it's a bit unadapted to use a docstring `///` "NOTE". Maybe we could first write a short description of what this function does, as normal documentation ("return the location of the workspace"), and then add the note.

yannham (approved):
I'm still not 100% sure about the `nickel-lang` vs `nickel-lang-cli` :sweat_smile: but I guess if we want to change that for the next release, it's just a small search and replace, isn't it? It's better to proceed with the new file hierarchy to avoid painful rebasing.

flake.nix
```diff
@@ -238,23 +242,23 @@
               };
 
           benchmarks = craneLib.mkCargoDerivation {
-            inherit src cargoArtifacts;
+            inherit src version cargoArtifacts;
 
-            pnameSuffix = "-bench";
+            pname = "nickel-lang-bench";
```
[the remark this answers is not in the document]
vkleen: That should be fine, we're overriding the build command to do the right thing. As far as I can tell, it's only our `buildPackage` wrapper function that use the `pname` to target a crate.

vkleen:
> I'm still not 100% sure about the `nickel-lang` vs `nickel-lang-cli` sweat_smile but I guess if we want to change that for the next release, it's just a small search and replace, isn't it? It's better to proceed with the new file hierarchy to avoid painful rebasing.

It's not quite a search and replace, but we're not using the string "nickel-lang" in all too many places, so it's not very painful. I agree that it's better to proceed as is and switch the names around later if we feel like it.

HACKING.md
```diff
@@ -13,11 +13,12 @@ long.
...
 
-The Nickel repository consist in 3 crates:
+The Nickel repository consist of 3 crates:
 
-- `nickel-lang` (path: `.`). The main crate containing the interpreter as a
-  library as well as the `nickel` binary.
-- `nickel-lang-lsp` (path: `lsp/nls/`). the Nickel Language Server (NLS), an LSP
+- `nickel-lang-lib` (path: `nickel-lang-lib`). The main crate containing the interpreter
+  as a library.
+- `nickel-lang` (path: `nickel-lang-cli`). The `nickel` binary.
```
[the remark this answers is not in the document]
vkleen: At some point, I had the CLI just under `cli`, like the `utilities` subdirectory now. I don't think there's any paritcular necessity to tie the directory name to the crate name, and just `nickel-lang` felt a bit misleading as you say.

PR #1358, opened 2023-06-13
Encode more pre-conditions of stdlib functions as additional contracts

### LINKED ISSUE #1354: Function error messages are unhelpful
- The failures are reported from nickel std internals
- There is no stack trace to figure out the full context that led to a failure

Here are some examples from the Nickel repl, but the issue is the same when evaluating nickel code from a file:

```
nickel> std.array.last []
error: elem_at expects its second argument to be a positive integer smaller than 18446744073709551615, got -1
   ┌─ <stdlib/std.ncl>:83:22
   │
83 │       = fun array => %elem_at% array (%length% array - 1),
   │                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here

```

```
nickel> std.array.drop_first []
error: array_slice: index out of bounds. Expected `start <= end <= 0`, but got `start=1` and `end=0`.
   ┌─ <stdlib/std.ncl>:97:22
   │
97 │       = fun array => %array_slice% 1 (%length% array) array,
   │                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here
```

```
nickel> std.array.at 1 []
error: elem_at: index out of bounds. Expected an index between 0 and 0, got 1
    ┌─ <stdlib/std.ncl>:154:20
    │
154 │       = fun n l => %elem_at% l n,
    │                    ^^^^^^^^^^^^^ here
```

yannham:
Fixes #1354. WIP

## Context

This pull request tries to fix a current defect of the stdlib functions wrapping primitive operations. We could expect those functions to encode all the required pre-conditions using contracts, for example that the argument of `std.array.first` is not only an array, but a non empty one.

However, that would make them hard to use inside statically typed code. For example, if we change the type of `first` from `forall a. Array a -> a` to `forall a. NonEmptyArray a -> a`, now arguments must have an opaque type that the typechecker doesn't understand. Meaning that each call site would need to explicitly convert the argument to `NonEmptyArray` using a contract check, which is both verbose and more costly.

Instead, we chose to have a static type annotation that is a bit less restrictive, but which is understood properly by the typechecker. If you think about it, the same dilemma happens for other statically typed language: Haskell or OCaml could have made their version of `first` such that it takes a custom type as an argument (a smart constructor). They chose not to, because that would make for a lot of wrapping and unwrapping, as well as more runtime checks. Instead, `first` has the same type as in Nickel and raise an exception at runtime if the argument is empty.

## Issue

As mentioned in #1354, the issue is that error messages for e.g. passing an empty array to `first` are really bad, because they don't have any relation to the call site of `first`. They merely point to inside the definition of `first`. Indeed, those are bare-bone dynamic type error from the corresponding Nickel primitive operation, and they don't carry much context.

## Idea

The idea developed in this PR is to add additional contract annotations which only check for those additional pre-conditions that aren't already enforced by the static type annotation (if any). For example, `first` is now:

```nickel
    first
      : forall a. Array a -> a
      | NonEmpty -> Dyn
```

The trick is that, in presence of multiple annotations, the typechecker has to select one to use for the static typechecking phase. The choice is to take the static type annotation in priority, if there's any, or to take the first contract annotation. Thus, with respect to static type checking, the new added annotation `NonEmpty -> Dyn` _doesn't change anything_, for both `first` and the consumers of `first`.

However, this contract annotation will fire at runtime, providing way better error messages (as well as better localized).

## Limits

One pain point of the current PR is that we had to define non trivial dependent function contracts (when the contract for one argument depends on the _value_ of another argument). It's not optimal, but still relatively fine as far as error reporting is concerned. It's just not easily re-usable for user-defined functions in a stable way. I think this mandate a proposal for a built-in syntax for dependent function contracts, but it's a separate matter. For now, we've put all those new ad-hoc depending function contracts and their specialized version into `contract.unstable`, with warning everywhere in the doc that those are just a tool for error reporting but aren't a stable part of the stdlib API.

## TODO

- [x] more generic dependent function contract
- [x] apply the dependent function contract to more stdlib functions
- [x] customize error messages for dependent function contracts
- [x] document somewhere our approach of static type + "differential" contract


stdlib/std.ncl
```diff
@@ -2284,7 +2559,8 @@
       = from,
 
     from_enum
-      | std.enum.Tag -> String
+      | forall a. [|; a |] -> String
```
silverraven691, line 2562: Was this meant to be the type rather than a contract?

stdlib/std.ncl
```diff
@@ -1103,6 +1130,254 @@
         "%
       = fun contract label value =>
         %assume% contract (%label_push_diag% label) value,
+
+    unstable
+      | doc m%"
+          The unstable module gathers contracts that are used right now in the
+          standard library to check additiona pre-condition, but aren't yet
```
silverraven691, line 1137:
```suggestion
          standard library to check additional pre-condition, but aren't yet
```

stdlib/std.ncl
```diff
@@ -2326,7 +2603,8 @@
       = fun s => %num_from_str% s,
 
     to_bool
-      | BoolLiteral -> Bool
+      | String -> Bool
```
silverraven691, line 2606: Same here, was this meant to be a type instead of a contract?

tests/snapshot/snapshots/snapshot__error_stderr_array_at_empty_array.ncl.snap
```diff
@@ -0,0 +1,29 @@
+---
+source: tests/snapshot/main.rs
+expression: err
+---
+error: contract broken by the caller of `at`: invalid array indexing
+    ┌─ <stdlib/std.ncl>:162:9
+    │
+162 │       | std.contract.unstable.IndexedArrayFun
+    │         ------------------------------------- expected type
+    │
```
silverraven691, line 10: This block is given a lot of importance visually in my opinion, but the information I get from it is kind of not terribly helpful because I don't know what `std.contract.unstable.IndexedArrayFun` is or means. The blame message shown ("Can't index into an empty array") is very helpful but my eyes don't jump to it immediately. Maybe it looks different with colors.
Either way, this isn't a blocker, this is a good improvement already.

stdlib/std.ncl
yannham: It looks like it, but it's not. Ideally that would be the case but the implementation is using `from` (which is just a wrapper around `%to_str%`), which has a type that is too generic. However, it has the same effect as being a type annotation for consumers of the function, because when several annotations are present, the first one is selected as the "official" type.

You can see it as a simpler/more efficient way of writing something like:

```nickel
from_enum
      : forall a. [|; a |] -> String
      | std.enum.Tag -> Dyn
      | doc m%"
        Converts an enum variant to its string representation.

        # Examples

        ```nickel
        std.string.from_enum 'MyEnum
          => "MyEnum"
        ```
      "%
      = from | forall b. [|; b|] -> String,
```

But maybe it's confusing, when you read the interface of the function...

stdlib/std.ncl
```diff
@@ -2312,7 +2588,8 @@
       = from,
 
     to_number
-      | NumberLiteral -> Number
+      | String -> Number
```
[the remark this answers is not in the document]
yannham: This one is a genuine overlook, though, thanks!

stdlib/std.ncl
yannham: This one as well, yes.

tests/snapshot/snapshots/snapshot__error_stderr_array_at_empty_array.ncl.snap
yannham: One solution would be to set the "Can't index into an empty array" at the top with the main label ("invalid array index"), although in general I try to keep the main label more generic and the notes more specific. But I agree with you.

I don't have a great solution right now on the top of my head: underlining the failing contract is very useful in most cases, and it's a generic process (it's not controlled by the contract implementation). I wanted to open an issue to have a dedicated syntax for dependent function contracts; maybe in that case, we wouldn't have those kind of "opaque" contracts anymore, and the underlining would be more useful. You can imagine something like:

```
Number ->dep contract.IndexedArray -> Dyn
```

with `contract.IndexedArray` being underlined.

stdlib/std.ncl
```diff
@@ -30,6 +30,20 @@
 #
 # On the other hand, the `std` module in the environment is properly typed, as
 # the typed interface is elaborated upfront.
+#
+# ## Both a type annotation and a contract annotation
+#
+# [^type-and-contract]: Some functions have both a static type annotation and
+# additional contract. The additional contract annotation is ignored by
```
vkleen, line 37:
```suggestion
# an additional contract. The additional contract annotation is ignored by
```

stdlib/std.ncl
```diff
@@ -1103,6 +1130,254 @@
...
+            "%
+          =
+            # capture the `contract.label` module to avoid name collisions with
+            # local contract argument `label`
+            let label_module = label in
+            let attach_message = label_module.with_message "invalid range" in
+
+            fun Codomain =>
+              let RangeSecond = fun start label value =>
+                if %typeof% start == 'Number && %typeof% value == 'Number then
```
vkleen, line 1228: Do we need this check if we're relying on an additional static type annotation anyway?

stdlib/std.ncl
```diff
@@ -1103,6 +1130,254 @@
...
+              passes.
+            "%
+          =
+            # capture the `contract.label` module to avoid name collisions with
+            # local contract argument `label`
+            let label_module = label in
+            let attach_message = label_module.with_message "invalid array indexing" in
+
+            let ArrayIndexFirst = fun label value =>
+              if %typeof% value == 'Number then
```
vkleen, line 1269: Same question here, do we need this `typeof` check?

stdlib/std.ncl
```diff
@@ -1103,6 +1130,254 @@
...
+                  |> attach_message
+                  |> label_module.append_note "Expected array index to be a positive integer, got %{%to_str% value} "
+                in
+                std.contract.apply std.number.Nat label value
+              else
+                value
+            in
+
+            let ArrayIndexSecond = fun min_size label value =>
+              if %typeof% min_size == 'Number && %typeof% value == 'Array then
```
vkleen, line 1281: Here as well, do we need the `typeof` check?

vkleen (approved):

stdlib/std.ncl
yannham: Yes, because of the way function contracts work, the order of checks for the arguments are actually reversed:

```nickel
foo
  | Domain1 -> Codomain1
  | Domain2 -> Codomain2
```

which will be elaborated as (more or less)

```
let first_app = fun x =>
  Codomain1 (f (Domain1 x))
in
first_app | Domain2 -> Codomain2

===>

fun x =>
  Codomain2 (first_app (Domain2 x))
```

So, the contracts are applied in that order, because the second function contract is the outer one:

- `Domain2` to `x`
- `Domain1` to `x`
- `Codomain1` to `f x`
- `Codomain2` to `f x`

So, when the new additional contracts are being applied, the static type annotation hasn't kicked in yet. This was observed in practice as well during testing. The additional contract thus have to yield the control flow to the other one, without failing, and without performing any operation that could cause another primop error (like trying to compare a number and a string).

vkleen: Oh, I see. Thanks for explaining. This might be subtle enough to deserve an explanation somewhere canonical, maybe the manual.

yannham: Do you mean about applying several function contract in general, or about the "differential approach" specifically?

vkleen: It's about the evaluation order of several function contracts applied to a field. That has implications about this differential approach, namely what can be relied on by which contract, but I feel it's a more general point to make.

yannham: Ok. I'll proceed with this PR but open a corresponding issue.

PR #1371, opened 2023-06-16
Topiary integration as `nickel format`

vkleen:


cli/Cargo.toml
```diff
@@ -27,6 +28,11 @@ clap = { workspace = true, features = ["derive"] }
 serde = { workspace = true, features = ["derive"] }
 directories.workspace = true
 
+topiary = { workspace = true, optional = true}
+tree-sitter-nickel = { workspace = true, optional = true }
+tempfile.workspace = true
+thiserror.workspace = true
```
yannham, line 34: Why do we need to pull those dependencies non-optionally?

flake.nix
```diff
@@ -225,21 +227,24 @@
...
+              cargoExtraArgs = "${cargoBuildExtraArgs} ${extraBuildArgs} --package ${pname}";
             } // extraArgs);
         in
         rec {
           inherit cargoArtifacts;
           nickel-lang-core = buildPackage { pname = "nickel-lang-core"; };
-          nickel-lang-cli = buildPackage { pname = "nickel-lang-cli"; };
+          nickel-lang-cli = buildPackage {
+            pname = "nickel-lang-cli";
+            extraBuildArgs = "--features format";
```
yannham, line 246: Don't we want to disable it by default for this PR, until we figure out the static binary build story?

yannham (changes requested):
The structure of the CLI looks better like this :+1: I'm requesting changes to not forget to disable the format feature for now, until we figure out the static build story.

flake.nix
vkleen: We can keep it enabled for the normal CLI build, and just disable it for the static binary. That way at least `nix run github:tweag/nickel` gets the formatting feature.

cli/Cargo.toml
vkleen: Good catch. In fact, `thiserror` isn't used at all anymore and I've made `tempfile` optional.

flake.nix
yannham: Ah, what I understood as "principle of least surprise" from the last planning meeting was that the binary built/run from nix with the default command and the one provided as a static binary shouldn't differ, as you should expect as a user. I would vote for disabling everywhere for now, but just merge it, so that this PR doesn't drift away from the master branch.

aspiwack: It's what I meant. But I don't think you need to choose how you're going to handle it before you release.

vkleen: Alright, I've disabled the `format` feature for the nix build 👍

yannham (approved):

PR #1373, opened 2023-06-16
Add the record builder interface from tf-ncl

vkleen:


vkleen:
This will be part of #1475, and the version in that PR is updated for recent Nickel changes. Closing this one.

PR #1374, opened 2023-06-16
In memory imports

### LINKED ISSUE #1359: Imports in nls get read from disk, not the in-memory version
According to the LSP spec, once we send a `DidOpenTextDocument` to the language server, that data should be treated as the source of truth for the file at that uri. So for example if I send a `DidOpenTextDocument` for the uri `file:///foo.ncl` and then I send some other document that tries to `import "/foo.ncl"`, then that import's contents should be resolved to the data that was sent in the `DidOpenTextDocument` notification. Instead, nls tries to read `/foo.ncl` from disk.

This is easily reproducible in the (WIP) test harness, which sends `DidOpenTextDocument` notifications for files that don't even exist on disk (and nls is unable to import them). It can also be reproduced in an editor, if you edit files without saving them.

jneem:
Changes the file cache to make an explicit distinction between "in-memory" and "on-filesystem" sources.

In normal nickel usage, the stdlib gets loaded "in-memory" and just about everything else comes from the filesystem.

In REPL usage, various things like repl inputs and queries get loaded "in-memory", and they get assigned paths like `<repl-input-3>` that are unlikely to clash with real files on the filesystem (although they aren't namespaced separately so technically they could). Anything that gets imported comes from the filesystem.

In the LSP, files opened in an editor buffer get loaded "in-memory", and they get assigned paths that are likely to coincide with their actual path in the filesystem. (But they might not, for example if the file was newly created in the editor and has not been written.)

The main change in the PR is that if an import gets loaded "in-memory" then we will never try to load it from the filesystem, even if there exists a file with that path that was modified recently. This is the correct behavior for the LSP. Two other changes:
- the path normalization has changed, because the old version only worked for paths that existed on the filesystem. This changes behavior in the presence of symbolic links, as we will no longer notice if two different paths point to the same file.
- the integration tests moved from lsp-harness to nls; we're testing the nls binary, so this is the way to have cargo realize that it needs to build nls before running the tests.

Fixes #1359


lsp/lsp-harness/Cargo.toml
```diff
@@ -7,11 +7,6 @@ edition = "2021"
 anyhow = "1.0.71"
 env_logger = "0.10.0"
 log = "0.4.19"
-lsp-types = "0.94.0"
+lsp-types = "0.88.0"
```
yannham, line 10: Any reason for a downgrade? We might update NLS to work with the latest version as well, if it's trivial. But maybe you prefer to do that in a subsequent PR

lsp/nls/src/cache.rs
```diff
@@ -22,15 +22,7 @@ pub trait CacheExt {
...
-        let path: OsString = path.into();
-        if let Some(file_id) = self.id_of(path.clone()) {
-            self.files_mut().update(file_id, source);
-            // invalidate cache so the file gets parsed again
-            self.terms_mut().remove(&file_id);
-            Ok(file_id)
-        } else {
-            Ok(self.add_string(path, source))
-        }
+        Ok(self.add_tmp(path, source))
```
yannham, line 25: If it's just an alias to `add_tmp`, we can just get rid of `update_content` and use `add_tmp` directly instead. Maybe we can also rename `add_tmp` to `add_in_memory` or something like that, because I reckon that using a method named `add_tmp` inside the LSP to update a document might be a bit confusing.

src/cache.rs
```diff
@@ -817,32 +817,42 @@ impl Cache {
...
-            .and_then(|entry| match entry.timestamp {
-                Some(ts) if ts == timestamp => Some(entry.id),
-                _ => None,
-            })
+    /// Only returns `Ok` if the source is up-to-date; if the source is stale, returns
+    /// either the new timestamp of the up-to-date file or the error we encountered when
+    /// trying to read it (which most likely means there was no such file).
+    ///
+    /// The main point of this awkward signature is to minimize I/O operations: if we accessed
+    /// the timestamp, keep it around.
```
yannham, line 832: It looks like we are encoding a custom ternary type (`UpToDate(FileId), Stale(Timestamp), Err(err)`). I would suggest to write down this type explicitly and use in the signature. Or at least a type for the first layer of `Result`, because it seems the `Err` branch has nothing do to with an error (or, we're using `Result` as an either). So having a binary type `enum SourceState { UpToDate(FileId), Stale(Timestamp) }` and use `io::Result<SourceState>` would also make sense as well (and might be slightly easier to use with `?` and the like).

src/cache.rs
```diff
@@ -1186,9 +1190,41 @@ fn with_parent(path: &OsStr, parent: Option<PathBuf>) -> PathBuf {
...
-pub fn normalize_path(path: &Path) -> io::Result<OsString> {
-    path.canonicalize().map(|p_| p_.as_os_str().to_os_string())
+/// This implementation (including the commend below) was taken from cargo-util.
+///
+/// CAUTION: This does not resolve symlinks (unlike
+/// [`std::fs::canonicalize`]). This may cause incorrect or surprising
+/// behavior at times. This should be used carefully. Unfortunately,
+/// [`std::fs::canonicalize`] can be hard to use correctly, since it can often
+/// fail, or on Windows returns annoying device paths. This is a problem Cargo
+/// needs to improve on.
```
yannham, line 1200: Do you know what are the "incorrect and surprising" behavior at times? In particular, if it's just that it fails to normalize two paths to the same thing although they are equals, that's fine (we're just going to waste a bit of cache by duplicating the file). However, if the behavior is that it can equate two paths that aren't actually the same, that's a whole different story (with correctness and security implications)

yannham (approved):

lsp/lsp-harness/Cargo.toml
jneem: I had to make it compatible with the version in the nls crate. I originally tried bumping the version in nls, but it's incompatible with `codespan-lsp`. (BTW, codespan doesn't seem actively maintained, so we may need to do something about that)

src/cache.rs
jneem: I think the surprises are mostly to do with symlinks. For example, this implementation will normalize `a/../c` to `c`,
but if `a` is a symlink to `b/a` then other implementations (e.g. `std::fs::canonicalize`) will normalize it to `b/c` because they resolve the symlinks before the `..`.

I looked a bit more into what rust-analyzer does, and they've decided that correct canonicalization is just [too hard](https://github.com/rust-lang/rust-analyzer/pull/14430#pullrequestreview-1411530295), and so they use absolute-but-not-canonical paths and rely on the OS to figure out what files they resolve to.

lsp/nls/src/cache.rs
jneem: Ok, I went with renaming to `replace_string`, because it made more sense next to `add_string` (which is also in memory). I've removed the `CacheExt` method, but I'm worried that nls needs to do some more work to handle invalidation properly, for example if you modify a file and then query something in a file that imports it. I'll be looking into this more later...

src/cache.rs
yannham: Ok, so it sounds like indeed cargo's normalization tries to be less smart and might equate less things (which would make sense, and is fine). Thanks for the explanation :+1:

PR #1377, opened 2023-06-19
Revert GitHub actions back to macos-latest

vkleen:
According to https://github.com/cachix/install-nix-action/issues/183 the MacOS problem should be resolved by #1376

Depends on #1376


torhovland:
I think you will also need `cachix/install-nix-action@v22` 🙂

vkleen:
That what #1376 is supposed to do :laughing:

yannham (approved):

PR #1381, opened 2023-06-19
add install info using homebrew on macos

mipmip:
nickel 1.0 is availble in homebrew so why not tell the macOS people about it.


yannham (approved):
Ah, I didn't know that. Thanks!

PR #1382, opened 2023-06-19
Fix LSP panic when importing JSON

### LINKED ISSUE #1378: Importing json/yaml crashes nls
**Describe the bug**

Importing json/yaml crashes `nls`. It works fine in `nickel`.

**To Reproduce**

```
let a = import "test.json" in
a
```

```
[2023-06-19T08:37:03Z DEBUG nls::linearization] adding term: Import("test.json") @ Original(RawSpan { src_id: FileId(5), start: ByteIndex(8), end: ByteIndex(26) })
[2023-06-19T08:37:03Z DEBUG nls::linearization] Add wildcard item: Import("test.json")
[2023-06-19T08:37:03Z DEBUG nls::linearization] adding term: Var(Ident { label: "a" }) @ Original(RawSpan { src_id: FileId(5), start: ByteIndex(30), end: ByteIndex(31) })
[2023-06-19T08:37:03Z DEBUG nls::linearization] adding usage of variable a followed by chain None
[2023-06-19T08:37:03Z DEBUG nls::linearization] linearizing
[2023-06-19T08:37:03Z DEBUG nls::linearization] unresolved references: []
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', lsp/nls/src/cache.rs:104:97
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /private/tmp/nix-build-nls-1.0.0.drv-0/nls-1.0.0-vendor.tar.gz/lsp-server/src/stdio.rs:29:37
```

**Expected behavior**
Should not crash.

**Environment**
 - OS name + version: MacOS 13.4 
 - Version of the code: nickel-lang-lsp 1.0.0

yannham:
Closes #1378

## Context

Nickel terms imported from a JSON file currently don't have any position set, which is making the LSP panic. Indeed, import expressions are registered as an usage site, to enable go to definition on them (which jumps to the corresponding file). The LSP expects that parsed terms have all their positions set, and in particular the one of the root term (corresponding to the whole file).

This commits stores the original input format alongside cached terms, and makes this information available to the public API of `cache`. The LSP can thus use it and abort any usage analysis when the term comes from an external import.

## Improvement

We might be able to set the position of the root term at least, even when it comes from an external format, which would enable go to definition. I'll try that as well

### DIFF
--- a/lsp/nls/src/linearization/mod.rs
+++ b/lsp/nls/src/linearization/mod.rs
@@ -3,7 +3,7 @@ use std::{collections::HashMap, marker::PhantomData};
 use codespan::FileId;
 use log::debug;
 use nickel_lang_lib::{
-    cache::ImportResolver,
+    cache::{ImportResolver, InputFormat},
     identifier::Ident,
     position::TermPos,
     term::{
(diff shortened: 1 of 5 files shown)


lsp/nls/src/linearization/mod.rs
```diff
@@ -460,14 +460,31 @@ impl<'a> Linearizer for AnalysisHost<'a> {
...
                 let position = final_term_pos(&term);
 
-                // This unwrap fails only when position is a `TermPos::None`, which only happens
-                // if the `RichTerm`, has been transformed or evaluated. None of these happen before
-                // linearization, so this is safe.
+                // unwrap(): this unwrap fails only when position is a `TermPos::None`, which only
+                // happens if the `RichTerm` has been transformed or evaluated. None of these
+                // happen before linearization. At this point, we also ensured that we are only
+                // considering a Nickel source (and not a term deserialized from e.g. JSON), and
+                // terms parsed from Nickel must have a position set. So
```
vkleen, line 487: Not that it really matters, but in essence we're `get`ting the term out of the linearization cache twice. First to check that the input format is `Nickel`. Then we throw away the entry and find it again, this time to just unwrap it. Maybe a `let else` construct could be better here?

vkleen (approved):

lsp/nls/src/linearization/mod.rs
yannham: I think `cache.get` returns a `RichTerm`, without the metadata. We probably don't want to change this API, which is used in several places. What I might do is to either add a method that returns the whole `CachedTerm`, but I'm not sure it's a useful interface. Or one that returns a term together with the `input_format`. On the other hand, it's just one additional hashmap get by import, which is probably totally negligible.

vkleen: I agree, it's probably negligible. Let's not hold up this PR with that 👍

yannham: In the end I added a method `get_with_input_format`, which looks reasonable.

PR #1404, opened 2023-06-26
Add support for completion requests to the lsp harness

jneem:
The first iteration of the lsp harness only supported testing goto-definition requests. This adds completion requests to the list.

### DIFF
--- a/lsp/lsp-harness/src/lib.rs
+++ b/lsp/lsp-harness/src/lib.rs
@@ -4,8 +4,8 @@ mod output;
 pub use jsonrpc::Server;
 use log::error;
 use lsp_types::{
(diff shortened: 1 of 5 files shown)


lsp/lsp-harness/src/lib.rs
```diff
@@ -92,6 +93,28 @@ impl Parse for Request {
                 work_done_progress_params: Default::default(),
                 partial_result_params: Default::default(),
             })),
+            // The format of a Completion request is the file:line:char location,
```
yannham, line 96: It's not too important for now, but I wonder if we should follow the same approach as for the integration tests in the future, where we simply write TOML inside comments - might make both using the format and parsing it uniform and less idiosyncratic.

yannham (approved):

PR #1405, opened 2023-06-26
Full vscode extension

szlend:
Ported from: https://github.com/szlend/vscode-nickel

- Add language definition
- Add syntax highlighting
- Replace npm with yarn (npm2nix doesn't support v3 lock format, seems unmaintained)
- Fix *.ncl file watcher
- Fix direnv integration


lsp/client-extension/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
szlend, line 21: Without a LICENSE we get a warning.

lsp/client-extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "nls",
-  "description": "Nickel Language Server",
+  "name": "vscode-nickel",
+  "description": "Nickel Language",
```
szlend, line 3: This will be the title (name) and description in the vscode marketplace.

lsp/client-extension/src/extension.ts
```diff
@@ -1,61 +1,54 @@
...
-			command: serverModule, transport: TransportKind.stdio, options: (enableDebug ? debugOptions : {}),
+			command: serverPath,
 			args: traceFile ? ["--trace", traceFile.toString()] : [],
+			options: (enableDebug ? debugOptions : options),
 		},
 		debug: {
-			command: debugServerModule,
+			command: serverPath,
 			args: traceFile ? ["--trace", traceFile.toString()] : [],
-			transport: TransportKind.stdio,
```
szlend, line 47: Removed this as it was trying to pass `--stdin` to `nls`.

flake.nix
```diff
@@ -401,25 +402,25 @@
...
-            echo y | npx vsce package -o $vsix/${pname}.vsix
-          '';
-        }).vsix;
+      # Build the Nickel VSCode extension
+      vscodeExtension = pkgs.mkYarnPackage {
+        pname = "vscode-nickel";
+        src = pkgs.lib.cleanSource ./lsp/client-extension;
+
+        buildPhase = ''
+          export HOME="$TMPDIR"
```
szlend, line 411: It's trying to create `.yarn` in `HOME`. There's probably a better way to fix this but this works for now.

yannham, line 411: Sounds reasonable, but it might be useful to persist this information

```suggestion
          # yarn tries to create a .yarn file in $HOME. There's probably a
          # better way to fix this but setting HOME to TMPDIR works for now.
          export HOME="$TMPDIR"
```

lsp/client-extension/package.json
yannham: After a quick look at the VSCode market place, many extensions (not necessarily languages) seem to follow the `vscode-xxx` naming scheme, many language extensions don't, but that name sounds reasonable :+1:

lsp/client-extension/src/extension.ts
```diff
@@ -1,61 +1,54 @@
...
-	const debugServerModule = context.asAbsolutePath(
-		path.join('..','..','target', 'debug', 'nls')
-	);
-
-	const serverConfigutration = workspace.getConfiguration("nls.server")
-	const enableDebug: Boolean = serverConfigutration.get("debugLog")
-	const traceFile: PathLike = serverConfigutration.get("trace")
-	const serverModule: string = serverConfigutration.get("path")
+export async function activate(context: ExtensionContext) {
+	const serverConfigutration = workspace.getConfiguration("nls.server");
```
yannham, line 14: I know this wasn't introduced by this PR, but taking this opportunity...

```suggestion
	const serverConfiguration = workspace.getConfiguration("nls.server");
```

lsp/client-extension/syntaxes/nickel.tmLanguage.json
```diff
@@ -0,0 +1,137 @@
...
+        },
+        {
+          "name": "storage.type",
+          "match": "\\b(fun)\\b"
+        }
+      ]
+    },
+    "strings_interpolation": {
+      "name": "string.interpolated",
+      "begin": "m%\"$",
```
yannham, line 48: As a note, this won't work with variable length string delimiters: `m%%\" some string "%%`, but honestly we can add this in a second step. To do that properly we need to share some state between `begin`, `end`, and `patterns`, which would be set by `begin` (and is the number of `%` in the opening delimiter).

lsp/client-extension/syntaxes/nickel.tmLanguage.json
```diff
@@ -0,0 +1,137 @@
...
+      "name": "constant.numeric",
+      "patterns": [
+        {
+          "name": "constant.numeric",
+          "match": "(\\d+)(\\.\\d+)"
+        }
+      ]
+    },
+    "bools": {
+      "name": "constant",
```
yannham, line 118: I'm not sure about this one, but given how numbers are named

```suggestion
      "name": "constant.boolean",
```

lsp/client-extension/syntaxes/nickel.tmLanguage.json
```diff
@@ -0,0 +1,137 @@
...
+      "patterns": [
+        {
+          "name": "constant",
```
yannham, line 121:
```suggestion
          "name": "constant.boolean",
```

lsp/client-extension/syntaxes/nickel.tmLanguage.json
```diff
@@ -0,0 +1,137 @@
...
+          "match": "\\b(true|false)\\b"
+        }
+      ]
+    },
```
yannham, line 125: Should we add the `null` value as well, here?

```suggestion
    },
    "null": {
      "name": "constant.null",
      "patterns": [
        {
          "name": "constant.null",
          "match": "\\b(null)\\b"
        }
      ]
    },
```

yannham (commented):
Many thanks for this @szlend ! It's fine by me, I just don't approve yet such that you can take a look at the review comments and decide what to do with them. When it's done, please ping me or re-request my review and I'll merge it.

lsp/client-extension/package.json
szlend: Ah, there's also `displayName` as `name` can't contain spaces. Do we want something like  "Nickel Language" instead? And add some more details in "description“?

lsp/client-extension/syntaxes/nickel.tmLanguage.json
szlend: Yeah I'm not too familiar with the `tmLanguage` spec. I added a fix that should at least handle the start and end of variable length string delimiters.

lsp/client-extension/syntaxes/nickel.tmLanguage.json
yannham: I think this one went under the radar , so I took the liberty to commit the suggestion

yannham (approved):
Looks good! Before publishing we'll take a look at the name/description first, but otherwise this should be a reasonable first version.

PR #1410, opened 2023-06-28
Make the pretty printer correctly escape field names

vkleen:
Previously, the pretty printer relied on `Ident::label_quoted` for printing record field names. This function blindly placed an identifier between quotes if it was deemed necessary. This produces incorrect results if the identifier label itself contains quotes, for example.

### DIFF
--- a/core/src/pretty.rs
+++ b/core/src/pretty.rs
@@ -1,6 +1,7 @@
 use crate::destructuring::{self, RecordPattern};
 use crate::identifier::Ident;
 
+use crate::parser::lexer::KEYWORDS;
 use crate::term::{
     record::{Field, FieldMetadata},
     *,
@@ -8,6 +9,7 @@ use crate::term::{
 use crate::types::*;
 
 use malachite::num::{basic::traits::Zero, conversion::traits::ToSci};
+use once_cell::sync::Lazy;
 pub use pretty::{DocAllocator, DocBuilder, Pretty};
 use regex::Regex;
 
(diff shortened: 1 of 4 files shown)


jneem (approved):
Unrelated to the current patch, but aren't we compiling a regex every time we pretty-print a label?

core/src/pretty.rs
```diff
@@ -42,6 +42,13 @@ fn sorted_map<K: Ord, V>(m: &'_ IndexMap<K, V>) -> Vec<(&'_ K, &'_ V)> {
     ret
 }
 
+pub fn escape(s: &str) -> String {
```
yannham, line 45: Nitpick, but shouldn't this function be moved to `indentifier` or even to `term::string` if it's used inside `identifier`? it's not obvious that `identifier` should be depending on `pretty`

vkleen: That's a good point. It felt too unrelated to `Ident` to go into `identifier` to me. In fact, `Ident::label_quoted` is used only in the pretty printer. So maybe the correct course of action to remove this dependency would be to move `label_quoted` into `pretty`? I'm not sure.

yannham: Hmm, good point as well. I think both are reasonable:

- Leave `label_quoted` in identifier and bring back escape either in `identifier` or `term::string`, because I believe what you're doing in the end is to escape a Nickel string. It would like implementing `Display`: I think it's fair to have it inside identifier (or, should we just implement `Display`, by the way, instead of `label_quoted`?)
- Move both to `pretty`, because it's about pretty-printing stuff

I guess the current in-between with one inside `identifier` and one inside `pretty` is what's feeling itchy

PR #1412, opened 2023-06-28
Stop recompiling the quoting regex for printing identifiers

vkleen:
As observed by @jneem we were constantly recompiling the regular expression determining whether quoting an identifier is necessary. This change improves the pretty printer performance on a non-synthetic example by a factor of 3.


jneem (approved):

yannham (approved):
2 lines change, x3 multiplier, sounds like a good operation :+1:

PR #1419, opened 2023-06-29
Trees that grow notes

Radvendii:


jneem (approved):
Very nice summary!

One thing that didn't occur to me before was the question of how to handle recursion. Like, if `Term` has a `Children(Vec<Term>)` element then presumably the macro has to change it to `Children(Vec<TermBase>)` and `Children(Vec<TermFoo>)` and `Children(Ext::Children, Vec<ThatGrows<Ext>>)`, right?

I also thought a little more about how to avoid conversion costs between the two representations, and I think can be done by making the trees-that-grow interface shallow and reference-based. So something like

```rust
enum ThatGrows<'a, Ext: Extension, T> {
    NoChange(&'a Ext::NoChange),
    Extend(&'a Ext::Extend, &'a i32),
    More(&'a Ext::More),
    // If there's recursion, we just take a shallow reference, not a deep copy.
    Children(&'a Vec<T>),
}

impl<'a> From<&'a TermBase> for ThatGrows<'a, ext::Base, TermBase> { ... }
impl<'a> From<&'a TermFoo> for ThatGrows<'a, ext::Foo, TermFoo> { ... }
```

(And there could also be a `&mut` variant)

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,167 @@
+As nickel grows, it becomes unwieldy to use a single AST for every stage of compilation. Some things are kept out of the AST, even though they would be useful in some places, because they don't make sense and would get in the way at others (e.g. types for the LSP, closures at runtime). Other things are in the AST, but aren't actually filled in until later on in the process (e.g. forall's var_kind). This causes `unwrap()`s and `unreachable!()`s and hurts code maintainability.
```
yannham, line 1: Nitpick, but Nickel is (not yet?) compiled :stuck_out_tongue: 

```suggestion
As Nickel grows, it becomes unwieldy to use a single abstract syntax tree (AST) for every stage of the pipeline. Some data are kept out of the AST, even though they would be useful in some places, because they don't make sense and would get in the way at others (e.g. types inferred during typechecking for the LSP, a closure - i.e. a term together with its environment - at runtime). Other data placeholder are in the AST, but aren't actually filled in until later on in the process (e.g. forall's var_kind). This causes `unwrap()`s and `unreachable!()`s and hurts code maintainability.
```

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Strategies
+
+One strategy to deal with this is to repeat a large AST enum in multiple places with small differences, and convert from one to another as we move through compilation. This creates a large maintenance burden, since many changes have to be propagated to all the trees.
```
yannham, line 5:
```suggestion
One strategy to deal with this is to repeat a large AST enum in multiple places with small differences, and convert from one to another as we move through program manipulations. This creates a large maintenance burden, since many changes have to be propagated to all the trees.
```

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,167 @@
...
+
+Another strategy is outlined by the paper [Trees that Grow](https://www.microsoft.com/en-us/research/uploads/prod/2016/11/trees-that-grow.pdf). This describes an extendable tree structure using type families and pattern synonyms for ergonomics. While this is quite a bit DRYer, there is still a decent amount of boilerplate, and actually using these trees is less intuitive. Especially in rust, where there are no pattern synonyms.
```
yannham, line 7:
```suggestion
Another strategy is outlined by the paper [Trees that Grow](https://www.microsoft.com/en-us/research/uploads/prod/2016/11/trees-that-grow.pdf). This describes an extendable tree structure using type families and pattern synonyms for ergonomics. While this is quite a bit DRYer, there is still a decent amount of boilerplate, and actually using these trees is less intuitive. Especially in Rust, where there are no pattern synonyms (the paper is targeting Haskell).
```

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,167 @@
...
+
+I (@Radvendii) have explored using proc macros to get the best of both worlds, and remove the boilerplate along the way. There are three ways we could use proc macros, but they share a common syntax.
+
+## Macro Syntax
+
+The naming of things is up in the air, but the general idea is to annotate the tree definition with how fields get modified in the extensions. There are two kinds of extensions, product extensions which add data to an existing term type, and sum extensions which add a new term type.
+
+```rust
```
yannham, line 15: It could help to add a written description of what the following program is supposed to mean, that is:

> The following example models 3 different ASTs, such that ...

It's not obvious what `Foo` is here (a type? a name? something else), and it's a bit hard to follow the translation if we don't really know what we are translating to begin with.

yannham (commented):
It's a valuable document, but I feel like you presented even more material to us. For example, we can see the generated code (which is arguably useful to understand the translation), but not _how does one use the generated code in concrete examples_, which is IMHO even more important, because that should be the main criterion to choose an approach. How can we write a simple function that is generic over any extension (ie which works morally on `Base`) in the different approaches? How do we pattern match, how do we transform between reprs, etc. ?

yannham:
@Radvendii what is the status of the PR? It's been approved, meaning that you can implement whatever you want from the last review comment and then go ahead and merge :slightly_smiling_face:

Radvendii:
Sorry it's taken me so long. I changed quite a bit, so it would be good if someone could look it over. Especially if @jneem could look over the part referencing his idea and make sure I understood it correctly. 

If not, I'll merge it at the beginning of next week.

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,399 @@
+As Nickel grows, it becomes unwieldy to use a single abstract syntax tree (AST) for every stage of the pipeline. Some data are kept out of the AST, even though they would be useful in some places, because they don't make sense and would get in the way at others (e.g. types inferred during typechecking for the LSP, a closure - i.e. a term together with its environment - at runtime). Other data placeholder are in the AST, but aren't actually filled in until later on in the process (e.g. forall's var_kind). This causes `unwrap()`s and `unreachable!()`s and hurts code maintainability.
```
jneem, line 1:
```suggestion
As Nickel grows, it becomes unwieldy to use a single abstract syntax tree (AST) for every stage of the pipeline. Some data are kept out of the AST, even though they would be useful in some places, because they don't make sense and would get in the way at others (e.g. types inferred during typechecking for the LSP, a closure - i.e. a term together with its environment - at runtime). Other data placeholders are in the AST, but aren't actually filled in until later on in the process (e.g. forall's var_kind). This causes `unwrap()`s and `unreachable!()`s and hurts code maintainability.
```

notes/trees-that-grow-proc-macro.md
```diff
@@ -0,0 +1,399 @@
...
+    type Var;
+    type Ann;
+    type Abs;
+    type App;
+    type More;
+}
+enum Exp<Ext: Extension> {
+    Lit(Ext::Lit, i32),
+    Var(Ext::Var, Var),
+    Ann(Ext::Ann, Box<Exp>, Typ),
```
jneem, line 106:
```suggestion
    Ann(Ext::Ann, Box<Exp<Ext>>, Typ),
```
(and also below)

jneem (approved):

PR #1421, opened 2023-06-29
manual docs: fix mismatch between input command and its echo in the output

foo-jin:
Hey, I noticed a little typo while browsing the manual.


vkleen (approved):
Thank you for noticing this! In fact, the position reporting got even more precise with the latest version.

PR #1423, opened 2023-06-30
Format benches as well using Topiary

yannham:
tests and benches were excluded from auto-formatting because they were causing non-idempotency errors from Topiary, but in fact only tests do so. While it's not criticial, it's still nicer to have the benchmarks properly formatted, if we can.

This PR runs Topiary on `core/benches/**/*.ncl` and remove the exclusion of benches from the pre-commit-hooks.


vkleen (approved):
Did we ever investigate why some of the test files fail the idempotency check? That seems like it should be a bug in topiary.

yannham:
@vkleen I had the same thought (and this was in fact my initial goal when formatting the benches again, to find examples to report on Topiary), I've opened https://github.com/tweag/topiary/issues/547 just after this PR

PR #1424, opened 2023-06-30
Specify the test harness requests as toml instead of a custom format

jneem:
The new format is just a toml-ization of the LSP protocol itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1310,6 +1310,7 @@ dependencies = [
  "serde",
  "serde_json",
  "test-generator",
+ "toml",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


yannham (approved):
Sounds good and easier to extend now! I wonder if at one point describing the full request would be too verbose (for example, I'm not sure I know what is the `triggerKind` field), but we can think about that when it happens. We might simply merge a partial request specification with some default values.

jneem:
With the disclaimer that I've only looked at a few request types so far, the LSP already has a bunch of optional fields with sane defaults. So I think there's a reasonable chance that the extra verbosity will continue to be manageable.

PR #1426, opened 2023-07-03
Invalidate importers in NLS (plus a couple other import-related issues)

jneem:
Fix a few related import issues.

  - record imports correctly even if the imported file is already cached.
  - avoid infinite recursion with circular imports. This previously
    worked only because the imports weren't correctly recorded.
  - track reverse-dependencies also, and invalidate the LSP correctly.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -510,14 +510,13 @@ impl Cache {
                     self.update_state(file_id, EntryState::Typechecking);
                     self.wildcards.insert(file_id, wildcards);
 
-                    // Here and below, update the state before recursing so that circular
-                    // imports don't cause an infinite loop.
-                    self.update_state(file_id, EntryState::Typechecked);
                     if let Some(imports) = self.imports.get(&file_id).cloned() {
                         for f in imports.into_iter() {
                             self.typecheck(f, initial_ctxt)?;
                         }
                     }
+
+                    self.update_state(file_id, EntryState::Typechecked);
                 }
                 // The else case correponds to `EntryState::Typechecking`. There is nothing to do:
                 // cf (grep for) [transitory_entry_state]
@@ -550,13 +549,13 @@ impl Cache {
                             ..cached_term
                         },
                     );
-                }
 
-                self.update_state(file_id, EntryState::Transformed);
-                if let Some(imports) = self.imports.get(&file_id).cloned() {
-                    for f in imports.into_iter() {
-                        self.transform(f)?;
+                    if let Some(imports) = self.imports.get(&file_id).cloned() {
+                        for f in imports.into_iter() {
+                            self.transform(f)?;
+                        }
                     }
+                    self.update_state(file_id, EntryState::Transformed);
                 }
                 Ok(CacheOp::Done(()))
             }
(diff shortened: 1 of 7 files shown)


core/src/cache.rs
```diff
@@ -507,13 +510,14 @@ impl Cache {
                     self.update_state(file_id, EntryState::Typechecking);
                     self.wildcards.insert(file_id, wildcards);
 
+                    // Here and below, update the state before recursing so that circular
+                    // imports don't cause an infinite loop.
+                    self.update_state(file_id, EntryState::Typechecked);
```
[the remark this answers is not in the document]
jneem: Good call, I overlooked those early returns. I made another try that is hopefully correct, based on the comments on `EntryState::Transforming` and the previous implementation of `typecheck`

lsp/nls/src/linearization/mod.rs
```diff
@@ -51,6 +51,11 @@ impl LinRegistry {
         let lin = self.map.get(&id.file_id).unwrap();
         lin.get_item(id)
     }
+
+    /// Remove a file's linearization from the registry.
+    pub fn remove_file(&mut self, file_id: FileId) {
```
yannham, line 56: I don't mind adding such helpers, but it must be noted that in many many places, we are simply using directly `lin_registry.map.xxx`. In particular `insert`. It looks a bit inconsistent to add a `remove_file` but not an `insert_file`, for example. I'm fine either way (using directly `registry.map.xxx` or adding common methods): adding such methods achieve better encapsulation, but it's also noisy boilerplate because we simply wrap `self.map.xxx`, which isn't too bad either. But at least I think we should be consistent.

yannham (commented):
I'm not sure to understand this part:

>- record imports correctly even if the imported file is already cached.
>- avoid infinite recursion with circular imports. This previously
> worked only because the imports weren't correctly recorded.

What do you mean exactly by recorded? I think circular imports also currently do have a specific codepath to avoid infinite loops, and they did make the interpreter loop at some point in time. Maybe there's a mutual misunderstanding about some of the mechanisms of the cache or the implementation somewhere?

jneem:
> What do you mean exactly by recorded? I think circular imports also currently do have a specific codepath to avoid infinite loops, and they did make the interpreter loop at some point in time. Maybe there's a mutual misunderstanding about some of the mechanisms of the cache or the implementation somewhere?

By "recorded" I just meant that if `a.ncl` imports `b.ncl` then there should be a corresponding entry in `cache.imports`. At least, that was my assumption; is it not correct?

Prior to this PR, if `a.ncl` imported `b.ncl` but `b.ncl` was already cached, then `cache.imports` would not get that entry. This could happen in with imports that follow a "diamond" pattern, and it can also easily happen in the LSP if the editor opens `b.ncl` before it opens `a.ncl`

lsp/nls/src/linearization/mod.rs
jneem: Sounds good, I've removed it for now.

yannham (approved):
Ah, that makes sense. Yes, I now believe that those were incorrectly handled, but that this was fine in the case of the interpreter, because what's required is probably just that each import exists _at least somewhere_ as the import of some file. For the LSP you do probably want `imports` to be correct, though.

Issue #1429, opened 2023-07-04
Wrong parsing of multiline string ending as an interpolation sequence


yannham:
**Describe the bug**

If a closing delimiter of a multiline string is followed by zero or more `%` and a final `{`, it is unduly parsed (or should I say lexed, in this case) as an interpolation sequence instead of a closing delimiter.

**To Reproduce**

```nickel
nickel> m%%""%%%{std.to_string 1}" World"%%
"\"%1\" World"

```

**Expected behavior**
The previous program should fail with a syntax error, as it is an empty string `m%%""%%` followed by gibberish. 

**Environment**
 - Version of the code: ec3b9c4a35bced37d321a01636b80b4f91533924

**Additional context**

The example comes from a test. I suspect that when the closing delimiter of multiline strings was `"%m` (pre 1.0), lexing `"%{` correctly implied to favor string interpolation over a potential end of a string (as the presence of `{` was disqualifying the whole sequence as a closing delimiter). This code has stayed the same, but now we don't have a trailing `m` anymore, and `"%` must be treated as a closing delimiter, no matter what comes next. See https://github.com/tweag/topiary/issues/547#issuecomment-1618430667.

yannham:
The bug is actually worse, since if we close a string with a number of `%` higher than the expected count, it just doesn't close the string:

```nickel
nickel> m%""%%%"%
"\"%%%"
```

PR #1430, opened 2023-07-04
[Refactoring] Better separation of lexer mode

yannham:
Preliminary work toward fixing #1429. Refactors the lexer for a better and cleaner separation between different lexer modes, making it more readable and statically forbid more invalid states.

The added error `StrEndMismatch` is more related to #1429, but at this point it's easier to leave it there than to artificially split it from this PR - the corresponding code is relatively small.

### DIFF
--- a/core/src/parser/lexer.rs
+++ b/core/src/parser/lexer.rs
@@ -426,7 +426,7 @@ pub enum MultiStringToken<'input> {
     /// variable number of `%` character, so the lexer matches candidate end delimiter, compare the
     /// number of characters, and either emit the `End` token above, or turn the `CandidateEnd` to a
     /// `FalseEnd` otherwise
-    #[regex("\"%+")]
+    #[regex("\"%+")] // m%%""%%{x}"%%m
     CandidateEnd(&'input str),
     /// Same as `CandidateEnd`, but for interpolation
     #[regex("%+\\{")]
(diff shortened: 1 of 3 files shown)


core/src/parser/lexer.rs
```diff
@@ -426,7 +426,7 @@ pub enum MultiStringToken<'input> {
     /// variable number of `%` character, so the lexer matches candidate end delimiter, compare the
     /// number of characters, and either emit the `End` token above, or turn the `CandidateEnd` to a
     /// `FalseEnd` otherwise
-    #[regex("\"%+")]
+    #[regex("\"%+")] // m%%""%%{x}"%%m
```
vkleen, line 429: Is this comment left-over from some experiments? It seems to use the old, symmetric delimiters.

core/src/parser/lexer.rs
```diff
@@ -630,168 +680,271 @@ impl<'input> Lexer<'input> {
...
-                } else {
-                    return Some(Err(ParseError::Lexical(
-                        LexicalError::InvalidAsciiEscapeCode(span.start + 2),
-                    )));
-                }
+            // Strictly speaking, a candidate end delimiter with more than the required count of
+            // `%` should be split between multistring end token, plus a variable number of `%`
+            // tokens. This is annoying because we only buffer one token currently. We could use a
+            // stack instead of a 1-length buffer, but in practice a string such as `m%" "%%` is
+            // almost surely meaningless: there's now meaningful way of interpreting it
```
vkleen, line 857:
```suggestion
            // almost surely meaningless: there's no meaningful way of interpreting it
```

vkleen (approved):

core/src/parser/lexer.rs
yannham: Ah, probably.

yannham, line 429:
```suggestion
    #[regex("\"%+")]
```

PR #1434, opened 2023-07-07
Add notes on partial parsing

jneem:
I made enough progress on partial parsing for `nls` to realize that the current approach has some issues. I wrote up these notes describing the problem and exploring some alternatives.


yannham (approved):
Thanks for writing this up, I think it's indeed useful to record your investigation in those notes.

PR #1435, opened 2023-07-07
[Fix] Fix bad lexing of mutline string closing delimiter

### LINKED ISSUE #1429: Wrong parsing of multiline string ending as an interpolation sequence

yannham:
Closes #1429.

We introduced a hack in the lexer because of a Logo issue related backtracking. Pre Nickel 1.0, the problem was related to strings like:

```nickel
let x = "string" in m%%"some "%%{x}"%%m
```

Here, `"%%{x}` had to be interpreted as a double quote `"` followed by an interpolation sequence `%%{`. Backtracking during matching is required, because up to `{`, we don't know yet if we are lexing an end delimiter `"%%m` or a `"` followed by an interpolation sequence. Logos doesn't handle those well (see https://github.com/maciejhirsz/logos/issues/200). So we added another token for this "ambiguous" (not literally) sequence, the `CandidateQuoteInterpolation`. However, now that that the final `m` is not required, a sequence of the form `"%%{` is either an end delimiter followed by a normal, non-string token `{` (if there's as many `%` as in the opening delimiter), a "closing delimiter too long" if there's more `%` than in the opening delimiter), or otherwise a verbatim `"%%{` (if there's less `%` than in the opening delimiter). This can _never_ be interpreted as a double quote followed by an interpolated sequence.

Thus, we can get rid of the ad-hoc `CandidateQuoteInterpolation` and related special treatments. This also get rid of regression tests, because this subtlety had caused several issues in the past but they are now moot given the new 1.0 syntax.

### DIFF
--- a/core/src/parser/lexer.rs
+++ b/core/src/parser/lexer.rs
@@ -859,7 +859,7 @@ impl<'input> Lexer<'input> {
             // modulo operator `%` - which will fail anyway at runtime with a type error).
             // Thus, we prefer to emit a proper error right here.
             MultiStringToken::CandidateEnd(s) if s.len() > data.percent_count => {
-                return Some(Err(ParseError::Lexical(LexicalError::StringEndMismatch {
+                return Some(Err(ParseError::Lexical(LexicalError::StringDelimiterMismatch {
                     opening_delimiter: data.opening_delimiter.clone(),
                     closing_delimiter: span,
                 })))
--- a/core/src/parser/lexer.rs
+++ b/core/src/parser/lexer.rs
@@ -431,12 +431,6 @@ pub enum MultiStringToken<'input> {
     /// Same as `CandidateEnd`, but for interpolation
     #[regex("%+\\{")]
     CandidateInterpolation(&'input str),
-    /// Unfortunate consequence of Logos' [issue #200](https://github.com/maciejhirsz/logos/issues/200).
-    /// The other rules should be sufficient to match this as a double quote followed by a
-    /// `CandidateInterpolation`, but if we omit this token, the lexer can fail unexpectedly on
-    /// valid inputs because of #200.
-    #[regex("\"%+\\{")]
-    QuotesCandidateInterpolation(&'input str),
     /// Token emitted by the modal lexer for the parser once it has decided that a `CandidateEnd` is
     /// an actual end token.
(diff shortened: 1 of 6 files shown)


core/src/parser/lexer.rs
```diff
@@ -467,8 +461,10 @@ pub enum ModalLexer<'input> {
     MultiString {
         mode_data: MultiStrData,
         /// A token that has been buffered and must be returned at the next call to `next()`. This is
-        /// made necessary by an issue of Logos (<https://github.com/maciejhirsz/logos/issues/200>). See
-        /// [`MultiStringToken::QuotesCandidateInterpolation`].
+        /// related to lexing an possible interpolation sequence, such as `%%%{`, which requires to
+        /// split a candidate interpolation token in two. In this case, we need to emit the first
+        /// token on the spot, and bufferize the second one, to be emitted on the following call to
```
jneem, line 466:
```suggestion
        /// token on the spot, and buffer the second one, to be emitted on the following call to
```

jneem (approved):

core/src/parser/lexer.rs
jneem: I guess I can't argue with wiktionary :shrug:

PR #1438, opened 2023-07-10
Add `term::make::static_access`

vkleen:
Add a helper function to the `term::make` module to construct chains of static record accesses.

### DIFF
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -1962,6 +1962,19 @@ pub mod make {
     pub fn integer(n: impl Into<i64>) -> RichTerm {
         Term::Num(Number::from(n.into())).into()
     }
+
+    pub fn static_access<I, S>(record: S, fields: I) -> RichTerm
+    where
+        I: IntoIterator<Item = S>,
+        I::IntoIter: DoubleEndedIterator,
+        S: Into<Ident>,
+    {
+        let mut term = make::var(record);
+        for f in fields.into_iter() {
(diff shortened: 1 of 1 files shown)


core/src/term/mod.rs
```diff
@@ -1962,6 +1962,29 @@ pub mod make {
...
+        T: Into<RichTerm>,
+    {
+        let mut term = record.into();
+        for f in fields.into_iter() {
+            term = make::op1(UnaryOp::StaticAccess(f.into()), term);
+        }
+        term
+    }
+
+    pub fn static_access_<I, S>(record: S, fields: I) -> RichTerm
```
yannham, line 1980: I'm not too fond of naming a public symbol with a trailing `_`. Maybe something like `static_access_var`, which better reflects the semantics?

Tangentially, static access is not a great name, and inherited from the very first prototype, but it's probably not the role of this PR to change that :) I would vouch for something with `field` in it, at least

yannham: In the end, I wonder if we even need this variant. `make::var` is like 9 characters, and if it's really heavily used somewhere (JSON schema converter), maybe the consuming crate can span its own `static_access_var`?

yannham (approved):

core/src/term/mod.rs
vkleen: Thinking about it again, I think you're right that `static_access_` isn't that useful. I'll get rid of it and just keep the other variant.

PR #1440, opened 2023-07-10
Fix current version in manual and README

yannham:


vkleen (approved):

README.md
```diff
@@ -92,8 +92,9 @@ the `nickel-lang-core` crate documentation).
      `$PATH` and is available anywhere.
    - If you're running macOS you can use Homebrew to install the Nickel binary
      with `brew install nickel`.
-   - Without Nix, you can use `cargo run` after [building](#build), passing
-     arguments with an extra `--` as in `cargo run --bin nickel -- -f program.ncl`.
+   - Without Nix, you can use `cargo run --bin nickel` after [building](#build),
```
jneem, line 95: Maybe it's worth adding a [default-run](https://doc.rust-lang.org/cargo/reference/manifest.html#the-default-run-field) field so `cargo run` just works?

vkleen: I looked into that and it seems like `default-run` only works for multiple binaries in a single crate, not across a workspace.

PR #1442, opened 2023-07-11
Add a Types variant to Term

jneem:
Types can get converted into terms (by turning them into contracts). Previously, this conversion was carried out at the parsing stage. This PR postpones the conversion to evaluation (or typechecking), by allowing the AST to store a `Types` directly, and then converting it as-needed.

One part I wasn't sure about: is it worth adding a mutable pass (like `subst`) to convert the `Types` to contracts? Right now I'm doing the conversion as-needed, so it might get converted twice (in typechecking, and then again in evaluation).

I also rewrote `RichTerm::without_pos` to use `traverse` instead of `apply_to_rich_term`, and then removed `apply_to_rich_term` because nothing else was using it.

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -293,9 +293,12 @@ pub enum TypecheckError {
         RichTerm, /* the inferred flat type */
         TermPos,
     ),
-    /// Types can appear in term position, like the `Number` in `let c = Number in 5 | c`.
-    /// These types are treated as contracts, but they are not allowed to contain flat types.
-    /// For example, `let c = { foo : 5 } in ...` is disallowed.
+    /// Within statically typed code, the typechecker must reject terms containing nonsensical
+    /// contracts such as `let C = { foo : 5} in ({ foo = 5 } | C)`, which will fail at runtime.
+    /// The typechecker is currently quite conservative and simply forbids to store any custom
+    /// contract (flat type) in a type that appears in term position. Note that this restriction
+    /// doesn't apply to annotations, which aren't considered part of the statically typed block.
+    /// For example, `{foo = 5} | {foo : 5}` is accepted by the typechecker.
     FlatTypeInTermPosition {
         /// The term that was in a flat type (the `5` in the example above).
         flat: RichTerm,
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -293,6 +293,15 @@ pub enum TypecheckError {
         RichTerm, /* the inferred flat type */
         TermPos,
     ),
+    /// Types can appear in term position, like the `Number` in `let c = Number in 5 | c`.
+    /// These types are treated as contracts, but they are not allowed to contain flat types.
+    /// For example, `let c = { foo : 5 } in ...` is disallowed.
+    FlatTypeInTermPosition {
+        /// The term that was in a flat type (the `5` in the example above).
+        flat: RichTerm,
+        /// The position of the entire type (the `{foo : 5}` in the example above).
+        pos: TermPos,
+    },
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Default)]
@@ -2050,6 +2059,25 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                         .with_notes(vec![
                             "Type variables introduced in a `forall` range over all possible types.".to_owned(),
                         ]),
+                ]
+            }
+            TypecheckError::FlatTypeInTermPosition { flat, pos } => {
+                vec![
+                    Diagnostic::error()
+                        .with_message(
+                            "types containing user-defined contracts cannot be converted into contracts",
+                        )
+                        .with_labels(
+                            pos.into_opt().map(|span|
+                                    primary(&span).with_message("This type (in contract position)")
+                                )
+                                .into_iter()
+                            .chain(
+                                flat.pos.into_opt().map(|span|
+                                    secondary(&span).with_message("contains this user-defined contract")
+                                ))
+                            .collect()
+                    ),
                 ]
             }
         }
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -968,7 +968,7 @@ pub fn subst<C: Cache>(
         | v @ Term::ResolvedImport(_)
         // We could recurse here, because types can contain terms which would then be subject to
         // substitution. Not recursing should be fine, though, because a type in term position
-        // turns into a contract, and we don't serialize contracts.
+        // turns into a contract, and we don't substitute inside contracts either currently.
         | v @ Term::Types(_) => RichTerm::new(v, pos),
         Term::Let(id, t1, t2, attrs) => {
             let t1 = subst(cache, t1, initial_env, env);
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -966,6 +966,9 @@ pub fn subst<C: Cache>(
         | v @ Term::Enum(_)
         | v @ Term::Import(_)
         | v @ Term::ResolvedImport(_)
+        // We could recurse here, because types can contain terms which would then be subject to
+        // substitution. Not recursing should be fine, though, because a type in term position
+        // turns into a contract, and we don't serialize contracts.
         | v @ Term::Types(_) => RichTerm::new(v, pos),
         Term::Let(id, t1, t2, attrs) => {
             let t1 = subst(cache, t1, initial_env, env);
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -709,6 +709,11 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                         env,
                     }
                 }
+                // Evaluating a type turns it into a contract.
+                Term::Types(ty) => Closure {
+                    body: ty.contract()?,
+                    env,
+                },
                 // Continuation of operations and element update
                 _ if self.stack.is_top_idx() || self.stack.is_top_cont() => {
                     clos = Closure {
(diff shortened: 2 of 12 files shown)


vkleen (approved):

core/src/eval/mod.rs
```diff
@@ -960,7 +965,8 @@ pub fn subst<C: Cache>(
         | v @ Term::SealingKey(_)
         | v @ Term::Enum(_)
         | v @ Term::Import(_)
-        | v @ Term::ResolvedImport(_) => RichTerm::new(v, pos),
+        | v @ Term::ResolvedImport(_)
+        | v @ Term::Types(_) => RichTerm::new(v, pos),
```
yannham, line 969: I wonder if this is the expected behavior. Morally, types can contain terms (imagine `Contract1 -> Contract2`), so we should very much continue the substitution by descending into the term, strictly speaking. This is the current behavior as `Contract1 -> Contract2` is turned into `$arrow Contract1 Contract2`, and thus is a normal term subject to substitution.

On the other hand, we don't substitute types/contracts in annotation. I suspect not substituting here is fine, because in the end putting  a type inside a variable means that you're going to use it as a contract anyway, and not say, serialize it, but at least this deserve a comment to say why we don't substitute under types.

core/src/term/mod.rs
```diff
@@ -181,6 +181,12 @@ pub enum Term {
     #[serde(skip)]
     ResolvedImport(FileId),
 
+    /// A type.
```
yannham, line 184:
```suggestion
    /// A type in term position, such as in `let my_contract = Number -> Number in ...`.
```

core/src/term/mod.rs
```diff
@@ -570,82 +576,6 @@ impl<E> StrChunk<E> {
 }
 
 impl Term {
-    /// Recursively apply a function to all `Term`s contained in a `RichTerm`.
-    pub fn apply_to_rich_terms<F>(&mut self, func: F)
```
yannham, line 574: wasn't this function used inside tests? I guess the CI will tell

core/src/term/mod.rs
```diff
@@ -1443,14 +1378,19 @@ impl RichTerm {
     ///
     /// It allows to use rust `Eq` trait to compare the values of the underlying terms.
     //#[cfg(test)]
```
yannham, line 1380:
```suggestion
```

core/src/term/mod.rs
```diff
@@ -1443,14 +1378,19 @@ impl RichTerm {
...
+                let term = match t.term.into_owned() {
+                    Term::Types(ty) => Term::Types(ty.without_pos()),
+                    t => t,
+                };
+                Ok(RichTerm::new(term, TermPos::None))
+            },
+            &mut (),
+            TraverseOrder::BottomUp,
+        )
+        .unwrap()
```
yannham, line 1393: Ah, right, we can use `traverse`.

core/src/typecheck/mod.rs
```diff
@@ -1589,6 +1590,11 @@ fn check<L: Linearizer>(
             );
             unify(state, &ctxt, ty, ty_import).map_err(|err| err.into_typecheck_err(state, rt.pos))
         }
+        // Types in term position are converted to contracts, so convert it before type-checking
+        // it. Note that we'll throw away the converted contract here and then do the conversion
+        // again during evaluation. If we had a mutable pass instead (like `subst`) we could
+        // do this work once.
+        Term::Types(typ) => check(state, ctxt, lin, linearizer, &typ.contract()?, ty),
```
yannham, line 1597: I don't think this makes a lot of sense in the long term. It makes sense because this is the current behavior, but truth is, this is never going to be well-typed (beside very trivial cases like primitive types, maybe):

```nickel
nickel> (let x = Number -> Number in 5) : _
error: incompatible types
 = Expected an expression of type `_a -> _b`
 = Found an expression of type `Dyn`
```

This is turned into applications of builtin contracts, here `$arrow $number $number`, but `$arrow` isn't typed, so it has type `Dyn`. I can see several possibilities:

1. Simply walk the type, instead of checking it. This would act as if the type was somewhere in an annotation. The cons is that we are too accepting: something absurd such as `Number -> 5` now typechecks. 
2. Show an error saying that types in term position aren't supported in statically typed blocks (probably the easiest, but the more restrictive)
3. Accept types which don't have any contracts (flat types) in them. Indeed, any "pure" type is well-formed and should be accepted. The issue is that the type of the content of a flat type can't be expressed in Nickel's type system (it's a union of `{_ : _}` and `Dyn -> Label -> Dyn` more or less), so there's no chance we can typecheck it in all generality currently. We could still accept some special cases, maybe.

I think I would favor 2., with the possibility of going to 3. incrementally. We probably have to go there before the next release though, otherwise 2. alone would be a regression as currently this typechecks (although this is a dubious behavior to rely on, as it's highly dependent of internal implementation details such as is this primitive type converted to a contract expression that has an application in it or not):

```nickel
(let x = Number in 5) : _
```

yannham (changes requested):
`impl Traverse<RichTerm> for RichTerm` must be updated to handle `Types`. Otherwise, I wonder how all of this interact with the LSP, and how - and if - we should linearize a type.

core/src/term/mod.rs
```diff
@@ -1377,7 +1377,7 @@ impl RichTerm {
     /// Erase recursively the positional information.
     ///
     /// It allows to use rust `Eq` trait to compare the values of the underlying terms.
-    //#[cfg(test)]
+    #[cfg(test)]
```
yannham, line 1380: I wonder why it was commented out, though. @vkleen do you use that in any of your libraries?

vkleen: I'm not using it in tf-ncl or json-schema-to-nickel.

vkleen: Git blame points to #683, so I think it was just an old oversight.

jneem: It turns out it was necessary, because we're using it in the integration tests, and it apparently the integration tests build the nickel crate without `cfg(test)`

yannham: I suspected something like that. Well, at lest let's get rid of the old comment

core/src/typecheck/mod.rs
jneem: I wasn't super keen on leaving a task that needed to be done before release, so I took a stab at the least permissive version of part 3: accept if and only if there are no flat types. If I understood it correctly, this is more permissive than the current rules but possibly less permissive than the eventual rule so it should be ok to release.

jneem:
Thanks for the review, I think this is ready for another look.

The main unanswered question is how this should interact with linearization/LSP. I'm planning to look into that next.

core/src/typecheck/mod.rs
yannham: Yup, sounds good :+1: I think it's fine to postpone a task to an immediately follow-up PR, just as a way of easing review and splitting work (not very different from splitting into commits), and not some vague unspecified later. But yeah the change isn't too big, so it's fine as it is

core/src/error.rs
```diff
@@ -293,6 +293,15 @@ pub enum TypecheckError {
         RichTerm, /* the inferred flat type */
         TermPos,
     ),
+    /// Types can appear in term position, like the `Number` in `let c = Number in 5 | c`.
+    /// These types are treated as contracts, but they are not allowed to contain flat types.
+    /// For example, `let c = { foo : 5 } in ...` is disallowed.
```
yannham, line 298: I think this comment is a tad confusing, because it never alludes to the fact that this restriction is only valid for statically typed code. In statically typed code, we also reject valid things, such as `MyContract -> Number`.

Conversely, we don't reject `let c = { foo : 5}` inside dynamically typed code. I think we could say something along the lines : 

> within statically typed code, the typechecker must reject terms containing nonsensical contracts such as `let C = { foo : 5} in ({ foo = 5 } | C)`, which will fail at runtime. The typechecker is currently quite conservative and simply forbids to store any custom contract (flat type) in type that appears in term position. Note that this restriction doesn't apply to annotations (so `{foo = 5} | {foo : 5}` is accepted), which are type positions, and aren't considered part of the statically typed block per se.

Might need some rewording.

core/src/error.rs
```diff
@@ -2052,6 +2061,25 @@ impl IntoDiagnostics<FileId> for TypecheckError {
                         ]),
                 ]
             }
+            TypecheckError::FlatTypeInTermPosition { flat, pos } => {
+                vec![
+                    Diagnostic::error()
+                        .with_message(
+                            "types containing user-defined contracts cannot be converted into contracts",
```
yannham, line 2068: I feel this message is a bit vague. But to be honest, this error is quite specific to our current implementation choices, so it's not too easy to get right in a succinct way. We can meditate on this and improve on later, so let's not block this PR on this.

core/src/eval/mod.rs
```diff
@@ -960,7 +965,11 @@ pub fn subst<C: Cache>(
         | v @ Term::SealingKey(_)
         | v @ Term::Enum(_)
         | v @ Term::Import(_)
-        | v @ Term::ResolvedImport(_) => RichTerm::new(v, pos),
+        | v @ Term::ResolvedImport(_)
+        // We could recurse here, because types can contain terms which would then be subject to
+        // substitution. Not recursing should be fine, though, because a type in term position
+        // turns into a contract, and we don't serialize contracts.
```
yannham, line 971: Just in case, because substitution isn't only used for serialization, I believe (for example to pretty print a result in the REPL)

```suggestion
        // turns into a contract, and we don't substitute inside contracts either currently.
```

core/src/typecheck/mod.rs
```diff
@@ -1589,6 +1590,13 @@ fn check<L: Linearizer>(
             );
             unify(state, &ctxt, ty, ty_import).map_err(|err| err.into_typecheck_err(state, rt.pos))
         }
+        Term::Types(typ) => {
+            if let Some(flat) = typ.find_flat() {
+                Err(TypecheckError::FlatTypeInTermPosition { flat, pos: *pos })
+            } else {
+                Ok(())
```
yannham, line 1597: I was going to say that you might need to `walk` the type at least there, but this may not be true since we know it doesn't contain any flat type, so `walk` is probably useless.

yannham (approved):

PR #1447, opened 2023-07-11
Improve query inteface

### LINKED ISSUE #1446: Unintuitive multiple ways to query a value in the repl
Let say that I want to query the documentation of `std.array.all`. The documentation says (`:help query`):

```
nickel> :help query
:query <identifier> [field path]
Aliases: `q`

Print the metadata attached to a field
<identifier> is valid Nickel identifier representing the record to look into.
<field path> is a dot-separated sequence of identifiers pointing to a field.

Example: `:query mylib contracts."special#chars".bar`
```

So, it seems I should do `:query std array.all`. The fact that I need to separate the first element of the path is a bit unintuitive, but ok. However, the documentation is lying, because in fact the first argument of query can be more complex, and be a path itself. So for example `:query std.array all` works as well, which is probably confusing.

Finally, if I put everything in the first argument, given how query works, this is accepted but I don't get all the information anymore:

```
nickel>:query std array.all
• type: forall a. (a -> Bool) -> Array a -> Bool
[...]

nickel>:query std.array all
• type: forall a. (a -> Bool) -> Array a -> Bool
[...]

nickel> :query std.array.all
```

The interface for `query` should just be to accept one path. Internally, the implementation can decide how to split this path and which part to evaluate or not, but that isn't relevant to the user. What you want to express is usually "give me information about this field", and "this field" is represented by one complete record path.

- Actually implementing `query` as advertised in the help message. This is still artificial (need to separate the first element of the path), but would at least error out on `:query std.array.all` instead of deceptively pretending there's no metadata for `all`.
- Keep the current status quo, which is IMHO confusing and deceptive (when using the full path version).

yannham:
Closes #1446.

Previously, the query command of the REPL would accept a first argument that could be a record access chain, followed by a second one which could also be a field path. This means that, for querying a given path `foo.bar.baz.blo`, there are 3 different but equivalent query invocations:

- `:query foo bar.baz.blo`
- `:query foo.bar baz.blo`
- `:query foo.bar.baz blo`

Moreover, `:query foo.bar.baz.blo` without a second argument wouldn't fail but pretend that there are no metadata, even if there are.

This PR simplifies the overall interface of the query command by just accepting one argument which is a field path.

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -59,7 +59,7 @@ impl QueryPath {
     /// # Post-conditions
     ///
     /// If this function succeeds and returns `Ok(query_path)`, then `query_path.0` is non empty.
-    /// Indeed, there's no such thing as a valid empty field path. If `input` is empty, or consist
+    /// Indeed, there's no such thing as a valid empty field path. If `input` is empty, or consists
     /// only of spaces, `parse` returns a parse error.
     pub fn parse(cache: &mut Cache, input: String) -> Result<Self, ParseError> {
         use crate::parser::{
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -55,6 +55,12 @@ impl QueryPath {
     /// Identifiers can be enclosed by double quotes when they contain characters that aren't
     /// allowed inside bare identifiers. The accepted grammar is the same as a sequence of record
     /// accesses in Nickel, although string interpolation is forbidden.
+    ///
+    /// # Post-conditions
+    ///
+    /// If this function succeeds and returns `Ok(query_path)`, then `query_path.0` is non empty.
+    /// Indeed, there's no such thing as a valid empty field path. If `input` is empty, or consist
+    /// only of spaces, `parse` returns a parse error.
     pub fn parse(cache: &mut Cache, input: String) -> Result<Self, ParseError> {
         use crate::parser::{
             grammar::FieldPathParser, lexer::Lexer, utils::FieldPathElem, ErrorTolerantParser,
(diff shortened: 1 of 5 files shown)


core/src/program.rs
```diff
@@ -55,6 +55,12 @@ impl QueryPath {
     /// Identifiers can be enclosed by double quotes when they contain characters that aren't
     /// allowed inside bare identifiers. The accepted grammar is the same as a sequence of record
     /// accesses in Nickel, although string interpolation is forbidden.
+    ///
+    /// # Post-conditions
+    ///
+    /// If this function succeeds and returns `Ok(query_path)`, then `query_path.0` is non empty.
+    /// Indeed, there's no such thing as a valid empty field path. If `input` is empty, or consist
```
vkleen, line 62:
```suggestion
    /// Indeed, there's no such thing as a valid empty field path. If `input` is empty, or consists
```

core/src/repl/mod.rs
```diff
@@ -415,15 +421,18 @@ pub fn print_help(out: &mut impl Write, arg: Option<&str>) -> std::io::Result<()
...
-                writeln!(out, ":{c} <identifier> [field path]")?;
+                writeln!(out, ":{c} <field path>")?;
                 print_aliases(out, c)?;
                 writeln!(out, "Print the metadata attached to a field")?;
                 writeln!(
                     out,
-                    "<identifier> is valid Nickel identifier representing the record to look into."
+                    "<field path> is a dot-separated sequence of identifiers pointing to a field. \
+                    Fields can be quoted if they contain special characters, \
+                    very much as in normal Nickel source code.\n"
```
vkleen, line 431:
```suggestion
                    just like in normal Nickel source code.\n"
```
This would sound better to me, I think.

vkleen (approved):

Issue #461, opened 2021-11-12
[Dogfooding] Nullable of several contracts (or, union of intersections)


yannham:
**Is your feature request related to a problem? Please describe.**
During dogfooding on the mantis repo, I had to convert CUE types that are morally unions of intersections, [as this one](https://github.com/input-output-hk/mantis-ops/blob/81ad4cb0ec5ff9faeb9065a2ddac060cef6452ac/schemas/nomad/types.cue#L59). In this example, `Weight` is either:

1. an unsigned int AND an int smaller than 100 AND an int greater than -100.
2. or `null`

Intersections alone (on simple contracts like this, at least: see the paper reference at the end for more info) can just be represented as several contract applications in Nickel. If we take `Weight` to be *NON* nullable, we can simply translate this as:

```
Weight | #nums.Nat
       | #(GreaterEq -100)
       | #(SmallerEq 100)
```

A nullabe alone is also not so bad. If we forget the bound `-100` and `100`, then we could write it like:

```
Weighy | #(Nullable nums.Nat)
```
Or even `Weight | #Nullable #nums.Nat` (under the proposal #445).

The problem is that when we mix everything, we cannot pass a series of contract application `| #nums.Nat | ...` to `Nullable`: that is, intersection of contracts is not first class. Currently we need to do something like:

```
let AllOf = fun contracts label value => lists.fold
  (fun c acc => c label acc) contracts value in
{
  Weight | #(Nullable (AllOf [nums.Nat, GreaterEq -100, SmallerEq 100]))
}
```

While not awful, it is quite more verbose and different than without the `Nullable`: in one case we just use standard contract application in sequence, but in the other we need a new combinator `AllOf`.

**Describe the solution you'd like**

I don't really know of a great solution. We can't have general union and intersection contracts for deep reasons: we actually [published a paper](https://dl.acm.org/doi/10.1145/3486602.3486767) about this very specific subject this year. So adding a syntax to do that, even in limited cases, could be very misleading and error prone in case of misuse. I just wanted to report this quirk to see if someone has a better solution on how to handle this.

regnat:
I’m a bit skeptical of the `x | #C1 | #C2 | #C3` way of writing intersections. As much as combining contracts that way makes sense for merging (say `(x | #C1) & (x | #C2)`), I feel like it’s a weird way of directly writing an intersection − because what we want is one contract that says “C1 and C2 and C3” − modulo the fact that it’s not _really_ an intersection, not a succession of each.
[...]

PR #1452, opened 2023-07-13
add std.contract.Sequence

Radvendii:
This new function isn't so useful for applying directly, but there;s currently no way to store multiple contracts in a variable to be applied later. In other words
```nickel
'foo | [| 'foo |]
```
is to
```nickel
let Foo = [| 'foo |] in
'foo | Foo
```
as
```nickel
"foo" | std.contract.TagOrString | [| 'foo |]
```
is to
???

Well, before there were no options. with this PR we can do
```nickel
let Foo = std.contract.Sequence [ std.contract.TagOrString, [| 'foo |] ] in
"foo" | Foo
```

Eventually, it would be nice if the LSP could recognize this and get the same niceties as if you'd directly applied the two contracts.

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -934,6 +934,27 @@
         "%
       = fun pred label value => if pred value then value else %blame% label,
 
+    Sequence
+      | doc m%"
+          Apply multiple contracts from left to right.
+
+          Type: `Array Contract -> Contract`
+          (for technical reasons, this function isn't actually statically typed)
+
+          # Examples
+
+          ```nickel
+          x | std.contract.Sequence [ Foo Bar ]
+          ```
+
+          is equivalent to
+
+          ```nickel
+          (x | Foo) | Bar
+          ```
+        "%
+      = fun contracts label value =>
+        std.array.fold_right (fun contract acc => std.contract.apply contract label acc) value contracts,
     label
       | doc m%"
           The label submodule provides functions that manipulate the label
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -957,13 +957,13 @@
           such as an argument to another contract:
 
           ```nickel
-          x | Array (std.contract.sequence [ Foo, Bar ])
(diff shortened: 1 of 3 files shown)


Radvendii:
related to #461

core/stdlib/std.ncl
```diff
@@ -934,6 +934,27 @@
...
+      | doc m%"
+          Apply multiple contracts from left to right.
+
+          Type: `Array Contract -> Contract`
+          (for technical reasons, this function isn't actually statically typed)
+
+          # Examples
+
+          ```nickel
+          x | std.contract.Sequence [ Foo Bar ]
```
vkleen, line 947:
```suggestion
          x | std.contract.Sequence [ Foo, Bar ]
```

vkleen (approved):
I like it, but I may be biased 😅

core/stdlib/std.ncl
```diff
@@ -934,6 +934,27 @@
         "%
       = fun pred label value => if pred value then value else %blame% label,
 
+    Sequence
+      | doc m%"
+          Apply multiple contracts from left to right.
```
yannham, line 939: Might be worth slightly more documentation, such as why do we need that if it's equivalent to just apply several contracts with `x | Foo | Bar`. A simple example with `Array (std.contract.sequence [Foo, Bar])` that can't be expressed currently with simple `|`s could be sufficient.

core/stdlib/std.ncl
```diff
@@ -934,6 +934,27 @@
...
+          ```
+
+          is equivalent to
+
+          ```nickel
+          (x | Foo) | Bar
+          ```
+        "%
+      = fun contracts label value =>
+        std.array.fold_right (fun contract acc => std.contract.apply contract label acc) value contracts,
```
yannham, line 957:
```suggestion
        std.array.fold_right (fun contract => std.contract.apply contract label) value contracts,
```

yannham (approved):

PR #1454, opened 2023-07-15
Add linker directive for *-darwin to fix PyO3

nrhtr:
Without these flags I get linking errors like this when building from nixpkgs:

```
  = note: Undefined symbols for architecture arm64:
            "_PyExc_ConnectionResetError", referenced from:
                pyo3::type_object::PyTypeInfo::type_object::hd73bc9c256c52717 in libpyo3-c59aeb8b74bb57e0.rlib(pyo3-c59aeb8b74bb57e0.pyo3.e60578ba-cgu.0.rcgu.o)
```

See: https://pyo3.rs/main/building_and_distribution#macos


vkleen:
I'm wondering if it wouldn't be better to use the build script option mentioned in the documentation you linked. The cargo config might not be picked up by the nix build. How exactly did you try building from nixpkgs to get these linker errors?

nrhtr:
I just ran `nix build nixpkgs#nickel`. Interestingly, I don't get any errors building the flake in this repo, only in nixpkgs. I guess the build is quite different but I'm not sure of the exact reason.

It does appear to be picked up by the nixpkgs build. I tested changing the derivation source to 1.1.1 w/ this patch and it worked fine.

I can change it to use the build script option. I wasn't sure if it would impact other platforms somehow... 🤷‍♂️

vkleen:
The build script version feels a bit cleaner to me. If we do that, it should be a new build script in the `pyckel` subdirectory. At least that would seem sensible to me without having tested it.

This makes me wonder why the nixpkgs derivation is trying to build the python bindings at all. On the face of it they seem to only request building `nickel-lang-cli` but the build logs do show `pyckel` being build.

Anyways, thanks for finding this and making a PR!

vkleen:
The `build.rs` version doesn't seem to impact the build on x86_64-linux, from a quick experiment locally. So I'd prefer that one.

yannham:
@vkleen could you have the occasion to test this on `*-darwin`? @nrhtr , does the latest version of this PR appear to solve the issue ? Thanks!

vkleen:
I checked on `aarch64-darwin`. It solves the issue there.

jneem (approved):

PR #1455, opened 2023-07-17
Static linking with `nickel format`

vkleen:
This is the result of experiments together with @jneem and @Radvendii. Using the musl libc and clang with libc++ we can statically link a Nickel binary which includes the new `format` feature.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -262,7 +262,7 @@
                   CARGO_BUILD_TARGET = pkgs.rust.toRustTarget pkgs.pkgsMusl.stdenv.hostPlatform;
                   # For some reason, the rust build doesn't pick up the paths
                   # to `libcxx` and `libcxxabi` from the stdenv. So we specify
-                  # them explicitly. Also, `libcxx` expects to be linked fith
+                  # them explicitly. Also, `libcxx` expects to be linked with
                   # `libcxxabi` at the end, and we need to make the rust linker
                   # aware of that.
                   RUSTFLAGS = "-L${pkgs.pkgsMusl.llvmPackages.libcxx}/lib -L${pkgs.pkgsMusl.llvmPackages.libcxxabi}/lib -lstatic=c++abi";
--- a/flake.nix
+++ b/flake.nix
@@ -241,10 +241,7 @@
         rec {
           inherit cargoArtifacts;
           nickel-lang-core = buildPackage { pname = "nickel-lang-core"; };
-          nickel-lang-cli = buildPackage {
-            pname = "nickel-lang-cli";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -241,18 +241,38 @@
         rec {
           inherit cargoArtifacts;
           nickel-lang-core = buildPackage { pname = "nickel-lang-core"; };
-          nickel-lang-cli = buildPackage { pname = "nickel-lang-cli"; };
+          nickel-lang-cli = buildPackage {
+            pname = "nickel-lang-cli";
+            extraBuildArgs = "--features format";
```
yannham, line 246: If we want  format support by default, I think we should rather make it a default feature in `Cargo.toml`, to 1. avoid spreading feature configuration across Nix and Rust 2. have `cargo install` and `cargo build` produce the same binary as `nix build` by default.

flake.nix
```diff
@@ -241,18 +241,38 @@
...
+            # tried building with libstdc++ but without success.
               buildPackage {
                 pname = "nickel-lang-cli";
+                extraBuildArgs = "--features format";
                 extraArgs = {
                   CARGO_BUILD_TARGET = pkgs.rust.toRustTarget pkgs.pkgsMusl.stdenv.hostPlatform;
-                  CARGO_BUILD_RUSTFLAGS = "-C target-feature=+crt-static";
+                  # For some reason, the rust build doesn't pick up the paths
+                  # to `libcxx` and `libcxxabi` from the stdenv. So we specify
+                  # them explicitly. Also, `libcxx` expects to be linked fith
```
yannham, line 265:
```suggestion
                  # them explicitly. Also, `libcxx` expects to be linked with
```

flake.nix
```diff
@@ -241,18 +241,38 @@
...
+          nickel-lang-cli = buildPackage {
+            pname = "nickel-lang-cli";
+            extraBuildArgs = "--features format";
+          };
           lsp-nls = buildPackage { pname = "nickel-lang-lsp"; };
 
+          # Static building isn't really possible on MacOS because the system call ABIs aren't stable.
           nickel-static =
             if pkgs.stdenv.hostPlatform.isMacOS
             then nickel-lang-cli
```
yannham, line 253: Should we silently fall back to non static here, or rather `throw` that static binaries aren't supported on MacOS? Wouldn't the former be at risk of being misleading?

vkleen: My main concern was the MacOS CI checking that all derivations in `packages` actually evaluate successfully. I suppose the most principled option would be to remove the `nickel-static` output for `*-darwin` systems.

yannham (approved):

Issue #1361, opened 2023-06-14
support scientific notation for numbers


MMesch:
**Is your feature request related to a problem? Please describe.**

I am always frustrated when I can't write numbers as `1e12` and have to write `1000000000000` instead. It's easy to make a mistake that way :)

**Describe the solution you'd like**

Support for scientific notation for numbers such as `1e12` would be a good solution, in my opinion.

yannham:
I would add to the mix: should we also support common computer units? For example, I think you can write `12MiB` or `12MB` in CUE, which can come in handy to configure e.g. space allocation.

yannham:
And second suggestion: we should probably support a delimiter, such as `_`, to make numbers like `1000000001` more readable, like `1_000_000_001`.

vkleen:
For scientific notation, `malachite` already supports converting from scientific notation, so we'll only need to change the lexer to allow this.

yannham:
Let's focus on scientific notation then, and open other issues accordingly for the other issues.

vkleen:
With #1456 we have scientific notation in Nickel.

PR #1456, opened 2023-07-17
Make the lexer accept scientific notation

vkleen:
Since we were alread using `malachite`'s scientific notation parser for parsing numbers, the only thing stopping #1361 was the lexer not supporting scientific notation. This would technically be a breaking change, since previously `f 1e6` would be parsed as `f` applied to the number `1` and the variable `e6`, while now it would be parsed as `f` applied to one million. I hope that nobody relies on this behaviour but it's worth considering, given our (informal, for now) stability guarantee.


yannham (approved):
Might be worth a documentation update. The fact that we are currently parsing `1e6` as `1 e6` is both useless and should be considered a bug in the current implementation (not very pressing, but still), so I don't think it should be part of the informal backward compatibility guarantee.

PR #1463, opened 2023-07-19
Evaluate terms before extracting documentation

### LINKED ISSUE #1462: `nickel doc` appears not to work for `Let x in y` expressions
Hello - running `nickel doc -f test.ncl --stdout` for this: 
```
# test.ncl
{
  port
    | doc "a  port"
}
```
works fine, and outputs the documentation I'd expect. But if I change the file to this: 
```
# test.ncl
let Port = {
  port
    | doc "a  port"
}
in Port
```
then I get an error: 
```
error: no documentation found
  ┌─ test.ncl:1:1
  │  
1 │ ╭ let Port = {
2 │ │   port
3 │ │     | doc "a  port"
4 │ │ }
5 │ │ in Port
  │ ╰───────^
  │  
  = documentation can only be collected from a record.
```
I'd expect to get the same behaviour from both of these examples, I'd have thought that the second one would just resolve to the record shown in the first one.

Unless there's something I'm missing, this stops documentation being generated for schemas which use imports.

Thanks!

vkleen:
With this change, `nickel doc` evaluates a term before trying to extract
documentation. The evaluation mode needed here is somewhat between evaluating to
WHNF and a full deep evaluation. Namely, we first evaluate the term to WHNF. If
it isn't a record then there is no documentation to extract and we're done. If
it is, we iterate through all fields with a defined value and call the evaluator
recursively.

The result will be a tree of records in WHNF which our previous documentation
extraction procedure knows how to handle. In fact, the only difference to a full
deep evaluation is that we do not descend into arrays. Consequently, this change
will incur a performance penalty compared to the previous `nickel doc`. For
now, we're willing to accept this penalty for two reasons. First, documentation
extraction is expected to be rarer than fully evaluating a Nickel configuration.
Second, the evaluation algorithm presented here does strictly less work than a
full evaluation and therefore shouldn't be any slower.

There is another caveat with this implementation. For simplicity, I've
implemented an explicitly recursive version of the algorithm described above.
This means that we're using the Rust callstack when descending into deeply
nested records and we might encounter stack overflows for particularly deep
records. The Nickel evaluator itself uses a custom stack stored in the heap to
circumvent this problem. If we ever encounter stack overflows in `nickel doc` in
the wild, we should do the same here.

Fixes #1462.

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -259,18 +259,45 @@ impl<EC: EvalCache> Program<EC> {
         String::from_utf8(buffer.into_inner().into_inner()).unwrap()
     }
 
+    /// Evaluate a program into a form suitable for extracting documentation.
+    /// This needs to evaluate to WHNF, but then still descend into subfields
+    /// whose values are records.
+    #[cfg(feature = "doc")]
+    pub fn eval_for_docs(&mut self) -> Result<RichTerm, Error> {
+        use crate::eval::{Closure, Environment};
+        use crate::match_sharedterm;
+        use crate::term::Term;
+
+        let (t, initial_env) = self.prepare_eval()?;
+
+        fn do_eval<EC: EvalCache>(
+            vm: &mut VirtualMachine<Cache, EC>,
+            t: RichTerm,
+            initial_env: &Environment,
+        ) -> Result<RichTerm, Error> {
+            vm.reset();
+            let (rt, env) = vm
+                .eval_closure(Closure::atomic_closure(t), initial_env)
+                .map_err(Error::from)?;
+            let eval_field =
+                |_id, rt: RichTerm| -> RichTerm { do_eval(vm, rt.clone(), &env).unwrap_or(rt) };
+
+            match_sharedterm! {rt.term, with {
+                    Term::Record(data) => Ok(Term::Record(data.map_defined_values(eval_field)).into())
+                } else Ok(rt)
+            }
+        }
+
+        do_eval(&mut self.vm, t, &initial_env)
+    }
+
     /// Extract documentation from the program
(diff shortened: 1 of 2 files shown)


core/src/program.rs
```diff
@@ -259,18 +259,72 @@ impl<EC: EvalCache> Program<EC> {
         String::from_utf8(buffer.into_inner().into_inner()).unwrap()
     }
 
+    /// Evaluate a program into a form suitable for extracting documentation.
+    /// This needs to evaluate to WHNF, but then still descend into subfields
+    /// whose values are records.
+    #[cfg(feature = "doc")]
+    pub fn eval_for_docs(&mut self) -> Result<RichTerm, Error> {
```
yannham, line 266:
```suggestion
    pub fn eval_for_doc(&mut self) -> Result<RichTerm, Error> {
```

core/src/program.rs
```diff
@@ -259,18 +259,72 @@ impl<EC: EvalCache> Program<EC> {
...
+                    },
+                    initial_env,
+                )
+                .map_err(Error::from)?;
+
+            match_sharedterm! {rt.term, with {
+                    Term::Record(data) => {
+                        let fields = data
+                            .fields
+                            .into_iter()
```
yannham, line 295: I would have thought you would need to do some substitution at some point, but it turns out `eval_closure` is already evaluating variables away, mapping over fields is enough. Nice!

yannham (approved):
As discussed before, this approach sounds reasonable. We might actually slower than evaluation in some cases (we reset the machine, we map over fields, etc.), but most probably still close to the performance of full evaluation in the worst case, so it's fine.

Issue #1431, opened 2023-07-05
Poor man's support of Nix interoperability


thufschmitt:
(Mostly a wild idea so far, testing the water to get an idea of how unreasonable it is. Also, this isn't really Nix-specific and could potentially be adapted to any language)

**Is your feature request related to a problem? Please describe.**

For https://github.com/nickel-lang/nickel-nix, having a first-class(ish) Nix interoperability would be a must. @francois-caddet started an ambitious effort to make Nickel able to evaluate Nix natively. However, there's (afaik) still a bunch of work to be done for this to fully work on Nixpkgs (which is the real goal here). So the best solution so far is a somewhat contrived and rigid one using JSON as the interop, and requiring a few back-and-forth between the different languages.

**Describe the solution you'd like**

As a first-order approximation, a Nickel unary operator `%import_nix% : String -> Dyn` that would evaluate a Nix expression (using Nix's own evaluator), and return its result as a Nickel value, replacing the functions and other non-representable values by errors.

This wouldn't be enough for a fully transparent interface, because this interface would be strict, while Nix expressions leverage heavily the laziness of the language. However, we could emulate this laziness with another primitive operator `%nix_thunk%` that takes a pointer to a (lazy) Nix value, and returns it in whnf, with all the thunks replaced by calls to `%nix_thunk%`. So for instance,

```nickel
%import_nix% "{ x = 1; y = 1+1; }"
```

would evaluate to

```nickel
{ x = 1, y = %nix_thunk% {a pointer to the Nix value representing the thunk "1+1"} } 
```

which could be further forced to

```nickel
{ x = 1, y = 2 }
```

The advantage of this approach is that the integrations is somewhat light as it “only” requires adding two builtins, meaning that it could be made optional (arguably not everyone cares about pulling in Nix when building Nickel) or even a plugin if there was such a mechanism.

**Describe alternatives you've considered**

- Fully-fledged Nix evaluator in Nickel. Cleaner (and would support functions), but bigger effort
- Keep (and try to improve) the current JSON-based FFI. I'll try to open another issue for suggestions here as we could maybe make it a bit better with some scoped changes to Nickel
- Rewrite Nixpkgs in Nickel

yannham:
[...]

PR #1465, opened 2023-07-20
add %eval_nix% primop

Radvendii:
add an %eval_nix% primop which calls out to the nix C++ bindings. At the moment, this goes through JSON and therefore requires the nix to fully evaluate

Partly addresses #1431


core/build.rs
vkleen, line 1: Usually, people use something like https://crates.io/crates/pkg-config to find include and library paths for external C and C++ libraries. I don't know if nix includes pkg-config files that would make that work. In any case, `build.rs` should add those paths somehow instead of using `-isystem` in `NIX_CFLAGS_COMPILE`.

core/src/term/mod.rs
```diff
@@ -1088,6 +1088,10 @@ pub enum UnaryOp {
     /// polymorphic contracts to check if they are being invoked to generate
     /// a dual contract, as part of the preliminary fix for [#1161](https://github.com/tweag/nickel/issues/1161).
     Dualize(),
+
+    /// Evaluate a string of nix code into a resulting nickel value. Currently completely (strictly) evaluates the nix code, and must result in a value serializable into JSON.
+    #[cfg(feature = "nix-experimental")]
+    EvalNix(),
```
vkleen, line 1094: If we gate the existence of this variant behind `nix-experimental` but we aren't able to gate the lalrpop parser, then the code probably won't compile without `nix-experimental` anymore. I suppose the workaround would be to not gate the existence of `UnaryOp::EvalNix` but use `unimplemented!()` when evaluating in case `nix-experimental` isn't enabled.

core/build.rs
Radvendii: Good call on using `pkg-config`. with a little finagling, it works.

core/src/term/mod.rs
Radvendii: It now works with the feature on or off. If the feature is off it will produce an error saying that you have to recompile with `--feature nix-experimental`

flake.nix
```diff
@@ -225,8 +224,7 @@
             cargoExtraArgs = "${cargoBuildExtraArgs} --workspace";
             # pyo3 needs a Python interpreter in the build environment
             # https://pyo3.rs/v0.17.3/building_and_distribution#configuring-the-python-version
-            # XXX: I still need to add "-isystem ${pkgs.nix}/include/nix" to $NIX_CFLAGS_COMPILE
-            buildInputs = [ pkgs.python3 pkgs.nixVersions.nix_2_16 pkgs.boost ];
+            buildInputs = with pkgs; [ python3 pkg-config nixVersions.nix_2_16 boost ];
```
vkleen, line 227: Cross-building is probably broken for other reasons already, but `pkg-config` should be in `nativeBuildInputs`. That might be true for `python3` as well, or that might need to be in both `buildInputs` and `nativeBuildInputs` for `pyo3`.

flake.nix
```diff
@@ -328,6 +328,8 @@
         inputsFrom = builtins.attrValues (mkCraneArtifacts { inherit rust; });
 
         buildInputs = [
+          # pkgs.nix.dev
+          # pkgs.boost # implicit dependency of Nix
```
vkleen, line 332:
```suggestion
```

vkleen (approved):

vkleen (approved):
I wonder if we can make the flake outputs more configurable to avoid the heavy dependency footprint in e.g. `nix run github:tweag/nickel` if one is just looking for a minimal Nickel interpreter. But that's probably better tackled in a separate PR.

core/src/error.rs
```diff
@@ -493,6 +496,10 @@ pub enum ParseError {
         /// The previous instance of the duplicated identifier.
         prev_ident: LocIdent,
     },
+    DisabledFeature {
```
yannham, line 499: Even if short, a description comment would be nice, in line with other's variant.

core/src/error.rs
```diff
@@ -1820,6 +1836,11 @@ impl IntoDiagnostics<FileId> for ParseError {
                         secondary(&prev_ident.pos.unwrap()).with_message("previous binding here"),
                         primary(&ident.pos.unwrap()).with_message("duplicated binding here"),
                     ]),
+            ParseError::DisabledFeature { feature, span } =>
+                Diagnostic::error()
+                    .with_message("Interpreter compiled without required features")
```
yannham, line 1841:
```suggestion
                    .with_message("interpreter compiled without required features")
```

core/src/error.rs
```diff
@@ -1820,6 +1836,11 @@ impl IntoDiagnostics<FileId> for ParseError {
...
+                    .with_labels(vec![primary(&span).with_message(format!("this syntax is only supported with the `{}` feature", feature))])
```
yannham, line 1842:
```suggestion
                    .with_labels(vec![primary(&span).with_message(format!("this syntax is only supported with the `{}` feature enabled", feature))])
```

core/src/nix_ffi/cpp/nix.cc
```diff
@@ -0,0 +1,44 @@
...
+#include <nix/shared.hh>
+#include <nix/canon-path.hh>
+#include <nix/nixexpr.hh>
+#include <nix/value-to-json.hh>
+#include <nix/command.hh>
+#include <nix/value.hh>
+
+#include "nickel-lang-core/src/nix_ffi/mod.rs.h"
+
+struct DummyEvalCommand : virtual EvalCommand {
```
yannham, line 17: Could be worth documenting why this is needed

flake.nix
```diff
@@ -209,7 +216,11 @@
             cargoExtraArgs = "${cargoBuildExtraArgs} --workspace";
             # pyo3 needs a Python interpreter in the build environment
             # https://pyo3.rs/v0.17.3/building_and_distribution#configuring-the-python-version
-            buildInputs = [ pkgs.python3 ];
+            nativeBuildInputs = with pkgs; [ pkg-config python3 ];
+            buildInputs = with pkgs; [
+              nix-input.packages.${system}.default
+              boost # implicit dependency of nix
```
yannham, line 222: Wait, you mean that Nix has implicit dependencies...in Nix? :scream:

yannham (approved):
Nice! The footprint isn't as bad as I thought (it doesn't add to much nix-experimental-related code).

yannham:
@Radvendii is there anything left to do to merge this PR (beside rebasing, now there's a conflict)? I fear this drifts further away from master

Radvendii:
There's a phantom bug which only occurs in CI, Not even on my local x86_64-darwin machine which is using the same version of `nix` to build , and building the same derivation.

I tried to track down the phantom bug using valgrind, but the nix package for valgrind is broken, and when I used an old version it didn't actually work. It seems from internet searches that valgrind is just kind of broken on mac os?

Given that this is only a problem with a single test, I've [filed the bug with upstream](https://github.com/NixOS/nix/issues/9107) and disabled tests for now.

PR #1470, opened 2023-07-21
Error on serializing very large numbers

### LINKED ISSUE #1457: "Nearest" rounding for large numbers is surprising
Something that has only really become a concern with the new support for scientific notation:
```
nickel> std.serialize 'Json 1.7e400
"1.7976931348623157e308"
```
This should really be a serialization error, because the number doesn't fit into the specified 64bit float. Just clamping it to the maximum representable number may be technically valid ("the closest float64 to the input number") but it's very surprising.

This is a result of us using `malachite`'s `f64:rounding_from(n, RoundingMode::Nearest)` for serializing numbers that are not representable as 64-bit integers. `RoundingMode::Nearest` is explicitly specified to always return the nearest finite floating point number, which is of course `f64::MAX` for very large inputs.

vkleen:
Instead of blindly using the nearest `f64` for all Nickel `Number`s when serializing, with this change we only use nearest rounding when a number is in the range `[f64::MIN, f64::MAX]`. Otherwise we trigger a newly introduced serialization error.

This stops the previous, arguably surprising, result that for example `e400` gets serialized as `1.7976931348623157e308`. While techincally this is the nearest `f64` to `1e400`, one wouldn't expect serialization to incur an absolute error of about `1e92`.

Fixes #1457


yannham (approved):
LGTM. We could add a snapshot test, to see how it looks, but I let you decide if it's useful.

vkleen:
Ah, thanks for reminding me. I had wanted to add snapshot and integration tests but evidently I forgot.

PR #1471, opened 2023-07-21
[Refactor] Introduce RemoveRow trait

yannham:
The free-standing functions `erows_add` and `rrows_add` are very similar. The name and documentation was also misleading: what those methods do is rather to remove a row than adding a new one (although they first ensure the row is present, which might lead to adding it first).

This refactoring put them in their own trait, `RemoveRow`, and update the documentation accordingly. Doing so, we also get rid of an `Option` wrapper for a parameter of `UnsatConstr` error, which was unwrapped when reporting anyway.

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -257,7 +257,7 @@ pub enum TypecheckError {
     /// the direct failure to unify `{ .. , x: T1, .. }` and `{ .., x: T2, .. }`.
     RowConflict(
         Ident,
-        /* the second type assignment which violates the constraint */ Option<Types>,
+        /* the second type assignment which violates the constraint */ Types,
         /* the expected type of the subexpression */ Types,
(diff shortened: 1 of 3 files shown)


jneem (approved):
The refactor lgtm. I'm not really sure about the changes to `UnifRecordRows` (in particular, why we now create a fresh unification variable whereas before we returned the passed-in type), but that's mostly just down to me not really understanding the type-checking algorithm yet...

yannham:
I would say this change is mostly cosmetic, in order to avoid having to pass the original type to `remove_row`, which I found to be a bit confusing when trying to understand the interface. Now, when the row type is extended, we extend it with a unification variable, that we unify just after with the type that was originally passed to `rrow_add`. There is a small additional step, but after the unification, the result should be strictly the same.

Issue #1027, opened 2023-01-02
Type unification error when using a `forall` inside a row type


matthew-healy:
**Describe the bug**

This program:

```
let r : { id : forall a. a -> a } = { id = fun r => r } in
r.id "x"
```

currently fails with an "incompatible rows declaration" error, failing to unify `forall a. a -> a` with `_a -> _a` (where `_a` is a unification variable).

**Expected behavior**

These types should successfully unify, and the code should typecheck.

yannham:
I think this is an instance of, or at least closely related to #360, and would be fixed by closing #584.

yannham:
Closed probably by #1473. The issue is anyway not present anymore in current 1.2.1.

PR #1473, opened 2023-07-24
Proper type instantiation discipline

### LINKED ISSUE #584: Implement a proper bidirectional algorithm

yannham:
Closes #584 #360 #1027

This PR is the last missing piece toward a reasonable bidirectional typechecking algorithm that properly handles higher-rank polymorphic types.

Until now, instantiation of polymorphic types has been sometimes ad-hoc and bizarrely limited. Simple programs such as:

```nickel
let r : { id : forall a. a -> a } = { id = fun r => r } in
r.id "x"
```

or

```nickel
let eval : forall a. (forall b. b -> b) -> a -> a = fun f x => f x in
(eval (fun x => x) 1) : Number
```

wouldn't typecheck (cf related issues at the top). Variable levels introduced in #1372 were an important preliminary step before we could attack this issue.

This PR is building on #1372 to use a proper instantiation discipline, derived from [Complete and Easy Bidirectional Typechecking for Higher-Rank
Polymorphism](https://www.cl.cam.ac.uk/~nk480/bidir.pdf) (ideally from the specification of the type system in this repository, but it's unfortunately not complete enough yet to be used as a reliable source of truth).

This commit gets rid of the ad-hoc cases of instantiation for `Op1` applications, the variable rule and others: those are now all subsumed by the subsumption rule, because they are special cases of a switch from inference mode to checking mode. Now, `check` and `subsumption` are the only place where the checked type might be instantiated with rigid type variable (I suspect we might even get rid of the instantiation of the checked type in `subsumption`, because it's always done by `check` before anyway, but this has to be verified first), and `subsumption` is the only rule to instantiate an inferred type with fresh unification variables.

The result is that the previously mentioned cases, plus other examples of higher-rank polymoprhic functions are now typechecking just fine.

Funnily, one test in `typecheck/fail` was actually well-typed, but didn't typecheck because of the described previous limitations. It's just been transformed to a passing test.

## TODO

- [ ] actually enforce that the instantiation is predicative (mark unification variables coming from the instantiation of a polytype and avoid unifying them with non instantiated polytypes). Currently we do naive impredicative instantiation, but it's just because I didn't want to load this PR too much: it's going to be predicative before the next release. Not that the current version isn't unsound, but doing naive impredicative instantiation makes typechecking sensitive to arguments order for such impredicative cases.


core/src/typecheck/mod.rs
```diff
@@ -2106,16 +2055,26 @@ fn check<L: Linearizer>(
...
-/// Currently, there is no subtyping (until RFC004 is implemented), hence this function simply
-/// performs unification (put differently, the subtyping relation is the equality relation). In the
-/// future, this function might implement a non-trivial subsumption rule.
+/// Currently, there is no subtyping (until RFC004 is implemented), hence this function performs
+/// polymorphic type instantiation with unification variable on the left (on the inferred type),
+/// and then simply performs unification (put differently, the subtyping relation is the equality
+/// relation).
+///
+/// The type instantiation corresponds to the zero-ary case of application in the current
+/// specification (which is based on [A Quick Look at Impredicativity][quick-look], also we
```
yannham, line 2064:
```suggestion
/// specification (which is based on [A Quick Look at Impredicativity][quick-look], although we
```

core/src/typecheck/mod.rs
```diff
@@ -2235,32 +2194,107 @@ fn check_with_annot<L: Linearizer>(
...
+            )?;
+
+            Ok(ty_res)
+        }
+        Term::App(e, t) => {
+            // If we go the full Quick Look route (cf [quick-look] and the Nickel type system
+            // specification), we will have a more advanced and specific rule to guess the
+            // instantiation of the potentially polymorphic type of the head of the application.
+            // Currently, we limit ourselves to predicative instantiation, and we can get away
+            // by eagerly instantiating heading `foralls` with fresh unification variables.
```
yannham, line 2270:
```suggestion
            // with eagerly instantiating heading `foralls` with fresh unification variables.
```

core/src/typecheck/mod.rs
```diff
@@ -2129,46 +2088,57 @@ fn check_field<L: Linearizer>(
...
-        field.value.as_ref().map(|v| v.pos).unwrap_or(id.pos),
-    )
+    // If there's no annotation, we simply check the underlying value, if any.
+    if field.metadata.annotation.is_empty() {
+        if let Some(value) = field.value.as_ref() {
+            check(state, ctxt, lin, linearizer, value, ty)
+        } else {
+            // It might make sense to accept any type for a value without definition (which would
+            // act a bit like a function parameter). But for now, we play safe and implement a more
+            // restrictive rule, which is that a value without a definition has type `Dyn`
```
vkleen, line 2098: Is this a change in behaviour or was this just implicit in the previous typechecking code?

vkleen (approved):
This looks reasonable, with the caveat that I'm not an expert in the current Nickel typechecker 😅

core/src/typecheck/mod.rs
yannham: This is a change in behavior, but I think the previous behavior was accidental, and neither officially supported nor documented. The previous behavior was indeed to let the type of fields without values to be unconstrained.

PR #1474, opened 2023-07-25
Fix REPL panic on transitive imports

Radvendii:
I also refactored one of the functions i had trouble understanding. I don't feel too strongly about it though if someone else objects to the refactor.

Something that took me a while to track down, and I still am having a little trouble understanding:
We can safely call `unwrap_error()` because the only case where `.resolve_imports()` returns `NotParsed` is when the call to `entry_state()`, in other words `terms.get()` is `None`.

And we've already called `import-resolution::strict::resolve_imports()`, and if I'm reading the code correctly everything passed back in `resolved_ids` must have been put in `terms` (ultimately by `parse_lax_multi()`)

My confusion is... we haven't parsed the file at that point though, right? We clearly parse them during hte call to `.resolve_imports()`, because that's where parse errors are being generated. But they must already have state `>= EntryState::Parsed` before that call.

---

Also, do we have any mechanism for testing the repl?

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -688,69 +688,69 @@ impl Cache {
         file_id: FileId,
     ) -> Result<CacheOp<(Vec<FileId>, Vec<ImportError>)>, CacheError<ImportError>> {
         match self.entry_state(file_id) {
-            Some(state) if state >= EntryState::ImportsResolved => {
-                Ok(CacheOp::Cached((Vec::new(), Vec::new())))
-            }
-            Some(state) if state >= EntryState::Parsed => {
-                let (pending, errors) = if state < EntryState::ImportsResolving {
-                    let TermEntry { term, .. } = self.terms.get(&file_id).unwrap();
-                    let term = term.clone();
+            Some(EntryState::Parsed) => {
+                let TermEntry { term, .. } = self.terms.get(&file_id).unwrap();
+                let term = term.clone();
 
-                    let import_resolution::tolerant::ResolveResult {
-                        transformed_term,
-                        resolved_ids: pending,
-                        import_errors,
-                    } = match self.error_tolerance {
-                        ErrorTolerance::Tolerant => {
-                            import_resolution::tolerant::resolve_imports(term, self)
-                        }
-                        ErrorTolerance::Strict => {
-                            import_resolution::strict::resolve_imports(term, self)?.into()
-                        }
-                    };
-
-                    // unwrap!(): we called `unwrap()` at the beggining of the enclosing if branch
-                    // on the result of `self.terms.get(&file_id)`. We only made recursive calls to
-                    // `resolve_imports` in between, which don't remove anything from `self.terms`.
-                    let cached_term = self.terms.get_mut(&file_id).unwrap();
-                    cached_term.term = transformed_term;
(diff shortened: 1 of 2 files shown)


yannham:
Is there an issue linked to this PR? A small sum up of what the problem was and how it was solved could help a bit, because honestly I have a bit of hard time understanding the description as it is.

yannham:
Thanks @jneem . I still think a short summary of why the bug occurred (which isn't really contained in the issue) and how it has been fixed would be useful

Radvendii:
There are two kinds of errors that could in theory occur at that point in the code, `NotParsed` and some error that resulted from the parsing. Having looked through it carefully, I'm reasonably sure `NotParsed` can't occur, which is maybe why someone felt comfortable putting an `.unwrap()` there, but the regular parsing errors were still possible.

So I changed it to propagate those errors instead of ignoring them. The whole change is essentially just this:
```rust
.unwrap()
becomes
.map_err(|cache_err| { cache_err.unwrap_error("repl::eval_(): expected imports to be parsed")})?
```

core/src/cache.rs
```diff
@@ -688,69 +688,69 @@ impl Cache {
...
+            // Nickel pipeline should however fail if `resolve_imports` failed at some
+            // point, anyway.
+            Some(EntryState::ImportsResolving) => Ok(CacheOp::Done((Vec::new(), Vec::new()))),
+            // >= EntryState::ImportsResolved
+            Some(
+                EntryState::ImportsResolved
+                | EntryState::Typechecking
+                | EntryState::Typechecked
+                | EntryState::Transforming
+                | EntryState::Transformed,
```
yannham, line 751: I believe one case such as `Some(state) if state >= EntryState::ImportResolved` is equally clear and a bit less verbose. Enumerating the cases has the advantages that if we ever add a phase, this forces us to update this part of this code (and thus to think about what we should do), but I fee like the ordering is enough here: whatever are the next phases, if we're past `ImportResovled`, there's nothing to do.

I really don't feel strongly about it though, feel free to ignore

core/src/repl/mod.rs
```diff
@@ -113,7 +113,12 @@ impl<EC: EvalCache> ReplImpl<EC> {
             resolved_ids: pending,
         } = import_resolution::strict::resolve_imports(t, self.vm.import_resolver_mut())?;
         for id in &pending {
-            dbg!(self.vm.import_resolver_mut().resolve_imports(*id)).unwrap();
```
yannham, line 116: Ah, an `unwrap` without explaining why it's assumed to be ok...good example that this is a bad thing :upside_down_face:

core/src/cache.rs
Radvendii: I agree that would be better. The disadvantage is that then rust can't tell the pattern is exhaustive, so we have to either add a `_ => unreachable!()` or change `None` to a wildcard match. Before, `None` was a wildcard match, and that was one of the things that confused me. ("what can this be besides None?? Oh, it can't.")

yannham: Oh, I see. Haven't thought about that. If it confused you, it's a sign that it might be better indeed (given than this version is more verbose but not unclear).

PR #1475, opened 2023-07-25
set record arguments with --

### LINKED ISSUE #1408: CLI arguments to pass data to and retrieve data from Nickel configurations
It often happens that some configuration values are not known when writing Nickel code but only when we finally run the `nickel` CLI invocation. Or that those values vary. See for example #1202 and #413. Another related example is secrets (an SSH key or other credentials) that you don't want to hardcode into the configuration but to pass either via the CLI or environment variables.

We might also want to override quickly the field of an existing configuration on-the-fly, without having to write a separate file which just imports the old one and do the override.

Similarly, it may happen that the configuration we want to extract is just part of a larger one: we need `<config>.ubuntu.system` exported to a YAML file.

Currently, the only existing solution to do those kind of things is to inject Nickel code in the `nickel` invocation: for example:

```console
$ nickel export <<< 'let as_function = import "base.ncl" in as_function "somearg" "coming from CLI"'
...
$ nickel export <<< 'let as_record = import "record-base.ncl" in as_record & {foo.bar.baz = "set/overridden"}'
...
$ nickel export <<< '(import "big-config.ncl").ubuntu.system'
```

While possible today, those operations aren't very elegant. They require to generate Nickel code with boilerplate (importing the original file, and applying whatever operations needed, such as merging), and can also have security implications: if a field is set to a value coming from an external source, the Nickel part is subject to Nickel injections (see #1202).

The proposal to add new CLI arguments to ease such operations: the overall goal is to make Nickel configuration _composable_ with respect to shell operations, not only between Nickel configurations themselves.

Typically, one would have at least one argument to set a value of a configuration and one to select one output of the configuration:

```console
$ nickel export -f "base.ncl" --bikeshed-arg-for-function '"somearg"' --bikeshed-arg-for-function '"coming from CLI"'
...
$ nickel export -f "record-base.ncl" --bikeshed-arg-for-record foo.bar.baz '"set/overridden"'
...
$ nickel export -f "big-config.ncl" --bikeshed-output ubuntu.system
...
```

**Bikeshedding**

There are several things to consider:

- should we pass the arguments as arguments to functions, as setting record values, or both (with different CLI arguments)?
- should we provide a different CLI argument to set a field, and to override a field (distinguishing the "normal" usage of filling a missing value versus intentionally overriding existing behavior)
- what should be the name of the new CLI argument? Proposals
  -  `--arg`, `--input`, `--param`, `--set`, `--set-field` for arguments
  - `--access`, `--select`, `--field`, `--path`, `--field-path` for output

My very personal view is that we should encourage the recursive records (aka partial configurations) approach, and thus support in priority passing data to records rather than to functions (also, this has the advantage of forcing users to name such inputs, while functions can be purely positional). And that we should differentiate between "passing an argument from the CLI" and "overriding an existing value", so we should have an additional `--override` CLI argument (if needed).

Radvendii:
[Rewritten by @yannham]

Partially fixes #1408 (it doesn't extract specific fields).

This PR adds a new "customize mode" to the CLI export subcommand, which generates a new CLI interface from the configuration to be exported. The user can then interact with this configuration by providing value directly on the CLI or by exploring the interface of the configuration via the help command of this CLI.

In practice, the customize mode is set by the freeform argument `--`. Then, the any field without definition (or with a default priority) is converted to an argument of the same name, and fields with higher priorities can be also overridden vie the `--override` argument.

## Example

```console
$ nickel export -f examples/config-gcc/config-gcc.ncl -- --help
Usage: customize-mode [OPTIONS]

Options:
      --path_libc <Path,SharedObjectFile>
          • documentation: Path to libc.
           [default: "/lib/x86_64-linux-gnu/libc.so"]
      --override <field> <value>
          Override any field of the configuration with a valid Nickel expression provided as a
          string. The new value will be merged with the configuration with a `force` priority.
          
          Overridable fields:
          - flags
          - optimization_level
  -h, --help
          Print help

$ nickel export -f examples/config-gcc/config-gcc.ncl -- --path_libc '"/foo/bar/libc.so"' --override optimization_level 1
{
  "flags": [
    "Wextra",
    "ostuff.o"
  ],
  "optimization_level": 1,
  "path_libc": "/foo/bar/libc.so"
}
```

## Content

This PR is quite big because it pulls the term builder interface written by @vkleen from tf-ncl into the main Nickel codebase. It consist of:

- A new `term::make::builder` module, with a builder interface to easily generate Nickel terms programmatically
- New code in `nickel_lang_cli::export` to extract data from a configuration through the so-called record spine form (also used for documentation extraction), and convert this information into a Clap CLI parser
- New code in `nickel_lang_cli::export` that parses the generated command line and generate the corresponding Nickel values in the cache, merges them with the configuration to be exported, and export the result

## Shortcomings/Future work

This PR uses the special argument `--override` to override fields that have higher priority than default. This means that if a field is named `--override`, we can't currently override it. There are possible solutions:

- If a field is named `override`, we could generate a different argument name and advertise it on the help message. But this new generated argument name could itself conflict with an existing field...
  - We could generate an argument that is sure to not conflict with any other, for example by abusing a short name (e.g. using `-override` in this case
  - Conversely, we could change `--override` to be something that doesn't clash with existing arguments, by using `-override` for example.
  - Use a second free-form separator for overrides (but this starts to be clumsy)
- I used the same code as the query printer to display documentation, to ensure that the result of `nickel query` and the documentation printed through `-- --help` has a consistent styling. However, it adds a `* documentation:` in the customize CLI help message which isn't very nice. We could either get rid of it for queries as well, or do something specific for the customize mode.
-  The customize mode could ultimately apply to other subcommands than `export`. This would make sense for evaluating as well, for example.


yannham:
Update: this is almost ready, but unfortunately there are some last minute new-Ident-types-related chores, and I won't have time to fix that before taking time off, so this PR will have to wait a bit.

yannham:
TODO:

- [ ] Move the `Combine` trait somewhere up, as it's now used in several components
- [ ] Self review

yannham:
@Radvendii could you please set the PR as ready for review?

yannham:
@Radvendii I can't request your review on your own PR, but if you want to take a look, that would be valuable

core/src/error.rs
```diff
@@ -2281,3 +2285,62 @@ impl IntoDiagnostics<FileId> for ReplError {
...
+        }
+    }
+}
+
+/// Pretty-print an error on stderr.
+///
+/// # Arguments
+///
+/// - `cache` is the file cache used during the evaluation, which is required by the reporting
+/// infrastructure to print point at specific locations and print snippet when needed.
```
vkleen, line 2321:
```suggestion
/// infrastructure to point at specific locations and print snippets when needed.
```

core/src/program.rs
```diff
@@ -201,10 +212,48 @@ impl<EC: EvalCache> Program<EC> {
...
-        let (t, initial_env) = self.prepare_eval()?;
+    /// Same as `eval`, but proceeds to a full evaluation. Optionally take a set of overrides that
+    /// are to be applied to the term (in practice, to be merged with).
+    ///
+    /// Skips record fields marked `not_exported`.
+    ///
+    /// # Arguments
+    ///
+    /// - `override` is a list of overrides, where the first element is a field path represented as an
+    ///   array of strings, and the second is Nickel source code as string. Each override is imported
```
vkleen, line 223:
```suggestion
    /// - `override` is a list of overrides in the form of an iterator of [`FieldOverride`]s. Each override is imported
```
Something like that. It's not a list of pairs anymore since we introduced `FieldOverride`.

core/src/program.rs
```diff
@@ -201,10 +212,48 @@ impl<EC: EvalCache> Program<EC> {
...
+                    record = record
+                        .path(ovd.path)
+                        .priority(ovd.priority)
+                        .value(Term::ResolvedImport(value_file_id));
+                }
+
+                let (t, initial_env) = self.prepare_eval()?;
+                let built_record = record.build();
+                let wrapper =
+                    mk_term::op2(BinaryOp::Merge(Label::default().into()), t, built_record);
```
vkleen, line 253: I'm not sure if we can do better than the `Label::default()` here for the merge label. But it certainly isn't ideal if the error reporting path ever ends up looking at it. At the moment we might not be able to hit that, but it's certainly a future footgun.

core/src/program.rs
```diff
@@ -275,19 +324,77 @@ impl<EC: EvalCache> Program<EC> {
...
+    ///       subbaz = [some_func "some_arg", "snd" ++ "_elt"],
+    ///     },
+    ///   },
+    ///   input,
+    ///   depdt = input & {extension = 2},
+    /// }
+    /// ```
+    ///
+    /// To evaluate a term to a record spine, we first evaluate it to a WHNF and then:
+    /// - If the result is a record, we recursively evaluates subfields to record spines
```
vkleen, line 363:
```suggestion
    /// - If the result is a record, we recursively evaluate subfields to record spines
```

core/src/term/make/builder.rs
```diff
@@ -0,0 +1,683 @@
...
+//! let t : RichTerm = b
+//!     .field("bar")
+//!     .value(Term::Num(42.into()))
+//!     .into();
+//! ```
+//!
+//! ## Warning
+//!
+//! `RichTerm`s generated by this module are not suitable for evaluation by Nickel. They should
+//! only ever be pretty printed.
```
vkleen, line 22: We're ignoring my warning here in the override code in export 😅 I think this was originally just because of fake `TermPos`s that I used. We seem to only use these as part of `Label::default()` now, and it seems we accept that the error reporting might do silly things if it encounters those elsewhere already.
```suggestion
```

vkleen (approved):

cli/src/export.rs
```diff
@@ -24,17 +47,499 @@ pub struct ExportCommand {
...
+    }
+}
+
+trait ExtractInterface {
+    fn extract_interface(&self) -> Option<TermInterface>;
+}
+
+impl ExtractInterface for &Term {
+    fn extract_interface(&self) -> Option<TermInterface> {
+        if let Term::Record(rd) = self {
```
jneem, line 220: Does `RecRecord` ever happen here, or have those all been turned into `Record` by this point?

cli/src/export.rs
```diff
@@ -24,17 +47,499 @@ pub struct ExportCommand {
...
+    }
+}
+
+/// A command dynamically built from a Nickel term.
+struct TermCommand {
+    /// The corresponding clap command
+    clap_cmd: clap::Command,
+    /// A mapping between clap argument ids and the corresponding field path (for inputs).
+    args: HashMap<clap::Id, Vec<String>>,
+    /// A maping between a field path and the corresponding override data.
```
jneem, line 457:
```suggestion
    /// A mapping between a field path and the corresponding override data.
```

core/src/program.rs
```diff
@@ -201,10 +212,47 @@ impl<EC: EvalCache> Program<EC> {
...
+    ///
+    /// - `override` is a list of overrides in the form of an iterator of [`FieldOverride`]s. Each override is imported
+    ///   in a separate in-memory source, for complete isolation (this way, overrides can't
+    ///   accidentally or intentionally capture other fields of the configuration). A stub record is
+    ///   then built, which has all fields defined by `overrides`, and values are an import referring
+    ///   to the corresponding isolated value. This stub is finally merged with the current program
+    ///   before being evaluated for import.
+    pub fn eval_full_for_export(
+        &mut self,
+        overrides: Option<impl IntoIterator<Item = FieldOverride>>,
```
jneem, line 230: The `Option` in the signature here seems a little redundant, given that the iterator could just be empty.

jneem (approved):

cli/src/error.rs
```diff
@@ -20,6 +25,31 @@ pub enum Error {
...
+        _files: &mut Files<String>,
+        _stdlib_ids: Option<&Vec<FileId>>,
+    ) -> Vec<Diagnostic<FileId>> {
+        match self {
+            CliUsageError::InvalidOverride { path } => {
+                vec![Diagnostic::error()
+                    .with_message(format!("invalid override: unknown field `{path}`"))
+                    .with_notes(vec![format!(
+                        "`{path}` doesn't refer to record field accessible from the root of the \
+                        configuration and thus can't be the target of `--override`."
```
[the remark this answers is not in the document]
yannham: Right, and even more because `clap` does this automatically for arguments, hence fields with a definition should benefit from it already, but not overridable fields, which isn't very consistent.

cli/src/export.rs
yannham: `RecRecord` shouldn't be possible here, because we do some evaluation (the "record spine form") before extracting the interface. The only unevaluated forms are expressions whose evaluation depends on a field that doesn't have a definition yet. But this is impossible for a `RecRecord`: the latter always evaluates to a `Record` in one step.

core/src/program.rs
yannham: We can indeed hit that error, if we try to override something that is already `force`, which can cause various merge issues. In my tests it's not too bad, but this is probably sheer luck (in that the generated default label always point to the first source loaded, which happens to be the original program), and could probably go very wrong.

That being said, I don't think we can really fix that here. The "all labels must refer to a precise span" policy has probably lived for too long, now that we're generating terms programmatically, and we probably can't avoid sprinkling a bunch of optionals (in practice, probably `RawSpan -> TermPos`) here and there.

yannham: Documented in 7882a96d

PR #1476, opened 2023-07-26
Ignore missing field definition errors during documentation extraction

vkleen:
My first implementation of a shallow evaluation mode for extracting documentation tried to evaluate to WHNF every record field that has a value. Then it reports any error that it encounters. But we actually expect to hit `EvalError::MissingFieldDef` in normal partial configurations because presumably an undefined field is intended to be used recursively somewhere. With this change, we report all errors except for `MissingFieldDef`.

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -264,6 +264,7 @@ impl<EC: EvalCache> Program<EC> {
     /// whose values are records.
     #[cfg(feature = "doc")]
     pub fn eval_for_doc(&mut self) -> Result<RichTerm, Error> {
+        use crate::error::EvalError;
         use crate::eval::{Closure, Environment};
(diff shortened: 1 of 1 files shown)


core/src/program.rs
```diff
@@ -278,15 +279,18 @@ impl<EC: EvalCache> Program<EC> {
...
+            let result = vm.eval_closure(
+                Closure {
+                    body: t.clone(),
+                    env: current_env,
+                },
+                initial_env,
+            );
+
+            let (rt, env) = match result {
+                Err(EvalError::MissingFieldDef { .. }) => return Ok(t),
```
yannham, line 291: Might be worth adding a comment to say why we do that in the code (basically, your PR description)

yannham (approved):

PR #1483, opened 2023-07-28
Actually persist output files for formatting

vkleen:
All the code for properly formatting files in-place was there, but in the original PR I forgot to persist the output file. As a result, formatting works for as a pipeline from stdin to stdout but when an output file is specified, the formatted Nickel code is silently discarded.

With this change we actually persist the formatted result.

### DIFF
--- a/cli/src/format.rs
+++ b/cli/src/format.rs
@@ -44,6 +44,7 @@ impl Display for FormatError {
     }
 }
 
+#[derive(Debug)]
 pub enum Output {
     Stdout,
     Disk {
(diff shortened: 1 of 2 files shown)


yannham (approved):

yannham:
By the way, I don't remember, but have we updated the LSP documentation and implementation to now rely on `nickel format` instead of topiary?

vkleen:
We haven't, as far as I'm aware. The LSP probably should link against topiary directly when built with a `format` feature flag and not rely on the CLI, though.

vkleen:
@yannham I added a new CLI integration test suite to check for regressions here. You might want to take a look again.

yannham (approved):
Looks good, but I wonder if a snapshot test wouldn't be more valuable, so that we can see the output as well. But the CLI being under-tested is a different matter, so this can be added in a future PR.

vkleen:
The snapshot tests don't really help with the issue here, because all the commands are producing the correct output on `stdout`. The problem with `format` was really the output file handling. The integration test isn't intended to test any behvaviour other than the correct creation of the file.

PR #1486, opened 2023-07-28
Restore --version for CLI, include git revision

yannham:
Let's see if this passes the CI


flake.nix
```diff
@@ -227,15 +227,21 @@
...
               inherit
                 pname
                 src
                 version
-                cargoArtifacts;
+                cargoArtifacts
+                env;
 
               cargoExtraArgs = "${cargoBuildExtraArgs} ${extraBuildArgs} --package ${pname}";
+
```
yannham, line 244: Is this on purpose?

cli/bin/nickel.rs
```diff
@@ -199,7 +198,7 @@ fn main() {
             "{} {} (rev {})",
             env!("CARGO_BIN_NAME"),
             env!("CARGO_PKG_VERSION"),
-            git_version!()
+            git_version!(fallback = env!("NICKEL_NIX_BUILD_REV"))
```
yannham, line 201: Ah, didn't know we could do that. Nice!

flake.nix
vkleen: The empty line wasn't on purpose. I wonder how that happened...
```suggestion

```

yannham (commented):
I can't accept my own PR, but changes look good to me @vkleen. I just wondered if the version printed when built with Nix and printed when built with cargo would have the same format, but it's not tremendously important.

vkleen:
It's slightly different for dirty worktrees. The cargo one will print the current `HEAD` with suffix `-dirty` (or something to that effect) while the nix build will refuse to give a git revision and print only `dirty`. I don't think we can do better with the Nix flake machinery.

vkleen (approved):

yannham:
Ok, if it's the same for clean trees, then it's all good :+1:

yannham:
Thanks for fixing this PR!

PR #1488, opened 2023-07-28
Linearize imports dependents-first

### LINKED ISSUE #1480: Values from transitive imports are only completed in the LSP if the transitive import is at toplevel
(At least that's my understanding of the issue)

**Describe the bug**

If `foo.x` is a record `{ y = 0 }`, then I'd expect `foo.x.<CURSOR>` to trigger a completion suggesting `y`. However, if foo comes from an `import`-ed file, in which `foo.x` is itself defined by an `import`, then the LSP will only offer the completion if the file defining `foo.x` is imported by a toplevel let-binding.

More concretely, if I have:

- `1.ncl`
    ```
    let foo = import "2.ncl" in foo.x.<CURSOR>
    ```
- `2.ncl`
    ```
    let fooX = import "3.ncl" in { x = fooX }
    ```
- `3.ncl`
    ```
    { y = 0 }
    ```

then the LSP will offer `y` for completion.

However, if (from a fresh LSP session, there also seems to be some caching involved) I change `2.ncl` to `{x = import "3.ncl" }`, then the LSP won't offer any completion any more.

**To Reproduce**

See above ^

**Expected behavior**

Just work :)

**Environment**
 - Version of the code: 1.1.1

**Additional context**

This might have the same cause as #1479  (and annoying in the same context)

jneem:
When file 1 imports file 2, which imports file 3, and we load file 1 into nls, we need to linearize 3 first, then 2, then 1. Prior to this PR, we would linearize 2, 3, 1.

I wasn't able to write a test for this, unfortunately. Our nls testing harness only sees files that you explicitly send to it, and so either you send the files in the order 3, 2, 1 (in which case there's no difference in behavior before/after this PR) or else you send them in some other order and some import (correctly) fails because the imported file doesn't exist anywhere nls can find it.

Fixes #1480

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -680,6 +680,11 @@ impl Cache {
     /// component: this return value is currently used by the LSP to re-run code analysis on new
     /// files/modified files.
     ///
+    /// The resolved imports are ordered by a pre-order depth-first-search. In
+    /// particular, earlier elements in the returned list might import later
+    /// elements but -- unless there are cyclic imports -- later elements do not
+    /// import earlier elements.
+    ///
     /// It only accumulates errors if the cache is in error tolerant mode, otherwise it returns an
     /// `Err(..)` containing  a `CacheError`.
     #[allow(clippy::type_complexity)]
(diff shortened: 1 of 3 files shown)


thufschmitt (approved):
I can't say anything about the implementation (well, except that it looks very sensible), but I can confirm that it fixes the original issue on the nickel-nix side

vkleen (approved):

PR #1489, opened 2023-07-28
Convert paths to absolute before normalization

### LINKED ISSUE #1482: Relative imports to the parent dir behave weirdly
**Describe the bug**

Given a directory tree like

```
.
├── bar.ncl
└── foo
    └── foo.ncl
```

where `foo.ncl` contains `import "../bar.ncl"`, running `nickel export -f foo/foo.ncl` from the toplevel works, but `cd foo && nickel export -f foo.ncl` fails with

```
error: import of ../bar.ncl failed: No such file or directory (os error 2)
  ┌─ foo.ncl:1:1
  │
1 │ import "../bar.ncl"
  │ ------------------- imported here
```

**To Reproduce**

Create a file tree like above and try

**Expected behavior**

Either both should fail (if we want to forbid imports going up in the fs hierarchy) or both should succeed

**Environment**
 - Version of the code: 1.1.1

jneem:
The `normalize_path` function that we store from cargo doesn't handle paths starting with `../` (or, more generally, paths that go above their starting directory). This PR avoids the issue by converting paths to absolute paths before normalization (which is also what cargo does). An alternative would be to change the normalization function so that paths going above their starting directory get normalized to start with however many `../` are necessary.

Fixes #1482

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@
 # Python virtual env
 .venv/
 venv/
+
+# helix configuration
+.helix/
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -289,7 +289,7 @@ impl Cache {
     pub fn add_file(&mut self, path: impl Into<OsString>) -> io::Result<FileId> {
         let path = path.into();
         let timestamp = timestamp(&path)?;
-        let normalized = normalize_path(PathBuf::from(&path).as_path());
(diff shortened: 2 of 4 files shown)


vkleen (approved):
Maybe it would be a good idea to use `normalize_path` in `cli/src/format.rs` as well instead of `fs::canonicalize()`, reading the comment from cargo-util.

PR #1491, opened 2023-08-01
Add a `Nullable` contract to the stdlib

thufschmitt:
Add a contract function that extends the domain of its argument by also
allowing `null`.

This is a very widely used construct, so it deserves being implemented
once and for once here.

From https://github.com/nickel-lang/nickel-nix/pull/101#discussion_r1277651805

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -869,6 +869,28 @@
                 |> from_predicate ((==) constant) ctr_label,
           },
 
+    Nullable
+      | doc m%"
+          `Nullable sub_contract` is a contract that gets verified if the value
+          is either `null` or verified by `sub_contract`.
+
+          # Examples
+
+          ```
+          1 | std.contract.Nullable Number
+            => 1
+          null | std.contract.Nullable Number
+            => null
+          "foo" | std.contract.Nullable Number
+            => error
+          ```
+        "%
+      = fun sub_contract label value =>
+        if value == null then
+          null
+        else
+          std.contract.apply sub_contract label value,
+
     blame
       | doc m%"
           Raises blame for a given label.
--- a/core/tests/integration/fail/contracts/nullable.ncl
+++ b/core/tests/integration/fail/contracts/nullable.ncl
@@ -1,0 +1,7 @@
+# test.type = 'error'
+#
+# [test.metadata]
+# error = 'EvalError::BlameError'
+
+3 | std.contract.Nullable String
+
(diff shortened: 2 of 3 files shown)


core/stdlib/std.ncl
```diff
@@ -869,6 +869,28 @@
...
           },
 
+    Nullable
+      | doc m%"
+          `Nullable sub_contract` is a contract that gets verified if the value
+          is either `null` or verified by `sub_contract`.
+
+          # Examples
+
+          ```
```
vkleen, line 879:
```suggestion
          ```nickel
```

core/stdlib/std.ncl
```diff
@@ -869,6 +869,28 @@
                 |> from_predicate ((==) constant) ctr_label,
           },
 
+    Nullable
+      | doc m%"
+          `Nullable sub_contract` is a contract that gets verified if the value
+          is either `null` or verified by `sub_contract`.
```
vkleen, line 875:
```suggestion
          A value satisfies `Nullable sub_contract` if it is either `null` or
          satisfies `sub_contract`.
```

vkleen (approved):

Radvendii:
If this is going in `std.ncl` it can probably be removed from `core/benches/mantis/lib.ncl` and all the references changed (in `core/benches/mantis/schemas/nomad/types.ncl`)

thufschmitt:
> If this is going in `std.ncl` it can probably be removed from `core/benches/mantis/lib.ncl` and all the references changed (in `core/benches/mantis/schemas/nomad/types.ncl`)

Indeed. I've changed mantis's `lib.ncl` to make `Nullable` an alias for `std.contract.Nullable`. Let me know if you prefer me to directly refer to `std.contract.Nullable` everywhere

yannham (changes requested):
I'll write a bit more about the why very soon (tm), but I just want to avoid having this PR merged yet

yannham:
For more context: I'm convinced that `Nullable` should be a builtin static type, not a mere contract, because it's so ubiquitous in a JSON-based DSL with `null` that can pop up everywhere (and because we thankfully made our static types non nullable by default).

A static Nullable requires us to think about how we deconstruct those things in statically typed way (we need something like flow-sensitive typing, either `if x = null then ... else ... ` or a new construct in match case), but we can already implement a first version without that. This would be at least as featureful as a contract in the stdlib (in fact even more: you could already produce `Nullable T` in statically typed code without issues), and people would use the right syntax/notion right away.

If this is something you need, that should not be very long to implement, @thufschmitt.

thufschmitt:
Fair enough,

I assumed you wouldn't want it as a static type because of the complexity it introduces (the deconstruction you talk about, probably some form of subtyping). But if you think it's worth it, then I'm all for it!

There's no urge for me, I can totally live with my custom contract for as long as needs to be (this isn't in typed code, so I don't need the static version).

PR #1492, opened 2023-08-02
Traversal by reference

jneem:
Here is an attempt at adding a function for (short-cuttable) tree traversals by reference. I only iterated top-down traversal, since (1) it's the one I needed and (2) short-cutting a subtree doesn't really make sense in bottom-up traversal.

I originally made the method generic over the `FnMut`, but I ran into rustc's recursion limit. Maybe this is one reason that the other traverse function separates out the mutable state? I'll try that tomorrow...

### DIFF
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -304,10 +304,7 @@ impl Traverse<RichTerm> for RuntimeContract {
         Ok(RuntimeContract { contract, ..self })
     }
 
-    fn traverse_ref<F, U>(&self, f: &mut F) -> Option<U>
-    where
-        F: FnMut(&RichTerm) -> TraverseControl<U>,
-    {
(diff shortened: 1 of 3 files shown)


vkleen (approved):
@yannham ran into the same recursion limit problem when implmenting recursion schemes for the Nickel AST. If I recall correctly, he was able to solve it by strategically adding indirection in the right place in the implementations of what here is `traverse_ref`. But I wasn't able to do the same thing here quickly, and dynamic dispatch shouldn't be a big problem here.

If you want to keep trying to make it generic, that would be cool, but I don't think it's absolutely necessary 👍

yannham:
For the record, I hit this two times I believe. I don't remember the first fix, but the second was to remove a spurious `&` prefix before a `Fn`, which didn't change anything for typechecking as it was auto-deref by the rust compiler, but did create an infinite type at template instantiation time/codegen.

PR #1497, opened 2023-08-03
Import YAML files containing multiple documents as arrays

vkleen:
For example, in Kubernetes deployments it is common practice to package multiple YAML documents into a single file. For processing such files with Nickel we need to support importing multiple YAML documents from a single file. The concern with this is: how does such a file fit into Nickel's data model? Importing it as an array of Nickel values felt most reasonable to me. At the same time, a YAML file containing a single document should still be transparently serialized as that document, not as a single element array.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -6,6 +6,7 @@ use crate::eval::Closure;
 use crate::parser::{lexer::Lexer, ErrorTolerantParser};
 use crate::position::TermPos;
 use crate::stdlib::{self as nickel_stdlib, StdlibModule};
+use crate::term::array::Array;
 use crate::term::record::{Field, RecordData};
 use crate::term::{RichTerm, SharedTerm, Term};
 use crate::transform::import_resolution;
@@ -14,12 +15,14 @@ use crate::typecheck::{self, type_check, Wildcards};
 use crate::{eval, parser, transform};
 use codespan::{FileId, Files};
 use io::Read;
+use serde::Deserialize;
 use std::collections::hash_map;
 use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
+use std::rc::Rc;
 use std::result::Result;
 use std::time::SystemTime;
 use void::Void;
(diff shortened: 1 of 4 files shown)


core/src/cache.rs
```diff
@@ -483,9 +486,39 @@ impl Cache {
...
-                .map_err(|err| (ParseError::from_serde_yaml(err, file_id))),
+            InputFormat::Yaml => {
+                // YAML files can contain multiple documents. If there is only
+                // one we transparently deserialize it. If there are multiple,
+                // we deserialize the file as an array.
+                let de = serde_yaml::Deserializer::from_str(self.files.source(file_id));
+                let mut terms = de
+                    .map(|de| {
+                        RichTerm::deserialize(de)
+                            .map(attach_pos)
```
jneem, line 497: It would make more sense to attach the sub-span corresponding to the single document, I guess. Doesn't seem like this can be easily extracted from `serde_yaml`, though...

core/src/cache.rs
```diff
@@ -483,9 +486,39 @@ impl Cache {
...
-            InputFormat::Yaml => serde_yaml::from_str(self.files.source(file_id))
-                .map(|t| (attach_pos(t), ParseErrors::default()))
-                .map_err(|err| (ParseError::from_serde_yaml(err, file_id))),
+            InputFormat::Yaml => {
+                // YAML files can contain multiple documents. If there is only
+                // one we transparently deserialize it. If there are multiple,
+                // we deserialize the file as an array.
+                let de = serde_yaml::Deserializer::from_str(self.files.source(file_id));
+                let mut terms = de
+                    .map(|de| {
```
jneem, line 495: TIL that `serde_yaml::Deserializer` implements `Iterator<Item=serde_yaml::Deserializer>` :scream_cat:

jneem (approved):

core/src/cache.rs
vkleen: Me too 🙀

core/src/cache.rs
vkleen: Yeah, I guess that's also the reason that the subterms of a term imported from something other than Nickel don't carry positions. I wonder if that breaks some assumptions in the evaluator...

PR #1502, opened 2023-08-04
Refactor the Nickel CLI

vkleen:
This is a refactor of the Nickel CLI structure with the goal of easing maintenance and later on streamlining the user interface. Notable changes are:

- Create an explicit `eval` subcommand, while keeping the `eval` the default when no subcommand is provided
- Add an option to generate shell completion files: `nickel gen-completions bash`, for instance.
- Restructure the CLI code into one Rust module per subcommand.
- Create a structured CLI error type

Depends on #1486

### DIFF
--- a/cli/src/format.rs
+++ b/cli/src/format.rs
@@ -47,21 +47,14 @@ pub enum Output {
 }
 
 impl Output {
-    pub fn new(path: Option<&Path>) -> CliResult<Self> {
-        match path {
-            None => Ok(Self::Stdout),
-            Some(path) => {
-                let path = nickel_lang_core::cache::normalize_path(path)?;
-                Ok(Self::Disk {
-                    staged: NamedTempFile::new_in(path.parent().ok_or_else(|| {
-                        FormatError::NotAFile {
-                            path: path.to_owned(),
-                        }
-                    })?)?,
-                    output: path.to_owned(),
-                })
-            }
-        }
+    pub fn from_path(path: &Path) -> CliResult<Self> {
+        let path = nickel_lang_core::cache::normalize_path(path)?;
+        Ok(Self::Disk {
+            staged: NamedTempFile::new_in(path.parent().ok_or_else(|| FormatError::NotAFile {
+                path: path.to_owned(),
+            })?)?,
+            output: path.to_owned(),
+        })
     }
 
     pub fn persist(self) {
@@ -101,8 +94,8 @@ pub struct FormatCommand {
 impl FormatCommand {
     pub fn run(self, global: GlobalOptions) -> CliResult<()> {
         let mut output: Output = match (&self.output, &global.file, self.in_place) {
-            (None, None, _) | (None, Some(_), false) => Output::new(None)?,
-            (None, Some(file), true) | (Some(file), _, _) => Output::new(Some(file))?,
+            (None, None, _) | (None, Some(_), false) => Output::Stdout,
+            (None, Some(file), true) | (Some(file), _, _) => Output::from_path(file)?,
         };
         let mut input: Box<dyn Read> = match global.file {
             None => Box::new(stdin()),
--- a/cli/src/query.rs
+++ b/cli/src/query.rs
@@ -30,27 +30,37 @@ pub struct QueryCommand {
 }
 
 impl QueryCommand {
-    pub fn run(self, global: GlobalOptions) -> CliResult<()> {
+    fn attributes_specified(&self) -> bool {
+        self.doc || self.contract || self.typ || self.default || self.value
+    }
+
+    fn query_attributes(&self) -> query_print::Attributes {
+        // Use a default selection of attributes if no option is specified
+        if !self.attributes_specified() {
+            query_print::Attributes::default()
+        } else {
+            query_print::Attributes {
+                doc: self.doc,
+                contract: self.contract,
+                typ: self.typ,
+                default: self.default,
+                value: self.value,
+            }
+        }
+    }
+
+    pub fn run(mut self, global: GlobalOptions) -> CliResult<()> {
         let mut program = self.evaluation.prepare(&global)?;
 
         program
-            .query(self.path)
+            .query(std::mem::take(&mut self.path))
             .map(|term| {
-                // Print a default selection of attributes if no option is specified
-                let attrs =
-                    if !self.doc && !self.contract && !self.typ && !self.default && !self.value {
-                        query_print::Attributes::default()
-                    } else {
-                        query_print::Attributes {
-                            doc: self.doc,
-                            contract: self.contract,
-                            typ: self.typ,
-                            default: self.default,
-                            value: self.value,
-                        }
-                    };
-
-                query_print::write_query_result(&mut std::io::stdout(), &term, attrs).unwrap()
+                query_print::write_query_result(
+                    &mut std::io::stdout(),
+                    &term,
+                    self.query_attributes(),
+                )
+                .unwrap()
             })
             .report_with_program(program)
     }


yannham:
I haven't looked at it yet, but this PR looks like it mixes many different things - code refactoring, changing the interface, adding new features wrt merging - would it be feasible to break it down into as much independent PR as possible, if reasonable (that is, if that doesn't incur too much work)? In particular, I would be happy to merge refactoring ASAP, but redesigning the UX probably mandates more research and discussion.

vkleen:
That makes sense, I'll try to make this PR only a refactoring without changing the interface.

vkleen:
The current version should now keep the previous user-facing structure with two exceptions. First, there is hidden support for an `eval` subcommand which is still the default when no subcommand is specified. Second, there is a `gen-completions` subcommand using `clap_complete` to generate shell completion files.

cli/src/error.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn from(error: nickel_lang_core::repl::InitError) -> Self {
+        Error::Repl { error }
+    }
+}
+
+pub trait WithProgram<T> {
+    fn with_program(self, program: Program<CBNCache>) -> CliResult<T>;
+}
+
+impl<T> WithProgram<T> for Result<T, nickel_lang_core::error::Error> {
```
yannham, line 47: I wonder if the naming is correct, as this is clearly an extension trait (cf https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html). Given the linked RFC, this might probably be called `ResultErrorExt` or maybe `CliResultExt`, even if `CliResult` is defined within this crate...or, `ResultErrorExt`? The latter is probably the closest to the RFC, but it's really obscure.

Also, I feel like the `T` should be a parameter of `with_program` and not of `WithProgram`, but it's purely aesthetic, as I believe the two version are isomorphic.

Another solution is to make `with_program` a free-standing function, which is perhaps less surprising but you lose the ability to chain with `.`. Or to simply inline `with_program`, which doesn't sound too bad either.

cli/src/doc.rs
```diff
@@ -34,66 +40,85 @@ impl fmt::Display for DocFormat {
...
+    /// The output format for the generated documentation.
+    #[arg(long, value_enum, default_value_t)]
+    pub format: crate::doc::DocFormat,
+
+    #[command(flatten)]
+    pub sources: Files,
+}
+
+impl DocOptions {
+    pub fn run(self, global: GlobalOptions) -> CliResult<()> {
```
yannham, line 59: It's maybe not very important, but I find it a bit strange conceptually to implement a `run` method on something called `XXXOptions`, which I would expect to be pure data. I could see either free-standing functions in each related module. On the other hand, I can see how the `XXXOptions` are also subtypes of the `Command` type morally, so maybe it's just a naming thing after all. Maybe we should name them `XXXSubcommand` or something? I don't have a strong opinion either, so if you think this is best as it is, fine by me.

cli/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+use eval::EvalOptions;
+
+use crate::cli::{Command, Options};
+
+fn main() {
+    let opts = <Options as clap::Parser>::parse();
+
+    let result = match opts.command.unwrap_or(Command::Eval(EvalOptions {})) {
+        Command::Eval(eval) => eval.run(opts.global),
+
```
yannham, line 28: It's a very syntactic nitpick, but I feel like new lines don't render very well, when cases are one-liners.

yannham (approved):
After installing this version, I don't get any completion in ZSH. Is there any step that I'm missing to get it to work?

yannham:
Sorry, this maybe wasn't ready for review yet :sweat_smile:

vkleen:
The more reviewing the better :laughing: To get zsh completions, you'd need to source the output of `nickel gen-completions zsh`. There seems to be unstable dynamic completion support for `clap`, but that would still need to be registered with zsh and it would mostly be useful for tab-completing the new freeform arguments to `export` (and maybe `eval` etc.) that we're planning.

vkleen:
About the completions, I actually can't get them to work ephemerally 😞 But `nickel gen-completions zsh > ~/.config/zsh/vendor-completions/_nickel` followed by `compinit` does give me completion in zsh.

yannham:
Ah, it did work for me with `source` as you proposed, using `nickel gen-completions zsh > completions.sh && source completions.sh && rm -f completions.sh` or something like that.

vkleen:
Oooh, thanks 👍 Even nicer for zsh: `source <(nickel gen-completions zsh)`. That let's zsh handle the temporary file.

cli/src/doc.rs
vkleen: A freestanding `run` felt worse to me, although my reasons aren't really well defined. Ultimately I agree this is really an issue with the naming of the `XXXOptions` and I changed them to `XXXCommand`.

cli/src/error.rs
yannham: > Making T just a parameter of the function doesn't work, I think, because I'm implementing WithProgram for Result<T, _> and the type parameter can't actually change for each invocation of with_program. That is to say, this would really be trait for things of kind Type -> Type. But that can't actually be expressed in Rust traits at the moment, as far as I know.

Ah, good point.

> My idea here was to get as close to just a ? at the use site as possible. Just implementing an instance of From isn't good enough here because I need to provide the Program that is to be used for reporting the error. Inlining with_program ends up exposing the structure of the error type at each use site, which feels a bit wrong to me.

I see your point, but I would argue that it's an extension trait currently because your return type is `CliResult<T>` and it seems to me that the only obvious implementer is `CliResult<T>`. I don't really see what other type would or could implement it, excepted maybe something that converts to `CliResult<T>`. Put differently, had we defined `CliResult` as a newtype wrapper, I believe `with_program` would naturally be a method of `CliResult` rather than having its own trait.

As per the RFC:

> Of course, adding methods via a new trait happens all the time. What makes it an extension trait is that the trait is not designed for generic use, but only as way of adding methods to a specific type or family of types.
>
>This is of course a somewhat subjective distinction. Whenever designing an extension trait, one should consider whether the trait could be used in some more generic way. If so, the trait should be named and exported as if it were just a "normal" trait. But traits offering groups of methods that really only make sense in the context of some particular type(s) are true extension traits.

But, as mentioned in this extract, this is a subjective distinction. All in all this is not very important and I'll let you decide what you think is best. I just wanted to bring the attention to existing conventions.

vkleen: It's not `CliResult<T>` that implements the trait but `Result<T, nickel_lang_core::error::Error>`. Note the different `Error` type. In a way, it's really a specialized `Into` that needs to take an extra parameter. But I do see the point about it maybe being an extension trait per the RFC now. And I really don't see how this particular trait could be generically useful 😅

So, I agree with renaming it to something like `ResultErrorExt`. Naming things is truly one of the hard problems :laughing:

cli/src/eval.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub fn run(self, global: GlobalOptions) -> CliResult<()> {
+        let mut program = prepare(&global)?;
+        program
+            .eval_full()
+            .map(|t| println!("{t}"))
+            .report_with_program(program)
+    }
+
+    pub fn prepare(&self, global: &GlobalOptions) -> CliResult<Program<CBNCache>> {
+        prepare(global)
```
yannham, line 21: I imagine `prepare` below isn't inlined here because it's used somewhere else outside of an `EvalCommand` value?

cli/src/format.rs
```diff
@@ -54,7 +47,7 @@ pub enum Output {
 }
 
 impl Output {
-    pub fn new(path: Option<&Path>) -> Result<Self, FormatError> {
+    pub fn new(path: Option<&Path>) -> CliResult<Self> {
```
yannham, line 50: Since `new` seems to be called in two different code paths below with definite value `None` and `Some(path)`, would that be clearer to have a `stdout()` and a `file(path)` instead (or `new_stdout()` and `new_from_path(path)` or whatnot) ?

cli/src/query.rs
```diff
@@ -0,0 +1,57 @@
...
+impl QueryCommand {
+    pub fn run(self, global: GlobalOptions) -> CliResult<()> {
+        let mut program = self.evaluation.prepare(&global)?;
+
+        program
+            .query(self.path)
+            .map(|term| {
+                // Print a default selection of attributes if no option is specified
+                let attrs =
+                    if !self.doc && !self.contract && !self.typ && !self.default && !self.value {
```
yannham, line 41: I probably implemented that, but reading it again, might be clearer to move this condition into an associated method `no_attrs_specified(&self) -> bool` or whatever is a good name

cli/src/typecheck.rs
```diff
@@ -0,0 +1,15 @@
...
+    error::{CliResult, ResultErrorExt},
+    eval,
+};
+
+#[derive(clap::Parser, Debug)]
+pub struct TypecheckCommand {}
+
+impl TypecheckCommand {
+    pub fn run(self, global: GlobalOptions) -> CliResult<()> {
+        let mut program = eval::prepare(&global)?;
```
yannham, line 12: Ah, here is the usage. You can resolve both conversations :)

yannham (approved):

PR #1504, opened 2023-08-04
Term-based completion

jneem:
This is the start (and just the start!) of doing completion using the AST instead of the linearization. The old completion is still active; we're just adding an extra source of completions and unioning them.

Besides the fact that most of the proposed analysis is not implemented, the AST-based completion currently assumes that the input parses fully.

The main new user-visible change is that you can now get completion for static accesses when the first expression in the path is a bit complicated. Like, `(import "foo.ncl").completeMe` and `(a & b).completeMe` now work.


lsp/nls/src/field_walker.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Creates a completion item from this definition.
+    pub fn to_completion_item(&self) -> CompletionItem {
+        /// Attach quotes to a non-ASCII string
+        fn adjust_name(name: &str) -> String {
+            if name.is_ascii() {
+                String::from(name)
+            } else {
+                format!("\"{name}\"")
+            }
```
vkleen, line 59: Just `is_ascii` might not be good enough. For example, a field name might contain spaces and should then be quoted in Nickel. Or is this something that the LSP spec takes care of already?

We have `QUOTING_REGEX` in `core/src/pretty.rs` and `KEYWORDS` in `core/src/parser/lexer.rs` to check if an identifier can be used unquoted. Not sure if you can use those without trouble in the LSP, but if not we should probably add some function to `Ident` to help.

core/src/term/mod.rs
```diff
@@ -1453,6 +1453,19 @@ impl RichTerm {
...
+        self.traverse_ref(&mut |rt: &RichTerm| {
+            if rt.pos.contains(pos) {
+                ret = Some(rt.clone());
+                TraverseControl::<()>::Continue
+            } else {
+                TraverseControl::SkipBranch
+            }
+        });
+        ret
+    }
```
vkleen, line 1468: I'm wondering if it would be reasonable to construct an interval tree like index structure, so we don't have to traverse the entire term each time. But maybe I'm reinventing the linearization concept now 😅

In any case, we shouldn't do that in this PR; just a thought.

vkleen (approved):

yannham (commented):
I think this PR is a bit too short on context, motivations and planning forward. What do we expect to gain from doing the analysis on terms? Is that for code simplicity? Will this provide better completion?

What is the vision of the end result (possibly after many PRs)? The fact is, the linearization part is already obscure and not that easy to update and maintain. Now we're adding yet another component, which based itself on the AST. It might be fine as a transitory state, but in this case it could be useful to have an idea of the desired final state, and how to get there; otherwise, I fear we're at risk of being stuck in the middle, with competing representations for code analysis.

lsp/nls/src/linearization/completed.rs
```diff
@@ -110,16 +160,15 @@ impl Completed {
                 .take_while(|item| {
                     // Here because None is smaller than everything, if binary search succeeds,
                     // we can safely unwrap the position.
-                    let pos = item.pos.unwrap();
-                    (pos.src_id, pos.start) == *locator
+                    item.pos.unwrap().start_pos() == pos
```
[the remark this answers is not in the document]
jneem: Yes, because `pos` is now a `RawSpan` so it has the file id

core/src/term/mod.rs
jneem: I think in this case the linearization can also be linear. The parsed expression nests like `((({} & {}) & {}) & {})` and so the ranges of the various Op2 terms will be like [0, 5], [0, 8], [0, 11], and so on, and only after all of the [0, n] ranges do we start getting ranges like [m, m+2] for the record terms. The linearizer does a binary search for the start of the range and then it does a linear scan among all ranges with that same start. So if you query a range near the start of the file, I think it will be linear.

Anyway, I think it shouldn't be hard to add an interval map for constant-time lookups. I even have a [crate](https://crates.io/crates/range-map) for it.

core/src/term/mod.rs
```diff
@@ -2089,6 +2108,40 @@ mod tests {
         assert_ne!(res.typ, None);
     }
 
+    fn parse(s: &str) -> (FileId, RichTerm) {
```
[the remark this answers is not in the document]
jneem: The issue I had here was that I wanted to interact with the AST of the parsed program and I didn't see how to do that in the integration test framework. Is there something there that I missed?

core/src/term/mod.rs
yannham: Ok, in the very specific case of querying a range between 0 and the next well-defined item, that's probably true. Still, most of the time, you query a well-defined sub-node, and in that case the current implementation should perform well (for example any of the record operands), while this one always start from the root and navigate through the tree - it doesn't have the possibility of jumping in the middle of the tree, I believe. So querying the `n-th` operator is `O(size of the AST)`. It's also probably possible to do a second binary search as well, to find the closest match, if that's required.

Anyway, as you say, that's something that can be improved later on :+1:

jneem:
Regarding the context and plans, the linearization provides us with a few things:
- lookups by position
- tracking of references/definitions
- type checking and inference

But (in addition to being quite complex for what it does) the linearization drops various other important pieces of information that are present in the AST. The eventual goal is to drop the linearization altogether and do everything with the AST plus some simple auxiliary structures:
- for lookups by position, we use an AST traversal to build some sort of interval tree (or maybe using [this](https://crates.io/crates/range-map) crate that I wrote years ago)
- for tracking of references/definitions, we use an AST traversal to build some hashmaps mapping references to definitions and vice versa
- for type checking and inference, we make a modified AST that stores `Type`s, possibly by making `Term` generic on `RichTerm` and plugging in a modified `RichTerm` that also stores a `Type`.

The problem is that everything in nls is based on the linearization now and it will take a little while to move it all over. The state of things in this PR is
- for lookups by position, I do it by traversal. As discussed, this won't be great for performance but I'll fix it soon
- for tracking of references/definitions, I added a hacky link between the AST and the linearization, so that we can reuse the definition-tracking that already exists in the linearization
- type inference is not yet used by the term-based completion.

Next plans are:
- improve the ast-based lookup
- add a pass for ast-based references/definitions
- do the type-annotated ast
- handle input that doesn't completely parse
- bring the term-based completer up to feature parity. I think the main thing missing by this point will be making use of contract annotations

I think the first two steps are fairly small (i.e. maybe later this week). I'm less sure about the others. After all this, we'll be able to remove the linearization from completion, and I'll move on to "goto definition"

jneem:
(@yannham I'll hold off on merging because I'd like to hear your thoughts on the plan)

yannham (approved):
Thanks for writing this up! It'll probably evolve alongside implementation, but it's clearer to bystanders (at least to me) where we wanna go now, and why. In a first step, I imagine you can avoid modifying the AST by just storing everything in an auxiliary structure `HashMap<SharedTermPtr, AdditionalData>` (which is isomorphic and have the same features, but make for a smaller diff before refactoring `Term`). Maybe refactoring `Term` isn't so bad, because a lot of code use `RichTerm` though... You'll probably know better :slightly_smiling_face:

PR #1510, opened 2023-08-07
[Fix] Fix contract.Sequence reversing order

yannham:
This PR started by noticing in an unrelated code review that some tests involving contracts were incorrect, because they never evaluated any value (thus, by lazyness, contract checks wouldn't fire).

One of those tests failed after being fixed, indicating that `std.contract.Sequence` wasn't preserving the order of applied contracts as required. This PR fixes the aforementioned tests and the ordering issue by using `fold_left` instead of `fold_right` in the implementation of `Sequence`.

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -2975,4 +2975,3 @@
     "%
     = fun msg => null | FailWith msg,
 }
-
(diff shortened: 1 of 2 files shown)


jneem (approved):

Radvendii (approved):
Good catch. `fold_right` seemed simpler, and it passed the test so we didn't think about it too hard 😅  But the contracts are applied left-to-right. This makes more sense when I think of them after they're converted into functions like `$dyn` and the like.

PR #1512, opened 2023-08-08
Improve type variables name allocation when reporting type errors

yannham:
The implementation of the conversion from unification type to human-readable types was a bit confused, using two name tables without really making it clear which table corresponds to which. The implementation didn't correctly update correctly the set of variables already in use, and failed to propagate position information for user-written type variables.

This PR simplifies the implementation, moving free-standing functions into `NameReg`, diminishing the number of arguments and gets rid of the additional table, with the effect of correctly propagating position information for user-written type variables.

This improves the error reporting of the variable level mismatch error. Previously, the offending type variable wasn't pointed at, but after this PR, it is:

```
let eval : forall a. (forall b. b -> b) -> a -> a =
fun f x => f x
in (let g = fun x => x
in eval g) : _
error: function types mismatch
  │
1 │ b -> b
  │ - this part of the expected type
  │
  │
1 │ _a -> _a
  │ -- does not match this part of the inferred type
  │
  ┌─ repl-input-0:4:9
  │
4 │ in eval g) : _
  │         ^ this expression
  │
  = Expected an expression of type `b -> b`
  = Found an expression of type `_a -> _a`
  = Could not match the two function types

error: while matching function types: invalid polymorphic generalization
  ┌─ repl-input-0:1:30
  │
1 │ let eval : forall a. (forall b. b -> b) -> a -> a =
  │                              - this polymorphic type variable
  │
  = While the type of this expression is still undetermined, it appears indirectly in the type of another expression introduced before the `forall` block.
  = The type of this expression escapes the scope of the corresponding `forall` and can't be generalized to the polymorphic type variable `b`
```

### DIFF
--- a/core/src/typecheck/error.rs
+++ b/core/src/typecheck/error.rs
@@ -109,16 +109,6 @@ pub enum UnifError {
 
 impl UnifError {
     /// Convert a unification error to a typechecking error.
-    ///
-    /// Wrapper that calls [`Self::into_typecheck_err_`] with a [name registry][reporting::NameReg]
-    /// initialized from the current state.
-    pub fn into_typecheck_err(self, state: &State, pos_opt: TermPos) -> TypecheckError {
-        let mut names = reporting::NameReg::new(state.names.clone());
-        self.into_typecheck_err_(state, &mut names, pos_opt)
-    }
-
-    /// Convert a unification error to a typechecking error.
-    ///
     /// There is a hierarchy between error types, from the most local/specific to the most high-level:
     /// - [`RowUnifError`]
     /// - [`UnifError`]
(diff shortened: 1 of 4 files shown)


core/src/typecheck/error.rs
```diff
@@ -112,8 +112,9 @@ impl UnifError {
     ///
     /// Wrapper that calls [`Self::into_typecheck_err_`] with an empty [name
     /// registry][reporting::NameReg].
-    pub fn into_typecheck_err(self, state: &State, pos_opt: TermPos) -> TypecheckError {
-        self.into_typecheck_err_(state, &mut reporting::NameReg::new(), pos_opt)
+    pub fn into_typecheck_err(self, state: &mut State, pos_opt: TermPos) -> TypecheckError {
+        let mut names = reporting::NameReg::new(std::mem::take(state.names));
+        self.into_typecheck_err_(state, &mut names, pos_opt)
```
vkleen, line 117: This function now replaces the `NameReg` in `state` by an empty one and calls `into_typecheck_err_` with the old, filled one. This might end up being very surprising. I wouldn't expect that `into_typecheck_err` removes name allocations from the state. Maybe it would make more sense to take ownership of `state`?

In any case, the doc-comment needs to be updated.

vkleen (approved):

core/src/typecheck/error.rs
yannham: Good point, but doing that at the caller site might be very annoying, because `State` is taken by reference everywhere, so we would end up with a lot of `std::mem::take`.

In practice, when you call to `into_typecheck_err`, you go the error path, so you don't need the state anymore. But you're right that the behavior isn't necessarily natural. Maybe cloning is the path of least resistance. Another solution is to separate the name tables between the one coming from state, that is would be immutably borrowed, and the one corresponding to generated names, that is owned, but that also sounds like more boilerplate (lifetimes, wrapper around `get`, etc.) just to avoid one clone.

yannham: I'll just clone, it's the simplest.

PR #1517, opened 2023-08-09
Add warning for empty query path on the CLI

### LINKED ISSUE #1487: Issue an error when providing no field path to `--query`
**Describe the bug**
The query subcommand takes a program to evaluate, and a field path. It then extracts the metadata - if any - of the corresponding field.

Since RFC005 (#819), querying a configuration without providing a field path doesn't make sense: metadata can only be located at the level of a record field. However, probably for historical reasons, the current CLI accepts it and just always spit "Metadata not found for this value.". This can be confusing, in particular because in the REPL, one provide just one argument which is field path. Thus, `:query std.record.map` in the REPL correctly return the documentation of `map`, but `nickel query <<< std.record.map` doesn't. Indeed, the correct way is `nickel query record.map <<< std` or `nickel query map <<< std.record`.

**To Reproduce**
```console
$ nickel query <<< std.record.map
```

**Expected behavior**
The `query` command should make the field path a required argument and error out on such an invocation.

yannham:
Closes #1487.

I initially only emitted the warning when the expression wasn't a record, but in the end, even in this case, using a proper query path can show additional information. For example:

```console
$ nickel query <<< std.contract.label

Available fields
• append_note
• with_message
• with_notes

$ nickel query label <<< std.contract
• documentation

The label submodule provides functions that manipulate the label
associated with a contract application[..]

Available fields
• append_note
• with_message
• with_notes
```

After this PR:

```console
$ nickel query <<< std.contract.label
warning: empty query path
 = You queried a value without requesting a specific field path. This operation can't find any metadata, beside listing the fields of a record.
 = Try to query the root configuration and provide a query path instead.
 = For example, instead of querying the expression `(import "config.ncl").module.input` with an empty path, query `config.ncl` with the `module.input` path: 
   `nickel query module.input -f config.ncl`

Available fields
• append_note
• with_message
• with_notes
```

In the future we might be smarter and make the suggestion more specific by introspecting the term and see if it's a record access, but for now I think this will do.

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -24,9 +24,9 @@ pub enum Error {
 
 /// Warning emitted by the CLI.
 pub enum Warning {
-    /// The user queried a program without providing any path, and the result isn't a record. In
-    /// this case, querying won't show anything useful, and it's most probably not what the user
-    /// want.
+        /// The user queried a program without providing any path. In this case,
+        /// querying won't show most information, and it's most probably not
+        /// what the user wanted.
     EmptyQueryPath,
 }
 
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -1,4 +1,8 @@
-use nickel_lang_core::{eval::cache::lazy::CBNCache, program::Program};
+use nickel_lang_core::{
+    error::{Diagnostic, Files, IntoDiagnostics},
+    eval::cache::lazy::CBNCache,
+    program::Program,
+};
 
 pub enum Error {
     Program {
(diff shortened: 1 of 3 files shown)


cli/src/error.rs
```diff
@@ -18,6 +22,36 @@ pub enum Error {
     },
 }
 
+/// Warning emitted by the CLI.
+pub enum Warning {
+    /// The user queried a program without providing any path, and the result isn't a record. In
+    /// this case, querying won't show anything useful, and it's most probably not what the user
+    /// want.
```
vkleen, line 29:
```suggestion
        /// The user queried a program without providing any path. In this case,
        /// querying won't show most information, and it's most probably not
        /// what the user wanted.
```

vkleen (approved):

PR #1518, opened 2023-08-09
Use the pretty printer in the  `Display` implementation for `Type`

vkleen:
We previously used a bespoke formatting algorithm for `Type`. I replaced the analogous code for `Term`s by the pretty printer in #1262 but we were worried about some questionable code for contract error reporting before doing the same for types. Namely, at some point it relied on hard coded string offsets for pointing at parts of types that were inferred by Nickel and consequently had no `TermPos`. In #1229 we ripped out that code and replaced it by reparsing the pretty printer output when necessary.

Incidentally, this change also fixes some terms being truncated when formatted. For example, previously

```
SomeUserDefinedContract "that" "takes" "many" "arguments"
```

would be printed as `SomeUserDefinedContract "that" …`. This is somewhat useful to prevent huge screenfuls of error messages sometimes, but it makes the `Display` implementation useless for other natural purposes.

### DIFF
--- a/cli/tests/snapshot/inputs/pretty/simple_record.ncl
+++ b/cli/tests/snapshot/inputs/pretty/simple_record.ncl
@@ -3,6 +3,6 @@
 { 
   a | Number | default = 1, 
   b : String | force = "some long string that goes past the 80 character line limit for pretty printing", 
-  c : { x : Number, y: Number } = { x = 999.8979, y = 500 },
+  c | { x : Number, y: Number } = { x = 999.8979, y = 500 },
   d | Array std.string.NonEmpty = ["a", "list", "of", "non", "empty", "strings"],
 }
--- a/core/src/pretty.rs
+++ b/core/src/pretty.rs
@@ -755,10 +755,7 @@ where
                 .append(allocator.hardline()),
             // TODO
             Sealed(_i, _rt, _lbl) => allocator.text("#<sealed>").append(allocator.hardline()),
-            Annotated(annot, rt) => allocator
-                .atom(rt)
-                .append(allocator.space())
-                .append(allocator.annot(annot)),
+            Annotated(annot, rt) => allocator.atom(rt).append(allocator.annot(annot)),
             Import(f) => allocator
                 .text("import")
                 .append(allocator.space())
@@ -790,9 +787,7 @@ where
                     .text("'")
                     .append(allocator.text(ident_quoted(row)));
                 let builder = if let EnumRowsF::Extend { .. } = tail.0 {
-                    builder
-                        .append(allocator.text(","))
-                        .append(allocator.space())
+                    builder.append(allocator.text(",")).append(allocator.line())
                 } else {
                     builder
                 };
@@ -833,9 +828,7 @@ where
                     .append(typ.pretty(allocator));
 
                 let builder = if let RecordRowsF::Extend { .. } = tail.0 {
-                    builder
-                        .append(allocator.text(","))
-                        .append(allocator.space())
+                    builder.append(allocator.text(",")).append(allocator.line())
                 } else {
                     builder
                 };
@@ -888,19 +881,31 @@ where
                     .group()
                     .append(allocator.intersperse(
                         foralls.iter().map(|i| allocator.as_string(i)),
-                        allocator.space(), //allocator.softline(),
+                        allocator.space(),
                     ))
                     .append(allocator.text("."))
-                    .append(allocator.line())
+                    .append(allocator.softline())
                     .append(curr.to_owned().pretty(allocator))
             }
-            Enum(erows) => erows.pretty(allocator).enclose("[|", "|]"),
-            Record(rrows) => rrows.pretty(allocator).braces(),
+            Enum(erows) => allocator
+                .line()
+                .append(erows.pretty(allocator))
+                .nest(2)
+                .append(allocator.line())
+                .group()
+                .enclose("[|", "|]"),
+            Record(rrows) => allocator
+                .line()
+                .append(rrows.pretty(allocator))
+                .nest(2)
+                .append(allocator.line())
+                .group()
+                .braces(),
             Dict {
                 type_fields: ty,
                 flavour: attrs,
             } => allocator
-                .line()
+                .softline()
                 .append(allocator.text("_"))
                 .append(allocator.space())
                 .append(match attrs {
@@ -909,7 +914,7 @@ where
                 })
                 .append(allocator.space())
                 .append(ty.pretty(allocator))
-                .append(allocator.line())
+                .append(allocator.softline())
                 .braces(),
             Arrow(dom, codom) => match dom.typ {
                 Arrow(..) | Forall { .. } => dom
--- a/core/src/typ.rs
+++ b/core/src/typ.rs
@@ -1045,7 +1045,7 @@ impl Type {
             | TypeF::Var(_)
             | TypeF::Record(_)
             | TypeF::Enum(_) => true,
-            TypeF::Flat(rt) if matches!(*rt.term, Term::Var(_)) => true,
+            TypeF::Flat(rt) if rt.as_ref().is_atom() => true,
             _ => false,
         }
     }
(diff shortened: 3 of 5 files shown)


cli/tests/snapshot/inputs/pretty/simple_record.ncl
```diff
@@ -3,6 +3,6 @@
 { 
   a | Number | default = 1, 
   b : String | force = "some long string that goes past the 80 character line limit for pretty printing", 
-  c : { x : Number, y: Number } = { x = 999.8979, y = 500 },
+  c | { x : Number, y: Number } = { x = 999.8979, y = 500 },
```
Radvendii, line 6: Do we want to be replacing existing tests, or adding new ones?

core/src/pretty.rs
```diff
@@ -833,9 +828,7 @@ where
                     .append(typ.pretty(allocator));
 
                 let builder = if let RecordRowsF::Extend { .. } = tail.0 {
-                    builder
-                        .append(allocator.text(","))
-                        .append(allocator.space())
+                    builder.append(allocator.text(",")).append(allocator.line())
```
Radvendii, line 831: What is the difference between using `space()` and `line()`? I thought the pretty printer would automatically insert line breaks when it needs to

core/src/typ.rs
```diff
@@ -1045,7 +1045,7 @@ impl Type {
             | TypeF::Var(_)
             | TypeF::Record(_)
             | TypeF::Enum(_) => true,
-            TypeF::Flat(rt) if matches!(*rt.term, Term::Var(_)) => true,
+            TypeF::Flat(rt) if rt.as_ref().is_atom() => true,
```
Radvendii, line 1048: This whole thing is a big `matches!()`
```
matches!(
    &self.typ,
    TypeF::Dyn
    | TypeF::Number
    | TypeF::Bool
    | TypeF::String
    | TypeF::Var(_)
    | TypeF::Record(_)
    | TypeF::Enum(_)
    | TypeF::Flat(rt) if rt.as_ref.is_atom()
)
```

core/src/typ.rs
```diff
@@ -1149,89 +1149,14 @@ impl Traverse<RichTerm> for Type {
...
-                flavour: DictTypeFlavour::Contract,
-            } => write!(f, "{{ _ | {type_fields} }}"),
-            TypeF::Arrow(dom, codom) => match dom.typ {
-                TypeF::Arrow(_, _) | TypeF::Forall { .. } => write!(f, "({dom}) -> {codom}"),
-                _ => write!(f, "{dom} -> {codom}"),
-            },
-            TypeF::Wildcard(_) => write!(f, "_"),
-        }
+        let doc: DocBuilder<_, ()> = self.clone().pretty(&allocator);
+        doc.render_fmt(80, f)
```
Radvendii, line 1159: We have this definition for `Display` elsewhere for other types IIRC. Is there a way to deduplicate or remove that boilerplate?

Radvendii (approved):

cli/tests/snapshot/inputs/pretty/simple_record.ncl
vkleen: I made that change thinking originally that a type signature here wouldn't work and `pprint-ast` just didn't evaluate and so doesn't notice. That's nonsense, of course. I'll revert this.

Regarding adding new tests for this change, there are a few round-trip tests in `core/src/typ.rs` that make sure the `Display` implementation for `Type` doesn't diverge far from what we expected.

core/src/pretty.rs
vkleen: When a particular `group()` doesn't fit on a single line, the pretty printer will first turn *every* `line()` into a line break and after that try to see if any more line breaks are needed, for instance at `space()`. So, this change is the difference between getting
```
{
  a, b, c,
  d, e,
}
```
and
```
{
  a,
  b,
  c,
  d,
  e,
}
```

It seems that's our formatting preference (also in topiary): if there are already multiple lines required for a record, then each record field should get its own.

core/src/typ.rs
vkleen: Right, at least we could make a small helper in `core/src/pretty.rs`, I think.

core/src/typ.rs
vkleen: It's actually not, sneakily. Your suggested pattern isn't valid, because `rt` isn't bound in every branch :disappointed:

core/src/pretty.rs
```diff
@@ -755,10 +755,7 @@ where
                 .append(allocator.hardline()),
             // TODO
             Sealed(_i, _rt, _lbl) => allocator.text("#<sealed>").append(allocator.hardline()),
-            Annotated(annot, rt) => allocator
-                .atom(rt)
-                .append(allocator.space())
-                .append(allocator.annot(annot)),
+            Annotated(annot, rt) => allocator.atom(rt).append(allocator.annot(annot)),
```
yannham, line 758: Wasn't the space necessary here? Or did it move to `allocator.annot` ?

core/src/pretty.rs
```diff
@@ -888,19 +881,31 @@ where
                     .group()
                     .append(allocator.intersperse(
                         foralls.iter().map(|i| allocator.as_string(i)),
-                        allocator.space(), //allocator.softline(),
+                        allocator.space(),
                     ))
                     .append(allocator.text("."))
-                    .append(allocator.line())
+                    .append(allocator.softline())
```
yannham, line 887: I wonder why we used `line()` instead of a space here. At least `softline` sounds better.

core/src/pretty.rs
```diff
@@ -888,19 +881,31 @@ where
...
                     .append(curr.to_owned().pretty(allocator))
             }
-            Enum(erows) => erows.pretty(allocator).enclose("[|", "|]"),
-            Record(rrows) => rrows.pretty(allocator).braces(),
+            Enum(erows) => allocator
+                .line()
```
yannham, line 891: should we always add a newline here? Does this mean that something like `[| 'foo |]` will always be rendered with a new line just for `foo` ?

yannham (approved):
The diff and the removal of the ad hoc printing code feels good :upside_down_face:

core/src/pretty.rs
vkleen: `annot_part` adds a `line()` before a type or contract annotation starts. That produces a space if everything in the current group fits on a line and a linebreak otherwise. So the space here was actually producing a double space before a contract assertion. On master:
```
❯ nickel pprint-ast <<<'foo | Number'
foo  | Number
```

core/src/pretty.rs
vkleen: `line()` only turns into a newline if the current `group()` doesn't fit on a line. So if `[| 'foo |]` fits on a line, it stays as it is. If it doesn't, the formatting becomes
```
[|
  'foo
|]
```

With `space()` instead of `line()` here it would turn into
```
[| 'foo
|]
```
if the first line just fits into the line length limit.

yannham: I see, thanks for the explanation. I thought this was softline, but the documentation of `pretty` indeed makes it different: a softline is a `space()` if the _document_ doesn't fit on a page and is a `line()` otherwise, and the latter is a space of the group doesn't fit on a line. Given this definition, is there a reason anymore to turn the `.line()` after the `forall` into a `softline()`? It seems that we don't really care about the notion of `page` in the pretty printer, or do we?

vkleen: I don't think "page" in the `pretty` documentation refers to actual pages. Rather, it seems to really just be about maximum line length. I originally made that change because `Number -> (forall a. a -> String) -> String` was formatted like
```
Number -> (forall a.
a -> String) -> String
```
which looks horrible. `softline()` seemed to work and I didn't think about it too much. But looking back I think the entire pretty printing of arrow types is inconsistend about line breaks. I'll try to play with it a little and see if I can improve it.

yannham: Ok, this was just a question, I think it's fair to merge like this and then see if we do better in a follow up PR.

PR #1520, opened 2023-08-09
Add a linebreak between contract and the documentation in the doc output

thufschmitt:
The markdown output of `nickel doc` for record fields with both a contract and a documentation currently looks like

```markdown
- `foo | Dyn`
documentation
```

which in strict markdown is equivalent to

```markdown
- `foo | Dyn` documentation
```

and isn't really nice once converted to Html.

Change that to

```markdown
- `foo | Dyn`
\
documentation
```

which renders `documentation` on a newline (still whithin the same bullet point though).

Screenshots on a basic `pandoc` rendering:

- before 
  image
- after
    image

### DIFF
--- a/cli/tests/snapshot/snapshots/snapshot__doc_stdout_record.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__doc_stdout_record.ncl.snap
@@ -4,5 +4,6 @@ expression: out
 ---
 # `Hello`
 
+\
 World\!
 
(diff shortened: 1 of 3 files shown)


vkleen (approved):

yannham (approved):
I was going to say something about affecting the user manual, but the latter is built from the JSON export I believe, right, @vkleen ? @thufschmitt might or might not be useful for you, but AFAIR it was just easier to generate HTML from a json export of the documentation, which you can do with `nickeld doc --format json`.

vkleen:
That's correct, the user manual is formatted on the website from the JSON output.

PR #1531, opened 2023-08-14
Sort out idents and positions

jneem:
The motivation for this change is `Ident`'s somewhat surprising treatment of positions: it stores them, but ignores them for equality checking and hashing. This causes problems for the LSP because we want to map from `Ident`s to usages of those identifiers, and in this case we want two `Ident`s with different positions to be considered as different.

This PR goes from a single `Ident` type to three types (names subject to bikeshedding):
- `Symbol` is just the name.
- `Ident` is a name and a position.
- `HashIdent` is a wrapper around `Ident` that implements `Hash` while ignoring the position

The eventual goal would be to get rid of `HashIdent`, the idea being that if you actually want to keep the position then you don't want to ignore it while hashing because you'll end up throwing away positions that you care about. I didn't manage to get that far, though -- I didn't even manage to drop the position-ignoring `PartialEq` for `Ident`, because it caused breakage in the type checker.

Anyway, I'm mainly just posting this to get some perspective on how much time is reasonable to spend on this. A quick workaround for the LSP would be just to store a `(Ident, TermPos)`...


core/src/identifier.rs
```diff
@@ -1,35 +1,107 @@
...
+    fn into(self) -> String {
+        self.into_label()
+    }
+}
+
+/// An identifier is a [`Symbol`] together with a position.
+///
+/// The position is ignored for equality comparison and hashing; it's mainly
+/// intended for error messages. If you want to combine a symbol with a semantically
+/// meaningful position, use [`RichIdent`] instead.
```
yannham, line 91: Is `RichIdent` planned for a future PR? By the way, this naming scheme is unfortunate (one day, the sooner the better, I will rip `RichTerm` out as well). I think we could use a better convention. Maybe something like `LocIdent` or `IdentLoc`? (so that there's no possible confusion with `XXXPos`, such as `TermPos`, which is "the position of a term", while here it's more like a "located ident"/ident+location).

yannham (approved):
Looks good in general, but I wonder about the naming. `Symbol` was an internal type, so the naming was less important, but now it's practically "the new Ident". I believe we import it and use it more than `Ident`, and `Symbol` is a bit vague. I proposed more descriptive names (but just on the top of my head, there are most definitely better solutions out there!)

Another solution would be to call `Symbol` `Ident`. And use a longer name for the current `Ident`. Though we still need a way to distinguish `Ident` and planned `RichIdent`.

core/src/eval/merge.rs
```diff
@@ -237,8 +238,10 @@ pub fn merge<C: Cache>(
 
             match mode {
                 MergeMode::Contract(label) if !r2.attrs.open && !left.is_empty() => {
-                    let fields: Vec<String> =
-                        left.keys().map(|field| format!("`{field}`")).collect();
+                    let fields: Vec<String> = left
+                        .keys()
+                        .map(|field| format!("`{}`", field.symbol()))
```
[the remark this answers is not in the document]
jneem: Sorry, that was left over from my attempts at a hashable wrapper for `Ident`.

core/src/identifier.rs
jneem: Ah, I had planned it when I wrote the comment, but then I decided that it will live in `nls` instead of `nickel-core` (in a future PR). I'll remove it from the comment.

PR #1537, opened 2023-08-18
Refactor destructuring bindings

vkleen:
Simplify the parsing of record destructuring and in particular stop applying a `Dyn` contract if no contract is specified in the binding. This could conceivably change error reporting (which seems to be the original reason this was done) but I haven't been able to find a case where the label attached to the `Dyn` contract was actually used.

While there, I refactored the code merging contract annotations and default values in record patterns and added the (expectedly) failing destructuring tests from the `nickel-lang-core` integration tests into snapshot tests in the CLI. This way we can hopefully have an easier time noticing regressions in the error messages, particularly if they start pointing at the wrong location.


jneem (approved):
I don't know the implications of that dummy contract, but the refactor part definitely looks good.

PR #1540, opened 2023-08-23
Use `std:io::IsTerminal` instead of `atty`

vkleen:
The `atty` crate appears to be unmaintained and `std::io::IsTerminal` provides the same functionality since rust 1.70.


jneem (approved):
Do we have a MSRV policy documented anywhere?

PR #1541, opened 2023-08-24
Initial handling of incomplete input in nls

### LINKED ISSUE #1045: No LSP completion for quoted fields with spaces
**Describe the bug**
When we have the following program: 
```
let data = { foo."bar baz".blarg  = 10 } in 
data.foo."bar baz"[.]
```
We do not get any completion

**Expected behavior**
The LSP should give completion as usual

**Additional context**
This is because of the way the completion engine does string manipulation to extract identifiers. A proper solution will involve the completion engine using some components of the lexer/parser

jneem:
This adds support for parsing incomplete input to nls, enabling the new term-based completer to be used in situations like `let x = some.path` (where the cursor is at the end).

Fixes #1045.


lsp/nls/src/field_walker.rs
```diff
@@ -202,20 +202,31 @@ impl FieldDefs {
...
                 .map(|def| {
                     log::info!("got def {def:?}");
-                    let defs = def.resolve_terms(linearization, server);
+
+                    // The definition of this identifier is unlikely to belong to the
+                    // environment we started with, especially because the enviroment
+                    // mechanism is only used for providing definitions to incompletely
+                    // parsed input.
+                    let env = Environment::new();
+                    let defs = def.resolve_terms(&env, server);
```
vkleen, line 214: Is that just an optimization? I.e. would it still be correct to pass along the environment given to `resolve` instead of a new environment?

vkleen (approved):

lsp/nls/src/field_walker.rs
jneem: I think it can actually make a difference, although maybe not in any case that matters. For example:

```rust
let y = x in
let x = { bar = 1 } in
[ y.
```
(with the cursor at the end). The incomplete part of the input has initial environment has y bound to `Term::Var(x)` and `x` bound to `{bar = 1}` When trying to provide completions, we follow the initial environment to replace y by `Term::Var(x)`. Then the correct thing to do is to see that `x` is undefined and stop looking for completions, but if we don't reset the environment we'll fall back to the other `x` and erroneously provide "bar" as a completion.

vkleen: But would
```
let x = { bar = 1 } in
let y = x in
[ y.
```
provide the `bar` completion then? I'm probably missing something about the environment discipline.

jneem: Yes, because for the part of the input that parses, environments get attached to every term. So the `Term::Var(x)` in `y = x` will get inserted into the defs usage table with its own environment.

The extra `env` parameter to field resolution is an extra environment that only applies for the first lookup. It's to compensate for the fact that `[ y.` doesn't get parsed and so it doesn't get an environment attached. I'll add some better docs...

vkleen: Thanks for the explanation!

PR #1544, opened 2023-08-25
Fix contract application order in let bindings and annotations

vkleen:
In a let binding `let foo | A | B = ...` and in an infix contract annotation `foo | A | B` the contracts `A` and `B` were actually being applied in reverse order. That is, first `B`, then `A`. The contract application order for record fields was still correct, however.

The root cause was the fact that there were two places with code that merged separate annotations into a single annotation containing multiple contracts. With this change, both spots call into the same code and I've added regression tests to ensure the correct behaviour.

### DIFF
--- a/cli/tests/snapshot/inputs/pretty/let_annotations.ncl
+++ b/cli/tests/snapshot/inputs/pretty/let_annotations.ncl
@@ -1,0 +1,3 @@
+# capture = 'stdout'
+# command = ['pprint-ast']
+let foo : String | String | std.string.NonEmpty = "some very long string to blow past the 80 character line length limit" in foo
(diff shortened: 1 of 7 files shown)


jneem (approved):
I was just slightly confused by the PR description, because you said there was an issue for both let bindings and infix contract annotations, but then it looks like the order was only changed in `TypeAnnotation` because `LetMetadata` already had the right order?

vkleen:
That's correct. Both infix annotations and `LetMetadata` used the `TypeAnnotation::combine` function, whereas `FieldMetadata` had a custom `flatten` function that implemented the opposite order.

PR #1546, opened 2023-08-27
Fix cursor desyncing when using REPL

DeoTimeTheGithubUser:
When using the REPL on windows, the cursor appears out of sync because of [an issue](https://github.com/kkawakam/rustyline/issues/562) with rustyline causing colored prompts to behave strangely:

https://github.com/tweag/nickel/assets/89555032/8c129db5-09d8-4038-b3fb-e17269782e33

This changes it to use the intended `highlight_prompt` helper instead.

### DIFF
--- a/core/src/repl/mod.rs
+++ b/core/src/repl/mod.rs
@@ -25,6 +25,8 @@ use std::result::Result;
 use std::str::FromStr;
 
 #[cfg(feature = "repl")]
+use ansi_term::{Colour, Style};
+#[cfg(feature = "repl")]
 use rustyline::validate::{ValidationContext, ValidationResult};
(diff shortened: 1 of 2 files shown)


core/src/repl/mod.rs
```diff
@@ -387,6 +389,14 @@ impl InputParser {
     }
 }
 
+impl rustyline::highlight::Highlighter for InputParser {
```
vkleen, line 392:
```suggestion
#[cfg(feature = "repl")]
impl rustyline::highlight::Highlighter for InputParser {
```

vkleen (approved):
Looks good! 👍

PR #1547, opened 2023-08-27
Fix "wich" typo in documentation

DeoTimeTheGithubUser:
Changes "wich" to "which"


vkleen (approved):
Thanks for fixing that!

PR #1549, opened 2023-08-28
Speed up CI

Radvendii:
- [x] Shorten version string so more things hit the cache
- [x] Fix double pname suffix (doesn't speed anything up; was just bothering me)
- [x] Fix doc build rebuilding all dependencies
- [x] Seems to fix wasm rebuilding dependencies as well
- [x] ~~Buffer CI stdout/stderr~~ get rid of status bar in logs
  - On my machine, doing this sped up a nix build by 33%.
  - this only applied to some terminal emulators (gnome-terminal, st) but less so or not at all for others (xterm, kitty)
  - it can't hurt though. we don't need the status bar in CI certainly

---

I also looked into speeding up local builds / the LSP. One thing that struck me is that rust-analyzer rebuilds every crate, even if you've only changed one. This seems to be intentional, as people were claiming that you get downstream errors that way (if changing `nickel-lang-core` results in an error in `nickel-lang-cli`). I've never actually seen this myself, and in fact I've definitely had downstream errors missed by the LSP before, so I don't know if / how this works.

In any case, even if we wanted to "fix" this and only recompile the changed crate, it seems we're blocked on [an upstream issue](https://github.com/rust-lang/rust-analyzer/issues/12882). Helix's `workspace-lsp-roots` is not enough.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -54,7 +54,15 @@ jobs:
 
     - name: Run all checks
       run: |
-        nix flake check --print-build-logs
+        # this is just `nix flake check --print-build-logs`
+        # everything else is to buffer the output for speed
+        (
+          set +e
+          nix flake check --print-build-logs >out 2>&1
+          errcode=$?
+          cat out
+          exit $errcode
+        )
 
     - name: Typecheck benchmarks
       run: find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nix run . -- -f file typecheck
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -54,15 +54,7 @@ jobs:
 
     - name: Run all checks
       run: |
-        # this is just `nix flake check --print-build-logs`
-        # everything else is to buffer the output for speed
-        (
-          set +e
-          nix flake check --print-build-logs >out 2>&1
-          errcode=$?
-          cat out
-          exit $errcode
-        )
+          nix flake check --log-format raw-with-logs
 
     - name: Typecheck benchmarks
       run: find benches -type f -name "*.ncl" -print0 | xargs -0 -I file nix run . -- -f file typecheck
--- a/flake.nix
+++ b/flake.nix
@@ -218,23 +218,24 @@
             NICKEL_NIX_BUILD_REV = self.shortRev or "dirty";
           };
 
-          buildPackage = { pname, extraBuildArgs ? "", extraArgs ? { } }:
+          buildPackage = { pnameSuffix, extraBuildArgs ? "", extraArgs ? { } }:
             craneLib.buildPackage ({
               inherit
                 pname
+                pnameSuffix
                 src
                 version
                 cargoArtifacts
                 env;
 
-              cargoExtraArgs = "${cargoBuildExtraArgs} ${extraBuildArgs} --package ${pname}";
+              cargoExtraArgs = "${cargoBuildExtraArgs} ${extraBuildArgs} --package ${pname}${pnameSuffix}";
             } // extraArgs);
         in
         rec {
           inherit cargoArtifacts;
-          nickel-lang-core = buildPackage { pname = "nickel-lang-core"; };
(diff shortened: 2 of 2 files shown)


Radvendii:
I moved the other ideas I had to #1554

flake.nix
```diff
@@ -196,6 +196,9 @@
         let
           craneLib = crane.lib.${system}.overrideToolchain rust;
 
+          # suffixes get added via pnameSuffix
```
vkleen, line 199:
```suggestion
```
Since you scrapped the `pnameSuffix` code.

.github/workflows/continuous-integration.yml
```diff
@@ -54,7 +54,15 @@ jobs:
...
-        nix flake check --print-build-logs
+        # this is just `nix flake check --print-build-logs`
+        # everything else is to buffer the output for speed
+        (
+          set +e
+          nix flake check --print-build-logs >out 2>&1
+          errcode=$?
+          cat out
+          exit $errcode
+        )
```
vkleen, line 65: Is the subshell necessary here? I would have naively expected GitHub Actions to run each step in its own shell.

vkleen (approved):

.github/workflows/continuous-integration.yml
yannham, line 65: It might be a naive question, but what's the point of using an intermediate file? My (ignorant) understanding is that `stderr` is unbuffered, and `stdout` is buffered. If redirecting a unbuffered stream to a buffered stream makes it buffered, would

```bash
nix flake check --print-build-logs 2>&1
```

be sufficient?

flake.nix
```diff
@@ -54,7 +54,7 @@
       cargoTOML = builtins.fromTOML (builtins.readFile ./Cargo.toml);
       cargoLock = builtins.fromTOML (builtins.readFile ./Cargo.lock);
 
-      version = "${cargoTOML.workspace.package.version}_${builtins.substring 0 8 self.lastModifiedDate}_${self.shortRev or "dirty"}";
+      inherit (cargoTOML.workspace.package) version;
```
yannham, line 57: Does that impact `nickel --version` ?

flake.nix
```diff
@@ -263,9 +265,9 @@
               };
 
           benchmarks = craneLib.mkCargoDerivation {
-            inherit src version cargoArtifacts env;
+            inherit pname src version cargoArtifacts env;
 
-            pname = "nickel-lang-bench";
+            pnameSuffix = "-bench";
```
yannham, line 270: Does this make any difference with respect to building? For example, why using `pnameSuffix` here but setting a complete name in other places (`nickel-lang-cli`, `nickel-lang-cli-static`, etc.)?

flake.nix
```diff
@@ -204,9 +207,8 @@
 
           # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
           cargoArtifacts = craneLib.buildDepsOnly {
-            pname = "nickel-lang";
-            inherit src;
-            cargoExtraArgs = "${cargoBuildExtraArgs} --workspace";
+            inherit pname src;
+            cargoExtraArgs = "${cargoBuildExtraArgs} --workspace --all-features";
```
yannham, line 211: Why do we need to specify `--all-features` when building the dependencies?

yannham:
I have the impression that the rust-analyzer situation has improved recently, for some reason. Maybe the issue I experienced has been fixed upstream at some point? Anyway, as a discrete datapoint, I now find the startup time ok.

flake.nix
vkleen: I don't think it does. The version mechanism in the nix build uses `self.shortRev` and gets the version number from `Cargo.toml` itself.

flake.nix
Radvendii: Because the `doc` build uses `--all-features`, so without it the doc build will rebuild all the dependencies. (at least i think that's what was happening, and this change seemed to fix that). 

Does it cause any problems?

yannham: I don't think there's any issue, I was just curious. Thanks for the clarification!

PR #1550, opened 2023-08-28
Two issues with incomplete input

jneem:
This fixes an off-by-one error in the handling of incomplete input. It also makes import resolution in incomplete input more reliable -- previously it would only work when the file being opened is in the current working directory.

### DIFF
--- a/lsp/nls/src/incomplete.rs
+++ b/lsp/nls/src/incomplete.rs
@@ -5,6 +5,8 @@
 //! information from the unparseable part. This is necessarily somewhat
 //! heuristic.
 
+use std::path::PathBuf;
+
 use nickel_lang_core::{
     parser::lexer::{self, NormalToken, SpannedToken, Token},
     position::RawSpan,
@@ -141,7 +143,12 @@ pub fn parse_path_from_incomplete_input(range: RawSpan, server: &mut Server) -> 
     let to_parse = subtext[tokens[start].0..tokens.last().unwrap().2].to_owned();
     log::info!("extracted (hopefully) reparseable input `{to_parse}`");
 
-    let file_id = server.cache.replace_string("<incomplete-input>", to_parse);
+    // In order to help the input resolver find relative imports, we add a fake input whose parent
+    // is the same as the real file.
+    let mut path = PathBuf::from(server.cache.files().name(range.src_id));
+    path.pop();
+    path.push("<incomplete-input>");
+    let file_id = server.cache.replace_string(&path, to_parse);
 
     match server.cache.parse_nocache(file_id) {
         Ok((rt, _errors)) => Some(resolve_imports(rt, server)),
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -716,15 +716,24 @@ pub fn handle_completion(
     id: RequestId,
     server: &mut Server,
 ) -> Result<(), ResponseError> {
-    // -1 because at the current cursor position, there is no linearization item there,
-    // so we try to get the item just before the cursor.
-    let mut pos = server.cache.position(&params.text_document_position)?;
-    pos.index.0 -= 1;
+    // The way indexing works here is that if the input file is
+    //
+    // foo‸
+    //
+    // then the cursor (denoted by ‸) is at index 3 and the span of foo is [0,
+    // 3), which does not contain the cursor. For most purposes we're interested
+    // in querying information about foo, so to do that we use the position just
+    // *before* the cursor.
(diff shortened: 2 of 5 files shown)


lsp/nls/src/requests/completion.rs
```diff
@@ -716,15 +716,24 @@ pub fn handle_completion(
...
-    let mut pos = server.cache.position(&params.text_document_position)?;
-    pos.index.0 -= 1;
+    // The way indexing works here is that if the input file is
+    //
+    // foo‸
+    //
+    // then the cursor (denoted by ‸) is at index 3 and the span of foo is [0,
+    // 3), which does not contain the cursor. For most purposes we're interested
+    // in querying information about foo, so to do that we use the position just
+    // *before* the cursor.
```
Radvendii, line 726: does this mean that if the cursor is over the first letter of a variable name, we won't get completion for that variable (unless it's the first character of the line)?

lsp/nls/src/requests/completion.rs
```diff
@@ -716,15 +716,24 @@ pub fn handle_completion(
...
+    let pos = RawPos {
+        index: (cursor.index.0.saturating_sub(1)).into(),
+        ..cursor
+    };
 
     let term = server.lookup_term_by_position(pos)?.cloned();
     let sanitized_term = term
         .as_ref()
-        .and_then(|rt| sanitize_term_for_completion(rt, pos, server));
+        .and_then(|rt| sanitize_term_for_completion(rt, cursor, server));
```
Radvendii, line 736: why is this using `cursor` rather than `pos`?

lsp/nls/tests/inputs/completion-basic.ncl
```diff
@@ -60,7 +60,7 @@ in
 ### [[request]]
 ### type = "Completion"
 ### textDocument.uri = "file:///completion-basic.ncl"
-### position = { line = 12, character = 30 }
+### position = { line = 12, character = 26 }
```
Radvendii, line 63: are we just changing what these test are testing for here? should we add new tests rather than replacing these ones, or are these ones not serving a useful purpose?

lsp/nls/src/incomplete.rs
```diff
@@ -141,7 +143,12 @@ pub fn parse_path_from_incomplete_input(range: RawSpan, server: &mut Server) ->
     let to_parse = subtext[tokens[start].0..tokens.last().unwrap().2].to_owned();
     log::info!("extracted (hopefully) reparseable input `{to_parse}`");
 
-    let file_id = server.cache.replace_string("<incomplete-input>", to_parse);
+    // In order to help the input resolver find relative imports, we add a fake input whose parent
+    // is the same as the real file.
+    let mut path = PathBuf::from(server.cache.files().name(range.src_id));
+    path.pop();
+    path.push("<incomplete-input>");
```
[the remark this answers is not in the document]
jneem: Yeah, it's a fair point. We have various other inputs (like "<repl-typecheck>") following the same pattern, but I'll check how much effort it is to have an enum separating synthetic inputs from real files.

lsp/nls/src/requests/completion.rs
jneem: I think it should be fine, basically because we're using the surrounding context to provide completion; we don't actually care about the identifier under the cursor.

To be specific, In the context of a record path you should get completion, because if the input is `foo.bar` and the cursor is "over" the `b` and so you'll get the list of completions implied by `foo.`. (This is the usual behavior for LSP; the editor is in charge of doing any filtering based on prefixes etc.)

In the context of a "bare" variable name, like the `foo` in `{ foo = 1 } | { foobar | Number }` it should also be fine, because we'll see that we're inside a record and use that record's fields and annotations for completion.

lsp/nls/src/requests/completion.rs
jneem: This was the off-by-one error. The `sanitize_term_for_completion` function passes the range `[error_start, cursor)` to get reparsed. It's an exclusive end (like most ranges in rust), and so using `pos` instead of `cursor` drops one character too many.

lsp/nls/tests/inputs/completion-basic.ncl
jneem: The change shouldn't have an effect: it's moving the cursor test location from `(import "included.ncl").lala‸` to `(import "included.ncl").‸lala` to check that off-by-one behavior. One could argue that we should test both, but most of the test cases have the cursor at the end to that case is pretty well covered already...

vkleen (approved):

yannham (approved):
Beside agreement with some of @Radvendii 's comment, LGTM

PR #1551, opened 2023-08-28
Handle let, static access, and contract annotations in completion

jneem:
This improves the term-based completion to handle three more of the rules laid out in the lsp semantics notes. With this change, the only thing missing compared to the old completer is handling static types.

### DIFF
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -4,7 +4,8 @@ use lsp_types::{CompletionItem, CompletionItemKind, Documentation, MarkupContent
 use nickel_lang_core::{
     identifier::Ident,
     pretty::ident_quoted,
-    term::{record::FieldMetadata, BinaryOp, RichTerm, Term},
+    term::{record::FieldMetadata, BinaryOp, RichTerm, Term, UnaryOp},
+    typ::TypeF,
 };
 
 use crate::{identifier::LocIdent, server::Server, usage::Environment};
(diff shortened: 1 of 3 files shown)


lsp/nls/src/field_walker.rs
```diff
@@ -239,6 +240,30 @@ impl FieldDefs {
             Term::Op2(BinaryOp::Merge(_), t1, t2) => {
                 FieldDefs::resolve(t1, env, server).merge_from(FieldDefs::resolve(t2, env, server))
             }
+            Term::Let(_, _, body, _) | Term::LetPattern(_, _, _, body) => {
+                FieldDefs::resolve(body, env, server)
+            }
```
Radvendii, line 245: It seems like this works because the test passes, but I'm confused how it works. Don't we need to know what's bound in the let binding? does `env` include that binding already?

Radvendii (approved):

lsp/nls/src/field_walker.rs
jneem: `env` probably doesn't (it's a special-case thing for incomplete inputs), but variable definitions within the `let` body should already be resolved in the usage/def lookup table. So when autocompleting something like

```
(let x = { foo = 1 } in x).fo
```

we do a global traversal to resolve the fact that the inner x references the outer x. Then we ask "what fields are defined on `(let x = { foo = 1 } in x)`?" and this match case says that this is the same as "what fields are defined on `x`?". (Now, I just wrote `x`, but it's really `Term::Var(x)`, which includes the location information, so we know *which* `x` it is.) Answering this query ends up in the `Term::Var` match case, which looks up the definition of `x` and sees that it's `{ foo = 1 }`.

PR #1552, opened 2023-08-28
Add check in CLI that the input file is a Nickel type

DeoTimeTheGithubUser:
This ensures that the input file being passed with `-f` in `nickel export/doc/format` is a valid nickel file extension (currently just `.ncl`), before running actions that assume nickel code on it.

I'm not completely sure whether this is the correct course to take here, made this after accidentally corrupting an entire repo while attempting to format a directory (lol) since format doesn't also apply to dependencies of a file like does currently, but that could also be avoided by checking the syntax validity before making modifications. I can't really think of a case where you would be running nickel actions on non `.ncl` files, but if there are some just LMK and will close this.

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -1,8 +1,11 @@
 //! Command-line options and subcommands.
+use std::ffi::OsStr;
 use std::path::PathBuf;
+use std::string::String;
 
 use git_version::git_version;
 
+use crate::error::{CliResult, Error};
 use crate::{
     completions::GenCompletionsCommand, eval::EvalCommand, export::ExportCommand,
     pprint_ast::PprintAstCommand, query::QueryCommand, typecheck::TypecheckCommand,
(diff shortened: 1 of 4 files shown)


DeoTimeTheGithubUser:
Going to close this as don't think its really necessary and are probably better ways of doing this efficiently.
Only use case I could see if for a future `--dir` argument for format, but that would't really be necessary if dependency resolving is supported on it like rustfmt.

vkleen:
I'm wondering how you accidentally corrupted a repo. Did you pass the directory to Nickel using `-f` and Nickel decided to overwrite it? Or did something else happen?

DeoTimeTheGithubUser:
> I'm wondering how you accidentally corrupted a repo. Did you pass the directory to Nickel using `-f` and Nickel decided to overwrite it? Or did something else happen?

I used it with `-i`, which overwrote the files it was used on. As for how, I was experimenting with a build script that would format  all the files in a directory, and didn't check if it was a .ncl file first. Really just my fault and didn't cause much trouble overall, was just confusing at first

vkleen:
As far as I can tell, `nickel format` shouldn't try to overwrite the input file unless `topiary` managed to parse it as Nickel using the tree-sitter grammar. Maybe that grammar is too permissive sometimes? Do you have an example file where that happens that can be shared publicly?

DeoTimeTheGithubUser:
Sure, here's an example rust file:
```rust
fn main() {
    let width1 = 30;
    let height1 = 50;

    println!(
        "The area of the rectangle is {} square pixels.",
        area(width1, height1)
    );
}

fn area(width: u32, height: u32) -> u32 {
    width * height
}
```
After running with `nickel format -f example_rust.rs -i` becomes:
```rust
fnmain(
    ) {
    let width1 = 30;
    let height1 = 50;

    println!(
        "The area of the rectangle is {} square pixels.",
        area(width1, height1)
    );
}fn area (width: u32,height : u32) -> u32
      {
        width *height
      }

```

vkleen:
Thanks! So `nickel format` for some reason doesn't encounter a parse error with that. The `topiary` CLI does report a parse error, though. So something is very fishy with `nickel format` here. I'll investigate.

PR #1553, opened 2023-08-28
Add multiline string support for VSCode autoclosing pairs

DeoTimeTheGithubUser:
Adds support for multiline string literals `m%""%` to be automatically surrounded in VSCode.


vkleen (approved):
The change looks good to me. I wonder if we can do something reasonable about delimiters with multiple `%` signs. I'm also not really familiar with VSCode's behaviour here, but maybe the `m%"` delimiters would also make sense to have in `brackets` and `surroundingPairs`?

yannham:
What we do for vim is simply to expand variable-length delimiters up to `n`, for `n = 4`. It's a bit dirty but in practice we can expect that almost no one will be using more than four heading `%%%%` as a delimiter, and it's cheap to do. So, probably adding:

```
["m%%\"", "\"%%"]
["m%%%\"", "\"%%%"]
["m%%%\"", "\"%%%"]
```

to the list?

yannham (approved):

yannham:
Thanks for contributing! I'll try to release a minor update of the VSCode extension ASAP.

PR #1559, opened 2023-08-30
Render hover documentation as Markdown in LSP

deotimedev:
Renders hover documentation as markdown instead of plaintext in LSP

## Before:
image

## After:
image

### DIFF
--- a/lsp/nls/src/requests/hover.rs
+++ b/lsp/nls/src/requests/hover.rs
@@ -3,7 +3,6 @@ use lsp_server::{RequestId, Response, ResponseError};
 use lsp_types::{Hover, HoverContents, HoverParams, LanguageString, MarkedString, Range};
 use nickel_lang_core::position::TermPos;
 use serde_json::Value;
-use unindent::Unindent;
 
 use crate::{
     cache::CacheExt,
@@ -56,18 +55,7 @@ pub fn handle(
         value: ty.to_string(),
     }));
 
-    let meta = meta
-        .iter()
-        .map(|s| {
-            s.lines()
-                .map(|s| if s.is_empty() { " " } else { s })
-                .collect::<Vec<_>>()
-                .join("\n")
-                .unindent()
-        })
-        .map(MarkedString::String)
-        .collect::<Vec<_>>();
-    contents.extend(meta);
+    contents.extend(meta.iter().cloned().map(MarkedString::String));
 
     server.reply(Response::new_ok(
         id,
--- a/lsp/nls/src/requests/hover.rs
+++ b/lsp/nls/src/requests/hover.rs
@@ -57,10 +57,13 @@ pub fn handle(
                     language: "nickel".into(),
                     value: ty.to_string(),
                 }),
-                MarkedString::LanguageString(LanguageString {
-                    language: "plain".into(),
-                    value: meta.join("\n"),
-                }),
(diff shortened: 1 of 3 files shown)


lsp/nls/src/requests/hover.rs
```diff
@@ -57,10 +57,13 @@ pub fn handle(
...
                     value: ty.to_string(),
                 }),
-                MarkedString::LanguageString(LanguageString {
-                    language: "plain".into(),
-                    value: meta.join("\n"),
-                }),
+                MarkedString::String(
+                    meta.iter()
+                        .flat_map(|s| s.lines())
+                        .map(|s| s.trim())
```
vkleen, line 63: This seems to trim the identation from code snippets, which is probably not intended.

DeoTimeTheGithubUser: True, it seems like the desired behavior should be like that of the `indoc!` macro which removes only the common indentation in the lines. Should I add the implementation for that macro, [unindent](https://github.com/dtolnay/indoc/tree/master/unindent), as a dependency, or make a seperate function for that?

vkleen: Since `unindent` is in our dependency tree already for `indoc`, using that seems like a reasonable thing to do 👍

lsp/nls/src/requests/hover.rs
```diff
@@ -57,10 +58,18 @@ pub fn handle(
...
                 }),
-                MarkedString::LanguageString(LanguageString {
-                    language: "plain".into(),
-                    value: meta.join("\n"),
-                }),
+                MarkedString::String(
+                    meta.iter()
+                        .map(|s| {
+                            s.lines()
+                                .map(|s| if s.is_empty() { " " } else { s })
```
DeoTimeTheGithubUser, line 65: Fixes a strange regression with the `unindent` crate that causes strings like this to work:
image

But not this:
image

yannham (commented):
LGTM, thanks! However, I'm a bit confused as to why we need to unindent here. The doc is parsed without the common indentation prefix (which I just checked by `println!` the raw documentation of `std.contract.apply` after parsing), that is, Nickel multiline strings should have the same effect as unindent already. Do you have any idea where does this indentation come from?

DeoTimeTheGithubUser:
I see here that multiline strings should be getting parsed with indent removed:
image
However, it still seems like the indent remains in these strings anyways:
image
image

Not entirely sure why this is happening, also found this `strip_indent_doc` function, but it doesn't seem like its being used anywhere currently
image

DeoTimeTheGithubUser:
Also seems like the documentation created from the `nickel doc` command keeps the indent in it:
image
image

vkleen:
I'm tentatively suspecting that's a line ending issue and should be fixed. With `\n` line endings the `nickel doc` output is correctly unindented for me.

yannham:
Ah, `\r\n` is probably not considered an empty line, and thus the indentation prefix is set to `0` because of this. This is why we need more Windows tester :upside_down_face: Then this is a bug in Nickel itself, and the LSP shouldn't try to handle this. @DeoTimeTheGithubUser, would that be ok to get rid of unindent here? We'll take care of the issue in the parsing of multiline strings separately.

yannham:
Perfect, many thanks!

PR #1561, opened 2023-09-01
(LSP) Import completion

deotimedev:
Provides autocompletion for local files when using an `import` statement:

https://github.com/tweag/nickel/assets/89555032/8538bc3d-d1ab-4243-9d70-f63a073a63fe

### DIFF
--- a/lsp/nls/src/cache.rs
+++ b/lsp/nls/src/cache.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use codespan::{ByteIndex, FileId};
 use lsp_types::TextDocumentPositionParams;
+use nickel_lang_core::position::TermPos;
 use nickel_lang_core::{
     cache::{Cache, CacheError, CacheOp, EntryState, TermEntry},
     error::{Error, ImportError},
@@ -93,7 +94,7 @@ impl CacheExt for Cache {
                 // The unwrap is safe here because (1) we have linearized `file_id` and it must be
                 // in the `lin_registry` and (2) every resolved import has a corresponding position in
                 // the linearization of the file that imports it.
-                let pos = lin_registry.map.get(&file_id).and_then(|lin| lin.import_locations.get(&id)).unwrap();
+                let pos = lin_registry.map.get(&file_id).and_then(|lin| lin.import_locations.get(&id)).unwrap_or(&TermPos::None);
                 let name: String = self.name(id).to_str().unwrap().into();
                 ImportError::IOError(name, String::from(message), *pos)
             });
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -18,15 +18,10 @@ use lsp_types::{
     CompletionItem, CompletionItemKind, CompletionParams, Documentation, MarkupContent, MarkupKind,
 };
 use nickel_lang_core::cache::InputFormat;
-use nickel_lang_core::{
-    identifier::{Ident, LocIdent},
-    position::RawPos,
-    term::{
-        record::{Field, FieldMetadata},
-        RichTerm, Term, TypeAnnotation, UnaryOp,
-    },
-    typ::{RecordRows, RecordRowsIteratorItem, Type, TypeF},
-};
+use nickel_lang_core::{cache, identifier::{Ident, LocIdent}, position::RawPos, term::{
+    record::{Field, FieldMetadata},
+    RichTerm, Term, TypeAnnotation, UnaryOp,
+}, typ::{RecordRows, RecordRowsIteratorItem, Type, TypeF}};
 use std::ffi::OsString;
 use std::io;
 
@@ -742,7 +737,7 @@ pub fn handle_completion(
         // Don't respond with anything if trigger is a `.`, as that may be the
         // start of a relative file path `./`, or the start of a file extension
         if !matches!(trigger, Some(".")) {
-            let completions = handle_import_completion(&import, &params).unwrap_or_default();
+            let completions = handle_import_completion(import, &params).unwrap_or_default();
             server.reply(Response::new_ok(id.clone(), completions));
         }
         return Ok(());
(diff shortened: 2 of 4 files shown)


lsp/nls/src/cache.rs
```diff
@@ -93,7 +94,7 @@ impl CacheExt for Cache {
                 // The unwrap is safe here because (1) we have linearized `file_id` and it must be
                 // in the `lin_registry` and (2) every resolved import has a corresponding position in
                 // the linearization of the file that imports it.
-                let pos = lin_registry.map.get(&file_id).and_then(|lin| lin.import_locations.get(&id)).unwrap();
+                let pos = lin_registry.map.get(&file_id).and_then(|lin| lin.import_locations.get(&id)).unwrap_or(&TermPos::None);
```
DeoTimeTheGithubUser, line 97: A random panic was happening here sometimes when imports were invalid, and we're already in the process of creating an import error here.

lsp/nls/src/requests/completion.rs
```diff
@@ -780,6 +792,48 @@ pub fn handle_completion(
...
+) -> io::Result<Vec<CompletionItem>> {
+    debug!("handle import completion");
+
+    let current_file = params
+        .text_document_position
+        .text_document
+        .uri
+        .to_file_path()
+        .unwrap()
+        .canonicalize()?;
```
jneem, line 807: We've been using `nickel_lang_core::cache::normalize_path` instead of `canonicalize` because it behaves better for files that don't exist on the filesystem (which I think could happen here if they're editing a new file that hasn't yet been saved).

lsp/nls/src/requests/completion.rs
```diff
@@ -780,6 +792,48 @@ pub fn handle_completion(
...
+    let mut current_path = current_file.clone();
+    current_path.pop();
+    current_path.push(import);
+    let dir = std::fs::read_dir(&current_path)?;
```
jneem, line 811: Just sort of thinking out loud here (and definitely not a blocker for this PR), but maybe it's worthwhile also looking for completions in the filenames cached in `server.files`? There are some niche cases where that might help (files that have been created but not yet saved) but more importantly it would make this feature testable with the current testing harness (which synthesizes inputs without writing them to disk).

lsp/nls/src/requests/completion.rs
```diff
@@ -729,11 +731,27 @@ pub fn handle_completion(
...
+    let trigger = params
+        .context
+        .as_ref()
+        .and_then(|context| context.trigger_character.as_deref());
 
     let term = server.lookup_term_by_position(pos)?.cloned();
+
+    if let Some(Term::Import(import)) = term.as_ref().map(|t| t.term.as_ref()) {
+        // Don't respond with anything if trigger is a `.`, as that may be the
+        // start of a relative file path `./`, or the start of a file extension
```
jneem, line 743: What's wrong with completing the file extension?

lsp/nls/src/cache.rs
jneem: Yeah, we still have a lot of unwraps floating around in `nls` :disappointed:

jneem (changes requested):
Thanks for this! A few random comments and questions, but I think the only actual change I'd ask for is changing `canonicalize` to `normalize_path`

lsp/nls/src/requests/completion.rs
deotimedev: If we send a new completion while trying to complete the file extension (something like: `thing.<cursor>`, then if the user tab-completes the new completion suggestion it will end up as `thing.thing.ncl`

There are maybe some ways around this, like using `EditText` instead of insert or only giving the file extensions as completion, but by just ignoring the `.` it will continue as normal and `thing.` gets tab completed to `thing.ncl`

deotimedev:
^ clippy fixes

lsp/nls/src/requests/completion.rs
deotimedev: True, testing was something I was thinking about trying to implement but couldn't really think of a way without having files on disk. I'll see if I can implement using the server-cached files

lsp/nls/src/requests/completion.rs
```diff
@@ -780,6 +792,48 @@ pub fn handle_completion(
...
+    current_path.push(import);
+    let dir = std::fs::read_dir(&current_path)?;
+
+    let completions = dir
+        .filter_map(|i| i.ok().and_then(|d| d.file_type().ok().zip(Some(d))))
+        .filter(|(file_type, d)| {
+            // don't try to import a file into itself
+            d.path().canonicalize().unwrap_or_default() != current_file
+                // check that file is importable
+                && (file_type.is_dir() || InputFormat::from_path_buf(d.path().as_path()).is_some())
```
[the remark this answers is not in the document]
deotimedev: Yeah lol, was sort of confusing

lsp/nls/src/requests/completion.rs
deotimedev: Got it working
image
image

lsp/nls/src/requests/completion.rs
```diff
@@ -780,6 +796,80 @@ pub fn handle_completion(
...
+        .text_document_position
+        .text_document
+        .uri
+        .to_file_path()
+        .unwrap();
+    let current_file = cache::normalize_path(current_file)?;
+
+    let mut current_path = current_file.clone();
+    current_path.pop();
+    current_path.push(import);
```
jneem, line 816: Did you try not adding `import` to the path? If I understood correctly, your intention is that if they type `import "foo"` then you'll only try completing filenames starting with "foo"?

Because my understanding for how completion works in LSP is you're not supposed to filter the responses based on prefix: just return all the files in the right directory, and let the editor sort it out. Because the editor might want to filter by fuzzy prefix or something...

deotimedev:
Not sure if this is just a windows thing, but all the snapshot tests fail for me because of a filenotfound error when creating the test harness
image

jneem (approved):
Thanks! Let's not block this on the test issues -- I've opened an issue for that. (Edit: and it probably is a windows thing, because they work in CI and also for me locally)

PR #1562, opened 2023-09-01
Normalize line endings in string literals during parsing

vkleen:
With this change, line endings are normalized to `\n` (instead of `\r\n` on some systems) as soon as a string literal enters the parser. This fixes a problem noticed in #1559 where documentation strings weren't correctly stripped of their common indentation prefix on Windows.

The tradeoff with the simplicity of this patch is that now post-processing is necessary to produce `\r\n` line endings as output. This is in line with [precedent set by Rust](https://github.com/rust-lang/rust/issues/62865).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "wasm-repl",
     "pyckel",
 ]
+resolver = "2"
 
 [workspace.package]
 version = "1.1.1"
--- a/core/src/parser/grammar.lalrpop
+++ b/core/src/parser/grammar.lalrpop
@@ -524,10 +524,10 @@ StrChunks: RichTerm = {
         );
 
         let chunks: Vec<StrChunk<RichTerm>> = fst.into_iter()
-            .map(StrChunk::normalized_literal)
+            .map(StrChunk::Literal)
             .chain(chunks.into_iter()
                 .map(|(mut es, s)| {
-                    es.push(StrChunk::normalized_literal(s));
+                    es.push(StrChunk::Literal(s));
                     es
                 })
                 .flatten())
@@ -583,7 +583,7 @@ ChunkLiteral : String =
     <parts: ChunkLiteralPart+> => {
         parts.into_iter().fold(String::new(), |mut acc, part| {
             match part {
-                ChunkLiteralPart::Str(s) => acc.push_str(s),
+                ChunkLiteralPart::Str(s) => acc.push_str(&s),
                 ChunkLiteralPart::Char(c) => acc.push(c),
             };
 
@@ -603,7 +603,7 @@ StandardStaticString = DelimitedStaticString<"\"", "\"">;
 
 MultilineStaticString: String = DelimitedStaticString<"m%\"","\"%"> => {
     // strip the common indentation prefix
-    let chunks: Vec<StrChunk<RichTerm>> = vec![StrChunk::normalized_literal(<>)];
+    let chunks: Vec<StrChunk<RichTerm>> = vec![StrChunk::Literal(<>)];
     match strip_indent(chunks).pop().unwrap() {
         StrChunk::Literal(s) => s,
         // We build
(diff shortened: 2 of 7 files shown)


jneem (approved):
Do we also want to fail if we encounter "\r" on its own, as described in the rust issue?

core/src/parser/utils.rs
```diff
@@ -845,18 +845,8 @@ pub fn strip_indent(mut chunks: Vec<StrChunk<RichTerm>>) -> Vec<StrChunk<RichTer
     chunks
 }
 
-/// Strip the indentation of a doc metadata. Wrap it as a literal string chunk and call
-/// [`strip_indent`].
-pub fn strip_indent_doc(doc: String) -> String {
```
yannham, line 850: I guess this was just an unused function, that wasn't flagged such because of how LALRPOP works?

core/src/parser/utils.rs
```diff
@@ -845,18 +845,8 @@ pub fn strip_indent(mut chunks: Vec<StrChunk<RichTerm>>) -> Vec<StrChunk<RichTer
...
-    let chunk = vec![StrChunk::Literal(doc)];
-    strip_indent(chunk)
-        .into_iter()
-        .map(|chunk| match chunk {
-            StrChunk::Literal(s) => s,
-            _ => panic!("expected literal string after indentation of documentation"),
-        })
-        .next()
-        .expect("expected non-empty chunks after indentation of documentation")
+pub fn mk_strchunk_literal<E>(s: impl AsRef<str>) -> StrChunk<E> {
```
yannham, line 848: Should this be a function of `StrLiteral` instead? `StrChunk::literal` or something more explicit if we want to avoid confusion around the fact that this does some normalization, like `normalize_literal` or whatnot?

yannham (approved):
Same questions as @jneem :stuck_out_tongue:

core/src/parser/utils.rs
vkleen: It wasn't flagged because it's declared as `pub`. But it's not used anywhere in the repository.

core/src/parser/utils.rs
vkleen: That sounds like a good idea 👍

core/src/parser/utils.rs
yannham: Hm, I wonder if this is the case of many more such functions, as we flag quite a bunch of things with `pub`. There's probably a flag somewhere to say "flag unused stuff even if it's pub"...

vkleen:
I've moved the line ending normalizing into the lexer. This also makes sure that we reject lone carriage returns in the source file. The code has changed significantly, so this may warrant another review.

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
     "wasm-repl",
     "pyckel",
 ]
+resolver = "2"
```
yannham, line 11: Ah, thanks, almost did it in the previous PR, then forgot about it

core/src/parser/grammar.lalrpop
```diff
@@ -946,9 +946,9 @@ extern {
 
     enum Token<'input> {
         "identifier" => Token::Normal(NormalToken::Identifier(<&'input str>)),
-        "str literal" => Token::Str(StringToken::Literal(<&'input str>)),
+        "str literal" => Token::Str(StringToken::Literal(<String>)),
```
yannham, line 949: Ah, this means that now, we have to allocate each and every string literal, while we could refer to the original input before. It's not necessary a showstopper, but I didn't think about this consequence. Not for this PR, but I wonder if using something like `Cow<'input>` could reduce the memory footprint, if it's ever an issue

yannham (approved):

core/src/parser/grammar.lalrpop
vkleen: We are allocating a `String` for each `StrChunk::Literal` in the parser anyway, so this might not be as big of a deal as it seems.

yannham: But we don't reuse the allocation, do we? I think it's used in the `acc.push_str(&s)` when building a string literal. If I understand correctly, the previous situation was: for each `ChunkLiteralPart` (or whatever it's called) we refer directly to the input, so no allocation, and the when building the actual `ChunkLiteral`, we're accumulating, where we allocate. Now we allocate for each `ChunkLiteralPart`, and then we consume them into `ChunkLiteral`, so the allocation are released but not reused. So we should be doing `total_count_of_chunk_literal_part` new allocations as compared to before? That being said, memory isn't really an issue, because they should be short-lived. It might hurt performance a bit though.

A possiblity would be to make a simple optimization for the case of "only one chunk literal part", where we could directly take the inner String instead of starting from `String::new()` and then pushing it once with `push_str`.

PR #1566, opened 2023-09-01
Revert to special casing the parsing of `"%...%{` in multiline strings

yannham:
This reverts PR #1435.

After more investigation, it turns out the original change made it very
annoying to interpolate something between double quotes, because it
conflicts with the multiline string end delimiter:

```nickel
m%" "%{hello}" "%
```

After the reverted change, this wouldn't parse correctly anymore. One
can escape the first quote, but it's verbose and arguably harder to read:

m%" %{"\""}%{hello}" "%

Additionally, we can't see any case where an interpolation sequence with
heading double quotes would have a valid meaning in the reverted PR
model: it's always an error at runtime (result in applying a string to something or taking the modulo of a string and a record...). Thus, if there's a sequence `"%{` in a string, there is no other meaningful
interpretation than the "double quote followed by zero or more percents followed by an interpolation
sequence".

We decided to revert the original change, although it derives a bit from
the most straightforward interpretation of the syntax of multiline string. Documentation is added in the manual for this special case.


core/src/parser/lexer.rs
```diff
@@ -823,9 +829,27 @@ impl<'input> Lexer<'input> {
                 self.enter_normal();
                 Token::MultiStr(tok)
             }
+            // If we encounter a `QuotesCandidateInterpolation` token with as much `%` characters
```
jneem, line 832:
```suggestion
            // If we encounter a `QuotesCandidateInterpolation` token with as many `%` characters
```

jneem (approved):

PR #1569, opened 2023-09-04
Add infrastructure for on-demand ARM64 runners on AWS

vkleen:
With this change, ARM64 release artifacts will be built automatically by a GitHub workflow. Since GitHub doesn't offer hosted runners running on ARM64, we're spinning up an EC2 spot instance on demand and run the jobs building ARM64 artifacts there.

As a fun side note, the Terraform infrastructure code is written entirely in Nickel.

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -8,13 +8,84 @@ on:
         description: "The release tag to target"
 
 permissions:
+  id-token: write
   contents: write
   packages: write
 
 jobs:
+  start-runner:
+    name: Start EC2 runner
+    runs-on: ubuntu-latest
+    outputs:
+      instance_id: ${{ steps.invoke-start.outputs.INSTANCE_ID }}
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v3
+        with:
+          role-to-assume: ${{ secrets.EC2_ROLE }}
+          aws-region: ${{ vars.EC2_REGION }}
+      - name: Start EC2 instance
+        id: invoke-start
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          EC2_START: ${{ secrets.EC2_START }}
+        run: |
+          RUNNER_TOKEN=$(gh api -X POST -q '.token' /repos/${{ github.repository }}/actions/runners/registration-token)
+          aws lambda invoke \
+            --cli-binary-format raw-in-base64-out \
+            --function-name "$EC2_START" \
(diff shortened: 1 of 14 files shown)


.github/workflows/release-artifacts.yaml
```diff
@@ -8,13 +8,84 @@ on:
...
+          echo 'Waiting for GitHub runner to start'
+          while [[ -z "$(gh api /repos/${{ github.repository }}/actions/runners | jq '.runners[] | select(.name == "ec2-spot")')" ]]; do
+            sleep 60
+          done
+          echo 'Done 🎉'
+
+  stop-runner:
+    name: Stop EC2 runner
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
```
yannham, line 50: What is the semantics of this line?

infra/README.md
```diff
@@ -0,0 +1,49 @@
+# GitHub Runner Infrastructure
+
+To redeploy and update GitHub variables, make sure you are logged into AWS and
```
yannham, line 3: I think this part could be slightly clearer about what is needed when. I would suggest something like starting by what is the trigger that needs to do additional work (as opposed to just let the CI do its stuff), and then order the guide by steps:

> If you make any change to the infrastructure code located in this directory,
> you will have to redeploy it. To do that, you'll need to update GitHub variables as well. Do the following:
>
> 1. Make sure you are logged into AWS. You can check that using `awscli2`:
>
> ```console
> nix run nixpkgs#awscli2 -- sts get-caller-identity
>{
>  # CENSORED
>}
>```
>
> If not, log in with AWS SSO credentials, follow [their guide][aws-sso-guide].
>
> 2. Make sure your a logged into GitHub.  You need the requisite permissions
on the Nickel repository to proceed further [..].
> 3. Update the infrastructure:
>   ```console
>   nix develop ..#infra -c update-infra
>   ```

yannham (approved):
I'm not really an expert on the Terraform part, but it looks good as far as I can tell

.github/workflows/release-artifacts.yaml
vkleen: It makes sure the stop action runs even if the previous jobs failed or got canceled.

yannham: Ah, so the `if:` default value isn't always, but sometimes :stuck_out_tongue:  that's a detail but it might be worth adding it as a comment.

vkleen: I think there's a silent default "only run if dependencies succeeded" that this explicit `if` overrides. But it feels really quite silly :laughing:

PR #1578, opened 2023-09-06
Implement curried dot operator

suimong:
per #1568 , so that `(.) field record` works. It is very pleasant to see that all it takes is just adding four lines of code, and everything works, including the language server. Thank you @yannham & @vkleen for the kind help.


core/src/parser/grammar.lalrpop
```diff
@@ -673,6 +673,9 @@ MatchCase: MatchCase = {
 
 // Infix operators by precedence levels. Lowest levels take precedence over
 // highest ones.
+InfixBOp1: BinaryOp = {
```
vkleen, line 676:
```suggestion
InfixBOpAccess: BinaryOp = {
```
A minor nitpick, but the other `InfixBOp` productions are named according to their operator precedence in `InfixExpr`. In fact, I'm tempted to say, that it would be better to put a case like
```
<l: @L> "." <r: @R> =>
    InfixOp::from(BinaryOp::DynAccess())
        .eta_expand(mk_pos(src_id, l, r)),
```
into the `CurriedOp` production, because `.` is never parsed as a generic infix operator. It rather has its own parsing rules as part of  `RecordOperationChain`.

suimong: Makes sense. Moved implementation inside `CurriedOp`.

PR #1580, opened 2023-09-06
Cache more

Radvendii:
Right now in the nix build we set the `NICKEL_NIX_BUILD_REV` environment variable, which affects the string printed by `nickel --version`. However, this invalidates the cache on every commit (and when building dirty), even if nothing in the code was changed it has to be recompiled in CI, and also locally if the repo is dirty.

This change puts a static string in that spot, which can then be replaced in the resulting binary with a cheap fixup derivation. That derivation will need to be rebuilt on every commit, but it's extremely fast. This only needs to be done for `nickel-lang-cli`, since that's the only derivation that actually uses the git revision.


flake.nix
```diff
@@ -229,16 +260,15 @@
...
+                cargoArtifacts;
 
               cargoExtraArgs = "${cargoBuildExtraArgs} ${extraBuildArgs} --package ${pname}${pnameSuffix}";
             } // extraArgs);
         in
         rec {
           inherit cargoArtifacts;
           nickel-lang-core = buildPackage { pnameSuffix = "-core"; };
-          nickel-lang-cli = buildPackage { pnameSuffix = "-cli"; };
+          nickel-lang-cli = fixupGitRevision (buildPackage { pnameSuffix = "-cli"; extraArgs = { inherit env; }; });
```
vkleen, line 271: `nickel-static` also needs the `env` in `extraArgs`.

flake.nix
```diff
@@ -206,6 +206,37 @@
...
+          mkdir -p $out/bin
+          for srcBin in $src/bin/*; do
+            outBin="$out/bin/$(basename $srcBin)"
+            # [dirty] must have 7 characters to match dummyRev (hard coded in nickel-lang-cli)
+            # we have to pad them out so they fit in the same spot in the binary
+            bbe -e 's/${dummyRev}/${padWith dummyRev (self.shortRev or "[dirty]")}/' \
+              $srcBin > $outBin
+            chmod +x $outBin
+          done
+        '';
```
vkleen, line 237:
```suggestion
        '';
        meta.mainProgram = pkgs.meta.mainProgram;
```

yannham (approved):

flake.nix
```diff
@@ -194,6 +194,37 @@
...
+      dummyRev = "DUMMYREV_THIS_SHOULD_NOT_APPEAR_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+
+      # pad a string with the tail of another string
+      padWith = pad: str:
+        str +
+        builtins.substring
+          (builtins.stringLength str)
+          (builtins.stringLength pad)
+          pad;
+
```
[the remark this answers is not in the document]
Radvendii: I had a little bit of that info up above `dummyRev`, but I added a more extensive comment on `fixupGitRevision`

PR #1583, opened 2023-09-07
Fix LSP regression

### LINKED ISSUE #1574: LSP regression when hovering on the `function` field in `std.xxx.function`
**Describe the bug**

Hovering on an item from the stdlib doesn't always provide the associated metadata (type and documentation), but just shows `Dyn`. This is a regression since 1.1.1, because the latter does show the correct information.

**To Reproduce**

Take the following snippet:

```nickel
let map_cartesian
  : forall a b c. (a -> b -> c) -> Array a -> Array b -> Array c
  = fun f left right =>
    std.array.fold_left
      (
        fun acc_ext left =>
          acc_ext
          @ (
            std.array.fold_left (fun acc_int right => std.array.append (f left right) acc_int) [] right
          )
      )
      []
      left
  in
map_cartesian (++) ["1", "2", "3"] ["a", "b", "c"]
```

And hover on one of the occurrence of `fold_left` in `std.array.fold_left`. The LSP just shows `Dyn`.

**Expected behavior**
I expect the LSP to show the type and the documentation of `std.array.fold_left`, as it does in version 1.1.1.

**Environment**
 - Version of the code: 795e2677fe0bc314984223067b704669723449fb

yannham:
Closes #1574.

WIP. Trying to fix a new indexing error in the linearization.


core/src/typecheck/linearization.rs
```diff
@@ -88,13 +89,17 @@ pub trait Linearizer {
     ///
     /// In practice this mainly includes environment information. Providing this
     /// state is the responsibility of [Linearizer::scope]
+    ///
+    /// Returns an item id that can then be used to complete e.g. type information. Some terms
+    /// might give rise to several items in the linearization, but usually, there's a main one that
+    /// is relevant for amending the type. The result can be `None` if the term doesn't have a
+    /// location, in which the LSP currently ignores it.
```
vkleen, line 96:
```suggestion
    /// location. In this case, the LSP currently ignores it.
```

core/src/typecheck/mod.rs
```diff
@@ -2109,27 +2133,34 @@ fn check_field<L: Linearizer>(
...
+    infer_with_annot(state, ctxt, lin, linearizer, annot, Some(rt), item_id)
 }
 
 /// Function handling the common part of inferring the type of terms with type or contract
 /// annotation, with or without definitions. This encompasses both standalone type annotation
 /// (where `value` is always `Some(_)`) as well as field definitions (where `value` may or may not
 /// be defined).
+///
+/// As for [check_visited] and [infer_visited], the additional `item_id` is provided when the term
+/// has been added to the linearizer before but can still benefit from an updating its information
```
vkleen, line 2162:
```suggestion
/// has been added to the linearizer before but can still benefit from updating its information
```

core/src/typecheck/mod.rs
```diff
@@ -2224,7 +2284,10 @@ fn infer<L: Linearizer>(
                 .cloned()
                 .ok_or(TypecheckError::UnboundIdentifier(*x, *pos))?;
 
-            linearizer.add_term(lin, rt, x_ty.clone());
+            if let Some(ref id) = item_id {
+                linearizer.retype(lin, id, x_ty.clone());
+            }
```
vkleen, line 2289: The amount of `if let Some(...)` in this function makes we wonder if it wouldn't make more sense to have `retype` operate on `Option<ItemId>`? Or, if that feels iffy semantically, we could have a wrapper that does that.

lsp/nls/src/linearization/building.rs
```diff
@@ -88,7 +88,7 @@ impl<'b> Building<'b> {
             // specific case, we simply ignore it (meaning that one can't do goto references for
             // `std` currently).
             TermKind::Record(_) => (),
-            // unreachable()!: add_usage can only called on let binding, functions and record
+            // unreachable()!: add_usage can only be called on let binding, functions and record
```
vkleen, line 91:
```suggestion
            // unreachable()!: add_usage can only be called on let bindings, functions and record
```

vkleen (approved):

lsp/nls/src/linearization/building.rs
```diff
@@ -306,8 +308,11 @@ impl<'b> Building<'b> {
         }
     }
 
-    pub(super) fn id_gen(&self) -> IdGen {
-        IdGen::new(self.linearization.len())
+    /// Return the id of the next item to be inserted. Note that `next_id` doesn't mutate anything:
+    /// as long as no new item in inserted in the linearization, calling `next_id` multiple time
```
yannham, line 312:
```suggestion
    /// as long as no new item is inserted in the linearization, calling `next_id` multiple time
```

core/src/typecheck/mod.rs
yannham: That's a good idea. `retype` is only ever called from the typechecker, and the linearizer will be scrapped ultimately, so it's probably not very rewarding to try to keep the `retype` signature more "natural" and write a second wrapper somewhere else, so I'll probably just change the signature of `retype`.

yannham: Addressed in [f39e49b](https://github.com/tweag/nickel/pull/1583/commits/f39e49b5571473d2208b69cb33dd4efd3d31ef6b)

core/src/typecheck/mod.rs
```diff
@@ -2204,18 +2237,35 @@ fn infer_with_annot<L: Linearizer>(
...
+    lin: &mut Linearization<L::Building>,
+    linearizer: L,
+    rt: &RichTerm,
+) -> Result<UnifType, TypecheckError> {
+    infer_visited(state, ctxt, lin, linearizer, rt, None)
+}
+
+/// Variant of [infer] taking an additional `item_id` to indicate that inference is done on a term
+/// that has already been visited by the linearizer, for example when coming directly from [check].
+/// When `item_id` is defined, the linearizer isn't called. [infer] is just `infer_visited` called
```
jneem, line 2251:
```suggestion
/// When `item_id` is defined, the term `rt` will not be added to the linearizer. [infer] is just `infer_visited` called
```

core/src/typecheck/mod.rs
```diff
@@ -2204,18 +2237,35 @@ fn infer_with_annot<L: Linearizer>(
...
 ) -> Result<UnifType, TypecheckError> {
     let RichTerm { term, pos } = rt;
 
-    // For the following infer rules, we need to be careful to call to `linearizer.add_term` at the
-    // right moment. We need to provide a type, so we can't do it before having inferred a type for
-    // `rt`. However, we must do it **before** any subsequent calls to `check` or `infer`, because
-    // the linearizer relies on terms being visited in order (pre-order).
+    // We don't have a type yet, so we put `Dyn` for now, but we'll update once we infer one via
+    // `linearizer.retype`
+    let item_id = item_id.or_else(|| linearizer.add_term(lin, rt, mk_uniftype::dynamic()));
```
jneem, line 2267: I'm a little worried about the fact that `item_id == None` could have two meanings: it wasn't added to the linearizer, or it was added but the linearizer returned `None`. In the second case, we could still call `add_term` twice, right?

jneem (approved):

core/src/typecheck/mod.rs
yannham: Technically, yes. Adding such a term twice is currently not an issue, because if the linearizer returned `None` once, it'll always return `None` (the term doesn't have a position). On those terms `add_term` is basically a no-op. Though it's a bit fragile and specific to the current linearizer implementation.

I'll try to think if there's a simple change that makes it better, or if that'll do for now as we're going to scrap the linearizer soon anyway.

jneem: If it doesn't break anything right now, I'd be inclined to just not worry :shrug:

PR #1595, opened 2023-09-11
Cleanup the code a bit (a lot)

gtsiam:
I was going to work on simplifying the `ImportResolver` trait, but I got very sidetracked with fixing various paper cuts in the code base. So I thought I'd make a pr with only those changes now so I don't go to merge conflict hell later on.

Don't let the huge diff scare you, the only meaningful change is the first one where I removed the explicit state from the traverse trait. Rust closures can capture state, so it shouldn't be necessary. That said, I had to do some copy-pasting to make it work, as the `RecordRows::try_map_state` family of methods you had before are incompatible with this. I might (if I find the time) rework the relevant datastructures to use `Vec`s underneath at some point, which should also drastically bring down the allocations.

The rest is all formatting plus a few `Default` impls to make clippy happy.


core/src/eval/tests.rs
```diff
@@ -338,6 +338,13 @@ fn substitution() {
...
         subst(&eval_cache, t, &initial_env, &env),
-        parse("match {'x => [1, 1], 'y => (if false then 1 else \"Glob2\"), 'z => {id = true, other = false}} true").unwrap()
+        parse(
+            "match {\
+                'x => [1, 1], \
+                'y => (if false then 1 else \"Glob2\"), \
+                'z => {id = true, other = false}\
+            } true"
+        )
+        .unwrap()
```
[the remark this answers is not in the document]
gtsiam: 1. I didn't to avoid adding extra whitespace to the test. Since I was making a lot of changes I tried to avoid changing the meaning of any code. In this case, yes. We can. It would technically add some whitespace to the string, but eh ¯\\\_(ツ)\_/¯.
2. I have vertical ruler at col. 100 on vscode, and I'm guessing the other one wasn't past the 100 mark. Again, rapid fire changes.

gtsiam:
> We're also having a hacking session tomorrow at 15:15 UTC. I think we'll take a look at the Traverse stuff then; let me know if you want to join for that.

I can't, unfortunately. But I'd say traverse is fine-ish for now. It might be a good idea to convert it to a `Visitor` similar to the one in `syn` at some point, but I don't know enough about the code base to say that.

 If you want something to look at, I'd say look at `RecordRowsF` / `EnumRowsF`. I gave fixing that code a shot, but I didn't know much about some of the code that interacted with (say the `UnifRecordRows` stuff). The diff became to large and intricate, so I just gave up. Specifically:

```rust
pub enum RecordRowsF<Ty, RRows> {
    Empty,
    Extend { row: RecordRowF<Ty>, tail: RRows },
    TailVar(LocIdent),
    TailDyn,
}
```

becomes:
```rust
pub enum RecordRows {
    rows: Vec<RecordRow>,
    tail: Option<LocIdent>,
}
```

and the same applies to all similar linked list-y code. Unfortunately, Rust isn't Haskell. It has functional aspects, but still.

To quote [Learning Rust With Entirely Too Many Linked Lists](https://rust-unofficial.github.io/too-many-lists/index.html#an-obligatory-public-service-announcement):
> 99% of the time you should just use a Vec (array stack), and 99% of the other 1% of the time you should be using a VecDeque (array deque)

Not bashing linked lists in functional languages, but in systems languages they're just a bad idea (at the very least from a memory layout perspective).

gtsiam:
Actually, I forgot: After making the pr, I found that [`IdentKind`](https://github.com/tweag/nickel/blob/aa1d9335c7ae66ccca1dc709675ff5026746b243/core/src/eval/mod.rs#L842)  is not used anywhere, so I removed it. I'll push this commit here too while I'm at it.

gtsiam:
And I completely forgot to address some of the things in your comment, so here goes:

> It seems like a good number of them should be getting caught by cargo fmt. I'll put that on my list of things to look into.

This is largely `match_sharedterm!`'s fault.

> It would be nice also if the comments would get reflowed in the pre-commit-hook / CI. @vkleen mentioned that unstable versions of cargo fmt do that, so maybe we should switch to that.

Didn't know nightly `cargo fmt` did that! TIL. I use the [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) extension.

> Why don't we merge this PR quickly without the Traverse changes, ...

I'll consider this nicely solved by your change :)

yannham:
Hey @gtsiam, many thanks for contributing! I haven't looked at the code yet (I was out sick), but I'll give it a review today. First things first, I would like to remind this passage in the preamble of CONTRIBUTING.MD:

> Before contributing any non-trivial change to this repository, please first check for the existence of related work such as issues or open pull requests. If there are none, it's better to discuss the change you wish to make via an issue, by email, or using any other method with the maintainers of this repository (listed below) before actually submitting something.

Sometimes things are what they are for non obvious technical reasons (and sometimes for no good reason though, beside  historical ones, so it's good to kick the anthill :slightly_smiling_face: !). The easiest is probably to join us on the [our Matrix (on the Developer chan)](https://matrix.to/#/#nickel-lang:matrix.org) for example.

About the linked list representation of row types, it has been done this way because it's more amenable to unification during type inference. You can think of them as a form of difference list (although the concatenation is not really O(1), because with Rust's borrowing it's probably annoying to keep a direct pointer to the tail, but it could be). So we need to keep _some_ form of linked-listness at least, so that we can unify the tail with something new. In this case the right representation isn't totally obvious to me: if you create a lot of one-element lists that you unify together one by one (which is what the current algorithm is doing when unifying two record rows), then you might end up with just a linked list but where elements are one-element `Vec<T>` instead of just being a `T`. We could have different representation for `Type` and `UnifType` but now this duplicates quite a bit of code.

That being said, the current setting is far from being the only, or even the best solution: we could probably do something better with vectors. But it's just to say that changing the representation to `Vec` without considering the rest isn't guaranteed to be automatically better (it can, but this has to be measured). It might make the code simpler as well though, so it's worth to investigating. Maybe in a separate PR?

core/src/repl/mod.rs
```diff
@@ -270,17 +272,16 @@ impl<EC: EvalCache> Repl for ReplImpl<EC> {
...
         let wildcards =
             typecheck::type_check(&term, self.env.type_ctxt.clone(), self.vm.import_resolver())?;
-        // Substitute the wildcard types for their inferred types
-        // We need to `traverse` the term, in case the type depends on inner terms that also contain wildcards
+        // Substitute the wildcard types for their inferred types We need to `traverse` the term, in
+        // case the type depends on inner terms that also contain wildcards
         let term = term
             .traverse(
-                &|rt: RichTerm, _| -> Result<RichTerm, std::convert::Infallible> {
+                &mut |rt: RichTerm| -> Result<RichTerm, Infallible> {
```
yannham, line 279: I'm not necessarily convinced that importing a symbol for a single use is better than inlining it. Do you happen to know of any kind of official and widespread guideline telling to do so?

core/src/term/mod.rs
```diff
@@ -926,18 +955,20 @@ pub enum UnaryOp {
...
+    /// content of the field. `RecordMap` then replaces the content of each field by the result of
+    /// the function: i.e., `recordMap f {a=2;}` evaluates to `{a=(f "a" 2);}`.
     RecordMap(),
 
     /// Inverse the polarity of a label.
     ChangePolarity(),
 
     /// Get the polarity of a label.
     Pol(),
+
```
yannham, line 971: It's not tremendously important, but I wonder if adding new lines everywhere makes it more readable: op used to be grouped somehow by theme (e.g `StrXXX`), but now there is no structure anymore. That being said the grouping was somehow historical.

Interestingly I wondered if inserting new lines between variants or fields was a common practice, thus I went to check in the source code of `rustc`. It seems that there is no logic there - sometimes enums and structs don't have a newline between fields and variants, sometimes they do. There seems to be, at least empirically, a notion of size - large structs/enums tend to be split.

yannham (approved):
Thanks a lot for cleaning that up :slightly_smiling_face:  ! To be frank, it was quite hard to digest. I would suggest, next time, to split things up - here for example separate the traverse change, the removal of `IdentKind`, and the cosmetic changes. The reason is that not only a big diff is harder to review, but it's even harder when most of it is mechanical but _some_ of it is not only formatting but more semantically meaningful.

That being said, I think it's better for the most part. I left a few comments but I don't see any blocker otherwise.

Radvendii (approved):

core/src/error.rs
```diff
@@ -1635,15 +1641,19 @@ impl IntoDiagnostics<FileId> for ParseError {
             ParseError::InvalidAsciiEscapeCode(span) => Diagnostic::error()
                 .with_message("invalid ascii escape code")
                 .with_labels(vec![primary(&span)]),
-            ParseError::StringDelimiterMismatch { opening_delimiter, closing_delimiter } => Diagnostic::error()
+            ParseError::StringDelimiterMismatch {
+                opening_delimiter,
+                closing_delimiter,
+            } => Diagnostic::error()
```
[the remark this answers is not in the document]
Radvendii: Yeah. Some of the time the issue is `match_sharedterm!`, but e.g. here it seems something else is going wrong. @jneem poked around at it a bit, but it's still not clear yet when/why some code isn't getting formatted.

gtsiam:
Sorry, I've been away from my computer for a few days.

> About the linked list representation of row types, it has been done this way because it's more amenable to unification during type inference. ...

Thanks for the explanation! Still, I do think I'll try to kick that particular ant hill with some code sooner or later. Definitely in another pr.

core/src/repl/mod.rs
gtsiam: I don't think there's any. That said, I tend to value type clarity over avoiding `use`s. It's not like it adds any real overhead, but long type names definitely make code that much harder to understand. So the cost calculation for me becomes:
- near zero cost to adding the use
- definitely non-zero cost to removing it

So I tend to add it. That said, not like this is complicated code; We could remove it if you prefer.

core/src/term/mod.rs
gtsiam: > I wonder if adding new lines everywhere makes it more readable

Definitely does for me, at least when each "item" is more than one line.

> Interestingly I wondered if inserting new lines between variants or fields was a common practice

You're not likely to find guidance on this, since rust APIs don't tend to expose the internals of their types. Grouping is also mostly fine, but in my experience it stops making sense rather quickly. If you wanted to variants to be that similar, you'd have merged them into one anyway.

core/src/repl/mod.rs
yannham: I was mostly curious about the status quo elsewhere. I personally don't entirely agree (sometimes I find full qualified names clearer, especially when there are a lot of symbols floating around, because you immediately see what namespace they relate to without having to go to their introduction site). But `Infallible` is pretty standard, it's mostly a matter of taste, and above all it's quite unimportant, so please don't bother removing this :slightly_smiling_face:

core/src/term/mod.rs
yannham: > You're not likely to find guidance on this, since rust APIs don't tend to expose the internals of their types. 

I was curious about any kind of unspoken standard, even if used only internally inside rustc. As for the `use` comment it's a detail anyway. I'm fine with the current way

yannham:
@gtsiam got it! I took the freedom to fix the merge conflict in the meantime, so that it's in a state to be merged.

PR #1599, opened 2023-09-12
Statically merge partial record definitions

### LINKED ISSUE #1427: Stack overflow on a rather big, but not deep record.
**Describe the bug**

I've been playing with generating `.ncl` file with `nixpkgs` metadata, and got stack overflow when trying to evaluate it with Nickel.

**To Reproduce**

When in try to evaluate [result.ncl](https://github.com/tweag/nickel/files/11937153/result.ncl.txt), I get this with current master:

```
 % nix run github:tweag/nickel -- export -f result.ncl

zsh: abort      nix run github:tweag/nickel -- export -f result.ncl
```

I tried removing some lines from it, and at some point the overflow doesn't happen anymore. I can provide a bigger example with all of nixpkgs data (5 Mb) if needed.

**Expected behavior**

It shouldn't fail, it should generate JSON output.

**Environment**
 - OS name + version: macOS Ventura 13.4.1
 - Version of the code: `github:tweag/nickel/b1458762817c1711e1502d5da6cde2df38edf54b`

**Additional context**

### LINKED ISSUE #1597: Large piecewise records create deeply nested ASTs and lead to stack overflow when type checking
Consider the following code: https://gist.github.com/Quantum64/6c82bd35cb386b9d5d7e8cfd6d028629

`nickel pprint-ast` yields a deeply nested AST:  https://gist.github.com/Quantum64/24bbf4cdf4ad183e32fd6b7498b6e536

`nickel typecheck` crashes with a stack overflow:
```
```

vkleen:
This is an attempt implementing @yannham's suggestion for resolving #1427. Instead of always translating a partial definition into an application of `&`, I'm statically resolving merges of non-recursive records.

Unfortunately, I've had to duplicate some of the merging logic from `eval::merge::merge_fields`. The latter function is too tightly coupled to the evaluator to use in the parser and I haven't found a nice way to factor out the common elements.

Fixes #1427
Fixes #1597

### DIFF
--- a/core/src/eval/merge.rs
+++ b/core/src/eval/merge.rs
@@ -26,12 +26,13 @@
 //! depend on each metadata.
 
 use super::*;
+use crate::combine::Combine;
 use crate::error::{EvalError, IllegalPolymorphicTailAction};
 use crate::label::{Label, MergeLabel};
 use crate::position::TermPos;
 use crate::term::{
     record::{self, Field, FieldDeps, FieldMetadata, RecordAttrs, RecordData},
-    BinaryOp, IndexMap, RichTerm, Term, TypeAnnotation,
+    BinaryOp, IndexMap, RichTerm, Term,
 };
 use crate::transform::Closurizable;
 
@@ -320,7 +321,7 @@ Append `, ..` at the end of the record contract, as in `{some_field | SomeContra
                     // about them anymore, and dependencies are stored at the level of revertible
                     // cache elements directly.
                     Term::RecRecord(
-                        RecordData::new(m, RecordAttrs::merge(r1.attrs, r2.attrs), None),
+                        RecordData::new(m, RecordAttrs::combine(r1.attrs, r2.attrs), None),
                         Vec::new(),
                         None,
                     ),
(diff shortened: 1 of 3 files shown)


core/src/parser/utils.rs
```diff
@@ -539,17 +525,56 @@ where
 /// Merge two fields by performing the merge of both their value (dynamically, by introducing a
 /// merging operator) and their metadata (statically).
```
jneem, line 526: Perhaps it's worth updating this comment to note that we don't always introduce the merge operator.

jneem (approved):

core/src/eval/merge.rs
```diff
@@ -407,47 +408,23 @@ fn merge_fields<'a, C: Cache, I: DoubleEndedIterator<Item = &'a LocIdent> + Clon
...
+            priority,
+        },
         value,
         pending_contracts,
     })
 }
 
 /// Merge two optional documentations.
-fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
+pub(crate) fn merge_doc(doc1: Option<String>, doc2: Option<String>) -> Option<String> {
```
yannham, line 427: Should we implement this as `Combine`? Although the semantics is a bit arbitrary, we still use it as such. And I believe we have something like `impl<T: Combine> Combine for Option<T>` somewhere?

core/src/parser/utils.rs
```diff
@@ -536,30 +522,88 @@ where
...
+        // [`FieldMetadata::combine`] produces subtly different behaviour from
+        // the runtime merging code, which is what we need to replicate here
+        metadata: FieldMetadata {
+            doc: merge_doc(field1.metadata.doc, field2.metadata.doc),
+            annotation: Combine::combine(field1.metadata.annotation, field2.metadata.annotation),
+            opt: field1.metadata.opt && field2.metadata.opt,
+            not_exported: field1.metadata.not_exported || field2.metadata.not_exported,
+            priority,
+        },
+        pending_contracts: Vec::new(),
```
yannham, line 606: It's not very important, but the idea of using `field1.pending_contracts` was to avoid dropping an empty `Vec` just to recreate one just after.

yannham (approved):
LGTM. I find the duplication minimal :+1: : there is so much more going on in the `eval::merge` function, which as you say is totally coupled to the environment, the cache, etc.

core/src/parser/utils.rs
vkleen: My thought process was: creating an empty `Vec` doesn't allocate and if for some reason `field1.pending_contracts` was at some point nonempty, then got cleared and then we end up here, we might as well drop its allocation.

In the end, this probably never happens and we're talking about 16 bytes being initialized to 0. 🤷

core/src/eval/merge.rs
vkleen: We do. But `String` doesn't implement `Combine` (yet) and as the comment indicates, we're not entirely sure how to combine documentation. So this should probably be another refactor, replacing `Option<String>` by a newtype and implementing an explicit `Combine` instance for it.

core/src/parser/utils.rs
yannham: I think it's 24 bytes with cap (ptr + size + cap), which is huge! Jokes aside, yes, it's unimportant. And TIL that an empty vector doesn't allocate before something is pushed in it.

PR #1604, opened 2023-09-14
Put embedded topiary formatting behind a feature flag for NLS

vkleen:
Unfortunately, in the current state we can't publish Nickel to crates.io with a dependency on topiary. However, we can pretend the `format` feature doesn't exist and hide the dependency from crates.io this way.

This change puts the NLS formatting code that depends on topiary behind a feature flag, just like we have in `nickel-lang-cli`. To not lose formatting completely in the version published on crates.io, it restores the old formatting code when the `format` feature is disabled. This code just calls out to an external `topiary` binary which must be in `PATH`.


lsp/nls/src/requests/formatting.rs
```diff
@@ -31,7 +31,9 @@ pub fn handle_format_document(
         file: params.text_document.uri,
     })?;
 
-    let result = Some(vec![TextEdit {
+    // TODO: instead of always sending a huge edit, we should compute a diff
+    // between `text` and `formatted` and send more granular edits.
+    let result = (text != formatted).then_some(vec![TextEdit {
```
yannham, line 36: Do you think testing for equality on a full text is necessarily a good tradeoff? The cost is the offset of the first differing character, which can be big if the source is big and the formatting only happen at the end (or is equal). I wonder if we really gain something over sending the whole text anyway...

yannham (approved):

lsp/nls/src/requests/formatting.rs
vkleen: I cribbed that from rust-analyzer and wasn't really thinking of performance, but of not sending edits that do nothing. That prevents, for example, helix from selecting the entire file contents on every format, even if nothing changed.

vkleen: From looking at that very briefly just now, I think we're actually sending a slightly wrong text edit. I'm suspecting there's an off by one error for a trailing newline somewhere, but that's better left for a follow-up PR.

yannham: IIRC, LSP numbergin starts at 1 for both chars and lines, if that matters. Otherwise, if rust-analyzer is doing it, then it's probably reasonable

jneem: At least in the completion messages, I'm pretty sure they both start at zero...

vkleen: They do indeed start at zero: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range

PR #1611, opened 2023-09-16
Testing Nickel snippets in the manual

### LINKED ISSUE #1154: Check evaluation of snippets in the manual
During our quest to rename the stdlib I hit a pain point with keeping the manual up-to-date. It would have been very helpful if we had some automated way to check that the Nickel snippets in it are valid.

I'm imagining some tooling that extracts code snippets marked as `nickel` from the manual and tried to evaluate them. This could then be part of the CI. There are probably lots of edge cases with incomplete code fragments and intentionally failing programs, so this might get a bit tricky. But maybe it's still worth the effort.

vkleen:
Add a new test suite in `core/tests/manual` which parses the Markdown files in `doc/manual` to extract all Nickel code blocks and checks that they parse or evaluate correctly. For now, REPL sessions in the manual aren't evaluated during testing but only parsed.

To facilitate this, code blocks in the manual are annotated with `nickel`, `nickel-lines`, `nickel-parse`, `nickel-repl` or `nickel-no-check` depending on the parsing or evaluation mode desired.

Closes #1154

### DIFF
--- a/core/tests/manual/main.rs
+++ b/core/tests/manual/main.rs
@@ -16,6 +16,8 @@ enum CodeBlockType {
     /// A code block containing multiple lines, each of which should be
     /// evaluated separately
     Lines,
+    /// A code block that should only be parsed, not evaluated
+    Parse,
     /// A code block containing a REPL session, e.g.
     /// ```nickel repl
     /// > let foo =
@@ -45,6 +47,7 @@ impl CodeBlockType {
             "nickel" => Some(CodeBlockType::Single),
             "nickel-lines" => Some(CodeBlockType::Lines),
             "nickel-repl" => Some(CodeBlockType::Repl),
+            "nickel-parse" => Some(CodeBlockType::Parse),
             _ => None,
         }
     }
@@ -72,6 +75,11 @@ fn check_parse_extended(program: String) {
     test_program::parse_extended(&program).unwrap();
 }
 
+fn check_parse(program: String) {
+    eprintln!("{program}"); // Print the program to stderr to make tracking test failures easier
+    test_program::parse(&program).unwrap();
+}
+
 impl CodeBlock {
     fn check(self) {
         match self.typ {
@@ -81,6 +89,7 @@ impl CodeBlock {
                     check_eval(line.to_owned());
                 }
             }
+            CodeBlockType::Parse => check_parse(self.content),
             CodeBlockType::Repl => check_repl_parts(self.content),
         }
     }
--- a/core/tests/manual/main.rs
+++ b/core/tests/manual/main.rs
@@ -19,7 +19,7 @@ enum CodeBlockType {
     /// A code block that should only be parsed, not evaluated
     Parse,
     /// A code block containing a REPL session, e.g.
-    /// ```nickel repl
+    /// ```nickel #repl
     /// > let foo =
     ///     fun x => x
     ///   in foo 5
@@ -45,9 +45,9 @@ impl CodeBlockType {
     fn from_info(info: impl AsRef<str>) -> Option<CodeBlockType> {
         match info.as_ref() {
             "nickel" => Some(CodeBlockType::Single),
-            "nickel-lines" => Some(CodeBlockType::Lines),
-            "nickel-repl" => Some(CodeBlockType::Repl),
-            "nickel-parse" => Some(CodeBlockType::Parse),
+            "nickel #lines" => Some(CodeBlockType::Lines),
+            "nickel #repl" => Some(CodeBlockType::Repl),
+            "nickel #parse" => Some(CodeBlockType::Parse),
             _ => None,
         }
     }
--- a/core/tests/manual/main.rs
+++ b/core/tests/manual/main.rs
@@ -26,10 +26,10 @@ enum CodeBlockType {
     /// > std.function.id 5
     /// 5
     /// ```
-    /// We interpret lines starting with `> ` and followed by indented lines
-    /// (starting with at least one `' '`) as a single Nickel program and try
-    /// to parse them. For now, we don't check evaluation. REPL inputs must be
-    /// separated by an empty line.
+    /// We interpret lines starting with `> ` directly after an empty line
+    /// and followed by indented lines (starting with at least one `' '`) as
+    /// a single Nickel program and try to parse them. For now, we don't check
+    /// evaluation. REPL inputs must be separated by an empty line.
     Repl,
 }
 
@@ -43,8 +43,8 @@ impl CodeBlockType {
     fn from_info(info: impl AsRef<str>) -> Option<CodeBlockType> {
         match info.as_ref() {
             "nickel" => Some(CodeBlockType::Single),
-            "nickel lines" => Some(CodeBlockType::Lines),
-            "nickel repl" => Some(CodeBlockType::Repl),
+            "nickel-lines" => Some(CodeBlockType::Lines),
+            "nickel-repl" => Some(CodeBlockType::Repl),
             _ => None,
         }
     }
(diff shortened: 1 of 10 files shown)


doc/manual/contracts.md
```diff
@@ -19,13 +19,16 @@ using the `|` operator:
 let x = (1 + 1 | Number) in x
 ```
 
-Contracts can also be attached to identifiers in a definition:
+Contracts can also be attached to identifiers in a definition. In let bindings,
+the following is equivalent to the previous example:
```
jneem, line 23:
```suggestion
Contracts can also be attached to identifiers in a let binding.
The following is equivalent to the previous example:
```

doc/manual/contracts.md
```diff
@@ -100,16 +103,16 @@ In `IsFoo`, we first test if the value is a string, and then if it is equal to
 `"foo"`, in which case the contract succeeds. Otherwise, the contract fails with
 appropriate error messages. Let us try:
 
-```nickel
+```nickel-repl
 nickel> 1 | IsFoo
```
jneem, line 107:
```suggestion
> 1 | IsFoo
```

doc/manual/syntax.md
```diff
@@ -147,13 +147,13 @@ present on all lines of the string is stripped. This way, multiline strings can
...
 
-```nickel
+```nickel-repl
 > m%"
-  This line has no indentation.
-    This line is indented.
-      This line is even more indented.
-  This line has no more indentation.
-"%
+    This line has no indentation.
```
jneem, line 152: I'm a little confused by this example. I guess you had to indent it extra so that the final `"%` will be caught by the parser? But even before this change, the "This line has no indentation" line had indentation...

doc/manual/typing.md
```diff
@@ -290,25 +294,23 @@ let snd : forall a b. a -> b -> b = fun x y => y in
...
-<!-- ``` -->
+Or even nest them:
+
+```nickel
+let higherRankId : forall a. (forall b. b -> b) -> a -> a
+  = fun id x => id x in
+let id : forall a. a -> a
+  = fun x => x in
+higherRankId id 0 : Number
+```
```
jneem, line 305: :tada:

jneem (approved):

doc/manual/syntax.md
vkleen: The multiline string parser strips common indentation. So this line doesn't have any indentation in the final string and that's what this example is supposed to illustrate.

doc/manual/typing.md
```diff
@@ -676,6 +678,13 @@ can be sure that something will eventually be a port number, or will fail with
 the correct error message: when using a contract application.
 
 ```nickel
+# hide-start
+let Port = std.contract.from_predicate (fun value =>
+  std.is_number value
+  && value % 1 == 0
```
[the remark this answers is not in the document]
vkleen: If `value` isn't an integer, `value % 1` returns the fractional part of `value`. I suppose that's in line with writing `x mod 1` for the residue class of `x` in $\mathbb R/\mathbb Z$...

core/tests/manual/main.rs
```diff
@@ -0,0 +1,123 @@
...
+enum CodeBlockType {
+    /// A code block intended to be evaluated in its entirety
+    Single,
+    /// A code block containing multiple lines, each of which should be
+    /// evaluated separately
+    Lines,
+    /// A code block that should only be parsed, not evaluated
+    Parse,
+    /// A code block containing a REPL session, e.g.
+    /// ```nickel repl
```
yannham, line 22: The example doesn't check out with practice: it seems that you are requiring a `-` instead of a space, ie `nickel-repl`.

doc/manual/contracts.md
```diff
@@ -711,34 +717,37 @@ value and continues with the second argument (here, our wrapped `value`).
...
-  "0" | doc "Some information" = true,
-}
-nickel>:query config "0"
+```nickel-repl
+> let config | NumberBoolDict = {
+    "1" | std.FailWith "ooch" = null, # same as our previous "fail"
+    "0" | doc "Some information" = true,
+  }
+
+# Don't parse this in tests hide-line
```
yannham, line 726: What's the purpose of this comment, and where is it handled? I think I missed it somehow... Does it appear in the final manual? Also, I hope we can get rid of it eventually - starting an REPL engine using `ReplImpl` to actually test the output of the REPL examples should be doable.

yannham (commented):
A good step forward!

I wonder if putting a space after `nickel` in the textinfo of fenced code block isn't more idiomatic markdown. In particular, according to the common mark spec, only the first word of the textinfo (up to the first space) is considered for highlighting purpose (see eg https://spec.commonmark.org/0.28/#example-112). We adapted the website, but for example I often read the manual inside NeoVim which has syntax coloring for Nickel snippet, and who knows what markdown rendered people will use to read the manual outside of the website. Would that make sense to do:

- `nickel-repl` => `nickel repl`
- `nickel-parse` => `nickel parse`

It could be something even more specific to indicate that it doesn't concern rendering, such as `nickel @test.parse-only` (this is an absolutely random suggestion, not necessarily a good one)

doc/manual/contracts.md
vkleen: It's a bit convoluted. The code parsing out REPL inputs expects to find a `>` directly following an empty line, so it will skip this particular input because of the comment. The `hide-line` is for gatsby to not display the comment on the website.

core/tests/manual/main.rs
vkleen: Right, I tried what you were suggesting about spaces before. Unfortunately, gatsby doesn't seem to ignore the extra information after `nickel`. In fact, it seems to remove the space after `nickel` and try to look for a Prismjs language `nickelrepl` if given `nickel repl` in the code block. Relying on that behaviour felt a bit too spooky for me, though...

vkleen: On the other hand, relying on it and just continuing to check when updating the website should make syntax highlighting work in saner markdown renderers, I guess. It's easy to change back here, so whichever you think would make more sense long term.

doc/manual/contracts.md
yannham: Ah. I would have suggested to add a comment to explain that, but...that's probably a bad idea :stuck_out_tongue: is the `hide-line` a vanilla Gatsby feature or something we filter out specifically in nickel-lang.org ?

vkleen: The `hide-line` is a feature of the [`gatsby-remark-prismjs` plugin](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-prismjs/README.md#line-hiding).

core/tests/manual/main.rs
yannham: Yeah, and what about using a special character and a space, something like `nickel #repl` or `nickel !repl`? Gatsby would parse that as `nickel#repl` which has almost no chance of conflicting with anything, keeping the alias approach (could even be `nickel -repl`, even if that feels...slightly wrong for some reason), and such that other markdown parser would be ok with it? I tried it into Neovim and some random HTML markdown renderer extension I have and indeed adding a space preserves highlighting, while the dash doesn't. It seems like it doesn't make a difference for Gatsby in one direction or the other, but preserve other renderers

yannham (approved):
Whatever approach you choose for the textinfo string, it's not a blocker.

core/tests/manual/main.rs
vkleen: I like `nickel #...` 👍 I'll make the change here and update the gatsby config on the website.

PR #1615, opened 2023-09-20
Run a REPL session for `nickel #repl` snippets

vkleen:
This is an enhancement of #1611 to improve testing REPL snippets. Instead of just making sure that REPL inputs parse correctly, we're now starting up a REPL session and run each command. This also enables us to compare the claimed output with what the REPL would actually produce.

### DIFF
--- a/core/tests/manual/main.rs
+++ b/core/tests/manual/main.rs
@@ -1,11 +1,18 @@
-use std::{fs::File, io::read_to_string};
+use std::{fs::File, io::read_to_string, iter::once};
 
+use codespan_reporting::term::termcolor::NoColor;
 use comrak::{
     arena_tree::{Node, NodeEdge},
     nodes::{Ast, AstNode, NodeCodeBlock, NodeValue},
     parse_document, ComrakOptions,
 };
+use nickel_lang_core::{
+    error,
+    eval::cache::CacheImpl,
+    repl::{EvalResult, Repl, ReplImpl},
+};
 use nickel_lang_utils::{project_root::project_root, test_program};
+use pretty_assertions::assert_str_eq;
 use test_generator::test_resources;
 use typed_arena::Arena;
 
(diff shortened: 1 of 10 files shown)


core/tests/manual/main.rs
```diff
@@ -53,14 +67,98 @@ impl CodeBlockType {
...
+    assert_str_eq!(
+        actual
+            .as_ref()
+            .lines()
+            .flat_map(|l| once(l.trim_end()).chain(once("\n")))
+            .collect::<String>()
+            .trim_end(),
+        expected.as_ref().trim_end()
+    );
+}
```
Radvendii, line 118: why do we have trailing whitespace somewhere?

core/tests/manual/main.rs
```diff
@@ -53,14 +67,98 @@ impl CodeBlockType {
...
+        MessageExpectation::Full(expected) => assert_str_eq_approx(actual, expected),
+        MessageExpectation::Abridged(prefix) => assert_prefix(actual, prefix),
+    }
+}
+
+fn check_repl(content: String) {
+    let mut repl = ReplImpl::<CacheImpl>::new(std::io::sink());
+    repl.load_stdlib().unwrap();
     for piece in content.split("\n\n") {
         if let Some(piece) = piece.strip_prefix('>') {
```
Radvendii, line 141: what happens if there isn't a leading `>`? Should we silently ignore it or error out?

doc/manual/contracts.md
```diff
@@ -718,6 +729,36 @@ value and continues with the second argument (here, our wrapped `value`).
...
+          fun name value =>
+            let label_with_msg =
+              std.contract.label.with_message "field `%{name}` is not a boolean" label
+            in
+            std.contract.apply Bool label_with_msg value
+        )
+      |> std.seq check_fields
+    else
+      std.contract.blame_with_message "not a record" label
+
```
Radvendii, line 761: Can we have the repls chain together (as if they're one repl) rather than repeating these preambles everywhere?

core/tests/manual/main.rs
vkleen: The error reporting code seems to produce trailing whitespace from time to time. It looks like that's due to `codespan` and not anything that we're doing, so this seems to be the best option to deal with it.

core/tests/manual/main.rs
vkleen: If there isn't a leading `>` this code just silently ignores the `piece`. That's actually pretty useful in the manual. For example, you can make the tests skip a REPL statement, that I haven't bothered implementing; `:query` comes to mind.

doc/manual/contracts.md
vkleen: I think, this is actually the only place in the manual where this would have been useful. It would probably be feasible to add a REPL session id to the code block info string, but I'm not sure it's worth the effort right now.

core/tests/manual/main.rs
Radvendii: Okay, great. Just making sure it was intentional. That also might be worth a comment since I often encounter things like this that are silently ignoring errors.

maybe an empty `else` branch saying `// ignore unknown REPL statement e.g. :query`

core/tests/manual/main.rs
Radvendii: Ah, okay. A comment to that effect might be useful for posterity.

doc/manual/contracts.md
Radvendii: Yeah, fair enough.

Radvendii (approved):
I think the comments would be useful, but nothing essential.

PR #1617, opened 2023-09-21
Assume -> ApplyContract

yannham:
At the very beginning of Nickel, the duality between a type annotation and contract annotation was reflected in their original naming, the assume construct (contract annotation) and the promise construct (a type annotation).

This terminology isn't used anymore, and isn't very telling for new contributors. This PR gets rid of old references to `Assume` and `Promise`, and rename the `Assume` primitive operation to the more boring but more explicit `ApplyContract`.


vkleen (approved):
Is there a reason why we keep with the "functional" enum syntax, e.g. `BinaryOp::ApplyContract()`, for the operators? Maybe this could be a good point to move them all over to not using `()` if we agree that it would look nicer.

yannham:
No good reason, it's just historical. Let's get rid of them in a follow-up PR.

PR #1623, opened 2023-09-21
Convert symbols request away from the linearizer

jneem:
I'm not completely sure about what should count as a symbol. Right now, let bindings and record fields create "symbols" but function arguments do not.

Returning all locally scoped symbols feels a little noisy to me right now. Maybe the best way to improve it would be to make a symbol hierarchy instead of a flat list. LSP supports this, but we aren't doing it yet.

### DIFF
--- a/lsp/lsp-harness/src/lib.rs
+++ b/lsp/lsp-harness/src/lib.rs
@@ -4,8 +4,8 @@ mod output;
 pub use jsonrpc::Server;
 use log::error;
 use lsp_types::{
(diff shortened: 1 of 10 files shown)


vkleen (approved):

yannham:
@jneem : right, I wonder if we should even put the symbols defined within let-binding at all. Usually, in my limited experience, you're rather looking for record fields - although as you say, even there, it's not clear for the LSP which one are "more top-level" than the others, because Nickel doesn't have a notion of top-level symbols. But returning all the locally defined variable does feel noisy in practice.

jneem:
I filed #1627 for future improvements.

PR #1624, opened 2023-09-22
Pin and update Topiary

ErinvanderVeen:
Updates Topiary to the latest commit on `main` and pins that revision to prevent future unwanted updates to Topiary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,8 @@ typed-arena = "2.0.2"
 unicode-segmentation = "1.10.1"
 void = "1"
 
-topiary = { version = "0.2.3", git = "https://github.com/tweag/topiary.git", rev = "refs/heads/main" }
+topiary = { git = "https://github.com/tweag/topiary.git", rev = "8299a04bf83c4a2774cbbff7a036c022efa939b3" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -90,7 +90,8 @@ typed-arena = "2.0.2"
 unicode-segmentation = "1.10.1"
 void = "1"
 
-topiary = { version = "0.2.3", git = "https://github.com/tweag/topiary.git", rev = "refs/heads/main" }
+topiary = { git = "https://github.com/tweag/topiary.git", rev = "8299a04bf83c4a2774cbbff7a036c022efa939b3" }
+topiary-queries = { git = "https://github.com/tweag/topiary.git", rev = "8299a04bf83c4a2774cbbff7a036c022efa939b3", package = "topiary-queries", default-features = false, features = ["nickel"] }
```
ErinvanderVeen, line 94: I'm sure there is a better way to do this, but I don't know it.

vkleen (approved):

Cargo.toml
yannham: Oh, so @vkleen , our version of Topiary wasn't actually pinned. Which is a bit reassuring about the behavior of `cargo install`.

yannham (approved):
Thanks for the fix! We should have pinned Topiary from the beginning.

PR #1626, opened 2023-09-22
More CI speed

Radvendii:
- switch `nickelWasm` to release build so it hits the cache
- build packages separately in `cargoArtifacts` derivation, so that when we build them separately later they hit the cache correctly
  - see https://github.com/ipetkov/crane/issues/396

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -618,9 +618,11 @@
           nickel-lang-cli
           nickel-lang-core
           rustfmt;
-        # An optimizing release build is long: eschew optimizations in checks by
-        # building a dev profile
-        nickelWasm = buildNickelWasm { profile = "dev"; };
+        # There's a tradeoff here: "release" build is in theory longer than
+        # "dev", but it hits the cache on dependencies so in practice it is
+        # shorter. Another option would be to compile a dev dependencies version
+        # of cargoArtifacts. But that almost doubles the cache space.
+        nickelWasm = buildNickelWasm { profile = "release"; };
         inherit vscodeExtension;
         pre-commit = pre-commit-builder { };
       };
--- a/flake.nix
+++ b/flake.nix
@@ -268,7 +268,22 @@
           # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
           cargoArtifacts = craneLib.buildDepsOnly {
             inherit pname src;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -268,7 +268,28 @@
           # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
           cargoArtifacts = craneLib.buildDepsOnly {
             inherit pname src;
-            cargoExtraArgs = "${cargoBuildExtraArgs} --workspace --all-features";
+            cargoExtraArgs = "${cargoBuildExtraArgs} --all-features";
```
yannham, line 271: Should we keep all features here? Or keep it but make sure that the subsequent `buildPackage` also have all-features? From our last discussion I had the impression that what we want is that the features set used to build the packages should be - as much as possible - the same feature set as the one used in `cargoArtifacts`. In fact I think on the CI we should probably always enable all features by default (the best would be to test different feature sets, but if we have to chose one feature set, all-features is probably the one with the maximal covering)

flake.nix
```diff
@@ -268,7 +268,28 @@
...
+            cargoTestCommand = "cargoWorkspace test";
+            cargoCheckCommand = "cargoWorkspace check";
+            preBuild = ''
+              cargoWorkspace() {
+                command=$(shift)
+                for packageDir in $(${pkgs.yq}/bin/tomlq -r '.workspace.members[]' Cargo.toml); do
+                  (
+                    cd $packageDir
+                    pwd
+                    cargoWithProfile $command "$@"
```
yannham, line 288: Where's `cargoWithProfile` defined?

yannham (approved):

flake.nix
Radvendii: somewhere in `craneLib`. it's the default value of `cargo{Build,Test,Check}Command`

flake.nix
Radvendii: It's only an issue if we have upstream dependencies whose feature sets depends on the set of features enabled on our crates. At the moment, we don't have any such situations, but in theory it could come up in the future. My impulse is to deal with it if/when it comes up, but the downside of that is that nobody will notice anything is wrong, CI will just silently start taking a little longer.

PR #1629, opened 2023-09-25
Hover support without the linearizer

jneem:
I think this is the last piece that needs to be converted to the AST-based nls infrastructure, meaning we can get rid of the linearizer.

### DIFF
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -236,7 +236,6 @@ impl<'a> FieldResolver<'a> {
 
                     self.resolve_def_with_path(def)
                 })
-                // FIXME: it isn't finding 'std'. Need to provide an intial env containing std...
                 .unwrap_or_else(|| {
                     log::info!("no def for {id:?}");
                     Default::default()
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -236,7 +236,12 @@ impl<'a> FieldResolver<'a> {
 
                     self.resolve_def_with_path(def)
                 })
-                .unwrap_or_default(),
+                // FIXME: it isn't finding 'std'. Need to provide an intial env containing std...
+                .unwrap_or_else(|| {
+                    log::info!("no def for {id:?}");
+                    Default::default()
+                }),
+            //.unwrap_or_default(),
             Term::ResolvedImport(file_id) => self
                 .server
                 .cache
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -745,12 +745,6 @@ fn env_completion(rt: &RichTerm, server: &Server) -> Vec<CompletionItem> {
     let resolver = FieldResolver::new(server);
     let mut items: Vec<_> = env
         .iter_elems()
-        // The name `internals` is always in scope, but let's not advertise the fact. We do want
-        // to complete `internals` if someone else uses that name, so we only filter out the
-        // `internals` that doesn't have a position -- anything user-defined should have one.
-        .filter(|(_, def_with_path)| {
-            def_with_path.ident.ident != "internals".into() || def_with_path.ident.pos.is_def()
-        })
         .map(|(_, def_with_path)| def_with_path.completion_item())
         .collect();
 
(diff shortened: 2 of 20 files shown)


lsp/nls/src/requests/hover.rs
```diff
@@ -1,68 +1,169 @@
...
+            ret.metadata = metadata;
+        } else if def.path.is_empty() {
+            ret.values.extend(def.value.iter().cloned());
+        }
+    }
+
+    Some(ret)
+}
+
+fn term_hover(rt: &RichTerm, server: &Server) -> Option<HoverData> {
```
yannham, line 83: It's a detail, and probably not for this PR, but I wonder if all those `term_xxx`, `type_xxx`, `contract_xxx` for several `xxx` (finding completion information, getting hover information, and so on) wouldn't be a bit more organized as traits. For example a `trait Hover { .. hover(..); }` that is implemented by `RichTerm`, `Ident` and whatnot.

lsp/nls/src/requests/hover.rs
```diff
@@ -1,68 +1,169 @@
...
 
-    let item = item.unwrap().to_owned();
+        contracts.sort();
+        contracts.dedup();
 
-    debug!("Found item {:?}", item);
+        contents.extend(contracts.into_iter().map(nickel_string));
 
-    let (ty, meta) = linearization.get_type_and_metadata(&item, &server.lin_registry);
+        // Not sure how to do documentation merging yet, so pick the first non-empty one.
```
yannham, line 152: I think it's the current behavior of merging as well. We should probably abstract that into Nickel (maybe define `struct Doc(Rc<String>)` and implement `Combine` on it, but I reckon it's not yet the case)

lsp/nls/src/usage.rs
```diff
@@ -115,11 +111,20 @@ impl UsageLookup {
...
-            .as_opt_ref()
-            .and_then(|span| self.def_table.get(span))
-            .and_then(|env| env.get(&ident.ident))
+        // First try to look up the definition in our symbols table. If that fails,
+        // find the active environment and look up the ident in it.
+        //
+        // We check the symbols table first so that we can retrieve the definition
+        // in the case that the ident is already pointing straight at it. For example, in
+        // `let x = 3`, the environment containing `x` doesn't define `x` but we still
+        // want `def(x)` to return this definition.
```
yannham, line 120: I think I didn't understand this part. Do you have an example where you do a goto definition or hover, and where NOT looking into the symbol table would be an issue?

yannham (approved):

lsp/nls/src/requests/completion.rs
```diff
@@ -745,6 +745,12 @@ fn env_completion(rt: &RichTerm, server: &Server) -> Vec<CompletionItem> {
     let resolver = FieldResolver::new(server);
     let mut items: Vec<_> = env
         .iter_elems()
+        // The name `internals` is always in scope, but let's not advertise the fact. We do want
+        // to complete `internals` if someone else uses that name, so we only filter out the
+        // `internals` that doesn't have a position -- anything user-defined should have one.
+        .filter(|(_, def_with_path)| {
+            def_with_path.ident.ident != "internals".into() || def_with_path.ident.pos.is_def()
```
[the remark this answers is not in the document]
jneem: Got it, thanks!

lsp/nls/src/field_walker.rs
```diff
@@ -236,7 +236,12 @@ impl<'a> FieldResolver<'a> {
 
                     self.resolve_def_with_path(def)
                 })
-                .unwrap_or_default(),
+                // FIXME: it isn't finding 'std'. Need to provide an intial env containing std...
```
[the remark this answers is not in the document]
jneem: Ah sorry, I fixed it but forgot the comment...

lsp/nls/src/usage.rs
jneem: If the input is `let x = 3 in x` and you hover on either `x`, you want to see that `x` is a `Number`. But if you look up the first `x` in the `def_table` it returns the empty (well, apart from `std`) environment because that's the active environment at that location. (Maybe the `def_table` should really be called the `env_table` because that's actually what it's storing: the environment of each term.)

Issue #1484, opened 2023-07-28
Performance improvement ideas: tracking issue


yannham:
Performance is a tricky matter. We want to avoid spending time on implementing changes wishfully improving performance without:

1. A concrete use-case showing that a performance issue is either an actual problem in practice, or is subpar when compared to other languages/tools
2. Principled measurements pointing to functions to be improved, and after implementing a performance-related change, confirming the gain

Still, with time, we've spotted many potential points of congestion and candidates for performance improvements. This issue serves to record those insights, and might be used as a starting point when investigating a performance issue or looking for performance improvements for specific parts of the Nickel pipeline.

This issue isn't concerned with big paradigm shifts such as incremental evaluation or moving towards a bytecode compiler + virtual machine approach. Such topics deserve their own issue/notes/RFC/paper. The ideas here are closely tied to the current implementation, and we want to avoid losing them or ignoring them while they could be relevant.

Feel free to add your own, that I'll try to edit into this description.

## Interpreter

- **Get rid of generated variables**. Currently, we can't directly store a closure (a term and an environment together) into a Nickel AST. One reason is that we only have one AST, and a closure doesn't make any sense at e.g. parsing time, typechecking time and program transformation time. It would be strange to depend on runtime-related definitions for the core AST type. But, during evaluation, we morally need to store closures inside expression (typically, data structures such as arrays and records store closures). The trick we use is to generate a unique variable, bind the closure in the environment, and store this variable in the term instead. This trick leads to a lot of variable generation and variable lookup, which go through the environment, meaning they incur a hashmaps lookup (and, in the worst case, a linear-time lookup, because the environment is more or less a linked list of hashmaps). This is in particular a problem for the share normal form, which, for big terms (large records), might generate a huge load of heading lets binding generated variables, degrading the performance of even just parsing then exporting a Nickel program which is mostly already evaluated (#1428)
[...]

PR #1631, opened 2023-09-26
Implement contract deduplication optimization

### LINKED ISSUE #1622: Investigate degradation of performance for medium-size codebase (5~10 kLoc)
**Describe the bug**
While I can't produce the code here for privacy reason, I've been personally given a Nickel codebase where the export time is very long (for 5~10kLoC, between 10 and 20s on my laptop). There is no reason that a program of this size is that slow. The code looks rather reasonable, and it means that we are at risk of similar performance degradation for any codebase that reaches this size.

Let's investigate, profile and rely on #1484 to see what's happening and bring that time down to something reasonable.

**To Reproduce**
Can't share the offending code here for privacy reason, unfortunately.

**Expected behavior**
That the export time is low (as a starter, say under a second).

```[tasklist]
### Tasks
- [ ] https://github.com/tweag/nickel/issues/1646
- [ ] https://github.com/tweag/nickel/issues/1648
- [ ] https://github.com/tweag/nickel/issues/1647
- [ ] https://github.com/tweag/nickel/issues/1668
```

yannham:
Partially fixes #1622. Related to #1484.

Implement the contract deduplication optimization, which avoid applying many times the same contract to a field value. This optimization requires that contracts are idempotent, or it could change the meaning of programs.

## Content

- Add `push_dedup` method on `RuntimeContract`, which conditionally pushes a contract to an existing list of pending contracts, given that it's not already in the list. Theoretically, this makes the complexity of merging two pending contract lists `size(left)*size(right)`. In practice, we expect pending contracts list to be very small (< 10), and in fact the deduplication optimization typically participate in keeping the size small.
- Add a `combine_dedup` method to combine contract annotation and deduplicate them based on an unsound, "syntax equality" like check. This is to avoid accumulating and printing many copies of the same initial contract when showing the result of evaluation.
- Fix the contract equality check for the eval environment. In particular, the contract equality checker actually needs to access to the initial environment, which comes with a few techcal complications as to how to properly define the signature in the `TermEnvironment` trait.
- Add a unsound mode to contract equality checking, corresponding to comparing them syntactically, used by |
- The previous point required to thread the initial environment into me functions, to make it accessible when deduplicating contracts. In the end, I put the initial environment into the virtual machine, because it looks like it shouldn't change much over the course of any Nickel subcommands - it's always the environment populated with the stdlib - and got rid of a lot of environment threading, a bit of simplification around the `query` subcommand as well.

## Results

On the codebase mentioned in #1622, this got the runtime down by roughly 50%, equivalently across `nickel export` or the normal evaluation.

## Future work

- [ ] Apply the same optimization to lazy array contracts

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -22,8 +22,7 @@
 //! Each such value is added to the initial environment before the evaluation of the program.
 use crate::{
     cache::*,
-    error::{report, ColorOpt, Error, EvalError, IOError, IntoDiagnostics, ParseError},
-    eval,
+    error::{report, ColorOpt, Error, IntoDiagnostics, ParseError, EvalError, IOError},
     eval::{cache::Cache as EvalCache, VirtualMachine},
     identifier::LocIdent,
     label::Label,
@@ -237,7 +236,7 @@ impl<EC: EvalCache> Program<EC> {
 
     /// Retrieve the parsed term and typecheck it, and generate a fresh initial environment. Return
     /// both.
-    fn prepare_eval(&mut self) -> Result<(RichTerm, eval::Environment), Error> {
+    fn prepare_eval(&mut self) -> Result<RichTerm, Error> {
         // If there are no overrides, we avoid the boilerplate of creating an empty record and
         // merging it with the current program
         if self.overrides.is_empty() {
@@ -258,7 +257,7 @@ impl<EC: EvalCache> Program<EC> {
                 .value(Term::ResolvedImport(value_file_id));
         }
 
-        let (t, initial_env) = self.vm.prepare_eval(self.main_id)?;
+        let t = self.vm.prepare_eval(self.main_id)?;
         let built_record = record.build();
         // For now, we can't do much better than using `Label::default`, but this is
         // hazardous. `Label::default` was originally written for tests, and although it
@@ -267,22 +266,21 @@ impl<EC: EvalCache> Program<EC> {
         // generate a dummy file id).
         // We'll have to adapt `Label` and `MergeLabel` to be generated programmatically,
         // without referring to any source position.
-        let wrapper = mk_term::op2(BinaryOp::Merge(Label::default().into()), t, built_record);
-        Ok((wrapper, initial_env))
+        Ok(mk_term::op2(BinaryOp::Merge(Label::default().into()), t, built_record))
     }
 
     /// Parse if necessary, typecheck and then evaluate the program.
     pub fn eval(&mut self) -> Result<RichTerm, Error> {
-        let (t, initial_env) = self.prepare_eval()?;
+        let t = self.prepare_eval()?;
         self.vm.reset();
-        self.vm.eval(t, &initial_env).map_err(|e| e.into())
+        self.vm.eval(t).map_err(|e| e.into())
     }
 
     /// Same as `eval`, but proceeds to a full evaluation.
     pub fn eval_full(&mut self) -> Result<RichTerm, Error> {
-        let (t, initial_env) = self.prepare_eval()?;
+        let t = self.prepare_eval()?;
         self.vm.reset();
-        self.vm.eval_full(t, &initial_env).map_err(|e| e.into())
+        self.vm.eval_full(t).map_err(|e| e.into())
     }
 
     /// Same as `eval`, but proceeds to a full evaluation. Optionally take a set of overrides that
@@ -299,18 +297,18 @@ impl<EC: EvalCache> Program<EC> {
     ///   an import referring to the corresponding isolated value. This stub is finally merged with
     ///   the current program before being evaluated for import.
     pub fn eval_full_for_export(&mut self) -> Result<RichTerm, Error> {
-        let (t, initial_env) = self.prepare_eval()?;
+        let t = self.prepare_eval()?;
         self.vm.reset();
         self.vm
-            .eval_full_for_export(t, &initial_env)
+            .eval_full_for_export(t)
             .map_err(|e| e.into())
     }
 
     /// Same as `eval_full`, but does not substitute all variables.
     pub fn eval_deep(&mut self) -> Result<RichTerm, Error> {
-        let (t, initial_env) = self.prepare_eval()?;
+        let t = self.prepare_eval()?;
         self.vm.reset();
-        self.vm.eval_deep(t, &initial_env).map_err(|e| e.into())
+        self.vm.eval_deep(t).map_err(|e| e.into())
     }
 
     /// Wrapper for [`query`].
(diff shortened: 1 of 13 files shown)


core/src/typecheck/eq.rs
```diff
@@ -190,21 +294,47 @@ impl State {
...
+        VarEq::Direct,
+        &GenericUnifType::<NoEnvironment>::from_type(t1.clone(), &NoEnvironment),
+        &NoEnvironment,
+        &GenericUnifType::<NoEnvironment>::from_type(t2.clone(), &NoEnvironment),
+        &NoEnvironment,
+    )
 }
 
 /// Decide type equality on contracts in their respective environment and given the remaining gas
 /// `gas`.
```
jneem, line 330:
```suggestion
/// in `state`.
```

core/src/term/mod.rs
```diff
@@ -319,6 +321,35 @@ impl RuntimeContract {
     {
         contracts.fold(rt, |acc, ctr| ctr.apply(acc, pos))
     }
+
+    /// Push a pending contract to a vector of contracts if the contract to add isn't already
+    /// present in the vector, according to the notion of contract equality defined in
+    /// [crate::typecheck::eq].
+    pub fn push_dedup(
```
jneem, line 328: I guess this is dwarfed by the cost of evaluating the contracts, but will we ever need to worry about the fact that this will be quadratic if we dedup a lot of (distinct) contracts?

jneem (approved):

core/src/term/mod.rs
Radvendii: We've discussed before that it might be sufficient to just check the last contract applied, since most of the time duplicates come together. Is that indeed how it panned out in the real-world case we tested? 

If it **is** a duplicate, we're likely to see that immediately since we do check the last applied contract first, but if it's not a duplicate we have to iterate through all of them.

core/src/term/mod.rs
```diff
@@ -576,6 +607,34 @@ impl TypeAnnotation {
     pub fn is_empty(&self) -> bool {
         self.typ.is_none() && self.contracts.is_empty()
     }
+
+    /// **Warning**: the contract equality check used in this function behaves like syntactic
+    /// equality, and doesn't take the environment into account. It's unsound for execution (it
+    /// could equate contracts that are actually totally distinct), but we use it only to trim
+    /// accumulated contracts before pretty-printing. Do not use prior to any form of evaluation.
```
Radvendii, line 614: Shouldn't this be sound because all the contracts are being applied in the same environment? There's no way for `foo | Foo | Foo` to sneak in a shadow of for `Foo` in the second contract application without being different syntactically.

If it's not sound, it seems problematic for pretty-printing as well. We've run into issues with the pretty-printer producing output that doesn't get parsed back the same way, right?

core/src/term/mod.rs
Radvendii: Come to think of it, isn't it problematic to dedup in this way? 
```
foo | PosNumber | Number | PosNumber
``` 
is different than 
```
foo | PosNumber | Number
```
e.g. the latter can't get passed to a function that takes a `PosNumber`

core/src/typecheck/eq.rs
```diff
@@ -85,12 +103,27 @@ impl Default for SimpleTermEnvironment {
...
     where
         F: FnOnce(Option<(&RichTerm, &SimpleTermEnvironment)>) -> T,
     {
-        f(self.0.get(&id).map(|(rt, env)| (rt, env)))
+        f(env.0.get(&id).map(|(rt, env)| (rt, env)))
+    }
+}
+
+impl<'a> AsRef<SimpleTermEnvironment> for &'a SimpleTermEnvironment {
+    fn as_ref(&self) -> &SimpleTermEnvironment {
```
Radvendii, line 125:
```suggestion
    fn as_ref(&self) -> &Self {
```

core/src/typecheck/eq.rs
```diff
@@ -105,19 +138,58 @@ impl std::iter::FromIterator<(Ident, (RichTerm, SimpleTermEnvironment))> for Sim
...
 }
 
-/*
-impl<C: Cache> TermEnvironment for eval::Environment {
-    fn get_then<F, T>(&self, id: &Ident, f: F) -> T
+#[derive(Clone, Copy)]
+pub struct EvalEnvsRef<'a> {
+    pub eval_env: &'a eval::Environment,
+    pub initial_env: &'a eval::Environment,
+}
```
Radvendii, line 145: It seems the issue here is that we can't construct a `&EvalEnvsOwned` from two `&Environment`s. Do I have that right? Is this a common pattern in rust?

core/src/term/mod.rs
yannham: This is discussed in the PR description:

> Add push_dedup method on RuntimeContract, which conditionally pushes a contract to an existing list of pending contracts, given that it's not already in the list. Theoretically, this makes the complexity of merging two pending contract lists size(left)*size(right). In practice, we expect pending contracts list to be very small (< 10), and in fact the deduplication optimization typically participate in keeping the size small.

At least, that's the assumption. It could be interesting to add a bit of infrastructure to enable different level of optimizations, such that we can continue to check that the assumption is verified (at least it is on the codebase I'm experimenting on).

@Radvendii : deduplication currently only happens at runtime, for things that are merged together. That is, it doesn't impact typechecking at all (there's no dedup at this pointà), and if you write explicitly a duplicated contract annotation, such as `{foo | Number | Number}`, it **will not** be deduplicated (after all, that's your intent). Deduplication would happen for example when doing `{foo | Number} & {foo | Number} & {foo = 1}`.

yannham: > We've discussed before that it might be sufficient to just check the last contract applied, since most of the time duplicates come together. Is that indeed how it panned out in the real-world case we tested?

I think in the real world case we have, most contracts are in fact singleton, so it probably doesn't change anything. In practice I'm not sure about this assumption of order though, because merge order is left unspecified (who knows, we might decide to reorder merges at some point for other optimizations - it's assumed to be commutative and idempotent), so I'm a bit wary of introducing any effect that depends on the order, unless we have practical clues that doing what we're doing now is actually an issue in practice (in that we exhibit quadratic behavior AND it has a noticeable impact: in practice, I'm yet to see a value with more than 3 distinct contracts applied, linear in 3 elements is not really an issue I suspect)

core/src/typecheck/eq.rs
yannham: Exactly: what I want is not a reference to a tuple but a tuple of references instead. I don't know if it's pattern to define a companion reference type and call it `Ref` like that, but at least it seemed natural.

core/src/program.rs
```diff
@@ -235,9 +234,9 @@ impl<EC: EvalCache> Program<EC> {
             .clone())
     }
 
-    /// Retrieve the parsed term and typecheck it, and generate a fresh initial environment. Return
+    /// Retrieve the parsed term, typecheck it, and generate a fresh initial environment. Return
     /// both.
```
vkleen, line 238:
```suggestion
    /// Retrieve the parsed term, typecheck it, and generate a fresh initial environment.
```

PR #1635, opened 2023-09-27
Track more metrics

vkleen:
Add a facility for recording numeric samples and aggregating them as a metric. Added tracked metrics:
- number of calls to `Thunk::new`
- number of calls to `Thunk::new_rev`
- number of calls to `Environment::insert`
- number of calls to `Environment::get`
- the size of an environment when it is actually cloned in `Environment:clone()`
- the size of each environment layer inspected during `Environment::get`
- the number of environment layers traversed during `Environment::get`

### DIFF
--- a/cli/src/metrics.rs
+++ b/cli/src/metrics.rs
@@ -1,33 +1,23 @@
 use std::{
-    collections::HashMap,
-    hash::BuildHasherDefault,
-    sync::{atomic::Ordering, Arc, PoisonError, RwLock},
+    fmt::Display,
+    sync::{atomic::Ordering, Arc},
 };
 
-use metrics::{
-    atomics::AtomicU64, Counter, Gauge, Histogram, Key, KeyHasher, KeyName, SharedString, Unit,
-};
+use metrics::{Counter, Gauge, Key, KeyName, SharedString, Unit};
+use metrics_util::registry::AtomicStorage;
 
-/// The actual store for recorded metrics and their descriptions. We only need
-/// to take the write lock once when creating a new metric, in all other cases
-/// we only take a read lock. The `metrics` crate implements the necessary
-/// trait [`::metrics::CounterFn`] for `Arc<AtomicU64>` and in any case expects its
-/// counter types to be `Arc`s. For this reason, we use `Arc<AtomicU64>` as the
-/// value type of the counters `HashMap`.
-///
-/// The `metrics` crate expects to be used in a thread safe manner, so follow
-/// suit here.
-#[derive(Default)]
-pub(super) struct Registry {
-    counters: RwLock<HashMap<Key, Arc<AtomicU64>, BuildHasherDefault<KeyHasher>>>,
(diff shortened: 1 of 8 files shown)


cli/src/metrics.rs
```diff
@@ -65,67 +48,117 @@ impl metrics::Recorder for Recorder {
...
+    max: f64,
+    avg: f64,
+    count: usize,
+}
+
+impl BucketStatistics {
+    fn new() -> BucketStatistics {
+        BucketStatistics {
+            min: f64::NAN,
+            max: f64::NAN,
```
yannham, line 76: Any reason to use `NaN` rather than `f64::MIN` and `f64::MAX` respectively? The current code might be ok but I believe `NaN` makes min and max order-sensitive, which is a bit hard to reason about.

core/src/environment.rs
```diff
@@ -42,6 +42,7 @@ impl<K: Hash + Eq, V: PartialEq> Clone for Environment<K, V> {
     fn clone(&self) -> Self {
         increment!("Environment::clone");
         if !self.current.is_empty() && !self.was_cloned() {
+            sample!("Environment.size_at_clone", self.current.len() as f64);
```
yannham, line 45:
```suggestion
            sample!("Environment.curr_layer_size_at_clone", self.current.len() as f64);
```

core/src/environment.rs
```diff
@@ -81,7 +83,16 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
...
     pub fn get(&self, key: &K) -> Option<&V> {
-        self.iter_layers().find_map(|hmap| hmap.get(key))
+        increment!("Environment::get");
+
+        let mut layer_count = 0;
+        let r = self.iter_layers().find_map(|hmap| {
+            sample!("Environment.hashmap_size_get", hmap.len() as f64);
+            layer_count += 1;
+            hmap.get(key)
+        });
```
yannham, line 93: I don't know if that's objectively better, but if you despise `mut`, here is a variant:

```suggestion
        let (layer_count, r) = self.iter_layers().enumerate().find_map(|(count, hmap)| {
            sample!("Environment.hashmap_size_get", hmap.len() as f64);
            Some((count, hmap.get(key)?))
        });
```

yannham (approved):

core/src/environment.rs
vkleen: I was really looking for something where I'm confident that the compiler can eliminate the dead store with minimal analysis in case `metrics` are disabled. Apparently hiding expressions behind `#[cfg(feature = "metrics")]` is experimental 😞

cli/src/metrics.rs
vkleen: `NaN` doesn't make them order-sensitive: "If one of the arguments is NaN, then the other argument is returned." That seems to be exactly the semantics we want. That's in fact different from the nightly `minimum` and `maximum` functions, which propagate NaNs. Also, I rather like this code returning Nan when no samples are present; even though that can't actually happen in our setup.

PR #1636, opened 2023-09-27
Sign executables after patching revision on Darwin

YorikSar:
https://github.com/tweag/nickel/pull/1580 introduced patching of resulting binary in a separate derivation that ends up with bad (or missing) code signature which prevents it from running:

```
% nix run
zsh: killed     nix run
```

and Console.app revealing the reason:

```
Exception Type:        EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
Exception Codes:       UNKNOWN_0x32 at 0x00000001054e4199
Exception Codes:       0x0000000000000032, 0x00000001054e4199
```

This change uses `autoSignDarwinBinariesHook` to sign the result that seems to fix the problem.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -234,9 +234,12 @@
         pname = pkg.pname + "-rev-fixup";
         inherit (pkg) version meta;
         src = pkg;
-        buildInputs = [ pkgs.bbe ];
+        buildInputs = [ pkgs.bbe ]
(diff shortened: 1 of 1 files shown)


vkleen (approved):
Looks like a good fix, thanks! @yannham is that worth making a 1.2.2 release for?

yannham (approved):

PR #1637, opened 2023-09-27
build lalrpop in a separate derivation

Radvendii:
lalrpop takes a significant amount of the build time of nickel-lang-core (1:30 / 4:30 on my machine). It's even worse than that in CI because pretty much every package depends on nickel-lang-core and rebuilds it separately.

The good news is that the lalrpop file doesn't change very often, so most of the time we can cache it. I made it into its own package that gets used as the cargoArtifacts of all the downstream packages.

### DIFF
--- a/core/src/typ.rs
+++ b/core/src/typ.rs
@@ -126,8 +126,8 @@ pub enum EnumRowsF<ERows> {
 /// type variables are introduced, that is, on forall quantifiers.
 #[derive(Clone, PartialEq, Eq, Debug, Default)]
 pub enum VarKind {
-    Type,
     #[default]
+    Type,
     EnumRows,
     /// `excluded` keeps track of which rows appear somewhere alongside the tail, and therefore
     /// cannot appear in the tail. For instance `forall r. { ; r } -> { x : Number ; r }` assumes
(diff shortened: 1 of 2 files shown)


Radvendii:
Remaining problem:
- [ ] we stick the built `cargoArtifacts` into the `devShell`, so right now lalrpop has to run in order to get a `devShell`, which we don't want. I'm not sure why we have `cargoArtifacts` in `devShell` to begin with. Does it help with anything? We rebuild them anyways when we run `cargo build`.

vkleen:
I think the intent of `inputsFrom` in `devShell` is that all inputs (i.e. C libraries and rust toolchain) needed for the build are included. Maybe it would be good enough to use only `inputsFrom = [ (mkCraneArtifacts {}).cargoArtifacts ]` to get all the necessary things?

vkleen (approved):

yannham:
Honestly I'm to not sure to know why it is the case. It might be make use of the cache, and have the pre-compiled dependencies available already?

vkleen:
It really doesn't. In the nix devshell, cargo still insists on building all rust dependencies on its own.

yannham (approved):

PR #1658, opened 2023-09-29
Nuke the linearizer

jneem:
A little bit of refactoring/renaming, and a lot of deletion.


jneem:
Good point, I'd renamed the linearizer stuff in nls but forgot about nickel_lang_core. I'll do that in a follow-up.

PR #1663, opened 2023-10-02
Remove/rename the Linearizer trait

jneem:
Renames Linearizer to TypeVisitor and simplifies the interface. (The interface may even be too simple now, but it's what `nls` currently wants.)


yannham (approved):
Nitpick: it's a bit confusing to call something `TypeVisitor` while it's actually visiting terms (because XXXVisitor is usually used for the visitor trait of XXX). Possible suggestions:

- just call it `TermVisitor`
- call it `TypecheckVisitor` or something like that where the first word isn't a datastructure which could have a visitor trait

jneem:
I went with `TypecheckVisitor` because I thought that `TermVisitor` sounds like what `traverse_ref` does...

PR #1665, opened 2023-10-03
Add goto support for pattern bindings

jneem:
Adds support for going to the "definition" of bindings in a let pattern. For example, in

```nickel
let { foo } = bar in ...
```

a "goto definition" request for `foo` will jump to wherever the "foo" field of `bar` is defined.

### DIFF
--- a/lsp/nls/src/requests/goto.rs
+++ b/lsp/nls/src/requests/goto.rs
@@ -8,16 +8,16 @@ use crate::{
     server::Server,
 };
 
-fn get_defs(term: &RichTerm, server: &Server) -> Vec<LocIdent> {
-    match term.as_ref() {
-        Term::Var(id) => {
-            if let Some(loc) = server.analysis.get_def(&(*id).into()).map(|def| def.ident) {
-                vec![loc]
-            } else {
-                vec![]
-            }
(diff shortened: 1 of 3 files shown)


lsp/nls/src/requests/goto.rs
```diff
@@ -8,25 +8,44 @@ use crate::{
...
 
-fn get_defs(term: &RichTerm, server: &Server) -> Vec<LocIdent> {
-    match term.as_ref() {
-        Term::Var(id) => {
-            if let Some(loc) = server.analysis.get_def(&(*id).into()).map(|def| def.ident) {
-                vec![loc]
-            } else {
-                vec![]
-            }
+fn get_defs(term: &RichTerm, ident: Option<LocIdent>, server: &Server) -> Option<Vec<LocIdent>> {
```
vkleen, line 11: Is there a semantic difference between `None` and `Some(vec![])` here?

vkleen (approved):

lsp/nls/src/requests/goto.rs
jneem: No, but changing it to `Option<Vec>` allows `?` for early return

PR #1670, opened 2023-10-09
Add cousin search to goto and hover

jneem:
When encountering trees of merges and annotations in the AST, we've been doing some analysis to include "cousins" into the local scope when generating completions. This PR extends this analysis to hover and goto definition/references. For example, hovering over the first `foo` in the following example now correctly brings up the doc metadata.

```nickel
{
   inner = { foo = 1 }
} | { inner = { foo | doc "hello" } }
```

### DIFF
--- a/lsp/nls/src/analysis.rs
+++ b/lsp/nls/src/analysis.rs
@@ -9,7 +9,7 @@ use nickel_lang_core::{
 };
 
 use crate::{
-    field_walker::DefWithPath,
+    field_walker::Def,
     identifier::LocIdent,
     position::PositionLookup,
     term::RichTermPtr,
@@ -139,6 +139,8 @@ impl<'a> ParentChainIter<'a> {
         let is_fieldy_term = |rt: &RichTerm| {
             matches!(
                 rt.as_ref(),
+                // There is also NAryOp::MergeContract, but only at eval time so we don't
+                // expect it.
                 Term::Op2(BinaryOp::Merge(_), _, _)
                     | Term::Annotated(_, _)
                     | Term::RecRecord(..)
@@ -249,7 +251,7 @@ impl AnalysisRegistry {
         self.analysis.remove(&file_id);
     }
 
-    pub fn get_def(&self, ident: &LocIdent) -> Option<&DefWithPath> {
+    pub fn get_def(&self, ident: &LocIdent) -> Option<&Def> {
         let file = ident.pos.as_opt_ref()?.src_id;
         self.analysis.get(&file)?.usage_lookup.def(ident)
     }
--- a/lsp/nls/src/analysis.rs
+++ b/lsp/nls/src/analysis.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use codespan::FileId;
 use nickel_lang_core::{
     identifier::Ident,
-    term::{RichTerm, Term, Traverse, TraverseControl},
+    term::{BinaryOp, RichTerm, Term, Traverse, TraverseControl},
     typ::{Type, TypeF},
     typecheck::{reporting::NameReg, TypeTables, TypecheckVisitor, UnifType},
 };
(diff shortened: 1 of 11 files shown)


lsp/nls/src/analysis.rs
```diff
@@ -134,6 +134,42 @@ impl<'a> ParentChainIter<'a> {
         }
     }
 
+    /// Like `next`, but skips over everything except for merges, annotations, and records.
+    pub fn next_merge(&mut self) -> Option<RichTerm> {
+        let is_fieldy_term = |rt: &RichTerm| {
+            matches!(
+                rt.as_ref(),
+                Term::Op2(BinaryOp::Merge(_), _, _)
```
yannham, line 142: It's probably not important for the LSP, but FYI theoretically there is also the `MergeContract` operator in `NAryOp`, which is the variant of merge used for contract application. That being said, it's only generated at runtime from `ApplyContract`, which themselves come from annotations, which you handle.

lsp/nls/src/field_walker.rs
```diff
@@ -128,6 +141,11 @@ pub struct DefWithPath {
     pub value: Option<RichTerm>,
     /// The path within the value that this binding refers to.
     pub path: Vec<Ident>,
+    /// If this definition came from a record field, this is the containing
+    /// record. Note that having a parent record is mutually exclusive with
+    /// having a non-empty path, because non-empty paths only come from let
+    /// pattern bindings.
```
yannham, line 147: Would that make sense to encode this invariant as a type, instead of a comment (a simple enum with maybe some helper functions)?

lsp/nls/src/field_walker.rs
```diff
@@ -200,6 +227,28 @@ impl<'a> FieldResolver<'a> {
...
 
+    pub fn get_cousin_defs(&self, def: &DefWithPath) -> Vec<(LocIdent, Field)> {
+        let mut ret = Vec::new();
+        if let Some(parent) = &def.parent_record {
+            if let Some(mut ancestors) = self.server.analysis.get_parent_chain(parent) {
+                while let Some(ancestor) = ancestors.next_merge() {
+                    // We're traversing up the tree starting at the parent, so this is the
+                    // path to the parent (not the original def).
+                    if let Some(parent_path) = ancestors.path() {
+                        let uncles = self.resolve_term_path(&ancestor, parent_path.iter().copied());
```
yannham, line 238: So the uncles are the other leaf of the merge expression enclosing the current term? I think those method could benefit from being described at a high level in the documentation (even more if they're public API, despite "public API" being a bit less meaningful for NLS)

lsp/nls/src/requests/goto.rs
```diff
@@ -100,15 +105,36 @@ pub fn handle_references(
...
+                let resolver = FieldResolver::new(server);
+                let cousins = resolver.get_cousin_defs(&def);
+                def_locs.extend(cousins.into_iter().map(|(loc, _)| loc))
+            }
+        }
+    }
 
+    // TODO: This usage map is based only on static scoping, and not on our "extended"
+    // scopes that we build up dynamically based on merges. Improving this probably
+    // requires building the extended scopes at static analysis time.
```
yannham, line 137: What exactly do you mean by that? We don't have dynamic scoping in Nickel, so you can't use a field that isn't part of the record, even if it will be brought by merging later. Or are you talking about context completion, e.g. in the case `{foo | String} & { [.] }` you would like to provide completion for `foo`?

yannham (approved):

lsp/nls/src/analysis.rs
jneem: Thanks for the info! I guess for now I'll just make a note of this in a comment.

lsp/nls/src/requests/goto.rs
jneem: For context completion this already works, but one example that doesn't currently work is a "goto references" request for the first `foo` in 

```
{ foo = 1 } & { bar = foo }
```

jneem: Sorry, that's what I get for trying to minimize an example without actually checking it :sweat_smile:. But yes, currently we go to only the definition on the RHS whereas I'd like to go to both definitions (GotoDefinition can have multiple responses). Also, when asking for References on the LHS foo, I'd like to respond with the RHS foo.

yannham: Ah, I didn't even know goto definition could show multiple locations (and haven't noticed on Nickel). Now it makes sense, thanks!

Issue #285, opened 2021-02-02
Cast elision inside statically typed code


yannham:
**Is your feature request related to a problem? Please describe.**
Contracts are nice, but they do incur some runtime cost. There are probably many optimization opportunities, one being the elision of unneeded checks. Currently, every type annotation gives rise to a contract at run-time:
```
let f : Num -> Num =
  let id : forall a. a -> a = fun x => x in
  fun x => (id x) + (id 5)
```
This example is contrived but corresponds to a real use-case: using a type annotation inside a statically context, either to introduce a polymorphic type, to debug a type error, or just as a documenting annotation for a subterm.

In that case, the contract should be useless, since as the annotated sub-term is well-typed, it wouldn't elicit positive blame, and since the context is well-typed, it shouldn't elicit negative blame, assuming type soundness and blame safety.

**Describe the solution you'd like**
Any annotation in a statically typed block would be erased at run-time. This could be done at program transformation time, since typechecking occurs before, rewriting such `Promise(ty, term)` as `term` directly. If the typing information is still needed later, e.g. for the `query` subcommand, we can introduce a new node in the AST, which would be the same as a promise during typechecking but would be transparent at run-time. We could also store typing information as meta-values, with more control over which one give rise to a contract or not.

aspiwack:
The contract is useless _only if_  `id` is never used in a dynamically typed context.

It's actually a fairly non-trivial transformation, which I've been considering as out of scope for the near future.

yannham:
> The contract is useless _only if_ `id` is never used in a dynamically typed context.

I am not sure. Since the enclosing block `f` is also statically typed, either it consumes the function internally (and, if well-typed, should not elicit blame), or the function escapes the block, but then it must appear in the codomain of `f` (say `... -> (forall a. a -> a)`), and will thus be guarded by the outer contract anyway.

This example is particular  because `id` has a polymorphic type that we happen to not infer, but another intuition in this direction is that if we take the same example with a monomorphic type `id: Num -> Num`, or imagine our system but with HM type inference, then the annotation is entirely superfluous and adding it or not shouldn't change the runtime behavior of the program.

> It's actually a fairly non-trivial transformation, which I've been considering as out of scope for the near future.

I agree it's not urgent, and that it needs more justification than "it seems to work": probably a description of the type system together with a proof of the soundness of the transformation under reasonable hypotheses. I just wanted to write down the idea first.

aspiwack:
[...]

PR #1671, opened 2023-10-09
Contract elision for static types

yannham:
## Context

As part of #1622, metrics show that the base used for benchmarks is performing an awful lot of array contract applications and their derived subcontracts, in particular polymorphic contracts. This is probably due to the usage of stdlib functions, which are all typed with a polymorphic type.

Indeed, consider `std.record.map`, whose type is `forall a b. (a -> b) -> Array a -> Array b`. If we apply it to a 5000 elements array, this will cause the execution of 20 000 small contracts: one to seal all elements of the array argument with `a`, one to unseal the same elements when `f` is applied (corresponding to `_a_ -> b`), one sealing the result of the application to `b` (the `b` in `a -> _b_`), and finally one that will map the unsealing key over the resulting array.

What's sad is that the function is statically typechecked. Thus, under the assumption that the type system is reasonably behaved (blame safety), the polymorphic contract _can never blame_. In the same spirit, it's not useful to check anything on the return value of the function: the typechecker already proves that it must be `Array b`.

I discussed contract elision rules in #285, which is a bit different, but related. This PR follows the same idea overall: use the guarantees of the static type system to elide some of the dynamic checks.

## Content

This PR applies the following rewriting when generating the contract coming from a static type annotation:

- `forall` in positive position are removed, and their bound variable is substituted for `Dyn`. That is, `forall a b. (a -> b) -> Array a -> Array b` is rewritten to `(Dyn -> Dyn) -> Array Dyn -> Array Dyn`.
- first-order contracts (anything but a function type) in positive position are substituted for `Dyn` as well. Indeed, according to blame safety, a well-typed function can never be blamed, thus positive occurrences of contracts can never fail. For map, we get after this second transformation `(Dyn -> Dyn) -> Array Dyn -> Dyn`

Those optimizations are coupled with the following specializations, which aren't restricted to static type annotations:

- `Dyn -> Dyn` is specialied to `func_dyn`, `Dyn -> T` to `func_dom T`, etc. That is, the general function contract is specialized when the domain, the codomain or both are `Dyn`, to eschew useless checks.
- Similarly, `Array Dyn` is specialized to a constant-time contract `$array_dyn` which just checks that its argument is an array (indeed, `$array $dyn` maps `$dyn` and is thus linear time - not when applied, thanks to lazy array contracts, but still when the array is finally forced)
- Similarly, `{_ : Dyn}` and `{_ | Dyn}` are specialized to `$dict_dyn`.

Those two changes combined give even better results: once the type has been rewritten, many specialized version of builtin contracts can be used. On our `map` example, the final contract becomes `$func $func_dyn ($func_dom $array_dyn)`. This contract only checks that the first argument to `map` is a function and the second is an array, all in constant time. From `4n` contract checks, we are now consistently applying 3 checks (we still check that `map` is a function, which is useless - but it's not very important and can be improved upon later).

## Result

Tested on the codebase of #1622, this gave a consistent 50% reduction in running time, which is encouraging. I will test it on kav (https://github.com/tweag/nickel/discussions/1667) as well.

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -703,10 +703,21 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                 // avoiding repeated contract application. Annotations could then be a good way of
                 // remembering which contracts have been applied to a value.
                 Term::Annotated(annot, inner) => {
-                    let contracts = annot.all_contracts()?;
+                    // We apply the contract coming from the static type annotation separately as
+                    // it is optimized.
+                    let static_contrat = annot.static_contract();
+                    let contracts = annot.pending_contracts()?;
                     let pos = inner.pos;
+                    let inner = inner.clone();
+
+                    let inner_with_static = if let Some(static_ctr) = static_contrat {
+                        static_ctr?.apply(inner, pos)
+                    } else {
+                        inner
+                    };
+
                     let inner_with_ctr =
-                        RuntimeContract::apply_all(inner.clone(), contracts.into_iter(), pos);
+                        RuntimeContract::apply_all(inner_with_static, contracts.into_iter(), pos);
 
                     Closure {
                         body: inner_with_ctr,
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -705,12 +705,12 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                 Term::Annotated(annot, inner) => {
                     // We apply the contract coming from the static type annotation separately as
                     // it is optimized.
-                    let static_contrat = annot.static_contract();
+                    let static_contract = annot.static_contract();
                     let contracts = annot.pending_contracts()?;
                     let pos = inner.pos;
                     let inner = inner.clone();
 
-                    let inner_with_static = if let Some(static_ctr) = static_contrat {
+                    let inner_with_static = if let Some(static_ctr) = static_contract {
                         static_ctr?.apply(inner, pos)
                     } else {
                         inner
--- a/core/src/typ.rs
+++ b/core/src/typ.rs
@@ -929,17 +929,18 @@ impl Type {
 
     /// Static typing guarantees make some of the contract checks useless, assuming that blame
     /// safety holds. This function simplifies `self` for contract generation, assuming it is part
-    /// of a static type annotation, for contract generation by removing some of these useless
-    /// checks.
+    /// of a static type annotation, by eliding some of these useless subcontracts.
     ///
     /// # Simplifications
     ///
     /// - `forall`s in positive positions are removed, and the corresponding type variable is
-    /// turned to a `Dyn` contract, with the hope of making [Self::contract()] to generate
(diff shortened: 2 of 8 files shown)


core/src/eval/mod.rs
```diff
@@ -703,10 +703,21 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                 // avoiding repeated contract application. Annotations could then be a good way of
                 // remembering which contracts have been applied to a value.
                 Term::Annotated(annot, inner) => {
-                    let contracts = annot.all_contracts()?;
+                    // We apply the contract coming from the static type annotation separately as
+                    // it is optimized.
+                    let static_contrat = annot.static_contract();
```
jneem, line 708:
```suggestion
                    let static_contract = annot.static_contract();
```

core/src/eval/mod.rs
```diff
@@ -703,10 +703,21 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                    let contracts = annot.pending_contracts()?;
                     let pos = inner.pos;
+                    let inner = inner.clone();
+
+                    let inner_with_static = if let Some(static_ctr) = static_contrat {
```
jneem, line 713:
```suggestion
                    let inner_with_static = if let Some(static_ctr) = static_contract {
```

core/src/typ.rs
```diff
@@ -927,6 +927,99 @@ impl Type {
...
+                    Box::new(optimize(*dom, vars_elide, polarity.flip())),
+                    Box::new(optimize(*codom, vars_elide, polarity)),
+                ),
+                // TODO: don't optimize only VarKind::Type
+                TypeF::Forall {
+                    var,
+                    var_kind: VarKind::Type,
+                    body,
+                } if polarity == Polarity::Positive => {
+                    let mut env = vars_elide.clone();
```
jneem, line 969: This is technically quadratic, right? I guess we don't expect the foralls to be deeply nested, though...

core/src/typ.rs
```diff
@@ -966,10 +1059,24 @@ impl Type {
             TypeF::Number => internals::num(),
             TypeF::Bool => internals::bool(),
             TypeF::String => internals::string(),
-            //TODO: optimization: have a specialized contract for `Array Dyn`, to avoid mapping an
-            //always successful contract on each element.
+            // Array Dyn is specialied to array_dyn, which is constant time
```
jneem, line 1062:
```suggestion
            // Array Dyn is specialized to array_dyn, which is constant time
```

core/stdlib/std.ncl
```diff
@@ -1971,7 +1971,7 @@
         "%
       = fun field content r =>
         let r =
-          if %has_field% field r then
+          if %has_field_all% field r then
```
jneem: Did this sneak in from the other pr?

jneem (approved):

core/stdlib/std.ncl
yannham: Yes, and it's obsolete. I'll revert this change, thanks for the catch.

yannham: Ah, it's outdated in fact, I did that already. Resolving this conversation

PR #1672, opened 2023-10-09
First prototype of evaluation

### LINKED ISSUE #1639: Evaluate the whole term on-demand through code actions
As a first simple step toward #1586, we could add a code action to evaluate the current document from the LSP and report contract errors (and in fact other errors) in-code.

As opposed to the end goal of #1586, this would require user action, wouldn't handle partial configuration (although we could maybe make it so) and can only report one error at a time, but it's cheap to implement and already better than having to switch to the command line to run the program and switch back to the editor to find the offending line.

[Edit: @yannham]

jneem:
This is an initial (and pretty dumb) version of evaluation: there's a code action that evaluates the whole file (from scratch every time you ask for it) and presents the diagnostics.

Fixes #1639

### DIFF
--- a/lsp/nls/src/command.rs
+++ b/lsp/nls/src/command.rs
@@ -27,7 +27,7 @@ pub fn handle_command(
 
 fn eval(server: &mut Server, uri: &Url) -> Result<(), Error> {
     if let Some(file_id) = server.cache.file_id(uri)? {
-        // TODO: avoid cloning the cache. Maybe we can have a VM with a &mut ImportResolver?
+        // TODO: avoid cloning the cache. Maybe we can have a VM with a &mut Cache?
         let mut vm = VirtualMachine::<_, CacheImpl>::new(server.cache.clone(), std::io::stderr());
         let rt = vm.prepare_eval(file_id)?;
         if let Err(e) = vm.eval_full(rt) {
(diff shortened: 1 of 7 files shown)


lsp/nls/src/command.rs
```diff
@@ -0,0 +1,39 @@
...
+            eval(server, &doc.uri)?;
+            Ok(())
+        }
+        _ => Err(Error::CommandNotFound(params.command).into()),
+    }
+}
+
+fn eval(server: &mut Server, uri: &Url) -> Result<(), Error> {
+    if let Some(file_id) = server.cache.file_id(uri)? {
+        // TODO: avoid cloning the cache. Maybe we can have a VM with a &mut ImportResolver?
```
yannham, line 30: Hmm, this would be annoying because we'd have to add a lifetime parameter to `VirtualMachine`. I wonder if you can't just implement `ImportResolver` for `&'a mut R` when `R: ImportResolver`, thus passing a mutable reference instead? Or just wrap it as a dedicated type `ImportResolverRef<'a, R: ImportResolver>(&'a mut ImportResolver)` and implement the trait. After all `VirtualMachine` is generic over the resolver.

yannham (approved):

lsp/nls/src/command.rs
jneem: Oh, that's a good point. I'll give it a quick try, and if it works I'll add it to this PR

jneem: Implementing `ImportResolver` for `&'a mut R` worked nicely, but it didn't solve the issue because `prepare_eval` is only defined when the import resolver parameter is `ImportCache`. I'll just leave the TODO for now...

PR #1677, opened 2023-10-11
Make rustfmt work in match_sharedterm!

Radvendii:
macros can have syntax that rustfmt doens't know how to deal with, so it only triggers on macros that
1. are called with `()` or `[]` delimiters (not `{}`)
2. contain only valid rust syntax

our `match_sharedterm!` failed on both those counts, so this commit fixes that. we move the `else` clause back inside the match statement as a wildcard arm, and use the `match` keyword so that rustfmt understands how we want the code formatted

TODO: 
- [x] ~~now that we are using the `match` keyword, the macro name `match_sharedterm` seems redundant. I couldn't come up with a better name, but am open to suggestions.~~
- [x] ~~I couldn't figure out a way either to make the `()` around the expression being matched optional, or to trigger a nice error message in that case. This seems bad to me, because it's an easy mistake to make (this is the only way I could see that the syntax differs from a normal `match`. I think this might be unavoidable, but if anyone else can find a solution that would be nice.~~

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -407,25 +407,23 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     // a definition, we complete the error with the additional information of where
                     // it was accessed:
                     let Closure { body, env } = self.cache.get(idx);
-                    let body = match_sharedterm! {body.term, with {
+                    let body = match_sharedterm!(match (body.term) {
+                        Term::RuntimeError(EvalError::MissingFieldDef {
+                            id,
+                            metadata,
+                            pos_record,
+                            pos_access: TermPos::None,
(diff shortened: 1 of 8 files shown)


yannham (approved):
We can think of a better solution for the parenthesis, but honestly, this already looks ten times better than the previous syntax.

As for the name, I don't have a great inspiration, but maybe something like `take_if_needed` or `consume_if_needed` would do the trick? Eventually reader can still refer to the macro's documentation, so as long as it's good enough...

Radvendii:
I feel weird about a name that implies broader applicability, because this still only works in the specific case of matching on a sharedterm.

Radvendii:
We can figure out the name later; for now we should merge the syntax change

PR #1678, opened 2023-10-12
Fix array closurization just forgetting common contracts

yannham:
As the title says. The logic of the previous PR introducing array contracts dedup (#1674) was wrong, and just dropped a contract if it was common to two operands of array concatenation.

Additionally, the index calculation was wrong, as it was performed on the array _after having filtered `None` elements out_, but was then used to perform the deletion in the original array, which is incorrect.

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -1851,27 +1851,36 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                                             initial_env: &self.initial_env,
                                         };
 
+                                        // We check if there is a remaining contract in
+                                        // `ctrs_right_sieve` which matches `ctr`: in this case,
+                                        // `twin_index` will hold its index.
                                         let twin_index = ctrs_right_sieve
                                             .iter()
-                                            // Keep track of the index for later deletion
-                                            .enumerate()
-                                            // Filter out contracts already deduplicated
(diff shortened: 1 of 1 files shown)


core/src/eval/operation.rs
```diff
@@ -1853,18 +1853,27 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
 
                                         let twin_index = ctrs_right_sieve
                                             .iter()
-                                            .filter_map(|ctr| ctr.as_ref())
-                                            .position(|other_ctr| {
+                                            // Keep track of the index for later deletion
+                                            .enumerate()
+                                            // Filter out contracts already deduplicated
+                                            .filter_map(|(i, other_ctr)| Some((i, other_ctr.as_ref()?)))
+                                            .find_map(|(i, other_ctr)| {
```
jneem, line 1860: Would it be simpler to have `ctrs_right_sieve.iter().position(|maybe_other_ctr| maybe_other_ctr.map_or(false, |other_ctr| /* check contract equality */))`. The point being to fix the indexing by checking the `Some` inside the predicate instead of filtering it out.

yannham: It's still quite a bit of nesting, but I agree it's using less iterators functions in total, so let's do this :+1: implemented in the latest HEAD

jneem (approved):

PR #1689, opened 2023-10-19
Fix invalid example code in doc of blame_with_message

bgni:
Fix wrong method name: "std.contract.blame_with_message_message"
should be: "std.contract.blame_with_message"


yannham (approved):
Ah, good catch! Thanks for the fix. We already automatically check the examples (well, in practice, code blocks) in the manual. Maybe we should do the same for the stdlib as well :man_shrugging:

PR #1692, opened 2023-10-19
Fix doc, example code for pipe lacks prefix

bgni:
The example code for "pipe" doesn't compile because the method "pipe" lacks prefix. Fixed by changing "pipe" to "std.function.pipe"

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1589,9 +1589,9 @@
         # Examples
 
         ```nickel
-        pipe 2 [ (+) 2, (+) 3 ]
+        std.function.pipe 2 [ (+) 2, (+) 3 ]
           => 7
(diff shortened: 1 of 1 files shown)


yannham (approved):
I think we use this style in several place in the documentation of the the stdlib, assuming that the prefix is obvious from the context. That being said, I agree that having examples that can be run by copy pasting without having to do any additional guessing or patching is probably a better idea. Let's merge this, but ideally, we should probably make a pass on the whole stdlib to make it consistent

bgni:
I agree with all of that, for someone new to nickel (like me) it's very valuable to have at least one example for each method that is easy to understand and is complete, so you don't have to spend mental resources on figuring out how you can make it work without the prefix (maybe you can do that, I actually don't know).

The ideal in my opinion would be having one or more links to a playground with a executable example of using the method.

PR #1696, opened 2023-10-23
Improve hover output for let patterns

### LINKED ISSUE #1536: LSP hover help is wrong when pattern bindings are involved.
Take the input `(let { x : Number } = { x = 1 } in x) : _` and hover over the first or last `x`. The LS gives the type hint `{x: Number}` but it should just give the type `Number`.

Also, if you remove the `: _` type annotation at the end, the LS says the first `x` is `Dyn` despite the `: Number` annotation.

jneem:
This makes the LSP aware of the types of let pattern bindings (when the typechecker knows them, at least).

I'll say that this fixes #1536 even though there's a bigger question about more aggressive type inference, because (1) the original confusion about whether the bound variable refers to the pattern or the thing inside it is fixed (not by this PR), and (2) we're now getting the right type in the fully-typechecked case.

### DIFF
--- a/lsp/nls/src/requests/hover.rs
+++ b/lsp/nls/src/requests/hover.rs
@@ -26,9 +26,9 @@ struct HoverData {
 }
 
 impl Combine for HoverData {
-    fn combine(mut left: Self, right: Self) -> Self {
-        left.values.extend_from_slice(&right.values);
-        left.metadata.extend_from_slice(&right.metadata);
+    fn combine(mut left: Self, mut right: Self) -> Self {
+        left.values.append(&mut right.values);
+        left.metadata.append(&mut right.metadata);
         left.ty = left.ty.or(right.ty);
         left.span = left.span.or(right.span);
         left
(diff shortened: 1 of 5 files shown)


lsp/nls/src/requests/hover.rs
```diff
@@ -16,14 +17,24 @@ use crate::{
...
     metadata: Vec<FieldMetadata>,
-    span: RawSpan,
+    span: Option<RawSpan>,
     ty: Option<Type>,
 }
 
+impl Combine for HoverData {
+    fn combine(mut left: Self, right: Self) -> Self {
+        left.values.extend_from_slice(&right.values);
+        left.metadata.extend_from_slice(&right.metadata);
```
yannham, line 31: I believe `extend_from_slice` is cloning, which is not required here. Any reason to not use `append` instead?

yannham (approved):
LGMT. I think we can close the original issue, whose example was really about statically typed pattern matching, and leave any potential extension to another issue/feature request.

lsp/nls/src/requests/hover.rs
jneem: No reason, good catch.

PR #1699, opened 2023-10-25
Fix contract equality on revertible thunks

### LINKED ISSUE #1664: Panic when trying to evaluate an Organist project
**Describe the bug**

Trying to evaluate the `project.ncl` from organist with the latest Nickel master results in

```console
$ nickel export project.ncl          
thread 'main' panicked at 'tried to get data from a revertible thunk without a cached value', core/src/eval/cache/lazy.rs:242:33
```

**To Reproduce**

(with Nickel at 7f1be1812fe0da4d07b4e2f9f3a8d28393ab8f59)

```console
git clone https://github.com/nickel-lang/organist
cd organist
git checkout aad58699b3d43bfa7762d29754183311f9c3b603
nickel export project.ncl
```

**Expected behavior**

Should work™

**Environment**
 - Version of the code: 7f1be1812fe0da4d07b4e2f9f3a8d28393ab8f59

yannham:
Closes #1664. Depends on #1698.

## Context

The introduction of contract deduplication led to a panic reported in #1664. Merging two records produces a new `RecRecord`, that is, a fixpoint. Every part of the previous records that could depend recursively on other fields are reverted to their original expression. This is a form of intermediate state that isn't supposed to be observed: after `merge` returns, the eval loop kicks in again, and will finish to build the reverted thunks by creating and injecting the recursive environment, practically "applying" the fixpoint to get a concrete record.

However, contract deduplication acts at the end of merging, before the fixpoint is recomputed at the next evaluation tick. Thus, it can observe those partially-built values and an `assert` complains somewhere in the implementation of thunks.

The issue is that, morally, we must use the new fixpoint, but we don't have access to it yet. One possibility is to built it as part of merging, but this is annoying, because this would duplicate in part the code already present inside `eval_closure`, even if we may factor out some of it. Secondly, we need to compute the fixpoint before merging the pending contracts, then perform deduplication, and finally combine the (deduplicated, fixpointed) pending contracts. Nothing impossible, but a bit clumsy.

### Fix the issue

This PR instead relies of the already existing but not really exploited behavior of contract equality check, which is that free variables (alternatively, unbound variables) are considered equal if they have the same name. That is, contract equality considers _open terms_, and thus check their equality _assuming they are living in the same environment_. As long as we can ensure there's no unbound variable in practice - which is verified as part of typechecking, this stays sound.

We introduced the `borrow_orig` method to get the cached value of a thunk or its original expression if it's an unitialized revertible thunk. This function is intended to be used only for contract equality check, and is somehow unsatisfying as a leaky abstraction - but it's such a small and simple change, which is appealing. 

### Embrace free variables

Now that we rely on the behavior of contract equality checking on free variables, we might as well bite the bullet and get rid of `NoEnvironment` and the threading of the initial environment through contract equality check. Those were a way to avoid to rely on simple name comparison for variables, but at the cost of more complexity.

Now, contract equality checks simply compares variable by name when encountering symbols from the standard library or the internal module, which is equivalent. This get rids of a lot of complexity, turbofish operators and abstractions (`EnvRefs/EnvOwned`) around contract equality checking.

Passing by, we've updated/reworded some old documentation.

### DIFF
--- a/core/src/closurize.rs
+++ b/core/src/closurize.rs
@@ -295,6 +295,8 @@ pub fn should_share(t: &Term) -> bool {
     }
 }
 
+/// Closurize all the inner terms of a recursive record, including pending contracts and
+/// dynamically defined fields.
 pub fn closurize_rec_record<C: Cache>(
     cache: &mut C,
     data: RecordData,
--- a/core/src/eval/cache/lazy.rs
+++ b/core/src/eval/cache/lazy.rs
@@ -244,6 +244,17 @@ impl ThunkData {
         }
     }
 
+    /// Return a reference to the closure currently cached. Return the original expression if this
+    /// thunk is a revertible thunk that hasn't been built yet.
+    pub fn closure_(&self) -> &Closure {
+        match self.inner {
+            InnerThunkData::Standard(ref closure) => closure,
+            InnerThunkData::Revertible { ref cached, ref orig, .. } => {
+                cached.as_ref().unwrap_or(orig)
+            }
+        }
+    }
+
     /// Return a mutable reference to the closure currently cached.
     pub fn closure_mut(&mut self) -> &mut Closure {
         match self.inner {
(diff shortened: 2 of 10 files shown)


vkleen (approved):
Big fan of fewer `initial_env`s 🎉

thufschmitt:
This is actually causing another breakage in Organist. I'll try to come with a reasonable repro, but it seems to come from a double contract application [here](https://github.com/nickel-lang/organist/blob/b29713c47476e30417a5b7f451ae39f391317441/project.ncl#L36-L45) where the second contract somehow cancels the default values that the first one (`NixpkgsPkgs`) should have set.

PR #1703, opened 2023-10-28
Fix unsound contract equality fast check

### LINKED ISSUE #1700: Contracts applied sequentially can shadow each other
**Describe the bug**

Under some (yet to be defined) circumstances, if a value has two contracts applied to it sequentially, then the first one can be dropped.

**To Reproduce**

Evaluate the following expression:

```ncl
let False = std.contract.from_predicate (fun x => false) in
{ bli = {} }
  | { _| {..} }
  | { _| False }
```

It should fail because of the ` { _ | False }` contract, but somehow the `{ _ | {..} }` contract before makes it pass.

**Expected behavior**

The above code shouldn't evaluate

**Environment**
 - Version of the code: 1921c316ad81bca8100c3a0c6ae2e3da974cdd51

yannham:
Closes #1700.

For performance reason, the contract equality checker performs a quick pointer check to see if the two terms to compare are physically equal. In this case, it used to return `true` directly, eschewing more costly recursive checks.

However, this is unsound, because the same term (physically) might be in two different environments, and thus evaluate in fine to two different values. This is typically the case with function application: when evaluating `f 1` and `f 2`, both terms will physically point to the body of `f`, but in a different environment. Thus, we also need to ensure that environment are equals as well in this fast check.

This commit adds a fast `ptr_eq` check to the environment as well, and now checks that both terms **and** their respective environments are pairwise physically equals.

Passing by, this commits also fix unrelated clippy warning, after the udpate to clippy 1.73.

### DIFF
--- a/core/src/environment.rs
+++ b/core/src/environment.rs
@@ -155,12 +155,8 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
     /// Checks quickly if two environments are obviously equal (when their components are
     /// physically equal as pointers or obviously equal such as being both empty).
     pub(crate) fn ptr_eq(this: &Self, that: &Self) -> bool {
-        // Check if the pointer to the previous layers are physically equal or both `None`
         let prev_layers_eq = match (&*this.previous.borrow(), &*that.previous.borrow()) {
-            (Some(ptr_this), Some(ptr_that)) => {
-                // Check that the current layers are equal, either both equal as pointers or both empty
-                Rc::ptr_eq(ptr_this, ptr_that)
-            }
+            (Some(ptr_this), Some(ptr_that)) => Rc::ptr_eq(ptr_this, ptr_that),
             (None, None) => true,
             _ => false,
         };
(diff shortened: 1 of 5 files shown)


core/src/environment.rs
```diff
@@ -151,6 +151,26 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
...
         Rc::strong_count(&self.current) > 1
     }
+
+    /// Checks quickly if two environments are obviously equal (when their components are
+    /// physically equal as pointers or obviously equal such as being both empty).
+    pub(crate) fn ptr_eq(this: &Self, that: &Self) -> bool {
+        // Check if the pointer to the previous layers are physically equal or both `None`
+        let prev_layers_eq = match (&*this.previous.borrow(), &*that.previous.borrow()) {
+            (Some(ptr_this), Some(ptr_that)) => {
+                // Check that the current layers are equal, either both equal as pointers or both empty
```
jneem, line 161: previous layers, right? Actually, given the preceding comment I think this one is just unnecessary.

jneem (approved):

core/src/environment.rs
yannham: Oops, indeed, this whole comment should actually move to line 168. Or be deleted.

PR #1710, opened 2023-11-09
Add suggestions when misspelling a record field

### LINKED ISSUE #1708: Suggest alternatives when trying to access a non-existent record field

Typos are always easy to make (though the LSP autocompletion helps a lot), and not always easy to spot.
It would make the developer's life easier (at least mine) if the evaluator could suggest similar (but valid) field names in the case of an error. Also, it's really cool :sunglasses: 

For an expression like 

```nickel
{ foo = 1 }.boo
```

Have the evaluator error out with something like

```
error: missing field `boo`
  ┌─ <stdin>:1:1
  │
1 │ { foo = 1 }.boo
  │ ^^^^^^^^^^^^^^^
  │ │
  │ this requires the field `boo` to exist
  │ field `boo` is missing here
  | Did you mean `foo`?
```

ø

**Additional context**

Nix for instance does that, and it can be quite helpful:

```console
$ nix eval --expr '{ foo = 1; }.boo'               
error: attribute 'boo' missing

       at «string»:1:1:

            1| { foo = 1; }.boo
             | ^
       Did you mean foo?
```

yannham:
Closes #1708.

This pull request shows suggestions of similar field names when trying to access a non-existent field of a record.

This first implementation uses the Damerau-Levensthein distance. The choice of the algorithm is inspired from the Rust compiler, although they use a more involved and hand-written variant with corrections around substrings: we just take one from the seemingly most popular string similarity library on crates.io, `strsim`.

More specifically, we try to find either a perfect match ignoring cases, or the field with the highest similarity with the user input, as long as it's above some threshold (to avoid suggesting things that have nothing to do with the input). The threshold was chosen rather arbitrarily, by trial and error: it seems that choosing a higher threshold didn't work that well for very small words (4 letters word), which can have low similarity even with an edit distance of 1 or 2.


cli/tests/snapshot/snapshots/snapshot__eval_stderr_record_access_suggestion.ncl.snap
```diff
@@ -0,0 +1,17 @@
...
+---
+error: missing field `some_fild_nam`
+  ┌─ [INPUTS_PATH]/errors/record_access_suggestion.ncl:3:1
+  │    
+3 │ ╭ ╭ ({
+4 │ │ │   some_field_name = "some value",
+5 │ │ │   another_field_name = "another value",
+6 │ │ │ }).some_fild_nam
+  │ ╰─│────────────────^ this requires the field `some_fild_nam` to exist
+  │   ╰──' field `some_fild_nam` is missing here
```
jneem, line 13: I know this was here previously, but to me "field `some_fild_nam` is missing here" suggests that the field is supposed to be present but isn't. I think something like "this record lacks the field `some_fild_nam`" might be clearer in a situation like this.

core/src/error/suggest.rs
```diff
@@ -0,0 +1,49 @@
...
+//!
+//! The current implementation uses the normalized Damereau-Levenshtein edit distance to find the
+//! closest match.
+
+use strsim::normalized_damerau_levenshtein;
+
+/// The minimum similarity between the user's input and an existing symbol for the symbol to be
+/// considered a serious candidate. The current threshold is rather low, because short words with
+/// edit distance 1 such as `bar` and `bare` might have a similarity which isn't that high.
+pub const MIN_SIMILARITY: f64 = 0.60;
```
jneem, line 13: If the right threshold changes depending on the length, is it worth having a different threshold for short words?

jneem (approved):

core/src/error/suggest.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Find the closest match to the user input in the given list of symbols, to be used as a
+/// suggestion. If there is an exact match (modulo casing), it is returned immediately. Otherwise,
+/// the symbol with the highest normalized Damerau-Levenshtein similarity is returned, as long as
+/// the similarity is above the fixed threshold [MIN_SIMILARITY].
+pub fn find_best_match<'syms, 'input, S, I>(
```
thufschmitt, line 19: I would rather suggest a few best matches rather than the best one. But I also don't have any proof that it's superior, so I'll leave it to your judgment.

core/src/error/suggest.rs
```diff
@@ -0,0 +1,49 @@
...
+        return None;
+    }
+
+    for sym in symbols {
+        // If there is an exact match (modulo case), return it immediately.
+        if sym.as_ref().to_lowercase() == user_input.as_ref().to_lowercase() {
+            return Some(sym.as_ref());
+        }
+
+        let similarity = normalized_damerau_levenshtein(sym.as_ref(), user_input.as_ref());
```
thufschmitt, line 40: (nit): Since we're first comparing things in a case-insensitive manner, would it make sense to also normalise the case when computing the distance?

core/src/eval/operation.rs
```diff
@@ -1571,15 +1567,14 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
-                                        String::from("(.$)"),
-                                        RichTerm {
-                                            term: t2,
-                                            pos: pos2,
-                                        },
+                                    _ => Err(EvalError::FieldMissing {
+                                        id: ident,
+                                        field_names: record
+                                            .field_names(RecordOpKind::IgnoreEmptyOpt),
+                                        operator: String::from("(.$)"),
```
thufschmitt, line 1574: Not related to this PR, but where is that `.$` coming from?

thufschmitt (approved):
Thanks!

core/src/eval/operation.rs
yannham: Oh, good catch. In fact the operator wasn't shown before this PR (basically the `notes` were created but never used - it's unrelated to this PR, it's just that I needed to add the suggestion to the notes, so I realized, and it was in fact surprising that the rust compiler didn't yell at me for unused value). So we might have let old name slipping through. But I believe this was the original syntax for dynamic access field, like `foo.$x` is the current `foo."%{x}"`, because we used to have `$` as the interpolation sign and we didn't have quoted fields yet, such as `foo."bar^baz"`. When we added the later syntax, it felt better to avoid keeping the strange `.$` which could just be subsumed by quoted field access + interpolation.

core/src/error/suggest.rs
yannham: I _think_ Nix, rust and others generally provide only one matches, do they? I'm not sure. I don't have anything against providing a few ones, yeah.

core/src/error/suggest.rs
yannham: I'm not sure. It would be worth it if this threshold was giving obviously bad suggestion for long words, which doesn't seem to the case (but it could be, I've just toyed with a few examples, to be honest). All in all it's just a threshold, so we still pick the most similar world possible. I would say let's see and if this does happen, maybe we will switch to several threshold, or just use the goold old edit distance instead of the normalized one.

core/src/error/suggest.rs
yannham: I'm leaving one suggestion for now, and let multiple suggestion for future work, in order to not delay even more 1.3.

core/src/error/suggest.rs
yannham: I just copied rustc here, and [they do keep the original case] for edit distance computations. I imagine the idea is, if there is an exact match modulo casing, you might have messed up the casing (like a camelCase error) with high probability, but if there's more than casing, then casing is considered as any other edit/mispelling (for example we don't consider keystrokes that are closer to have somehow a smaller cost in the edit distance - of course that would be hard to do across all keyboard layouts, and that's maybe why).

Rustc doing it doesn't mean it's the best approach necessarily, but I guess it at least means it's not entirely stupid :man_shrugging:

core/src/error/suggest.rs
thufschmitt: Nix at least provides several (but I'm the one who wrote it, so…). But yeah, it's certainly not worth blocking anything on this. We can see in the future whether it seems to become an annoyance or not.

PR #1712, opened 2023-11-14
Add a `--field` argument to subcommands whenever it makes sense

### LINKED ISSUE #1408: CLI arguments to pass data to and retrieve data from Nickel configurations

yannham:
Depends on #1709
Closes #1408

This PR is the last step toward closing #1408. It adds a `--field <FIELD_PATH>` parameter to several subcommands, to allow to act or to query only a specific subfield of the configuration. One of the usual motivating use case is to generate several configurations from the same Nickel file:

```console
$ nickel export all_machines.ncl --field machines.pikachu -o pikachu.json
$ nickel export all_machines.ncl --field machines.taupiqueur -o taupiqueur.json
$ nickel export all_machines.ncl --field machines.grolem -o grolem.json
```

This also subsumes the path provided as an argument to `nickel query`, which was the default positional argument in prior versions (<= 1.2) and was temporarily set as `--path` after the CLI rework on master (#1502). It's thus not a breaking change, because `nickel query --path` hasn't made in a release yet. `field` seemed to be more explicit and less confusing than `path` to me.

## Implementation

To avoid code duplication, this PR uses the same trick as for customize mode, which is to ship common options (but not global options) as extra parameter to main commands using clap's `#[command(flatten)]` inclusion mechanism. In particular, although it's part of the customize mode per se - the part after the `--` -- it was just handy to pack the `--field` argument description and implementation there (after all, it can be seen as a form of customization, and adding it as part of the customize mode CLI has been discussed before).

For the field extraction logic, this was mostly already implemented by `VirtualMachnine::query`, so the code has been repurposed as a more general `extract_field` function, which is then used by `program` directly before proceeding further with any operation.

### DIFF
--- a/cli/src/customize/mod.rs
+++ b/cli/src/customize/mod.rs
@@ -252,13 +252,13 @@ impl CustomizeOptions {
         customizable_fields: CustomizableFields,
         mut program: Program<CBNCache>,
     ) -> CliResult<Program<CBNCache>> {
-        let assignment_overrides: Result<Vec<_>, super::error::CliUsageError> = self
+        let assignment_overrides: Result<Vec<_>, CliUsageError> = self
             .assignment
             .into_iter()
             .map(|assignment| {
                 let ovd = program
                     .parse_override(assignment, MergePriority::default())
-                    .map_err(|error| super::error::CliUsageError::AssignmentParseError { error })?;
+                    .map_err(|error| CliUsageError::AssignmentParseError { error })?;
 
                 if !customizable_fields.inputs.contains_key(&ovd.path) {
                     if customizable_fields.overrides.contains_key(&ovd.path) {
@@ -386,8 +386,6 @@ fn program_with_field(
     mut program: Program<CBNCache>,
     field: Option<String>,
 ) -> CliResult<Program<CBNCache>> {
-    use super::error::{CliUsageError, Error};
-
     if let Some(field) = field {
         match program.parse_field_path(field) {
             Ok(field_path) => program.field = field_path,
--- a/cli/src/customize/mod.rs
+++ b/cli/src/customize/mod.rs
@@ -31,8 +31,14 @@ const ASSIGNMENT_SYNTAX: &str = "FIELD_PATH=NICKEL_EXPRESSION";
 const EXPERIMENTAL_MSG: &str =
     "[WARNING] Customize mode is experimental. Its interface is subject to breaking changes.";
 
+/// For commands where full customization is allowed, combining extracting a target field and
+/// overriding arbitrary values of the configuration.
 #[derive(clap::Parser, Debug)]
 pub struct CustomizeMode {
+    /// Query a specific field of the configuration. If omitted, the top-level value is queried
+    #[arg(long, short, value_name = "FIELD_PATH")]
+    pub field: Option<String>,
+
     /// \[WARNING\] Customize mode is experimental. Its interface is subject to breaking changes.
     ///
     /// Customize mode turns the nickel invocation into a new CLI based on the configuration to be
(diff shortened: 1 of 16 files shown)


cli/src/customize/mod.rs
```diff
@@ -258,10 +266,18 @@ impl CustomizableFields {
 }
 
 pub trait Customize {
+    /// Customize the program with the given options.
     fn customize(&self, program: Program<CBNCache>) -> CliResult<Program<CBNCache>>;
+    /// Return the value of the `field` argument, if specified and if supported by the current
+    /// customize variant.
+    fn field(&self) -> Option<&String> {
```
Radvendii, line 273: Is there a reason to use `&String` here instead of `&str`?

Radvendii (approved):
It would be nice if this could be orthogonal to customize mode in the code, but I don't know clap well enough to think about how that would be done. Looks good enough for now, and we can always come back to it if/when there's more flags that are shared between different sets of commands.

jneem (approved):

cli/src/customize/mod.rs
yannham: Not really. In fact we don't even use the content, what I use it for is to check if `field` is set in `nickel query` and issue a warning if not. Maybe the interface should then be minimal, that is `is_field_set(&self) -> bool`, which eschews the question entirely

yannham:
> It would be nice if this could be orthogonal to customize mode in the code, but I don't know clap well enough to think about how that would be done.

You're right, I believe it should be its own `extract_field` submodule, and I don't foresee any particular difficulty. Also it's somehow vaguely related to customize (if only because it's also a form of customization), in practice it's not the same part of the command line at all. I'll post-pone this task post-release, though.

PR #1721, opened 2023-11-22
Add a cli param to specify the search path

jneem:
This adds the cli argument version of the `NICKEL_PATH` environment variable. A couple points that could be bike-shedded:

- comma-separated vs colon-separated. I feel like comma-separated is "traditional" in arguments while colon-separated is "traditional" in environment variables, but I have no other justification for this choice.
- the name of the cli arg. Maybe the `nickel` prefix is unnecessary? I think it's justified for the env var because it lives in a global namespace.

### DIFF
--- a/cli/src/input.rs
+++ b/cli/src/input.rs
@@ -14,6 +14,15 @@ pub struct InputOptions<Customize: clap::Args> {
     #[arg(long, global = true)]
     pub nostdlib: bool,
 
+    /// Specifies a (comma-separated) list of paths to search for imports in.
+    ///
+    /// When importing a file, nickel searches for it relative to the file doing the
+    /// import. If not found, it searches in the paths specified by `--nickel-path`.
+    /// If not found there, it searches in the (colon-separated) list of paths contained
+    /// in the environment variable `NICKEL_PATH`.
+    #[arg(long, global = true, value_delimiter = ',', num_args = 1..)]
+    pub nickel_path: Option<Vec<PathBuf>>,
+
     #[command(flatten)]
     pub customize_mode: Customize,
 }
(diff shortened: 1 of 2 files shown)


yannham (approved):
The code looks fine. As for the bikeshedding, I would be tempted to say that having a parameter of the same name as the environment variable might avoid confusion (also, just having `--path` is not really clear; we could have `--include-path`, but then the two names are really distinct and not even in a prefix relationship).

As for the delimiter, I honestly have no idea, but we should probably pick 5 compilers/interpreters that are both mainstream and known for having a good CLI UX, and research what they do, and why?

vkleen:
About the separator, the tradition with compilers is to just pass multiple `include-path` arguments. For example, `gcc -Iinclude -Imore-includes`.

yannham:
@jneem what do you think about Viktor's answer? Should we make the CLI argument just take multiple values and sidestep the question of the delimiter entirely? At first sight it sounds more appealing/less error-prone than having two different delimiters.

jneem:
Yes, that sounds like a good plan, I'll try to find some time today to change it.

I also had a quick look at some mainstream compilers (clang, gcc, rustc, ocamlc, ghc, javac). The multiple argument approach does indeed seem the most common (exceptions: ghc supports both multiple arguments and colon-separated arguments; javac supports only colon-separated arguments (I think: I only read the docs, didn't test)).

The naming between env var and cli flag doesn't have a clear precedent. Most compilers seem to *only* support the short flag `-I` (or `-i`, or `-L`). Exception is javac, with --class-path and --module-path and so on. Support for environment variables seems not so common: rustc, ocamlc, and ghc don't support them. clang and gcc's environment variable names have nothing to do with the cli flag. javac has the same name for environment variable and cli flag.

yannham:
Another solution is to rename the env var `NICKEL_INCLUDE_PATH` (which is somehow more explicit, IMHO) and call the argument `--include-path` (with short `-I`) which would be more or less consistent, and be in line with existing tools. Allowing colon-delimited paths probably don't hurt (as it's consistent with the environment variable), although it's probably less useful because you can just use the environment variable for that.

jneem:
While implementing this, it occurred to me that maybe "import" is better than "include", since that's what we call the operator?

yannham:
> While implementing this, it occurred to me that maybe "import" is better than "include", since that's what we call the operator?

Sounds reasonable, indeed!

Issue #1656, opened 2023-09-29
Default output path of `nickel doc` is unexpected


gkleen:
I was quite surprised that `nickel doc` by default wants to write it's output to a file (and [accidentally commited](https://gitlab.uniworx.de/uniworx/machines/srv01/-/tree/9e0066e47240c003f82860c73eafcfd551da60cd/srv02/kubernetes/external-dns) one of those output files, even).

I think it would make more sense for `nickel doc` to emit to stdout by default?

yannham:
I think the rationale is that it's more common to generate doc somewhere to then publish it or commit it, while it's more uncommon to post-process this documentation directly on the command line. Most documentation tools I know needs to generate multiple HTML files anyway - and nickel could do the same -, which don't make much sense to print on stdout.

That being said, I think we currently don't even print a message to say "Written doc in ...". Would that help to add such an output? Also, what I said might not apply to outputting the doc as `json`, where it might be indeed common to pipe the output to something else. I'm not sure yet it would be a good idea to have a different default behavior depending on the format, though.

gkleen:
A message stating the path of the output file would certainly have alleviated my surprise.  
I think it also wouldn't seem out of place in other automations like the generation of html docs like you describe :)

P.S.: Sorry for the late response – seem to have had some imap issues that went unnoticed; getting to the other issues now, too…

PR #1729, opened 2023-11-30
Print something when nickel doc succeeds

yannham:
Closues #1656.

The doc subcommand has been reported as confusing, because it generates file in a default location but just returns silently upon success. This PR adds a print statement that says where the document has been generated instead, and serves a confirmation that everything went well as well.

### DIFF
--- a/cli/src/doc.rs
+++ b/cli/src/doc.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Cow,
     fmt, fs,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


cli/src/doc.rs
```diff
@@ -67,59 +70,82 @@ impl DocCommand {
...
             DocFormat::Markdown => doc.write_markdown(&mut out),
+        }?;
+
+        if !self.stdout {
+            eprintln!(
+                "Documentation written to {}",
+                out_path
+                    .as_ref()
+                    .map(Cow::as_ref)
+                    .unwrap_or(DEFAULT_OUT_DIR)
```
vkleen, line 145:
```suggestion
                    .expect("There should always be an output path")
```
Since we're always producing an output path if `self.stdout` is not set, I feel it would be better to panic here if no output path gets set. Just as an insurance policy against the path logic changing in the future.

vkleen (approved):

cli/src/doc.rs
yannham: Good point, but then I think I prefer to destructure the option instead of relying on the related-but-the-compiler-can't-prove-it boolean flag. That is, if `out_path` is Some, we print it, otherwise we don't print anything.

PR #1730, opened 2023-11-30
Add kcl comparison

jneem:


RATIONALE.md
```diff
@@ -432,6 +432,22 @@ The main difference between Jsonnet and Nickel are types. Jsonnet does not
 feature static types, contracts or metadata, and thus can't type library code
 and has no principled approach to data validation.
 
+### KCL vs Nickel
+
```
yannham, line 436: Would that be possible to add a small sentence first to say that they have many things common as well, and that KCL seems to be a capable configuration language? (types, validation, modularity, merging, functions)

yannham (commented):
Thanks for this! Would you mind add KCL to the table at the end of this document, and to the copy that is at the end of the README as well ?

yannham (approved):
Perfect, thanks!

PR #1734, opened 2023-12-03
Support importing txt files as strings

Quantum64:
Files that contain raw text shouldn't be interpreted as Nickel code

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -41,7 +41,7 @@ pub enum InputFormat {
     Toml,
     #[cfg(feature = "nix-experimental")]
     Nix,
-    String,
+    Raw,
 }
 
 impl InputFormat {
@@ -54,7 +54,7 @@ impl InputFormat {
             Some("toml") => Some(InputFormat::Toml),
             #[cfg(feature = "nix-experimental")]
             Some("nix") => Some(InputFormat::Nix),
-            Some("txt") => Some(InputFormat::String),
+            Some("txt") => Some(InputFormat::Raw),
             _ => None,
         }
     }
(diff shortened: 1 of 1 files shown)


core/src/cache.rs
```diff
@@ -53,6 +54,7 @@ impl InputFormat {
             Some("toml") => Some(InputFormat::Toml),
             #[cfg(feature = "nix-experimental")]
             Some("nix") => Some(InputFormat::Nix),
+            Some("txt") => Some(InputFormat::String),
```
yannham, line 57:
```suggestion
            Some("txt") => Some(InputFormat::Raw),
```

core/src/cache.rs
```diff
@@ -616,6 +618,10 @@ impl Cache {
                     .map(|t| (attach_pos(t), ParseErrors::default()))
                     .map_err(|err| ParseError::from_serde_json(err, file_id, &self.files))
             }
+            InputFormat::String => Ok((
```
yannham, line 621:
```suggestion
            InputFormat::Raw => Ok((
```

yannham (approved):
Thanks for contributing!

I think we've discussed an extended import syntax somewhere in package management-related issues, that would allow to specify what is the format of an import, and could be a more robust mechanism (or at least, an additional one, if you don't use standard extensions).

However, we already use file extensions to decide the format currently, and I can't think of another reasonable way to interpret a `.txt` file than as `String`, so this makes sense to me.

Quantum64:
> I think we've discussed an extended import syntax somewhere in package management-related issues, that would allow to specify what is the format of an import

That would be great. File extension based semantics feels a bit shaky.

PR #1740, opened 2023-12-18
Add --error-format flag to serialize err diagnostics

### LINKED ISSUE #1738: Serialize errors as JSON

People have reported using Nickel in a CI or an automated pipeline. There, if there's an error, the output is not very machine-friendly: they have to strip the decorations and parse the output, which could be fragile.

Add an option `--error-format json` or something like that to make `nickel` output errors as JSON instead of ANSI terminal text.

Do nothing and parse the errors, as is already possible.

**Additional context**

This shouldn't be very hard to do, as before being printed, errors are transformed to `codespan`'s diagnostics which are probably straightforward to serialize.

yannham:
Closes #1738.

This PR adds an `--error-format` global flag (copied over from Rust, after benchmarking what a few other compilers are donig, this seemed to be the better tradeoff between being intuitive and not conflicting with other meaning of "format as JSON", which we have a lot in Nickel - export format, input format, etc.).

The user can choose between JSON, TOML or YAML - I believe JSON is the most common one, but it's virtually free to support more formats with `serde` (at the least the format that we export Nickel configurations to), hence I don't see an obvious reason to not support them.

Incidentally, this PR formats more CLI errors using codespan diagnostics (they were just printed as text before), so that the flag really affects all error messages. As a bonus, those error messages are now nicely colored like the others.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3362992a0d9f1dd7c3d0e89e0ab2bb540b7a95fea8cd798090e758fda2899b5e"
 dependencies = [
  "codespan-reporting",
+ "serde",
 ]
 
 [[package]]
@@ -424,6 +425,7 @@ version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3538270d33cc669650c4b093848450d380def10c331d38c768e34cac80576e6e"
 dependencies = [
+ "serde",
  "termcolor",
  "unicode-width",
 ]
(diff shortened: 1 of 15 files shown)


jneem (approved):
Does this imply any stability guarantees for the error messages? I'd imagine that if people are consuming them programmatically then they're expecting at least part of the format to be stable...

yannham:
> Does this imply any stability guarantees for the error messages? I'd imagine that if people are consuming them programmatically then they're expecting at least part of the format to be stable...

Good question. I would be tempted to say "no", because it makes our life simpler :stuck_out_tongue: and also because currently it's just serializing the library datatype that we happen to use (`codespan_reporting`). It's probably not too hard to convert it if we ever change the error reporting infrastructure, but still. In the meantime, I should probably add a mention in the documentation of the CLI argument that the format isn't stable.

PR #1746, opened 2023-12-21
Completion in arrays

### LINKED ISSUE #1732: Arrays confuse the auto-completer
The auto-completer is not good at auto-completing records inside arrays. For example, in

```nickel
let LanguageConfiguration = {
  name | String,
}
in
let HxLang = { language | Array LanguageConfiguration, .. } in
{
  language = [
    { na }
  ]
}
  | HxLang
```

we don't get an autocompletion for "name". If we drop the arrays, we get the autocompletion.

jneem:
This makes the completer aware of array contracts, allowing it to provide completion in more circumstances. The tl;dr is that `[{ a }] | Array { aardvark | String }` will now provide the completion "aardvark" after the "a".

Fixes #1732

### DIFF
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -372,21 +372,6 @@ impl<'a> FieldResolver<'a> {
                     Vec::new()
                 }
             }
-            Term::Array(_, attrs) => attrs
-                .pending_contracts
-                .iter()
-                .filter_map(|ctr| {
-                    if let Term::Type(ty) = ctr.contract.as_ref() {
-                        if let TypeF::Array(elt_ty) = &ty.typ {
-                            Some(Container::Array((**elt_ty).clone()))
-                        } else {
-                            None
-                        }
-                    } else {
-                        None
-                    }
-                })
-                .collect(),
             Term::ResolvedImport(file_id) => self
                 .server
                 .cache
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -42,7 +42,7 @@ impl From<Ident> for EltId {
 }
 
 impl Container {
-    /// If this `FieldHaver` has a field named `id`, returns its value.
+    /// If this `Container` has a field named `id`, returns its value.
     fn get(&self, id: EltId) -> Option<FieldContent> {
         match (self, id) {
             (Container::RecordTerm(data), EltId::Ident(id)) => data
@@ -58,7 +58,7 @@ impl Container {
         }
     }
 
-    /// If this `FieldHaver` is a record term, try to retrieve the field named `id`.
+    /// If this `Container` is a record term, try to retrieve the field named `id`.
     fn get_field_and_loc(&self, id: Ident) -> Option<(LocIdent, &Field)> {
         match self {
             Container::RecordTerm(data) => data
@@ -81,7 +81,7 @@ impl Container {
         }
     }
 
-    /// Returns all fields in this `FieldHaver`, rendered as LSP completion items.
+    /// Returns all fields in this `Container`, rendered as LSP completion items.
     pub fn completion_items(&self) -> impl Iterator<Item = CompletionItem> + '_ {
         match self {
             Container::RecordTerm(data) => {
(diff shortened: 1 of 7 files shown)


lsp/nls/src/field_walker.rs
```diff
@@ -13,35 +13,55 @@ use nickel_lang_core::{
...
-/// A `FieldHaver` is something that... has fields.
+/// A `Container` is something that has elements.
 ///
-/// You can use a [`FieldResolver`] to resolve terms, or terms with paths, to `FieldHaver`s.
+/// The elements could have names (e.g. in a record) or not (e.g. in an array).
+///
+/// You can use a [`FieldResolver`] to resolve terms, or terms with paths, to `Container`s.
 #[derive(Clone, Debug, PartialEq)]
-pub enum FieldHaver {
+pub enum Container {
```
yannham, line 22: I wonder: would that make sense to make this a trait instead? It seems that most of what you do below is to match on the type of container, and then apply specific behavior. It sounds a lot like what traits are for.

Also, it feels slightly odd to mix array and records, leading to a rather vague notion of container (I guess it's why this is hard to give it a meaningful name) and an `EltId` which also feels like it smashes two different things together to then match on the different possibilities and apply specific behavior (also, if I'm understanding correctly, some illegal cases such as trying to get an ident from an array aren't statically excluded but just return `None`).

At this point I might be oblivious to important details that make the current approach actually justified, sorry if this is the case. But from a distance, would that lead to much code duplication to have:

A trait `FieldHaver` with the methods you already implemented before this PR, and implement it for `RecordData`, `Type`, and `RecordRows`.
A trait `Indexable` (that's a bad name and it doesn't relate to `FieldHaver`, so it's doubly bad...) that is mostly doing the same but for array elements, which doesn't use `Ident` nor `field` in the method's name, and implement that for `Array`.

What do you think?

lsp/nls/src/field_walker.rs
```diff
@@ -347,6 +372,21 @@ impl<'a> FieldResolver<'a> {
                     Vec::new()
                 }
             }
+            Term::Array(_, attrs) => attrs
+                .pending_contracts
```
yannham, line 376: I'm surprised that there's anything in `pending_contracts` at this point. If I recall correctly, `pending_contracts` on array attributes is only filled at evaluation type when applying an `Array` contract, which updates this field. It's a runtime thing. Do you get anything out of it here in practice?

yannham (commented):
I like the fact that adding this feature is mostly a small chunk of boring but expected code, which is a good sign that the current architecture of the LSP is way better than before :slightly_smiling_face:

lsp/nls/src/field_walker.rs
jneem: Hm, I'm not actually sure. I'll check if removing it changes anything.

lsp/nls/src/field_walker.rs
jneem: Good points, I will think more about this over the break. I think there is at least one situation where you do want to dynamically handle either a record field or an array element, but it's probably a smaller use case and there's probably a better organization.

To elaborate, suppose we're trying to complete at the `a` in `{foo = [{ a }]} | {foo | Array { aardvark | String }}`. The way we currently do this without any evaluation is to first construct the path to the place we want completion. In some weird notation I just made up, this path looks like `. foo / !!` (where the `!!` means an array index -- we don't track what the index actually was). Then we follow the same path on the contract `{foo | Array { aardvark | String }}` to find the `aardvark` completion. 

So the path does need to alternate at runtime between field accesses and array indices, and it might also happen that the path-follower tries to match an array index against a record (say, if the contract had been `{foo | {_: {aardvark | String}}}`). I don't know of any *valid* nickel code where this will happen, but of course nls also needs to handle invalid nickel code.

lsp/nls/src/field_walker.rs
yannham: @jneem let's proceed with this version, code structure is less important. However, maybe we should flag this now with either a comment or remove it altogether (at worse re-add it later if it turns out to do something in practice, but I doubt it)

yannham (approved):
While code structure is important for long-term maintainability, the code in this PR is still entirely reasonable and brings a nice feature for 1.4. Let's proceed in this state (beside my comment around pending_contracts, to solve first), include this in 1.4 and from there think harder about the right way to structure this.

PR #1748, opened 2024-01-03
Add a release script

yannham:
Automate part of the releasing process (which is becoming tedious) in a dedicated script.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -7,9 +7,44 @@ crates and dependent repositories (such as the website) in a consistent state.
 
 ## Steps to make a release
 
+### Releasing script
+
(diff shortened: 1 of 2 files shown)


jneem (approved):
Nice to have this automated!

Did you see if there's anything useful that `cargo release` has to offer? The topiary stuff probably means we can't just use it out-of-the-box, but for example the `cargo release version` subcommand might get rid of the version-related sed stuff.

yannham:
> Did you see if there's anything useful that cargo release has to offer?

I haven't looked at it. Thanks for the suggestion, I'll see if it can make some parts of the script simpler!

PR #1749, opened 2024-01-05
format: don't fail silently on invalid input anymore

yannham:
`nickel format` is silently doing nothing and exit with success code when the input file doesn't parse, which can be surprising (you think that you correctly formatted your codebase, but in fact the file doesn't even parse). This is due to using `tolerate_parsing_errors: true`, which as reported by @vkleen was a fix for a time when Topiary would simply erase some files upon parsing errors, which was of course problematic, to say the least.

This isn't the case anymore, and thus we can get rid of this option and correctly report parse errors when formatting.

### DIFF
--- a/core/src/format.rs
+++ b/core/src/format.rs
@@ -32,7 +32,7 @@ pub fn format(mut input: impl Read, mut output: impl Write) -> Result<(), Format
         &grammar,
         topiary::Operation::Format {
             skip_idempotence: true,
-            tolerate_parsing_errors: true,
+            tolerate_parsing_errors: false,
         },
     )
     .map_err(FormatError)


vkleen:
When trying to format a file containing a single `{` with this PR, while `nickel` doesn't erase the contents of the file anymore, it still silently pretends to succeed. Maybe we need to do something more when calling `topiary`?

yannham:
In fact I noticed that unexpected end of file doesn't seem to cause any error, and the file is also actually formatted (although the last unclosed block is a bit messed up sometime). So something (Topiary, tree-sitter, or the Nickel grammar) is probably wired to handle incomplete input.

However, if you add an expression like `1 + + 2`  somewhere in your file, this should fail. At least it does on my side.

vkleen (approved):
You're right, with `1 + + 2` in the file it fails for me as well.

PR #1752, opened 2024-01-08
Fix the order of argument in (.)

yannham:
The curried dot operator `(.)` has been added to Nickel 1.3. However, due to an overlook of the order of arguments of the underlying primitive operator (`DynAccess`), the arguments of this operator have been flipped: in 1.3, `(.) foo bar` is `bar."%{foo}"` instead of `foo.%{"bar"}`.

While it's tempting to keep this way - in particular because it's very useful combined with the reverse application operator - this is both unintentional and inconsistent with every other curried operator. Although it's a breaking change, I believe it's the right price to pay to maintain consistency as much as possible and keep to the principle of least surprise for newcomers.

In addition to fixing the order of arguments for `(.)`, a new function `std.record.get` has been added to the stdlib to make up for the previous usage of the flipped `(.)` in sequence of `|>` applications: any previous usage of `(.)` can safely be replaced with `std.record.get`.


RELEASES.md
```diff
@@ -1,3 +1,24 @@
...
+Breaking changes
+----------------
+
+* The curried dot operator added in Nickel 1.3 was implemented the wrong way:
+    the arguments were flipped, meaning that `(.) foo bar` was `bar."%{foo}"`
+    instead of the expected `foo."%{bar}"`. While the initial flipped
+    implementation seems more useful for piping operations using the reverse
+    application operator `|>`, it's inconsistent with all other curried
+    operators, where `(<operator>)` is always define as `fun x y => x <operator>
+    y`. To ensure consistency, and because the initial behavior was an overlook
```
jneem, line 13:
```suggestion
    y`. To ensure consistency, and because the initial behavior was an oversight
```

jneem (approved):

PR #1756, opened 2024-01-11
Add "goto definition" support for the import term

### LINKED ISSUE #1754: Regression 1.2.3 -> 1.3.0: goto definition on an import expression doesn't work anymore
**Describe the bug**

In the LSP, requesting a _go to definition_ when the cursor is on an import statement used to jump to the corresponding file. I checked and this behavior is still there for 1.2.2.

However, since 1.3.0, the 

**To Reproduce**

```nickel
# file a.ncl
null

# file b.ncl
{
  a = import "./a.ncl",
}
```

Now go to definition when the cursors is somewhere within `import "./a.ncl"` will just show "definition not found".

**Expected behavior**

I expect that requesting the definition anywhere within `import "./a.ncl"` open `a.ncl` and jumps to the beginning of the file, as it used to.

**Environment**
 - Version of the code: 1.3.0-release

**Additional context**

I think we got rid of the old linearizer between 1.2.2 and 1.3.0, so this is the most probable suspect. cc @jneem

jneem:
After this PR, asking for "goto definition" on an import term will go to that file.

Most of this PR is relaxing a previous assumption that the "definition" is always an identifier (because in the case of an import, the definition is an arbitrary term). Instead, the answer for a definition query is now an arbitrary `RawSpan`.

Fixes #1754

### DIFF
--- a/lsp/nls/src/requests/goto.rs
+++ b/lsp/nls/src/requests/goto.rs
@@ -1,6 +1,9 @@
 use lsp_server::{RequestId, Response, ResponseError};
 use lsp_types::{GotoDefinitionParams, GotoDefinitionResponse, Location, ReferenceParams};
-use nickel_lang_core::term::{record::FieldMetadata, RichTerm, Term, UnaryOp};
+use nickel_lang_core::{
+    position::RawSpan,
+    term::{record::FieldMetadata, RichTerm, Term, UnaryOp},
+};
 use serde_json::Value;
 
 use crate::{
@@ -11,7 +14,7 @@ use crate::{
     server::Server,
 };
 
-fn get_defs(term: &RichTerm, ident: Option<LocIdent>, server: &Server) -> Option<Vec<LocIdent>> {
+fn get_defs(term: &RichTerm, ident: Option<LocIdent>, server: &Server) -> Option<Vec<RawSpan>> {
     let resolver = FieldResolver::new(server);
     let ret = match (term.as_ref(), ident) {
         (Term::Var(id), _) => {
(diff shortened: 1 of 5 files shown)


lsp/nls/src/requests/goto.rs
```diff
@@ -113,20 +131,26 @@ pub fn handle_references(
...
-                let def = Def::Field {
-                    ident: id,
-                    value: None,
-                    record: parent.clone(),
-                    metadata: FieldMetadata::default(),
-                };
-                let resolver = FieldResolver::new(server);
-                let cousins = resolver.get_cousin_defs(&def);
-                def_locs.extend(cousins.into_iter().map(|(loc, _)| loc))
+        if let Some(span) = id.pos.into_opt() {
```
yannham, line 134: Nitpick: should we combine the two `if`, to avoid one level of nesting? Such as

```rust
if let Some(span) = server.lookup_ident_by_position(pos)?.and_then(|id| id.pos.into_opt())
```

lsp/nls/src/usage.rs
```diff
@@ -37,7 +37,8 @@ pub struct UsageLookup {
     // This could be made more general (we might want to map arbitrary positions to
     // environments) but its enough for us now.
     def_table: HashMap<RawSpan, Environment>,
-    usage_table: HashMap<LocIdent, Vec<LocIdent>>,
+    // Maps from spans of idents to the places where they are referenced.
+    usage_table: HashMap<RawSpan, Vec<LocIdent>>,
```
yannham, line 41: Just to make sure to understand: before, the table was hashing on _both the identifier and the position_ ? And now we just use the position, which should be entirely the same for located idents (that is, hashing on the ident as well was useless anyway) ?

yannham (approved):

lsp/nls/src/usage.rs
jneem: Exactly. In fact, we are still hashing a little too much, because the `RawSpan` contains an unnecessary `FileId`...

lsp/nls/src/requests/goto.rs
jneem: Oh wait, now I remember why it was like this: I needed to bind the `LocIdent` because it's used a few lines down to construct the `Def::Field`.

PR #1760, opened 2024-01-12
Stop using spot instances for release builds

vkleen:
We now had a failure to build release binaries due to a spot instance getting canceled early for the second time. For now, I propose not using a spot instance. While this will increase the cost, with it only getting run once per release, I don't think there's cause for concern.

This change is change is already deployed to the AWS account to build the 1.4 release artifacts.

### DIFF
--- a/infra/spot_lambdas/start.py
+++ b/infra/spot_lambdas/start.py
@@ -22,13 +22,6 @@ def lambda_handler(event, context):
             "LaunchTemplateId": launch_template_id,
             "Version": "$Latest"
         },
-        InstanceMarketOptions={
-            "MarketType": "spot",
-            "SpotOptions": {
-                "SpotInstanceType": "one-time",
-                "InstanceInterruptionBehavior": "terminate",
-            },
-        },
         TagSpecifications=[
             {
                 "ResourceType": "instance",


infra/spot_lambdas/start.py
```diff
@@ -22,13 +22,6 @@ def lambda_handler(event, context):
             "LaunchTemplateId": launch_template_id,
             "Version": "$Latest"
         },
-        InstanceMarketOptions={
-            "MarketType": "spot",
-            "SpotOptions": {
-                "SpotInstanceType": "one-time",
-                "InstanceInterruptionBehavior": "terminate",
```
yannham, line 29: Out of curiosity, would there any other behavior that make sense here, and would avoid restarting from zero?

yannham (approved):

infra/spot_lambdas/start.py
vkleen: The only other option is to stop hibernate the instance, I think. But this will still cause the GitHub job to be canceled and we'd still need some way to claen up stale instances and retry automatically. Also, with the GitHub job canceled, we'd restart from scratch anyway.

PR #1768, opened 2024-01-19
Allow metadata keywords in field position

### LINKED ISSUE #1765: It shouldn't be necessary to quote keywords
This works:
```ncl
{
  value = "Hello",
  "optional" = false,
}
```
and this doesn't:
```ncl
{
  value = "Hello",
  optional = false,
}
```

Shouldn't the compiler be able to recognize it as a key definition at this point? I don't see any ambiguity there (as far as my knowledge of the Nickel syntax goes).

yannham:
Closes #1765.

Metadata keywords (optional, default, etc.) are numerous and they can't clash with field names, because they don't appear in the same positions. Thus, to avoid having user to quote their fields, this PR allows unambiguously the usage of metadata keyword in field position.

I suspect we could actually proceed with even more keywords of the language, but let's start small and avoid shooting ourselves in the foot. The metadata keywords is the set which isn't small and is likely to grow, while the other keywords should stay relatively stable.

Note that currently metadata keyword are rejected within let destucturing, because destructuring binds variables, and variables which are keywords can't be accessed in the current syntax. We might change that if we ever implement raw identifiers.

### DIFF
--- a/core/src/parser/grammar.lalrpop
+++ b/core/src/parser/grammar.lalrpop
@@ -321,7 +321,7 @@ TypeArray: Type = "Array" <t: AsType<Atom>> =>
 
 // A record operation chain, such as `{foo = data}.bar.baz`.
 RecordOperationChain: RichTerm = {
-    <t: AsTerm<Atom>> "." <id: Ident> => mk_term::op1(UnaryOp::StaticAccess(id), t).with_pos(id.pos),
+    <t: AsTerm<Atom>> "." <id: ExtendedIdent> => mk_term::op1(UnaryOp::StaticAccess(id), t).with_pos(id.pos),
     <t: AsTerm<Atom>> "." <t_id: WithPos<StrChunks>> => mk_access(t_id, t),
 };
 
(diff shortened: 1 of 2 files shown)


core/tests/integration/pass/core/records.ncl
```diff
@@ -1,141 +1,165 @@
 # test.type = 'pass'
```
yannham, line 1: This field has a big diff because I formatted it with Topiary. You can ignore most of the changes but the very end, where I added a test for metadata keyword in field position.

core/src/parser/grammar.lalrpop
```diff
@@ -557,6 +557,30 @@ Match: Match = {
 // A default annotation in a pattern.
 DefaultAnnot: RichTerm = "?" <t: Term> => t;
 
+// A metadata keyword returned as an indent. In some positions, those are
```
vkleen, line 560: Can we do this for all (or most) of our keywords? I think even a `let` or `rec` should be unambiguous in field name positions.

yannham: I would vote for doing this for metadata keyword only at first. Although I don't have right now any example of an existing keyword or a future extension that could cause an issue, but let's just give it a bit more thinking (metadata keyword are much more restricted in where they can appear). At worst we can do that in a follow-up PR.

vkleen (approved):

Issue #1448, opened 2023-07-12
Algebraic Data Types


yannham:
# Algebraic Data Types

Algebraic data types (ADTs) - also called tagged unions - and the ability to pattern match on them is a common and very well appreciated feature of most typed functional languages. If record types can be understood as product types, that is a logical `and` (`{n : Number, s : String}` is a string and a number together), ADTs are sum types, corresponding to `or`. Having both of them allow to represent most common shapes of data.

Nickel enums are a very limited form of tagged unions, which can't hold any data, but have only a tag.

## Motivation

ADTs are widely useful, so it's difficult to write a precise and exhaustive list of their applications, but we missed them in many occasions while working on Nickel code:

- Represent nullable values (especially in statically typed code), although this one might be better special cased in the language as a bare (untagged) union
- Representing errors (Rust/OCaml/Haskell's `Result`)
- Writing folds. Nickel doesn't have a looping construct, hence iteration is performed using fold functions such as `std.array.fold`. The "state" of the loop is passed along, and it's been often the case that this state is naturally an ADT.
- Better represent the possible return values of some stdlib functions, and more generally user-defined helper functions. Typically, `std.string.find` returns `{matched = "", index = -1, groups = []}` if the pattern isn't found in the string, which is not very elegant.

## Representation

Currently, all Nickel primitive values (numbers, strings, booleans, arrays and records) directly map to an equivalent in the JSON data model. This is deliberate as JSON interop is of high importance.

This property isn't true for ADTs. Take a hypothetical Nickel ADT: `type Either = 'Left Number | 'Right String`. A reasonable encoding in JSON could be something like a record `{tag, value}` where `tag : [| 'Left, 'Right |]` and `value` is either of type `Number` or `String`, depending on the `tag`.

This is in fact how we've encoded tagged union in Nickel in practice, up to now (of course, by doing it manually, we don't have any pattern matching capability and we can represent many invalid values).

This encoding isn't unique. For once, unions are used in the wild in configurations, but they are not necessarily tagged: a more natural and direct representation of `Either` could be just one value that is either a string or a number, corresponding to the contract:

```nickel
Either = std.contract.from_predicate
  (fun value => std.is_number value || std.is_string value)
```

Second, even if the tag+value encoding is used, the choice of field names is arbitrary. We could have chosen `discriminant` and `data` for example.

[...]

PR #1770, opened 2024-01-22
Structural ADTs: first step

yannham:
MVP for adding structural ADTs to Nickel (see #1448), see what works and what breaks, and iterate from there.

## Content

This PR adds some preliminary elements of structural algebraic data types, although some functionalities are still missing. The goal is to produce a reviewable chunk of work, that doesn't impact the normal usage of the language, and doesn't commit permanently when bikshedding needs to be involved. More specifically, this PR:

- Adds a new `EnumVariant(Ident, RichTerm)` node to the AST, which represent an applied enum tag (ie an ADT with one argument). Extend the existing code to support it/fix compiler errors.
- Adds a temporary, intentionally ugly (but easy to add and non-ambiguous) syntax for experimentation: `'Foo...(arg)` (yes, it's a literal ellipsis, not the expression of some kind of repetition)
- Extend enum row types to take an additional optional argument. Current enum tags correspond to a row `EnumRow { id, typ: None}`, and an applied tag corresponds to `EnumRow {id, typ: Some(ty_arg)`
- Add the most likely final (because non ambiguous and uncontroversial) syntax for those new rows, which is just a tag applied to another type. For example, now `[| 'Foo, 'Bar Number, 'Baz {foo : String}, 'Motto (Dyn -> Dyn) |]` is a valid type.
- Extend the typechecker to handle those new rows and the new AST node

I've also expanded/reworded some of the existing in-code documentation: going through our own comments a few month later is a good exercise to see what is useful, and what is missing, when the technical aspects aren't so fresh anymore :slightly_smiling_face: 

The bulk of this PR, beside adding new definitions and fixing compiler errors, is to extend row types and adapt the typechecker to handle the corresponding new cases. In particular, a lot of cases were previously simpler for enum rows (as compared to record rows) because those couldn't contain any other type inside, beside an enum row unification variable in tail position. Now, as with record rows, enum rows might contain normal types inside, which thus must be walked/folded over/variable-level updated, etc. It actually makes the handling of 

## Missing features

The following important aspects are missing for the moment, and will be covered in follow-up PRs:

- extracting data/matching an ADT. There's currently no way to destructure or unwrap an enum variant (pedantically, there's no elimination form). In general, I think this might be time to allow destructuring patterns in match expressions - put differently, allow the same patterns for destructuring and match expressions. If the latter proves to be too big and to interfere with getting ADTs down the finish line, we can add ad-hoc pattern matching to match expression just for ADTs, but I suspect that wouldn't be much easier than to go for the general approach directly.
- add code to handle enum variants in the implementation of `==`: currently `'Foo..(1) == 'Foo(1)` is false.
- use the normal application syntax instead of the temporary `..( )`. Distinguishing between zero-ary enum tags and 1-ary enum tags would be done at parsing time alone.
- add relevant user documentation

## Maintainability shortcomings

In order to keep the diff reasonable, some visible shortcomings aren't addressed right now but should be explored nonetheless:

- after implementing variable levels #1372, on of the follow-up was:
    >look for code deduplication opportunities. It seems at least some type definitions and generic functions could be shared between record rows variables and enum rows variables.
  This seems even truer after this PR, which makes enum rows and record rows closer in behavior. A lot of the implementation of missing functions for enum rows felt like copy pasting the one for record rows and changing a few names (the biggest difference is the `Option` in enum rows).
- related to the previous point (but still a thing on its own), I wonder if the case of an unapplied enum tag could be better represented morally as an argument of type `Unit`. Ie, `[| 'Foo |]` would be surface syntax for `[| 'Foo Unit |]` and `'Foo` alone would be surface syntax for `'Foo ()`. The type `Unit` doesn't exist currently, but could be added  even only internally, as a special value, if we don't want to create confusion around two different ways of writing the same thing (`'Foo == 'Foo ()`). This might make some of the special casing because of the optional type argument of enum rows go away, and we could use the exact same definition for record rows and enum rows.
- In a similar spirit, should we have only one ast node for enums, whether they're applied to something or not? The proposal just above would say yes, with the argument being `() : Unit` for unapplied enum tags. Another solution, if we don't go the unit route, is to have one AST node with an optional argument. It's hard to say currently if that would be a win or not.
- some type errors around row constraints are fishy. It seems their parameters aren't used consistently across different parts of the typechecker; however this is unrelated to this PR but was just noticed during this work.


core/src/typecheck/eq.rs
```diff
@@ -582,9 +584,34 @@ fn type_eq_bounded<E: TermEnvironment>(
                         && type_eq_bounded(state, t1, env1, t2, env2)
                 }
                 (TypeF::Enum(uty1), TypeF::Enum(uty2)) => {
-                    let rows1 = rows_as_set(uty1);
-                    let rows2 = rows_as_set(uty2);
-                    rows1.is_some() && rows2.is_some() && rows1 == rows2
+                    // let rows1 = rows_as_set(uty1);
+                    // let rows2 = rows_as_set(uty2);
+                    // rows1.is_some() && rows2.is_some() && rows1 == rows2
+                    //
```
yannham, line 590:
```suggestion
````

File: core/src/typecheck/mk_uniftype.rs [hunk 2]
```diff
@@ -19,7 +19,10 @@ macro_rules! mk_uty_arrow {
 }
 
 /// Multi-ary enum row constructor for types implementing `Into<TypeWrapper>`.
-/// `mk_uty_enum_row!(id1, .., idn; tail)` correspond to `<id1, .., idn | tail>.
+/// `mk_uty_enum_row!(id1, .., idn; tail)` correspond to `[| 'id1, .., 'idn; tail |]. With the
+/// addition of algebraic data types (enum variants), individual rows can also take an additional
+/// type parameter, specificed as a tuple: for example, `mk_uty_enum_row!(id1, (id2, ty2); tail)`
+/// is `[| 'id1, 'id2 ty2; tail |]`.
```
yannham, line 25: I need to update the documentation: this macro also accepts tuples `(id, ty)` as individual rows now.

core/src/typecheck/mod.rs
```diff
@@ -1853,6 +1978,19 @@ fn check<V: TypecheckVisitor>(
             ty.unify(mk_uty_enum!(*id; row), state, &ctxt)
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))
         }
+        Term::EnumVariant(id, t) => {
+            let row_tail = state.table.fresh_erows_uvar(ctxt.var_level);
+            let ty_arg = state.table.fresh_type_uvar(ctxt.var_level);
+
+            // We match the expected type against `[| 'id ty_arg, row_tail |]`, where `row_tail` is
```
yannham, line 1985:
```suggestion
            // We match the expected type against `[| 'id ty_arg; row_tail |]`, where `row_tail` is
```

core/src/parser/grammar.lalrpop
```diff
@@ -1011,24 +1024,30 @@ TypeAtom: Type = {
...
     "{" "_" ":" <t: WithPos<Type>> "}" => {
         Type::from(TypeF::Dict {
             type_fields: Box::new(t),
             flavour: DictTypeFlavour::Type
         })
     },
-    // Although a dictionary contracts isn't really a type, we treat it like as
-    // a type for now - at least syntactically - as they are represented using a
+    // Although a dictionary contracts isn't really a type, we treat them as
+    // types for now - at least syntactically - as they are represented using a
```
jneem, line 1044:
```suggestion
    // Although dictionary contracts aren't really types, we treat them as
    // types for now - at least syntactically - as they are represented using a
```

core/src/parser/grammar.lalrpop
```diff
@@ -1011,24 +1024,30 @@ TypeAtom: Type = {
...
     // `TypeF::Dict` constructor in the AST. This just simpler for many reasons
     // (error reporting of contracts and in particular type paths, LSP, and so
     // on.)
     //
     // However, note that we use a fixed type as an argument. This has the
-    // effect of preventing dictionary contracts from catpuring type variables
+    // effect of preventing dictionary contracts from capturing type variables
```
jneem, line 1050: :cat:

core/src/term/mod.rs
```diff
@@ -116,8 +116,12 @@ pub enum Term {
     #[serde(skip)]
     Var(LocIdent),
 
-    /// An enum variant.
+    /// An enum tag, or equivalently, an enum variant without any argument.
     Enum(LocIdent),
+    /// An applied enum variant (an algebraic data type). In Nickel ADTs can have at most one
```
jneem, line 121: If `EnumVariant` will be treated separately from `Enum`, it's *exactly* one variant, right?

core/src/typ.rs
```diff
@@ -78,8 +78,11 @@ pub struct RecordRowF<Ty> {
 /// `EnumRowF` is the same as `EnumRow` and doesn't have any type parameter. We introduce the alias
```
jneem, line 78: Obsolete comment?

jneem (approved):
Exciting! I'd vote for a single ast node with a unit type, especially if that would make it easier to deduplicate the rrows/erows stuff.

core/src/parser/grammar.lalrpop
yannham: Haha, nice substitution indeed

core/src/term/mod.rs
yannham: I've reworded this to be hopefully a bit clearer (it wasn't entirely clear if a bare enum tag is considered an ADT or not; I think it is, so I've reflected that in the latest version)

core/src/typ.rs
yannham: Indeed. Updated. Thanks!

core/src/typecheck/unif.rs
```diff
@@ -1162,18 +1345,59 @@ impl Unify for UnifEnumRows {
...
+                    erows2 @ EnumRowsF::Extend { .. },
+                ) => {
                     let uerows2 = UnifEnumRows::Concrete {
                         erows: erows2,
                         var_levels_data: var_levels2,
                     };
-                    let (_, t2_without_row) = uerows2.remove_row(&id, state, ctxt.var_level)?;
+                    let (ty2_result, t2_without_row) =
+                        //TODO[adts]: it's ugly to create a temporary Option just to please the
+                        //Box/Nobox types, we should find a better signature for remove_row
```
vkleen, line 1375: I may be missing something obvious, but where is the temporary option? Is this comment outdated?

core/src/typ.rs
```diff
@@ -128,13 +134,11 @@ pub enum EnumRowsF<ERows> {
...
     #[default]
     Type,
-    EnumRows,
     /// `excluded` keeps track of which rows appear somewhere alongside the tail, and therefore
     /// cannot appear in the tail. For instance `forall r. { ; r } -> { x : Number ; r }` assumes
-    /// `r` does not already contain an `x` field.
-    RecordRows {
-        excluded: HashSet<Ident>,
-    },
+    EnumRows { excluded: HashSet<Ident> },
```
vkleen, line 139: Does this mean the typechecker will now refuse to instantiate `a` in say `forall a. [| 'x; a |]` with `'x`? Could this have unintended consequences somewhere?

yannham: Indeed, the typechecker won't instantiate `a` with an enum row type that has a _conflicting declaration_ for `a`. Note that, as for record rows, it means that `forall a. [| 'x; a |]` can totally be unified with a type with a `'x` row inside. However, it will refuse to instantiate to a type with e.g. `'x Number`.

In practice, I think we had this code before the row refactoring even for enum rows, and we scrapped it because for enum rows duplication didn't matter: it was a set (without associated type), so conflict couldn't happen.

Now, I _believe_ this change might only be backward-incompatible for row types that were especially written with duplicate rows, such as `[| 'x, 'x |]`, which we should probably forbid anyway. Otherwise, unification should take care of matching existing rows. I can't think of a situation where the typechecker would generate an enum type with a duplicate row by itself.

core/src/typecheck/unif.rs
yannham: Sorry, I think I wrote that most for myself than for the reader :stuck_out_tongue: what I mean is that we have an `Option<Box<T>>`, and that we clone this whole thing and dereference the inner pointer to get `Option<T>`, while I think `remove_erows` might be better off taking an `Option<Box<_>>` to begin with. But I agree the text of the todo is a bit off.

PR #1773, opened 2024-01-23
correctly drop Array::IntoIter

Radvendii:
Since #1683 has been open a while, and I'm not sure when I'm going to have the time to work out what's going on with `Interner`, I wanted to separate out the changes to Array::IntoIter.

If you called `Array::into_iter()`, you could leak elements in two circumstances:
1. If you did not iterate through all elements before dropping, since they had been converted to `ManuallyDrop`
2. If the Array had multiple references when the iterator was created, but the other references were dropped before the iterator was dropped, then the existence of the iterator would prevent those Rcs from dropping their contents, but then because the iterator is of type ManuallyDrop, it wouldn't drop those elements either

### DIFF
--- a/core/src/term/array.rs
+++ b/core/src/term/array.rs
@@ -218,7 +218,7 @@ impl Iterator for IntoIter {
                     Rc::get_mut(inner)
                         .expect("non-unique Rc after checking for uniqueness")
                         .get_mut(self.idx)
-                        .map(|t| ManuallyDrop::take(t))
+                        .map(ManuallyDrop::take)
                 },
             };
 
--- a/core/src/term/array.rs
+++ b/core/src/term/array.rs
@@ -159,11 +159,6 @@ impl IntoIterator for Array {
         // - Move out the elements inside out inner view.
         // - Drop the rest of the elements when we're dropped
         // Otherwise, we clone everything.
-        //
-        // If we start as a shared reference, we could become the only reference
-        // later, but we clone everything anyways, so we don't end up in an
-        // invalid in-between state where some terms have been freed manually
-        // and others haven't.
 
         let inner = if Rc::strong_count(&self.inner) != 1 || Rc::weak_count(&self.inner) != 0 {
             IntoIterInner::Shared(self.inner)
@@ -214,14 +209,14 @@ impl Iterator for IntoIter {
         } else {
             let term = match &mut self.inner {
                 IntoIterInner::Shared(inner) => inner.get(self.idx).cloned(),
-                IntoIterInner::Owned(inner) => unsafe {
-                    Rc::get_mut(inner)
-                        .expect("non-unique Rc after checking for uniqueness")
-                        .get_mut(self.idx)
-                        .map(ManuallyDrop::take)
-                },
+                IntoIterInner::Owned(inner) => Rc::get_mut(inner)
+                    .expect("non-unique Rc after checking for uniqueness")
+                    .get_mut(self.idx)
+                    // SAFETY: We already checcked that we have the only
+                    // reference to inner, and after this we increment idx, and
+                    // we never access elements with indexes less than idx
+                    .map(|t| unsafe { ManuallyDrop::take(t) }),
             };
-
             self.idx += 1;
             term
         }
@@ -234,7 +229,13 @@ impl Drop for IntoIter {
         if let IntoIterInner::Owned(inner) = &mut self.inner {
             let inner = Rc::get_mut(inner).expect("non-unique Rc after checking for uniqueness");
             for term in &mut inner[self.idx..self.end] {
-                unsafe { ManuallyDrop::drop(term) }
+                // SAFETY: Wwe already checked that we have the only reference
+                // to inner, and once we are done dropping the remaining
+                // elements, `inner` will get dropped and there will be no
+                // remaining references.
+                unsafe {
+                    let _ = ManuallyDrop::take(term);
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


core/src/term/array.rs
```diff
@@ -207,6 +228,18 @@ impl Iterator for IntoIter {
     }
 }
 
+impl Drop for IntoIter {
+    fn drop(&mut self) {
+        // drop the rest of the items we haven't iterated through
+        if let IntoIterInner::Owned(inner) = &mut self.inner {
+            let inner = Rc::get_mut(inner).expect("non-unique Rc after checking for uniqueness");
+            for term in &mut inner[self.idx..self.end] {
+                unsafe { ManuallyDrop::drop(term) }
```
yannham, line 237: This unsafe usage should have an associated safety comment.

core/src/term/array.rs
```diff
@@ -191,14 +212,14 @@ impl Iterator for IntoIter {
...
-                Some(slice) => slice[..self.end]
-                    .get_mut(self.idx)
-                    .map(|t| unsafe { ManuallyDrop::take(t) }),
+            let term = match &mut self.inner {
+                IntoIterInner::Shared(inner) => inner.get(self.idx).cloned(),
+                IntoIterInner::Owned(inner) => unsafe {
+                    Rc::get_mut(inner)
+                        .expect("non-unique Rc after checking for uniqueness")
+                        .get_mut(self.idx)
+                        .map(|t| ManuallyDrop::take(t))
```
yannham, line 221:
```suggestion
                        .map(ManuallyDrop::take)
```

yannham (changes requested):
Missing safety comment (but beside other remarks/questions, LGTM)

yannham:
Revisiting this, I think like we're going through a lot of trouble for no obvious gain. IIUC, what this iterator does is to drop stuff live _as it is iterating_ if we're the only reference to the backing array. I guess the original motivation was that you could be the only remaining live reference, but have a very small view into the array (say, the `[10,15]` interval in a 10k elements array).

But, I wonder if we should really care at this point about such optimizations (we can wait for the iterator be consumed or dropped). Besides, I'm more concerned with running time than with memory consumption for now. And it's not obvious (or even probably false in a lot of common cases) than dropping things in many steps is faster than dropping everything at once.

This is why I'm tempted to get rid of this complexity altogether and just a normal iterator. What do you think @Radvendii ?

core/src/term/array.rs
```diff
@@ -157,29 +157,50 @@ impl IntoIterator for Array {
...
-            let idx = self.start;
-            let end = self.end;
-
-            if let Some(slice) = Rc::get_mut(&mut inner) {
-                for term in &mut slice[..idx] {
+        //
+        // If we start as a shared reference, we could become the only reference
+        // later, but we clone everything anyways, so we don't end up in an
+        // invalid in-between state where some terms have been freed manually
+        // and others haven't.
```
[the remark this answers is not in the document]
Radvendii: Yeah, I was saying something else. In retrospect this comment wasn't super clear. This wasn't meant to explain why this code is safe, it's meant to explain why we don't optimize further. In theory if we become the only reference while we're iterating, we could at that point not bother cloning things. This is what the code did before this change. But because this could leave us in an invalid state that is hard to clean up, I think we shouldn't bother with that optimization.

I'll try to reword the comment to be clearer.

And yes, you're right that we don't clone anything at this point in the code, we just mark it as "to be cloned".

Radvendii:
> IIUC, what this iterator does is to drop stuff live as it is iterating if we're the only reference to the backing array.

I thought the point was more that if you're, say, mapping over an Array and you're the only reference, you don't have to clone all the elements, you can modify-in-place. But I'm not super confident in that understanding.

> This is why I'm tempted to get rid of this complexity altogether and just a normal iterator.

Oh, hmm... I'm not opposed. What's the status of testing the real-world speed? Can we run that medium-sized code base someone provided and see if simplifying this to a normal iterator slows it down?

core/src/term/array.rs
Radvendii: Turns out it was this way for a reason (`take` is unsafe), but I think it's clearer to move the `unsafe` block into the lambda anyways.

yannham:
> I thought the point was more that if you're, say, mapping over an Array and you're the only reference, you don't have to clone all the elements, you can modify-in-place. But I'm not super confident in that understanding.

Like this, from the top of my head, I would say that the fact that we can consume the backing memory of the array and reuse the allocation doesn't depend on this code, as this code seems to be concerned with dropping the individual elements or not, not the enclosing array. Given that we have a hand-written iterator, I wonder if there's any chance that `array.into_iter().map(f).collect()` can reuse the inner array - I can't see how the compiler could do so. But I must admit I don't have a good idea of how it works for Rust stdlib structures like `Vec`, so let me investigate a bit. That might actually be a better performance goal (than cloning or not the individual elements).

As for individual elements, it's true that the current code avoid cloning - but honestly, cloning terms is cheap. Unless it prevent some reuse-in-place later, maybe. Though, about the initial motivation, maybe you're right: `ManuallyDrop` might have been used just for `ManuallyDrop::take` in `next()`.

> Oh, hmm... I'm not opposed. What's the status of testing the real-world speed? Can we run that medium-sized code base someone provided and see if simplifying this to a normal iterator slows it down?

Yes, I have a couple sizeable real world codebases, although they have been given to me as private, I can try with and without and see if there's any noticeable difference. With the proviso that variance is always huge - I've always suspected the hashmap randomization and the fact that we stress hashmaps a lot, but have never come to actually test this hypothesis.

Thanks for your input. I think the situation isn't all black-and-white, so while waiting for further investigations on whether the current implementation is worth it, or if we could do better with respect to mutating-in-place, I don't see a good reason to block this PR which strictly improves the current situation.

yannham (approved):

PR #1778, opened 2024-01-24
Fix --field not applying pending contracts

### LINKED ISSUE #1774: Contract attached to record key not getting evaluated
**Describe the bug**
I'm seeing a weird behavior with v1.4.0 (not sure about how far back it goes though), that contracts attaching to a key is not evaluated, only contracts that attach after the value gets evaluated.

**To Reproduce**
Using the sample code from the [documentation](https://nickel-lang.org/user-manual/correctness/#using-a-contract-annotation-1):

```nickel
# sample.ncl

{
    OptLevel = std.contract.from_predicate (fun value =>
      std.array.elem value ["O0", "O1", "O2", "O3"])
    ,
    level = 4
    ,
    opt_level | OptLevel = "A" ++ std.string.from_number level,
    opt_level2 = ("A" ++ std.string.from_number level) | OptLevel,
}
```

```bash
nickel eval sample.ncl --field=opt_level   # evaluates to "A4", which is wrong
nickel eval sample.ncl --field=opt_level2  # contract broken by a value, which is expected
```

**Environment**
 - Version of the code: nickel 1.4.0 (rev 1fa5a79)

**Additional context**

yannham:
Closes #1774

The `--field` argument has been added in 1.3 to act only on a specific field of the configuration. However, pending contracts (contracts attached to the field by metadata annotations or propagated through merging) weren't properly applied, leading to illegal values being accepted.

This commit fixes the issue by adding the missing contract application step. A snapshot test is added as a regression test. While it's not ideal (we don't encode in the test what is the expected result), it's the best we can do currently for CLI-only tests.

I plan to publish a minor release with this patch, as silently not applying contract is serious enough.


vkleen (approved):
This seems like a good quick fix. Long term, I wonder if it wouldn't be better to encapsulate the "extract value out of field" operation so that we'd have a harder time forgetting to do that. Now, there are quite a few places in the code base that need to call `RuntimeContract::apply_all` and the whole setup feels quite fragile.

yannham:
This is a good point. A simple helper first, like `value_with_ctrs(&self) -> Option<RichTerm>` would probably go a long way, at least for deduplication. For correctness, we might end up making the `value` field private, even if general I don't like accessors proliferation when the language provide a built in way to access struct fields, but I guess encapsulation (for correctness) is a valid motivation.

PR #1782, opened 2024-01-26
Improve the FieldResolver api

jneem:
This has some improvements to the `FieldResolver` api in nls.

It turns out that all the public usage of `FieldResolver` only cares about records (some places only care about record terms; some want record terms or record types). This means that while the `FieldResolver` also has to care about arrays and dicts in order to "see through" them in its field resolution, they can be hidden from the public API.

I did experiment a little with also making the distinction between `Record` and `Container` *within* the `field_walker` module. This would be more performant than the current approach of using `Container` internally and filtering out the non-`Record`s at the API boundary. But I couldn't figure out how to do it without either repeated code or fussy indirection through traits. And I doubt the performance matters here...

### DIFF
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -295,7 +295,7 @@ impl<'a> FieldResolver<'a> {
 
     /// Finds all the containers that are descended from `rt` at the given path.
     ///
-    /// The path can mix field access and array "accesses." The array accesses are only used
+    /// The path can mix field access and array "accesses". The array accesses are only used
     /// in array types -- we never actually index an array value -- but they can be used,
     /// for example, to see that `{ foo | Array { bar | { baz | Number } } }` evaluated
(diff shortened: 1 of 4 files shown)


lsp/nls/src/field_walker.rs
```diff
@@ -236,28 +280,44 @@ impl<'a> FieldResolver<'a> {
...
+    /// If this term resolves to one or more records, return them all.
+    pub fn resolve_record(&self, rt: &RichTerm) -> Vec<Record> {
+        filter_records(self.resolve_container(rt))
+    }
+
+    /// Finds all the containers that are descended from `rt` at the given path.
     ///
-    /// Returns all the field-having objects that the final path element refers to.
-    pub fn resolve_term_path(
+    /// The path can mix field access and array "accesses." The array accesses are only used
```
yannham, line 298:
```suggestion
    /// The path can mix field access and array "accesses". The array accesses are only used
```

yannham (approved):

PR #1787, opened 2024-01-29
Implement equality on ADTs

yannham:
Small follow-up of #1770, adding the missing cases to compare enum variants.

Depends on #1770

### DIFF
--- a/core/tests/integration/pass/core/eq.ncl
+++ b/core/tests/integration/pass/core/eq.ncl
@@ -77,5 +77,11 @@ let {check, ..} = import "../lib/assert.ncl" in
         [[1,2,3,4], [1,2,3,4], [1,2,3,4], [1,2,3,4]]
       == false
   ),
+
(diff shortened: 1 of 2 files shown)


core/tests/integration/pass/core/eq.ncl
```diff
@@ -77,5 +77,10 @@ let {check, ..} = import "../lib/assert.ncl" in
         [[1,2,3,4], [1,2,3,4], [1,2,3,4], [1,2,3,4]]
       == false
   ),
+
+  # ADTs
+  'Left..(1+1) == 'Left..(2),
```
jneem, line 82:
```suggestion
  'Left..(1+1) == 'Left..(2),
  'Left..(1) != 'Left(2),
```

(just because the case of equal discriminants but unequal payloads isn't covered yet)

jneem (approved):

PR #1789, opened 2024-01-29
Fix laziness closurization bug, add support for ADTs

### LINKED ISSUE #1788: Some sub-expressions of thunks are re-evaluated several times
**Describe the bug**
During unrelated tests, I noticed that some expressions which should be only evaluated once given the lazy evaluation model of Nickel are actually re-evaluated potentially many time. This bug hits in particular thunks which are in weak head normal form from the beginning, such as arrays or records.

**To Reproduce**

Put a computation inside an array and that array inside a thunk, and request the evaluation of the element several times:

```nickel
nickel> let trace_eval = fun x => std.trace "evaled!" x in let y = [trace_eval null] in [std.array.elem 0 y, std.array.elem 0 y]
std.trace: evaled!
std.trace: evaled!
[ false, false ]
```

**Expected behavior**

I expect that the expression `trace_eval null`, part of the thunk `y`, is only evaluated at most once:

```nickel
nickel> let trace_eval = fun x => std.trace "evaled!" x in let y = [trace_eval null] in [std.array.elem 0 y, std.array.elem 0 y]
std.trace: evaled!
[ false, false ]
```

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

The issue should to be related to the elimination of generated variables, which caused some invariant previously held to become false. In particular, the function `Thunk::should_update`, which decides if a thunk is worth updating, is at fault. A fix is on its way.

yannham:
Depends on #1787  
Closes #1788 

This PR started as a small follow-up of #1770 adding closurization to ADTs, such that in an expression `'Foo exp` (in the ideal final syntax), `exp` is properly shared and not recomputed many times, as for other datastructures (arrays and records).

Testing this unveiled #1788 which is related and rather simple to fix, so this PR does both: fixing other datastructure subexpressions being sometimes unduly re-evaluated, and add missing sharing (and thus avoid re-evaluation) for ADTs.


core/src/eval/mod.rs
```diff
@@ -666,6 +666,21 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                         }
                     }
                 }
+                // Closurize the argument of a enum variant if it's not already done. Usually this is done at the first
+                // time this record is evaluated.
```
yannham, line 670:
```suggestion
                // time the variant is evaluated.
```

jneem (approved):

core/src/eval/mod.rs
```diff
@@ -666,6 +666,21 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                         }
                     }
                 }
+                // Closurize the argument of a enum variant if it's not already done. Usually this is done at the first
```
vkleen, line 669:
```suggestion
                // Closurize the argument of an enum variant if it's not already done. Usually this is done at the first
```

vkleen (approved):

PR #1799, opened 2024-02-02
[Refactor] Pattern matching

yannham:
This PR is preliminary work for the introduction of elimination form for ADTs. It is just a pure refactoring, which shouldn't incur any user-facing change.

This refactoring was motivated by the fact that the existing structure of the codebase used names that weren't very consistent, and relied heavily on the assumption that the only possible patterns are record patterns, which would make the addition of new patterns (ADTs, but not limited to them - constants patterns as well for e.g. strings, array patterns, regular expressions, etc.) impossible.

## Content

- renaming several symbols for better consistency and generality. In particular, `destructuring` is now the submodule `term::pattern` (and same for other destructuring submodules, which are now named `pattern`), and the strange distinction between matches and destructuring is ditched: everything is a pattern.
- restructuring of the representation of a pattern. In particular, this PR introduces a structure that better mimics the actual ADT of a pattern, while the previous one mangled nodes together
- better handling of aliased patterns: make them an optional field instead of a separate variants, which streamline their handling in several part of the code base
- add span information to every subpattern
- factor out operations that apply to several type of patterns to proper traits, instead of declining free-standing function like `do_stuff_record_pattern`, `do_stuff_field_pattern`, etc.
- simplify the implementation of the typechecking of patterns. The previous implementation was building the type of a pattern in a first step, which is always expected to be a record type, and in a second time would extract from this type the type of the bindings brought into scope by the pattern. This extraction was a bit complicated (building a hashmap representation of the row type), and a bit unnecessary. The current implementation builds both the type of the whole pattern and the type of each binding in a single pass, which makes the implementation easier, but also doesn't have to assume that the type of the pattern will be nested record types, which is a prerequisite for extending patterns in the future.
- avoid illegal state with respect to the tail of the pattern (previously `open = false, rest = Some("captured_var")`
- simplify `LetPattern` and `FunPattern` to not take an additional top-level alias as an argument, and simply store this data inside the pattern itself

## Review

Unfortunately, the git diff doesn't seem to understand that `destructuring.rs` was renamed to `term/pattern.rs` (and similarly for `typecheck::destructuring -> typecheck::pattern`, `nickel_lang_lsp::destructuring -> nickel_lang_lsp::pattern`). That being said, the changes are quite heavy, so maybe it's not a bad idea to just review as new code (but take a look at the deleted file to have an idea of what was the previous state).


core/src/transform/desugar_destructuring.rs
```diff
@@ -1,222 +1,203 @@
...
-//! It will first destruct the pattern and create a new var for each field of the pattern.
-//! After that, it will construct a new Record/Array from the extracted fields.
+//! Replace a let-binding with destructuring by a sequence of normal let-binding.
 //!
 //! # Example
 //!
-//! ## The let pattern:
+//! ## Let-binding
+//!
+//! The following destruring let-binding:
```
vkleen, line 9:
```suggestion
//! The following destructuring let-binding:
```

core/src/transform/desugar_destructuring.rs
```diff
@@ -1,222 +1,203 @@
...
 //! let x = {a=1,b=2,c="ignored"} in
 //! let a = x.a in
 //! let d = x.b in
 //! ...
 //! ```
 //!
-//! ## The function pattern
+//! ## Function
+//!
+//! The following desctructuring function:
```
vkleen, line 26:
```suggestion
//! The following destructuring function:
```

core/src/transform/desugar_destructuring.rs
```diff
@@ -1,222 +1,203 @@
...
+        id,
+        RichTerm::new(
+            Term::LetPattern(pat, Term::Var(id).into(), body),
+            // TODO: should we use rt.pos?
+            pos_body,
+        ),
+    )
+}
+
+/// Elaborate a contract from the pattern if it is a record pattern and applies to the value before
```
vkleen, line 83:
```suggestion
/// Elaborate a contract from the pattern if it is a record pattern and apply it to the value before
```

core/src/transform/desugar_destructuring.rs
```diff
@@ -1,222 +1,203 @@
...
-        RecordPattern {
-            matches,
-            open: true,
-            rest: None,
-            ..
-        } => (matches, LocIdent::fresh()),
+impl Desugar for FieldPattern {
+    // For a filed pattern, we assume that the `destr` argument is the whole record being
+    // destructured. We extract the field from `destr` and desugar the rest of the pattern against
+    // `destr.matched_id`.
```
vkleen, line 150:
```suggestion
    // For a field pattern, we assume that the `destr` argument is the whole record being
    // destructured. We extract the field from `destr` and desugar the rest of the pattern against
    // `destr.matched_id`.
```

core/src/transform/desugar_destructuring.rs
```diff
@@ -1,222 +1,203 @@
...
+
+impl Desugar for RecordPattern {
+    fn desugar(self, destr: RichTerm, body: RichTerm) -> Term {
+        let pos = body.pos;
+        // The body is the rest of the term being transformed, which contains the code that uses
+        // the bindings introduced by the pattern. After having extracting all fields from the
+        // value, we potentially need to capture the rest in a variable for patterns with a
+        // capturing tail. For example, `let {foo, bar, ..rest} = destr in body` should be
+        // desugared as `let foo = destr.foo in let bar = destr.bar in let rest = <rest> in body`
+        // (where `<rest>` is some expression removing `foo` and `bar` from `destr`).
```
vkleen, line 165:
```suggestion
        // The body is the rest of the term being transformed, which contains the code that uses
        // the bindings introduced by the pattern. After having extracted all fields from the
        // value, we potentially need to capture the rest in a variable for patterns with a
        // capturing tail. For example, `let {foo, bar, ..rest} = destr in body` should be
        // desugared as `let foo = destr.foo in let bar = destr.bar in let rest = <rest> in body`
        // (where `<rest>` is some expression removing `foo` and `bar` from `destr`).
```

vkleen (approved):

core/src/transform/desugar_destructuring.rs
yannham: It's definitely hard to type, or so it seems :laughing:

PR #1803, opened 2024-02-06
fixed std.array.split_at behavior at right boundary.

### LINKED ISSUE #1802: `std.array.split_at` broken if index equals length of array
**Describe the bug**

The following example from the documentation produces broken contract error on nickel 1.4.1:

```ncl
std.array.split_at 3 [ "Hello", "world", "!" ]

```

```bash
error: contract broken by the caller of `split_at`: invalid array indexing
    ┌─ <stdlib/std.ncl>:573:9
    │
573 │       | std.contract.unstable.IndexedArrayFun
    │         ------------------------------------- expected type
    │
    ┌─ <repl-input-5>:1:22
    │
  1 │ std.array.split_at 3 [ "Hello", "world", "!" ]
    │                      ------------------------- evaluated to this expression
    │
    = Expected an array index between 0 and 2 (included), got 3
```

**Expected behavior**

```ncl
std.array.split_at 3 [ "Hello", "world", "!" ]
  => { left = [ "Hello", "world", "!" ], right = [  ] }
```

**Environment**
 - Version of the code: nickel 1.4.1 (rev 25ea79e)

**Additional context**

The issue seems to stem from [this line](https://github.com/tweag/nickel/blob/6cf2902d3db768618e1d990c549671e308dd3ff4/core/stdlib/std.ncl#L1365). A simple fix would be to just remove the "-1" part however I'm not 100% certain if this fix conforms to the intended semantics of the contract. Happy to submit a PR.

suimong:
fixes #1802 .

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -66,7 +66,7 @@ nickel-lang-lsp 0.1.0
 
 To only build the main crate `nickel-lang-core`, run:
 
-```console
+```shell
 cargo build -p nickel-lang-core
 ```
 
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -578,7 +578,7 @@
 
           # Preconditions
 
-          In `split_at inded value`, `index` must be a positive integer such
+          In `split_at index value`, `index` must be a positive integer such
           that `0 <= index <= std.array.length value`.
 
           # Examples
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1325,6 +1325,7 @@
                 (fun start => RangeSecond start -> Codomain),
 
         IndexedArrayFun
+          | [| 'Index, 'Split |] -> Dyn -> Dyn
           | doc m%"
               **Warning**: this is an unstable item. It might be renamed,
               modified or deleted in any subsequent minor Nickel version.
@@ -1379,15 +1380,13 @@
               else
                 value
             in
-            let contract
-              | [| 'Index, 'Split |] -> Dyn -> Dyn
-              = fun type label function =>
-                DependentFun
-                  ArrayIndexFirst
-                  (fun index => ArrayIndexSecond type index -> Dyn)
-                  label
-                  function
-              in contract,
+            let contract = fun type label function =>
+              DependentFun
+                ArrayIndexFirst
+                (fun index => ArrayIndexSecond type index -> Dyn)
+                label
+                function
+            in contract,
 
         ArraySliceFun
           | doc m%"
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1362,7 +1362,7 @@
               if %typeof% min_size == 'Number && %typeof% value == 'Array then
                 if min_size >= %length% value then
                   let index_as_str = %to_str% min_size in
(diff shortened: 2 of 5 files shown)


vkleen (changes requested):

core/stdlib/std.ncl
```diff
@@ -1362,7 +1362,7 @@
               if %typeof% min_size == 'Number && %typeof% value == 'Array then
                 if min_size >= %length% value then
                   let index_as_str = %to_str% min_size in
-                  let max_as_str = %to_str% (%length% value - 1) in
+                  let max_as_str = %to_str% (%length% value) in
```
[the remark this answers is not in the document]
suimong: Thanks! I followed @yannham 's suggestion and parameterized the `IndexedArrayFun` to accommodate the two functions currently applying this contract.

vkleen (approved):
LGTM. The CI seems to be broken because of an unrelated problem. Thanks a lot for finding and fixing this!

core/stdlib/std.ncl
```diff
@@ -1378,9 +1379,15 @@
               else
                 value
             in
-            DependentFun
-              ArrayIndexFirst
-              (fun index => ArrayIndexSecond index -> Dyn),
+            let contract
+              | [| 'Index, 'Split |] -> Dyn -> Dyn
```
yannham, line 1383: In general we've refrained from contract annotations within the definition of a contract, although this one is rather harmless. That being said, if we want to add a contract, any reason to attach it here, instead of hoisting it up to `IndexedArrayFun`, so that it's visible in the documentation and the blame error tells that `IndexedArrayFun` is the culprit?

HACKING.md
```diff
@@ -66,7 +66,7 @@ nickel-lang-lsp 0.1.0
 
 To only build the main crate `nickel-lang-core`, run:
 
-```console
+```shell
```
yannham, line 69: Any reason for this change? It's not very important, but usually we've used `console` throughout the documentation because IIRC it's more adapted to showing multiple lines when some of them are output and some are commands (an interactive session, basically). In this case I'm not sure it makes any difference, but just for the sake of consistency

suimong: It's just that this is the only code block in Hacking.md that uses `console` (and `shell` everywhere else) and I changed it for consistency. Perhaps I could instead change all `shell` to `console`?

core/stdlib/std.ncl
suimong: My bad, that's not how I usually annotate contract either... moved to the top.

HACKING.md
yannham: Oh, I see. Maybe I've been confusing it with README, and HACKING is older. It's fine then, consistency is better - we can always change everything at once in a later PR.

yannham:
@suimong could you please rebase on latest master? This should solve the (unrelated) CI issue.

suimong:
Not sure why the two tests failed. I'll look into it tomorrow.

vkleen:
I think it's just outdated snapshots for the snapshot tests, you should be able to see them fail locally with `cargo test` and accept the new snapshots with `cargo insta review`.

suimong:
I see. Will do that tomorrow morning.

PR #1805, opened 2024-02-06
[Refactor] Typechecking errors

yannham:
Systematically transform typechecking errors (at least the ones internal to `typecheck`) from multi-parameter variants to variants using an anonymous struct, for self-documenting errors (`FooError(Type, Type, Type, Type)` isn't very clear).

Doing so, this PR also introduces proper distinct variants for enum rows and record rows, cleaning a follow-up TODO of #1770 introducing basic support for ADTs.

## Follow-up

This PR is a pure refactoring, and shouldn't change the behavior of error reporting. However, the error reporting for row conflicts is still slightly broken. I intend to address this point in a follow PR, just to keep this one more focused.

### DIFF
--- a/core/src/typecheck/error.rs
+++ b/core/src/typecheck/error.rs
@@ -319,7 +319,7 @@ impl UnifError {
     }
 
     /// Convert a unification error to a typechecking error, given a populated [name
-    /// registry][reporting::NameReg]. inferred meat of the implementation of
+    /// registry][reporting::NameReg]. Actual meat of the implementation of
     /// [`Self::into_typecheck_err`].
     fn into_typecheck_err_(
         self,


core/src/typecheck/error.rs
```diff
@@ -15,25 +15,58 @@ pub enum RowUnifError {
...
+        /// conflicting row.
+        rrows: UnifRecordRows,
+    },
+    /// A [row constraint][super::RowConstr] was violated.
+    EnumRowConflict {
+        /// The id of the conflicting row.
+        id: LocIdent,
+        /// The type of the new row that conflicts with an existing row.
+        row_type: Option<UnifType>,
+        /// The record rows that were checked for row constraints (and thus in the process of being
```
vkleen, line 54:
```suggestion
        /// The enum rows that were checked for row constraints (and thus in the process of being
```

core/src/typecheck/error.rs
```diff
@@ -82,42 +172,125 @@ impl RowUnifError {
 }
 
 /// Error during the unification of two types.
+///
+/// In each variant, `expected` and `inferred` refers to the two type that failed to unify.
```
vkleen, line 176:
```suggestion
/// In each variant, `expected` and `inferred` refers to the two types that failed to unify.
```

core/src/typecheck/error.rs
```diff
@@ -82,42 +172,125 @@ impl RowUnifError {
...
+        inferred: UnifType,
+    },
+    /// Tried to unify a unification variable with a row type violating the [row
+    /// constraints][super::RowConstr] of the variable.
+    EnumRowConflict {
+        /// The id of the conflicting row.
+        id: LocIdent,
+        /// The type of the new row that conflicts with an existing row.
+        row_type: Option<UnifType>,
+        /// The record rows that were checked for row constraints (and thus in the process of being
```
vkleen, line 258:
```suggestion
        /// The enum rows that were checked for row constraints (and thus in the process of being
```

core/src/typecheck/error.rs
```diff
@@ -146,7 +319,7 @@ impl UnifError {
     }
 
     /// Convert a unification error to a typechecking error, given a populated [name
-    /// registry][reporting::NameReg]. Actual meat of the implementation of
+    /// registry][reporting::NameReg]. inferred meat of the implementation of
```
vkleen, line 322:
```suggestion
    /// registry][reporting::NameReg]. Actual meat of the implementation of
```

vkleen (approved):

core/src/typecheck/error.rs
yannham: Ah, (too) wild search and replace leads to fun results :sweat_smile:

PR #1807, opened 2024-02-07
[WIP] Lazy Imports

vkleen:
Implement a prototype solution for lazy imports. This feature can, for now,
be accessed via the intentionally annoying syntax `import %lazy% "..."`.
Semantically, the intention is to delay any filesystem actions for the import
until evaluation actually demands its value. In particular, the exitence of the
imported file is not checked. For typechecking purposes, a lazy import always
gets assigned type `Dyn`. Apart from these two differences, lazy imports are
supposed to act exctly like normal imports when they get evaluated.

## To be done

- [x] Actually support lazy imports during evaluation
- [ ] Actually support lazy imports during evaluation without panicing on errors
- [ ] Make the import cache abstraction more modular

### DIFF
--- a/core/benches/stdlib.rs
+++ b/core/benches/stdlib.rs
@@ -1,7 +1,7 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 use pprof::criterion::{Output, PProfProfiler};
 
-use nickel_lang_core::cache::{Cache, ErrorTolerance};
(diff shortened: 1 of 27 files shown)


yannham:
Viktor won't have time to dedicate to this anymore, the branch is out of sync and the refactoring of the cache turned out to be harder than initially expected. Not saying that we are abandoning the idea of lazy imports, but for now, let's close this particular attempt.

PR #1811, opened 2024-02-08
Rename LSP action

### LINKED ISSUE #1776: [LSP] Support symbol rename
Using Nickel for a while, it occurs to me that the defining features of Nickel, i.e. recursion, laziness, merge system etc. makes writing program based on "single source of truth" a breeze, meaning that you define static data somewhere **once**, and everywhere else you can refer to that data with nickel symbol. Which brings the topic of this issue: supporting symbol rename in the LSP.

Right now with VSCode, the LSP does nothing when placing the cursor on a symbol and pressing "F2", but imagine if it does, the experience of refactoring nickel code with VSCode or any other LSP-supporting editor, would be comparable to, even better than that of Java with Intellij or Python with PyCharm.

In VSCode at least, pressing 'F2' on a symbol shows the rename dialog box, user enter a new name, VSCode shows preview of all usages of that symbol within the workspace, and do the job.

`ast-grep` + Nickel's tree-sitter grammar is a potentially viable alternative, though I haven't tried it yet.

jneem:
Adds a "rename" action for the LSP. The current behavior is that it renames exactly the same locations that are returned by "references"; I think this is probably the most predictable behavior, but we can tweak it if it turns out not to be.

Fixes #1776

### DIFF
--- a/lsp/lsp-harness/src/lib.rs
+++ b/lsp/lsp-harness/src/lib.rs
@@ -10,10 +10,10 @@ use lsp_types::{
     notification::{Notification, PublishDiagnostics},
     request::{
         Completion, DocumentSymbolRequest, Formatting, GotoDefinition, HoverRequest, References,
-        Request as LspRequest,
+        Rename, Request as LspRequest,
     },
     CompletionParams, DocumentFormattingParams, DocumentSymbolParams, GotoDefinitionParams,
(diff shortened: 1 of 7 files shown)


yannham (approved):
Awesome! It's almost scary that it's smart enough to rename siblings in a merge but it's probably the right semantics (otherwise you wouldn't use rename but just edit the field if you don't want to touch others).

PR #1814, opened 2024-02-09
LSP evaluation in the background

jneem:
This is a basic version of eval-in-the-background for the LSP. It has plenty of deficiencies, but I think it's a step in the right direction.

We spawn one subprocess in the background and keep its state up-to-date with the current files open in the editor. We evaluate files each time they are updated, monitoring the background process and killing it if it goes on too long. If a file times out, we blacklist it and don't try evaluating it again.

Among the many unfinished parts:
- Evaluation is from-scratch every time.
- ~We're just running `eval_full` and reporting the first error.~
- We aren't doing anything smart for incomplete configurations.
- Cross-file diagnostics need more thought. (Currently, we do eval on a file-by-file basis and only report the diagnostics we found in that file.)
- There are basically no tests; the harness doesn't handle diagnostics well.


lsp/nls/Cargo.toml
```diff
@@ -44,6 +44,9 @@ derive_more.workspace = true
 lazy_static.workspace = true
 csv.workspace = true
 thiserror.workspace = true
+crossbeam = "0.8.4"
+ipc-channel = "0.18.0"
+bincode = "1.3.3"
```
yannham, line 49: Any reason to not use the workspace version, as elsewhere, to avoid feature unification causing breakages and to ensure we use a common version if we ever need crossbeam in another crate of the workspace?

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,347 @@
...
+use lsp_types::Url;
+use nickel_lang_core::{
+    cache::SourcePath,
+    eval::{cache::CacheImpl, VirtualMachine},
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{diagnostic::SerializableDiagnostic, files::uri_to_path, world::World};
+
+const EVAL_TIMEOUT: Duration = Duration::from_secs(1);
```
yannham, line 20: I guess experience will tell, but I think one second might be a bit harsh

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,347 @@
...
+        // Deduplicate the evals back-to-front, and then evaluate front-to-back.
+        // This means that if we get requests for `foo.ncl` and `bar.ncl`, and
+        // then another request for `foo.ncl` before started working on the
+        // first one, then we'll throw away the first `foo.ncl` and evaluate
+        // `bar.ncl` first.
+        let mut seen = HashSet::new();
+        let mut dedup = VecDeque::new();
+        for path in evals.iter().rev() {
+            if seen.insert(path) {
+                dedup.push_front(path.clone());
```
yannham, line 99: Nitpick: couldn't we just use a `Vec`, `push` and then iterate in reverse for dedup? As it seems we always push only on one side (front)

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,347 @@
...
+    }
+}
+
+// Returning an Option, just to let us use `?` when channels disconnect.
+fn worker(cmd_rx: IpcReceiver<Command>, response_tx: IpcSender<Response>) -> Option<()> {
+    let mut evals = VecDeque::new();
+    let mut cmds = Vec::new();
+    let mut world = World::default();
+
+    drain_ready(&cmd_rx, &mut cmds);
```
yannham, line 73: Is this one really necessary? I guess if we remove it, the first for loop will just not run at all, and it'll have the same effect at the next loop iteration.

lsp/nls/src/main.rs
```diff
@@ -33,6 +36,9 @@ struct Opt {
     /// The trace output file, disables tracing if not given
     #[arg(short, long)]
     trace: Option<PathBuf>,
+
+    #[arg(long)]
+    main_server: Option<String>,
```
yannham, line 41: It's not user-facing so not very important, but I find it a bit surprising that (IIUC) starting `nls` with `--main-server` means that it just starts a background process to eval stuff. I would rather assume `--main-server` to be the main process.

yannham: (Another solution might be to use [fork](https://docs.rs/fork/latest/fork/))

yannham (commented):
What's the rationale for a separate subprocess vs a separate thread? Is it mostly to avoid having a panic shutting down the rest of the LSP? A subprocess offers some kind of isolation, but communication is also more costly

lsp/nls/src/background.rs
jneem: That's quite possible, and I need to do some more experimentation with big projects. My reason for being conservative here is that we re-evaluate from scratch pretty often (i.e. every time the file contents get updated) and I don't want to be constantly pegging the cpu.

lsp/nls/src/main.rs
jneem: Ah, but `--main-server` takes an argument so the background process gets run with `nls --main-server=foo`...

The fork crate looks nice, but I'm a little worried that it doesn't mention windows at all. I guess at some point I'll need to set up a windows vm to test on...

jneem:
Thanks for the comments! I'll try to get another revision up later this week. The main motivation for a subprocess was to allow cancellation. I'm not too worried about panics, but I am worried about long-running evaluations.

I guess another option would be a "gas"-powered evaluation, like the way equality is tested in the typechecker.

yannham:
> I guess another option would be a "gas"-powered evaluation, like the way equality is tested in the typechecker.

I think that might not prove easy, at least not without baking this in the interpreter (which would be the one that needs to count the gas). At first sight, I think a timeout is better, because time is probably what we want to bound in practice.

> I'm not too worried about panics, but I am worried about long-running evaluations.

FWIW, I found this short post of matklad, which seems to show that clean thread cancellation is not too cumbersome: https://matklad.github.io/2018/03/03/stopping-a-rust-worker.html

jneem:
I think matklad's solution doesn't work if the worker is doing a long-running computation, right? It relies on having a cooperative cancellation point (i.e. the point where the worker tries to receive it's next message).

I looked a bit into terminating threads. The best I found was [this](https://docs.rs/stop-thread/latest/stop_thread/index.html) crate (which uses `pthread_cancel` on linux and `TerminateThread` on windows). [This](https://github.com/rust-lang/unsafe-code-guidelines/issues/211) thread made me think it's a little scary, though.

yannham:
> I think matklad's solution doesn't work if the worker is doing a long-running computation, right? It relies on having a cooperative cancellation point (i.e. the point where the worker tries to receive it's next message).

You're right, I skimmed the post and thought you could just drop the thread object, but he's dropping the `Sender`.

> I looked a bit into terminating threads. The best I found was [this](https://docs.rs/stop-thread/latest/stop_thread/index.html) crate (which uses pthread_cancel on linux and TerminateThread on windows). https://github.com/rust-lang/unsafe-code-guidelines/issues/211 thread made me think it's a little scary, though.

Ok, then maybe a process is fair then.

lsp/nls/src/main.rs
yannham: Ok. Might be worth adding a description then.

jneem:
Ok, I think this is worth another look. I've been test-driving it a little and decided that the false positives on incomplete configurations was too annoying, so instead I've just stopped issuing errors on missing fields.

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,396 @@
...
+        let mut seen = HashSet::new();
+        let mut dedup = Vec::new();
+        for path in evals.iter().rev() {
+            if seen.insert(path) {
+                dedup.push(path.clone());
+            }
+        }
+        evals = dedup;
+
+        if let Some(uri) = evals.pop() {
```
yannham, line 149: What happens here if we have, say, 10 evals inside `evals`? It seems to me that we're going to pop one, and then do another round of loop. Say we receive new commands, so we recompute dedup. Then, L147 (two lines above), we assign `evals` to `dedup`. Won't this erase the `9` remaining evals that were saved from the previous loop? Shouldn't we prepopulate `dedup` with the evals from the previous round or something?

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,396 @@
...
+            cmd_tx,
+            response_rx,
+            banned_files: HashSet::new(),
+            eval_in_progress: None,
+        })
+    }
+
+    fn handle_command(&mut self, msg: Command) -> Result<(), SupervisorError> {
+        if let Command::UpdateFile { uri, text } = &msg {
+            self.contents.insert(uri.clone(), text.clone());
```
yannham, line 255: Should we also remove the file from the banned ones on update? Or maybe a timeout is indeed a better way, or a mix of the two (only remove after timeout + update). But that can wait for another PR, anyway.

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,396 @@
...
+            })?;
+        }
+
+        Ok(())
+    }
+
+    fn run(&mut self) {
+        loop {
+            match self.run_one() {
+                Ok(_) => unreachable!(),
```
yannham, line 337: Why `Ok` is unreachable? Might be worth a small comment (like an `unwap`)

lsp/nls/src/command.rs
```diff
@@ -26,12 +23,13 @@ pub fn handle_command(
 }
 
 fn eval(server: &mut Server, uri: &Url) -> Result<(), Error> {
-    if let Some(file_id) = server.cache.file_id(uri)? {
+    let cache = &mut server.world.cache;
```
yannham, line 26: Nitpick: it seems we only cache once below, and not as a ref (we call `clone()`), so we might just inline its definition as well.

lsp/nls/src/background.rs
yannham: Sorry, you can ignore that remark. We iterate on evals, I misunderstood the code above.

lsp/lsp-harness/src/lib.rs
```diff
@@ -49,7 +50,11 @@ pub enum Request {
 
 #[derive(Deserialize, Debug, Default)]
 pub struct Requests {
-    request: Vec<Request>,
+    request: Option<Vec<Request>>,
```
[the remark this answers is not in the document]
jneem: It affects the derived deserialization. With the `Vec<Request>`, there's a deserialization error if there are no `[[request]]` entries.

lsp/nls/src/background.rs
jneem: I'm mainly worried about working on files that take a long time to evaluate. The editor might update the contents several times per second, and so if we keep on trying to evaluate and timing out then we're just pegging their CPU for nothing.

I agree that banning the file *forever* is probably too aggressive and I haven't thought too carefully about better policies. I think in the longer term, incremental evaluation would solve this problem: even if evaluation is slow, it's making progress building the incremental cache and so future evals will be fast.

yannham (approved):
I initially had a blocking question, but I realized midway that I misunderstood the code and removed it, so in fact it's good to go (of course modulo what you want to make out of the other comments - but nothing blocking).

lsp/nls/src/background.rs
```diff
@@ -0,0 +1,396 @@
...
+                            v.pos,
+                        );
+                        eval_permissive(vm, errors, value_with_ctr);
+                    }
+                }
+            }
+            _ => {}
+        },
+    }
+}
```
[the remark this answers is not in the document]
jneem: I gave that a try and it does seem nicer. My main concern was that this behavior is a bit LSP-specific.

PR #1816, opened 2024-02-13
Record pattern compilation

yannham:
Following the introduction of ADTs, and the generalization of match expression to patterns, this PR implements pattern compilation for record patterns. Compilation means elaborating a basic Nickel expression, which doesn't make use of patterns anymore, to which patterns and match expressions can be compiled (or desugared) to.

Enums are left for follow-up work. In this specific PR, the code isn't actually used anywhere, but I've pushed the branch `tmp/matchv2` which introduces a separate `matchv2` keyword which accepts generic patterns instead of just enum tags, and used that branch to do tests and experiment. Reviewers are invited to try it out, if they wish to.

## Content

- Add a new submodule `compile` to the `term::pattern` module
- Add traits `Compile` and `CompilePart` to be implemented by various pattern AST nodes
- Add `MatchData` data type which will hold the future `Match` node
- Add new required primops to implement pattern compilation: `field_is_defined`, `pattern_branch`,
- Implement compile for record patterns

### Compiling a full match statement

As described in the `compile` module-level documentation, the current approach for a full-fledged match statement is the naive approach: we compile each pattern independently, and try them one by one. We don't expect this to have any impact on performance today. If it does, there are classic references to optimize the generated decision tree.

### Pass bindings to branches

When a branch is selected, we must bring the variable captured by the pattern in scope. The current solution is to build a record along the way - when checking if a pattern match a value. If it does match, the final bindings are added to the environment of the branch by a new special primop introduced here. This is a simple approach that reuses existing Nickel datastructure. We can imagine do something more performant than building the record step by step in the Nickel world: after all, we know in advance the number of pattern variables, and we could pre-allocate memory and fill it imperatively using special primops

### Manual compilation

The current implementation generates all the Nickel expression programmatically. It's a bit clunky, and we could offload the bigger blocks to, say, the internals module of the stdlib, as we do for built-in contracts. However, I believe the current approach is more amenable to more aggressive optimizations, as the compiling code has access to all the bits. It's already probably more performant as well - the blocks offloaded to the internals module would need to be functions, and we would need to apply them, while the current implementation morally inlines everything.

## Follow-up

- Implement compilation for enums
- Switch to a match statement that is able to handle the full patterns thanks to compilation
- Implement static typing of match statement with generic patterns


core/src/error/mod.rs
```diff
@@ -1303,6 +1311,26 @@ impl IntoDiagnostics<FileId> for EvalError {
...
+                        .with_message("this value doesn't match any branch"),
+                );
+
+                vec![Diagnostic::error()
+                    .with_message("non-exhaustive pattern matching")
+                    .with_labels(labels)
+                    .with_notes(vec![
+                        format!("This match expression isn't exhaustive"),
+                        "It has been applied to an argument which doesn't match any of the branches".to_owned(),
+                    ])]
```
jneem, line 1332:
```suggestion
                vec![Diagnostic::error()
                    .with_message("unmatched pattern")
                    .with_labels(labels)
                    ])]
```
I think the notes are redundant given the labels, and I also think "unmatched" is better than "non-exhaustive" here, because a match expression is non-exhaustive if there *exists* a value which isn't matched whereas this runtime error occurs if we've *encountered* a value which isn't matched. (Which of course implies the existence of one, but I think we can use the more specific error message)

core/src/typecheck/operation.rs
```diff
@@ -220,6 +220,17 @@ pub fn get_uop_type(
         // Note that is_variant breaks parametricity, so it can't get a polymorphic type.
         // Dyn -> Bool
         UnaryOp::EnumIsVariant() => (mk_uniftype::dynamic(), mk_uniftype::bool()),
+        // [crate::term::UnaryOp::WithEnv] shouldn't appear anywhere in actual code, because its
```
jneem, line 223:
```suggestion
        // [crate::term::UnaryOp::PatternBranch] shouldn't appear anywhere in actual code, because its
```

core/src/term/pattern/compile.rs
```diff
@@ -0,0 +1,475 @@
...
+//! strategy in the future and is already a more efficient in the current setting (combining
+//! building blocks from the standard library would require much more function applications, while
+//! we can generate inlined versions on-the-fly here).
+use super::*;
+use crate::{
+    mk_app,
+    term::{make, BinaryOp, MatchData, RecordExtKind, RecordOpKind, RichTerm, Term, UnaryOp},
+};
+
+/// Generate a standard `%record_insert` primop as generated by the parser.
```
jneem, line 31:
```suggestion
/// Generate a standard `%record_insert%` primop as generated by the parser.
```

core/src/term/pattern/compile.rs
```diff
@@ -0,0 +1,475 @@
...
+    fn compile(self, value: RichTerm, pos: TermPos) -> RichTerm;
+}
+
+impl Compile for MatchData {
+    // Compilation of a full match expression (code between < and > is Rust code, think of it
+    // as a kind of templating):
+    //
+    // let value_id = value in
+    //
+    // <for (pattern, body) in branches.rev()
```
jneem, line 396: fold?

jneem (approved):

core/src/term/pattern/compile.rs
yannham, line 396:
```suggestion
    // <fold (pattern, body) in branches.rev()
```

PR #1820, opened 2024-02-14
Full pattern matching

yannham:
Depends on #1819.

WIP.


core/src/term/pattern/compile.rs
```diff
@@ -185,26 +185,25 @@ impl CompilePart for RecordPattern {
...
-    //         (%record_insert% <rest>
-    //           final_bindings_id
-    //           (%static_access% <REST_FIELD> final_bindings_id)
-    //         )
-    //   <else if self.tail is open>
-    //     %record_remove% "<REST>" final_bindings_id
-    //   <end if>
+    //    if final_bindings_id == null then
+    //      null
+    //    else
```
yannham, line 190: The interesting part of this change is here: it was missing a null check which could lead at runtime to `%record_remove%` complaining that its argument is null on valid (albeit unmatching) patterns. The rest of the diff in this file is just the implementation of this change.

core/tests/integration/inputs/pattern-matching/basics.ncl
```diff
@@ -0,0 +1,62 @@
...
+
+  'Some..('Tag..(true)) |> match {
+    'None..('Tag..(x)) => !x,
+    'Some..('Other..(x)) => !x,
+    'Some..('Tag..(x)) => x,
+  },
+
+  'Point..({x = 1, y = 2, z = 3}) |> match {
+    'Line..(u) => false,
+    'Plan..({u, v}) => false,
```
vkleen, line 34:
```suggestion
    'Plane..({u, v}) => false,
```

vkleen (approved):

PR #1823, opened 2024-02-16
Support pattern contracts in match statement

yannham:
Depends on #1821.

Match expressions have been extended to accept the full range of patterns (that was reserved to destructuring before). Patterns allow, in particular, to annotate fields with metadata, such as contract annotations or default values. Those were, until now, simply ignored by the new match expressions.

This commit modifies the pattern compilation scheme to handle contracts and default values properly. It follows the current semantics of destructuring (which has been supporting contract in pattern). There is something to be said about default value being merged with the default priority. I'm not sure it aligns well with intuition and other PLs that `{foo | default = 1} |> match { {foo ? 2} => foo } }` currently fails with a merge conflict instead of returning `1`. But at least this PR adds the required machinery. I also think it's mostly forward-compatible with a shift toward a default value that would be used only if the field isn't defined in the original record, as the latter would accept strictly more programs (it's not entirely true, as it would change the meaning of e.g.  `{foo | default = {}} |> match {{foo ? {bar = 1}} => _`, but honestly relying on the fact that the two default values will be merged by the pattern sounds very cursed and twisted...)


core/tests/integration/inputs/pattern-matching/default_value_array_merge_conflict.ncl
```diff
@@ -0,0 +1,7 @@
+# test.type = 'error'
+#
+# [test.metadata]
+# error = 'EvalError::BlameError'
```
vkleen, line 4: Why does this fail with a blame error? Shouldn't it be a merge conflict because of the two default values?

vkleen (changes requested):

vkleen (approved):

core/tests/integration/inputs/pattern-matching/default_value_array_merge_conflict.ncl
yannham: Theoretically you're right, but merge for arrays is implemented as `array1 & array2 ~ std.contract.Equals array1 some_label array2`. So, technically, when that fails, the error is represented internally as a blame error (and probably reported as such, but at least with custom notes? I don't remember)

vkleen: Ah, of course. It is a merge failure but we're checking equality lazily via a contract that fails.

PR #1831, opened 2024-03-01
Relax enum row conflicts

yannham:
The current design of structural ADTs is to clearly separate syntactically unapplied variants (enum tags) from enum variants with an argument, both when constructing them or when deconstructing them (pattern matching).

However, the previous handling of row conflicts would still forbid types like `[| 'Foo, 'Foo Number |]`. Because enum tags and variants are constructed separately, and because they can be matched on separately, this type is actually unproblematic. Thus, this commit relaxes the notion of row conflicts for enums by basically ignoring unapplied enum rows (corresponding to enum tags) within enum row types.

This makes it possible to typecheck expressions such as `match { 'Foo => .., 'Foo x => ..}`, which are indeed entirely fine.

On paper, it seems like we should separate row conflicts in two distinct dimension: enum tags and enum variants. However, enum variants can never cause an actual conflict: `'Foo` is always compatible with another occurrence `'Foo` (and indeed, we didn't bother checking for conflicts just before ADTs where introduced). We thus boldly ignore enum rows that don't have any type argument (corresponding to enum tags) when handling conflicts in this PR.


core/src/typecheck/unif.rs
```diff
@@ -1091,16 +1109,30 @@ impl PropagateConstrs for UnifRecordRows {
...
+            constr: &mut RowConstrs,
             var_id: VarId,
             var_constr: HashSet<Ident>,
             erows: &UnifEnumRows,
         ) -> Result<(), RowUnifError> {
             match erows {
                 UnifEnumRows::Concrete {
-                    erows: EnumRowsF::Extend { row, .. },
+                    // If the row is an enum tag (ie `typ` is `None`), it can't cause any conflict.
+                    // See the documentation of RowConstrs for more details.
```
vkleen, line 1126: Do we want to use a GHC style `[^row-tag-constraints]` for these? The same reference seems to appear a few times.

vkleen (approved):

core/src/typecheck/unif.rs
yannham: I opted for rust-doc style `[RowConstrs]`, even if it's not checked in local comments. The rationale was that the documentation of `RowConstrs` is rust doc-style starting with `///`, and I didn't want to pollute it with a `[^row-tag-constraints]`. Assuming that the rustdoc form `[RowConstrs]` is easily understood and gotoref-able for any Rust developer.

PR #1832, opened 2024-03-01
[Doc] Add patterns to the syntax section of the manual

yannham:
Add a subsection covering patterns, destructuring and match expressions in the syntax section of the user manual. This doesn't cover the typechecking part of patterns, which is left for future work.


yannham:
The "check manual snippets" tests are failing because of issues that should be fixed by #1835 and #1833.

PR #1835, opened 2024-03-05
[Fix] Properly force enum variants

yannham:
The implementation of deep_seq and force primops haven't been properly updated to handle enum variants, letting them mostly unevaluated. This could in particular let contracts be unexpectedly unevaluated, and also shows unevaluated chunks in the REPL when inputting enum variants, which isn't ideal.

This commit properly forces enum variants.

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -707,6 +707,10 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                             env: Environment::new(),
                         })
                     }
+                    Term::EnumVariant { arg, .. } => Ok(Closure {
+                        body: seq_terms(std::iter::once(arg), pos_op),
+                        env,
+                    }),
                     _ => {
                         if let Some((next, ..)) = self.stack.pop_arg(&self.cache) {
                             Ok(next)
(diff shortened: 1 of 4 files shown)


core/tests/integration/inputs/adts/deep_seq_enum.ncl
```diff
@@ -0,0 +1,7 @@
+# test.type = 'error'
+#
+# [test.metadata]
+# error = 'EvalError::BlameError'
+
+# Check that deep_eq correctly evaluate the content of an enum variant
```
vkleen, line 6:
```suggestion
# Check that deep_seq correctly evaluates the content of an enum variant
```

core/tests/integration/inputs/adts/force_enum.ncl
```diff
@@ -0,0 +1,7 @@
+# test.type = 'error'
+#
+# [test.metadata]
+# error = 'EvalError::BlameError'
+
+# Check that force correctly evaluate the content of an enum variant
```
vkleen, line 6:
```suggestion
# Check that force correctly evaluates the content of an enum variant
```

core/src/eval/operation.rs
```diff
@@ -707,6 +707,10 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                             env: Environment::new(),
                         })
                     }
+                    Term::EnumVariant { arg, .. } => Ok(Closure {
+                        body: seq_terms(std::iter::once(arg), pos_op),
+                        env,
+                    }),
```
vkleen, line 713: Just to make sure: enum tags are already in normal form, so we didn't have to worry about them before, right?

vkleen (approved):

core/src/eval/operation.rs
yannham: Right, enum tags are considered weak head normal forms, the same as any compound datastructure.  They are the structural version of a type constructor, and in Haskell for example, type constructors are indeed WHNF.

PR #1836, opened 2024-03-05
Add subsection on enum types in the manual

yannham:
As the title says, add subsection on enum types in the typing section of the manual.

### DIFF
--- a/doc/manual/typing.md
+++ b/doc/manual/typing.md
@@ -220,21 +220,25 @@ The following type constructors are available:
   std.record.map (fun char count => count + 1) occurrences : {_ : Number}
   ```
 
-- **Enum**: ``[| 'tag1, .., 'tagn |]``: an enumeration comprised of alternatives
-  `tag1`, .., `tagn`. An enumeration literal is prefixed with a single quote and
-  serialized as a string. It is useful to encode finite alternatives. The
-  advantage over strings is that the typechecker handles them more finely: it is
-  able to detect incomplete matches, for example.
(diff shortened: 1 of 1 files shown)


doc/manual/typing.md
```diff
@@ -430,13 +434,134 @@ Because the `match` statement has a catch-all case `_`, this function is indeed
...
+  │   ^^^ this expression
+  │
+  = Expected an expression of type `[| 'Foo Number, 'Bar Number |]`, which contains the field `Bar`
+  = Found an expression of type `[| 'Foo Number |]`, which does not contain the field `Bar`
+```
+
+The match expression expects a type `[| 'Foo Number, 'Bar Number |]`, but `foo`
+is of type `[| 'Foo Number |]`. However, `[| 'Foo Number |]` *should be
+compatible* with `[| 'Foo Number, 'Bar Number |]`: if something is an enum `'Foo
+Number`, then it is surely either `'Foo Number` or `'Bar Number`. This
```
vkleen, line 493: Do the backticks around `'Foo Number` render correctly, even though there's a newline?

vkleen (approved):
The snippet testing framework demands changes to `<repl-input-***>` in a few snippets, but I can't add suggestions for it. With that, looks good :+1:

doc/manual/typing.md
yannham: I _think_ so, because it happens often when I use neovim to wrap lines around automatically. But let me double check

PR #1838, opened 2024-03-07
Rework format.rs to new Topiary API

ErinvanderVeen:
TODO! Wasm-bindgen issue needs to be solved. Which is a pain-point for both Nickel and Topiary alike. We just happen to pin different versions for the same reason.


yannham:
I've taken over this PR. The wasm-bindgen version is updated to the latest one from Nixpkgs/unstable, which is the one used by Topiary as well.

core/src/format.rs
```diff
@@ -19,19 +19,21 @@ impl Display for FormatError {
...
         &mut input,
         &mut output,
-        &query,
-        language,
-        &grammar,
-        topiary::Operation::Format {
-            skip_idempotence: true,
+        &language,
+        Operation::Format {
+            skip_idempotence: false,
```
vkleen, line 36: Is there guidance for when idempotence should be checked in a downstream usecase like here? I think I originally put it to "skip" because I assumed it's mostly a test for topiary itself.

yannham: This is my understanding too. I wouldn't like to make users unable to format their codebase because Topiary happens to fail the idempotence checks. On the other hand, it's useful data for debugging Topiary. Maybe we should enable it with the debug profile?

ErinvanderVeen: Heeeej!

The idempotency check takes up but a tiny fraction of Topiary's runtime, which is why we generally enable it by default.

But yes, in theory it should be fine to disable it.

yannham: I think I don't mind the extra time, but I fear user can't format their code if they run into an idempotency bug. Or maybe we can catch the error and just emit a warning kindly asking to report the issue?

yannham:
@vkleen I've disabled the idempotency check in release mode, as after checking with @ErinvanderVeen, an idempotency failure would mean Nickel would refuse to format the codebase, which is unacceptable. We could try first with idempotency, and then re-format without if the check fails, but this starts to be complicated. Let me know if that sounds good to you and if we can proceed with the PR in the current state

Issue #1534, opened 2023-08-17
Non-ASCII identifier support


toastal:
**Is your feature request related to a problem? Please describe.**

English is a weird language. It was the basis of ASCII, but many languages&mdash;even ones also using the Latin script don’t fit inside its limited character set. As a result, there is a bias towards Latin characters `[A-Za-z]` without accents. Since there doesn’t appear to be a bicameral distinction requirement, all writing scripts can & probably should be a considered valid for a modern language that doesn’t have the legacy bias of older languages. As such I get `unexpected token` errors for situations that _feel_ like they should be valid. Consider:

```nickel
let Pokémon = {
	ID | std.number.PosNat,
	name | String,
	# …
} in

let SomeNorseGods = [| 'Odin, 'Freyr, 'Freyja, 'Þórr, 'Loki, 'Höðr, 'Sága |] in

let SomeGreekGods = [| 'Ἀφροδίτη, 'Ἀπόλλων, 'Ἄρης, 'Περσεφόνη |]

let Buds = {
	คิว = { },
	แชมป์ = { },
	เมฆ = { },
} in

{ }
```

This gets `unexpected token` errors despite being valid (according to humans) writing scripts.

**Describe the solution you'd like**

If it’s a ‘letter’ in a writing system block, it’s valid. I understand errors for names with spaces or ‘symbol’ but all writing systems should be valid.

**Describe alternatives you've considered**

* ‘Romanize’ everything (tho this can lead to errors as many languages distinguish between ‘e’ & ‘é’) & [deburr](https://lodash.com/docs#deburr).
* Convert everything to English since English tends to remove all accents since English’s writing system is already a mess & since words aren’t phonemic, its speakers are used to memorizing weird or misspelled borrowings from other languages (tho exceptions where words like naïve & façade & jalapeño are often spelled with their accents which would still fail).
 
**Additional context**

vkleen:
For reference, we could and maybe should follow [Unicode Standard Annex #31](https://www.unicode.org/reports/tr31/tr31-37.html) like Rust itself is for identifiers.

jneem:
Just to note that this was discussed in the [weekly meeting](https://docs.google.com/document/d/1MlJdr-uqhHIY1Fe-Vqi4d9gnwMgKRiLqIvbBFzR1xDY/edit#heading=h.v7ik9i3ta1or), and it seems easily doable (the hardest part seems to be the vim highlighting).

toastal:
[...]

PR #1839, opened 2024-03-07
Improve the description of identifier syntax in the manual

### LINKED ISSUE #1837: Typo on the homepage (underscore and dash)
I quote

> Nickel identifiers start with an alphabetic character, followed by zero or more alphanumeric characters, _ (underscores) or ' (single quotes). For example, this-isn't-invalid is a valid identifier.

Underscores are allowed, and you proof that by giving an example with dashes. 

I guess either dashes are also ok, and you forgot that to mention, or you made a mistake in that last sentence. 

https://nickel-lang.org/user-manual/syntax

vkleen:
Explicitly mention the validity of dashes and leading underscores.

Fixes #1837

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -6,9 +6,11 @@ slug: syntax
 
 ## Identifiers
 
-Nickel identifiers start with an alphabetic character, followed by zero or more
-alphanumeric characters, `_` (underscores) or `'` (single quotes). For example,
-`this-isn't-invalid` is a valid identifier.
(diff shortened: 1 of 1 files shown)


yannham (approved):
By the way, I wonder if we have any good reason to not allow several heading underscores. It's probably an oversight

vkleen:
I don't think there's any reason, either. We snould probably just tackle that as part of #1534 .

ShalokShalom:
And once that is tackled, also bring it back into the documentation ;)

PR #1841, opened 2024-03-08
Add manual section on modular configurations

yannham:
This PR imports the Tweag blog post about the Nickel merge system (https://www.tweag.io/blog/2023-11-02-nickel-merge-system/) into the manual almost verbatim (beside a few trivial changes such as "post" -> "section"). This blog post explores how to use the merging system to write modular configurations, which isn't covered currently in the manual.


yannham:
I'm waiting for #1843, which might be required to make the CI pass.

yannham:
This is all good now, ready for review and CI is green @jneem @vkleen

doc/manual/introduction.md
```diff
@@ -74,18 +74,16 @@ website.
 
 ## Current state and roadmap
 
-Nickel is currently released in version `1.3`. We expect the core design of the
-language to be stable and the language to be useful for real-world applications.
-The next steps we plan to work on are:
+Nickel is currently released in version `1.4`. The core design of the language
```
vkleen, line 77: Did we forget updating that, again? Or did this change sneak in from somewhere elese?

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
+---
+slug: modular-configurations
+---
+
+# Modular configurations
+
+This section explains how to leverage the Nickel languages features, and in
```
vkleen, line 7:
```suggestion
This section explains how to leverage Nickel's features, and in
```

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
...
+  client1 = machine {ip = "0.0.0.0", open_ports = [22]},
+  client2 = machine {ip = "1.1.1.1", open_ports = [80]},
+}
+```
+
+In the zoology of programming abstractions, functions are the simplest and most
+ubiquitous concept of all. For the machine example, a ([pure][pure-function])
+function is exactly what we look for: an expression with unknown parameters
+which can be selected at usage time. Virtually every programming language has
+functions in some form, even if they're not called that way (for example, in
```
vkleen, line 41:
```suggestion
functions in some form, even if they're not called that (for example, in
```

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
...
+Terraform's configuration language, _modules_ are more or less
+functions[^terraform-modules]).
+
+You might even want not to hardcode some parameters directly in the
```
vkleen, line 45:
```suggestion
You might even want to not hardcode some parameters directly in the
```

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
...
+  pkgs.wasm-bindgen-cli // {version = wasm_bindgen_fixed_version;}
+```
+
+Sadly, this doesn't work. This code overrides the final `version` field
+correctly, but not the other values which depend on version (such as `src`).
+What we want to override is the _original_ `version` used when calling the
+package-as-a-function, not the one propagated in the end result. But once an
+arbitrary function has been applied, there's no coming back.
+
+The [actual Nix code][wasm-bindgen-override] to build Nickel calls to a
```
vkleen, line 235:
```suggestion
The [actual Nix code][wasm-bindgen-override] to build Nickel calls a
```

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
...
+apparent: there are just some holes to be filled.
+
+### Records remember
+
+As opposed to functions, recursive records remember the relation between their
+different fields even if they have all been given a value at this point.
+Overriding a field will automatically recompute the new value of its reverse
+dependencies. Remember the `wasm-bindgen` example, where we had to use Nixpkgs'
+`override`. In Nickel, you don't need any of that. The following is a similar
+example, albeit simplified, where the `cmd` field depends on the the `ip` field
```
vkleen, line 305:
```suggestion
example, albeit simplified, where the `cmd` field depends on the `ip` field
```

doc/manual/modular-configurations.md
```diff
@@ -0,0 +1,429 @@
...
+the current merge system can already get you pretty far in an intuitive way.
+
+## Conclusion
+
+In this section, we explored the merge system of Nickel and how to leverage it
+to write natural configuration parts that are easy to query, to combine and to
+tweak. We supported our claim that partial configuration is a better suited
+model than functions (or other abstractions, for that matter) for configuration
+management, and showed that the combination of merging and field metadata is an
+expressive combo.
```
vkleen, line 419: I feel like we should remove this conclusion. It doesn't really fit the style of the other manual sections.

vkleen (approved):

doc/manual/modular-configurations.md
yannham: Fair enough. I wondered about it too, because we have "take-away" subsections in some sections (typing I believe), but I agree it doesn't really fit.

doc/manual/modular-configurations.md
yannham: I think both are valid, although the emphasize is a bit different, right?

doc/manual/modular-configurations.md
yannham, line 419:
```suggestion
```

doc/manual/modular-configurations.md
yannham: I don't have a very strong opinion, but this part is verbatim from the blog which went under the review of the editing team. Is it a stylistic issue, or a correctness? If the former, I think I'll stick to the published version. If the latter, let me know, because then I need to update the blog post as well.

vkleen: It feels to me like "called that way" is about how something gets called, e.g. by phone. But I'm not a native speaker, so take that with a grain of salt. I don't think anyone will care, though :)

doc/manual/modular-configurations.md
vkleen: I'm not sure; "not want to" or "want to not" would feel right to me and have this difference in emphasis. But again, not a native speaker...

PR #1847, opened 2024-03-11
Remove some spurious diagnostics from background eval

jneem:
The first pass at background eval intended to ignore missing field errors, but it didn't do a thorough job. Specifically, it ignored missing field errors that came from forcing undefined fields. But it kept missing field errors coming from anything that evaluated those undefined fields. This PR removes those errors, and also some spurious "infinite recursion" errors coming from not resetting the thunks on error.

I decided to make `eval_permissive` return all the errors it finds, and filter out the missing field errors in nls. This is a little less efficient than filtering them out in `eval_permissive`, but it felt more principled. (Also, if we ever figure out a way to guess when a record is intended for export, maybe we won't always filter them out.)

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -938,12 +938,10 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
 
     /// Evaluate a term, but attempt to continue on errors.
     ///
-    /// This differs from `VirtualMachine::eval_full` in 3 ways:
+    /// This differs from `VirtualMachine::eval_full` in 2 ways:
     /// - We try to accumulate errors instead of bailing out. When recursing into record
     ///   fields and array elements, we keep evaluating subsequent elements even if one
     ///   fails.
-    /// - We ignore missing field errors. It would be nice not to ignore them, but it's hard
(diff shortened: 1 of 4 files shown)


core/src/eval/mod.rs
```diff
@@ -938,21 +938,23 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
             slf: &mut VirtualMachine<R, C>,
             acc: &mut Vec<EvalError>,
             rt: RichTerm,
         ) {
+            let pos = rt.pos;
             match slf.eval(rt) {
-                Err(e) => acc.push(e),
+                Err(e) => {
+                    acc.push(e);
+                    slf.stack.reset(&mut slf.cache);
```
yannham, line 956: I think `VirtualMachine` has a reset function. Mostly just call `stack.reset()` and `call_stack.clear()`, but you don't need to access inner fields of the virtual machine.

yannham (approved):
LGTM. Initially I thought we would take the converse approach: have a whitelist of errors that make sense to report in an incomplete program (and the first candidate would be to start with contract errors). But I guess most of errors shouldn't happen (beside parse errors), so we can do the blacklist approach as well. We'll figure it out with experience anyway

PR #1853, opened 2024-03-13
1.5.0 release

yannham:
Backport 1.5 update and small fixes made along the way to master.

### DIFF
--- a/doc/manual/introduction.md
+++ b/doc/manual/introduction.md
@@ -74,9 +74,9 @@ website.
 
 ## Current state and roadmap
 
-Nickel is currently released in version `1.4`. The core design of the language
-is stable and Nickel is useful for real-world applications. The next steps we
-plan to work on are:
+Since version 1.0 released in May 2023, the core design of the language is
+stable and Nickel is useful for real-world applications. The next steps we plan
+to work on are:
 
 - Custom merge functions (second part of the
   [overriding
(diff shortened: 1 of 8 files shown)


doc/manual/introduction.md
```diff
@@ -74,9 +74,9 @@ website.
 
 ## Current state and roadmap
 
-Nickel is currently released in version `1.4`. The core design of the language
```
yannham, line 77: To avoid getting out of sync every version, I've just reworded this part to not mention the current version. I think the main message to convey is that Nickel is post 1.0 and that you should use it.

lsp/nls/src/requests/formatting_external.rs
```diff
@@ -17,8 +17,8 @@ pub fn handle_format_document(
     server: &mut Server,
 ) -> Result<(), ResponseError> {
     let path = uri_to_path(&params.text_document.uri)?;
-    let file_id = server.cache.id_of(&SourcePath::Path(path)).unwrap();
-    let text = server.cache.files().source(file_id).clone();
+    let file_id = server.world.cache.id_of(&SourcePath::Path(path)).unwrap();
```
yannham, line 20: those build failures were only revealed when trying to build without the format feature, which doesn't happen outside of crates.io releases.

jneem (approved):

PR #1854, opened 2024-03-13
Propagate pending array contracts in permissive_eval

jneem:
I don't think this had an issue, but it was mentioned in the discord...

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -956,12 +956,17 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     slf.reset();
                 }
                 Ok(t) => match t.as_ref() {
-                    Term::Array(ts, _) => {
+                    Term::Array(ts, attrs) => {
                         for t in ts.iter() {
                             // After eval_closure, all the array elements  are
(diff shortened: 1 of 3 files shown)


core/src/eval/mod.rs
```diff
@@ -956,12 +956,17 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
                 }
                 Ok(t) => match t.as_ref() {
-                    Term::Array(ts, _) => {
+                    Term::Array(ts, attrs) => {
                         for t in ts.iter() {
                             // After eval_closure, all the array elements  are
                             // closurized already, so we don't need to do any tracking
                             // of the env.
-                            inner(slf, acc, t.clone());
+                            let value_with_ctr = RuntimeContract::apply_all(
```
yannham, line 964: Oh, so we did evaluate arrays, but just didn't apply lazy contracts?

yannham (approved):

core/src/eval/mod.rs
jneem: Yep, so `[ "a" | Number ]` was working before, but not `["a"] | Array Number`

PR #1862, opened 2024-03-22
Extend merge to enum variants

yannham:
Enum variants (ADTs) were introduced in 1.5, but they currently can't be merged, which breaks the idempotency of merge on values. This commit extend merging to enum variants, by merging their content recursively if their tags are equal, or failing otherwise.

## Semantics

Two different semantics have been discussed for enum variants:

1. As for other primitive values, only merge variants that are strictly equal, including their arguments.
2. Merge the arguments of variants if they have the same tag, that is `'Tag arg1 & 'Tag arg2 ~ 'Tag (arg1 & arg2)`.

The current statuos quo is lending toward 1., as this policy has been applied to other primitive values. However, there's no other reasonable semantics for constants (booleans, strings and numbers). So we should really compare variants to other datastructures, arrays and functions. For those two, we decided not to merge them to avoid silent and unwanted merging to happen (and because, for arrays, there are multiple possible reasonable semantics).

However, given the feedback of some users, it seems that merging enum variants with similar tags is both useful and natural. The risk of unwanted merging also seems to be limited, as when the tag differs, or the variant's argument is not a record (or only composed of variants of records), the merge will fail anyway. After consideration, we decided to go with semantics 2.


core/src/error/mod.rs
```diff
@@ -1201,18 +1201,93 @@ impl IntoDiagnostics<FileId> for EvalError {
...
+                                while the other is of type {right_ty}"
+                            ));
+                            notes.push("Values of different types can't be merged".to_owned());
+                        }
+                        "String" | "Number" | "Bool" | "Array" | "EnumTag" => {
+                            push_merge_note(&mut notes, &left_ty);
+                        }
+                        "Function" | "MatchExpression" => {
+                            notes.push(
+                                "Both values are functions (including match expressions)"
```
jneem, line 1233:
```suggestion
                                "Both values are functions (or match expressions)"
```

doc/manual/merging.md
```diff
@@ -160,8 +160,11 @@ Evaluates to the record:
 When one or both of the common fields are not records, the merge will fail
 unless one of the following condition hold:
 
-- They are both of the same primitive data type (`Number`, `Bool`, `Enum`, or
+- They are both of the same primitive data type (`Number`, `Bool`, `EnumTag`, or
   `String`) and they are equal
+- They are both enum variants, and their tags are equal. In this case, the
+  arguments the variants are merged recursively: that is,
```
jneem, line 166:
```suggestion
  arguments are merged recursively: that is,
```

jneem (approved):

PR #1863, opened 2024-03-22
Update release infrastructure NixOS configuration

vkleen:
Upstream has changed the configuration schema for GitHub runners in
NixOS. Update our configuration to follow suit.

The change is deployed and building Nickel 1.5 artifacts right now.

### DIFF
--- a/infra/github-runner.nix
+++ b/infra/github-runner.nix
@@ -1,4 +1,4 @@
-### https://nixos.org/channels/nixos-unstable nixos
+### https://nixos.org/channels/nixos-23.11 nixos
 { pkgs, lib, ... }:
 let
   runner-name = "ec2-spot";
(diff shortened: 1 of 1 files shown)


infra/github-runner.nix
```diff
@@ -7,10 +7,9 @@ in
   imports = [ <nixpkgs/nixos/modules/virtualisation/amazon-image.nix> ];
   ec2.hvm = true;
 
-  services.github-runner = {
+  services.github-runners."$${runner-name}" = {
```
yannham, line 10: Oh, is that a double-level interpolation? :smiling_face_with_tear:

yannham (approved):

infra/github-runner.nix
vkleen: We heard you like interpolating, so here's how to interpolate while you interpolate... No, this is actually `$$` escaping the terraform string interpolation, because nix and terraform syntax conflicts. This wouldn't have happened if we could write NixOS configurations in Nickel :stuck_out_tongue:

PR #1868, opened 2024-03-25
Add a GitHub workflow for testing the release infrastructure

vkleen:
The `release-infra-test.yaml` workflow can be called manually to build
`nixpkgs#hello` on the ARM64 runner. The plan is to use it as a test
while making a release to catch problems with the infrastructure early.

TODO: this doesn't work yet, because the infrastructure is only allowed to be run on release tags. I'll look into a good way to address this later.

### DIFF
--- a/.github/workflows/release-infra-test.yaml
+++ b/.github/workflows/release-infra-test.yaml
@@ -1,0 +1,97 @@
+name: Test the ARM64 runner infrastructure
+on:
+  workflow_dispatch:
+
+# TODO: ideally this and `release-artifacts.yaml` should depend on a runner workflow, so that we can't have the two setups diverge. For now, this copyint is quick and good enough.
+
+permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-infra-test.yaml
```diff
@@ -0,0 +1,98 @@
+name: Test the ARM64 runner infrastructure
+on:
+  pull_request: # TODO: make absolutely sure to remove this before merging
```
yannham, line 3: I'm adding a review comment and request changes so that we don't forget to do so.

yannham (changes requested):
LGTM, beside the thing we must not forget to remove, which is why I'm requesting changes.

yannham:
I'm going to close this, since we shifted away from AWS and now use GH native runners for release artifacts.

PR #1869, opened 2024-03-25
Don't leak memory in background eval

jneem:
This fixes a memory leak in background evaluation, by running each background eval job in a separate process. This might be a little extreme: there's a middle ground where we run several eval jobs in a single process before killing it to reclaim the memory. This extreme approach has the advantage of being simpler and not having an additional knob to tune.


lsp/nls/src/background.rs
```diff
@@ -15,305 +14,236 @@ use nickel_lang_core::{
...
 
 struct SupervisorState {
     cmd_rx: Receiver<Command>,
     response_tx: Sender<Diagnostics>,
-    child: Child,
-    cmd_tx: IpcSender<Command>,
-    response_rx: Receiver<Response>,
 
     contents: HashMap<Url, String>,
+    deps: HashMap<Url, Vec<Url>>,
```
yannham, line 128: Is this related to running the process in the background? Why do we need to keep deps now?

yannham (commented):
If I understand correctly, does this mean we don't dedup anymore, and just blindly spawn a new process as soon as we have a notification? Is that at risk of spawning many many processes during live editing?

lsp/nls/src/background.rs
jneem: Because we spin up a new process for each job, we need to transfer over all the file contents (because the LSP's in-memory contents might be different to what's on disk). The dependency tracking allows us to send over only the file contents that the current eval depends on.

yannham (approved):
I see, thanks for the precision. The approach looks good to me: while there's space for fine-tuning, even if we waste a bit of resources spanning a lot of processes, as long as it doesn't happen in the main LSP thread, it's reasonable.

PR #1872, opened 2024-03-27
[Fix] Polymorphic field typechecking

### LINKED ISSUE #1690: Typechecking issues with polymorphic types annotations on record fields
**Describe the bug**

Some record fields with polymorphic type annotations don't typecheck properly: programs are unduly rejected and polymorphic types are being unduly monomorphized. Moving the type annotation from the record field to the record value (that is `{foo : T = value} ~> {foo = (value : T)}`) makes it work, while both should have the same behavior.

**To Reproduce**

```console
$ nickel repl
nickel> { id : forall a. a -> a = std.function.id } : _             
error: incompatible types
  ┌─ <stdin>:1:27
  │
1 │ { id : forall a. a -> a = std.function.id } : _
  │                           ^^^^^^^^^^^^^^^ this expression
  │
  = Expected an expression of type `forall a. a -> a`
  = Found an expression of type `a -> a`

nickel> { id = (std.function.id : forall a. a -> a) } : _
{ id = fun x => x, }
```

**Expected behavior**
I expect the first program to typecheck and return the same result as the second program.

**Environment**
 - Version of the code: 2a727baa

yannham:
Closes #1690 

Using a polymorphic type annotation on a record field, as in `{id : forall a. a -> a = std.function.id}` would unexpectedly fail with a failure to unify `forall a. a -> a` with `a -> a`, which indicates that the type annotation wasn't properly instantiated before being used to check the value of the field.

It turns out the typechecking of the field per se is fine, but the issue comes from the recursive environment: in a recursive record, we need to build an environment with the types of all record fields _before_ we can start to actually typecheck each field. When the field isn't annotated, we use a fresh unification variable. When typechecking the corresponding field, we unify what's in the type environment (the unification variable, but which might have been unified with another type at this point) with the type inferred for the field. Doing so, we reject things like `{a = true, b = a + 1} : _`.

However, when the field has a type annotation, we use this annotation for the recursive environment instead (which is important e.g. to keep polymorphic types...polymorphic). In this case, the additional unification step described above would unify a polymorphic type with the inferred type, which is the place that was missing the instantiation.

One possibility is to just add the missing instantiation, but in fact this unification is useless when there is a type annotation, as the typechecking of the field with a polymorphic annotation will perform the same work anyway (infer a type and check that it's a subtype of the instantiated annotation). We thus rather remember which fields are annotated and we entirely skip the unification in this case.

### DIFF
--- a/core/src/typecheck/mod.rs
+++ b/core/src/typecheck/mod.rs
@@ -1789,8 +1789,8 @@ fn check<V: TypecheckVisitor>(
     // When checking against a polymorphic type, we immediatly instantiate potential heading
     // foralls. Otherwise, this polymorphic type wouldn't unify much with other types. If we infer
     // a polymorphic type for `rt`, the subsumption rule will take care of instantiating this type
-    // with unification variables, such that terms like
-    // `(fun x => x : forall a. a -> a) : forall b. b -> b` typecheck correctly.
+    // with unification variables, such that terms like `(fun x => x : forall a. a -> a) : forall
+    // b. b -> b` typecheck correctly.
     let ty = instantiate_foralls(state, &mut ctxt, ty, ForallInst::Constant);
 
     match t.as_ref() {
@@ -2115,7 +2115,6 @@ fn check<V: TypecheckVisitor>(
             ty.unify(mk_uniftype::dict(ty_dict.clone()), state, &ctxt)
                 .map_err(|err| err.into_typecheck_err(state, rt.pos))?;
 
-            //TODO: should we insert in the environment the checked type, or the actual type?
             for id in record.fields.keys() {
                 ctxt.type_env.insert(id.ident(), ty_dict.clone());
                 visitor.visit_ident(id, ty_dict.clone())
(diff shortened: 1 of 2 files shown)


core/src/typecheck/mod.rs
```diff
@@ -2857,7 +2900,8 @@ fn instantiate_foralls(
         ..
     } = ty
     {
-        let kind = (&var_kind).into();
+        let kind: VarKindDiscriminant = (&var_kind).into();
```
yannham, line 2903: Cosmetic change: I find it better to put an explicit type annotation when there's an `into()` that isn't used right away.

core/src/typecheck/mod.rs
```diff
@@ -2846,7 +2878,18 @@ fn instantiate_foralls(
...
+    // really introduce a new block of rigid type variables.
+    if matches!(
+        ty,
+        UnifType::Concrete {
+            typ: TypeF::Forall { .. },
+            ..
+        }
+    ) {
+        ctxt.var_level.incr();
+    }
```
yannham, line 2892: Unrelated small changes that was on my radar for some time. It doesn't change anything with respect to correctness (the typechecker was already correct before), but would just introduce more levels that necessary.

jneem (approved):

PR #1876, opened 2024-03-28
Add empty optional fields-aware record operator variants

yannham:
The design guideline around empty optional fields has been to make most normal record operation ignore them, to avoid issues like trying to list the field of a record and access them just to fail on an empty optional introduced by a contract, which we shouldn't care about.

However, in some cases (typically operating on record contracts), users might want to actually take empty optional fields into account. Most primops were already existing in two variants internally (ignore empty opts and consider all fields). This PR just makes them available as clearly documented stdlib functions, and update the documentation of the existing record functions to make their behavior with respect to empty optionals more explicit.

## Naming

I use the `_all` prefix for now, but I don't find it optimal. `insert_all` and `remove_all` looks like they're taking several fields to insert and remove. I'm open to suggestions. An alternative could be `extd` for extended. It's not very telling, but at least doesn't set up false expectations.


jneem (approved):
LGTM. I think `with_opts` was the suffix we settled on in the weekly.

yannham:
During the renaming `_all` -> `_with_opts`, I realized I forgot to add `has_field_with_opts` as well.

PR #1879, opened 2024-04-03
Markdown documentation generation: do not insert line breaks

### LINKED ISSUE #1706: The markdown `nickel doc` output prints some spurious backslashes
**Describe the bug**

`nickel doc`'s markdown output contains some backlashes that mess up with the rendering. For instance, `{ x : Number = 1, y | doc "Blah" }` will be rendered as

```markdown
# `x`

- `x : Number`
\
# `y`

\
Blah
```

`cmark` as well as GitHub's markdown renderer and `pandoc --from=commonmark` will show (some of) these backslashes literally:

image

Pandoc's own markdown format won't render them but interpret them as escaping the start of the next line:

image

**To Reproduce**

```console
$ nickel doc --stdout <<<'{ x : Number = 1, y | doc "Blah" }' | cmark
```

**Expected behavior**

An output that is at least commonmark-compatible (and if possible compatible with other common markdown dialects like gfm or pandoc)

**Environment**
 - OS name + version:
 - Version of the code: 28672ee45bfe78dc3dc828ac1fe4fa40f99965e9

**Additional context**

yannham:
Fixes #1706.

The code generating markdown documentation from a Nickel configuration used to unconditionally insert hard line breaks between a field's signature and the rest of its documentation. However, one can't insert a hard line break immediately followed by a blocks in commonmark (or put differently: it's not representable as a source), and those line breaks were just rendered as spurious backslashes at best, or sometimes interpreted differently (see #1706 for more details).

A bit of experimentation showed that what comes after this line break is always a block in the sense of commonmark (even when it's simple documentation text, it's wrapped as a Markdown paragraph). Thus this PR just removes the line break altogether, which is not useful and never legal anyway.

### DIFF
--- a/cli/tests/snapshot/snapshots/snapshot__doc_stdout_record.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__doc_stdout_record.ncl.snap
@@ -4,6 +4,4 @@ expression: out
 ---
 # `Hello`
 
-\
 World\!
-
--- a/cli/tests/snapshot/snapshots/snapshot__doc_stdout_types.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__doc_stdout_types.ncl.snap
@@ -6,6 +6,4 @@ expression: out
 
 - `Hello : Number`
 - `Hello | std.string.Stringable`
-\
 World\!
-
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -789,8 +789,6 @@ mod doc {
                     ))
                 }
 
-                document.append(arena.alloc(AstNode::from(NodeValue::LineBreak)));
-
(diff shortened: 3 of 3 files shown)


vkleen (approved):

thufschmitt (approved):

jneem:
Doesn't this regress the example in #1520?

yannham:
Argh, that's a good point. That being said, I feel this time this is really a comrak issue: it should insert a newline between a list and the following block, but it doesn't. @jneem if you have a branch with the latest comrak, could you see if this has changed in the latest version?

Otherwise another solution (which I actually implemented initially) is to check if the next element is a paragraph manually, and insert the hard line only in this case. It's not too hard, but just doesn't feel right after having read the commonmark spec.

PR #1883, opened 2024-04-05
Dedup diagnostics

### LINKED ISSUE #1882: NLS reports same diagnostic message multiple times
**Describe the bug**

NLS reports same diagnostic message multiple times when the recursive definitions is present. Related to #1875.

https://github.com/tweag/nickel/assets/13243167/dca92462-5493-4891-aa1a-83765bb27518

**To Reproduce**

```nickel
{
  applications = {
    apps | not_exported = applications,
    message = "message",
    blah | String = 5,
  }
} 
```

**Expected behavior**

The diagnostic message will be reported only once per case.

**Environment**
 - OS name + version:

```
NAME="Fedora Linux"
VERSION="39.20240327.0 (Sericea)"
```

 - Version of the code: [262ac8fa36df3268414750094ddaa358344decf2](https://github.com/tweag/nickel/commit/262ac8fa36df3268414750094ddaa358344decf2)

jneem:
Unfortunately the `lsp_types` crate doesn't provide a lot of `Ord` impls, so I had to write some wrappers. Is there a better way? (Of course we could use hashmaps to dedup, but `lsp_types` also doesn't provide `Hash` impls)

Fixes #1882

### DIFF
--- a/lsp/nls/src/background.rs
+++ b/lsp/nls/src/background.rs
@@ -112,6 +112,8 @@ pub fn worker_main() -> anyhow::Result<()> {
             );
         }
 
+        diagnostics.sort();
+        diagnostics.dedup();
         let diagnostics = Diagnostics { path, diagnostics };
 
         // If this fails, the main process has already exited. No need for a loud error in that case.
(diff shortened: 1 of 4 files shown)


yannham (approved):
I don't know of an obvious better way to do that on a foreign types (which means it's not possible to derive anything automatically either). Although I wonder if we really need the newtype wrappers boilerplate, and if we couldn't just move `cmp` to free-standing comparison functions instead which we then pass to e.g. `sort_by`. That might save a few `map` and unwrapping?

PR #1887, opened 2024-04-11
Extend the symbol range to include the rhs

### LINKED ISSUE #1886: VSCode Extension should not disregard cursor position and collapse all symbol breadcrumbs to "..." in all occasions
**Describe the bug**
The latest version of `nls` provides proper symbol outlines in the "Outlines" Tab. Normally the same symbol information is also displayed as "breadcrumbs" at the top of the editor view, continuing after the file path and scrolls horizontally if the width exceeds the window (below is an example from a rust file in the nickel codebase, powered by rust-analyzer)

image

However with a nickel file, all symbol information are collapsed in "..." after the file path (see below, note the cursor is positioned at `_source2`). While one can always click on "..." and navigate to the symbol, IMHO this behavior greatly diminishes the primary utility of breadcrumb that is providing direct visual guidance to the user without keyboard / mouse interaction.

image

I searched for vscode setting entries and couldn't find a knob that tweaks this behavior.

**To Reproduce**

**Expected behavior**
Expand the symbol information inline in the breadcrumbs bar.

**Environment**
 - Version of the code: `nickel-lang-cli nickel 1.5.0 (rev 13ffc85)`

**Additional context**

jneem:
Fixes #1886.

The symbols response has two ranges: the "selection range" is just the position of the symbol name, but the "range" is supposed to include the "body" of the symbol. VSCode uses this range for the breadcrumb bar: it shows the symbols in the bar when your cursor is in that symbol's body. Previously, we were just reporting the position of the symbol name for both ranges.


lsp/nls/src/requests/symbols.rs
```diff
@@ -36,10 +36,47 @@ fn symbols(
...
+                if let Some(val_span) = field.value.as_ref().and_then(|val| val.pos.into_opt()) {
+                    last_pos = last_pos.max(val_span.end.to_usize());
+                }
+
+                if let Some(last_ty_pos) = field
+                    .metadata
+                    .annotation
+                    .contracts
+                    .iter()
+                    .chain(field.metadata.annotation.typ.as_ref())
```
yannham, line 60: There's an `iter()` method on `TypeAnnotation` that is returning the type chained with the contracts as an iterator.

lsp/nls/src/requests/symbols.rs
```diff
@@ -36,10 +36,47 @@ fn symbols(
...
+                    .filter_map(|ty| ty.typ.pos.as_opt_ref())
+                    .map(|pos| pos.end.to_usize())
+                    .max()
+                {
+                    last_pos = last_pos.max(last_ty_pos);
```
yannham, line 65: I don't know if it's useful to reduce boilerplate, but there's a `fuse()` method on `RawSpan`, which returns the smallest enclosing span that contains both spans given as arguments.

lsp/nls/tests/inputs/symbols-basic.ncl
```diff
@@ -9,6 +9,8 @@ let func = fun x => 1 in
   type_checked_block = {
     inner_name = { name = "value" },
   } : _,
+  annotated | String,
+  docced | doc "The symbol range doesn't current include docs",
```
yannham, line 13:
```suggestion
  docced | doc "The symbol range doesn't currently include docs",
```

yannham (approved):

lsp/nls/src/requests/symbols.rs
jneem: That did turn out nicer, thanks!

PR #1888, opened 2024-04-12
Leverage function contract information

### LINKED ISSUE #1885: VSCode extension does not show hints for functions annotated with record contract

Consider the following example:
```nickel
{
  Schema = {
    foo,
    bar
      | {
        baz,
      }
  },
  some_func
    | Schema -> Dyn
    = fun src =>
      {
        some_val = src.bar.baz
      }
      ,
  some_record = {
    src | Schema,
    some_val  = src.bar.baz
  }
}
```
within `some_record`, the extension beautifully shows hints on the structure of `Schema` as I type after `some_val = src.`, whereas within some_func, no hint is available after `some_val = src.`

Enable hints for function parameters that are annotated with record contract.

**Additional context**
```
> nickel --version  
nickel-lang-cli nickel 1.5.0 (rev 13ffc85)
```

yannham:
Closes #1885.

When a function is annotated with a function contract, use this additional type information to enable completion in the LSP (and also use the domain as the type of the function's argument in the typing environment, even if we are in walk mode). Contract data are easy to fetch during typechecking when available and use them improve the developer experience by bringing in more static information.

I hesitated to updated the documentation about the fact that the typechecker is marginally smarter about function contracts, but I'm not sure yet we want to commit to it fully yet. Honestly I don't expect it to be used a lot in the wild, I mostly did it because it was virtually free (performance-wise and code complexity-wise).

### DIFF
--- a/core/src/typecheck/mod.rs
+++ b/core/src/typecheck/mod.rs
@@ -1462,7 +1462,9 @@ fn walk<V: TypecheckVisitor>(
                 })
         }
         Term::Fun(id, t) => {
-            // The parameter of an unannotated function is always assigned type `Dyn`.
+            // The parameter of an unannotated function is always assigned type `Dyn`, unless the
+            // function is directly annotated with a function contract (see the special casing in
+            // `walk_with_annot`).
             ctxt.type_env.insert(id.ident(), mk_uniftype::dynamic());
             walk(state, ctxt, visitor, t)
         }
@@ -1698,7 +1700,7 @@ fn walk_annotated<V: TypecheckVisitor>(
 /// type or contract annotation. A type annotation switches the typechecking mode to _enforce_.
 fn walk_with_annot<V: TypecheckVisitor>(
     state: &mut State,
-    ctxt: Context,
+    mut ctxt: Context,
     visitor: &mut V,
     annot: &TypeAnnotation,
     value: Option<&RichTerm>,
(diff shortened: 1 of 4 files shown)


lsp/nls/tests/inputs/completion-fun-parameter-contract.ncl
```diff
@@ -0,0 +1,17 @@
...
+{
+  Schema = {
+    foo | { bar }
+  },
+  some_func
+    | Schema -> Dyn
+    = fun src =>
+      {
+        # TODO: we shouldn't need to type the first letter to get completion.
+        some_val = src.foo.b
```
yannham, line 11: I realized that for some reason, the completion doesn't fire if we don't start to typing a prefix of an existing field, while this doesn't happen for other kind of completion (for type-based completion for example, if you start to type `std.`, you get a list of all available functions - same for a record literal, if you define `record = {...}` and then later type `record.`, you get a full list - but here we get nothing until we type the first letter of a field). I'll open a separate issue.

yannham: In fact it's the good old NLS can't handle some incomplete inputs. If I add an identifier at the end of this line, something like `src_val = src.foo.  dummy_value`, which is then parsed correctly (because right now you can have a space between fields and the `.` character, for the best or the worst), then you get completion.

core/src/typecheck/mod.rs
```diff
@@ -1718,10 +1720,32 @@ fn walk_with_annot<V: TypecheckVisitor>(
...
+                contracts,
+            },
+            Some(value),
+        ) => {
+            // If we see a function annotated with a function contract, we can get the type of the
+            // argument for free. We use this information both for typechecking (you could see it
+            // as an extension of the philosophy of apparent types, but for function arguments
+            // instead of let-bindings) and for the LSP, to provide better type information and
+            // completion.
+            if let (Some(contract), Term::Fun(id, body)) = (contracts.first(), value.as_ref()) {
```
jneem, line 1735: Why only check the first contract?

jneem (approved):

core/src/typecheck/mod.rs
yannham: We could indeed try to get the first contract that is a function contract, if any.

PR #1889, opened 2024-04-12
Fix LSP not showing type signature in untyped code

yannham:
Most of the time, hovering over a symbol in untyped code would show `Dyn` even though the symbol has a type annotation - typically for records accessed via a non-trivial path, e.g. `foo.bar.baz`. This is an issue especially because we don't get to see the right types for stdlib symbols when hovering over them.

Capture d’écran du 2024-04-12 15-58-40

This PR fixes the issue by taking the type annotation (technically one of the aggregated type annotations) instead of the type provided by the typechecker when the latter is `Dyn`.

### DIFF
--- a/lsp/nls/tests/inputs/hover-stdlib-type.ncl
+++ b/lsp/nls/tests/inputs/hover-stdlib-type.ncl
@@ -1,0 +1,12 @@
+### /main.ncl
+let foo = { bar.baz : String -> Number = fun x => 1 }
+in
+foo.bar.baz "0" + std.string.to_number "1"
+### [[request]]
+### type = "Hover"
+### textDocument.uri = "file:///main.ncl"
+### position = { line = 2, character = 10 }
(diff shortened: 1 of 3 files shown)


lsp/nls/tests/inputs/hover-stdlib-type.ncl
```diff
@@ -0,0 +1,12 @@
...
+in
+foo.bar.baz "0" + std.string.to_number "1"
+### [[request]]
+### type = "Hover"
+### textDocument.uri = "file:///main.ncl"
+### position = { line = 2, character = 10 }
+### [[request]]
+### type = "Hover"
+### textDocument.uri = "file:///main.ncl"
+### position = { line = 2, character = 10 }
```
jneem, line 12: I guess you didn't mean to ask for the same position twice?

jneem (approved):

PR #1893, opened 2024-04-22
Update to latest Topiary and tree-sitter-nickel

yannham:
For some reason, the version of Topiary and of the Nickel tree sitter grammar were lagging behind, before the support for pattern matching was added, which means current master wasn't able to format a file with pattersn in it - maybe it was a rollback because of the Topiary install issue (https://github.com/tweag/topiary/issues/690).

The former issue has been fixed, so this commit update to latest Topiary in order to get formatting of patterns.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3011,7 +3011,7 @@ dependencies = [
 [[package]]
 name = "topiary-core"
 version = "0.3.0"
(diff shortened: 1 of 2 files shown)


ErinvanderVeen (approved):
It looks fine by me, but something seems wrong with CI?

yannham:
Well, it's awfully slow, probably because this triggered some rebuild. But so far I don't see anything red

PR #1897, opened 2024-04-25
Add support for constant patterns

yannham:
This PR adds support for constant patterns, that is patterns matching specific booleans, strings, numbers or the null value. This allows to write for exampe:

```nickel
match {
  "foo" => ...,
  "foo/bar" => ...,
  other => ...
}
```

Or

```nickel
match {
  {enabled = true, ..rest} => code_path_when_enabled rest,
  {enabled = false, ..} => 'Error "Is not enabled!",
}
```

For destructuring, the semantics is the same as for enum tags - which are actually constants as well, but were just supported as a side effect of supporting eneral enum patterns - : a constant pattern forces the value and works as an assertion.


core/src/term/pattern/compile.rs
```diff
@@ -258,6 +258,46 @@ impl CompilePart for PatternData {
...
+impl CompilePart for ConstantPattern {
+    fn compile_part(&self, value_id: LocIdent, bindings_id: LocIdent) -> RichTerm {
+        self.data.compile_part(value_id, bindings_id)
+    }
+}
+
+impl CompilePart for ConstantPatternData {
+    fn compile_part(&self, value_id: LocIdent, bindings_id: LocIdent) -> RichTerm {
+        let compile_constant = |nickel_type: &str, value: Term| {
+            // if %typeof% value_id == '<nickel_type> && value_id == <value> then
```
jneem, line 275: Is the type comparison an optimization? It looks to me like `BinaryOp::Eq` already handles mismatched types.

jneem (approved):

core/src/term/pattern/compile.rs
yannham: No, it's because we can't compare functions. So we want to avoid this runtime error and eschew the comparison instead:

```
error: cannot compare values for equality
  ┌─ <repl-input-1>:1:1
  │
1 │ (fun _x => null) == (fun _x => null)
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  │ │
  │ an argument has type Function, which cannot be compared for equality
  │ problematic argument evaluated to this
```

Since we have to do a pre-condition check anyway, instead of checking if one is a function, it's just simpler and probably faster to just avoid comparison for any value that doesn't have the expected type to begin with.

PR #1900, opened 2024-04-29
Add `new_from_sources` function to `Program`

ErinvanderVeen:
This is similar to the `new_from_path` and `new_from_paths` function. A `new_from_source` function already existed, but a `new_from_sources` was still missing.

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -288,6 +288,8 @@ impl<EC: EvalCache> Program<EC> {
         I: IntoIterator<Item = P>,
         P: Into<OsString>,
     {
+        // The File type parameter is a dummy type and not used.
+        // It just needed to be something that implements Read, and File seemed fitting.
         Self::new_from_inputs(
             paths.into_iter().map(Input::<std::fs::File, _>::Path),
             trace,
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -274,6 +274,47 @@ impl<EC: EvalCache> Program<EC> {
         })
     }
 
+    /// Create program from possibly multiple source. The main program will be
+    /// the [`BinaryOp::Merge`] of all the inputs.
+    pub fn new_from_sources<I, T, S>(
+        sources: I,
+        trace: impl Write + 'static,
+    ) -> std::io::Result<Self>
+    where
+        I: IntoIterator<Item = (T, S)>,
+        T: Read,
+        S: Into<OsString> + Clone,
+    {
+        increment!("Program::new");
+        let mut cache = Cache::new(ErrorTolerance::Strict);
+
+        let merge_term = sources
+            .into_iter()
+            .map(|(mut src, name)| {
+                let path = PathBuf::from(name.into());
+                // The main_id will be the final merged expression
+                let _ = cache
+                    .add_source(SourcePath::Path(path.clone()), &mut src)
+                    .unwrap();
+                RichTerm::from(Term::Import(path.into()))
+            })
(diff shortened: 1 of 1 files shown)


jneem (approved):

yannham (commented):
I think there's a bit of code duplication between `new_from_sources` and `new_from_files`. Although it's more salient here, there is in fact also a bit of code duplication between `new_from_source` and `new_from_file`.

It seems the only difference between `from_source(s)` and `from_file(s)` is that we use `cache.add_source` vs `cache.add_file`. After a quick look it doesn't seem reasonable to use `cache.add_source` generically because `cache.add_file`, although it does call to `cache.add_source`, also does more file-specific and non trivial operations like registering timestamps and stuff.

Do you think we could factor this out, by having a private generic function that e.g. takes a closure or just a bespoke enum `SourceOrFile`, and have `new_from_source(s)` and `new_from_file(s)` be simple wrapper around the corresponding underlying implementation?

ErinvanderVeen:
> I think there's a bit of code duplication between `new_from_sources` and `new_from_files`. Although it's more salient here, there is in fact also a bit of code duplication between `new_from_source` and `new_from_file`.
> 
> It seems the only difference between `from_source(s)` and `from_file(s)` is that we use `cache.add_source` vs `cache.add_file`. After a quick look it doesn't seem reasonable to use `cache.add_source` generically because `cache.add_file`, although it does call to `cache.add_source`, also does more file-specific and non trivial operations like registering timestamps and stuff.
> 
> Do you think we could factor this out, by having a private generic function that e.g. takes a closure or just a bespoke enum `SourceOrFile`, and have `new_from_source(s)` and `new_from_file(s)` be simple wrappers around the corresponding underlying implementation?

Absolutely! I'll get on it!

ErinvanderVeen:
We spoke before about how this is a good private function, but with a few minor tweaks, we could make it possible for users to mix input sources.

Topiary has a use for this actually. We want to create a single program from a Merge between the builtin source, and the configurations from the user (files). Our current solution is to read the files to a string, and create a cursor from that. Not ideal.

We would just have to expand the `Input` type to take another Type parameter:

```rust
enum Input<F, T, S> {
    /// A filepath
    Path(F),
    /// The source is anything that can be Read from, the second argument is the name the source should have in the cache.
    Source(T, S),
}
```

For use like: `Input::<std::fs::File, Cursor, String>`.

core/src/program.rs
```diff
@@ -192,28 +192,77 @@ pub struct Program<EC: EvalCache> {
     pub field: FieldPath,
 }
 
+/// The Possible Input Sources, anything that a Nickel program can be created from
+enum Input<T, S> {
```
yannham, line 196: So in the end, you didn't make it public? Or are you going to do that in a separate PR?

core/src/program.rs
```diff
@@ -256,22 +312,24 @@ impl<EC: EvalCache> Program<EC> {
...
-        S: Into<OsString> + Clone,
+        S: Into<OsString>,
     {
-        increment!("Program::new");
-
-        let mut cache = Cache::new(ErrorTolerance::Strict);
-        let path = PathBuf::from(source_name.into());
-        let main_id = cache.add_source(SourcePath::Path(path), source)?;
-        let vm = VirtualMachine::new(cache, trace);
+        Self::new_from_input(Input::Source(source, source_name.into()), trace)
```
yannham, line 317: It seems we are always coercing `source_name` to an `OsString` before putting it in the `Input::Source`. So is it worth it to have the additional type parameter `S` on `Input`? Couldn't we just replace it with `OsString` without having to change anything else?

yannham (approved):

PR #1904, opened 2024-05-02
Wildcard patterns

yannham:
This PR adds a special pattern that match any value but doesn't bind anything, the wildcard pattern `_`. 

This subsumes the previously special case of the catch-all case of match expression `_ => default_value`. Now a match expression is simply a list of patterns, and `_` is just one of them.

This addition allows to use `_` anywhere in a nested pattern, like `'Foo ('Bar _)`, which is especially useful when matching on enum variants. The ideas introduced for the typechecking of the catch-all case `_ =>` of match expressions naturally extend to wildcards pattern: if somewhere in a nested pattern, there is a wildcard which is in the same position as an enum variant in a sibling pattern, like `match { {foo = 'Bar} => something, {foo = _} => something_else}`, this makes the corresponding enum type open: the above match expression has type `forall r. {foo : [| 'Bar; r |]} -> SomeType`.

A wildcard pattern doesn't force the value being matched on, following the idea that it's equivalent to `let _x = value in body` (but without binding `_x`). In general, from now on, we'll try to follow the idea that the semantics of destructuring `let <pattern> = value in body` should be equivalent to `value |> match { <pattern> => body, _ => <NonExhaustiveDestructuringError> }`. Following this guideline, it confirms that `let _` shouldn't force the matched value. One last argument for this behavior is that `fun _ => body` is desugared to `fun fresh_var => let _ = fresh_var in body`, which means that making `let _` strict would make `fun _ => body` strict in its argument, which is arguably very surprising (for reviewers, this contradicts what I said today earlier in the weekly meeting, because I made a mistake with wildcard patterns specifically - the "destructuring should force" is still valid for other patterns though, and it agrees with everything said here)

### DIFF
--- a/core/src/term/pattern/compile.rs
+++ b/core/src/term/pattern/compile.rs
@@ -252,6 +252,7 @@ impl CompilePart for Pattern {
 impl CompilePart for PatternData {
     fn compile_part(&self, value_id: LocIdent, bindings_id: LocIdent) -> RichTerm {
         match self {
+            PatternData::Wildcard => Term::Var(bindings_id).into(),
             PatternData::Any(id) => {
                 // %record_insert% "<id>" value_id bindings_id
                 insert_binding(*id, value_id, bindings_id)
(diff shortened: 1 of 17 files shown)


core/src/term/pattern/compile.rs
```diff
@@ -670,31 +671,48 @@ impl Compile for MatchData {
...
-                let PatternData::Enum(EnumPattern {tag, ..}) = pat.data else {
-                    panic!("match compilation: just tested that all cases are enum tags, but found a non enum tag pattern");
-                };
-
-                (tag, body)
-            }).collect();
+            // We take the first wildcard pattern as the default case. In theory, we could discard
+            // all the tags coming after the default branch; in practice we expect any sane match
+            // expression to have the default branch at the end, so this isn't a very useful
+            // optimization to care about.
```
jneem, line 680: Unfortunately, I think you do need to discard the other tags. I just checked, and
```
'Bar |> match {
  'Foo => 1,
  _ => 2,
  'Bar => 3,
}
```
evaluates to 3, when it should evaluate to 2.

core/src/typecheck/pattern.rs
```diff
@@ -64,48 +71,58 @@ pub struct PatternTypeData<T> {
...
+    ///   {foo = 'Qux x} => <exp>,
+    ///   {foo = _} => <exp>,
+    /// }
+    /// ```
+    ///
+    /// Similarly, the type of the match expression is `{ foo: [| 'Bar: a, 'Qux: b; c |] } -> e`.
     ///
     /// See [^typechecking-match-expression] in [typecheck] for more details.
     pub enum_open_tails: Vec<(PatternPath, UnifEnumRows)>,
+    /// Paths of the occurrence of  wildcard patterns encountered. This is used to determine which
```
jneem, line 90:
```suggestion
    /// Paths of the occurrence of wildcard patterns encountered. This is used to determine which
```

core/src/term/pattern/compile.rs
yannham: Ah, right, this comment dates from the introduction of pattern matching and the specific optimization toward enums. It seems the accidental semantics has been that the catch-all case was independent of its position (and indeed, this was how it used to be parsed even). But now this is not the case anymore.

Good catch, thanks for noticing this!

jneem (approved):

PR #1910, opened 2024-05-06
Implement pattern guards

yannham:
This PR implements pattern guards, which are side conditions that can be added to the branch of a match expression, following the pattern, to further constrain the matching. A guard is introduced by the (already existing) `if` keyword. For example:

```nickel
nickel> {list = std.array.generate std.function.id 20}
|> match {
  {list} if std.array.length list > 10 => 'Big,
  {list} if list == [] => 'Empty,
  {list} => 'Small
}
'Big

nickel>
```

The compilation is rather straightforward, as a pattern is already compiled to a tree of if-then-else while also building the bindings introduced by pattern variables. After all the conditions coming from the pattern have been tested, we just additionally check for the guard (injecting the bindings in the condition since the guard can - and most often does - use variables bound by the pattern).

I expect this feature to be particularly nice for writing custom contracts, which are currently in practice sizeable nested if-then-else trees, which could hopefully be aggressively flattened with patterns and guards.

### DIFF
--- a/core/src/term/pattern/compile.rs
+++ b/core/src/term/pattern/compile.rs
@@ -668,11 +668,15 @@ impl Compile for MatchData {
     //      # this primop evaluates body with an environment extended with bindings_id
     //      %pattern_branch% body bindings_id
     fn compile(mut self, value: RichTerm, pos: TermPos) -> RichTerm {
-        if self.branches.iter().all(|MatchBranch { pattern, .. }| {
+        if self.branches.iter().all(|branch| {
+            // While we could get something working even with a guard, it's a bit more work and
+            // there's no current incentive to do so (a guard on a tags-only match is arguably less
+            // common, as such patterns don't bind any variable). For the time being, we just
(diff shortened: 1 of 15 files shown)


core/src/term/pattern/compile.rs
```diff
@@ -668,23 +668,28 @@ impl Compile for MatchData {
     //      # this primop evaluates body with an environment extended with bindings_id
     //      %pattern_branch% body bindings_id
     fn compile(mut self, value: RichTerm, pos: TermPos) -> RichTerm {
-        if self.branches.iter().all(|(pat, _)| {
```
jneem, line 671: For the TagsOnlyMatch optimization, don't you also need to check that there are no guards?

yannham: Ah, you're right (although we could still first do the hashmap access and then check the guard, but it's out of scope for this PR and also probably much less useful/common). Thanks

jneem (approved):

PR #1912, opened 2024-05-07
Implement array patterns

yannham:
Depends on #1910.

This commit adds a new class of patterns: array patterns. This is a natural extension, both syntactically and semantically, of existing data structure patterns in Nickel (in particular of records). Similarly, arrays pattern can also capture the rest of the pattern (the tail of the array that hasn't been matched yet) and bind it to a variable.


core/src/typecheck/pattern.rs
```diff
@@ -222,13 +223,6 @@ pub(super) trait PatternTypes {
     ) -> Result<Self::PatType, TypecheckError>;
 }
 
-/// Builds the type associated to a record pattern. When matching a value against a pattern in a
-/// statically typed code, for example in a let destructuring or via a match expression, the type
-/// of the value will be checked against the type generated by `build_pattern_type`.
-///
-/// The type of each "leaf" identifier will be assigned based on the `mode` argument. The current
-/// possibilities are for each leaf to have type `Dyn`, to use an explicit type annotation, or to
-/// be assigned a fresh unification variable.
```
yannham, line 231: This was somehow redundant with the documentation of the trait itself, and had not much to do here, attached to a specific implementation.

jneem (approved):

Issue #1909, opened 2024-05-03
Opaque foreign values


jneem:
This is some kind of baby step towards FFI: allow for opaque foreign values to be injected into nickel and copied around, but not otherwise interacted with or modified from within nickel.

This was a request from the keystone project, with some [discussion](https://discord.com/channels/1174731094726295632/1179430745727586407/1236018818711556146) on the discord.

Question: what should be payload of the opaque foreign values be? `u64` might be sufficient: that gives a lot of values, and the user could "embed" richer data by maintaining a lookup table on their side. Maybe `Box<dyn Any>` is also a possibility?

yannham:
I would close this, given #1931 has been merged. Foreign value are still quite bare-bone, but if there is demand for missing features, I think they would be better handled by new, focused issues.

PR #1913, opened 2024-05-10
Opaque values

jneem:
Here's a first stab at opaque values, as in #1909. The payload is a `u64` for now -- I think we could change it without breaking our backwards-compatibility promises, because you can only generate these when embedding nickel, and we don't promise much about the stability of nickel-as-a-library.

### DIFF
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -267,7 +267,7 @@ pub enum Term {
     ///
     /// This can be used by programs that embed Nickel, as they can inject these opaque
     /// values into the AST.
-    Opaque(u64),
+    ForeignId(u64),
 }
 
 // PartialEq is mostly used for tests, when it's handy to compare something to an expected result.
@@ -879,7 +879,7 @@ impl Term {
             Term::SealingKey(_) => Some("SealingKey".to_owned()),
             Term::Sealed(..) => Some("Sealed".to_owned()),
             Term::Annotated(..) => Some("Annotated".to_owned()),
-            Term::Opaque(_) => Some("Opaque".to_owned()),
+            Term::ForeignId(_) => Some("ForeignId".to_owned()),
             Term::Let(..)
             | Term::LetPattern(..)
             | Term::App(_, _)
@@ -926,7 +926,7 @@ impl Term {
             | Term::EnumVariant {..}
             | Term::Record(..)
             | Term::Array(..)
-            | Term::Opaque(_)
+            | Term::ForeignId(_)
             | Term::SealingKey(_) => true,
             Term::Let(..)
             | Term::LetPattern(..)
@@ -984,7 +984,7 @@ impl Term {
             | Term::Str(_)
             | Term::Lbl(_)
             | Term::Enum(_)
-            | Term::Opaque(_)
+            | Term::ForeignId(_)
             | Term::SealingKey(_) => true,
             Term::Let(..)
             | Term::LetPattern(..)
@@ -1027,7 +1027,7 @@ impl Term {
             | Term::Array(..)
             | Term::Var(..)
             | Term::SealingKey(..)
-            | Term::Opaque(..)
+            | Term::ForeignId(..)
             | Term::Op1(UnaryOp::StaticAccess(_), _)
             | Term::Op2(BinaryOp::DynAccess(), _, _)
             // Those special cases aren't really atoms, but mustn't be parenthesized because they
@@ -2180,7 +2180,7 @@ impl Traverse<RichTerm> for RichTerm {
             | Term::Import(_)
             | Term::ResolvedImport(_)
             | Term::SealingKey(_)
-            | Term::Opaque(_)
+            | Term::ForeignId(_)
             | Term::ParseError(_)
             | Term::RuntimeError(_) => None,
             Term::StrChunks(chunks) => chunks.iter().find_map(|ch| {
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -58,6 +58,9 @@ use std::{
     rc::Rc,
 };
 
+/// The payload of a `Term::ForeignId`.
+pub type ForeignIdPayload = u64;
+
 /// The AST of a Nickel expression.
 ///
 /// Parsed terms also need to store their position in the source for error reporting.  This is why
@@ -267,7 +270,7 @@ pub enum Term {
     ///
     /// This can be used by programs that embed Nickel, as they can inject these opaque
     /// values into the AST.
-    ForeignId(u64),
+    ForeignId(ForeignIdPayload),
 }
 
 // PartialEq is mostly used for tests, when it's handy to compare something to an expected result.
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -261,6 +261,13 @@ pub enum Term {
     ///
     /// This is a temporary solution, and will be removed in the future.
     Closure(CacheIndex),
+
+    #[serde(skip)]
+    /// An opaque value that cannot be constructed within Nickel code.
+    ///
+    /// This can be used by programs that embed Nickel, as they can inject these opaque
+    /// values into the AST.
+    Opaque(u64),
 }
 
 // PartialEq is mostly used for tests, when it's handy to compare something to an expected result.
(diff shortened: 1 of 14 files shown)


core/src/term/mod.rs
```diff
@@ -261,6 +261,13 @@ pub enum Term {
     ///
     /// This is a temporary solution, and will be removed in the future.
     Closure(CacheIndex),
+
+    #[serde(skip)]
+    /// An opaque value that cannot be constructed within Nickel code.
```
yannham, line 266: Could be worth to insist on the fact that they cannot be constructed, but also should _never_ be observable from within Nickel code (compared or distinguished by any means).

core/src/term/mod.rs
```diff
@@ -261,6 +261,13 @@ pub enum Term {
...
+    ///
+    /// This can be used by programs that embed Nickel, as they can inject these opaque
+    /// values into the AST.
+    Opaque(u64),
```
yannham, line 270: I don't know if we spell `u64` somewhere else, but I wonder if we should use a type alias instead to make it easy to switch to something else (ie define `Opaque(OpaquePayload)` with `type OpaquePayload = u64` or something).

core/src/typ.rs
```diff
@@ -269,6 +269,8 @@ pub enum TypeF<Ty, RRows, ERows> {
     ///
     /// See [`crate::term::Term::Sealed`].
     Symbol,
+    /// An opaque value, the type of `Term::Opaque`.
+    Opaque,
```
yannham, line 273: I wonder if we should even worry about having a type for that, given that the user can't do much with opaque values, and can't use this type right now (it's not in the grammar, and it would be a breaking change to add it). However it doesn't cost much, and maybe it could be useful for Rust binaries consuming the library? I'm not sure.

core/src/typ.rs
```diff
@@ -818,6 +821,7 @@ impl Subcontract for Type {
             TypeF::Number => internals::num(),
             TypeF::Bool => internals::bool(),
             TypeF::String => internals::string(),
+            TypeF::Opaque => internals::opaque(),
```
yannham, line 824: I think this contract can never be instantiated in practice (unless programmatically) . For example, we don't even bother elaborating a contract for `Symbol` below (which is also arguably not a very useful type). Maybe it would be more useful to have a `std.contract.Opaque` contract?

yannham (commented):
I wonder if we should have a type for that. A type is a bit useless in that opaque value can't be deconstructed or computed with. So maybe a contract in the stdlib is sufficient (and wouldn't reserve a new keyword).

On another front, I personally find `Opaque` to be a bit too generic, although it's rather descriptive. I wonder if we should insist on the fact that those values are intended to be injected by the environment, so maybe something like `ForeignValue`, `ForeignId`, `OpaqueForeign`, or something like that? But it's just a suggestion, though.

core/src/term/mod.rs
jneem: I made sure they can't be compared for equality; is there some other kind of observability that needs to be handled?

I was imagining that these values could be copied around (and possibly manipulated by functions provided by the nickel embedder). So you'd be able to write a contract like `{ username | String, token | ForeignId }`, and then by applying this contract you could observe the presence or absence of the token.

core/src/typ.rs
jneem: I was thinking that the nickel embedder might provide a function like `union: ForeignId -> ForeignId -> ForeignId`, so it would be nice if we could type-check it

yannham: Ah, you mean customizing the set of operations possible on opaque values with their own operations? There is still the problem that if the user can't write this type down, it is somehow a second class citizen (and once again, making it possible to spell out is a breaking change).

All of that being said, having the type internally is like 10 additional line of code, so let's not argue about it too much. It doesn't really hurt to have this type internally, and see later if it can be turned into something useful. I'm fine with keeping it as it is for now

yannham: A second idea to make it not a breaking change is to wait for the let-type RFC, and make it possible to define primitive types - think `%Opaque%` (just so it doesn't clash with user-land). Then, once we can export types from a record, we could define stdlib namespaced types, like `type Opaque = %Opaque%` in `std.foreign` or something.

core/src/term/mod.rs
yannham: Right, I said observable, but the right technical term is probably _separable_, which means "distinguishable".

That is, for every context `C` (that you can think of just as a function here), for any pair of foreign keys `k1` and `k2`, then if `C k1 k2 ~> v` (the context evaluates to the value `v`), then for all other foreign keys `k` and `l`, `C k l ~> v'` with `v ~ v'` (I won't define the precise meaning of `~`, but let's say it's an observational equivalence). Maybe we need to extend that to arbitrarily long finite lists of key.

Put differently, no result of an expression can depend on the actual values of the keys. They must be all interchangeable without affecting the semantics. We need to include error messages as well (so my above specification is incomplete), because that's another way for a malicious user to get the value of the key indirectly - but you properly don't print the content of the key in the pretty printer, which is all good.

aiverson: comparing by equality is actually specifically fine and useful. Even being able to use them as keys in maps is fine and useful as long as that doesn't reveal the content of the opaque reference directly or indirectly. (though there is some subtlety of exactly which equality is being checked, since there could be two copies of the same reference, two different references to the same local data, or two different references to promises which will eventually become the same)
It's being able to forge them or inspect the backing data that's the problem, and as long as there's no way to make a new opaque reference (from inside the language; making them from foreign functions into the trusted platform is fine) that is satisfied, and no way to look up what's behind the foreign reference. Being able to check the arbitrary number in the foreign reference that indexes the backing table isn't even a security issue as long as the table itself is impossible to access, but it is good practice to forbid that since any kind of behavioral reliance on the specific values of those numbers is fragile and should be prevented, and if the numbers are viewable in the first place someone accidentally allowing forgery is more likely.

core/src/typ.rs
aiverson: Perhaps a good way to do this would be to have a `newtype` construct to allow creating a distinct type family that is backed by an existing type, so that we can have this core Opaque/ForeignId type, then give names and possibly generics to various kinds of foreign data. That allows making more specific function signatures that work on foreign data too.

core/src/term/mod.rs
aiverson: It might even be fairly straightforward to make the interpreter generic over these if desired; I don't actually need that for my usecases so it might not be worth it, but it might allow generalizing to other systems nicely.

jneem: I think that would be a bit painful, since it would involve carrying around an extra type parameter everywhere that `Term` is used (which is basically everywhere). An easier version would be to have a `Box<dyn Any>` to allow downcasting.

yannham: I agree with Joe, on paper it's just adding a generic, but API-wise, this is painful, because you often have to propagate this type parameter everywhere (every site that uses a term, which is pretty much everywhere). For less fundamental datastructure we could a parameter and make an alias like `pub type Foo = ParametrizedFoo<usize>`, but for terms, I'm afraid this will leak in `typechecking`, `transform`, etc. which might not be worth it. The `Box<dyn Any>` doesn't look very nice, but yeah, if it's really needed, this might be a possibility.

core/src/typ.rs
yannham: Yeah, in fact the `let type` or `type = ` would precisely act like a `newtype`.  But the point stands that you need to be able to refer to a primitive `Opaque` type at some point, and just adding it would break backward compatibility (maybe someone called their contract `Opaque` already). So the idea would be to have the primitive type use an obscure internal syntax - as other primops work in nickel already - and the stdlib would just export `type Opaque = %BuiltinOpaqueType%`.

core/src/term/mod.rs
yannham: I think using them as map keys is much more invasive. Currently record fields in Nickel are all strings, that you can list with stdlib functions. So opaque value would need special casing here.

Regarding equality, this is easy to add - I just wonder if there are other use-cases where you would want to not even have equality. In any case, I would propose to start the most restrictive possible and move forward with this PR, and then see cases by case what we would add and why, if that sounds good to you all.

jneem:
Hm, it looks like I never updated `UnaryOp::Typeof` to support `'ForeignId`; that's added now

yannham (approved):
LGTM. Let's start like this and add operation on foreign ids as we see fit.

Issue #1300, opened 2023-05-12
Add testing infrastructure to the standard library


matthew-healy:
**Is your feature request related to a problem? Please describe.**

Currently there's no recommended/built-in way to write tests for Nickel code. It's not particularly hard to do (see the simple testing infra in `tests/integration/pass/lib/assert.ncl` for an example), but it seems like something enough projects will want that it's worth considering it for inclusion in the stdlib.

**Describe the solution you'd like**

It's probably sufficient to have some simple assertion primitives (in the form of contracts, I guess). It might be nice to spike out some ideas for more opinionated test structures, but I guess there'd be a higher bar there for inclusion in the stdlib.

**Describe alternatives you've considered**

The main alternative is to provide testing infrastructure in its own library. I'm not against this in principle, but given we don't have a simple way of importing Nickel packages yet (and we're unlikely to have one in the near future) I think it makes sense to at least add _something_ official to the stdlib.

vkleen:
A minimal step towards this could be putting `Assert` and `check` from `tests/integration/pass/lib/assert.ncl` into a stdlib namespace, e.g. `std.tests`. That seems to be all the Nickel specific functionality we have needed for a while; later we could maybe add some automation into the CLI to make it simpler to hook up.

matthew-healy:
Yeah, I agree that could be a reasonable first step. Though since stdlib releases are currently tightly coupled to language releases I wonder if we should explore the design space here a little more before making a final decision.

PR #1914, opened 2024-05-11
chore: add assert to the stdlib

ajbt200128:
Previously Nickel was using an Assert only for integration tests. This pr makes it officially part of the standard library.

Closes https://github.com/tweag/nickel/issues/1300


ajbt200128:
Note the original issue recommends adding these to `std.test` . I can move `Assert` and `check` there, and maybe `Equal` belongs there too. Added it to `std.contract` for now since `Equal` is there already

core/stdlib/std.ncl
```diff
@@ -792,6 +792,55 @@
...
+        1 == 1 | std.contract.Assert
+         => true
+        ```
+        "%
+        = fun l x => x || %blame% l,
+
+    # We can't use a static type because x | Assert is not of type bool. We
+    # could use additional contracts to make it work but it's not worth the
+    # hassle.
+    check
```
yannham, line 820: I think `check` is a bit too generic for the stdlib. When using it internally for Nickel tests it's fine, but if it's to move to the stdlib, I would pick a name that is more explicit. Maybe `assert_all` is more descriptive, for example. Same for the description, I would make it less "test"-oriented, and just says that this checks that the conditions provided as an array of boolean are true, or it will fail with a contract error pointing to the one that failed.

core/stdlib/std.ncl
```diff
@@ -792,6 +792,55 @@
...
+      | doc m%"
+
+        Check an array of tests. Apply the `std.contract.Assert` contract on
+        each element of the given array.
+
+        `std.contract.check tests` It is semantically equivalent to
```
yannham, line 826:
```suggestion
        `std.contract.check tests` is semantically equivalent to
```

core/tests/integration/inputs/contracts/contracts.ncl
```diff
@@ -1,5 +1,5 @@
 # test.type = 'pass'
-let {check, Assert, ..} = import "../lib/assert.ncl" in
+
```
yannham, line 2: As we use `Assert` a lot, I wonder if we shouldn't still alias it, with `let Assert = std.contract.Assert` instead of repeating the `std.contract` part everywhere. We usually use `check` only once, so `check` is fine as it is.

core/stdlib/std.ncl
ajbt200128: yea I agree, assert_all makes more sense. I don't see a reason not to add it to the stdlib, since it's such a common pattern

core/tests/integration/inputs/contracts/contracts.ncl
ajbt200128: Sure, we can do that

ajbt200128:
Looks like pre-commit is failing, must not have run properly on my machine

core/stdlib/std.ncl
```diff
@@ -792,6 +792,56 @@
...
+        (std.array.all function.id tests) | std.contract.Assert
+        ```
+
+        But will give precise error reporting when one boolean statement
+        fails.
+
+        # Example
+
+        ```nickel
+        [ (1 == 2), (1 == 1), (1 == 3) ] |> std.contract.check
```
yannham, line 838:
```suggestion
        [ (1 == 2), (1 == 1), (1 == 3) ] |> std.contract.assert_all
```

core/stdlib/std.ncl
```diff
@@ -792,6 +792,56 @@
...
+    assert_all
+      | doc m%"
+
+        Assert an array of boolean statements are true. Apply the
+        `std.contract.Assert` contract on each element of the given array.
+
+        `std.contract.assert_all tests` is semantically equivalent to
+
+        ```nickel
+        (std.array.all function.id tests) | std.contract.Assert
```
yannham, line 829: Initially I just wanted  to add  the missing `std` in front of `function.id`, but `std.function.id` is just longer than `fun x => x` :man_shrugging: 

```suggestion
        (std.array.all (fun x => x) tests) | std.contract.Assert
```

yannham (approved):
Beside a few comments on the doc, looks good!

yannham:
After some discussions, we chose to put it in a new dedicated module `std.test`.

yannham:
So, this PR was delayed for reasons that have nothing to do with the original change (mostly related to our flake being outdated because of unrelated C++ compilations errors for the nix-experimental feature), but I think I can finally proceed now.

PR #1916, opened 2024-05-13
Or patterns

yannham:
Depends on #1912.

This PR introduces or-patterns, which allows to express alternatives within patterns, including in a deep subpattern.

The compilation of or-patterns is rather simple: we simply try each alternative until one matches, and use the corresponding bindings.

Typechecking of or-patterns can be done following the same process as for typechecking a whole match expression (which is also a disjunction of patterns), although the treatment of bound variables is a bit different.

Most of the complexity of this PR comes from the fact that we don't want to make `or` a reserved language keyword, which would break backward compatibility. This is possible, because `or` in pattern can't be confused with an identifier, but it requires some tweaking to make our LALR(1) parser accept this.

## Syntax

Beside the technicalities around making the grammar unambiguous, this PR also imposes parentheses around enum variant patterns which are part of an or-pattern. That is, it forces to write `('Foo x) or ('Bar x)` instead of `'Foo x or 'Bar x`, which is arguably less readable. Even less so when the identifier is `or`: `'Foo or or 'Bar or` :cold_sweat:

### DIFF
--- a/core/src/parser/lexer.rs
+++ b/core/src/parser/lexer.rs
@@ -121,6 +121,10 @@ pub enum NormalToken<'input> {
     True,
     #[token("false")]
     False,
+    /// Or isn't a reserved keyword. It is a contextual keyword (a keyword that can be used as an
+    /// identifier because it's not ambiguous) within patterns.
+    #[token("or")]
+    Or,
 
     #[token("?")]
     QuestionMark,
--- a/core/src/term/pattern/mod.rs
+++ b/core/src/term/pattern/mod.rs
@@ -31,6 +31,8 @@ pub enum PatternData {
     Enum(EnumPattern),
     /// A constant pattern as in `42` or `true`.
     Constant(ConstantPattern),
+    /// A sequence of alternative patterns as in `'Foo _ or 'Bar _ or 'Baz _`.
+    Or(OrPattern),
 }
 
 /// A generic pattern, that can appear in a match expression (not yet implemented) or in a
@@ -137,6 +139,12 @@ pub enum ConstantPatternData {
     Number(Number),
     String(NickelString),
     Null,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct OrPattern {
+    pub patterns: Vec<Pattern>,
+    pub pos: TermPos,
 }
 
 /// The tail of a data structure pattern (record or array) which might capture the rest of said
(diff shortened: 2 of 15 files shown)


core/src/parser/lexer.rs
```diff
@@ -121,6 +121,10 @@ pub enum NormalToken<'input> {
     True,
     #[token("false")]
     False,
+    /// Or isn't a reserved keyword. It is a contextual keyword (a keyword that can be used as an
+    /// identifier because it's not ambiguous but) within patterns.
```
yannham, line 125:
```suggestion
    /// identifier because it's not ambiguous) within patterns.
```

core/src/term/pattern/mod.rs
```diff
@@ -31,6 +31,8 @@ pub enum PatternData {
     Enum(EnumPattern),
     /// A constant pattern as in `42` or `true`.
     Constant(ConstantPattern),
+    /// A sequence alternative patterns as in `'Foo _ or 'Bar _ or 'Baz _`.
```
yannham, line 34:
```suggestion
    /// A sequence of alternative patterns as in `'Foo _ or 'Bar _ or 'Baz _`.
```

core/src/transform/free_vars.rs
```diff
@@ -316,3 +311,18 @@ impl RemoveBindings for EnumPattern {
         }
     }
 }
+
+impl RemoveBindings for OrPattern {
+    fn remove_bindings(&self, working_set: &mut HashSet<Ident>) {
+        // Theoretically, we could just remove the bindings of the first pattern, as all
+        // branch in an or patterns should bind exactly the same variables. However, at the
```
yannham, line 318:
```suggestion
        // branches in an or patterns should bind exactly the same variables. However, at the
```

core/src/transform/free_vars.rs
```diff
@@ -316,3 +311,18 @@ impl RemoveBindings for EnumPattern {
         }
     }
 }
+
+impl RemoveBindings for OrPattern {
+    fn remove_bindings(&self, working_set: &mut HashSet<Ident>) {
+        // Theoretically, we could just remove the bindings of the first pattern, as all
+        // branches in an or patterns should bind exactly the same variables. However, at the
+        // time of writing, this condition isn't enforced at parsing type (it's enforced
```
yannham, line 319:
```suggestion
        // time of writing, this condition isn't enforced at parsing time (it's enforced
```

PR #1918, opened 2024-05-16
Update README.md for nix profile install

cloudyluna:


yannham:
Which makes me think, we should probably advise to install the stable branch, and not master. But another change for another PR :slightly_smiling_face:

PR #1926, opened 2024-05-24
Tell lsp about variable bindings in match statements

### LINKED ISSUE #1925: Context-sensitive completion in match cases

Within the branch of a match case, the Lsp isn't aware of the new variables brought in scope by the match, so doesn't offer to complete them.

For instance,

```nickel
match {
  'Foo bar => [CURSOR]
} baz
```

won't trigger a completion for `bar`.

Have the LSP take into account the match branch it's in

**Additional context**

Observed in Nickel 1.5.0

jneem:
Fixes #1925

### DIFF
--- a/lsp/nls/src/usage.rs
+++ b/lsp/nls/src/usage.rs
@@ -162,6 +162,34 @@ impl UsageLookup {
 
                         TraverseControl::ContinueWithScope(new_env)
                     }
+                    Term::App(f, value) => {
+                        if let Term::Match(data) = f.as_ref() {
+                            for branch in &data.branches {
+                                let mut new_env = env.clone();
+                                for (path, ident, _field) in branch.pattern.bindings() {
+                                    let def = Def::Let {
+                                        ident: ident.into(),
+                                        value: value.clone(),
+                                        path: path.into_iter().map(|x| x.ident()).collect(),
+                                    };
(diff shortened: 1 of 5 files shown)


lsp/nls/src/usage.rs
```diff
@@ -162,6 +162,34 @@ impl UsageLookup {
 
                         TraverseControl::ContinueWithScope(new_env)
                     }
+                    Term::App(f, value) => {
```
yannham, line 165: Should we also do this even when the match isn't applied, like a function definition `let foo = match { 'Bar payload => p[.] }` to get `payload` completion on `p` (and maybe even record completion if we have type information on the argument of `foo`)?

jneem: Good point, I've added that. I'm having some trouble coming up with a realistic test that uses type information to get completions. The problem is that if typechecking fails then completions just don't work at all right now. (I'll open an issue for this.) This means that something like
```
fun x => match { y => y.fo } : { foo: Number} -> Number
```
fails to type-check and therefore doesn't offer completions. I've made it "work" for now by writing
```
fun x => match { y => y.fo } : { foo: Number, fo: Number} -> Number
```

yannham (approved):

PR #1937, opened 2024-06-03
Rename and re-organize primitive operators

yannham:
Primitive operators haven't got much love, as opposed to user-facing interfaces like the stdlib, and they have grown organically since the beginning of the very first Nickel prototype. As a result, the naming is inconsistent, with several generations, both in the surface syntax of Nickel and internally in the Rust codebase.

This PR makes a cleaning pass on primitive operators, by:

1. Use full worlds in the style of the current stdlib: `str` ->
   `string`, `num` -> `number`, etc.
2. Introduce pseudo-namespaces: instead of `str_foo` and `record_bar`, we use `/` as a separator for categories. The previous examples become `string/foo` and `record/bar`. We don't use `.`, to make it clear that it's not a normal record access, but just a nice way to dinstinguish the different categories
3. Align old operators on the current naming in the standard library.

This PR is mostly read for review, but in order to appease the CI, I need to update the tree-sitter grammar first to accommodate for the `/` character in primops.


yannham:
Although this PR is also blocked on the `topiary-queries` update (https://github.com/tweag/topiary/pull/708), it's in reviewable state.

jneem (approved):

PR #1939, opened 2024-06-04
Add from/to_tag_and_arg

yannham:
Depends on #1937.

This PR adds two functions to convert enums to and back from records, represented as a tag and an optional argument. Such functions are useful to handle enums in a general, dynamic way, while pattern matching requires to know in advance the possible tags. They might prove useful if we have an `export_with` metadata or the like, one day.

Additionally, we also implement a `map` function, which can be derived from the conversions but is useful in its own right.

As far as naming is concerned, we initially hesitated with `to_record` and `from_record`, but this doesn't really give a hint as to what is the conversion. `tagged_value` was also considered, but in some sense an enum is a tagged value as well, so it's not telling much. I initially picked `to/from_tag_and_value`, but I think we are using the term "argument" pretty much everywhere else when talking about enum variants (in particular because the syntax is application-like, so it makes sense that `arg` is an argument in `'Foo arg`), so I switched to keep a consistent jargon.


yannham:
@jneem gentle ping in case it went under the radar, as this is the last PR to review and merge before cutting the next release :slightly_smiling_face:

core/stdlib/std.ncl
```diff
@@ -1566,15 +1566,31 @@
 
         ```nickel
         ('foo | std.enum.Tag) =>
-          `foo
+          'Foo
```
jneem, line 1569:
```suggestion
          'foo
```

core/stdlib/std.ncl
```diff
@@ -1566,15 +1566,31 @@
...
+          'FooBar
         ("tag" | std.enum.Tag) =>
           error
         ```
       "%
       = std.contract.from_predicate is_enum_tag,
 
+    Enum
+      | doc m%"
+        Enforces that the valie is an enum (either a tag or a variant).
```
jneem, line 1580:
```suggestion
        Enforces that the value is an enum (either a tag or a variant).
```

core/stdlib/std.ncl
```diff
@@ -1566,15 +1566,31 @@
...
+
+        # Examples
+
+        ```nickel
+        ('Foo | std.enum.Enum) =>
+          'foo
```
jneem, line 1586:
```suggestion
          'Foo
```

jneem (approved):
Sorry, I somehow missed this one

Issue #1466, opened 2023-07-20
Opaque Contract type


Radvendii:
It used to be that contracts were all functions (morally) of type `Label -> Dyn -> Dyn`
Now we also allow special record contracts, and we are discouraged from actually using the contracts as functions. (they should be used with `std.contract.apply` instead)

It would reduce confusion and mistakes, and allow for better error reporting and more features to have an opaque Contract type, and require calling some `std.contract.from_function` to convert from a function that used to just implicitly be a Contract.

---

**Is your feature request related to a problem? Please describe.**
- #1460 
- People not using `std.contract.apply`
- Can add extra information (e.g. "came from `std.contract.from_predicate`" means we can freely `||` or `&&` the contracts)
- Can require that any term in contract position (i.e. after `|`) must be of type Contract

**Describe the solution you'd like**
An opaque `Contract` type, with explicit conversion from a function that would right now just be a contract

**Describe alternatives you've considered**
The way we do it now works, though with the problems mentioned above. We could also add an opaque contract type, but have automatic conversion from functions `Label -> Dyn -> Dyn` when needed.

**Additional context**
This would be a breaking change, so it could only happen in Nickel 2.0

Radvendii:
@yannham I don't see a tag that signifies "breaking change"

yannham:
This was discussed in the weekly again, because of some reports similar to https://github.com/tweag/nickel/issues/1460 (on a very large codebase). A possible backward-compatible solution would be to add this new contract opaque type, but still allow implicit conversion from a function to the corresponding contract. Now, this would negate some of the benefits, as we can't complain when something in contract position is not a contract, and so we can't really solve #1460 directly.

However, we can issue a deprecation warning in those cases, and at least the users now have the possibility to move to the new way (the modification is likely to just add a call to a hypothetical `std.contract.from_function` here and there, so nothing major).

yannham:
This is more or less what's been implemented in #1987. This only concerns functions, but functions are the problematic bit (for the rest, records and static types can't be applied, and they are all valid contracts). We went the backward-compatibility route and still accept naked functions, but document this as legacy and will be deprecated.

PR #1955, opened 2024-06-12
Add a contract node to the AST

yannham:
Follow-up of #1954. Related to #1466.

This PR continues further the work of properly separating contracts in the AST in order to maintain more run-time information and be able, in the long run, to perform operations like boolean `or` on a larger class of contracts.

For now, to avoid breaking backward compatibility, we make those special node act like their non-contract counter-part: for example, this PR only adds one kind of contract, a `Predicate` - built from `std.contract.from_predicate` - which wraps a function and will behave as a function for all intent of purpose. However, as opposed to the result of `from_predicate`, this node makes it possible to remember the predicate and thus implement an `or` over several predicates-turned-into-contracts.


core/src/pretty.rs
```diff
@@ -5,6 +5,9 @@ use crate::parser::lexer::KEYWORDS;
 use crate::term::{
     pattern::*,
     record::{Field, FieldMetadata, RecordData},
+    // Because we use `Term::*`, we need to differentiate `Contract` from `Term::Contract`, so we
+    // alias the latter
+    CustomContract as ContractNode,
```
jneem, line 10: But it's called `CustomContract`, not `Contract` :thinking:

core/src/term/mod.rs
```diff
@@ -216,6 +216,21 @@ pub enum Term {
     #[serde(skip)]
     Type(Type),
 
+    /// A custom contract built using e.g. `std.contract.from_predicate`. Currently, custom
+    /// contracts can be partial identities (the most general form, which either blame or return
+    /// the value with potentital delayed checks burried inside) or a predicate. Ideally, both
```
jneem, line 221:
```suggestion
    /// the value with potential delayed checks buried inside) or a predicate. Ideally, both
```

core/src/term/mod.rs
```diff
@@ -216,6 +216,21 @@ pub enum Term {
...
+    /// would fall under then `CustomContract` node.
+    ///
+    /// For now, we only put predicates built using `std.contract.from_predicate` here.
```
jneem, line 224: I'm a bit confused as to why we'd want to support putting both forms in this node. Why now just convert everything to the general form, so there's only one representation to worry about?

core/src/term/mod.rs
```diff
@@ -884,7 +919,11 @@ impl Term {
             Term::Bool(_) => Some("Bool".to_owned()),
             Term::Num(_) => Some("Number".to_owned()),
             Term::Str(_) => Some("String".to_owned()),
-            Term::Fun(_, _) | Term::FunPattern(_, _) => Some("Function".to_owned()),
+            Term::Fun(_, _)
+            | Term::FunPattern(_, _)
+            // We could print a separate type for predicates. For the time being, we just consider
+            // it to be the function resulting of `$predicate_to_ctr pred`.
+            | Term::CustomContract(CustomContract::Predicate(..)) => Some("Function".to_owned()),
```
jneem, line 926: Since the new node is supposed to enable more precise errors, why not "Contract" here?

jneem (approved):

core/src/pretty.rs
yannham: Ah, but I initially called it `Contract`, and renaming didn't catch that, of course. I'm already working on a follow-up, so will fix it there instead.

core/src/term/mod.rs
yannham: Fixed in a follow-up coming PR.

core/src/term/mod.rs
yannham: You mean, generic custom contracts and predicates? Well, the most general form is also the one with the less information. Once you convert a predicate to a generic custom contract (exactly what `from_predicate` used to do before this PR), there's no coming back. So that won't help in implementing the `or` operator.

I think the whole idea of having this new node is to be more fine-grained, and be able to say not only "this is an entirely opaque and general function I know nothing about" but rather "this is a custom contract, supposed to be used in contract position, and furthermore it comes from predicate `xxx` or from the general partial identity form `yyy`". Converting everything to the most general form is exactly what we used to do (without a dedicated node), basically.

jneem: Makes sense, thanks for the explanation!

PR #1961, opened 2024-06-14
Bump comrak

jneem:


yannham (approved):
Oh, this was easier than expected. Thanks for taking care of this!

PR #1970, opened 2024-06-19
Introduce validators for building custom contracts

### LINKED ISSUE #1957: Variant of `std.contract.from_predicate` to specify an error message

It's a bit verbose and unintuitive to write a custom contract as a partial identity, passing a label around and returning the original value in case of success. For simple (and eager) contracts, sometimes we just want to return "accept" or "reject". This is in essence the role of `std.contract.from_predicate`, which takes a boolean function and make a contract out of it.

However, one limitation is precise error reporting: because predicates can only return `true` or `false`, they can't distinguish between different error cases with dedicated error messages. While predicates work well for simple contracts, where the user can just tell from the contract's name and the reported faulty value what went wrong, it doesn't scale to more complex cases. In those cases, the user is back to writing `fun label value => ...`.

The addition of enum variants open a new possibility in the middle: we can specify a contract as an eager, boolean-like predicate, but with additional data attached, such as e.g. an error message, notes, etc. I suspect the vast majority of contracts, that are currently written as general partial identities, are actually simple predicates that could be written using those "extended" predicates, which are simpler to write and more intuitive.

Another benefit of having user writing contracts this way instead of general function, beyond being more intuitive for users, is that given #1466 and the ongoing work on boolean combinators, more contracts would be amenable to boolean operations like `or` (because those extended predicates can trivially be converted to predicates). This is not the case of the same contract written as a partial identity.

While I'm not set on the exact data that can be returned, I think the most natural approach is to simply pass anything that can go into a label: a message, notes, etc. Using default values or optional, we can make it easily backward compatible if we ever add more stuff to labels (at the cost of not being a purely static type anymore). So, I would like to add a function to `std.contract`:

```nickel
bikeshedme_from_validator : [| 'Ok, 'Error { message | String, notes | Array String | default = [] } |] -> __Contract
```

yannham:
Depends on #1964. Closes #1957.

## Motivation

There's currently two constructors for custom contracts: `from_predicate`, which takes a boolean predicate and turn it into a contract, and `custom`, which takes a general partial identity.

The former is nice because it's simple to write (and will support boolean operations). However, it can't customize error messages, which is important for developer experience.

The latter is the most general, but it's not very natural to write, as we need to call blame and use label helpers to attach error data, and because we need to return the original value instead of just "ok" or "not ok".

This PR introduces a third form, in the middle, called validators. Validators leverage the enum variants of Nickel to cover the case of eager contracts (like predicates) but with all the customization power of `custom`. Basically, instead of returning a boolean, a validator returns `[| 'Ok, 'Error <ErrorData> |]` where `ErrorData` mimics the structure of a label, but as a normal Nickel record. Because it's eager, it's trivially convertible to a predicate, so validators will be usable with boolean operators.

The hope is that many concrete cases are actually validator, which are simpler to write than general custom contracts, and have much better properties (they can be negated, or-ed, etc.).

## Content

This PR:

- Adds a new custom contract variant in the AST, `Validator`
- Adds the corresponding primop to build them, `%contract/from_validator%`, as well as the corresponding stdlib wrapper `std.contract.from_validator`
- Update the examples and the stdlib to use validators whenever possible, instead of general custom contracts
- Update the contracts section of the manual to introduce validators


core/src/term/mod.rs
```diff
@@ -216,23 +216,24 @@ pub enum Term {
...
-    ///
-    /// The reason for having a separate node is that we can leverage the metadata for example to
-    /// implement a restricted `or` combinator on contracts, which needs to know which contracts
-    /// are built from predicates, or for better error messages in the future when parametric
-    /// contracts aren't fully applied ([#1460](https://github.com/tweag/nickel/issues/1460)).
-    ///
-    /// Also, custom contracts aren't supposed to be applied using the standard function
-    /// application, because we need to perform additional bookkeeping upon application. So there's
-    /// no strong incentive to represent them as naked functions.
+    /// A custom contract built.
```
jneem, line 219:
```suggestion
    /// A custom contract.
```

core/src/term/mod.rs
```diff
@@ -401,12 +402,16 @@ pub enum BindingType {
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum CustomContract {
-    /// A generic custom contract, represented as a partial identity function of type `Label -> Dyn
-    /// -> Dyn`.
-    PartialIdentity(RichTerm),
     /// A contract built from a predicate. The argument is a function of type
     /// `Dyn -> Bool`.
     Predicate(RichTerm),
+    /// A contract built from a validator. A validator is a function of type `Dyn -> [| 'Ok, 'Error
+    /// ReifiedLabel |}` where `ReifiedLabel` is a record with error reporting data (a message,
```
jneem, line 409:
```suggestion
    /// ReifiedLabel |]` where `ReifiedLabel` is a record with error reporting data (a message,
```

core/src/typecheck/operation.rs
```diff
@@ -562,3 +565,31 @@ pub fn get_nop_type(
...
+
+/// Return the type of a validator, which is one way of representing a custom contract. This static
+/// type is more rigid than the actual values accepted by `std.contract.from_validator`, because we
+/// can't represent optional fields in the type system. But it's ok to be stricter in statically
+/// typed code.
+///
+/// Also remember that custom contracts shouldn't appear directly in the source code of Nickel:
+/// they are built using `std.contract.from_xxx` and `std.contract.custom` functions. We implement
+/// typechecking for them mostly because we can (to avoid an `unimplemented!` or a `panic!`), but
+/// we don't expect this case to trigger at the moment, so it's of the utmost importance.
```
jneem, line 577:
```suggestion
/// we don't expect this case to trigger at the moment, so it isn't of the utmost importance.
```

core/stdlib/std.ncl
```diff
@@ -1080,6 +1090,54 @@
...
+          Build a custom contract from a validator.
+
+          # Validator
+
+          A validator is a function returning either `'Ok` or `'Error
+          {message, notes}` with an optional error message and notes to
+          display. A validator is similar to a predicate (see
+          `std.contract.from_predicate`) as both are contract implementations
+          that must decide right away if the value passes or fails (that is,
+          those contracts aren't lazy). But as opposed to predicates, but
```
jneem, line 1110:
```suggestion
          those contracts aren't lazy). But as opposed to predicates,
```

core/stdlib/std.ncl
```diff
@@ -1080,6 +1090,54 @@
...
+          validators have the additional ability to customize the error
+          reporting.
+
+          # Typing
+
+          Because Nickel doesn't currently have proper types for contracts, the
+          return type of `from_validator` is simply `Dyn` in the type signature.
+          You should think of `from_validator` as some returning a value of some
```
jneem, line 1118:
```suggestion
          You should think of `from_validator` as returning a value of some
```

jneem (approved):

core/src/term/mod.rs
yannham: Fixed in ongoing #1975.

PR #1974, opened 2024-06-21
Make the LSP configurable

thufschmitt:
Make the LSP options configurable

Take into account the configuration sent by the client through [`InitializeParams.initializationOptions`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams).

This is currently used to set the limits for the background evaluation, although it also creates the infrastructure for setting more things.

Depends on https://github.com/tweag/nickel/pull/1973 as they both touch the same code.

### DIFF
--- a/lsp/nls/src/config.rs
+++ b/lsp/nls/src/config.rs
@@ -1,0 +1,73 @@
+//! Configuration for the nls server
+use serde::{Deserialize, Serialize};
+
+use std::time::Duration;
+
+fn default_eval_timeout() -> Duration {
+    Duration::from_secs(1)
+}
+fn default_recursion_limit() -> usize {
+    128
+}
+fn default_blacklist_duration() -> Duration {
+    Duration::from_secs(30)
+}
+
+/**
+Limits to appy to the LSP background evaluator.
+If an evaluation reaches one of these limits, it will be canceled and the offending file will be
+temporarily blacklisted.
+*/
+#[derive(Debug, Deserialize, Serialize)]
+pub struct LspEvalLimits {
+    /// Time out at which to cancel the background evaluation
+    #[serde(default = "default_eval_timeout")]
+    pub timeout: Duration,
+    /// The maximum recursion level to allow in the background evaluator
+    #[serde(default = "default_recursion_limit")]
+    pub recursion_limit: usize,
+}
+impl Default for LspEvalLimits {
+    fn default() -> Self {
+        LspEvalLimits {
+            timeout: default_eval_timeout(),
+            recursion_limit: default_recursion_limit(),
+        }
+    }
+}
+
+/// The configuration of the LSP evaluator
+#[derive(Debug, Deserialize, Serialize)]
+pub struct LspEvalConfig {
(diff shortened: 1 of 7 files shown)


lsp/nls/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use std::time::Duration;
+
+fn default_eval_timeout() -> Duration {
+    Duration::from_secs(1)
+}
+fn default_recursion_limit() -> usize {
+    128
+}
+fn default_blacklist_duration() -> Duration {
```
yannham, line 12: Are those used anywhere else? It's usually more idiomatic to just inline the value in the `Default` implementation of the structure holding them.

lsp/nls/src/config.rs
```diff
@@ -0,0 +1,72 @@
+//! Configuration for the nls server
```
yannham, line 1: nitpick: NLS already means Nickel language server, so NLS server sounds odd

```suggestion
//! Configuration for Nickel Language Server
```

lsp/nls/tests/main.rs
```diff
@@ -111,3 +113,32 @@ fn reload_broken_imports() {
...
+        "{ C = fun n => if n == 0 then String else C (n - 1), res = 2 | C 5 }",
+    );
+
+    // Typecheck diagnostics. Empty because there's nothing to error on
+    let diags = harness.wait_for_diagnostics();
+    assert!(diags.diagnostics.is_empty());
+
+    // Evaluator diagnostics.
+    // These shouldn't be empty (because `C 5 == String` so `2 | C 5` is a contract
+    // violation), but are because `recursion_limit` is too low for the evaluator to be
```
yannham, line 140: Missing word in this sentence?

yannham (commented):
LGTM. Leaving the approval to Joe.

lsp/nls/src/config.rs
thufschmitt: They are (afaik) needed for serde's `default` pragma – which requires a function name if the `Default` instance for the field can't be used. Maybe there's a way around that (it's indeed quite ugly), but I couldn't find it.

lsp/nls/src/config.rs
thufschmitt:
```suggestion
//! Configuration for the Nickel nls Language server
```

? :smile:

lsp/nls/src/config.rs
yannham: I believe you can slap `#[serde(default)]` on the whole container and it will use its `Default` impl. Field defaults are used only when you want to have default values for some fields, but not all.

thufschmitt: Oooooh, indeed. Let me simplify that a bit then

thufschmitt: Changed in 7926278d. That's indeed much nicer

jneem (approved):

yannham:
@thufschmitt I'm not sure how to handle the conflict here

PR #1975, opened 2024-06-21
Split contracts into an immediate part and a delayed part

yannham:
Depends on #1970 #1978.

Continuing the preliminary work toward a `one_of` combinator, it appeared that explicitly separating contracts into an immediate part (more or less a predicate) and a delayed part (the partial identity) is very compelling, both theoretically and semantically, and as a user interface. It makes the behavior of boolean contract combinators trivial to explain and to implement.

This PR thus changes both the documentation and the representation of general custom contracts to adhere to this vision: that is, a custom contract is now a pair `immediate: Option<RichTerm>, delayed: Option<RichTerm>`. A nice consequence is that the Rust implementation is simplified: instead of having a lot of ad-hoc cases (predicates, validators, custom, delayed) and corresponding primops, they can all be encoded as general custom contract using only one primop constructor `%contract/custom% : Option ImmediatePart -> Option DelayedPart -> Contract`. Their treatment is more uniform throughout the code base.

### Builtin contracts

Ideally, all builtin contracts - that is the one derived from records and from static types - would be represented this way, or at least have explicit immediate and delayed part.  This is the role of the new primop `contract/get_immediate` and `contract/get_delayed`. For custom contracts, they can just extract the corresponding component. The next step is that they also work on records and static types. However, this requires a non trivial rewrite of builtin contracts, and how they are generated from static types. While migrating builtin contracts to the new paradigm will be required to have boolean operators that work on them, it's not strictly necessary - for now, we keep them as naked functions, which still works. Because this PR is already big, we'll defer the migration of builtin contracts to follow-up work.

### Parametrized contracts

 Currently, the immediate part can only return `'Ok` (early success), `'Proceed` (success but apply delayed part) and `'Error` (early failure). This means that, to be able to apply another contract, one needs to implement it in the delayed part. For example, the Nullable contract.

The consequence is that although we could (we have all the required information) very much compute a purely immediate contract for `Nullable Number`, we currently can't. I think one possibility is to add a new return value for the immediate part, `'Apply contract`, which  would indicates to the contract system that the immediate part should proceed with a contract application. The immediate part of the resulting contract would thus be the (obvious) composition of both immediate parts, and we would have that `Nullable {foo | String}` has a delayed part, but `Nullable Number` doesn't.

This is left for future work as well.

### DIFF
--- a/doc/manual/contracts.md
+++ b/doc/manual/contracts.md
@@ -198,10 +198,10 @@ error: contract broken by a value
 "foo"
 ```
 
-### Generic custom contracts
+### General custom contracts
 
 In some situations, even validators aren't sufficient. For example, when writing
-[lazy contracts](#laziness) or [parametrized
+[delayed contracts](#delayed-contracts) or [parametrized
 contracts](#parametrized-contracts), you might not be able to decide right away
 if a value satisfies a contract. Please refer to the aforementioned sections for
 more details. In this case, the most general form for creating a custom contract
@@ -229,9 +229,10 @@ A general custom contract is a function of two arguments:
 - A `label`.
 - The value being checked.
 
-Upon success, the contract must return the original value. We will see
-the reason why in the [laziness section](#laziness). To signal failure,
-a custom contract uses `std.contract.blame` which takes the label as an argument.
+Upon success, the contract must return the original value. We will see the
+reason why in the [dealyed contracts section](#delayed-contracts). To signal
+failure, a custom contract uses `std.contract.blame` which takes the label as an
+argument.
 
 The label is a special object that is automatically passed to the custom
 contract by the interpreter and which contains tracking information for error
@@ -382,9 +383,10 @@ function `Contract label value`, but use `std.contract.apply Contract label
 value` instead.
 
 Parameters that are also contracts are usually generic. In particular, they are
-not necessarily predicates or validators. In consequence, a contract
-parametrized by another unknown contract must usually be written using the most
-general constructor `std.contract.custom`.
+not necessarily immediate contracts. In consequence, a contract parametrized by
+another unknown contract must usually be written as [a delayed
+contract](#delayed-contracts), using the most general constructor
+`std.contract.custom`.
 
 One example is the `Nullable` contract, which accepts a value that is either
 `null` or of some other given format:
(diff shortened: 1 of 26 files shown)


doc/manual/contracts.md
```diff
@@ -782,134 +717,200 @@ error: contract broken by the value of `fail`
...
+      message | String | optional,
+      notes | Array String | optional
+    }
+  |]
+  ```
+
+- the *delayed part* which returns the value with delayed checks integrated
+    inside. Those checks will fire only when further data is requested. This is
+    the second argument of `std.contract.custom`. The delayed part is a function
+    taking a label, the value, and return the augmented value, of type:
```
jneem, line 758:
```suggestion
    taking a label, the value, and return the augmented value, of type:
```
```suggestion
    taking a label, the value, and returning the augmented value; it has type:
```

doc/manual/contracts.md
```diff
@@ -782,134 +717,200 @@ error: contract broken by the value of `fail`
...
+  Dyn -> Dyn -> Dyn
+  ```
+
+Take the record contract `{foo | FooContract}`:
+
+- the immediate part of this contract will check that the value is a record and
+    that its only field is `foo`.
+- the delayed part will push is the `FooContract` contract (more
+    precisely map lazily) inside the value. The resulting enriched value is
+    returned.
```
jneem, line 770: Maybe it's clearer (especially the delayed part) if you say what happens when applying `{foo | FooContract}` to the value `{foo = bar}`. The second point could be something like "the delayed part will push the `FooContract` contract inside the record, returning `{ foo = (bar | FooContract) }`"

jneem (approved):
I didn't quite understand the part about needing an extra return variant for the immediate part. If we could destructure contracts back into immediate and delayed, couldn't we write Nullable like this?

```
let Nullable = fun Contract =>
  std.contract.custom
    (fun value =>
      if value == null then
        'Ok
      else
        (%contract/immediate% Contract) value
    )
    (%contract/delayed% Contract)
```

yannham:
>I didn't quite understand the part about needing an extra return variant for the immediate part. If we could destructure contracts back into immediate and delayed, couldn't we write Nullable like this?
>
>```
>let Nullable = fun Contract =>
>  std.contract.custom
>   (fun value =>
>      if value == null then
>        'Ok
>      else
>        (%contract/immediate% Contract) value
>    )
>    (%contract/delayed% Contract)
>```

I believe, in this case, that the delayed part of `Contract` will be applied in all situations. This is not what you want if `value == null` (for `Nullable {foo | Number}`, it doesn't make sense to try to map `Number` somewhere in a `null` value). I think we need to let the immediate part decide if it should short circuit and stops now, or proceed with the delayed part. The delayed part might assume that the immediate part must have run and thus doesn't need to be defensive; for `{foo | Number}`, it would assume without checks that what it gets is a record.

yannham:
Oh, wait, were talking about the `'Apply` thingy?

jneem:
Yes, exactly. (I was also initially confused about `'Ok` vs `'Proceed`, but reading the changes sorted that part out for me)

yannham:
Then you're right, you don't technically need the `'Apply` variant. It's more of an API design question: would that make writing custom contracts nicer, more intuitive and less error prone?

```nickel
let Nullable = fun Contract =>
 std.contract.custom
  (fun value =>
     if value == null then
       'Ok
     else
       (%contract/immediate% Contract) value
   )
   (%contract/delayed% Contract)
```

versus

```nickel
let Nullable = fun Contract =>
  std.contract.parametrized (fun value
    if value == null then
      'Ok
    else
      'Apply {contract = Contract}
)
```

It's worth discussing the possibility, I think.

PR #1977, opened 2024-06-25
Add a subsumption rule between record types and dictionary types

Eckaos:
Modified the subsumption function to add the subsumption rule between a record type and a dictionary type.

### DIFF
--- a/core/src/typecheck/mod.rs
+++ b/core/src/typecheck/mod.rs
@@ -2443,29 +2443,6 @@ pub fn subsumption(
         ),
         (_, _) => checked.unify(inferred_inst, state, &ctxt),
     }
-
-    /*match (
-        inferred_inst.clone().into_type(state.table).typ,
-        checked.clone().into_type(state.table).typ,
-    ) {
-        (TypeF::Record(rrows), TypeF::Dict { type_fields, .. }) => {
-            let unif_type_fields = UnifType::from_type(*type_fields, &ctxt.term_env);
-            rrows
-                .iter()
-                .fold(Ok(()), |acc, row| -> Result<(), UnifError> {
-                    match (row, &acc) {
-                        (RecordRowsIteratorItem::Row(a), Ok(_)) => subsumption(
-                            state,
-                            ctxt.clone(),
-                            UnifType::from_type(a.typ.clone(), &ctxt.term_env),
-                            unif_type_fields.clone(),
-                        ),
-                        _ => acc,
-                    }
-                })
-        }
-        (_, _) => checked.unify(inferred_inst, state, &ctxt),
-    }*/
 }
 
 fn check_field<V: TypecheckVisitor>(
--- a/core/src/typecheck/mod.rs
+++ b/core/src/typecheck/mod.rs
@@ -2420,7 +2420,52 @@ pub fn subsumption(
     checked: UnifType,
 ) -> Result<(), UnifError> {
     let inferred_inst = instantiate_foralls(state, &mut ctxt, inferred, ForallInst::UnifVar);
-    checked.unify(inferred_inst, state, &ctxt)
+    match (inferred_inst.clone(), checked.clone()) {
+        (
+            UnifType::Concrete {
+                typ: TypeF::Record(rrows),
+                ..
+            },
+            UnifType::Concrete {
+                typ: TypeF::Dict { type_fields, .. },
+                ..
+            },
+        ) => rrows.iter().fold(
+            Ok(()),
+            |acc, row: GenericUnifRecordRowsIteratorItem<_>| -> Result<(), UnifError> {
+                match (row, &acc) {
+                    (GenericUnifRecordRowsIteratorItem::Row(a), Ok(_)) => {
+                        subsumption(state, ctxt.clone(), a.typ.clone(), *type_fields.clone())
+                    }
+                    _ => acc,
+                }
+            },
+        ),
+        (_, _) => checked.unify(inferred_inst, state, &ctxt),
+    }
+
+    /*match (
+        inferred_inst.clone().into_type(state.table).typ,
+        checked.clone().into_type(state.table).typ,
+    ) {
+        (TypeF::Record(rrows), TypeF::Dict { type_fields, .. }) => {
+            let unif_type_fields = UnifType::from_type(*type_fields, &ctxt.term_env);
+            rrows
+                .iter()
+                .fold(Ok(()), |acc, row| -> Result<(), UnifError> {
+                    match (row, &acc) {
+                        (RecordRowsIteratorItem::Row(a), Ok(_)) => subsumption(
+                            state,
+                            ctxt.clone(),
+                            UnifType::from_type(a.typ.clone(), &ctxt.term_env),
+                            unif_type_fields.clone(),
+                        ),
+                        _ => acc,
+                    }
+                })
+        }
+        (_, _) => checked.unify(inferred_inst, state, &ctxt),
+    }*/
 }
 
 fn check_field<V: TypecheckVisitor>(


core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,29 @@ pub fn subsumption(
     checked: UnifType,
 ) -> Result<(), UnifError> {
     let inferred_inst = instantiate_foralls(state, &mut ctxt, inferred, ForallInst::UnifVar);
-    checked.unify(inferred_inst, state, &ctxt)
+    match (inferred_inst.clone(), checked.clone()) {
```
yannham, line 2423:
```suggestion
    match (inferred_inst, checked) {
```

Those early clones are probably not needed. You might need to put a `ref` in the pattern matching though, as in:

```rust
           UnifType::Concrete {
                typ: TypeF::Record(ref rrows),
                ..
            },
```

core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,29 @@ pub fn subsumption(
...
+        (
+            UnifType::Concrete {
+                typ: TypeF::Record(rrows),
+                ..
+            },
+            UnifType::Concrete {
+                typ: TypeF::Dict { type_fields, .. },
+                ..
+            },
+        ) => rrows.iter().fold(
```
yannham, line 2433: If you don't build anything valuable in a fold, then you can probably get rid of it and just use a good old `for` loop:

```rust
for row in rrows {
    if  let GenericUnifRecordRowsIteratorItem::Row(a) = row {
        subsumption(state, ctxt.clone(), a.typ.clone(), *type_fields.clone())?;
    }
}

Ok(())
```

core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,29 @@ pub fn subsumption(
...
+            Ok(()),
+            |acc, row: GenericUnifRecordRowsIteratorItem<_>| -> Result<(), UnifError> {
+                match (row, &acc) {
+                    (GenericUnifRecordRowsIteratorItem::Row(a), Ok(_)) => {
+                        subsumption(state, ctxt.clone(), a.typ.clone(), *type_fields.clone())
+                    }
+                    _ => acc,
```
yannham, line 2440: Now I'm thinking about something that I haven't thought about before. During type inference, we might have open record types, such as `{foo : Number, bar : Number; tail}` where the `tail` at the end means that this value can have more fields that we don't know yet.

If we try to subtype that with a dictionary, as in `{foo : Number, bar : Number; ?a} <: {_ : Number}`, we can't let the `?a` be, because it might very much be instantiated with `baz : String` at some point. So we need to close the tail of the record.

Interestingly, we can put `Dyn` in the tail currently (instead of `?a`) to say that we don't know anything about it, that those are some fields with `Dyn` type. I wonder if we could extend the type system to add other types, for example `{foo : Number, bar : Number; Number}` (or `{foo : Number, bar : Number; _ : Number}`) to say that the tail is unknown but fields must be numbers. That would probably be a better representation.

Anyhow, I'm just thinking out loud. Let's forget that, for now, I don't think we can do better than closing the tail of the record. In practice, this means that we need to also consider the cases:

- `GenericUnifRecordRowsIteratorItem::UnifVar { id, init_level}`. In that case we need to unify this var with an empty row. Here is how it's done (**note**: it's done for _enum rows_ below, but it's exactly the same for record rows; just change any "enum" to "record"):

  https://github.com/tweag/nickel/blob/9969d9a9d40958d97f323420daba79c1b910522e/core/src/typecheck/pattern.rs#L129-L133
- `GenericUnifRecordRowsIteratorItem::TailConstant(id)`. In this case we must fail. We can raise a `WithConst` error - in the future we could have a dedicated error for better messages, but for now, that will do.

yannham (commented):
It's on the right track! There's one oversight though around the tail of record types. See the comments.

We probably also want to add a bunch of tests.

yannham:
@Eckaos feel free to re-request a review from me when you think it's ready for a new round!

Eckaos:
If you think there no need to think about ```GenericUnifRecordRowsIteratorItem::TailDyn``` and ```GenericUnifRecordRowsIteratorItem::TailVar(&'a LocIdent)```, I think it's ready for a new round ! @yannham

core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,39 @@ pub fn subsumption(
...
+                match row {
+                    GenericUnifRecordRowsIteratorItem::Row(a) => {
+                        subsumption(state, ctxt.clone(), a.typ.clone(), *type_fields.clone())?
+                    }
+                    GenericUnifRecordRowsIteratorItem::TailUnifVar { id, .. } => state
+                        .table
+                        .assign_rrows(id, UnifRecordRows::concrete(RecordRowsF::Empty)),
+                    GenericUnifRecordRowsIteratorItem::TailConstant(id) => {
+                        Err(UnifError::WithConst {
+                            var_kind: VarKindDiscriminant::EnumRows,
```
yannham, line 2444:
```suggestion
                            var_kind: VarKindDiscriminant::RecordRows,
```

Copy-paste leftover I guess, but we are unifying record rows, not enum rows :slightly_smiling_face:

core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,39 @@ pub fn subsumption(
...
+        (
+            UnifType::Concrete {
+                typ: TypeF::Record(rrows),
+                ..
+            },
+            UnifType::Concrete {
+                typ: TypeF::Dict { type_fields, .. },
+                ..
+            },
+        ) => {
```
yannham, line 2433: Nitpick: in Rust, when you have a match case of the form (that is with only one pattern and a default case)

```rust
match foo {
  Pattern => /* .. some code */,
  _ => /* .. default case */,
}
```

It's more idiomatic to use an `if-let`. It's exactly the same result but save one level of indentation:

```rust
if let Pattern = foo then {
  /* some code */
}
else {
 /* default case */
}
```

core/src/typecheck/mod.rs
```diff
@@ -2420,7 +2420,39 @@ pub fn subsumption(
...
+            for row in rrows.iter() {
+                match row {
+                    GenericUnifRecordRowsIteratorItem::Row(a) => {
+                        subsumption(state, ctxt.clone(), a.typ.clone(), *type_fields.clone())?
+                    }
+                    GenericUnifRecordRowsIteratorItem::TailUnifVar { id, .. } => state
+                        .table
+                        .assign_rrows(id, UnifRecordRows::concrete(RecordRowsF::Empty)),
```
yannham, line 2441: It could be good to also reproduce the comment that was at the other place where this code was taken from, saying that because we are unifying with a ground type, we don't need to care about level and stuff. Usually `assign_rrows` is a bit low-level and one shouldn't blindly use it directly. Here we can, but for future readers, it's good to say why.

yannham (commented):
The logic looks good to me.

What's still missing before we can merge:
 - Update the documentation of `subsumption`, which isn't up to date. It says that we don't perform any form of subtyping, but we do, know. We can also include a bit on why we close the tail when it's a unification variable.
 - Add tests. You can look at `core/tests/integration`, the README, and various inputs (in the `inputs` directory). It's sufficient to just add new test `.ncl` test files somewhere in `core/tests/integration/inputs/*` and they'll automatically be picked up in tests.

We should also update the user manual, but that can probably come in a second step (that is, another pull request).

Eckaos:
@yannham It must be good !

core/src/typecheck/mod.rs
```diff
@@ -2401,16 +2401,17 @@ fn check<V: TypecheckVisitor>(
...
+/// and then simply performs unification (put differently, the subtyping relation when it is not
+/// a record/dictionary subtyping is the equality
 /// relation).
 ///
 /// The type instantiation corresponds to the zero-ary case of application in the current
 /// specification (which is based on [A Quick Look at Impredicativity][quick-look], although we
 /// currently don't support impredicative polymorphism).
 ///
-/// In the future, this function might implement a non-trivial subsumption rule.
+/// In the future, this function might implement a other non-trivial subsumption rule.
```
yannham, line 2414:
```suggestion
```

core/src/typecheck/mod.rs
```diff
@@ -2401,16 +2401,17 @@ fn check<V: TypecheckVisitor>(
 
 /// Change from inference mode to checking mode, and apply a potential subsumption rule.
 ///
-/// Currently, there is no subtyping (until RFC004 is implemented), hence this function performs
+/// Currently, there is record/dictionary subtyping, if we are not in this case we fallback to perform
 /// polymorphic type instantiation with unification variable on the left (on the inferred type),
-/// and then simply performs unification (put differently, the subtyping relation is the equality
+/// and then simply performs unification (put differently, the subtyping relation when it is not
+/// a record/dictionary subtyping is the equality
 /// relation).
```
yannham, line 2408:
```suggestion
/// a record/dictionary subtyping is the equality relation).
```

yannham (approved):
I think more tests would be welcome, in particular when applying a function such as `std.record.insert` or `std.record.get` to a record literal (put in a variable). However, I tested those locally and they work, so let's proceed with this PR and add more compelling tests in a second step.

PR #1982, opened 2024-07-01
Add record/split_pair and record/disjoint_merge

yannham:
This is preliminary work for the rewriting of native contracts using the new immediate/delayed representation introduced in #1975, although this PR is fairly well-separated (and this is why this is a small separate PR).

In order to efficiently implement the contracts for static record types, which needs to know which fields to lock in the sealed tail, we need a way to split and recombine records which is efficient and preserves field metadata. The current implementation can do without because the immediate part and the delayed part are crammed into one function, and the current implementation is also probably quite inefficient (computing 2 record-as-sets differences in pure Nickel).

This PR adds two primops:

- `record/split_pair` corresponds exactly to what the Rust code `eval::merge::split::split` is doing: separate a pair of records into 4 parts, the fields that are in one of the operands but not the other, plus the fields that are common (but which is split in two because we have to choose if we take the value and field metadata's from the left operand or from the right operand).
- `record/disjoint_merge` is a fast merge operation when both operands are known to be disjoints. It's closer to a union than a merge, although both coincide on such a special case, in that it doesn't interact with metadata or recursive definitions (although it does preserve metadata). This operation just puts everything in one record.

Those primops might be also useful to efficiently implement other operations (record differences, record projection, etc.) in the stdlibin the future.


core/src/eval/operation.rs
```diff
@@ -2580,6 +2580,148 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        call_stack: std::mem::take(&mut self.call_stack),
+                    });
+                }
+
+                // Note that because of record closurization, we assume here that the record data
+                // of each record are already closurized, so we don't really care about
+                // environments. Should that invariant change, we might get into trouble (trouble
+                // meaning undue `UnboundIdentifier` errors).
+                debug_assert!(record1.attrs.closurized && record2.attrs.closurized);
+                record1.fields.extend(record2.fields);
```
jneem, line 2710: I don't see where we're checking for disjointness. Should we?

jneem (approved):

core/src/eval/operation.rs
yannham: It's intentional. The  reasoning was that I want this to be fast and simple especially in various record contract implementation when we *know* that the arguments are disjoints by construction (for example, when re-assembling an unsealed record tail with the rest of the record body).

Checking for disjointedness isn't free, and while it's still fast in Rust land, it kinda diminishes the interest of this function. It's also ok for primops to be unsafe (that is, the caller has to enforce some invariants) IMHO. Lastly, if one ever misuse this function, the worth that can happen is that the field of the right hand side will override the one on the left-hand side (so it's kind of a directed merge). In fact maybe we should brand this primop like this, so that it's not even a prerequisite anymore that the arguments are disjoints?

PR #1983, opened 2024-07-02
Package management rfc

jneem:
[Rendered](https://github.com/tweag/nickel/blob/dcb1ebcd3633d9e2fecf9c8c59f9825e950260e6/rfcs/006-package-management.md)

### DIFF
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -12,6 +12,16 @@ Nickel code. The mechanism needs to be predictable (it should fetch the code
 that the user expects to fetch) and reliable (if it works on my machine it
 should work on your machine).
 
+Explicit non-goals are:
+- Integration with other language ecosystems; this is only for Nickel code.
+  Anyone integrating Nickel with other languages should use a polyglot build
+  system like Bazel. (However, we might consider exporting package-management
+  metadata in a format suitable for consumption by other tools.)
+- Management of system (or other non-Nickel) dependencies. You can use Nix for
+  that.
+- Any sort of processing or compilation. Nickel is an interpreted language, so
+  package management is only in charge of distributing unmodified source files.
+
 ## The manifest file
 
 We will require a manifest file in order to import packages. Manifest files must
@@ -116,6 +126,12 @@ support, as you can just put
 ```
 
 in your package's `main.ncl`, to provide "other" and "blah" as other entry points.
+
+### Alternative: entry points are top-level files
+
+Instead of hardcoding `main.ncl`, we could say that every file in the package's
+top-level directory is publicly accessible. Package authors could keep implementation
+details private by putting code in subdirectories.
 
 ## Kinds of dependencies
 
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -57,7 +57,9 @@ So an example manifest might look like
 
 ```nickel
 {
-  # ...
+  name = "demo"
+  version = "0.1.0"
+  nickel-version = "^1.0"
   dependencies = {
     foo = 'Index { package = "github/tweag/foo", version = "1.2.0" }
     bar = 'Path "../my-bar",
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -14,22 +14,22 @@ should work on your machine).
 
 Explicit non-goals are:
 
-- Integration with other language ecosystems; this is only for Nickel code.
-  Anyone integrating Nickel with other languages should use a polyglot build
-  system like Bazel. (However, we might consider exporting package-management
-  metadata in a format suitable for consumption by other tools.)
 - Management of system (or other non-Nickel) dependencies. You can use Nix for
   that.
 - Any sort of processing or compilation. Nickel is an interpreted language, so
   package management is only in charge of distributing unmodified source files.
+- Integration with other language ecosystems; this is only for Nickel code.
+  Anyone integrating Nickel with other languages should use a polyglot build
+  system like Bazel. (However, providing integration points for other tools *is*
+  in scope. More detail on that below.)
 
 ## The manifest file
 
-We will require a manifest file in order to import packages. Manifest files must
-be named `package.ncl`, and they are found by searching up from the file being
-evaluated. That is, when the user invokes `nickel export path/to/foo.ncl`, we
-look for a manifest at `path/to/package.ncl` and then at `path/package.ncl`, and
-so on.
+We will require a manifest file in order to import packages. Manifest files
+must be named `electroplate.ncl`, and they are found by searching up from the
+file being evaluated. That is, when the user invokes `nickel export path/to/
+foo.ncl`, we look for a manifest at `path/to/electroplate.ncl` and then at
+`path/electroplate.ncl`, and so on.
 
 The manifest file format is defined by the contract `std.package.Manifest`,
 which is defined as
@@ -75,7 +75,14 @@ on the advantages and disadvantages of inline imports.
 ### Alternative: toml manifest
 
 Maybe the manifest should be in some plain-data format like toml. This would
-be easier to modify programmatically.
+be easier to modify programmatically, and it would prevent people from putting
+lots of complicated code in their manifest files.
+The disadvantage of a plain-data format is that you don't get a nice contract
+for it and you can't use ADTs for the dependencies.
+
+We discussed this point in office hours, and the general sentiment was that
+it's ok to allow the manifest to be interpreted. If someone wants to use
+that power to create a ridiculously complicated manifest, that's their problem.
 
 ### Alternative: shorthand for registry imports
 
@@ -94,13 +101,9 @@ instead of
 Since we expect registry imports to be the common case, maybe it's worth having
 a shorthand?
 
-### Question: manifest file name
+### Alternative: manifest file name
 
-The name `package.ncl` was chosen to be similar to npm's `package.json` or
-stack's `package.yaml`. One problem with this is that it could be confused for
-a nickel source file. Another possibility would be to use an extension-based
-name like `<package-name>.cabal`. Or just a stranger name that's less likely to
-conflict with something real.
+Bikeshed the name "electroplate.ncl."
 
 ## Import statements
 
@@ -112,7 +115,7 @@ without quote -- in which case it imports a package.
 The `import foo` expression evaluates to the contents of `main.ncl` in `foo`'s
 root directory.
 
-### Question: other entry points?
+### Alternative: other entry points
 
 We've hardcoded `main.ncl` as the entry point of every package, but what if
 they want to expose multiple entry points? For example, node allows a package's
@@ -133,6 +136,37 @@ in your package's `main.ncl`, to provide "other" and "blah" as other entry point
 Instead of hardcoding `main.ncl`, we could say that every file in the package's
 top-level directory is publicly accessible. Package authors could keep implementation
 details private by putting code in subdirectories.
+
+### Alternative: namespace the import tool
+
+Our initial intention for packaging was to allow for the usage of multiple
+different package management tools. This RFC only proposes one such tool, but
+maybe the import syntax could be designed with other tools in mind. For example,
+it could be `import foo from electroplate` with the idea that future nickel
+versions might add, say, `import foo from nix-flake`.
+
+## CLI interface
+
+We will add a new CLI tool (called `plate`) that wraps the nickel CLI and
+adds package management. It will offer a superset of the nickel CLI's commands
+and arguments. For example, `plate eval foo.ncl` is the same as `nickel eval
+foo.ncl`, except that it reads the manifest file, prepares the dependencies, and
+makes them available to the nickel interpreter before evaluating.
+
+### Alternative: built-in package management
+
+We could build package management straight into the nickel CLI. This might be
+more convenient and more discoverable, but it comes with stability hazards:
+we might want to evolve the `plate` interface more rapidly than the `nickel` one.
+Building in package management would also bloat the nickel CLI.
+
+### Alternative: two-stage interface
+
+Rather than wrapping the original `nickel` CLI, the `plate` command could
+*prepare* the packages for `nickel`, which would be in charge of loading
+them. The workflow would then be `plate install` followed by `nickel eval foo.ncl`.
+This would be more like how `npm` or `poetry` work (whereas the wrapping
+interface is more similar to `cargo`).
 
 ## Kinds of dependencies
 
@@ -170,28 +204,58 @@ dependencies, allowing those identical versions to be used every time.
   dependencies. This is because path dependencies can change at any time, so
   they can't be meaningfully locked.
 
-What happens if we have a lock-file, but we modify the manifest? We don't want
-to be too strict about requiring the exact versions in the lock-file, or we'll
-end up forcing the user to re-create the lock-file from scratch. In this case,
-we treat the lock-file as a suggestion instead of a hard constraint: during
-resolving, when choosing the next package version to try, it picks the locked
-version first. But if the locked version leads to a conflict, it will try
-another version without complaining. If nothing has changed since the lock-file
-was created, it should always resolve the same versions.
+## Restrictions on path dependencies
+
+Path dependencies can be problematic for reproducibility, because they require
+something to be present at a given path. In order to mitigate this:
+
+- Packages imported from the index are not allowed to have path dependencies.
+- Packages imported from git can have path dependencies, but only if they point
+  within the same git repo. The lock-file treats these path dependencies as though
+  they were git dependencies.
+
+Therefore your dependency tree can have a few path-dependency subtrees, but only
+at the root.
+
+## Stale lock files
+
+What happens if there's a lock-file, but someone modifies the manifest? The
+lock-file might need to be regenerated: certainly it might need some new
+entries, but also there might be new version conflicts that require a different
+resolution. In this case, we treat the lock-file as a suggestion instead of a
+hard constraint: during resolving, when choosing the next package version to
+try, we try to pick the locked version first. But if we run into a resolution
+conflict, we allow a different version to be chosen (and notify the user
+that a package was changed).
+
+This behavior is similar to what cargo does. It has the advantage that if the
+new manifest is compatible with the old lock-file, nothing will be changed.
+
+### Alternative: interactive prompt
+
+Instead of merely notifying the user that the lock-file changed, we could
+require them to approve the changes.
+
+### Alternative: explicit regeneration
+
+Instead of automatically updating the lock-file when the manifest changes,
+we could update it only on explicit commands: `plate eval` would use an
+old lock-file if it exists, while `plate lock` would read the manifest
+and update the lock-file. This alternative is potentially more efficient
+(as `plate eval` wouldn't need to re-read the manifest, and the manifests
+of path-dependencies, on every invocation), but makes it easy to accidentally
+use out-of-date packages.
 
 ## Version compatibility and resolution
 
 How do we handle a package that gets imported multiple times in the dependency
 tree?
 
-For path and git dependencies, there isn't much choice. Dependencies
-from the registry are the most interesting. Fortunately, there are fairly
-well-established conventions for specifying ranges of versions (like ">=1.0
-<3.0", or "^1.2"). What's less clear is how to handle multiple packages with
-overlapping ranges. Some languages (e.g. python) insist that each package
-resolves to a single version across the whole dependency tree. Other languages
-allow multiple versions, keeping track of which package in the dependency tree
-needs to import which version of a package.
+For path and git dependencies, there isn't much choice. Dependencies from the
+registry are the most interesting.  Some languages (e.g. python) insist that
+each package resolves to a single version across the whole dependency tree.
+Other languages allow multiple versions, keeping track of which package in the
+dependency tree needs to import which version of a package.
 
 I think we want to allow multiple versions of a package; the alternative can be
 fragile and annoying. But then we need to figure out how many different versions
@@ -207,12 +271,20 @@ versions into semver-delimited "bins" and allows resolution to choose at most
 one version from each bin. That is, we can have a `util@2.2` and a `util@1.2` in
 the same dependency tree, but not a `util@1.2` and a `util@1.1`.
 
+### Question: how to handle pre-1.0 minor versions?
+
+Officially, semver says that pre-1.0 versions are mutually incompatible. If we
+follow this, pre-1.0 versions would never get binned together. `cargo` modifies
+the semver rules, allowing `0.x.y` versions to be binned together if they share
+the same `x`. Should we do the same?
+
 ## The registry
 
 How should we manage the global registry? There's a potential for incurring
 substantial maintenance costs here, so we should be careful.
 
-We will provide a git repo, at a hard-coded location, to serve as the registry.
+We will provide a git repo, hard-coded to live at `github.com/tweag/nickel-mine`,
+to serve as the registry.
 This repo will contain the "index", but not the actual package contents. It
 will contain one file per package, each of which contains a line per version.
 Each entry specifies the location of the package (currently required to be on
@@ -221,17 +293,16 @@ doesn't stop them from disappearing: we don't keep a copy of the actual package
 contents.
 
 The registry entries are named like "github/\<org\>/\<package\>" (where in the
-future we might support places other than github). This allows the package
-registry to automatically discover new package versions: to find the latest
-versions of "github/tweag/json-schema-lib", we simply fetch the repository at
-`github.com/tweag/json-schema-lib` and look for tags that look like version
-numbers. Initially, we will scrape packages daily in a cron job. Eventually we
-will allow people to automatically request re-scrapes of specific packages.
+future we might support places other than github). This allows us to skip
+registration and authentication: if someone has the github permissions to create
+`jneem/foo` on github, they also have permissions to create and update the
+`github/jneem/foo` nickel package.
 
-Once a package version is stored in the index, it will never be overwritten.
-If a future scrape sees that a previously existing version tag is pointing at a
-different commit, we will make a note (maybe warn someone somehow?) and keep the
-old version.
+We will provide a backend service to update packages. Users can submit a request
+to the backend, asking to publish version `0.6.5` of package `jneem/foo` at a
+specific git revision. The service will check that the github repo `jneem/foo`
+has a `v0.6.5` tag pointing at that git revision; if so, it will add it to
+the index.
 
 ### Question: should we store a content hash too?
 
@@ -240,13 +311,34 @@ contents in the future, maybe we should also store a hash of the git tree
 contents? This would allow verification of package tarballs, without needing the
 whole git repo.
 
+## Integration point: package maps
+
+Other tools (like build systems) may need to (1) consume the dependency resolution
+that we produce, or (2) plug in their own dependencies to the nickel interpreter.
+For (1), they can consume our lock-file, which will be in JSON. This means its
+format needs to be stable.
+
+For (2), we divide the package management implementation into two parts: the `plate`
+command has all the logic for consuming the manifest, fetching packages, and so on.
+The other part is to teach the nickel interpreter about "package maps," which is
+just a map associating a filesystem path to each pair
+`(package filesystem path, package-local name)`. When the nickel interpreter
+is running a file that came from the package at path `/home/jneem/foo`, and that
+file contains `import bar`, the nickel interpreter looks up `(/home/jneem/foo, bar)`
+in the package map to figure out where to import from.
+
+After `plate` fetches dependencies, it provides the interpreter with the correct
+package map to find them. Other tools with their own dependency-fetching methods
+can invoke the interpreter with a custom package map to makes those packages
+available.
+
 ## CLI support
 
 We'll need some CLI commands for handling common package-management tasks. The
 current prototype has
 
-- a `nickel package generate-lockfile` command that updates the lock-file
-- a `nickel package debug-resolution` command that prints the full recursive
+- a `plate package generate-lockfile` command that updates the lock-file
+- a `plate package debug-resolution` command that prints the full recursive
   dependency tree
 
 We probably also want
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -165,25 +165,40 @@ include the name of the package manager in the nickel source.
 
 ## CLI interface
 
-We will build package management straight into the nickel CLI. `nickel eval`,
-`nickel export`,
-and similar commands will do an additional package-management step before the
-actual evaluation. They will start by searching for an `electroplate.ncl` file.
-If one is found, we will evaluate it. We will then search for a lock-file.
-If one is found, it will be used to guide dependency resolution; if not, we will
-do dependency resolution from scratch and write out the generated lock-file.
+We will build package management straight into the nickel CLI. There will
+be a `nickel package` subcommand. This command will start by searching for an
+`electroplate.ncl` file. If one is found, we will evaluate it. We will then
+search for a lock-file. If one is found, it will be used to guide dependency
+resolution; if not, we will do dependency resolution from scratch and write out
+the generated lock-file.
 
 Once dependencies are resolved, they will be downloaded if necessary
 (git dependencies will need to be downloaded during resolution), and then
-cached. Finally, the nickel interpreter will be invoked with the data
-necessary to find the downloaded dependencies (see the section on "package
-maps" below).
+cached.
+
+Other nickel commands like
+`nickel eval` and `nickel export` will read the lockfile before the actual
+evaluation.
+
+### Alternative: do the package management updates automatically
+
+We could do the main package management steps automatically when running
+commands like `nickel eval` and `nickel export`. This would ensure that
+the lock-file is never stale: if you update the manifest and then run
+`nickel eval`, it will take manifest changes into account. This automatic
+package management step is similar to how `cargo` works, whereas `poetry`
+and `npm` require an explicit package-management command.
 
 There will be command line flags to fine-tune this behavior. For example, there
 could be a `--locked` flag that triggers a failure if the lock-file is not
 present and up-to-date, or an `--offline` flag that triggers a failure if the
 dependencies aren't already available. There could also be a flag (`--no-electroplate`?)
 to disable package-management altogether.
+
+The main concern with this approach is performance: we need to check on
+each invocation whether the manifest is up-to-date. After initial implementation,
+we can measure the performance and decide whether we want to do things
+automatically.
 
 ### Alternative: a separate CLI tool
 
@@ -319,6 +334,9 @@ snapshots, though. Haskell gets to use compile-time checks to test compatibility
 and nixpkgs just does a lot of building and testing to check whether everything
 works. Nickel being dynamic and lazy might make this hard.
 
+It seems like global snapshots are something that can be added after the fact
+(like stackage did), so we can revisit this later.
+
 ### Alternative: minimal version selection
 
 Go uses a system called "[minimal version selection]" in which all version
@@ -430,20 +448,26 @@ to get the benefits of a registry without making your code public. Therefore
 we should support alternatives to the global registry. There are at least two
 distinct use-cases:
 
-1. replacing the default registry with an alternative, for example in order to
-   host or cache it locally
-2. providing an additional registry with private code, to be used alongside
+1. providing an additional registry with private code, to be used alongside
    the global registry
+2. replacing some registry with an alternative, for example in order to
+   host or cache it locally
 
-These two use-cases should be configured differently, because the first is
-a global setting and the second is per-package. For the global setting, we
-add a `source-replacement` field to the manifest: `std.package.Manifest` becomes
+The difference between these two settings is that the private code registries
+are configured per-package: when you import the `myorg/foo` package you should
+know that it's coming from the registry hosted by `myorg` and not from
+the global registry. (It's important not to transparently overlay private
+registries on global registries, because this can lead to
+[dependency confusion attacks](https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610)).
+
+On the other hand, the registry replacements are configured globally, possibly
+on a system-wide level. The main use is to set up, say, an organization-wide mirror
+of the registry so as not to depend on external infrastructure.
+For this global setting, we add a `source-replacement` field to some system-wide
+or project-local nickel configuration file. It takes the form
 
 ```nickel
 {
-  name | String,
-  # ...
-
   source-replacement | {
     registry | { _ : String },
     git | { _ : String },
@@ -467,8 +491,8 @@ source-replacement.registry."https://github.com/nickel-lang/nickel-mine.git"
 
 You can also replace git repository locations, because why not.
 
-For per-package alternate registries, we just add a `registry` field to the
-package spec: the contract on `dependencies` becomes
+For per-package alternate registries (use-case 1 above), we just add a
+`registry` field to the package spec: the contract on `dependencies` becomes
 
 ```nickel
 { _:
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -304,6 +304,11 @@ specific git revision. The service will check that the github repo `jneem/foo`
 has a `v0.6.5` tag pointing at that git revision; if so, it will add it to
 the index.
 
+### Alternative: manual pull requests
+
+If package volume is low enough (which it probably is, at first), index updates
+could be done manually via pull requests.
+
 ### Question: should we store a content hash too?
 
 We're storing a git tree hash in the index, but if we ever want to store package
--- a/rfcs/006-package-management.md
+++ b/rfcs/006-package-management.md
@@ -13,6 +13,7 @@ that the user expects to fetch) and reliable (if it works on my machine it
 should work on your machine).
 
 Explicit non-goals are:
+
 - Integration with other language ecosystems; this is only for Nickel code.
   Anyone integrating Nickel with other languages should use a polyglot build
   system like Bazel. (However, we might consider exporting package-management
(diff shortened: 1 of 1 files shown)


rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+author: Joe Neeman
+---
+
+# Package management
+
+People want to reuse code, but Nickel doesn't currently have a good way to do
+it. We should have a way to fetch packages and make them easily available to
+Nickel code. The mechanism needs to be predictable (it should fetch the code
+that the user expects to fetch) and reliable (if it works on my machine it
+should work on your machine).
```
yannham, line 13: It might sound obvious, but I think we might emphasize here clearly one of the (non)-goals: we only want to make Nickel code available to other Nickel code. We are not concerned with any kind of deeper system dependencies here, or cross-language ones (just use Nix or Bazel, I guess?)

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+
+## The manifest file
+
+We will require a manifest file in order to import packages. Manifest files must
+be named `package.ncl`, and they are found by searching up from the file being
```
yannham, line 18: I wonder if we should accept `package.toml`, `.json`, etc. Maybe it would be good hygiene to not have arbitrary Nickel code in here, as it could lead to some obfuscation. But `package` is so pervasive that we might try to find another word. We can just try to find a name for the package manager and use that instead.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+evaluated. That is, when the user invokes `nickel export path/to/foo.ncl`, we
+look for a manifest at `path/to/package.ncl` and then at `path/package.ncl`, and
+so on.
+
+The manifest file format is defined by the contract `std.package.Manifest`,
+while is defined as
```
yannham, line 24:
```suggestion
which is defined as
```

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+  name | String,
+  version | Semver,
+  nickel-version | Semver,
+
+  dependencies
+    | { _:
+      [|
+        'Path String,
+        'Git { url | String, branch | optional | String, rev | optional | String },
+        'Index { name | String, version | Semver },
```
yannham, line 37: Ok, this makes it harder to have a .toml file. Another solution is to emit a warning when the file isn't a full normal form, and ask the user if they want to continue. Or something like that. On the other hand `flake.nix` can have pretty complex logic, so maybe it's ok.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+a nickel source file. Another possibility would be to use an extension-based
+name like `<package-name>.cabal`. Or just a stranger name that's less likely to
+conflict with something real.
+
+## Import statements
+
+The manifest file assigns a name to each dependency; to import the dependency
+named `foo` you simply write `import foo`. That is, an `import` statement either
+takes a string in quotes -- in which case it imports a path -- or an identifier
+without quote -- in which case it imports a package.
```
yannham, line 99: I think there were discussions to make the import syntax more general, like (say we name our package manager `metalist`) `import foo from metalist`. Then, if we ever provide other integrations later on, you could do `import foo from flake` or something like that. What I like with the general import is that we're going the Nix route of names are resources that can be abstracted (or at least their resolution). Do you think that this will ever be useful? How does it fit in the current package management story?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+support, as you can just put
+
+```nickel
+{
+  other = import "./other.ncl",
+  blah = import "./blah.ncl",
+}
+```
+
+in your package's `main.ncl`, to provide "other" and "blah" as other entry points.
```
yannham, line 118: Another possibility would be to make all the top-level files separate entry points. But since we don't have public/private modifiers, it might prove annoying, as it makes it harder to hide internal modules, so to speak.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+  within the same repository, or for temporary patching of published packages.
+
+We require packages to have their own manifest file (at the package root), even
+if they don't import dependencies.
+
+## Lock files
+
+In order to ensure reproducibility across time and across machines, we build
+a lock-file (if there isn't yet one) when running `nickel eval` or `nickel
+export`. The lock-file specifies the exact versions of all (transitive)
```
yannham, line 143: If I understand correctly, this means that we entirely tie the interpreter and the package manager (of course, the interpreter needs to have some knowledge of the package manager, because it needs to find the file on the disk - but that interface can be very simple and light weight, as in I give you a name, you give me a concrete path). I wonder if we should also consider separating them, with Nickel complaining about a missing lockfile (and asking to run `metalist` or whatnot). Although it does have a cost in term of user experience

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+- For a repository dependency, the version specifier might allow for a range of
+  versions. The lock-file will record the exact version used.
+- For a path dependency, the lock-file will record that there was a path
+  dependency, but it won't record anything about it and it will ignore recursive
+  dependencies. This is because path dependencies can change at any time, so
+  they can't be meaningfully locked.
+
+What happens if we have a lock-file, but we modify the manifest? We don't want
+to be too strict about requiring the exact versions in the lock-file, or we'll
+end up forcing the user to re-create the lock-file from scratch. In this case,
```
yannham, line 158: I'm not sure about this one. Not respecting the lockfile sounds like a hug avenue for reproducibility bugs (that is clearly stated as an overarching goal in the introduction: if it works on my machine, it works on your machine). Do you have example of such behavior in other mainstream language-specific package managers?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+every time a package gets imported, solving the dependency graph is easy.
+But it increases the chance of getting incompatibilities at runtime: we might
+accidentally get a value from `util@1.1` and try to pass it to an incompatible
+function defined in `util@1.2`. Pulling in too many different versions also
+increases the total number of packages in the dependency graph.
+
+The current prototype uses a strategy similar to cargo: it divides package
+versions into semver-delimited "bins" and allows resolution to choose at most
+one version from each bin. That is, we can have a `util@2.2` and a `util@1.2` in
+the same dependency tree, but not a `util@1.2` and a `util@1.1`.
```
yannham, line 191: That sounds like a reasonable tradeoff indeed.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,243 @@
...
+doesn't stop them from disappearing: we don't keep a copy of the actual package
+contents.
+
+The registry entries are named like "github/\<org\>/\<package\>" (where in the
+future we might support places other than github). This allows the package
+registry to automatically discover new package versions: to find the latest
+versions of "github/tweag/json-schema-lib", we simply fetch the repository at
+`github.com/tweag/json-schema-lib` and look for tags that look like version
+numbers. Initially, we will scrape packages daily in a cron job. Eventually we
+will allow people to automatically request re-scrapes of specific packages.
```
yannham, line 212: I'm not sure to understand exactly when this process takes place. Is it to ease the addition of a new package in the registry (automatic discovery)? Or is it more than that, like re-generating the index with auto-updates for existing pacakges? In the latter case, isn't that a bit fragile if it's entirely automated but based on heuristics?

yannham (commented):
First, thanks a lot for putting that together. I think it's clear and well-articulated. Something that might be improved a bit (I left a related comment) is to clearly state the goal and non-goal of what we're trying to achieve here, to delimit the scope more explicitly.

rfcs/006-package-management.md
jneem: Cargo has this behavior (which is where I first learned it from). For example, if I make a new crate with a dependency on `serde = "1"` then it will generate a lock-file with `serde` 1.0.203. If I then edit my `Cargo.toml` to specify `serde = "=1.0.202"` and run `cargo build`, it will automatically re-create the lock-file with `serde` version 1.0.202.

I think the reason this doesn't kill reproducibility too badly is that it only gets triggered if a manifest file changes. If I have a working lock-file and you copy my files to your machine, nickel's first attempt to resolve versions will try the same ones as in the lock-file. Those choices will all succeed, and so you'll get the exact same versions as me. The "lock-file-as-a-suggestion" path only gets triggered if there was a conflict when trying to use the lock-file's versions.

Unfortunately I don't know enough about other package managers' corner cases to know how widespread this behavior is. I think poetry doesn't do it, because it only modifies the lock-file when you explicitly ask it to. I'll investigate some others...

rfcs/006-package-management.md
jneem: I was thinking initially of manual package addition + automatic discovery of new versions. The tag-scanning part might be too fragile, I agree. The only alternative I can think of would require a bit more effort but it might actually not be too bad: instead of running a cron job, have a service where you can POST to `nickel-mine.org/release/github/tweag/json-schema-lib?v=0.1.0&rev=df1234`, and if the specified repo has the '0.1.0' tag pointing at revision df1234 it will publish that version.

rfcs/006-package-management.md
jneem: So the current code structure of the prototype is:
- `nickel-lang-core` gets a new `PackageMap` type, which basically just answers the question "if package `x` imports package `y`, what path should it come from"
- there's a new crate `nickel-lang-package` that has all the logic of package fetching and version resolution. It depends on `nickel-lang-core`, and is capable of producing a `PackageMap` from the resolution.
- `nickel-lang-cli` adds a dep on `nickel-lang-package` to integrate the package management part. I've put some package-management tools under `nickel package xyz`, and added package support to `eval` and `export`

As an alternative, `nickel-lang-cli` could remain untouched and we could put the package-management utilities under, say, `metalist`.

There is one little annoyance related to path dependencies, because the lock-file doesn't currently include their recursive dependencies. This means that we (currently) need to do some resolution on-the-fly just before evaluation. So with the current state of things it would be hard to support the interface

```console
metalist lock
nickel export foo.ncl
```

and we'd probably need an interface that wraps nickel, like `metalist export foo.ncl`

rfcs/006-package-management.md
jneem: Sounds reasonable, I'll add it as an alternative. I think it isn't too hard to hide private modules -- just put them in a subdirectory.

rfcs/006-package-management.md
jneem: Yeah, the ability to have a nice contract was my main motivation for proposing a nickel manifest. There are definitely some advantages to a plain-data format, though...

rfcs/006-package-management.md
jneem: Ah, this is a good point. The syntax part is easy, but I think the main question is how integrated package management should be. I can see two main options:
1. package management is integrated into the interface (either built-in to the nickel cli, or by having a package-management cli that wraps the nickel cli)
2. package management is loosely coupled to nickel, where a package manager is expected to have some interface that nickel can use to query the map from package names to paths.

I think the choice mainly depends on how many different package managers we expect to have. If it's just `metallist` and `flake`, then maybe it makes sense to have them built-in. Last time we discussed this, I was under the impression that we could piggy-back off existing package repositories like `npm`. But since that seems not to be the case, I think there probably won't be many alternatives. (Which might be a good thing, as one of the concerns about the other plan was in fracturing the ecosystem)

rfcs/006-package-management.md
yannham: Ok, so maybe a different way of saying it, is that if there is a conflict between the manifest and the lockfile, the lockfile is automatically re-generated. Which I think might be ok.

About cargo, I really don't like that `cargo install` from a published crate entirely ignores the lockfile for example (even if it agrees with the manifest), and that has been a bad user experience to try to understand why it's been failing with `install` but not with `build` locally. While cargo is great, we should make sure we aren't too biased and we don't just blindly copy their decisions just because we work in Rust these days (and constraints might be quite different for a Rust package and a Nickel package, although to be fair, the lockfile/manifest sync sounds like a very generic problem)

jneem: > Ok, so maybe a different way of saying it, is that if there is a conflict between the manifest and the lockfile, the lockfile is automatically re-generated.

One subtlety here is that the lockfile isn't regenerated from scratch. It's regenerated by trying to run the previous resolution, but then backtracking when conflicts occur. Most of the time this should result in a fairly small change to the resulting resolution, as opposed to re-resolving from scratch which might bump lots of versions.

rfcs/006-package-management.md
jneem: It was intentional, rather than accidental, but it's possible I'm falling into the trap of copying cargo's behavior too much...

There are two inherent problems with path dependencies: they can change at any time, and they might not have the same contents across machines. This means that "locking" them doesn't do much to help reproducibility. It also means that the path dependencies are just as mutable as the top-level manifest -- every time we read and evaluate the top-level manifest we also need to read and evaluate all the path-dependency manifests. So I guess there are two possibilities:

- with package management integrated into the cli (as in the current prototype) path dependencies should not be locked, and their manifests should be re-read and re-resolved on every `nickel eval`
- with package management as a separate tool, path dependencies should be locked (because `nickel eval` will consume the lock-file and needs to know about them). Every time `metallist` is run, the path dependencies should be re-resolved. If the path dependencies are modified in between running `metallist` and running `nickel`, you'll just get an error. (Same will happen if you modify the top-level manifest.)

rfcs/006-package-management.md
jneem: Initial version could use pull requests to the index for updating.

rfcs/006-package-management.md
jneem: We could ask for confirmation before changing the resolution (or print out the changes)

jneem:
I've updated the RFC based on feedback here and discussion at the weekly.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,357 @@
...
+`path/electroplate.ncl`, and so on.
+
+The manifest file format is defined by the contract `std.package.Manifest`,
+which is defined as
+
+```nickel
+{
+  name | String,
+  version | Semver,
+  nickel-version | Semver,
```
GuillaumeDesforges, line 41: Why is it needed?
Should this be a version constraint? e.g. `>=1`, `^1`, `>1.1.3,<2` etc?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,357 @@
...
+    | default
+    = {},
+}
+```
+
+So an example manifest might look like
+
+```nickel
+{
+  # ...
```
GuillaumeDesforges, line 60: Can the example demonstrate all fields? 
```suggestion
  name = "demo"
  version = "0.1.0"
  nickel-version = "^1.0"
```

rfcs/006-package-management.md
```diff
@@ -0,0 +1,357 @@
...
+```nickel
+    foo = 'Index { package = "github/tweag/foo", version = "1.2.0" }
+```
+
+Since we expect registry imports to be the common case, maybe it's worth having
+a shorthand?
+
+### Alternative: manifest file name
+
+Bikeshed the name "electroplate.ncl."
```
GuillaumeDesforges, line 106: typo
```suggestion
Bikeshed the name "electroplate.ncl".
```

rfcs/006-package-management.md
```diff
@@ -0,0 +1,357 @@
...
+
+Our initial intention for packaging was to allow for the usage of multiple
+different package management tools. This RFC only proposes one such tool, but
+maybe the import syntax could be designed with other tools in mind. For example,
+it could be `import foo from electroplate` with the idea that future nickel
+versions might add, say, `import foo from nix-flake`.
+
+## CLI interface
+
+We will add a new CLI tool (called `plate`) that wraps the nickel CLI and
```
GuillaumeDesforges, line 150: Funnily enough, Nickel Package Manager would be... `npm` :skull_and_crossbones:

rfcs/006-package-management.md
```diff
@@ -0,0 +1,357 @@
...
+adds package management. It will offer a superset of the nickel CLI's commands
```
GuillaumeDesforges, line 151: This seems to be a huge argument in favor of embeding package management in nickel's CLI.

At some point, I can hardly imagine anybody _not_ using package management. That would mean everybody would use `plate` instead of `nickel` anyways.

rfcs/006-package-management.md
jneem: Ah, right. I guess `version` should be `SemverConstraint` (or `VersionConstraint`). `nickel-version` is the minimum nickel interpreter version required by the package. Maybe `min-nickel-version` is clearer?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+## Stale lock files
+
+What happens if there's a lock-file, but someone modifies the manifest? The
+lock-file might need to be regenerated: certainly it might need some new
+entries, but also there might be new version conflicts that require a different
+resolution. In this case, we treat the lock-file as a suggestion instead of a
+hard constraint: during resolving, when choosing the next package version to
+try, we try to pick the locked version first. But if we run into a resolution
+conflict, we allow a different version to be chosen (and notify the user
+that a package was changed).
```
aspiwack, line 231: An alternative is to never do any dependency resolution at all: target a particular commit of the registry, and just use the latest version of each package according to said commit. This is the Nixpkgs way (and also Stackage). It has pros and cons, but it has a fairly clear story for lockfile updates (_i.e._ just create lockfile entries for the new packages, never touch the existing package (unless asked to by the user, of course)).

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+follow this, pre-1.0 versions would never get binned together. `cargo` modifies
+the semver rules, allowing `0.x.y` versions to be binned together if they share
+the same `x`. Should we do the same?
+
+## The registry
+
+How should we manage the global registry? There's a potential for incurring
+substantial maintenance costs here, so we should be careful.
+
+We will provide a git repo, hard-coded to live at `github.com/tweag/nickel-mine`,
```
aspiwack, line 288: Or `github.com/nickel-lang/nickel-mine`.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+to serve as the registry.
+This repo will contain the "index", but not the actual package contents. It
+will contain one file per package, each of which contains a line per version.
+Each entry specifies the location of the package (currently required to be on
+github) and its git tree hash. This ensures that packages are immutable, but it
+doesn't stop them from disappearing: we don't keep a copy of the actual package
+contents.
```
aspiwack, line 295: This representation strategy means that the manifest files (hence the packages' metadata) isn't kept in the registry. It means that querying the registry can be quite expensive since it requires many calls to many Git repositories. Is this reasonable? (genuine question, I don't really have an opinion on this)

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+We will provide a backend service to update packages. Users can submit a request
+to the backend, asking to publish version `0.6.5` of package `jneem/foo` at a
+specific git revision. The service will check that the github repo `jneem/foo`
+has a `v0.6.5` tag pointing at that git revision; if so, it will add it to
+the index.
+
+### Alternative: manual pull requests
+
+If package volume is low enough (which it probably is, at first), index updates
+could be done manually via pull requests.
```
aspiwack, line 312: Opam manages very well by pull requests. The cli can automate writing a pull request for you. If the volume becomes too high, we can also automate merging the pull request. This sounds way easier than having to maintain a backend service. So I'm heavily biased toward this alternative.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+format needs to be stable.
+
+For (2), we divide the package management implementation into two parts: the `plate`
+command has all the logic for consuming the manifest, fetching packages, and so on.
+The other part is to teach the nickel interpreter about "package maps," which is
+just a map associating a filesystem path to each pair
+`(package filesystem path, package-local name)`. When the nickel interpreter
+is running a file that came from the package at path `/home/jneem/foo`, and that
+file contains `import bar`, the nickel interpreter looks up `(/home/jneem/foo, bar)`
+in the package map to figure out where to import from.
```
aspiwack, line 335: There are two big technical missing bits in your RFC:
- How does the `nickel` command line understand packages? This is your package-map bit, but extended beyond the interpreter: how do we provide said package map to the interpreter to begin with?
- Where/how are the packages stored on disk (I think this is easier than for most programming languages since we store the packages verbatim, so we don't have this thing where the same version of the same package is actually two different artefact because they were built with two different compilers).

aspiwack (commented):
My initial comments (I'm coming at this completely fresh, I haven't read the discussion above)

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+
+Maybe the manifest should be in some plain-data format like toml. This would
+be easier to modify programmatically, and it would prevent people from putting
+lots of complicated code in their manifest files.
+The disadvantage of a plain-data format is that you don't get a nice contract
+for it and you can't use ADTs for the dependencies.
+
+We discussed this point in office hours, and the general sentiment was that
+it's ok to allow the manifest to be interpreted. If someone wants to use
+that power to create a ridiculously complicated manifest, that's their problem.
```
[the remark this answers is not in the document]
jneem: Oh, this is one important thing I left out. The index entry in the registry (which is in json, not nickel) contains the dependency information from the manifest. So resolving dependencies for registry packages doesn't require evaluating their manifests. The slow evaluation only matters when they create their PR for uploading their package to the registry.

rfcs/006-package-management.md
jneem: This is also related to the thing I left out, that the registry index has enough information for dependency resolution. We don't store the whole manifest in the registry, but enough that we can do version resolution just by querying the index (and manifests from git/path dependencies, of course).

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+top-level directory is publicly accessible. Package authors could keep implementation
+details private by putting code in subdirectories.
+
+### Alternative: namespace the import tool
+
+Our initial intention for packaging was to allow for the usage of multiple
+different package management tools. This RFC only proposes one such tool, but
+maybe the import syntax could be designed with other tools in mind. For example,
+it could be `import foo from electroplate` with the idea that future nickel
+versions might add, say, `import foo from nix-flake`.
```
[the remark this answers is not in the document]
yannham: I think the idea isn't that we switch package manager entirely for a fixed set of pavkages, but rather that some packages are managed by package manager A and we want to add other packages are managed by package manager B, typically provided as a bare flake output. More of a multi package providers scenario. At least when we looked at not having a proper package manger just for Nickel but relying on existing ones. Maybe it's less relevant now.

rfcs/006-package-management.md
yannham: The lack of dependency resolution in flakes leads people to implement their own layer (https://discourse.nixos.org/t/introducing-flakehub/32044) which the usual lot of community split, bikeshedding and drama. I don't know about the Stack situation, though.

rfcs/006-package-management.md
yannham: Yes, I think this was the general feeling when discussed orally in the weekly: this is the simplest and most reasonable way to start.

rfcs/006-package-management.md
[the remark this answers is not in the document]
aspiwack: But you might have to switch package manager anyway. At any rate, I don't think the _tool_ used to provide package is relevant for namespacing.

rfcs/006-package-management.md
yannham: I think that the flakes and Stack models rely on the existence of a globally coherent versioned set of packages (is that what Stackage is)? In Nixpkgs those are NixOS releases. I agree this is usually great from a user experience point of view, but also a non negligible burden put on the people in charge of curating, updating and releasing this set.

aspiwack: Yes, Stackage is a collection of so-called snapshots. Each snapshot has at most one version of each package, which are tested to be able to build together.

I think one of the biggest obstacle that you'd have for Nickel is probably the testing bit. Since you don't necessarily have types to guarantee that everything works together well (and evaluating everything like Nixpkgs does is pretty prohibitive).

Possibly the approach of allowing several copies of the same package alleviates a large chunk of the issues with bounds and resolution. But I think this alternative should at least be addressed in the document.

PS: if we go with version bounds, we need to have an answer to the question: what is the process when we discover after the fact that a bound is too permissive?

rfcs/006-package-management.md
jneem: Fair enough. I guess the index ought to include whatever metadata we want to be easily queryable.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+
+The manifest file assigns a name to each dependency; to import the dependency
+named `foo` you simply write `import foo`. That is, an `import` statement either
+takes a string in quotes -- in which case it imports a path -- or an identifier
+without quote -- in which case it imports a package.
+
+The `import foo` expression evaluates to the contents of `main.ncl` in `foo`'s
+root directory.
+
+### Alternative: other entry points
```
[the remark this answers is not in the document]
jneem: > as long as imports aren't evaluated at all unless forced, because main.ncl may not exist

This might be an issue, because currently we do eagerly parse and typecheck imports.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,359 @@
...
+and arguments. For example, `plate eval foo.ncl` is the same as `nickel eval
+foo.ncl`, except that it reads the manifest file, prepares the dependencies, and
+makes them available to the nickel interpreter before evaluating.
+
+### Alternative: built-in package management
+
+We could build package management straight into the nickel CLI. This might be
+more convenient and more discoverable, but it comes with stability hazards:
+we might want to evolve the `plate` interface more rapidly than the `nickel` one.
+Building in package management would also bloat the nickel CLI.
```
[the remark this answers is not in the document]
jneem: Ok, I'm redoing that section with the built-in package management as the proposed solution, and a separate command as an alternative.

rfcs/006-package-management.md
jneem: - I've added some more details to the section on the CLI interface
- I added an extra section at the end on how we store packages

rfcs/006-package-management.md
jneem: I added an "Alternative" section about global snapshots as a placeholder, but it's probably worth discussing it in more detail in the weekly.

Anecdotally, I find that the rust defaults (dependencies use the "^" mode by default, and you can have multiple copies of a package that don't share the same semver bin) work pretty well. The point is that the default dependency bounds align with the no-duplication bounds, and so if those bounds are too restrictive then you just end up with multiple copies of a package. 

The bounds are usually not too permissive, because they are then someone has done semver wrong. But if it does happen, you can often work around it by pinning a dependency. For example, suppose I have a dependency on `dep`, which in turn depends on `dep2 ^0.1.0`, but then `dep2` releases a broken `0.1.1`. I can force version resolution to fall back to the old version of `dep2` by adding a dependency on `dep2 =0.1.0`.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+Packages will be updated by pull requests to the index. There can be some tooling
+in the nickel CLI to automate the creation of these pull requests. If the volume
+of updates becomes high enough, we can automate the merging.
+
+### Question: should we store a content hash too?
+
+We're storing a git tree hash in the index, but if we ever want to store package
+contents in the future, maybe we should also store a hash of the git tree
+contents? This would allow verification of package tarballs, without needing the
+whole git repo.
```
aspiwack, line 362: Storing a content hash is a good idea. But there's more than one way to do so, so you may end up not getting much more compatibility than you planned.

That being said, there is one content-hash standard, namely SWHID from our good friends at Software Heritage. It's specified here https://www.swhid.org/specification/v1.1/ . Maybe worth looking into it?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+
+Anything else?
+
+## Where do the packages live?
+
+Index and git dependencies need to be downloaded before they are used. We will
+use a single global cache directory for all dependencies. It will use the
+[`directories`](https://docs.rs/directories/latest/directories/struct.ProjectDirs.html#method.cache_dir)
+crate to deduce a platform-appropriate location (`~/.cache/nickel-lang/` on
+linux, unless `$XDG_CACHE_HOME` is set). Within this directory, git checkouts
```
aspiwack, line 411: Do we need some garbage-collection/cache-eviction policy?

rfcs/006-package-management.md
aspiwack: I'm sorry, my question was ambiguous. Here's the scenario that I have in mind: we have packages A<-B uploaded in the registry. And we discover that the latest version of B, says it's compatible with versions of A which it actually isn't compatible with (whatever the reason). We now need to update the registry somehow. It could be by altering the how the current version of B depends on A. Or it can be by marking this version of B as incorrect somehow, so that it's never picked up by the solver (and uploading a new version).

But, one way or another, we need to make sure that the bad combination is forbidden, otherwise, it'll be possible in the future for the solver to pick that combination, even if there are newer versions of B.

We need policy and workflows for to handle this case.

aspiwack (commented):
Thanks for the updates. A few extra comments below.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+{
+  name | String,
+  version | Semver,
+  nickel-version
+    | Semver,
+    | doc "The minimum version of nickel supported by this package",
+
+
+  dependencies
+    | { _:
```
aherrmann, line 47: should the key be restricted to string?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+### Alternative: inline dependencies
+
+Nix and Dhall allow for importing dependencies dynamically, using things like
+`fetchGit`. There was some discussion
+[here](https://github.com/tweag/nickel/issues/329#issuecomment-967372858)
+on the advantages and disadvantages of inline imports.
+
+### Alternative: toml manifest
+
+Maybe the manifest should be in some plain-data format like toml. This would
```
aherrmann, line 82: I didn't see it mentioned explicitly, so I wanted to raise it. Will `import` be forbidden in the manifest itself? I would imagine that that would be a good idea. But I haven't thought about it very deeply.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+(git dependencies will need to be downloaded during resolution), and then
+cached. Finally, the nickel interpreter will be invoked with the data
+necessary to find the downloaded dependencies (see the section on "package
+maps" below).
+
+There will be command line flags to fine-tune this behavior. For example, there
+could be a `--locked` flag that triggers a failure if the lock-file is not
+present and up-to-date, or an `--offline` flag that triggers a failure if the
+dependencies aren't already available. There could also be a flag (`--no-electroplate`?)
+to disable package-management altogether.
```
aherrmann, line 186: Such features would be great for integration with other package managers or build systems.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+
+Officially, semver says that pre-1.0 versions are mutually incompatible. If we
+follow this, pre-1.0 versions would never get binned together. `cargo` modifies
+the semver rules, allowing `0.x.y` versions to be binned together if they share
+the same `x`. Should we do the same?
+
+## The registry
+
+How should we manage the global registry? There's a potential for incurring
+substantial maintenance costs here, so we should be careful.
```
aherrmann, line 332: Are you considering support for custom registries?

When Bazel introduced its new dependency manager bzlmod they also allowed users to define custom registries and even use multiple registries. Commercial users often like this because they can place proprietary code into their own private registry.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+dependency tree needs to import which version of a package.
+
+I think we want to allow multiple versions of a package; the alternative can be
+fragile and annoying. But then we need to figure out how many different versions
+to allow. There's a trade-off: if we allow pulling in a different version
+every time a package gets imported, solving the dependency graph is easy.
+But it increases the chance of getting incompatibilities at runtime: we might
+accidentally get a value from `util@1.1` and try to pass it to an incompatible
+function defined in `util@1.2`. Pulling in too many different versions also
+increases the total number of packages in the dependency graph.
```
aherrmann, line 301: Just for reference, the Bazel bzlmod solution is to use [Go's MVS](https://research.swtch.com/vgo-mvs) and fail if they encounter incompatible version requests for the same package. But, the root package can declare [single or multi version overrides](https://bazel.build/external/module#overrides).

rfcs/006-package-management.md
```diff
@@ -0,0 +1,425 @@
...
+
+## Integration point: package maps
+
+Other tools (like build systems) may need to (1) consume the dependency resolution
+that we produce, or (2) plug in their own dependencies to the nickel interpreter.
+For (1), they can consume our lock-file, which will be in JSON. This means its
+format needs to be stable.
+
+For (2), we divide the package management implementation into two parts: the `plate`
+command has all the logic for consuming the manifest, fetching packages, and so on.
```
aherrmann, line 372: Build systems like Bazel prefer to do the fetching themselves. Is the intention that `plate` can just produce the URLs for dependencies but leave the fetching to an external tool?

aherrmann (commented):
Very interesting proposal!
I've added some comments from the Bazel perspective.

rfcs/006-package-management.md
jneem: I think the nickel dict type always has string keys; the underscore in `{_ : x}` is just part of the notation for a dict.

rfcs/006-package-management.md
jneem: Good question, I had originally thought that the manifest wouldn't support package management, but it would be allowed to import normal paths. But maybe let's start by forbidding imports altogether.

rfcs/006-package-management.md
jneem: Good point, I'll look into that.

rfcs/006-package-management.md
jneem: Yes, I think we can have a `nickel package list-all-urls` (or something) subcommand that outputs the information needed to fetch dependencies. Would that meet the needs of build systems?

rfcs/006-package-management.md
jneem: That's a very interesting read, thanks! This approach and the one of allowing duplicate semver-compatible versions seem to be pretty much mutually exclusive, so I guess we need to pick one or the other?

rfcs/006-package-management.md
jneem: Maybe, but it might also be ok to punt on it. The local cache doesn't have an index, so it's safe to just delete things manually.

rfcs/006-package-management.md
aherrmann: I guess the user controlled override is compatible with either approach. But, otherwise, yes, I think so.

rfcs/006-package-management.md
jneem: Ah, right. I'll add a section on that. I like the "mark B as incorrect" option, because it can also be used to work around other kinds of breakage.

rfcs/006-package-management.md
jneem: It looks like SWHID is very easy to support. As far as file and tree identifiers go (which I think are the ones we care about; the registry only needs to identify the tree, not the whole history), it's 100% git-compatible. That is, if the tree if is `d198bc9d7a6bcf6db04f476d29314f157507d505` then the SWHID is
`swh:1:dir:d198bc9d7a6bcf6db04f476d29314f157507d505`

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+accidentally get a value from `util@1.1` and try to pass it to an incompatible
+function defined in `util@1.2`. Pulling in too many different versions also
+increases the total number of packages in the dependency graph.
+
+The current prototype uses a strategy similar to cargo: it divides package
+versions into semver-delimited "bins" and allows resolution to choose at most
+one version from each bin. That is, we can have a `util@2.2` and a `util@1.2` in
+the same dependency tree, but not a `util@1.2` and a `util@1.1`.
+
+### Alternative: global snapshots à la Stackage
```
yannham, line 308: Reading again this section, I wonder if version resolution and snapshots are necessarily incompatible. We could get started with version resolution but at some point provide a snapshot to take stuff from, I suppose. A bit like `foo.workspace = true` in cargo, but with `snapshot-foo` instead of `workspace`.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+  This amounts to being opinionated that "the ecosystem is using semver, and your
+  version requirements should comply with semver." It's basically a semver analogue
+  of go's policy of only allowing lower bounds (plus pinning at the top level).
+
+### Question: how to handle pre-1.0 minor versions?
+
+Officially, semver says that pre-1.0 versions are mutually incompatible. If we
+follow this, pre-1.0 versions would never get binned together. `cargo` modifies
+the semver rules, allowing `0.x.y` versions to be binned together if they share
+the same `x`. Should we do the same?
```
yannham, line 368: I think so. Otherwise I don't understand why pre-1.0 version have both a minor and a patch version, if they don't have any useful distinction semantically.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+Package yanking will be done with the same process as package publishing
+(i.e., probably PRs to the index). See the next section.
+
+### Alternative: allow yanked packages if resolution would otherwise fail
+
+A slightly more permissive alternative would be to ignore yanked packages that
+aren't in the lock-file, but if resolution fails then go back and try allowing
+them. This would allow resolution to succeed, for example, if your manifest
+asks for the latest version of a package *and* that package was yanked *and*
+you don't have a lock-file. But this seems like maybe adding too many corner-cases.
```
yannham, line 389: I wonder how useful this is. That is, I don't expect it would happen a lot that there's a unique version that can be picked and it's a yanked version. Often either the yanked version is immediately followed by a backward-compatible update - minor or patch -, or there it wasn't a `X.0.0` version and there's other versions of the same major version. I think a yanked version would be the only choice if people started to depend on this specific version too fast (otherwise a previous version of the same bin could still be picked) AND the correction is a breaking change (otherwise the more up-to-date correction could still be picked).

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+distinct use-cases:
+
+1. replacing the default registry with an alternative, for example in order to
+   host or cache it locally
+2. providing an additional registry with private code, to be used alongside
+   the global registry
+
+These two use-cases should be configured differently, because the first is
+a global setting and the second is per-package. For the global setting, we
+add a `source-replacement` field to the manifest: `std.package.Manifest` becomes
```
yannham, line 440: I think I don't understand: you say it's a global setting, but we still configure it at the level of the manifest of a package? Or by global do you mean it applies to all the dependencies of one package,  instead of potentially one of the  dependency of one package?

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+```nickel
+source-replacement.registry."https://github.com/nickel-lang/nickel-mine.git"
+  = "/path/to/a/local/git/repo"
+```
+
+or
+
+```nickel
+source-replacement.registry."https://github.com/nickel-lang/nickel-mine.git"
+  = "https://my-site/nickel-mine"
```
yannham, line 465: Should we have any form of security here? Adding a source-replacement has the potential to substitute an existing package with pretty much anything. The security model is not as bad in Nickel as in say Nix, which can install arbitrary executable or inject hidden malware in your glibc, but in Nix you do need to add a public key and a there is mechanism to make sure the user really intended to use a substituter (need to be a trusted user, and will ask on the CLI if you want to use the substituters configured at the flake level).

The answer might be that we don't care in our case, I'm thinking out loud here.

rfcs/006-package-management.md
```diff
@@ -0,0 +1,560 @@
...
+
+## CLI interface
+
+We will build package management straight into the nickel CLI. `nickel eval`,
+`nickel export`,
+and similar commands will do an additional package-management step before the
+actual evaluation. They will start by searching for an `electroplate.ncl` file.
+If one is found, we will evaluate it. We will then search for a lock-file.
+If one is found, it will be used to guide dependency resolution; if not, we will
+do dependency resolution from scratch and write out the generated lock-file.
```
yannham, line 174: I wonder what's the cost of doing that on each evaluation. As opposed to compiled languages, we might have to pay the dependency resolution step on each run, including for small-ish script (as opposed to say `cargo` when you pay the price at build time but not at run time). Do you think that would be an argument in favor of better separated package-management, where `nickel eval` would just pick the existing lockfile or fail but not perform the update automatically? Or at least an argument for not updating the lockfile automatically? Or do you expect those costs to be negligible?

rfcs/006-package-management.md
jneem: I *think* it isn't a security risk because source replacement is only configurable at the top-level: it can't be triggered by some package deep in your dependency tree. If the top-level package wants to be malicious, it probably has sneakier ways of being so.

rfcs/006-package-management.md
jneem: Good question, I guess we'd have to measure it to be sure. I expect version resolution to be quick, especially if there's an up-to-date lockfile but I don't actually have data.

Having to manually trigger a re-resolution is definitely an option; I think poetry and npm do this. I personally find it a bit annoying, as it's an extra step that's easy to forget.

yannham (approved):
After the last review, and discussion in team meeting, we decided that the RFC is in a good state and answered most direction questions. We shall go forward with this version (although @jneem has minor updates to do still before merging).

PR #1984, opened 2024-07-02
More aggressive type/contract deduplication on hover

### LINKED ISSUE #1972: Filter out the contract annotations on hover/completion when they don't add anything

The LSP very helpfully displays all the types/contracts attached to values or fields on hover.
However, it's easy for this value to get a `Dyn` type, and unless there's no other information, that `Dyn` is just noise:

image

Likewise, a value might get a contract applied to it several times (or have both a type, and the same type applied as a contract):

image

Filter out these annotations as they don't bring anything to the table

jneem:
Fixes #1972.

While testing this, I noticed that some field documentation was missed when hovering over the exact field that was annotated, so I fixed that too.

### DIFF
--- a/lsp/nls/src/requests/hover.rs
+++ b/lsp/nls/src/requests/hover.rs
@@ -161,7 +161,9 @@ pub fn handle(
     if let Some(hover) = hover_data {
         let mut contents = Vec::new();
 
-        // Collect all the type and contract annotations we can find. (We don't distinguish between them.)
+        // Collect all the type and contract annotations we can find. We don't distinguish between them
+        // (and we deduplicate annotations if they're present as both types and contracts). However, we
+        // do give some special attention to the inferred static type if there is one: we list it first.
         let mut annotations: Vec<_> = hover
             .metadata
             .iter()
--- a/lsp/nls/src/requests/hover.rs
+++ b/lsp/nls/src/requests/hover.rs
@@ -156,57 +156,47 @@ pub fn handle(
     if let Some(hover) = hover_data {
         let mut contents = Vec::new();
 
-        // If we can't determine a static type through the typechecker because we are outside of a
-        // statically typed block, but the term points to a definition with a type annotation, we
-        // use this annotation insted.
-        let mut type_annots: Vec<_> = hover
+        // Collect all the type and contract annotations we can find. (We don't distinguish between them.)
+        let mut annotations: Vec<_> = hover
             .metadata
             .iter()
-            .filter_map(|m| Some(m.annotation.typ.as_ref()?.typ.to_string()))
+            .flat_map(|m| {
+                let maybe_typ = m.annotation.typ.as_ref().map(|typ| typ.typ.to_string());
+                let contracts = m
+                    .annotation
+                    .contracts
+                    .iter()
+                    .map(|c| c.label.typ.to_string());
+                maybe_typ.into_iter().chain(contracts)
+            })
+            .chain(
+                hover
+                    .values
+                    .iter()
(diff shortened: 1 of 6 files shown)


lsp/nls/src/requests/hover.rs
```diff
@@ -156,57 +160,47 @@ pub fn handle(
...
             .iter()
-            .filter_map(|m| Some(m.annotation.typ.as_ref()?.typ.to_string()))
+            .flat_map(|m| {
+                let maybe_typ = m.annotation.typ.as_ref().map(|typ| typ.typ.to_string());
+                let contracts = m
+                    .annotation
+                    .contracts
+                    .iter()
+                    .map(|c| c.label.typ.to_string());
+                maybe_typ.into_iter().chain(contracts)
```
yannham, line 174: You have an `iter()` function on `TypeAnnotation` that chains the type and then the contracts, so that you don't have to do it manually here (https://github.com/tweag/nickel/blob/3bafeff167b9130b43567f235a627dbad9d29d76/core/src/term/mod.rs#L758)

lsp/nls/src/requests/hover.rs
```diff
@@ -156,57 +160,47 @@ pub fn handle(
...
-            .iter()
-            .flat_map(|m| &m.annotation.contracts)
-            .chain(hover.values.iter().flat_map(annotated_contracts))
-            .map(|contract| contract.label.typ.to_string())
-            .chain(type_annots)
-            .collect();
+        // There's no point in repeating the static type in the annotations.
+        if let Some(idx) = annotations.iter().position(|a| a == &ty) {
+            annotations.remove(idx);
+        }
```
yannham, line 196: I don't understand why we remove the static type from the annotations.  `annotations` is just the list of the type, followed by all contracts, sorted and deduped, right? So, I would expect to find the static type in it, and that it's intentional. Why remove it?

lsp/nls/src/requests/hover.rs
```diff
@@ -156,57 +160,47 @@ pub fn handle(
...
+            .chain(
+                hover
+                    .values
+                    .iter()
+                    .flat_map(annotated_contracts)
+                    .map(|contract| contract.label.typ.to_string()),
+            )
             .collect();
+        annotations.sort();
+        annotations.dedup();
```
yannham, line 185: I'm not entirely sure it's ok to sort the annotations. The order of contracts is meaningful in Nickel (different order of application might give a different result). Additionally, if there's no static type, the first contract of the list is taken to be the static type, which also makes a difference statically, and not only at runtime.

lsp/nls/src/requests/hover.rs
yannham: Ok, I get it. You include the static type in the annotations so that it's deduped (if there is the same as a contract), but you still keep the distinction static type/contracts, so you display the static type separately, and don't want to repeat it in the annotations.

Which I think is the right behavior (the static type is special in that it's the one selected e.g. by the typechecker). Maybe you could add a short mention explaining this (we don't distinguish between types and contracts for the dedup but we will in the LSP answer) to the comment attached to `annotations` ?

yannham (approved):

lsp/nls/src/requests/hover.rs
jneem: Yes, exactly. I've added a comment so hopefully it's less mysterious now

lsp/nls/src/requests/hover.rs
jneem: Ok, I'm deduping them without sorting now.

PR #1986, opened 2024-07-04
flake: remove redundant transitive dependencies

jeremyschlatter:
I added the `github:tweag/nickel` flake to a project today and noticed that it pulled in a bunch of redundant transitive dependencies. This commit cleans those up. The impact is visible in the lock file, which now has 13 fewer `nodes`:

```
nickel on master fsh ❯ python -c 'import json; print(len(json.load(open("flake.lock"))["nodes"]))'
29
```

```
nickel on flake-inputs fsh ❯ python -c 'import json; print(len(json.load(open("flake.lock"))["nodes"]))'
16
```

Things done:

- added several more `follows` relationships for de-duplication

- updated to latest rust-overlay flake which drops flake-utils dependency

- set `follows = ""` for a few unused dependencies. See https://github.com/NixOS/nix/issues/7807.

### DIFF
--- a/core/src/error/suggest.rs
+++ b/core/src/error/suggest.rs
@@ -24,7 +24,7 @@ where
     S: AsRef<str>,
     I: AsRef<str>,
 {
-    let mut max = std::f64::MIN;
+    let mut max = f64::MIN;
     let mut arg_max: Option<&'syms str> = None;
 
     if symbols.is_empty() {
--- a/core/src/identifier.rs
+++ b/core/src/identifier.rs
@@ -301,7 +301,9 @@ mod interner {
             // deallocation, so references are valid until the arena drop, which is tied to the
             // struct drop.
             let in_string = unsafe {
-                std::mem::transmute(self.arena.lock().unwrap().alloc_str(string.as_ref()))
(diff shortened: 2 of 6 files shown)


yannham (commented):
Thanks, it does trim the flake.lock quite a bit.

What do you think about https://github.com/NixOS/nix/issues/7807#issuecomment-1427986568, related to `follows = ""`? I'm a bit reluctant with those kind of hacks, because if somebody else has to handle a CI breakage (say, upon a flake.lock update) and don't know about it, it can turn out to be painful.

jeremyschlatter:
I was conflicted about it. I think it's quite reasonable to not use `follows = ""` for that reason. I pushed a commit that removes it. You can see the diff here: https://github.com/tweag/nickel/pull/1986/commits/43125b506831187632cc4034c934d6c207f6a1fd

yannham:
@jeremyschlatter Thanks. I believe there's still one empty follows for the `flake-compat` input of `pre-commit-hooks`.

jeremyschlatter:
@yannham Whoops, thanks for catching that. Fixed.

yannham (approved):

PR #1987, opened 2024-07-04
Re-introduce unified representation for custom contracts, migrate builtin contracts

yannham:
#1975 introduced a new model for custom contracts, where they are split between an immediate and a delayed part.

The motivation is to eventually introduce boolean combinators (`any_of`, `not`, etc.), and also to make the user interface of contracts more friendly.

This PR started as a migration of all the builtin contracts to the new model. However, it didn't exactly go as expected.

### Data flow

Some cases in the stdlib required to pass data from the immediate part to the delayed part. Of course, this data can be recomputed, because the delayed part gets the original value as well, but those computations might be expensive: in our case, it's the projection of a record onto a record type (separating the tail from the rest), which is supposedly at least linear in the size of the records.

The original fix was to have a new possible return value for the immediate part, `'Ok {payload = _}`, and call the delayed part with this payload, so the delayed part becomes `fun payload label value => ...`. This adds complications in explaining this interface in the manual, and I originally kept two version of `std.contract.custom`, one without payload and one with payload.

What's more, the contract for record types makes more sense as one block. Separating it into the immediate part and the delayed part felt a bit artificial and made the logic harder to follow (although the forced separation is kinda the point of the current line of work).

### Contract composition

The split representation makes contract composition tricky:

```nickel
let Nullable = fun Contract =>
  std.contract.custom
    (fun value =>
      if value == null then
        'Done
      else
        'Ok
     )
    (std.contract.apply Contract)
  )
```

This is a natural definition of `Nullable`, but it's not the right one: `Nullable Number` is now a delayed contract (running the immediate part only tells you if the value is null, but not if it's a number, while it could be known immediately). So the right definition would be something like:

```nickel
let Nullable = fun Contract =>
  std.contract.custom
    (fun value =>
      if value == null then
        'Done
      else
        (std.contract.get_immediate Contract) value
     )
    (std.contract.get_delayed Contract)
  )
```

as noticed by @jneem in #1975, but this is more cumbersome and not very intuitive. 

### Implementation complexities

The split design also just makes for much more code in the implementation. Several things are duplicated for immediate and delayed part, each builtin contract now needs two internal implementations (the immediate and the delayed), this requires also more stdlib functions (`get_delayed`, `get_immediate`, `custom_piped`, etc.), and the design with `%contract/custom%` taking nullable parts was questionable (impossible to type statically, null tests in the internals module, etc.).

All in all, after the first round of migrating builtin contracts to the split representation, and migrating it back to the new representation of this PR, the diff was something like (before test fixes and documentation updates):

```
12 files changed, 685 insertions(+), 1147 deletions(-)
```

which does hint at the fact that the new approach is just simpler.

## A combined representation

This PR adopts a different representation. The fundamental observation is that we can encode the immediate/delayed distinction in the return type of a custom contract, using an enum:

```
CustomContractImp := Label -> Dyn -> [| 'Ok Dyn, 'Error {..} |]
```

This type is a mix of the type of the immediate part and the delayed part introduced in #1975. It takes both a value and a label, but return an enum. The idea is that immediate errors are reported using the return value (instead of blaming the label), which makes them catchable and subject to boolean operations. When the immediate checks succeed, the contract can return `'Ok`, but also provides the new value with the delayed checks. Now, it's quite easy to implement `get_immediate` and `get_delayed` in term of this unique representation:

```nickel
get_immediate : CustomContractImpl -> Validator = fun impl value =>
  impl null value
  |> match {
    'Ok _ => 'Ok,
    error => error,
  }

get_delayed: CustomContractImpl -> Delayed = fun impl label value =>
  impl label value
  |> match {
    'Ok value => value,
    error =>
      let label_with_error_data = [...] in
      std.contract.blame label_with_error_data,
  }
```

There is the minor detail that we need to provide a label in the immediate part, and we give `null`, which should be fine because the label shouldn't be used in the immediate check, but we can't predict that still. It's irrelevant anyway; we actually don't need proper `get_immediate` and `get_delayed` functions. The point is that this contract representation encodes the same information and is sufficient to implement boolean operations.

### Composition

With the all-smashed-in-one design, the naive composition has already the right properties:

```
let Nullable = fun Contract =>
  std.contract.custom
    (fun label value =>
      if value == null then
        'Ok
      else
        std.contract.apply_as_custom Contract label value
    )
```

There is one caveat: we can't use `std.contract.apply` here. I mean, we could:

```
let Nullable = fun Contract =>
  std.contract.custom
    (fun label value =>
      if value == null then
        'Ok
      else
        'Ok (std.contract.apply Contract label value)
    )
```

But we would be in the same situation as before: we artificially make `Contract` entirely delayed. So, this PR adds a `apply_as_custom`, whose name is to be bikeshed, which applies a contract but doesn't convet `'Error` to throwing blame error like `apply` does. Instead, it preserve the catchable errors and the immediate part of composition of contracts.

## Content

This PR changes the representation of custom contracts to the new one, and migrates all builtin contracts to the new representation. It adds some new primops and stdlib function, and update the documentation as well.

### Record contracts

Record contracts application was conveniently implemented just as a variant of the merge operation. This PR needs to adapt this code, because `merge` would signal immediate errors by returning a `BlameError`, which makes it delayed/non catchable. One possibility would have been to implement part of the contract in pure Nickel, but this is wasteful: merge has to perform a field splitting operation anyway, so checking for extra field immediately is basically free. Doing that in user-code would be duplicating this work, or would need to make yet a new version of merge that can take a partially prepared result.

In fact, it was simpler to just make the existing variant of merge to adapt to the custom contract signature: it wraps the result in `'Ok` and signal immediate errors by generating an `'Error` term.

## Future

- bikeshed the name of `std.contract.apply_as_custom`
- Implement the boolean operators on contracts

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -1314,7 +1314,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                 }),
                 Term::Record(_) => Ok(Closure {
                     body: mk_app!(
-                        stdlib::internals::record_immediate(),
+                        stdlib::internals::immediate::record(),
                         RichTerm { term: t, pos }
                     )
                     .with_pos(pos_op_inh),
@@ -1336,7 +1336,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                 }),
                 Term::Record(_) => Ok(Closure {
                     body: mk_app!(
-                        stdlib::internals::record_delayed(),
+                        stdlib::internals::delayed::record(),
                         RichTerm { term: t, pos }
                     )
                     .with_pos(pos_op_inh),
@@ -2595,17 +2595,26 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     return mk_type_error!("Record", 2, t2.into(), pos2);
                 };
 
-                // As for merge, we refuse to combine two records if one of them has a sealed tail
-                if let Some(record::SealedTail { label, .. }) =
-                    record1.sealed_tail.or(record2.sealed_tail)
-                {
-                    return Err(EvalError::IllegalPolymorphicTailAccess {
-                        action: IllegalPolymorphicTailAction::Merge,
-                        evaluated_arg: label.get_evaluated_arg(&self.cache),
-                        label,
-                        call_stack: std::mem::take(&mut self.call_stack),
-                    });
-                }
+                // As for merge, we refuse to combine two records if one of them has a sealed tail.
+                // However, if only one of them does, because we don't do any recursive
+                // re-evaluation here, it's fine to just pick this tail as the tail of the result.
+                //
+                // This behavior is actually useful, because disjoint_merge is used in the
+                // implementation of builtin contracts to combine an unsealed tail with the
+                // original body of the record. In that case, the unsealed tail might have an
+                // additional sealed tail itself (tail can be sealed multiple times in a nested
+                // way), and the right behavior is to just keep it.
+                let sealed_tail = match (record1.sealed_tail, record2.sealed_tail) {
+                    (Some(record::SealedTail { label, .. }), Some(_)) => {
+                        return Err(EvalError::IllegalPolymorphicTailAccess {
+                            action: IllegalPolymorphicTailAction::Merge,
+                            evaluated_arg: label.get_evaluated_arg(&self.cache),
+                            label,
+                            call_stack: std::mem::take(&mut self.call_stack),
+                        })
+                    }
+                    (tail1, tail2) => tail1.or(tail2),
+                };
 
                 // Note that because of record closurization, we assume here that the record data
                 // of each record are already closurized, so we don't really care about
@@ -2619,10 +2628,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     Term::Record(RecordData {
                         fields: record1.fields,
                         attrs: record1.attrs,
-                        // We need to set the tail to `None` explicitly to appease the borrow
-                        // checker which objects that `sealed_tail` has been moved in the if
-                        // let above.
-                        sealed_tail: None,
+                        sealed_tail,
                     }),
                     pos_op_inh,
                 )))
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -1561,7 +1561,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     label.arg_pos = self.cache.get_then(idx.clone(), |c| c.body.pos);
                     label.arg_idx = Some(idx);
 
-                    // Otherwise, we push the label back on the stack and we properly convert the
+                    // We push the label back on the stack and we properly convert the
                     // contract to a naked function of a label and a value, which will have the
                     // stack in the same state as if it had been applied normally to both
                     // arguments.
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -26,7 +26,6 @@ use crate::{
     position::TermPos,
     serialize,
     serialize::ExportFormat,
-    stdlib,
     stdlib::internals,
     term::{
         array::{Array, ArrayAttrs, OutOfBoundError},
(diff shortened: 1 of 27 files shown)


jneem (approved):
I didn't get through internals.ncl and std.ncl yet, but I have to run and I'll be OOO until next week. I'll give it the green button in case @vkleen is busy and you don't want to wait...

core/src/term/mod.rs
```diff
@@ -216,10 +216,50 @@ pub enum Term {
...
+    /// forces values)
+    ///
+    /// The delayed part is a partial identity which takes a label and the value and either blames or
+    /// return the value with potential delayed checks buried inside.
+    ///
+    /// Note that this is a conceptual distinction. It did happen that we experimented with making
+    /// this distinction explicit, with custom contracts being represented by two different
+    /// functions, one for each part. But this proved to be cumbersome in many ways (both for us
+    /// language developers and for users). Instead, we decided to make custom contracts just one
+    /// function of type `Label -> Dyn -> [| 'Ok, 'Ok Dyn, 'Error {..} |]`, which gives enough
```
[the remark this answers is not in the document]
yannham: We don't currently, thanks for catching this. I went through several design iterations

core/src/eval/operation.rs
```diff
@@ -1564,75 +1528,119 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
                         EvalError::NotEnoughArgs(3, String::from("contract/apply"), pos_op)
                     })?;
-                    let mut l = l.clone();
-                    l.arg_pos = self.cache.get_then(idx.clone(), |c| c.body.pos);
-                    l.arg_idx = Some(idx);
 
+                    label.arg_pos = self.cache.get_then(idx.clone(), |c| c.body.pos);
+                    label.arg_idx = Some(idx);
+
+                    // Otherwise, we push the label back on the stack and we properly convert the
```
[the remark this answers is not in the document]
yannham: The code was moved around, and I think at some point it was right after the special casing for `Type`. I guess I can just get rid of the "otherwise" now.

yannham, line 1564:
```suggestion
                    // We push the label back on the stack and we properly convert the
```

core/src/term/mod.rs
yannham, line 248:
```suggestion
    /// function of type `Label -> Dyn -> [| 'Ok Dyn, 'Error {..} |]`, which gives enough
```

yannham:
Thanks @jneem. I think this PR can wait a bit (and I was OOO as well anyway). It touches several core parts so it's probably better to wait for your second pair of eyes on the stdlib and internals changes. Although the internals are a bit intricate sometimes, so I understand that some parts might be hard to fully assess in one sweeping code review, but it's still good to have a look on it.

core/stdlib/std.ncl
```diff
@@ -1371,39 +1364,62 @@
...
+          where `label` is inserted by the Nickel interpreter.
 
           Type: `Contract -> Label -> Dyn -> Dyn`
           (for technical reasons, this function isn't actually statically typed)
 
-          Nickel supports user-defined contracts defined as predicates, as
-          general functions, as records, etc. The interpreter also performs additional
-          bookkeeping for error reporting when applying a contract in an
-          expression `value | Contract`. You should thus not use standard
+          Although contracts used to be naked functions a long time ago,
```
jneem, line 1399: I think there's too much discussion of the old state of affairs. The changes should be documented, but I suspect that it isn't what most people are interested in. What if this paragraph was just replaced with something like "Contracts can be constructed using the stdlib functions `std.contract.from_predicate`, `std.contract.from_validator`, or `std.contract.custom`."

And then the "Custom contracts" section below could be replaced by

# Legacy contracts

Old versions of Nickel allowed functions of type `Label -> Dyn -> Dyn` to be treated as contracts. While this is still supported for backwards compatibility, it is discouraged and will be deprecated in the future. Use the contract constructors in `std.contract` instead.

jneem (approved):

jneem:
As for the bikeshedding part, I'd want a name that emphasizes the distinction between asserting (or throwing, or panicking) and returning an error. If the name "apply" weren't already fixed, I might go with `std.contract.assert` for the throwing variant and `std.contract.check` for the non-throwing variant. Given that `apply` is already fixed, though, I'm not sure that there's a good intuitive distinction between `apply` and `check`...

Other possibilities for the non-throwing variant: `eval`, `apply_without_blaming`

core/stdlib/std.ncl
```diff
@@ -1068,49 +1046,30 @@
...
-          details.
-        "%
-      = fun delayed => %contract/custom% null delayed,
+      = fun contract => %contract/custom% contract,
 
     from_predicate
       | (Dyn -> Bool) -> Dyn
       | doc m%"
-          Build an immediate custom contract from a boolean predicate.
+          Build a custom contract from a boolean predicate.
```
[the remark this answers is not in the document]
yannham: I don't think there's a very important distinction, but they still have different shapes. Basically, until application, contracts might be record or static types as well, which aren't custom contract. That being said, `apply_as_custom` is basically a conversion to the custom contract form, so whatever you can do with builtin contracts, you can do with custom contracts and vice-versa.

doc/manual/contracts.md
```diff
@@ -835,74 +912,49 @@ value which is wrapping the original value with delayed checks inside**:
...
+We first define `with_delayed_checks`, the delayed part, that maps a function
+over the record which substitutes each field for the same value but wrapped in a
+`Bool` contract. Thanks to lazyness, the record mapping won't happen until `'Ok
+with_delayed_checks` is returned *and* some code actually uses the value. Even
+then, because records (and record mapping) are lazy, *this doesn't actually
+execute the `Bool` contracts right away*. Each contract will only be run when
+the individual fields will be accessed or exported.
+
+We proceed to the immediate checks. We first check if the value is a record and
+return `'Error {..}` otherwise. Then, we iterate over over all the record field
```
yannham, line 934:
```suggestion
return `'Error {..}` otherwise. Then, we iterate over all the record field
```

yannham:
> though, I'm not sure that there's a good intuitive distinction between apply and check...

I agree, although `check` has the slight connotation of trying something? There's `try` as well but I fear this would be dangerous because it sounds like `try` makes all errors catchable, but delayed checks might still throw later and it's not catchable. So it's a "half-try", or "try immediate, then run delayed if it works". 

`check` or `eval` would still be reasonable as an arbitrary variation of `apply` (we just need to distinguish the two).

Another possibility that came to mind is something like Rust's `and_then` (or something along the lines: `then`, `continue_with`, `proceed_with`, etc.). The idea is that the main usage of `apply_as_custom` is to be called in tail position in a custom contract implementation (hence the continue/proceed idea). So there's an idea of chaining checks. Also, as for `and_then` in Rust, it has a monad-like feeling: while `apply` just returns the original value naked, `apply_as_custom` returns a `Result<_, BlameError>`, morally.

core/stdlib/std.ncl
yannham: Right, the documentation was clumsy. In the end, the nature of contracts that can be given to `apply` are already detailed in the following section (`# Arguments`), so I just got rid of this paragraph and expanded slightly the arguments section. I trimmed the `# Custom contracts` and changed it to legacy contract as well: https://github.com/tweag/nickel/pull/1987/commits/01ca07a3e3fdc969e22c8d59166cfddd08d78522

jneem:
For the `and_then` analogy, I'd expect a signature more like `[| 'Ok .., 'Err .. |] -> Contract -> [| 'Ok .., 'Err .. |]`. What about `check_apply` or `try_apply`?

yannham:
> For the and_then analogy, I'd expect a signature more like [| 'Ok .., 'Err .. |] -> Contract -> [| 'Ok .., 'Err .. |].

This is a good point, for a Rustacean. We could also chose to use it nonetheless, as it's a bit similar to do `Ok(value)).and_then(other_contract)`. But indeed it doesn't map exactly. Or maybe just call it `then` ?

> What about check_apply or try_apply?

Honestly I don't see the advantage of `check_apply` over `apply`. Same for `try_apply`, I think it conveys the same semantics as `try`, just with more letters, so IMHO it suffers from the same problem (make it look like it can catch blame errors).

I think I'll merge this PR to avoid drifting and rebasing too much, and we'll bikeshed the name in the weekly.

PR #1990, opened 2024-07-08
Adding documentation for subtyping

Eckaos:
In this pull request, I will add the manual documentation for subtyping(Record/Dictionary, Array/Array, Dictionary/Dictionary) and an integration test to chain functions on records.


core/tests/integration/inputs/typecheck/chaining_dictionary_function_record_dictionary_subtyping.ncl
```diff
@@ -0,0 +1,8 @@
+# test.type = 'error'
+# [test.metadata]
+# error = 'EvalError::BlameError'
+let x = { foo = 1, bar =2} : {foo : Number, bar : Number} in
+let y = (std.record.insert "baz" 5 x) : { _ : Number } in
+let _ = (std.record.get "foo" y) : Number in
+let z = (std.record.remove "foo" y) : { _ : Number } in
+std.record.get "foo" z
```
yannham, line 8: I believe all the type annotations are exactly the type that should be inferred, so they can probably be omitted.

```suggestion
let result : _ =
  {foo = 1, bar = 2}
  |> std.record.insert "baz" 5 x
  |> std.record.get "foo" y
  |> std.record.remove "foo" y
  |> std.record.get "foo" z
in
result
```

doc/manual/typing.md
```diff
@@ -255,6 +255,26 @@ The following type constructors are available:
   }
   ```
 
+### Subtyping
+
+Subtyping is a relation between two types where
+you can use one of the type in place of the other one.
+In Nickel, there is :
```
yannham, line 262: The beginning is a bit dry. The best way to introduce a notion is to motivate it, with an example. For example, start by saying something like: 

> While distinct types are usually incompatible, some types might actually be safely converted to some other types. Take the following example:
>
>```nickel
>let extended : {_ : Number} =
>  let initial = {foo = 1} in
>  std.record.insert initial
>```

And then go on to explain that technically the types don't match, but it's legal to "forget" some information to go from a record type to dictionary type.

And go on to explain why this is accepted while the types don't really match specifically.

doc/manual/typing.md
```diff
@@ -255,6 +255,26 @@ The following type constructors are available:
...
+
+- **Record/Dictionary** subtyping : `{field1 : T1, ..., fieldn : Tn} <= { _ : T}`
```
yannham, line 264: Usually it's customary to use the symbol `<:`. It's also better to define it before (after the informal intro, you can say that it's call "subtyping" and we write `T <: U` if a value of type `T` can be freely used as a value of type `U`.

```suggestion
- **Record/Dictionary** subtyping : `{field1 : T1, ..., fieldn : Tn} <: { _ : T}`
```

Eckaos:
I added motivation for subtyping. I was wondering do I add constructors(Array, Dictionary) in this pull request too or do I make another pull request to update the manual ? @yannham

yannham:
I think you can add it to this PR. It feels a bit artificial to split this in several ones, given the small changes all focused on the same file.

Eckaos:
I don't understand why my array example does not work.

I didn't add to the manual Record/Record and Dyn subtyping for now because it is not implemented.

Eckaos:
I added a generic example for expanding a function to work around the fact that we don't have arrow subsumption.
Do you think, the documentation miss something else ? @yannham

doc/manual/typing.md
```diff
@@ -255,6 +255,89 @@ The following type constructors are available:
...
+let extended : { _ : Number } =
+  let initial : { foo : Number } = { foo = 5 } in
+  std.record.insert "bar" 5 initial in
+extended
+```
+
+In this example, there is a silent conversion from `{foo : Number}` to `{_ :
+Number}`. This is safe because `foo` is of type `Number`, and it's the only
+field, which means that a value of type `{foo : Number}` is effectively
+dictionary of numbers. In the typing jargon, `{foo : Number}` is said to be a
```
yannham, line 276:
```suggestion
a dictionary of numbers. In the typing jargon, `{foo : Number}` is said to be a
```

doc/manual/typing.md
```diff
@@ -255,6 +255,89 @@ The following type constructors are available:
...
+  ```
+
+  Here, `{_ : {a : Number}}` is accepted where `{_ : {_ : Number}}` is expected,
+  because `{a : Number} <: { _ : Number }`.
+
+**Remark**: if you've used languages with subtyping before, you might expect the
+presence of a rule for function types, namely that `T -> U <: S -> V` if `S <:
+T` and `U <: V`. This is not the case in Nickel, for various technical reasons.
+However, you can work around this limitation by expanding a given function.
+This means that instead of `f g`, you need to have `f (fun x -> g x))` where
```
yannham:
```suggestion
However, you can work around this limitation by expanding a given function:
If `f` has type `T -> U`, but a value of type `S -> V` is required, use `fun x => f x`
instead, which has the same runtime behavior but isn't typed in the same
way.
```

yannham (approved):

doc/manual/typing.md
```diff
@@ -255,6 +255,90 @@ The following type constructors are available:
...
+
+  Here, `{_ : {a : Number}}` is accepted where `{_ : {_ : Number}}` is expected,
+  because `{a : Number} <: { _ : Number }`.
+
+**Remark**: if you've used languages with subtyping before, you might expect the
+presence of a rule for function types, namely that `T -> U <: S -> V` if `S <:
+T` and `U <: V`. This is not the case in Nickel, for various technical reasons.
+However, you can work around this limitation by expanding a given function:
+If `f` has type `T -> U`, but a value of type `S -> V` is required,
+use `fun x => f x`instead, which has the same runtime behavior
```
yannham, line 339:
```suggestion
use `fun x => f x` instead, which has the same runtime behavior
```

PR #1991, opened 2024-07-08
adding array constructor to subsumption rule

Eckaos:


yannham (changes requested):
So, we don't add covariance of record types in this PR. This is fine, this makes it smaller and we can leave record types for a follow up. However, it's best to add some tests to check array covariance.

yannham:
@Eckaos the test isn't really testing what you tink, I fear. The application isn't under a type annotation, so it's not typechecked. The simplest is to just put everything under a type annotation, for example:

```nickel
let numbers : Array Number = 
  let array_of_records = [{foo = 1}, {foo = 2}, {foo = 3}] in
  let extract_numbers : Array {_ : Number} -> Array Number = fun dicts =>
    dicts
    |> std.array.map std.record.to_array
    |> std.array.flatten
    |> std.array.map (fun { field, value } => value)
  in
  extract_numbers array_of_records
in
numbers == [1, 2, 3]
```

PR #1992, opened 2024-07-08
clean up outdated primitive operator names

jeremyschlatter:
PR #1937 recently renamed many primitive operators. It missed some instances of the previous names, though. In particular:

- error messages
- comments
- documentation
- an unused function "apply_contract" in core/src/term/mod.rs

While most of these names were only made obsolete in #1937, some of them have been incorrect for longer, eg "%array_access%" in core/src/term/pattern/compile.rs and "recordMap" in core/src/term/mod.rs.

I caught as many as I could find. However it's hard to be sure I got all of them, given that some of the previous names are very general terms like "value", "fields", "length", and "map".

The full list of renames I identifiend are as follows, formatted as `<old name> <new name>`.

First, the easy cases:

    chng_pol label/flip_polarity
    record_map record/map
    str_trim string/trim
    str_chars string/chars
    str_uppercase string/uppercase
    str_lowercase string/lowercase
    str_length string/length
    str_from to_string
    num_from number/from_string
    enum_from enum/from_string
    str_is_match string/is_match
    str_find string/find
    str_find_all string/find_all
    record_empty_with_tail record/empty_with_tail
    label_push_diag label/push_diag
    enum_unwrap_variant enum/unwrap_variant
    enum_is_variant enum/is_variant
    enum_get_tag enum/get_tag
    apply_contract contract/apply
    array_lazy_app_ctr contract/array_lazy_app
    record_lazy_app_ctr contract/record_lazy_app
    elem_at array/at
    str_split string/split
    str_contains string/contains
    record_insert record/insert
    record_insert_with_opts record/insert_with_opts
    record_remove record/remove
    record_remove_with_opts record/remove_with_opts
    label_with_message label/with_message
    label_with_notes label/with_notes
    label_append_note label/append_note
    str_replace string/replace
    str_replace_regex string/replace_regex
    str_substr string/substr
    record_seal_tail record/seal_tail
    record_unseal_tail record/unseal_tail
    array_slice array/slice

Then the harder cases:

    polarity label/polarity
    go_dom label/go_dom
    go_codom label/go_codom
    go_array label/go_array
    go_dict label/go_dict
    embed enum/embed
    map array/map
    generate array/generate
    length array/length
    fields record/fields
    fields_with_opts record/fields_with_opts
    values record/values
    go_field label/go_field
    has_field record/has_field
    has_field_with_opts record/has_field_with_opts
    field_is_defined record/field_is_defined
    field_is_defined_with_opts record/field_is_defined_with_opts
    lookup_type_variable label/lookup_type_variable
    insert_type_variable label/insert_type_variable

Finally, two cases that I didn't understand and seem to be unused:

    rec_force_op op rec_force
    rec_default_op op rec_default

### DIFF
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -1235,8 +1235,8 @@ pub enum UnaryOp {
     /// Typecast an enum to a larger enum type.
     ///
     /// `EnumEmbed` is used to upcast enums. For example, if a value `x` has enum type `a | b`,
-    /// then `enum/embed c x` will have enum type `a | b | c`. It only affects typechecking as at
-    /// runtime `enum/embed someId` acts like the identity function.
+    /// then `%enum/embed% c x` will have enum type `a | b | c`. It only affects typechecking as at
+    /// runtime `%enum/embed% someId` acts like the identity function.
     EnumEmbed(LocIdent),
 
     /// A specialized primop for match when all patterns are enum tags. In that case, instead of
@@ -1257,7 +1257,7 @@ pub enum UnaryOp {
     ///
     /// The mapped function must take two arguments, the name of the field as a string, and the
     /// content of the field. `RecordMap` then replaces the content of each field by the result of
-    /// the function: i.e., `record/map f {a=2;}` evaluates to `{a=(f "a" 2);}`.
+    /// the function: i.e., `%record/map% f {a=2;}` evaluates to `{a=(f "a" 2);}`.
     RecordMap,
 
     /// Inverse the polarity of a label.
@@ -2721,7 +2721,7 @@ pub mod make {
         Term::OpN(op, args.into_iter().map(T::into).collect()).into()
     }
 
-    pub fn contract_apply<T>(
+    pub fn apply_contract<T>(
         typ: Type,
         l: Label,
         t: T,
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -1234,9 +1234,9 @@ pub enum UnaryOp {
 
     /// Typecast an enum to a larger enum type.
     ///
-    /// `Embed` is used to upcast enums. For example, if a value `x` has enum type `a | b`, then
-    /// `embed c x` will have enum type `a | b | c`. It only affects typechecking as at runtime
-    /// `embed someId` act like the identity.
+    /// `EnumEmbed` is used to upcast enums. For example, if a value `x` has enum type `a | b`,
+    /// then `enum/embed c x` will have enum type `a | b | c`. It only affects typechecking as at
+    /// runtime `enum/embed someId` acts like the identity function.
     EnumEmbed(LocIdent),
 
     /// A specialized primop for match when all patterns are enum tags. In that case, instead of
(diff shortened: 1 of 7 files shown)


core/src/term/mod.rs
```diff
@@ -1234,9 +1234,9 @@ pub enum UnaryOp {
 
     /// Typecast an enum to a larger enum type.
     ///
-    /// `Embed` is used to upcast enums. For example, if a value `x` has enum type `a | b`, then
-    /// `embed c x` will have enum type `a | b | c`. It only affects typechecking as at runtime
-    /// `embed someId` act like the identity.
+    /// `EnumEmbed` is used to upcast enums. For example, if a value `x` has enum type `a | b`,
+    /// then `enum/embed c x` will have enum type `a | b | c`. It only affects typechecking as at
```
yannham, line 1238: I think this was written a very long time ago, so let's use the current primop syntax in Nickel snippets.

```suggestion
    /// then `%enum/embed% c x` will have enum type `a | b | c`. It only affects typechecking as at
```

core/src/term/mod.rs
```diff
@@ -1257,7 +1257,7 @@ pub enum UnaryOp {
     ///
     /// The mapped function must take two arguments, the name of the field as a string, and the
     /// content of the field. `RecordMap` then replaces the content of each field by the result of
-    /// the function: i.e., `recordMap f {a=2;}` evaluates to `{a=(f "a" 2);}`.
+    /// the function: i.e., `record/map f {a=2;}` evaluates to `{a=(f "a" 2);}`.
```
yannham, line 1260:
```suggestion
    /// the function: i.e., `%record/map% f {a=2;}` evaluates to `{a=(f "a" 2);}`.
```

core/src/term/mod.rs
```diff
@@ -2721,7 +2721,7 @@ pub mod make {
         Term::OpN(op, args.into_iter().map(T::into).collect()).into()
     }
 
-    pub fn apply_contract<T>(
+    pub fn contract_apply<T>(
```
yannham, line 2724: Nitpick, but I think prefer the old name. The primop has the name `contract/apply`, which is a bit like a poor man's namespacing `contract::apply`. To avoid confusion, the corresponding variants `UnaryOp`, `BinaryOp`, and `NAryOp` have the exact same name (replacing the `/` by a separation in camel case). But I feel like we don't need to do that for such a standalone helper function.

yannham (approved):
Ah, right, we didn't update the primop names in the error messages and it kinda got forgotten. Thanks a lot for this, those aren't the most exciting changes but are very much needed to keep the codebase clean and maintainable!

yannham:
> rec_force_op op rec_force
> rec_default_op op rec_default

You can safely ignore those. They've been disabled several versions ago (or even before 1.0, I believe). The code is kept around in the vague hope of resuscitating them once we find a reasonable design, but for now, it's just dead code.

jeremyschlatter:
@yannham Thanks for the review! I addressed all of your comments.

yannham (approved):

PR #1995, opened 2024-07-15
Add boolean contract combinators to the stdlib

yannham:
Depends on #1987

This commit adds the boolean combinators `one_of`, `not` and `all_of` to the contract module of the stdlib. They are implemented in the obvious way, by making immediate decisions based on the immediate part of their contract operands. The limitations and possible work around are detailed in the documentation and in the manual.

I considered the addition of `one_of`, to adhere to the JSON schema terminology, but it turns out that it seems pretty useless for the time being, because record contracts are mostly delayed and I couldn't come up with a simple compelling example that would work in the current setting, where `one_of` would be needed over `any_of`. The standard example is two record contracts with an non-empty intersection, like `{employee_id, name, ..}` and `{student_id, name, ..}`, but this can't work as expected currently as the check for missing fields is delayed. I suspect we need to wait for proper full eager versions of builtin record contracts for a simple example (one can write the previous example with specialized custom contracts, but it starts to be involved - the fact that I can't come up with a short example to put in the documentation of the function leads me to think that it's too early to introduce this function)

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1527,17 +1527,17 @@
       | Array Dyn -> Dyn
       | doc m%"
           Builds a contract that checks if the value satisfies at least one of
-          the given contracts. More precisely, `one_of` applies each contract in
+          the given contracts. More precisely, `any_of` applies each contract in
           order and returns the result of the first one that doesn't fail
           immediately (that is, the first one which returns the value `'Ok _`).
 
           **Important** `any_of` is only an approximation of what you could
           expect from an `or` operation on contracts. Because `any_of` must
           preserve the lazyness of contracts, it means that it has to
-          pick the righ branch by relying only on the immediate part of each
+          pick the right branch by relying only on the immediate part of each
           contract, and not on the delayed part.
 
-          Fortunately, `any_of` an overstrict approximation, in that it might
+          Fortunately, `any_of` is an overstrict approximation, in that it might
           reject more values than what you expect, but it'll never let invalid
           values slip through.
 
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1175,6 +1175,7 @@
           ),
 
     Sequence
+      | Array Dyn -> Dyn
       | doc m%"
           Apply multiple contracts from left to right.
 
@@ -1381,6 +1382,7 @@
       },
 
     apply
+      : Dyn -> Dyn -> Dyn -> Dyn
       | doc m%"
           Applies a contract to a label and a value. Either aborts the execution
           with a broken contract error if the contract fails, or proceeds with
@@ -1481,6 +1483,13 @@
         %contract/apply% contract (%label/push_diag% label) value,
 
     apply_as_custom
+      : Dyn -> Dyn -> Dyn -> [|
+        'Ok Dyn,
+        'Error {
+          message | String | optional,
+          notes | Array String | optional
+        }
+      |]
       | doc m%"
           Variant of `std.contract.apply` which returns `'Error {..}` upon
           immediate failure of the contract instead of aborting the execution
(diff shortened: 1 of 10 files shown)


jneem:
I think `oneOf` is pretty useless in general, not just in nickel...

yannham:
> I think oneOf is pretty useless in general, not just in nickel...

Yeah, it's hard to think of a good, natural use-case. I imagine they might exist. Let's implement it lazily: wait for someone to come with a need for it.

Otherwise, it's ready to review now @jneem (sorry about the early ping, I requested your review too soon, before adding tests)

core/stdlib/std.ncl
```diff
@@ -1526,6 +1535,136 @@
       = fun contract label value =>
         %contract/apply_as_custom% contract (%label/push_diag% label) value,
 
+    any_of
+      | Array Dyn -> Dyn
+      | doc m%"
+          Builds a contract that checks if the value satisfies at least one of
+          the given contracts. More precisely, `one_of` applies each contract in
```
jneem, line 1542:
```suggestion
          the given contracts. More precisely, `any_of` applies each contract in
```

core/stdlib/std.ncl
```diff
@@ -1526,6 +1535,136 @@
...
+          order and returns the result of the first one that doesn't fail
+          immediately (that is, the first one which returns the value `'Ok _`).
+
+          **Important** `any_of` is only an approximation of what you could
+          expect from an `or` operation on contracts. Because `any_of` must
+          preserve the lazyness of contracts, it means that it has to
+          pick the righ branch by relying only on the immediate part of each
+          contract, and not on the delayed part.
+
+          Fortunately, `any_of` an overstrict approximation, in that it might
```
jneem, line 1552:
```suggestion
          Fortunately, `any_of` is an overstrict approximation, in that it might
```

core/stdlib/std.ncl
```diff
@@ -1526,6 +1535,136 @@
...
+      | doc m%"
+          Builds a contract that checks if the value satisfies at least one of
+          the given contracts. More precisely, `one_of` applies each contract in
+          order and returns the result of the first one that doesn't fail
+          immediately (that is, the first one which returns the value `'Ok _`).
+
+          **Important** `any_of` is only an approximation of what you could
+          expect from an `or` operation on contracts. Because `any_of` must
+          preserve the lazyness of contracts, it means that it has to
+          pick the righ branch by relying only on the immediate part of each
```
jneem, line 1549:
```suggestion
          pick the right branch by relying only on the immediate part of each
```

core/stdlib/std.ncl
```diff
@@ -1526,6 +1535,136 @@
...
+          in
+
+          { day = 1, month = 1, year = 1970 } | Date
+          ```
+        "%
+      = fun contracts =>
+        %contract/custom%
+          (
+            fun label value =>
+              std.array.fold_left
```
jneem, line 1585: Maybe it's worth having a short-circuiting version of fold_left at some point?

doc/manual/contracts.md
```diff
@@ -1058,3 +1057,148 @@ is wrapping the original value with delayed checks inside**. Doing so preserves
...
+you could expect from boolean operations. Fortunately, they are overstrict
+approximation, in that they might reject more values, but will never let invalid
+values slip through.
+
+The reason is, once again, the presence of delayed contracts.
+
+### Combining delayed contracts
+
+Let `Foo` be the contract `std.contract.any_of [{ foo | String }, {foo |
+Number}]`. `any_of` can't evaluate the field `foo` to determine if it's a number
```
jneem, line 1076: I was going to object to the word "can't" here, because in principle `any_of` could use `deep_seq` to force the evaluation. But that seems to only allow for applying the inner contract in the "blaming" way, not in the "checking" way. Maybe it would be useful to have a mechanism for forcing inner contract checks? I can imagine that sometimes you might really want the exact semantics but you don't care about laziness.

Edit: ok, I read to the end and maybe the future strict contracts make this unnecessary

doc/manual/contracts.md
```diff
@@ -1058,3 +1057,148 @@ is wrapping the original value with delayed checks inside**. Doing so preserves
...
+(Array Dyn)`, and will reject any array.
+
+On the other hand, `all_of` doesn't suffer from the same limitations, since it has
+to apply all contracts unconditionally. One exception worth noting is function
+contracts, which are mostly delayed, and might be overstrict with all boolean
+operators including `all_of`: `all_of [Number -> Number, String -> String]` will
+reject all values, including the identity function `fun x => x`, because it's
+the same contract as `std.contract.all_of [Number, String] ->
+std.contract.all_of[Number, String]`, but `std.contract.all_of [Number,
+String]` is void.
```
jneem, line 1204: Is this really overly strict? `(id | Number -> Number) "hi"` is an error, so the function contract `Number -> Number` means "a function that only takes numbers as inputs, and returns numbers". It doesn't mean "a function that, when given a number as input, returns a number." With these semantics, `id | Number -> Number` is the identity on numbers, `id | String -> String` is the identity on strings, and `id | all_of [Number -> Number, String -> String]` is an error (or at least, a function that cannot be called)

jneem (approved):

doc/manual/contracts.md
yannham: I think the "can't" here is meant as "assuming that this contract must preserve lazyness". Of course, it can otherwise extract the `foo` field and `apply_as_custom String` to get an immediate answer (as does the example of a custom contract that is more eager later, I believe?). But it might be good to emphasize it, I'll repeat the assumption of lazyness.

core/stdlib/std.ncl
yannham: Ah, yes, this is not the first time I write this :disappointed: let's bikeshed a name and an interface

doc/manual/contracts.md
yannham: Improved in 78b877c23e4fdb7334459e0591f6964d98c1cb05

doc/manual/contracts.md
yannham: I added one line of justification (among other things) of why you could expect this contract to work in 78b877c23e4fdb7334459e0591f6964d98c1cb05.

jneem: > Usually, the interpretation of a function contract T ->U is "if you give me a T, I give you an U. Otherwise, I can't promise anything"

Sure, but the interpretation in Nickel is different, right? Because `(f | T -> U) x` blames if `x` is not a `T`.

yannham: Ok, I think I see your point now. You're saying we could do nothing if the argument isn't a `T`, which would be a strange design, but is possible.

Then the strongest argument is maybe the correspondence with types, which is an important design guideline in Nickel, that contracts derived from static types are just run-time version and should morally check the same thing, even if they might accept more values (or, put differently, they check that something is a realizer of a type). Once again, in most interpretations and implementations of union types that I know of, the union is interpreted as a set-theoretic union *on terms that can be assigned the type*. Thus `T -> U \/ V -> W` is the type of the overloaded functions that supports both `T -> U` and `V -> W`, and not `T /\ V -> U /\ W`.

yannham: @jneem what do you think about the current version? Do you suggest I drop even more of the "doesn't work as expected" and just warn that this is the current semantics, or is it ok as it is?

PR #2000, opened 2024-07-18
Add short-circuiting fold_left

### LINKED ISSUE #1997: Short-circuiting version of `fold_left`

`fold_left` is a useful combinator, and the one most often used for general iteration in functional programming (preferred over `fold_right`). However, as opposed to `fold_right`, it's not short-circuiting. For example, here is an excerpt of the implementation of `std.contract.Sequence`:

```nickel
std.array.fold_left
  (
    fun acc Contract =>
      acc
      |> match {
        'Ok value =>
          std.contract.apply_as_custom Contract label value,
        # if we encountered an error at some point in the pipeline, we
        # just forward it from this point
        error => error
      }
  )
  ('Ok value)
  contracts
```

The idea is that we want apply the contracts from the given `contracts` list as long as they don't return an error, but stop at the first error. But we have to manually match on the accumulator and propagate the short-circuiting case all-along (`error => error`), which is a bit clumsy. This isn't the only place where this takes place.

Add a function in the stdlib that would have a signature like:

```nickel
try_fold_left: forall a b c. (a  -> c -> Result a b) -> a -> Array c -> Result a b
```

Where `Result a b := [| 'Ok a, 'Error b |]` (or something different, like `Some/None`, `Continue/Stop`, etc.). Whether it would be just sugar for the previous example, or it would use a special primop to actually be short-circuiting is an important question but even just a helper function to avoid the propagate-the-error boilerplate would be useful.

jneem:
Fixes #1997, subject to bikeshedding of names.

### DIFF
--- a/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_range_reversed_indices.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_range_reversed_indices.ncl.snap
@@ -4,9 +4,9 @@ expression: err
 ---
 error: contract broken by the caller of `range`
        invalid range
-    ┌─ <stdlib/std.ncl>:756:9
+    ┌─ <stdlib/std.ncl>:757:9
     │
-756 │       | std.contract.unstable.RangeFun Dyn
+757 │       | std.contract.unstable.RangeFun Dyn
     │         ---------------------------------- expected type
     │
     ┌─ [INPUTS_PATH]/errors/array_range_reversed_indices.ncl:3:19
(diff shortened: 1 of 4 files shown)


yannham (approved):
I think we can use it in a second function, which actually sparked your initial comment about adding a short-circuiting `fold_left`: `std.contract.any_of`. The `Ok`/`Error` cases are reversed though, so you might have to dualize each step and the final result with e.g.

```nickel
let flip = match {
    'Ok x => 'Error x,
    'Error x => 'Ok x,
  }
```

As far as I'm concerned, I like the name.

PR #2002, opened 2024-07-19
apply_as_custom -> check

### LINKED ISSUE #1996: Bikeshed a name for `apply_as_custom`
Remainder to find a better name for the function introduced as `apply_as_custom` in https://github.com/tweag/nickel/pull/1987. 

For more context, see in particular the section on Composition in the PR description, as well as the comment https://github.com/tweag/nickel/pull/1987#issuecomment-2229921637 and the ones following.

yannham:
Closes #1996.

After bikeshedding, it's been decided in the weekly meeting that we should use `check` for the new variant of `contract.apply` introduced in the latest contract system rework.

We couldn't find a meaningful name for this function: `try` and `try_apply` would be nice, but they are deceptive because they sound like `apply_as_custom` could catch all contract errors, while it only makes immediate errors catchable. Also, we might introduce an actual `contract.try` at some point, which is able to catch contract errors.

`then` and `and_then` are reflective of how they're used in custom contract, but that might not be the only possible usage. In the end, we just needed something different than `apply`, and `check` was the less bad candidate, with a slight connotation of trying something (as opposed to `eval`).


aspiwack:
I don't mean to relaunch a bikeshedding session, but the word “peek” is often used for such things. But “check” is reasonably intuitive too, I think.

yannham:
@jneem gentle ping, just in case you've missed this small one :slightly_smiling_face:

jneem (approved):

PR #2007, opened 2024-07-25
Add record constructor to subtyping

Eckaos:


core/src/typecheck/unif.rs
```diff
@@ -1599,7 +1599,7 @@ impl Unify for UnifRecordRows {
 }
 
 #[derive(Clone, Copy, Debug)]
-enum RemoveRowError {
+pub enum RemoveRowError {
```
yannham, line 1602: Is there a reason to make this symbol public? Public means that it's accessible to other crates. I think most of those symbols are specific to `nickel-lang-core`, or even to `typecheck`.

yannham:
> Do you know how to reconstruct a GenericUnifType from GenericUnifRecordRows.

The best is, if those record rows are coming from destructuring a unif type to begin with, is to keep the original `var_levels` around so that you can reconstruct the original type.

If this type is modified, or generated, or coming from some type inference transformation, you can just use `UnifType::concrete` (or `GenericUnifType::concrete`), the method with a lowercase `c` (and not the constructor `Concrete`), which takes care of updating and filling the levels accordingly.

core/src/typecheck/unif.rs
Eckaos: In the mod.rs file, rust does not know what is RemoveRowError. It says it is not accessible.

yannham: In which file? `typecheck/mod.rs`?

Eckaos:
The commit `58af17d` break stuff about record. 
Left to do : 
- [ ] Comments
- [ ] Fix the problem about records
- [ ] Remove subsumption function in `typecheck/mod.rs`

Eckaos:
- [ ]  Add documentation

core/src/typecheck/subtyping.rs
```diff
@@ -0,0 +1,238 @@
...
+/// The type instantiation corresponds to the zero-ary case of application in the current
+/// specification (which is based on [A Quick Look at Impredicativity][quick-look], although we
+/// currently don't support impredicative polymorphism).
+///
+/// In the future, this function might implement a other non-trivial subsumption rule.
+///
+/// [quick-look]: https://www.microsoft.com/en-us/research/uploads/prod/2020/01/quick-look-icfp20-fixed.pdf
+use super::*;
+
+pub(super) trait Subsume {
```
yannham, line 19: In Rust, it's customary to give the same name to the trait and to the provided function, when a trait provides one single well scoped method.

```suggestion
pub(super) trait IsSubsumedBy {
```

core/src/typecheck/subtyping.rs
```diff
@@ -0,0 +1,238 @@
...
+                    match row {
+                        GenericUnifRecordRowsIteratorItem::Row(a) => {
+                            a.typ
+                                .clone()
+                                .is_subsumed_by(*type_fields.clone(), state, ctxt)?
+                        }
+                        GenericUnifRecordRowsIteratorItem::TailUnifVar { id, .. } =>
+                        // We don't need to perform any variable level checks when unifying a free
+                        // unification variable with a ground type
+                        // We close the tail because there is no garanty that
```
yannham, line 64:
```suggestion
                        // We close the tail because there is no guarantee that
```

yannham (commented):
As a first review, the approach looks good. We can think about how to de-depulicate the code between subsume and unify later on. Maybe the code comments still need to be put at the right place and updated a bit, they've been mostly copy pasted or so it seems.

Eckaos:
To de-duplicate the code between `is_subsumed_by` and `unify`, we can just call `unify` where we need in `is_subsumed_by`. The cost will be the cost of pattern matching + function call (depends on what optimization the rust compiler will do). Or we could abstract cases that are the same in `is_subsumed_by` and `unify` into another function. The cost will be pattern matching + function call in the two functions (depends on what optimization the rust compiler will do). @yannham

yannham:
> To de-duplicate the code between is_subsumed_by and unify, we can just call unify where we need in is_subsumed_by. 

I'm not sure to understand what you mean here. For example, when subsuming record types, we do the same general work but we still recursively calls `is_subsumed_by` on each component, so we can't just replace the whole case by a call to `unify`. Indeed, I think we need to have a core function that is generic how it calls itself recursively. It's certainly doable, but we can do that in a second step.

P.S: it could be good to have tests for more complicated record types: types with tails, record types with subsumption acting on several components, etc.

Eckaos:
I was talking about cases where we do exactly the same work (with Constant and UnifVar etc...). In those cases, we can call `unify` from `is_subsumed_by`.
The other alternative was to factor out a function which does only cases where it is the same result for `unify` and `is_subsumed_by` and then call it from `unify` and `is_subsumed_by` when needed.

Eckaos:
I don't understand why cargo and the continuous integration pipeline give me an error about the lsp. @yannham

jneem:
It's because the lsp snapshot tests have changed -- you've changed docs in the stdlib and so the hover text in the lsp is different. Running `cargo test -p nickel-lang-lsp` and `cargo insta review` locally should let you update the snapshots.

yannham:
> It's because the lsp snapshot tests have changed -- you've changed docs in the stdlib and so the hover text in the lsp is different. Running cargo test -p nickel-lang-lsp and cargo insta review locally should let you update the snapshots.

I don't see any change to the stdlib. Also, the diff of the snapshot is dubious - we now get an additional monomorphic type with `_a` coming first in the LSP output. So it seems this PR has changed something related to either how `TypecheckVisitor` is called, or to the apparent type of values, etc.

core/src/typecheck/subtyping.rs
```diff
@@ -0,0 +1,244 @@
+//! Type subsumption
+/// Type subsumption is generally used when we change from inference mode to checking mode.
+/// Currently, there is one subtyping relations :
+/// - Record / Dictionary : `{a1 : T1,...,an : Tn} <: {_ : U}` if for every n `Tn <: U`
+///
+/// And we extend subtyping to type constructors :
+/// - Array / Array : `Array T <: Array U` if `T <: U`
+/// - Dictionary / Dictionary : `{_ : T} <: {_ : U}` if `T <: U`
+/// - Record / Record : `{a1 : T1,...,an : Tn} <: {b1 : U1,...,bn : Un}` if for every n `Tn <: Un`
```
yannham, line 9:
```suggestion
/// - Record / Record : `{a1 : T1,...,an : Tn} <: {a1 : U1,...,an : Un}` if for every n `Tn <: Un`
```

core/src/typecheck/subtyping.rs
```diff
@@ -0,0 +1,244 @@
+//! Type subsumption
+/// Type subsumption is generally used when we change from inference mode to checking mode.
```
yannham, line 2:
```suggestion
//! Type subsumption
//!
/// Type subsumption is generally used when we change from inference mode to checking mode.
```

core/src/typecheck/subtyping.rs
```diff
@@ -0,0 +1,244 @@
...
+/// Currently, there is one subtyping relations :
```
yannham, line 3: Here, and below, you should continue to use `//!` instead of `///`, because it's part of the top-level module documentation.

doc/manual/typing.md
```diff
@@ -330,6 +330,24 @@ Subtyping extends to type constructors in the following way:
 
   Here, `{_ : {a : Number}}` is accepted where `{_ : {_ : Number}}` is expected,
   because `{a : Number} <: { _ : Number }`.
+- **Record**: `{a1 : T1, ..., an : Tn} <: {b1 : U1, ..., bn : Un}` if for each
```
yannham, line 333: (the fields must be the same on both sides)

```suggestion
- **Record**: `{a1 : T1, ..., an : Tn} <: {a1 : U1, ..., an : Un}` if for each
```

lsp/nls/tests/snapshots/main__lsp__nls__tests__inputs__hover_field_typed_block_regression_1574.ncl.snap
```diff
@@ -2,7 +2,9 @@
 source: lsp/nls/tests/main.rs
 expression: output
 ---
-<0:24-0:37>[Applies a function to every element in the given array. That is,
+<0:24-0:37>[```nickel
+(_a -> Number) -> Array _a -> Array Number
```
yannham, line 6: This change  in the LSP is dubious. The user will see an additional monomorphic type instead of the actual polymorphic annotation, which we don't want, I guess. We should try to understand where it comes from. Usually, the typechecker communicates with the LSP through the `TypecheckVisitor` trait, by calling the corresponding `visit_ident` method to update the type affected to a variable or a binding in general. Maybe we call this too late now, once the type has been instantiated? Maybe we call it many times? I don't know, but we should fix that.

yannham (changes requested):
Overall looks good, but the LSP regression is suspicious.

yannham:
I'm taking over this PR for the foreseeable future, unless @Eckaos thinks he has the time to bring it down the finish line.

yannham:
After some thorough investigation, I think I understand what's happening and I convinced that it makes sense (although in the test case it adds a bit of noise, it's also more correct in some cases - see the commit message). It an artifact of when we visit the term compared to when the types are instantiated, more or less.

yannham (approved):

PR #2015, opened 2024-07-29
Add `--format` argument to `nickel query` command

### LINKED ISSUE #1976: Output query results as JSON
I need to programmatically list top-level fields of a nickel program.

```console
$ nickel query cbuild.ncl
warning: empty query path
 = You queried a value without requesting a specific field path. This operation can't find any metadata, beside listing the fields of a record.
 = Try to query the root configuration and provide a query path instead.
 = For example, instead of querying the expression `(import "config.ncl").module.input` with an empty path, query `config.ncl` with the `module.input` path: 
   `nickel query config.ncl --field module.input


Available fields
• hello
No metadata found for this field.
```

It is too hard to parse in my script.

```console
$ nickel query --json cbuild.ncl
{"fields":["hello"]}
```

**Additional context**
I'm calling nickel as a subprocess.

suimong:
Closes #1976 

This PR adds the ability for `nickel query` to export the result to JSON, YAML & TOML,  similar to that of `nickel export`. The output schema looks like:

```nickel
{
  doc | String | optional,
  "optional" | Bool,
  not_exported | Bool,
  priority | String,   # variants include: "default", "neutral", "force", and string representation of rational number ("1", "5/2", etc.)
  typ | String | optional,
  contracts | Array String | optional,
  sub_fields | Array String | optional,
  value | String | optional,
}
```

Two flags are added to nickel query: `--format` and `--output`
-  `--format` accepts 4 variants: markdown(default), json, yaml and toml. `nickel query --format markdown ...` behaves the same as the original `nickel query ...`
- `--output` is added to align with `nickel export`. There's one exception though: user is not able to output markdown formatted query result to a file. While this could be implemented, it is arguable that a markdown output is much more useful when used ad-hoc in a shell.

### DIFF
--- a/cli/src/query.rs
+++ b/cli/src/query.rs
@@ -1,4 +1,13 @@
-use nickel_lang_core::repl::query_print;
+use std::{fs, io::Write, path::PathBuf};
+
+use nickel_lang_core::{
+    error::{Error, IOError},
+    identifier::{Ident, LocIdent},
+    repl::query_print,
+    serialize::{self, MetadataExportFormat},
+    term::{record::Field, LabeledType, MergePriority, RichTerm, Term},
+};
+use serde::Serialize;
 
 use crate::{
     cli::GlobalOptions,
@@ -24,8 +33,80 @@ pub struct QueryCommand {
     #[arg(long)]
     pub value: bool,
 
+    /// Export the value and all metadata of selected field in the specified format.
+    ///
+    /// This flag cannot be used along with the following flags: --doc, --contract, --type, --default, --value
+    #[arg(long, short, value_enum, default_value_t, conflicts_with_all(["doc", "contract", "typ", "default", "value"]))]
+    pub format: MetadataExportFormat,
+
+    /// Output file. Standard output by default
+    #[arg(short, long, conflicts_with_all(["doc", "contract", "typ", "default", "value"]))]
+    pub output: Option<PathBuf>,
+
     #[command(flatten)]
     pub inputs: InputOptions<ExtractFieldOnly>,
+}
+
+#[derive(Clone, Debug, Serialize)]
+struct QueryResult {
+    pub doc: Option<String>,
+    pub typ: Option<LabeledType>,
+    pub contracts: Vec<LabeledType>,
+    pub optional: bool,
+    pub not_exported: bool,
+    pub priority: MergePriority,
+    pub sub_fields: Option<Vec<Ident>>,
+    pub value: Option<RichTerm>,
+}
+
+impl From<Field> for QueryResult {
+    fn from(field: Field) -> Self {
+        let sub_fields = match field.value {
+            Some(ref val) => match val.as_ref() {
+                Term::Record(record) if !record.fields.is_empty() => {
+                    let mut fields: Vec<_> = record.fields.keys().collect();
+                    fields.sort();
+                    Some(fields.into_iter().map(LocIdent::ident).collect())
+                }
+                Term::RecRecord(record, dyn_fields, ..) if !record.fields.is_empty() => {
+                    let mut fields: Vec<_> = record.fields.keys().map(LocIdent::ident).collect();
+                    fields.sort();
+                    let dynamic = Ident::from("<dynamic>");
+                    fields.extend(dyn_fields.iter().map(|_| dynamic));
+                    Some(fields)
+                }
+                // Empty record has empty sub_fields
+                Term::Record(..) | Term::RecRecord(..) => Some(Vec::new()),
+                // Non-record has no concept of sub-field
+                _ => None,
+            },
+            None => None,
+        };
+
+        QueryResult {
+            doc: field.metadata.doc,
+            typ: field.metadata.annotation.typ,
+            contracts: field.metadata.annotation.contracts,
+            optional: field.metadata.opt,
+            not_exported: field.metadata.not_exported,
+            priority: field.metadata.priority,
+            sub_fields,
+            value: field.value,
+        }
+    }
+}
+
+impl QueryResult {
+    pub fn filter_for_export(self) -> serde_json::Map<String, serde_json::Value> {
+        let mut query_res = serde_json::to_value(&self).unwrap();
+        let query_res = query_res.as_object_mut().unwrap();
+
+        if self.value.is_none() {
+            query_res.remove("value");
+        }
+
+        query_res.to_owned()
+    }
 }
 
 impl QueryCommand {
@@ -48,6 +129,34 @@ impl QueryCommand {
         }
     }
 
+    fn export<T>(self, res: T, format: MetadataExportFormat) -> Result<(), Error>
+    where
+        T: Serialize,
+    {
+        // This is a near-verbatim copy of ExportCommand::export
+
+        // We only add a trailing newline for JSON exports. Both YAML and TOML
+        // exporters already append a trailing newline by default.
+        let trailing_newline = format == MetadataExportFormat::Json;
+
+        if let Some(file) = self.output {
+            let mut file = fs::File::create(file).map_err(IOError::from)?;
+            serialize::to_writer_metadata(&mut file, format, &res)?;
+
+            if trailing_newline {
+                writeln!(file).map_err(IOError::from)?;
+            }
+        } else {
+            serialize::to_writer_metadata(std::io::stdout(), format, &res)?;
+
+            if trailing_newline {
+                println!();
+            }
+        }
+
+        Ok(())
+    }
+
     pub fn run(self, global: GlobalOptions) -> CliResult<()> {
         let mut program = self.inputs.prepare(&global)?;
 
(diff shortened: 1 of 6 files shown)


suimong (commented):
@yanham do you think it's ok that in this PR, the `--format` flag ignores all customzation flags (`--doc`, `--default` etc.) and just exports all metadata fields and value? I tried to make them compatible but after some experimentation I'm inclined to keep the two features independent. The main reason is that by default `--format` exports all metadata fields which I think is desirable,  but a few of them does not have a corresponding flag (namely `optional`, `not_exported`, `priority`), so customization for json export would be half baked and inconsistent with the current markdown output. That is, if we leave the markdown rendering code untouched.

yannham:
I think it's reasonable. Those options are useful to minimize the markdown output when one is only interested in one particular field, but It's probably less useful for a JSON export, which can be transformed afterwards by another tool (like...Nickel :imp: ?) or the end consumer if needed. I don't remember `clap` well enough to know if we can encode that constraint directly. I suppose we can (such that the CLI prints an error automatically if you specify both `--doc` and `--format`). In any case, as long as it's documented properly in the argument description and validated in one way or another, it's fine by me :+1:

suimong:
Cool! Will get the PR ready for review soon.

cli/src/query.rs
```diff
@@ -48,27 +82,61 @@ impl QueryCommand {
...
-            })
-            .report_with_program(program)?;
-
-        if !found {
-            eprintln!("No metadata found for this field.")
+        if let Some(format) = self.format {
+            let query_res = program.query().map(|field| QueryResult(field));
+            if let Ok(res) = query_res {
+                self.export(res, format).report_with_program(program)?
+            } else {
```
yannham, line 121: Nitpick: instead of duplicationg `"No metadata found for this field"`, could we define `found` outside of the `if let Some(format) = self.format {...} else {...}` and just set found to `true` in the`if` (or `false` in the else, depending on the initial value of `found`), and move the `if !found` below also after the whole if-then-else?

core/src/serialize.rs
```diff
@@ -47,6 +46,25 @@ impl fmt::Display for ExportFormat {
     }
 }
 
+/// Available common export formats.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, clap::ValueEnum)]
+pub enum ExportFormatCommon {
+    #[default]
+    Json,
```
yannham, line 53: I wonder if we shouldn't rather have an additional `Markdown` variant, which is the default (and should be the documented default in the CLI as well). I think it's good practice that the normal behavior (without `--format`) can be recovered in a more precise way with an argument - typically, `nickel export` is the same as `nickel export --format json`, etc. - so that the default behavior is just that, a default value.

On the other hand, it makes the logic of when we should allow `--doc`, etc. more complex. So I'm not entirely sure. Maybe we can start as it is now, but then I would remove the `Default` impl of `ExportFormatCommon`, because it doesn't really have a default value.

core/src/term/mod.rs
```diff
@@ -681,6 +681,20 @@ impl fmt::Display for MergePriority {
...
+        S: Serializer,
+    {
+        match self {
+            Self::Bottom => serializer.serialize_unit_variant("priority", 0, "default"),
+            Self::Neutral => serializer.serialize_unit_variant("priority", 3, "neutral"),
+            Self::Numeral(n) => serializer.serialize_newtype_variant("priority", 2, "numeral", &n),
+            Self::Top => serializer.serialize_unit_variant("priority", 3, "force"),
+        }
+    }
+}
```
yannham, line 696: This looks pretty mechanic - can't `serde` derive it automatically with `#[derive(Serialiaze)]` on `MergePriority` ?

core/src/serialize.rs
```diff
@@ -47,6 +46,25 @@ impl fmt::Display for ExportFormat {
     }
 }
 
+/// Available common export formats.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, clap::ValueEnum)]
+pub enum ExportFormatCommon {
```
yannham, line 51: `ExportFormatCommon` isn't very telling, IMHO (how it is different than `ExportFormat` or other formats described elsewhere in the codebase?). If there's no good general way to describe it, maybe something more descriptive such as `MetadataExportFormat` is less confusing?

core/src/repl/query_print.rs
```diff
@@ -1,11 +1,10 @@
 //! Rendering of the results of a metadata query.
-use crate::{
-    identifier::{Ident, LocIdent},
-    pretty::PrettyPrintCap,
```
yannham, line 4: This is what is failing the CI - the trait `pretty::PrettyPrintCap` has been added recently (maybe before you  branched?) and need to be kept included here.

suimong:
@yannham Thanks for the review! I'm very sorry that I forgot to push one last commit before marking this ready for review. 

Additionally, here are a few questions on which I'd like to have your opinion:
    - whether to use `Nullable T` or `T | optional` for some fields in the output schema
    - how to serialize the priority field. Currently everything is serialized to a string, including the `Numeral` variant. An alternative would be to separate the `Numeral` variant from `Bottom` and  `Top`, serializing the `Numeral` variant to json number. 
    - the logic that dictates the presence of the value field.

yannham:
Hi @suimong, sorry for the long pause - summer break it was!

To answer your previous questions:

> whether to use Nullable T or T | optional for some fields in the output schema

I would  tend to use `T | optional`, if it's not too annoying. I feel like this is the usual way in the JSON world/API world rather than putting explicit nulls. The only reason to use `null` for me is if the data might be used in a statically typed Nickel block, but here it's just for export on the CLI.

> how to serialize the priority field. Currently everything is serialized to a string, including the Numeral variant. An alternative would be to separate the Numeral variant from Bottom and Top, serializing the Numeral variant to json number.

This is a good question. I was going to say that a string is a good approximation, and is easy to parse back as an integer if needed, but on the other hand I don't know if lose anything by being more precise and serialize a number as a number. Maybe the schema is less nice to consume for statically typed language (although you can always serialize that as an enum in Rust I guess). I don't have a strong opinion here.

> the logic that dictates the presence of the value field.

Hmm. I'm tempted to revisit the original logic and to just always print the value (capped so that we don't print 20k lines of config when querying a top-level attribute). But this can be done in a second step; for now, if it's not too annoying, I would keep the same logic as the original query printer, if possible. We can then revisit both at once later.

yannham (commented):
After reading the code, I think the value logic is fine as it is, and we'll probably adapt the markdown renderer to do the same. Otherwise it looks good I believe, besides the potential updates regarding your questions and my answers.

suimong:
Hi @yannham hope you had a great summer break and thanks for the review!

I rebased the PR onto the latest master, and it should be ready to merge. To summarize the changes:

- two flags are added to `nickel query`: `--format` and `--output`
  - `--format` accepts 4 variants: `markdown`(default), `json`, `yaml` and `toml`. `nickel query --format markdown ...` behaves the same as the original `nickel query ...`
  - `--output` is added to align with `nickel export`. There's one exception: user is not able to write `markdown` formatted query result to a file. While this is simply due to a lack of implementation, I feel like it is acceptable for now as the rendered markdown output should be primarily used ad-hoc within the shell.

> I would tend to use T | optional, if it's not too annoying. I feel like this is the usual way in the JSON world/API world rather than putting explicit nulls. The only reason to use null for me is if the data might be used in a statically typed Nickel block, but here it's just for export on the CLI.

Thanks for the answer. Though coming back to this after a month, I honestly don't recall where the question came from😅 It has nothing to do with the rust code anyway.

> This is a good question. I was going to say that a string is a good approximation, and is easy to parse back as an integer if needed, but on the other hand I don't know if lose anything by being more precise and serialize a number as a number. Maybe the schema is less nice to consume for statically typed language (although you can always serialize that as an enum in Rust I guess). I don't have a strong opinion here.

I'm going to keep the original string representation which as you said is a fairly good approximation. There's something interesting here (I should've mentioned when I asked the question), as I found out that the actual type of the priority "number" is actually `Rational`, which means if a user specifies a priority of 2.336, it will be serialized to "292/125"😄 While this will arguably make parsing it into number somewhat harder, but in all practicality, since users will need to write custom logic to parse the "priority" field anyway, this should not cause them too much extra work.

On a side note, I realized that making the priority a rational number is a pretty neat feature, as you'll never run into a situation where you cannot "insert" a layer of config in between two layers. With integer, while in theory user could space the priority number of two adjacent layers far apart, like priority 10, priority 20 etc., but in practice a pretty common case is that one layer does not define priority (priority 0), and some user defines an override with priority 1 (because that's just intuition), and now you can not sneak in anything in between. But with rational numbers, we can just set priority to the average of the two number.

suimong:
I'm not sure why there's a merge conflict...I thought I solved all of them locally.

jneem:
It seems to be rebased on an older master. For example, I see [165eedd](https://github.com/tweag/nickel/commit/165eeddcca6a723b6e392220a18e5f5e41e7ed24) on master, but not in this branch

core/src/serialize.rs
```diff
@@ -370,6 +388,35 @@ pub fn validate(format: ExportFormat, t: &RichTerm) -> Result<(), ExportError> {
...
+    format: MetadataExportFormat,
+    item: &T,
+) -> Result<(), ExportError>
+where
+    W: io::Write,
+    T: ?Sized + Serialize,
+{
+    // This is a near-verbatim copy of `to_writer`
+    match format {
+        //
```
yannham, line 402: This comment is missing some text, I suppose :)

cli/src/query.rs
```diff
@@ -24,10 +33,94 @@ pub struct QueryCommand {
...
+        }
+    }
+}
+
+impl QueryResult {
+    pub fn filter_for_export(self) -> serde_json::Map<String, serde_json::Value> {
+        let mut query_res = serde_json::to_value(&self).unwrap();
+        let query_res = query_res.as_object_mut().unwrap();
+
+        if self.value.is_none() {
```
yannham, line 104: I looked it up and it seems that you can achieve the same a bit more declaratively by slapping the following serde attribute on the corresponding field of the enum:

```rust
struct QueryResult {
  #[serde(skip_serializing_if = "Option::is_none")]
  value: Option<RichTerm>,
  // etc..
}
```

And get rid of the `if`s here.

yannham (approved):
Beside small details, it looks good! Thanks again for tackling this. I like the final design with an `output` flag and a `format` flag.

yannham:
It's strange that a string literal is closurized here, but it's a technical detail that I think is indeed unrelated to this PR indeed - we might be missing some closure substitution somewhere. However it makes something else apparent: given how serde works, value is recursively serialized, which is an issue (typically in your case, it doesn't know how to serialize a closure - and there are many expressions that aren't serializable to JSON, such as `foo + 1`, `if true then 1 else 2` or `fun x => x + 1`, and they can all be the value of a field). So I think what we want is for value an `Option<String>` and pretty-print the value, instead of deeply serializing it (it will print the ugly closure as well, but it's a question that we can handle separately).

suimong:
Ah that's what you meant by "print" & "capped" in an earlier comment. I added a note in the CLI's help message regarding the value field.

PR #2017, opened 2024-07-30
Fix and improve contract generation optimization for static types

### LINKED ISSUE #2014: Record row polymorphism can be destructive
**Describe the bug**

A function with type ` forall tail. { a : Number ; tail } -> { a : Number ; tail }` can remove the record tail from the returned value.

**To Reproduce**

Evaluate the following expression:

```nickel
let f : forall tail. { a : Number ; tail } -> { a : Number ; tail }
  = fun o => o
  in f { a = 1, b = ".." }
```

The result is `{ a = 1, }`.

**Expected behavior**

I expect a result of `{ a = 1, b = ".." }`.

**Environment**
 - OS name + version: Windows 11
 - Version of the code:  master (a7dc3594)

yannham:
Fixes #2014.

## Motivation

This PR fixes a number of issues with the optimization of contract generation from a static type annotation. This optimization has been introduced to take advantage of the guarantees of static typing to simplify the generated contract and elide a number of useless checks at run-time.

However, the original optimization is too agressive: it turns any non-function type in a positive position to `Dyn`. 

Contracts in negative position must be kept around, because they check what's come from the (potentially untyped) external world inside the function: the fact that a function is statically typed doesn't ensure it'll be called with the
right arguments. Basic contracts in positive position can be elided though: if a function has been statically checked to be of type `Number ->  Number`, we now that it'll always return a `Number` at runtime, and we can simplify the contract to `Number -> Dyn` (which is further specialized to an implementation that doesn't check anything on the codomain). Alas, this isn't true for more complex types in positive positions.

### Unsoundness

The first issue is that a type constructor in positive position can hide negative contracts inside. Typically, take `let functions: Array (Number -> Number) in (std.array.at 0 functions) "a"`. The application is untyped, and wrongly passes a string to a function of the array. However, the current optimization will turn this contract to just `Dyn` at runtime and get rid of entirely, because there's an `Array _` in positive position, and won't blame at runtime (might get a primop dynamic type error, or nothing, depending what exactly does the function). The elision is only valid if `_` doesn't contain any arrow type, which it does, in this example.

Another issue stems from the sealing/unsealing symmetry of polymorphic row contracts. Take `let id : forall r. {foo : Number; r} -> {foo : Number; r} = fun x => x in id {foo = 1, bar = 2}`. The current version would return `{foo = 1}`, with the `bar` field missing. The issue is that the second `{foo : Number; r}` is a record type in positive position, which currently gets entirely elided, giving `forall r. {foo : Number; r} -> Dyn` as an optimized contract. The argument contract seals anything else than `foo` in the record's tail, but there's no dual contract to unseal it anymore, such that `bar` stays sealed in the tail forever and thus invisible.

This commit implements a more complicated strategy, but which is hopefully correct. Instead of getting rid of type constructor in positive positions, we try to simplify their content as well, and only if the content is simplified down to `Dyn` (or `{; Dyn}` for record rows, etc.), we get rid of the entire constructor. Otherwise, we try to pay only for the necessary checks. For example, `{foo : Number, bar : Number -> Number}` now gets simplified to `{bar : Number -> Dyn; Dyn}` instead of just `Dyn` before (which was incorrect).

The full simplification rules around polymorphic row contracts are a bit intricate and are detailed in the documentation of the corresponding methods.

### DIFF
--- a/core/src/typ.rs
+++ b/core/src/typ.rs
@@ -950,6 +950,73 @@ impl EnumRows {
             ty: std::marker::PhantomData,
         }
     }
+
+    /// Simplify enum rows for contract generation when the rows are part of a static type
+    /// annotation. See [Type::contract_static].
+    ///
+    /// The following simplification are applied:
+    ///
+    /// - the type of the argument of each enum variant is simplified as well
+    /// - if the polarity is positive and the rows are composed entirely of enum tags and enum variants whose argument's simplified type is `Dyn`, the entire rows are elided by returning `None`
+    /// - a tail variable in tail position is currently left unchanged, because it doesn't give
+    /// rise to any sealing at runtime currently (see documentation of `$forall_enum_tail` in the
+    /// internals module of the stdlib)
+    fn simplify(
+        self,
+        contract_env: &mut Environment<Ident, RichTerm>,
+        simplify_vars: SimplifyVars,
+        polarity: Polarity,
+    ) -> Option<Self> {
+        // Actual simplification logic. We can elide the initial enum row type if **each**
+        // individual row can be elided and this enum has an empty tail. Thus, when making
+        // recursive calls, we can't just return `None` when some subrows can be elided, as we
+        // might still need them. Instead, `do_simplify` returns a tuple of the simplified rows and
+        // a boolean indicating if this part can be elided.
+        fn do_simplify(
+            rows: EnumRows,
+            contract_env: &mut Environment<Ident, RichTerm>,
+            simplify_vars: SimplifyVars,
+            polarity: Polarity,
+        ) -> (EnumRows, bool) {
+            match rows {
+                EnumRows(EnumRowsF::Empty) => (EnumRows(EnumRowsF::Empty), true),
+                // Currently, we could actually return `true` here, because the tail contract for
+                // enums doesn't do anything (see documentation of `$forall_enum_tail` in the
+                // `internals` module of the stdlib). However, if we ever fix this shortcoming of
+                // enum tail contracts and actually enforce parametricity, returning `true` would
+                // become incorrect. So let's be future proof.
+                EnumRows(EnumRowsF::TailVar(id)) => (EnumRows(EnumRowsF::TailVar(id)), false),
+                EnumRows(EnumRowsF::Extend { row, tail }) => {
+                    let (tail, mut elide) =
+                        do_simplify(*tail, contract_env, simplify_vars.clone(), polarity);
+
+                    let typ = row.typ.map(|mut typ| {
+                        *typ = typ.simplify(contract_env, simplify_vars, polarity);
+                        elide = elide && matches!(typ.typ, TypeF::Dyn);
+                        typ
+                    });
+
+                    let row = EnumRowF { id: row.id, typ };
+
+                    (
+                        EnumRows(EnumRowsF::Extend {
+                            row,
+                            tail: Box::new(tail),
+                        }),
+                        elide,
+                    )
+                }
+            }
+        }
+
+        let (result, elide) = do_simplify(self, contract_env, simplify_vars, polarity);
+
+        if matches!((elide, polarity), (true, Polarity::Positive)) {
+            None
+        } else {
+            Some(result)
+        }
+    }
 }
 
 impl Subcontract for EnumRows {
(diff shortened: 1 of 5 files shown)


yannham:
@jneem if you have a chance to take a look - I'd like to include this in the 1.8 release (given the severity of the bug that is fixed)

core/src/typ.rs
```diff
@@ -1138,6 +1205,233 @@ impl RecordRows {
...
+    ///   Dyn; Dyn}`. `{foo : Number, bar : Dyn}` is simplified to `{; Dyn}` which is simplified
+    ///   further to `Dyn` by [Type::simplify].
+    ///
+    ///   We can't do that when the tail can't be simplified to `Dyn` (this is the case when the
+    ///   tail variable has been introduced by a forall in negative position).
+    /// - The case of a tail variable is a bit more complex and is detailed below.
+    ///
+    /// # Simplification of tail variable
+    ///
+    /// The following paragraphs distinguish between a tail variable (and thus and enclosing record
```
jneem, line 1227:
```suggestion
    /// The following paragraphs distinguish between a tail variable (and thus an enclosing record
```

core/src/typ.rs
```diff
@@ -1138,6 +1205,233 @@ impl RecordRows {
...
+    /// equal to the excluded fields for this row variable (it's always a subset of the excluded
+    /// fields by definition of row constraints), or if there are some excluded fields left to
+    /// check.
+    ///
+    /// ## Positive polarity
+    ///
+    /// In a positive position, we can replace the tail with `Dyn` if the tail isn't a variable
+    /// introduced in negative position, because a typed term can never violate row constraints.
+    ///
+    /// For a variable introduyced by a forall in negative position, we need to keep the unsealing
```
jneem, line 1277:
```suggestion
    /// For a variable introduced by a forall in negative position, we need to keep the unsealing
```

core/src/typ.rs
```diff
@@ -1483,10 +1816,166 @@ impl Traverse<RichTerm> for Type {
...
+// to have to reconstruct a type from `None` everywhere, which is why we didn't settle for
+// just returning `Option<Self>` in [Simplify::simplify]. We can also return `TailDyn` for
+// record rows.
+//
+// On the other hand, returning `Option<Self>` is useful when implementing [Simplify] on
+// `EnumRows` which don't have a neutral element. In this case, `Result` is `Option<Self>` and
+// we can just return `None` to indicate to the caller that the containing type constructor -
+// record or enum - can be entirely removed and replaced with `Dyn`.
+
+// Simplify a type appearing in a static type annotation for contract generation.
```
jneem, line 1868: These comment blocks look like they were left over from moving things around?

jneem (approved):

core/src/typ.rs
```diff
@@ -1138,6 +1205,233 @@ impl RecordRows {
...
+    /// ensured to be `Dyn`.
+    fn simplify(
+        self,
+        contract_env: &mut Environment<Ident, RichTerm>,
+        simplify_vars: SimplifyVars,
+        polarity: Polarity,
+    ) -> Self {
+        // This helper does a first traversal of record rows in order to peek the tail and check if
+        // it's a variable introduced by a forall in negative position. In this case, the second
+        // component of the result will be `true` and we can't elide any field during the actual
```
yannham, line 1291:
```suggestion
        // component of the result will be `false` and we can't elide any field during the actual
```

core/src/typ.rs
yannham: Indeed, thanks.

core/src/typ.rs
```diff
@@ -1483,10 +1816,166 @@ impl Traverse<RichTerm> for Type {
...
+    fn parse_type(s: &str) -> Type {
+        use codespan::Files;
+        let id = Files::new().add("<test>", s);
+
+        FixedTypeParser::new()
+            .parse_strict(id, Lexer::new(s))
+            .unwrap()
+    }
+
+    /// Parse a type, simplify it and assert that the result correspond to the second argument
```
yannham, line 1893:
```suggestion
    /// Parse a type, simplify it and assert that the result corresponds to the second argument
```

core/stdlib/internals.ncl
```diff
@@ -375,6 +375,30 @@
             extra_fields
         ),
 
+  # Specialized version of `$forall_record_tail` for a `forall` introduced in a
+  # positive position and a tail in a negative position which only check for
```
yannham, line 379:
```suggestion
  # positive position and a tail in a negative position which only checks for
```

PR #2021, opened 2024-08-05
[Performance] Use primop for the most basic contracts

yannham:
This is an experimentation that uses primops instead of the most basic contracts (number, booleans, strings and dyn) to see if that makes a difference, performance-wise, rather than have them defined in Nickel. Although they aren't very costly, they should be called _a lot_, which thus could make a difference.

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -2981,6 +2981,39 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     pos_op_inh,
                 )))
             }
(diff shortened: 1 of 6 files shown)


yannham:
It seems to show a consistent 2.5% perf improvement on the private bench, both small and big (~4sec eval and ~20sec eval). While it's not nothing, I don't know if it's worth the ad-hoc implementation right  now, given that there should be much more meaningful gain by either reducing the memory footprint of `Term` or by having a first step toward a bytecode VM (let's say, a flattened AST).

yannham:
Given the limited improvement, I'm closing this experiment.

PR #2022, opened 2024-08-05
[Performance] Reduce size of `term`

yannham:
A `sizeof` show that the main `Term` enum has a whopping size of 464 bytes. This is quite fat and might pressure the allocator, dropping code and various caches. This PR boxes a bunch of large structure to reduce the size of the overall enum.


yannham:
Also the code is a bit noisier (need to box and unbox quite a bit more), the preliminary results are encouraging. After a first trimming round from 464 to 74 bytes, I see a consistent gain of around 15% on the smaller real life benchmark, and up to 25-30% on the larger one (in running time). The allocated memory also goes down significantly (because it's more expansive to measure, I only ran bytehound on the smaller example, but I get more than 60% reduction in allocated memory).

I'm going to try to reduce it further down to approximately 48 bytes (which are 24x2, where 24 is the term + position information repr) and measure the difference. Then I'll check and see if there are other pervasive enum to optimize.

yannham:
I'm not sure of the optimum, but reducing it further to 56 bytes wasn't beneficial,and is actually less performant than 72. It also costed a lot more in term of code uglyness, so I'm tempted to keep it at 72 bytes for now, and look for other enums to optimize.

yannham:
This was an interesting experience, but this effort is superseded by #2045.

PR #2024, opened 2024-08-06
Fix benchmarks on Windows

cydparser:


yannham (approved):
Thanks a lot for fixing those! We don't have  much Windows user here, and it's not easy to test through GitHub actions, so help on this front is always welcome. I've tested locally and this does still work correctly on Linux, for all I can see.

cydparser:
Replaced by #2025

PR #2025, opened 2024-08-08
Fix windows tests

### LINKED ISSUE #1565: nls tests fail on Windows
The nls testing harness appears to not work at all on windows. See [here](https://github.com/tweag/nickel/pull/1561#issuecomment-1702934852)

cydparser:
- Includes a commit from #2024
- Fixes #1565
- Added a Windows CI job
- Improved unwrap panic messages

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -86,7 +86,6 @@ strsim = "0.10.0"
 pretty_assertions.workspace = true
 assert_matches.workspace = true
 criterion.workspace = true
-pprof = { workspace = true, features = ["criterion", "flamegraph"] }
 nickel-lang-utils.workspace = true
 similar.workspace = true
 test-generator.workspace = true
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -86,7 +86,7 @@ strsim = "0.10.0"
 pretty_assertions.workspace = true
 assert_matches.workspace = true
 criterion.workspace = true
-nickel-lang-utils.workspace = true
+nickel-lang-utils = { workspace = true, features = ["pprof"] }
 similar.workspace = true
 test-generator.workspace = true
 
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -19,3 +19,6 @@ criterion.workspace = true
 codespan.workspace = true
 serde = { workspace = true, features = ["derive"] }
 toml = { workspace = true, features = ["parse"] }
+
+[target.'cfg(target_family = "unix")'.dependencies]
+pprof = { workspace = true, features = ["criterion", "flamegraph"] }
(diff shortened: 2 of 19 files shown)


core/Cargo.toml
```diff
@@ -86,11 +86,13 @@ strsim = "0.10.0"
 pretty_assertions.workspace = true
 assert_matches.workspace = true
 criterion.workspace = true
-pprof = { workspace = true, features = ["criterion", "flamegraph"] }
 nickel-lang-utils.workspace = true
 similar.workspace = true
 test-generator.workspace = true
 
+[lints.clippy]
+doc_lazy_continuation = { level = "allow" }
```
jneem, line 94: Why turn this off? I don't see it firing, but even if it does it seems like a useful lint

utils/Cargo.toml
```diff
@@ -19,3 +19,6 @@ criterion.workspace = true
 codespan.workspace = true
 serde = { workspace = true, features = ["derive"] }
 toml = { workspace = true, features = ["parse"] }
+
+[target.'cfg(target_family = "unix")'.dependencies]
+pprof = { workspace = true, features = ["criterion", "flamegraph"] }
```
jneem, line 24: Can this be a dev-dependency?

jneem (changes requested):
Thanks for looking into this! It looks good, just a couple of minor questions.

jneem:
(Also, I'm happy to merge this in one go and forget about #2024, if that's more convenient)

utils/Cargo.toml
cydparser: It must be a dependency because this is a library. I initially used dev-dependency on the other PR and it caused CI failures.

core/Cargo.toml
cydparser: I agree that it is a useful hint. There are 49 places where the warning occurs (with clippy 0.1.80), so I assumed that the documentation violating the lint is a deliberate style choice. I can drop that commit.

utils/Cargo.toml
jneem: Ah, right. But it should be possible to make it optional (i.e. depend on a non-default feature flag) and then only bring in that feature in the dev-dependencies of nickel-lang-core?

yannham:
Oops, I almost merged #2024 before seeing this one. I've reviewed and tested the other. Can I let both this one and #2024 in your hands, @jneem ?

utils/Cargo.toml
cydparser: I've made that change here: https://github.com/tweag/nickel/pull/2025/commits/0fbcf71ba24526998164c4629a40ead59870403d

jneem (approved):
Thank you!

PR #2032, opened 2024-09-06
Update to latest topiary-queries (formatting)

### LINKED ISSUE #2003: Bikesheding: formatting of multi-line expressions enclosed in parentheses
After the addition of a new constructor for custom contracts (and in fact a primop for function-based contracts in general), the stdlib has many instances of the following pattern:

```
some_function (fun label value =>
   ...multiple lines of code
)
```

Or sometimes with multiple arguments (`some_function` is most of the time `std.contract.custom` or the primop `%contract/custom%`). Nickel currently formats such multi-line applications in the following way:

- put `some_function` on its own line
- indent arguments (here, the function) and put it on the next line
- because `( ... )` is multi-line as well, put `(` on the first line, and add another indentation level
- put the content of the function
- put a closing `)`

This leads the `std.contract.custom` pattern to add a lot of indentation. But it's not the only offending example: any higher-order function like `std.array.map` or various folds also exhibit a big shift to the right of the screen, and such calls might be nested. Here is an example from the stdlib:

```nickel
= fun contracts =>
  %contract/custom%
    (
      fun label value =>
        std.array.try_fold_left
          (
            fun _acc Contract =>
              let label =
                %label/with_message%
                  "any_of: a delayed check of the picked branch failed"
                  label
              in
              std.contract.apply_as_custom Contract label value
              # We want to short-circuit on contract success. Since try_fold_left
              # short-circuits on failure, we need to flip the two.
              |> match {
                'Ok value => 'Error value,
                'Error msg => 'Ok msg
              }
          )
          ('Ok null)
          contracts
        |> match {
          'Ok _ =>
            'Error {
              message = "any_of: value didn't match any of the contracts",
            },
          'Error value => 'Ok value,
        }
    ),
```

I think this formatting is sub-optimal: the content of the folded function is 6 indentation levels to the right of the original field definition. Another instance of high-indentation level, once again due to the way parentheses are formatted, is nested applications. Here is another example from the stdlib (here combined with anonymous function definitions)

```nickel
DependentFun
  SliceIndexFirst
  (
    fun start_index =>
      DependentFun
        (SliceIndexSecond start_index)
        (
          fun end_index =>
            ArraySliceArray end_index -> Dyn
        )
  ),
```

## Proposal(s)

I can see several ways to reduce the indentation level:

### Special casing

- as we do for field definitions, when the definitional `=` is followed by either a function, a record literal, an array, etc., apply a special casing that remove the indentation level and the new line after `(`
- special case of an unary application where the argument is a function definition inside parentheses, that is cases like of `std.contract.custom (fun label value ...)` to not even add a newline between `(` and `fun`, so that we could have
  ```nickel
  std.contract.custom (fun label value =>
    'Ok value
  )
  ```

Both solutions aren't mutually exclusive.

### Bulk un-indentation

A more brutal approach is to just not add a new line and an additional indentation level for stuff within parentheses. I think it's reasonable because containers (records and arrays) don't need parentheses anyway, so the most common use-cases for them is either to define anonymous functions or for nested applications. Those forms already add their own level of indentation. It's possible to find other cases, such as a long sequence of infix operators, but they don't seem be very common, and would probably fine as well. For example, here would be the result for a boolean expression:

```nickel
(foo
&& bar
&& baz
&& qux
)
```

I did this bulk change on an experimental branch. You can see [the result of formatting the stdlib](https://github.com/tweag/nickel/compare/junk/format-parentheses). I think all changes are reasonable.

### Closing parenthesis

A last minor point is, if we don't always add a newline after an opening `(` in a multi-line context, should we keep `)` on its own new line, as in:

```nickel
std.contract.custom (fun label value =>
  'Ok value
)
```

Or rather put it on the last line of the content, as in:

```nickel
std.contract.custom (fun label value =>
  'Ok value)
```

I'm personally in favor of the former, because I find it more readable to have matching delimiters on the same column. But one could argue that the latter is more symmetric (if we don't add a newline for `(`, we shouldn't add one for `)`).

## Bikeshedding

To sum up, here are the points that I would like to open for discussion:

- [ ] Should we special case `(` depending on what follows, or should we always remove the currently added new line and the indentation level?
- [ ] Should we put the last `)` on its own line like before, or rather at the end of the previous expression?
- [ ] Should we special case unary function application to a function definition to get something like
  ```nickel
  foo_fun (fun bar baz =>
    ... some stuff ...
  )
  ```

  instead of

  ```nickel
  foo_fun
    (fun bar => baz
      ... some stuff ...
    )
  ```

yannham:
Closes #2003.

Formatting of Nickel code has been updated in Topiary to better anticipate the usage of `std.contract.custom` and `std.contract.from_predicate`. This commit updates to the latest version, and update the formatting of the stdlib, the examples and the snippets of the manual.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3131,9 +3131,9 @@ dependencies = [
 
 [[package]]
 name = "topiary-queries"
-version = "0.4.1"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89106b9c504a9e3247b8e40f879a73a922f0d516e1051a6ef739b4cf7d54c8d9"
+checksum = "103bcbfd5c605f2ef94074ae4950473f9dcf159f8bb01959a13856f96a2ffc8c"
 
 [[package]]
 name = "topiary-tree-sitter-facade"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ metrics = "0.21"
 metrics-util = "0.15"
 
 topiary-core = "0.4.0"
-topiary-queries = { version = "0.4.1", default-features = false, features = ["nickel"] }
+topiary-queries = { version = "0.4.2", default-features = false, features = ["nickel"] }
 # This version should agree with the topiary-query version: Nickel queries
 # target a specific version of the Nickel grammar (though that dependency
 # doesn't appear explicitly in `topiary-queries`'s Cargo.toml file). For now you
--- a/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_at_empty_array.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_at_empty_array.ncl.snap
@@ -4,9 +4,9 @@ expression: err
 ---
 error: contract broken by the caller of `at`
        invalid array indexing
-    ┌─ <stdlib/std.ncl>:166:9
+    ┌─ <stdlib/std.ncl>:164:9
     │
-166 │       | std.contract.unstable.IndexedArrayFun 'Index
+164 │       | std.contract.unstable.IndexedArrayFun 'Index
     │         -------------------------------------------- expected type
     │
     ┌─ [INPUTS_PATH]/errors/array_at_empty_array.ncl:3:16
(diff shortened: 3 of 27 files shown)


jneem (approved):

jneem:
I guess the flake's topiary version will also need to be bumped for ci to be happy.

PR #2036, opened 2024-09-10
Explicit imports: `import 'Raw "sample.html"`

### LINKED ISSUE #1000: Add a string import builtin
Sometimes it would be useful to be able to import a file as a string without trying to parse it. E.g.
```nickel
{
  ssh_authorized_keys = m%"
    %{string.import "./id_rsa.pub"}
    %{string.import "./id_ed25519.pub"}
  "%
}
```

### LINKED ISSUE #2033: Where is  "import" formally documented?
I see `let ... = import ... in` several times though the [documentation](https://nickel-lang.org/user-manual/), but don't see a specific page or section regarding modules, imports and multi-file projects. [Standard library documentation](https://nickel-lang.org/stdlib) also does not contain something like an `import` built-in function.

Shall documentation about multi-file Nickel projects be extended?

vi:
Resolves #2033.
Resolves #1000.

Shall I continue this (i.e. try to make tests, error handling, docs, etc.) or it should be done some other way?

Shall it be based on master or e.g. on #2022?


core/src/term/mod.rs
```diff
@@ -207,7 +207,7 @@ pub enum Term {
 
     /// An unresolved import.
     #[serde(skip)]
-    Import(OsString),
+    Import{path: OsString, typ: Option<LocIdent>},
```
jneem, line 210: I'd suggest having `typ: InputFormat` here, and handling the supported types (including the auto-detection from the filename) while constructing the `Term`.

jneem (commented):
The approach looks reasonable to me. I guess we might want to bikeshed the syntax a little...

core/src/term/mod.rs
vi: 1. How would it interact with optionally present format like `Nix`? Ideally it should recognize it and preserve during round-tripping and only fail when actually trying to import the file.
2. Does it move complexity to the `grammar.lalrpop`? (It's the first time I edit such a file)

vi:
What do you think about fallback to `InputFormat::Nickel` when filename extension cannot be recognized? Maybe that `#[default]` is not a good idea?

Falling back from passive format to code format may affect security if there is untrusted data near filename, if `.txt` appender slips somehow, Nickel would suddenly start executing included (using plain syntax) file instead of just reading a config or text snippet.

core/src/term/mod.rs
jneem: 1. Good question. I think since `InputFormat::Nix` is experimental, it's ok to error if it isn't supported.
2. I'd suggest adding the actual logic in `core/src/parser/utils.rs`, so that the changes to `grammar.lalrpop` will be very simple.

vi:
> I'd suggest having typ: InputFormat here, and handling the supported types (including the auto-detection from the filename) while constructing the Term.
> I'd suggest adding the actual logic in core/src/parser/utils.rs, so that the changes to grammar.lalrpop will be very simple.

Done.

Some questions:

* Is it OK to requiring the tag when filename extension is unknown?
* Shall it preserve whether the import was explicit or implicit in `Term`, so that it can pretty-print it back? Or just always pretty-print with a tag? Or auto-detect when filename extension is good and omit the tag?
* What is `new_from_inputs`? Is it OK to hard code it to `Nickel` inputs or they should be auto-detected like other files?

jneem:
> Is it OK to require the tag when filename extension is unknown?

I think for backwards-compatibility we should fall back to the default when the extension is unknown. (But because the tag is a new feature, it's fine to error for an unknown tag.)

> Shall it preserve whether the import was explicit or implicit in Term, so that it can pretty-print it back? Or just always pretty-print with a tag? Or auto-detect when filename extension is good and omit the tag?

I *think* any of those options is ok; maybe @yannham has some input. The important thing for the pretty-printer is that it round-trips when doing (pretty-print -> parse). It doesn't need to round-trip when doing (parse -> pretty-print).

> What is new_from_inputs? Is it OK to hard code it to Nickel inputs or they should be auto-detected like other files?

You mean `Program::new_from_inputs`? I think all the `Program` constructors can assume nickel format.

> Shall there be aliases, like 'Txt or 'Text as alternative for 'Raw, or 'Ncl as alternative for 'Nickel?

My preference is to not have unnecessary aliases, but maybe other people feel differently...

core/src/error/mod.rs
```diff
@@ -583,6 +583,9 @@ pub enum ParseError {
     /// time, there are a set of expressions that can be excluded syntactically. Currently, it's
     /// mostly constants.
     InvalidContract(RawSpan),
+    /// Attempt to specify an import, type of which is not known at the moment of compilation.
+    /// `explicit` determines whether explicit import type annotation was used
```
yannham, line 587: Once again, I fear `type` is a bit ambiguous here (because imports can have a type as well). I suggest we use `format` consistently in the documentation and error reporting instead.

yannham (commented):
Thanks for tackling this! Beside the syntax bikeshedding (which can be done later), and the `type` -> `format` question, this PR looks good.

yannham:
By the way, the  CI is failing because some code in the `lsp/nls` crate needs to be updated now that `Term::Import` has changed.

vi:
> I think for backwards-compatibility we should fall back to the default when the extension is unknown.

Shall a warning be issues instead of parse error in that case?

(Does/should Nickel in general have warnings?)

vi:
Implemented additional changes:

* Rename `typ` to `format`. Moved code that deals with the format tags closer to the code that deals with filename extensions.
* Brought back the fallback to Nickel files (the function is deliberately kept one step away from implementing the error again though).
* Implement pretty-printer part (it omits the tag when needed and adds `'Nickel` if previous code relied on the fallback).
* Fix compilation of NLP (though not implemented completions for the format tags)
* Added imports section to the book.

yannham (commented):
Looks good! It's just missing some tests. You can take inspiration from `nickel/core/tests/integration/inputs/imports`, where the test files for imports are located, and `nickel/core/tests/integration/inputs/imports/imported`, which hosts the files imported by the former tests. If you put a `.ncl` file in `imported` which isn't a stand-alone test but just designed to be imported by another a test, as most of them are in `imported`, you'll need to add a heading annotation `# test.type = 'skip'`.

If you're not confident or can't devote more time to this, feel free to let us know and we can also take it from here.

vi:
Added some tests (without exhaustive tests for each possible input format).

Note that when the same file is imported twice with different formats, the cache remembers the first import and ignores the tag on further ones. Shall we address this potential confusion source? I see two main ways:

* Enforce the consistency. Remember the format in cache and verify that the format is correct when attempting to retrieve it from cache. Prototype: https://github.com/vi/nickel/commit/3ae48a9fbb2b5ff3ce2ffe7a30a323e4be79e12f
* Use input format as a part of the key for cache, allowing caching the same file multiple times (for distinct `InputFormat`s)

jneem:
I prefer the second option. I can imagine some use-case where you want to import something as a structured format *and* as raw text.

vi:
Shall I include the implementation (https://github.com/vi/nickel/commit/cfc46aa5901779cc101f6671176847da10739805) into this branch?

This is a more invasive change compared to just some explicit tag support.

yannham:
@vi Yes, I think it's fine to join both here. Splitting is good usually but the second change is really motivated by the current PR and has a ~ +30/-30 diff, which IMHO is entirely reasonable (and the current PR isn't too big either).

vi:
Something's failed on Windows, it is not very clear from CI logs.

Does Windows mangle `\n` to `\r\n` during checkout or file read?

yannham:
Mh, that's very likely, yeah. One possibility is to remove the newlines at the end of the files and call it a day. Another possibility is to do a regexp match, for example to just check the suffix, or by matching and end of line pattern that will fit both Unix and Windows `\\r?\\n` or something like that.

We should probably have something like an OS-dependent `std.string.new_line` in the future, but for this PR we'll have to do without.

vi:
Is there anything more to be done here (e.g. a syntax change for #2039) ?

yannham (approved):
Nope, we can handle this in a second step. Thanks a lot for carrying this through @vi !

vi:
Noticed possible regression:

```
$ cargo run -p  nickel-lang-cli eval q.q q.q
...
error: unbound identifier `import'Nickel`
  ┌─ <generated main>:1:2
  │
1 │ (import'Nickel "q.q") & (import'Nickel "q.q")
  │  ^^^^^^^^^^^^^ this identifier is unbound
```

yannham:
I  guess the pretty printer is at fault, for not inserting a space between `import` and `'Nickel`. However we will end up with a different syntax, so I think we shouldn't bother fixing this issue separately.

PR #2042, opened 2024-09-16
Fix unsound record contract deduplication

### LINKED ISSUE #2041: Multiple contracts on a record not working as intended
**Describe the bug**
Typically, the following code would apply multiple contracts on a value and if value is a negative number it would fail on the second part of the contract.

```nickel
-123 | Number | std.number.PosNat 
```

However, this is not the case when applying multiple record contracts and both have the same set of fields. The snippet below should have failed when evaluating `container` but it doesn't.  Based on my testing, this only happens when both inner and outer have the same of fields and adding an additional field to either contracts would give the expected behavior.

```nickel
let outer = {
  spec
    | {
      hostAliases | Number | optional,
      containers | Number | optional,
      ..
    }
    | optional,
  ..
}
in
let inner = {

  spec
    | {
      hostAliases
        | optional,
      containers
        | Array {
          ..
        },
      ..
    }
    | optional,
  ..
}
in
{
  spec = {
    hostAliases = "this should have failed",
    containers = [
      {
        name = "test",
        image = "nginx",
        ports = [
          {
            containerPort = 80,
            name = "http"
          }
        ],
      }
    ]
  }
} | inner | outer 
```

**Expected behavior**
`{} | contract1 |  contract2` should work when both contracts run checks against the same set of fields. 

**Environment**
 - OS name + version: NixOS 24..05
 - Version of the code: 1.8.0

yannham:
Closes #2041.

As hinted in the original issue, the culprit is indeed the contract deduplication optimization, and more specifically the contract equality checker. The checker only looks at the fields' values to establish equality without looking at the metadata, which is obviously wrong, as it would equate most contracts with the same structures such as `{foo | Number}` and `{foo | String}`, since the distinction is in the metadata.

This commit fixes the issue by comparing the field metadata as well.

### DIFF
--- a/core/src/typecheck/eq.rs
+++ b/core/src/typecheck/eq.rs
@@ -548,6 +548,11 @@ fn erows_as_map<E: TermEnvironment>(
 
 /// Check for contract equality between record fields. Fields are equal if they are both without a
 /// definition, or are both defined and their values are equal.
+///
+/// The attached metadata must be equal as well: most record contracts are written as field with
+/// metadata but without definition. For example, take `{ foo | {bar | Number}}` and `{foo | {bar |
+/// String}}`. Those two record contracts are obviously not equal, but to know that, we have to
+/// look at the contracts of each bar field.
 fn contract_eq_fields<E: TermEnvironment>(
     state: &mut State,
     field1: &Field,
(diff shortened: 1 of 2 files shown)


yannham:
TODO:

- [ ] add regression test
- [ ] check the impact on the performance on the private benchmark

core/src/typecheck/eq.rs
```diff
@@ -548,20 +548,69 @@ fn erows_as_map<E: TermEnvironment>(
...
     env1: &E,
     field2: &Field,
     env2: &E,
 ) -> bool {
-    match (&field1.value, &field2.value) {
+    // Check that the pending contracts are equal.
+    //
+    // [^contract-eq-ignore-label]: We mostly ignore the label here, which doesn't impact the fact
+    // that a contract blame or not. Different labels might lead to different error messages,
+    // though. Not that there is one important exception: the field `type_environment` does impact
```
jneem, line 567:
```suggestion
    // though. Note that there is one important exception: the field `type_environment` does impact
```

jneem (approved):

PR #2045, opened 2024-09-17
[RFC007] Bytecode interpreter

yannham:
Although the name is a bit pompous, the goal of this RFC is mostly to be a working document for designing a more compact and efficient run-time representation for Nickel expressions.

While this is something that won't be user-facing (at least in a direct way), and thus can be changed later without breaking backward-compatibility, I think the technical scope of this effort is such that I find it better to discuss it formally here before going for a first implementation.

### DIFF
--- a/rfcs/007-bytecode-interpreter.md
+++ b/rfcs/007-bytecode-interpreter.md
@@ -266,8 +266,74 @@ mostly for the GHCi REPL. This section describes what we know of the actual the
 bytecode interpreter, but also incorporates some aspects of the original STG
 machine which is the basis of the low-level operational semantics of Haskell.
 
-### Memory representation
+#### Memory representation
 
+Closures (thunks and functions) are represented uniformly as an environment and
+a code pointer. In some other machines for lazy lambda calculi, thunks hold
+additional data such as a flag (evaluated or suspended) and a potential value
+that is filled after the thunk has been forced. When accessing a thunk, the
+caller checks the state, and then either retrieve the value or initiate an
+update sequence and retrieve the code of the unevaluated expression. In Haskell,
+the update process performed by the callee instead (the thunk's code), such that
+thunk access is uniform: just jump to the corresponding code. As with other VMs,
+the environment - immediately inlined after the code pointer - stores the free
+variables of the expression that are captured by the closure. When entering a
+closure, a dedicated register holds the pointer to the environment for fast
+access.
+
+The STG paper argues that this simplifies the compilation process and gives room
+for some specific optimizations (vectored return for pattern matching, for
+example) that should be beneficial to Haskell programs.
+
+In Haskell, every data is considered to be an algebraic data type including
+primitive types such as integers, which is just `data Int = MkInt Int#`  where
+`Int#` is a native machine integer. The rationale is to avoid having types being
+blessed by the compiler with some magic while ADTs, which are ubiquitous in
+functional programming, are second-class citizen (with respect to optimizations
+and performance). In consequence, ADTs are the foundation of data structures.
+
+Both data and closures are represented the same way. A data value is a block
+with a code pointer, which corresponds to the corresponding ADT constructor.
+Instead of the environment in the case of closures, the code pointer is followed
+by the constructor's argument. As each constructor usage potentially generates
+very similar code, GHC is smart enough to share common constructors instead of
+generating it again and again. Specific optimizations or compilation schemes are
+implemented to alleviate the cost of the (code) pointer indirection for data
+values.
+
+In practice, when compiled to native code, the closures
+
+#### Virtual machine
+
+The STG machine has five components:
+
+- the code
+- the argument stack which contains values. Values are either a heap address
+    `Addr a` or a primitive  integer value `Int n`.
+- the return stack which contains continuations
+- the update stack which contains update frames
+- the heap storing (only) closures
+- the global environment which gives the addresses of closures defined at the top
+    level
+
+When compiling to native machine code, the three stacks (argument, return and
+update) are merged into two stacks - but it's only for garbage collection
+reasons: there is a pointer stack, and a value stack, as the garbage collector
+couldn't tell the difference otherwise. If possible,it's simpler and usually
+better to just merge all the stacks into one, as long as they grow
+synchronously.
+
+The environment, which is just an abstract map data structure in the STG
+operational semantics is split in the actual implementation between the stack
+(arguments), the environment register (captured variables or _upvalues_) and the
+heap register. Local variables introduced by let-bindings are allocated by
+bumping the heap register and are accessible directly from there: the heap is
+indeed just a bump-allocator than triggers copying garbage collection when full.
+In particular, local let-bound variables aren't pushed to the stack unless there
+is a context switch caused by eager evaluation, that is a case expression
+forcing an argument. In this case, each live variable not on the stack (id est
+let-bound variables and captured variables) are stored on the stack prior to the
+context switch.
 
 #### References
 
@@ -371,7 +437,28 @@ functions.
 
 The code consist of an instruction segment (`code`), a value segment
 (`constant`), and a span segment. The code is an array of operations each
-encoded on one byte.
+encoded on one byte. Some instructions actually have operands, that are decoded
+by the VM when needed.
+
+Although Nix is a lazy language, Tvix departs from the push-enter model
+implemented in OCaml's ZAM and Haskell's STG by having a proper `call`
+instruction. Compiling an expression in Tvix results in leaving some result on
+the stack. When evaluating an application `f x`, code is generated for `x`,
+followed by `f`, and finally a `CALL` opcode is generated. On paper at least,
+this looks less adapted for partial application and currying, at least without
+further optimizations. On the other hand, this plays well with the way Tvix
+represents local variables, which are directly stored on the stack (of course,
+the thunks bound by the variables are stored on the heap but their address
+remains on the stack).
+
+Looking at the implementation, it also seems that local stack slots are only
+deallocated after the whole code of the function has ben run. But in a lazy
+language a variable in head position is basically a tail-call, so it looks like
+Tvix is retaining much more memory than necessary: for example, in `let f = x:
+_continuation_; in let a = 1 + 1; in let b = a + 2; in f b`, `a` and `b` looks
+like they would remain alive on the stack for the whole duration of
+`_continuation_` while they aren't accessible anymore. But I might be missing
+something as well.
 
 #### References
 
@@ -383,10 +470,11 @@ encoded on one byte.
     STG-machine that copying the upvalues into the closure is actually one of
     the most efficient in practice (smarter strategies don't gain much time and
     can cost a lot of memory). So we should probably just do basic closure
-    conversion. Not sure the LUA approach is really useful as a first shot,
+    conversion. Not sure the Lua approach is really useful as a first shot,
     because as I understand it, you need to produce the upvalue move instruction
     when a scope ends.
-- Record representation: it's honestly a bit irrelevant, because it can evolve.
+- Record representation: it's honestly a bit irrelevant, because it can evolve
+    independently from the bytecode instructions and other representations.
     Maybe we should use persistent data structure? Still, there might be some
     specific stuff related to the recursive environment and the "lazyness" of
     the recursive environment. Maybe compile that in the same way as OCaml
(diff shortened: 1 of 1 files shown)


yannham:
Some parts might need refinement, but I think it's in a good shape for a first external review.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,834 @@
...
+operation call, closure constructions, thunks update, and primitive operations.
+
+### Intermediate representations
+
+We propose the usage of two intermediate representations.
+
+#### AST
+
+The first one is an AST and would more or less correspond to the current unique
+representation, minus runtime-specific constructors. We could have gone closer
```
jneem, line 641: I think we can also get rid of some efficiency-oriented duplication in the current representation, like the distinction between `LetPattern`/`Let` and `RecRecord`/`Record`. Getting rid of these would be convenient for both the LSP and the typechecker, I think.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,834 @@
...
+
+We leave the precise representation of records and arrays open for now. Their
+actual representation doesn't impact the rest of the proposal and can be
+optimized at will in the future. The current representation works fine.
+
+One simple optimization worth mentioning would be to special case the empty
+record and the empty array, for example with `enum Record { Empty,
+NonEmpty(RecordData) }`. This should use the same space as `RecordData` in Rust
+(if `RecordData` is a pointer, at least) and save an allocation for empty
+structures.
```
jneem, line 722: I'm a bit confused about what discriminants are present in the "top-level" representation (I'm not sure what's the right term, but I'm talking about the word-sized thing that packs in a pointer with some discriminant). Above, it sounded like we were only going to inline `null` and `boolean` in the discriminant; here, you're also proposing the put empty records and empty arrays? It seems like there isn't enough room in the pointer alignment for all of these.

By the way, x86_64, aarch64, and riscv-64 all max out at 48 bits of address space. So on these architectures we can pack lots more stuff at the most-significant end of the top-level representation.

jneem (approved):

rfcs/007-bytecode-interpreter.md
yannham: Agreed. In fact I started to draft the first representation and did get rid of the `Let` and `Fun` to keep only the pattern variants.

rfcs/007-bytecode-interpreter.md
yannham: > By the way, x86_64, aarch64, and riscv-64 all max out at 48 bits of address space. So on these architectures we can pack lots more stuff at the most-significant end of the top-level representation.

Yes, but it's less portable. Although we don't need Nickel to run on embedded, I think we can do with one bit for now, and explore those other possibilities later.

> I'm a bit confused about what discriminants are present in the "top-level" representation (I'm not sure what's the right term, but I'm talking about the word-sized thing that packs in a pointer with some discriminant). Above, it sounded like we were only going to inline null and boolean in the discriminant; here, you're also proposing the put empty records and empty arrays? It seems like there isn't enough room in the pointer alignment for all of these.

Your first understanding is right. At the top-level, there is only one discriminant for `bool`, `null` and `pointer`. If we follow the pointer, we find many representations: arrays, numbers, etc. Here I'm talking about the representation of the pointee, which can itself be a pointer to something else (typically I guess your immutable vec representation would be mostly a pointer to the root plus some parameters).  Somehow the 1-word representation can hold any data, and I'm talking about this specific data when the pointee represents a record. Does that make sense?

yannham: Also, as we need a discriminant at the beginning of the pointee (is it an Array? a Record? etc.) that will need to be aligned (although it might be merged with some other metadata as well), we'll probably have some more space here, and can even special case `EmptyArray` and `EmptyRecord` as special discriminants, instead of bothering making `Record` actually an enum.

jneem: Even on 32-bit, we can have lots more values at the top-level, right? Assuming our pointers are all 4-byte aligned, anything ending in 01, 10, and 11 is not a pointer. But then for each of those non-pointer values we still have 30 bits left to store actual data. So couldn't we have variants for `EmptyArray` and `EmptyRecord` without even following that single top-level pointer? And we'd still have room for small integers...

yannham: Sure, we have a lot more room. It's just that you can't encode anything that needs at least a full machine word or more - typically OCaml needs to have 31-bits and 63-bits integers so that they can unbox them, which is the case of most other non trivial data structures. But indeed special values like empty stuff could in theory be also directly put that top-level.

jneem: Ok, makes sense. I was just confused about what discriminants were where. Our current `Vector` has a `root: Option<Rc<Node>>`, so it's already avoiding allocation for empty arrays.

rfcs/007-bytecode-interpreter.md
yannham: In the end I've mentioned this simplification explicitly.

rfcs/007-bytecode-interpreter.md
yannham: I've tried to make this part clearer: instead of mentioning empty records, I've moved the discussion to when we introduce the first level representation, and list many possible inlining (small numbers, small strings) that include empty records and arrays. It feels less ad-hoc.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+
+- [The Implementation of LUA 5](https://www.lua.org/doc/jucs05.pdf)
+
+### OCaml
+
+#### Memory representation
+
+The following notes on the memory representation applies to the native code
+backend's representation. I'm not sure how closures are represented in the Zinc
+Abstract Machine.
```
aspiwack, line 211: I'm not sure it ought to be called Zinc anymore, but anyway: I'm pretty sure that the representation of values (including closures) is the same in native and bytecode. It must be so at least to some degree for the sake of the FFI, where Ocaml values can be manipulated.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+
+```text
+| header (1w)        | content   |
+----------------------------------
+| size | color | tag | data ...  |
+```
+
+ADTs are represented within the block's data as an integer for constructors with
+no argument (the tag byte then doesn't store the actual contructor's tag but has
+the same value than for a boxed `int`). For a variant with parameters, the tag
```
aspiwack, line 239: This is not quite it. A constructor without argument is represented as an _unboxed_ integer value. It doesn't point to a block, so there is no tag involved. Constructors with arguments are pointers, and point to a structure as above.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+  the ML language_ (1990)](https://inria.hal.science/inria-00070049/document)
+- [OCaml bytecode instructions](http://cadmium.x9c.fr/distrib/caml-instructions.pdf)
+- [OCaml memory representation of values](https://ocaml.org/docs/memory-representation)
+- [From Krivine's machine to the Caml implementations](https://xavierleroy.org/talks/zam-kazam05.pdf)
+- [OCaml's representation of closures](https://github.com/Gbury/ocaml-memgraph/blob/master/doc/closures.md)
+
+### Haskell
+
+Despite not being advertised, Haskell has an interpreter as well, which is used
+mostly for the GHCi REPL. This section describes what we know of the actual the
```
aspiwack, line 300: An Template Haskell.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+pointer and other garbage-collection related data. The closure info table is
+generated only once per binding appearing in the code and thus shared by dynamic
+instances of the same code. In some other machines for lazy lambda calculi,
+thunks hold additional data such as a flag (evaluated or suspended) and a
+potential value that is filled after the thunk has been forced. When accessing a
+thunk, the caller checks the state, and then either retrieve the value or
+initiate an update sequence and retrieve the code of the unevaluated expression.
+In Haskell, the update process is performed _by the callee_ instead (the thunk's
+code), such that thunk access is uniform: it's an unconditional jump to the
+corresponding code.
```
aspiwack, line 317: In fact, when pointing to an info table, you actually point directly to the code pointer. The metadata in the info table is accessed backwards by _subtracting_ from the pointer. This way entering a thunk is an indirect jump, which many processors support in a single instruction.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+
+As with other VMs, the environment -- immediately inlined after the code pointer
+-- stores the free variables of the expression that are captured by the closure.
+When entering a closure, a dedicated register holds the pointer to the
+environment for fast access.
+
+The STG paper argues that this uniform thunk representation (with "self-handled
+update") simplifies the compilation process and gives room for some specific
+optimizations (vectored return for pattern matching, for example) that should be
+beneficial to Haskell programs.
```
aspiwack, line 327: It's worth noting that since the STG paper was written, things have become a little more complex. GHC uses some pointer tagging to mark if the thunk is already forced in the form of one of the 3 (on 32 bits) or 7 (on 64 bits) constructors of the data type. So pattern-matching will check these bits before entering the closure. For efficiency.

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+low-level case, more primitive than the fancy pattern matching of the surface
+language), which forces its argument.
+
+Both data and closures are laid out in a similar way in memory. A data value is
+a block with a code pointer representing the corresponding ADT constructor. The
+code pointer is followed by the constructor's argument (instead of the
+environment in the case of closures). As each constructor usage potentially
+generates very similar code, GHC is smart enough to share common constructors
+instead of generating them again and again (typically the one for an empty
+list).
```
aspiwack, line 344: An interesting difference between GHC's memory representation and Ocaml's is the way unboxed and boxed values are distinguished. This is only a concern for the GC, so maybe it doesn't matter too much for this discussion.

- In Ocaml, if a value, viewed as an unsigned int, is odd, then it's unboxed (and the GC doesn't follow it), when it's even, it's a pointer to a boxed value. This way it's always evident, when looking at a value whether it's a pointer or not. The cost is that integers lose one bit, and that it becomes impossible to unbox floats in most cases.
- In Haskell, pointers and non-pointer values are indistinguishable. So the GC asks the info table which of the fields are pointers. In practice, this is just a number, and all the non-pointer fields are stored first, and all the pointer fields after (or maybe the other way around, I don't remember).

rfcs/007-bytecode-interpreter.md
```diff
@@ -0,0 +1,839 @@
...
+
+### Virtual machine architecture
+
+The virtual machine is a stack-based machine with a set of working registers.
+
+It follows the same general principles as the ZAM, the Tvix VM or the STG
+machine. Those machines still have a number of differences on how they handle
+fundamental operations such as function application. We take inspiration from
+the
+[call-by-push-value](https://www.cs.bham.ac.uk/~pbl/papers/thesisqmwphd.pdf)(CBPV)
```
aspiwack, line 561: :blush:

aspiwack (commented):
Some random comments.

rfcs/007-bytecode-interpreter.md
yannham: You're right, I don't know where I got this idea

rfcs/007-bytecode-interpreter.md
yannham: Ah, right; I think it's mentioned in the STG paper that this was difficult to do with the approach proposed there going through ANSI C, but easy to do with a custom native code generation backend, which I suppose is what GHC does today.

rfcs/007-bytecode-interpreter.md
yannham: *unevaluated thunk, rather

aspiwack: Maybe I said that, I don't know. But I think my broader point was that I was thinking, at the time, of metadata (doc, default, …) as being attached to thunks. And CBPV helps us understand what's going on, because of thunking being an explicit operation in the language, so you can change what's going on there.

That being said, I always think of any evaluation model in terms of CBPV/polarised system L. Even when I don't present it that way: it's almost certainly a translation of the CPBV I have in my head. And I'm keen to embrace it in abstract machines as well.

aspiwack: Speaking of, you can think of system L as representing an abstract machine with a structured stack. Implemented directly, it's not as cache friendly as an actual stack machine. But maybe it's a middle ground to consider, as you don't have to come up with a complex linearised continuation representation for pattern matching, in particular.

On the other hand, maybe it's too tree-like for this proposal. I don't know, I haven't thought about it. I just said the name, and this idea popped up in response. Do whatever you want to do with it.

PR #2051, opened 2024-09-24
Use let-blocks when apprioriate in stdlib & examples

yannham:
This PR updates to Topiary 0.5.1, which includes the support for formatting let-blocks (and the let-in formatting fix). Additionally, I made a pass on the examples and the stdlib to use let-blocks whenever it seemed appropriate, so that both remains idiomatic Nickel.

This is the last item on the checklist before being able to release Nickel 1.9

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,28 +178,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
-name = "async-stream"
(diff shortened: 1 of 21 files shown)


yannham:
This is actually also block on https://github.com/tweag/topiary/pull/747 to be able to update to latest Topiary in the flake.

yannham:
Now that we've got rid of Topiary as an input of the Nickel flake, we need Topiary to be compatible with tree-sitter 0.22 (used by the Nickel grammar) - which is the case, but this version hasn't been released yet. Waiting for a new version of Topiary to be released, so that we can finally format let-blocks.

PR #2052, opened 2024-09-26
Thunks for resolved imports

### LINKED ISSUE #1722: No detection of infinite recursion during evaluation of imports
**Describe the bug**

When a Nickel file attempts to do recursive imports, the evaluation process may hang indefinitely, and sometimes consumes all available RAM. This could happen due to user error.

**To Reproduce** 

1. Create a Nickel file (e.g., recursive_import.ncl) with the following content:

```nix
# recursive_import.ncl
# This only hangs
import "./recursive_import.ncl"

# recursive_import.ncl
# This hangs and consumes all RAM
let a = import "./recursive_import.ncl" in a
``` 

2. Attempt to evaluate the file using Nickel.
```shell
nickel eval ./recursive_import.ncl
```

**Expected behavior**

Nickel should detect recursive imports and gracefully handle them, either by preventing the evaluation recursion or by providing a clear error message, to avoid hanging during evaluation and consuming excessive resources.

**Environment**
 - OS name + version: Pop_OS! 22.04 (Ubuntu Jammy)
 - Version of the code: nickel-lang-cli nickel 1.3.0 (rev cargore)

**Additional context**

The example above is a simplified way of showing this bug. This issue was found during testing on my use case, with 2 files, due to a typo:
- File A evaluates to a list of records from imports, and a list of functions.
- File B requires some of those functions, and it's resulting record is present in file A's record list.

This wouldn't necessarily cause an evaluation recursion by itself, but I mistyped something and the imports did result in a recursion, and that crashed my PC by consuming all the RAM (and swap).

In comparison, Nix detects such recursions and reports it with the following error:
```nix
# recursive_import.nix
let a = import ./recursive_import.nix; in a
``` 

```shell
$ nix eval --file ./recursive_import.nix        
error: stack overflow (possible infinite recursion)
```

jneem:
Wraps resolved imports in a thunk, so that the blackholing checks in the evaluator detect infinite recursion.
 
Fixes #1722

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -1,5 +1,6 @@
 //! Source cache.
 
+use crate::closurize::Closurize as _;
 use crate::error::{Error, ImportError, ParseError, ParseErrors, TypecheckError};
 use crate::eval::cache::Cache as EvalCache;
 use crate::eval::Closure;
@@ -231,6 +232,8 @@ pub enum EntryState {
     Transforming,
     /// The entry and its transitive imports have been transformed.
     Transformed,
+    /// The entry has been closurized.
+    Closurized,
 }
 
 pub enum EntryOrigin {}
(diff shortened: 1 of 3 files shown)


core/src/cache.rs
```diff
@@ -811,6 +814,47 @@ impl Cache {
         }
     }
 
+    /// Replace a cache entry by a closurized version of itself. If it contains imports,
+    /// closurize them recursively.
+    ///
+    /// Closurization is not required before evaluation, but it has two benefits:
+    /// - the closurized term uses the evaluation cache, so if it is imported in multiple
+    ///   places then they will share a cache
+    /// - the eval cach's built-in mechanism for preventing infinite recursion will also
```
yannham, line 823:
```suggestion
    /// - the eval cache's built-in mechanism for preventing infinite recursion will also
```

yannham (approved):

Issue #1815, opened 2024-02-09
Infinite loop (and memory usage)


jneem:
When trying to export the following file, nickel gets stuck and starts using up all my memory:

```nickel
{ y = { foo = y } }
```

yannham:
I mean, it is an infinite loop, right? Although we should probably be able to detect that without black-holing, and I'm not sure why we don't.

What behavior do you think would be right?

jneem:
I'd expect the same infinite recursion error that you get from, say `let rec y = y in y`

yannham:
I haven't tested this hypothesis, but in fact I believe it's to be expected that black-holing doesn't detect such loops. In fact, as long as you don't use `deep_seq` (but both `nickel eval` and `nickel export` does use it, which is why it's looping), this term _doesn't loop_, in that it's always productive: if we evaluate `y`, we get a lazy record `{foo = y}`. If we then evaluate `foo`, we get in one the original record `{y = {foo = y}}`, and so on. It's basically an infinite record `{ y = { foo = {foo = { foo = ....`, which is - at each step - at most one variable substitution away from being in weak head normal form form.

So, evaluating `y` always terminates, and doesn't trigger the infinite recursion detection.

However, this is different when forcing/deep_sequing the record, of course, because we are now trying to recurse into an infinite data structure. I wonder if we can extend the blackholing mechanism to handle this case, by forbidding thunks to be evaluated again while we're still forcing the children of this thunk.

D-Brox:
Related to #1722

yannham:
I think also related to #1967 - both the issue and the solution should be pretty similar.

jneem:
I had a look at this today, but I couldn't figure out a version of:
> forbidding thunks to be evaluated again while we're still forcing the children of this thunk
that works. For example, this input is fine:

```nickel
{ y = { foo = z }, z = 1 }
```

but when forcing the child `{ foo = z }`, we end up needing to re-evaluate the outermost thunk.

I guess one thing that might work (and maybe helps in #1967 too) is to have special handling for a field that evaluates to a parent thunk. I don't see how to fit this into the existing black-holing framework though; it seems like it might need an extra stack?

yannham:
[...]

PR #2055, opened 2024-09-27
Detect infinite recursions in `nickel doc`

### LINKED ISSUE #1967: `nickel doc` overflows the stack with recursive schemas
**Describe the bug**
`nickel doc` aborts with a stack overflow on a real-world 1MB Nickel schema, which doesn't look very deep (`nickel doc` does use recursive calls, but the stack space should be linear in the depth of the schema, roughly).

**To Reproduce**
I unfortunately can't provide the test case, which has been provided as a courtesy but isn't public. This issue is mostly a reminder to investigate it.

yannham:
Closes #1967.

This commit fixes an issue where legit recursive configurations cause infinite recursion when trying to extract their documentation. Indeed, `nickel doc` calls to `eval_record_spine`, which is unable to detect loops that spans multiple evaluation steps (an evaluation step is evaluating a field to a WHNF).

This commit adds a `lock` and `unlock` method to the interface of thunks, which amounts to blackholing, and use them in the implementation of `eval_record_spine` such that the thunks of the parents of the field that we are currently evaluating - the active parents - are always locked. If there is a cycle, this will raise an infinite recursion error, that we ignore - we can still print the unevaluated value for the user - but will prevent the evaluation from going on forever.

This technique should be used to fix other similar issues with `force` and `deep_seq`, but this is left for future work.


core/src/eval/cache/lazy.rs
```diff
@@ -405,18 +405,50 @@ impl Thunk {
...
+        if let ThunkState::Blackholed = data_ref.state {
+            data_ref.state = ThunkState::Evaluated;
+        }
+    }
+
     /// Generate an update frame from this thunk and set the state to `Blackholed`. Return an
     /// error if the thunk was already black-holed.
     pub fn mk_update_frame(&mut self) -> Result<ThunkUpdateFrame, BlackholedError> {
-        if self.data.borrow().state == ThunkState::Blackholed {
+        let mut data_ref = self.data.borrow_mut();
```
yannham, line 445: The changes in this function are cosmetic (avoid two calls to `borrow()`/`borrow_mut()` in a row: just borrow mut from the beginning)

core/src/eval/cache/lazy.rs
```diff
@@ -647,18 +681,21 @@ impl Cache for CBNCache {
...
-                idx.mk_update_frame().map(Some)
-            }
-            // If the thunk isn't to be updated, directly set the evaluated flag.
-            else {
-                idx.set_evaluated();
-                Ok(None)
-            }
-        } else {
-            Ok(None)
+        // If the thunk is already evaluated, we don't return an update frame.
```
yannham, line 684: The changes are a bit cosmetic, but not only: previously, if `idx.should_update()` was `false`, we would never look at the state of the thunk and always proceed with evaluation. This is actually not correct, because if the thunk is blackholed, we should raise an infinite recursion error. For eval record spine in particular, the parents thunk are in weak head normal form (they are records), so `should_update()` is false and without this additional change, the detection of the infinite cycle fails.

jneem:
I think the recursion check has some false positives: with this PR, the input:

```
{
  outer = {
    mid = { inner | doc "this is inner" },
    z | doc "this is z" = outer.mid,
  }
}
```

doesn't generate documentation for `outer.z.inner`. But if I replace the value of `z` by `mid` instead of `outer.mid` then it does generate documentation for `outer.z.inner`.

yannham:
I wonder if we can do much better than that, at least using the thunk machinery, because when `z` refers to the thunk `outer`, we can't really decide if this use is problematic or not. It depends if another field of `outer` is then accessed, and even then, just seeing `mid` doesn't tell us if there's an issue or not, as we could have something like:

```
{
  outer = {
    mid = { inner = outer.z | doc "this is inner" },
    z | doc "this is z" = outer.mid,
  }
}
```

Which is looping. So either we think it's  ok to not print documentation in this case because the structure of the configuration is a bit intricate - do we expect configuration from which we extract doc to exhibit this sort of shape? I honestly don't know.

Or we can lower the detection power and accept to have more false negatives, but it becomes very ad-hoc. For example, we could only detect unguarded recursion, that is when a field directly hosts a naked variable as in `z = outer`.

Another possible solution is to try to count the number of times we unwrap the same thunk, and instead to halt at one, halt at some other higher empirical value (with the idea that if you recurse 10 times in the same thunk, there is a great change that this is an infinite loop). It has to be baked in the interpreter though, because  `eval_record_spine` only has access to the thunks stored directly at the level of fields, but in `z = outer.mid`, only the evaluator has the knowledge that `outer` has been used at some point.

yannham:
What makes me wonder if this example is really a good case is that you would rather refer to `mid` directly as a sibling here, and this is handled fine by the implementation of this PR.

However, while it might ok for documentation, the false positives preclude the usage of this technique for deep sequing and forcing, I believe.

yannham:
Maybe one last possible approach, which is I believe what you hinted at in https://github.com/tweag/nickel/issues/1815#issuecomment-2378374768, would be to detect regular trees, that is when a contract or a field evaluates to a parent thunk directly, rather than this PR that is more sensitive and detects mere usage. I would indeed require an additional stack/set of active thunks, but shouldn't be very hard to implement either. Le met give it a try

yannham:
@jneem I implemented the last proposal, which seems to handle both your example and the original repro. I'm not sure it's a good idea to do that for `deep_seq`, as it has a cost - albeit small. I don't know if we want to pay that price for each and every `deep_seq`. I think the trade-off is different for `nickel doc`, where we can afford a bit more machinery to avoid infinite recursion.

In any case, it's read for a review.

core/src/eval/cache/lazy.rs
```diff
@@ -405,18 +405,50 @@ impl Thunk {
...
+    /// Unlock a thunk previously locked (see [Self::lock]). If the thunk is locked or blackholed,
+    /// this method sets the state to [ThunkState::Evaluated]. Otherwise, it's a no-op.
+    pub fn unlock(&self) {
+        let mut data_ref = self.data.borrow_mut();
+
+        if let ThunkState::Blackholed = data_ref.state {
+            data_ref.state = ThunkState::Evaluated;
+        }
+    }
+
```
jneem, line 441: This lock/unlock are no longer used, right?

core/src/eval/cache/lazy.rs
jneem: I guess this code path is no longer triggered in `eval_record_spine`? But maybe it makes sense to do this anyway...

jneem (approved):
Looks good, except that I think there are some left-overs from the previous version.

I haven't thought too much about the semantics, but if performance is a concern during normal eval then I think the state could be moved into the thunks instead of having an external hash set: give each thunk a bit that says whether or not it's active.

yannham:
> I haven't thought too much about the semantics, but if performance is a concern during normal eval then I think the state could be moved into the thunks instead of having an external hash set: give each thunk a bit that says whether or not it's active.

This is a very good point; we could just "mix" the two solutions to get the best of both worlds (not false positive and fast check). For now I'm going to proceed with the current version, but it's good to keep in mind for the `deep_seq` issue.

core/src/eval/cache/lazy.rs
yannham: Yes, I suspect the previous code could miss some infinite recursion errors as well because of that.

core/src/program.rs
```diff
@@ -718,13 +759,27 @@ impl<EC: EvalCache> Program<EC> {
...
-            let result = match result {
-                Err(EvalError::MissingFieldDef { .. }) => return Ok(term),
-                _ => result,
-            }?;
+            result
+        }
 
-            match_sharedterm!(match (result.body.term) {
+        // Evaluates the closure, and if it's a record, recursively evaluate its fields and their
+        // contrats.
```
jneem, line 773:
```suggestion
        // contracts.
```

core/src/program.rs
```diff
@@ -692,23 +714,42 @@ impl<EC: EvalCache> Program<EC> {
...
-                env,
-            });
+
+            let curr_thunk = term.as_ref().try_as_closure();
+
+            if let Some(thunk) = curr_thunk.as_ref() {
+                // If the thunk is already locked, it's the thunk of some parent field, and we stop
+                // here to avoid infinite recursion.
+                if !thunk.lock() {
+                    return Ok(term);
```
jneem, line 742: This seems appropriate for docs, but maybe for the deep_seq case we want to raise an error? (If so, it could be in a subsequent PR)

jneem (approved):

core/src/program.rs
yannham: Sure, but `deep_seq` isn't using `eval_record_spine` or similar functions anyway, so currently this doesn't affect `deep_seq` in any way. I think it's appropriate for the semantics of `eval_record_spine` in general (doc or not).

yannham: (Put differently, `deep_seq` impl will handle locking itself and can very much raise an error upon double-lock)

PR #2061, opened 2024-10-08
Add a flake check for testing the stdlib

jneem:
An amusing little thing that I think is harmless: when a test in the stdlib uses the `std` name (like `std.contract.not`) then it's testing the copy of the stdlib that was compiled into the nickel binary. But when it uses a "local" name (like `HasField`) then it's testing the copy of the stdlib that it gets from the path `./core/stdlib/std.ncl`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -670,6 +670,12 @@
         pkgs.mkShell {
           buildInputs = [ terraform run-terraform update-infra ];
         };
+
+      stdlibTests = pkgs.runCommandLocal "stdlib-test" { }
+        ''
+          ${pkgs.lib.getExe self.packages."${system}".default} test ${./core/stdlib/std.ncl};
+          mkdir $out
+        '';
     in
     rec {
       packages = {
@@ -725,7 +731,7 @@
         # shorter. Another option would be to compile a dev dependencies version
         # of cargoArtifacts. But that almost doubles the cache space.
         nickelWasm = buildNickelWasm { profile = "release"; };
-        inherit vscodeExtension;
+        inherit vscodeExtension stdlibTests;
         pre-commit = pre-commit-builder { };
       };
     }
--- a/flake.nix
+++ b/flake.nix
@@ -673,8 +673,7 @@
 
       stdlibTests = pkgs.runCommandLocal "stdlib-test" { }
         ''
-          ${pkgs.lib.getExe self.packages."${system}".default} test ${./core/stdlib/std.ncl};
-          mkdir $out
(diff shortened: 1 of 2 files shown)


yannham:
> An amusing little thing that I think is harmless: when a test in the stdlib uses the std name (like std.contract.not) then it's testing the copy of the stdlib that was compiled into the nickel binary. But when it uses a "local" name (like HasField) then it's testing the copy of the stdlib that it gets from the path ./core/stdlib/std.ncl.

Yeah, it's the same for the actual stdlib code - often functions refer to other  function of the same module directly, but use `std` to access outer module (in some places this was required because of the typechecking not being the same - I don't know if it's the case anymore with the polymorphism rework but it might be).

flake.nix
```diff
@@ -670,6 +670,12 @@
         pkgs.mkShell {
           buildInputs = [ terraform run-terraform update-infra ];
         };
+
+      stdlibTests = pkgs.runCommandLocal "stdlib-test" { }
+        ''
+          ${pkgs.lib.getExe self.packages."${system}".default} test ${./core/stdlib/std.ncl};
```
yannham, line 676: Isn't the `;` at the end going to make the command pass whatever is the actual result? Maybe a sanity check would be to push a commit with a deliberate error in the stdlib somewhere, just to be sure.

yannham (commented):
Glad to see this moving forward!

flake.nix
jneem: I don't know exactly why (maybe `runCommandLocal` does a `set -xeu` or something?) but it seems to fail if I make it fail.

yannham: But why do you need the `;` ? Can't we use a `&&`? Do we also have to necessarily create `$out` ?

jneem: Sure, I can use `&&` instead; the behavior seems to be the same.

Yes, without making `$out` then `nix flake check` fails with
```
error: builder for '/nix/store/kyr0xs7ncr05i3frnphvvvr1q8p8ja1c-stdlib-test.drv' failed to produce output path for output 'out' at '/nix/store/kyr0xs7ncr05i3frnphvvvr1q8p8ja1c-stdlib-test.drv.chroot/root/nix/store/yil1zc0j0vy5p3sjsp0c1kqi8r0h6krj-stdlib-test'
```

yannham: Thanks, looks good! I think `&&` is indeed better as it doesn't rely on assumptions about how `runCommandLocal` works internally :+1:

yannham (approved):

PR #2063, opened 2024-10-08
Remove deprecated toRustTarget

jneem:
This fixes a deprecation warning and doesn't appear to break anything :crossed_fingers: 

Re-running the nix build after this change didn't actually build anything new, so I think the derivations are identical.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -103,7 +103,6 @@
       mkRust =
         let
           inherit (pkgs.stdenv) hostPlatform;
-          inherit (pkgs.rust) toRustTarget;
         in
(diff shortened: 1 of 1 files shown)


yannham (approved):
It might be good to check that the static binary still builds as well - I think this isn't currently tested in the CI because it's just too long.

jneem:
It does indeed take a while to build. I will finish checking it tomorrow, and merge if everything is ok.

jneem:
I mean, I was checking locally that the static binary builds but it was taking forever.

PR #2064, opened 2024-10-08
Remove the option for an external formatter in nls

### LINKED ISSUE #2062: Building nls with `--no-default-features` fails.
Some APIs changed, and the external formatter doesn't build anymore. We should either fix it (and check in CI that it builds) or remove the option to use an external formatter.

I seem to remember that it was originally made configurable because of some difficulty building pre-releases of topiary. Maybe we don't need the external formatting option anymore?

jneem:
Fixes #2062.


lsp/nls/Cargo.toml
```diff
@@ -14,10 +14,6 @@ version.workspace = true
 name = "nls"
 path = "src/main.rs"
 
-[features]
```
yannham, line 17: I'm not sure we want to remove the `format` feature though. I think the rationale around features right now is to leave the possibility to build a minimal nickel interpreter that can be embedded say on a server, which doesn't need to format, render markdown, etc. - so anything that isn't strictly necessary and pulls in a non trivial set of dependencies is probably better kept behind a feature gate.

jneem: This is just the nls format feature, though. Is there a need for a minimal nls?

yannham: Ah, you're right, my bad.

yannham (approved):

PR #2066, opened 2024-10-09
ci: Add shorter names to CI jobs

YorikSar:
By default GitHub lists all arguments in brackets after job name. This gets unwieldy when we include runner labels in https://github.com/tweag/nickel/pull/2037. For example, Linux job is now "build-and-test (self-hosted, Linux, X64, stable)". Shorten it to "build-and-test (linux, stable)" instead.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,15 +15,14 @@ jobs:
 
     strategy:
       matrix:
-        os:
-          - [self-hosted, Linux, X64]
-          - [self-hosted, macOS, ARM64]
-        rust_channel:
-          - stable
         include:
+          - os: [self-hosted, Linux, X64]
+            name: linux
           - os: [self-hosted, macOS, ARM64]
+            name: macos
             continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -15,15 +15,14 @@ jobs:
...
-        rust_channel:
-          - stable
         include:
+          - os: [self-hosted, Linux, X64]
+            name: linux
           - os: [self-hosted, macOS, ARM64]
+            name: macos
             continue-on-error: true
 
+    name: build-and-test (${{ matrix.name }}, stable)
```
yannham, line 25: As we've removed `rust_channel` from the parameters, shouldn't we even remove the `stable` at the end, if that's the only thing we test anyway ? (also because I suspect we would tweak the flake directly if we want to test other Rust channels, and not the workflow)

YorikSar: Good point. Removed it from everywhere, including windows.

yannham (approved):

PR #2067, opened 2024-10-10
Adds some array benchmarks to the ones tracked in CI

jneem:
This also rewrites the sum benchmark to be (IMO) more idiomatic. It doesn't affect performance much.

### DIFF
--- a/core/benches/arrays.rs
+++ b/core/benches/arrays.rs
@@ -32,6 +32,7 @@ fn ncl_random_array(len: usize) -> String {
     String::from_utf8(out).unwrap()
 }
 
+#[cfg(not(feature = "benchmark-ci"))]
 ncl_bench_group! {
 name = benches;
 config = criterion_config();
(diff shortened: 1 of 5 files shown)


core/benches/arrays.rs
```diff
@@ -32,6 +35,7 @@ fn ncl_random_array(len: usize) -> String {
     String::from_utf8(out).unwrap()
 }
 
+#[cfg(not(feature = "benchmark-ci"))]
```
yannham, line 38: If the whole `ncl_bench_group!` is gated behind the feature, is there a reason why we need to disable individual benchmark with a `#[cfg(not(...))]` as well ?

yannham (approved):

core/benches/arrays.rs
jneem: Without configuring out those things, CI failed with unused code warnings.

Issue #2043, opened 2024-09-17
Accessing environment variables


yannham:
**Is your feature request related to a problem? Please describe.**

A feature request that came up several times on various channels is to be able to access environment variables. This is possible today by performing some pre-processing step (for example, a script dumping the environment to a file or converting it to a `.ncl` file that can be imported), or by using the customize command line, for example:

```nickel
# file: config.ncl
{
  ENV_PATH | String,
  config = {
    new_path = "%{ENV_PATH}:/nix/store/x29idoa934zdoijgoin9nickel-1.8.1/bin",
  }
}
```

```console
$ nickel export config.ncl --field config -- ENV_PATH=\""$PATH"\"
{
  "new_path": "/home/johndoe/.local/bin:/nix/store/x29idoa934zdoijgoin9nickel-1.8.1/bin"
}
```

The pre-processing approach has the obvious drawback of requiring everyone to come up with their own scheme and scripts for a task that seems to be common to many use-cases. The customize-mode approach can quickly lead to very long and unreadable command line (needing some form of scripting anyway, probably). In both cases, it's also not trivial to handles cases where the environment variable has special characters that might be interpreted in the Nickel string, such as `"`, escape sequences or `%{`. It can be solved practically by using a multi-line strings with a complicated enough delimiters, but then new lines and indentation can become an issue as well...

Having a standard built-in solution to this problem is reasonable.

**Describe the solution you'd like**

Using variables from the environment have the potential of breaking Nickel's purity if done wrong. If the result of the program can suddenly depend on unspecified environment variables in an uncontrolled way, it can become very challenging to ensure reproducibility of evaluation across different environments. Thus, whatever we end up with, _we must ensure that which variables are taken from the environment (and from which environment variable) can be easily and immediately decided, without having to evaluate the program_.

This precludes a simple approach such as  a free-form dynamic `std.env.get` that can retrieve any value dynamically from the environment. An explicit mapping of some top-level variables to environment variables might be a better idea.

[...]

PR #2068, opened 2024-10-12
Implement raw string sigils for CLI field customization.

vi:
* `^` to use literal string without interpretation
* `@` to read string from file
* `%` to read string from environment variable

Related to #2043.


yannham:
I personally think sigils are a good way forward; they're simple, easy to add, and there's precedent of their use on the CLI. My personal preference would to not duplicate them too much, at the risk of having a cryptic CLI that resembles a Perl expression: for example we could just use `@` everywhere, but have a kind of URL/protocol specification that is maybe slightly less succinct but easier to understand. For example: `@file:foo.txt`, `@env:ENV_VAR`, etc.

We could have sensible but simplistic default: if nothing is specified, it's a file, and we use the extension to deduce the format. We probably also want a way to specify the format explicitly if needed, so that we can write `@file:text:foo.txt`. This also solves the issue of how to get a Number directly instead of a string: if an environment variable store a number that we want to interpret as such,, we could do `@env:nickel:ENV_VAR`. The default would be text, but this can be discussed.

> Shall they be parsed from grammar.lalrpop, just like regular assignments? If so, I am uncertain how to design it properly.

This isn't required, if the parsing is really trivial. I think field paths on the CLI as in `foo.bar.baz` in `nickel eval config.ncl -- foo.bar.baz=5` used to be parsed separately. But field paths are a bit more involved - you can have escaping, so stuff like `foo."bar.baz".balrg`, the code was already existing the grammar as Nickel needs to parse such field definitions as well. So it made sense to avoid duplication and do everything in the grammar.

This is not the case here though, as the syntax we propose isn't valid Nickel. So as long as we know exactly what characters we expect and we don't have any kind of stack to maintain like for well-bracketed expressions, parsing the sigil manually is most probably ok.

For the rest of your questions, I think `@foo.txt` should behave exactly as sugar for `foo='import "foo.txt"'`. We can even make it actual desugaring: that way, we don't have to duplicate any logic or think too hard about each detail. We can't really do that for environment variables, but this can be a guide: if for any question, the answer is probably how would that behave if we put the same content in a file and imported it instead.

vi:
> if nothing is specified, it's a file, and we use the extension to deduce the format

> if an environment variable store a number that we want to interpret as such

What is security standpoint for Nickel? Should be in general protect against mixing code and data?

Shortcuts like `@$MY_FILE` meaning `@file:text:$MY_FILE` can lead to surprising injections where `MY_FILE` variable can be bent to contain a prefix like `literal:nickel:evil_code` instead of a file name.

Should Nickel strive to make secure way the easiest and avoid easy to code, superficially working, but unsound pitfalls? Filename extension sniffing with a fallback from passive json/yaml/text to active Nickel code is already a step in the wrong direction if it does.

vi:
> But field paths are a bit more involved - you can have escaping, so stuff like foo."bar.baz".balrg, the code was already existing the grammar as Nickel needs to parse such field definitions as well.

Currently implementation does reuse proper parser for the part on the left side of `=` character.

It does assume that `=` cannot happen in the field name part. Is it the case?

vi:
Do I correctly assume that current way of specifying fields on CLI is not stable, so we break backwards compatibility and e.g. require some additional prefix to specify a nontrivial Nickel code as a field value?

yannham:
> Shortcuts like @$MY_FILE meaning @file:text:$MY_FILE can lead to surprising injections where MY_FILE variable can be bent to contain a prefix like literal:nickel:evil_code instead of a file name.

It's important to note that Nickel is entirely pure at the time, and can't dynamically load files. So all an evil file or Nickel snippet could do is...to produce a value (well, or loop indefinitely, there's that). While it's a valid point, code injection isn't the same for Nickel as for say for python.

On one hand I'm tempted to say that it's the user's job to use `@file:text:$MY_FILE` if `$MY_FILE` is untrusted content. On the other hand you're right that we should make the default, simplest approach the safest. One possibility is to always require the `file` part explicitly. Or to chose a syntax where the short one can't be a prefix of the longest one, in some sense. Anyway the `@file:text` was just a suggestion: I just want to vouch for readable syntax rather than having 5 ad-hoc Nickel-specific sigils. I hope we can find a syntax that satisfies both readability and avoid potential injections.

> It would be tricky to just transform @foo.txt to import "foo.txt" reliably, as foo.txt may contain embedded " or %{}s. Escaping, then unescaping looks ugly and can also be a security issue if done poorly.

I'm not sure to understand this part. Are you talking about the filename that can have embedded `"` or `%{}s`, or the _content_ of the file? Note we don't have to go through a round trip of actually printing and parsing Nickel code. We're in the interpreter, so we can generate an `Import` AST node directly without caring about escaping. What I mean is that _the semantics should be the same as if we did the round trip_, not that we have to actually do it exactly as text substitution. That should still answer all the questions like "should this go to the import cache" and whatnot.

> Currently implementation does reuse proper parser for the part on the left side of = character.
> It does assume that = cannot happen in the field name part. Is it the case?

Ah, this is a good point. I was thinking about parsing just the sigil part (the value) but we still have to find the value part and it can be non-trivial: indeed fields can contain `=` as in `foo."bar=baz?".value=true`. So maybe it's safest to do it in the parser. Or maybe we can just take the last `=`?

> Do I correctly assume that current way of specifying fields on CLI is not stable, so we break backwards compatibility and e.g. require some additional prefix to specify a nontrivial Nickel code as a field value?

In theory, yes. In practice, this means like we need to prefix stuff like `true`, `5`, etc. So we have to weight each possibility and see if it's really worth it. Even if it's unstable, people are still probably using it, so breakage must be justified.

vi:
> So all an evil file or Nickel snippet could do is...to produce a value

It can include arbitrary file from a filesystem, which may be unsandboxed. That value may contain secrets.

> Are you talking about the filename that can have embedded " or %{}s, or the content of the file?

The filename.

> Or maybe we can just take the last =?

What if the filename or a literal string contain `=`?

Maybe `CliFieldAssignment` should not parse both `Vec<LocIdent>` (the left-hand part) and `RichTerm` (today's right hand part) in one go, but do it two steps: firstly just `Vec<LocIdent>` with `=` and some arbitrary string chunk after `=`,  then (with the appropriate prefix to indicate arbitrary Nickel expression) it would be a `RichTerm`?

>> require some additional prefix to specify a nontrivial Nickel code as a field value

> In practice, this means like we need to prefix stuff like true, 5, etc.

`5` and `true` counts as a trivial Nickel code. Nontrivial means there can be imports, lets, string interpolation, etc.

If we go the `grammar.lalrpop` way then we can discriminate between trivial terms and complex ones.

yannham:
> What if the filename or a literal string contains a =?
>
>Maybe CliFieldAssignment should not parse both Vec<LocIdent> (the left-hand part) and RichTerm (today's right hand part) in one go, but do it two steps: firstly just Vec<LocIdent> with = and some arbitrary string chunk after =, then (with the appropriate prefix to indicate arbitrary Nickel expression) it would be a RichTerm?

Hmm. Parsing arbitrary stuff is annoying to do in the current parser infrastructure, because the lexer still produces tokens that are tailored to the Nickel language. It's not impossible to do, but might prove annoying - requiring something like a new lexer mode that is toggled by the parser. Maybe the simplest is to add lexer rules for stuff like `@....` instead.

In general I don't see how a prefix for complex expressions is going to help though, because if you do `foo=$VAR` then `$VAR` could just include this prefix. You probably want the converse - have a prefix to restrict what comes after, but then it becomes the user's responsibility to add it, which is a problem, because the natural `foo=$VAR` becomes dangerous (beside files, as you mention, I reckon it could also contain something like `@env:PATH` or `@env:SECRET`).

Maybe what we should do is the following: by default, restrict everything that can be given as a field value on the CLI. That is, we accept string literals, numbers, bools etc - all literals, and maybe arrays and records of literals - and that's it. It's easy to check after parsing, we can just validate the right hand side.

Then for anything more involved, like `foo.bar='import "other.json"'`, Nickel would require the usage of additional flags or permissions - like `--allow import`, and `--allow expression` for more complex expressions. There are small design questions around if you want those flags to apply to one field only, or all, or flags for both cases.

For the sigils, maybe not having shortcuts is the way forward: you always need to write `@file:` and `@env:`, which isn't too bad. And we can have a different separator for the format so that it can't be injected, for example `@file/json:foo.txt`. So when you do `@file:$MYVAR`, `MYVAR` can't change the protocol at all.

`@` is used for array concatenation in Nickel, but if we chose a different character, we can also put the protocol _before_, which would solve the injection issue similarly. Say we use `~` as a sigil (we don't want to, but for the sake of the argument). Then the syntax could be `file~foo.txt`, `file:json~foo.txt`, and `~$MY_VAR` would be a safe shortcut.

vi:
> `foo."bar=baz?".value=true`

How important are those field names containing `=`? Maybe it can be worth to just limit CLI to simpler, reasonable field names and require a workaround (e.g. a small Nickel file that copies one field to another) to set tricky names?

yannham:
> How important are those field names containing =? Maybe it can be worth to just limit CLI to simpler, reasonable field names and require a workaround (e.g. a small Nickel file that copies one field to another) to set tricky names?

I don't expect that to happen a lot. But I would still be a bit disappointed that an implementation detail (it's annoying to parsing) leaks to the interface and someone somewhere on day will get a failure because they use a `=` in their field name and won't understand why.

Thinking about this again, in fact we could find the right `=` by leveraging the lexer. The lexer is indeed able to make the handle strings, so we could lex the assignement and get the first normal `=` token that we found (which will provide a span). This `=` **must** be the one of the assignment, as strings are lexed as separate tokens.

vi:
> Thinking about this again, in fact we could find the right = by leveraging the lexer.

I also though about it.

> the whole thing

Can lalrpop parse prefix of a string; not expect EOF, just stop parsing when a specified token is found?

This can be used even to handle non-Unicode filenames that cannot be put in a `String` - remember proper byte position of the `=`, then apply it to original `OsString` argument to get the filename reliably. Should I try implementing it (introduces complexity) or avoid handling this for now to keep things simpler?

yannham:
> Can lalrpop parse prefix of a string; not expect EOF, just stop parsing when a specified token is found?

Not that I know. LALRPOP just pops tokens from the lexer until it can't parse stuff anymore, to the best of my knowledge. But honestly the more I think about it, the more I think the lexer is the easiest: it's cheap (lexing should be pretty fast), it's simple (take until `=`), it doesn't need to modify the grammar or to do strange LALRPOP hacks...and if there is no sigil, we can just fallback to parsing it again entirely in a normal way. There is no way the cost of lexing the field path two times is even measurable, I suspect.

And if we think that the sigils are the exception and that normal field assignment should be the fast path, we can do the converse: try to parse, if there is an error, restart a lexer and try to sigil-parse, and if both fails report the original parse error. But I'm not sure this will make any difference, honestly.

yannham:
This issue was discussed in the last weekly. We settled on using only one sigil `@` as a prefix, and impose that the user write either `file` or `env`, as in `@file:foo.txt` or `@env:FOOBAR`, even if the thread model discussed here is not very convincing (if you can control which files are imported, I'm not sure there's a lot more you can do by interpreting something as Nickel instead of text for example). Anyway, that decision sidesteps the injection issue.

@vi are you willing/able to take this PR through, or should we take over?

vi:
> impose that the user write either file or env

Shall there be also "@literal:" to embed arbitrary uninterpreted string?

Is the list of those prefixes expected to grow in further pull requests?

> `@file:foo.txt`

What will happen with `@file:foo.html`?

Will it still default to a Nickel file or just fail? Will explicit formats be handled by alternative prefixes like `@file_text:foo.html`?

> @vi are you willing/able to take this PR through, or should we take over?

Maybe someday (was doing other, non-Nickel-related things meanwhile), but not this/next month probably.

If needed, it can be taken over (with some comment here).

yannham:
> Shall there be also "@literal:" to embed arbitrary uninterpreted string?

We can discuss this further afterwards, but right now I just don't see the use case. You can already include literals by either enclosing the whole assignment within single quotes, as in `'foo.bar="hello,world!"'`, or double quotes (escaping the ones in the string): `"foo.bar=\"hello,world!\""`, or if you really don't like it, escaping the double quotes directly: `foo.bar=\"hello,world!\"` in this case since there's no space. I don't see the appeal of introducing a new way of doing the same thing which requires 7 more characters, and doesn't solve the issue of e.g. needing to add quotes anyway if you have spaces or other special characters.

> What will happen with @file:foo.html?

I would say this should behave like an import. So try to import it as Nickel by default. Another solution is to consider it text by default on the CLI, which is maybe the more common case, but then this might be surprising for users (there are two different default interpretations for files instead of one, depending on if they come from an import or a `@file`). My current preference would be to make things simple and uniform: `@file:foo.bar` is a shorthand for `'import "foo.bar"'. Once again, this is of course open to discussion.

> Will explicit formats be handled by alternative prefixes like @file_text:foo.html?

Yes, I think we want that. We can implement something like that first and then bikeshed the actual format specification, keeping in mind the injection possibility (that is: the default without an explicit format must NOT be a proper prefix of the version with an explicit format specified). Final syntax can be bikeshedded later, such as `@file/text:foo.html`.

> Maybe someday (was doing other, non-Nickel-related things meanwhile - my attention allotment ran dry for this one), but not this/next month probably.

Got it. I think the whole `@file` part is nice to have, but I wouldn't say it's urgent - we can already do the same with a tad more boilerplate today. However the environment variable part is harder to do properly, so that could be nice to have. In any case, we'll just take over this PR from here if needed. Thanks a lot for carrying it to this point already!

PR #2069, opened 2024-10-15
Fix completions in incomplete field name position.

jneem:
This fixes part of the problem in this [tf-ncl issue](https://github.com/tweag/tf-ncl/issues/49): when input failed to parse we would always try to treat it as a record *path*, meaning that the completions we'd return for

```
{ field. } | C
```

were the same as the ones we'd return for

```
{ blah = field. } | C
```

This PR fixes the problem by (when appropriate) completing the input as we would do for a record field.

There's still a performance issue, which I think is mostly caused by the massive size of our reply: we return the full text of the contract. We should probably detect when it's large and truncate.

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -70,8 +70,7 @@ fn extract_static_path(mut rt: RichTerm) -> (RichTerm, Vec<Ident>) {
     }
 }
 
-// Try to interpret `term` as a record path to offer completions for.
-fn sanitize_record_path_for_completion(
+fn parse_term_from_incomplete_input(
     term: &RichTerm,
     cursor: RawPos,
     world: &mut World,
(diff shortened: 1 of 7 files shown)


lsp/nls/src/requests/completion.rs
```diff
@@ -167,30 +173,61 @@ fn record_path_completion(term: RichTerm, world: &World) -> Vec<CompletionItem>
...
+// { bar = 1, foo = { blah = { ba } } }
+// ```
+//
+// so we don't encounter the path in this case. Instead, the non-empty path only comes
+// into play when the input fails to parse completely, like in (note the trailing dot)
+
+// ```
+// { bar = 1, foo.blah.ba. }
+//                        ^cursor
+// ```
```
yannham, line 205: Nitpick: we can use docstrings for private item (`///`), and I think the accepted style when using an empty line is to still prefix it with `///` (otherwise it's not clear that we're still documenting the same item)

lsp/nls/src/requests/completion.rs
```diff
@@ -70,8 +70,7 @@ fn extract_static_path(mut rt: RichTerm) -> (RichTerm, Vec<Ident>) {
     }
 }
 
-// Try to interpret `term` as a record path to offer completions for.
-fn sanitize_record_path_for_completion(
+fn parse_term_from_incomplete_input(
```
yannham, line 73: This function could deserve a small description.

PR #2070, opened 2024-10-15
Final import syntax for import with explicit format

yannham:
Change the syntax of import with explicit format specification from the temporary `import '<Format> <file>` to `import <file> as '<Format>`, as decided with the Nickel team.

Doing so, we also make the various input and export formats names consistent: we get rid of `raw`, and use `text` instead, which is more precise. For backward compatibility reason, we still support `raw` as an alias on the CLI.

The pretty-printer is also fixed: some spaces were previously missing around import with explicit format specification.

### DIFF
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -1321,11 +1321,12 @@ environment variable or similar command line option, with default being the
 current directory.
 
 One-argument import, like `import "myfile.ncl"`, uses filename extension to
-determine the file format. Nickel embeds a short list of known filename
-extensions: `ncl`, `json`, `yml`, `yaml`, `toml`, `txt` and `nix` with a
-fallback to a Nickel file if there is no extension or the extension is unknown.
+determine the file format. Nickel automatically recognizes the extensions
+`ncl`, `json`, `yml`, `yaml`, `toml` and `txt`. When compiled with experimental Nix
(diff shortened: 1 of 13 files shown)


doc/manual/syntax.md
```diff
@@ -1307,24 +1307,27 @@ record is serialized. This includes the output of the `nickel export` command:
...
 
-One-argument import, like `import "myfile.ncl"`, uses filename extension
-to determine the file format. Nickel embeds a short list of known filename
-extensions: `ncl`, `json`, `yml`, `yaml`, `toml`, `txt` and
-`nix` with a fallback to a Nickel file if there is no extetnsion
-or the extension is unknown.
+One-argument import, like `import "myfile.ncl"`, uses filename extension to
+determine the file format. Nickel embeds a short list of known filename
+extensions: `ncl`, `json`, `yml`, `yaml`, `toml`, `txt` and `nix` with a
+fallback to a Nickel file if there is no extension or the extension is unknown.
```
jneem, line 1326:
```suggestion
determine the file format. Nickel automatically recognizes the extensions
`ncl`, `json`, `yml`, `yaml`, `toml` and `txt`. When compiled with experimental Nix
support, it also recognizes `nix`. If the file's extension is not recognized, it
will default to Nickel format.
```

jneem (approved):

PR #2072, opened 2024-10-16
[RFC007] First step: AST representation

yannham:
# New AST representation for the (future) bytecode compiler

This is an implementation of the first step of RFC007: define the first AST representation, supposed to be output by the parser.

The whole RFC007 is a big chunk of work; we don't want to implement it all at once and make it the default. This PR creates a new `bytecode` module, which is only enabled under the `bytecode-experimental` feature, so that mainline Nickel is left unchanged, and we can experiment step by step on the side.

## Content

This PR is concerned with the AST design part. It defines a new immutable, arena-allocated AST, that has been cleaned from any runtime concern - this should be the AST as produced by the parser in the future.

Any subpart of `term::Term` that relies explicitely on the `RichTerm` representation has been copied, cleaned and adapted: mostly the patterns, record and array satellite datatypes.

Finally, this PR introduces a `bytecode::compat` module that converts from  the current mainline AST to the new AST, to make sure we haven't overlooked any part and that we have enough methods to build any ast node. This part isn't tested yet but this PR is already huge so we left this for future work.

## AST design guideline

**Important**: this PR doesn't set anything in stone. The size of the AST hasn't been checked or hardcore optimized yet, and we'll probably have to update the representation as we re-implement typechecking, etc. It's rather a first draft, trying to follow a systematic approach.

The AST is designed to be compact and adapted for processing by various analysis phases (mostly typechecking, code analysis by the LSP and compilation by the future bytecode compiler). Thus we've replaced any `Box`/`Rc` by plain immutable references, where the content has been allocated in a centralized arena (actually several ones of them).

For `struct`s, we have no reason to use references - for example, if `struct Foo` has a field `bar: Bar`, there is no good reason to add an indirection `bar: &'ast Bar`. Thus struct fields use owned data as much as possible.

This is the converse for `enum`s: to avoid size bloat, we add reference indirection for any variant where the arguments takes up more than a few words.

Because everything is immutable and shareable, and that we want to avoid heap allocation as much as possible for performance reason (arena allocation should be faster), we don't use `Vec<T>` but `&'ast [T]` instead, which is the immutable equivalent.

We've tried to reduce the variation of the same constructs: while the original AST has two `Let` and `LetPattern`, `Fun` and `FunPattern`, `Record` and `RecRecord`, and so on, this AST merges all those cases. Whilie we take a small size hit for the simplest cases (a `let x = y` now has one indirection, and store the size of a 1 element slice in a fat pointer), this makes the definition and the code consuming it arguably much simpler, and we won't pay this price at runtime since the AST will be compiled away.

Similarly, there's no difference anymore between `UnaryOp`, `BinaryOp`, and `NAryOp` and the corresponding `Op1`, `Op2`, `OpN`: there is only one `PrimOp` type, and one `PrimOpApp` node taking a slice of arguments. Function application is also made multi-ary, which is a more efficient representation for application to multiple arguments and can also help give better error messages during typechecking for over or under-application.

`Type` is the only satellite data that hasn't be replicated. It's a problem because it includes a `Contract` constructor that still refer to the old representation (a `RichTerm`). We can also wonder if we we'd like to arena-allocate the `Type` AST as well, but this is non trivial work and we left it for a follow-up PR.

## Reviewing

The diff is really big, but keep in mind that a lot of code needed to be copy pasted and almost mechanically adapted. I've also added  even more documentation (such as the arguments on primops).

The most interesting is probably type definitions: `Node`, `Ast`, `Pattern`, `PrimOp`, etc. The rest is mostly type-guided implementation.


core/Cargo.toml
```diff
@@ -19,14 +19,15 @@ links = "nix"
 bench = false
 
 [features]
-default = ["markdown", "repl", "doc", "format", "spanned-deser"]
+default = ["markdown", "repl", "doc", "format", "spanned-deser", "bytecode-experimental"]
```
yannham, line 22: TODO: temporary change to make rust-analyzer work by default. Change that before merging.

aspiwack:
For my understanding, do you actually intent the flat AST to be an output of the parser?

jneem:
If I understood this correctly, there's no flat AST: just a tree AST and a bytecode format. This PR is the simplified tree AST produced by the parser. It's simplified compared to current nickel because it doesn't have to support evaluation.

aspiwack:
Ah. That's the point I'd missed. Thanks @jneem .

core/src/bytecode/ast/pattern/bindings.rs
```diff
@@ -0,0 +1,181 @@
...
+
+impl<'ast> InjectBindings<'ast> for EnumPattern<'ast> {
+    fn inject_bindings(
+        &self,
+        bindings: &mut Vec<(Vec<LocIdent>, LocIdent, Field<'ast>)>,
+        path: Vec<LocIdent>,
+        _parent_extra: Option<&Field<'ast>>,
+    ) {
+        //TODO: I'm not sure we should just transparently forward to the variant's argument. Maybe
+        //we need a more complex notion of path here, that knows when we enter an enum variant?
```
jneem, line 163: I think it isn't urgent. The two users of `bindings` are let pattern blocks (which ignores the path) and nls. Nls will ignore a path that passes through an enum.

core/src/bytecode/ast/pattern/mod.rs
```diff
@@ -0,0 +1,234 @@
+//! Pattern matching and destructuring of Nickel values.
+use std::collections::{hash_map::Entry, HashMap};
+
+use super::{Annotation, Ast, Number};
+
+use crate::{identifier::LocIdent, parser::error::ParseError, position::TermPos};
+
+pub mod bindings;
+
+/// A small helper to generate a
```
jneem, line 10: to generate a?

jneem (approved):
I like the new simple `Term` -- it almost fits within one editor window!

core/src/bytecode/ast/compat.rs
```diff
@@ -0,0 +1,541 @@
...
+
+/// Convert from the mainline Nickel representation to the new AST representation. This trait is
+/// mostly `From` with an additional argument for the allocator.
+///
+/// # Parameters
+///
+/// - `'ast`: the lifetime of the AST nodes, tied to the allocator
+/// - `'a`: the lifetime of the reference to the mainline Nickel object, which doesn't need to be
+///   related to `'ast` (we will copy any required data into the allocator)
+/// - `T`: the type of the mainline Nickel object ([term::Term], [term::pattern::Pattern], etc.)
```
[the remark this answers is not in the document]
yannham: Hmm, it's true that it sounds like `'a` should be a "free" lifetime. I think it's useful to keep both direction of the conversion around just in case; maybe we'll want to test both mainline and the bytecode vm for full runs and compare the result of each phase for example, I don't know. We'll get rid of both of them eventually anyway. So let me try to move the lifetime to `from_mainline`.

core/src/bytecode/ast/pattern/bindings.rs
yannham: Yeah, this comment was probably just copy pasted from `term::pattern`.

core/src/bytecode/ast/compat.rs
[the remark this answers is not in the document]
yannham: Since it becomes the only lifetime parametrized at the level of the function (`'ast` is coming from `impl<'ast>`), we can just elide it entirely, which is even cleaner. Thanks!

core/src/bytecode/ast/pattern/mod.rs
yannham: Seems to be an old leftover, which was already orphan in `term::pattern`. Got rid of it in both places.

PR #2073, opened 2024-10-21
Bound the length of nls completions

jneem:
When there's a big contract, the completions can be very large. This can slow down nls substantially, with most of the time being spent in pretty-printing. The current `PrettyPrintCap` mechanism isn't useful here, because it does all the slow pretty-printing first and then truncates the output.

This PR adds a crude mechanism to bound the size of pretty prints. With this change, completion in the [tf-ncl](https://github.com/tweag/tf-ncl/issues/49) example is instant. There are probably other places that truncated pretty-printing could be used in nls.

I'm not completely happy with how this implementation threads the bounds through, but I couldn't figure out how else to do it:
- I could use a wrapper type like `BoundedTerm<T: Pretty>(T, SizeBounds)` to wrap the things  that are actually getting pretty-printed, but it requires lots of changes to the pretty-printer to explicitly pass that state around everywhere
- I tried keeping the state in the allocator and recursing with smaller "child" allocators, but the lifetimes in `DocAllocator` don't allow this: since the lifetime of the allocator leaks into the returned `DocBuilder`, you can't format children using a local allocator.

The current approach with interior mutability was from me trying to work around the second issue.

I also wonder if `NickelAllocatorExt` needs to be a trait. Maybe we can just use `BoundedAllocator` (possibly renamed) concretely everywhere?


core/src/pretty.rs
```diff
@@ -152,17 +153,197 @@ fn needs_parens_in_type_pos(typ: &Type) -> bool {
     }
 }
 
-pub fn fmt_pretty<'a, T>(value: &T, f: &mut fmt::Formatter) -> fmt::Result
+pub fn fmt_pretty<T>(value: &T, f: &mut fmt::Formatter) -> fmt::Result
 where
-    T: Pretty<'a, pretty::BoxAllocator, ()> + Clone,
+    T: for<'a> Pretty<'a, BoundedAllocator, ()> + Clone,
```
yannham, line 158: Out of curiosity, what is the reason to move the `forall 'a` into `T` instead of `fmt_pretty`?

core/src/pretty.rs
```diff
@@ -152,17 +153,197 @@ fn needs_parens_in_type_pos(typ: &Type) -> bool {
...
+            bound: Some(Cell::new(SizeBound {
+                depth: max_depth,
+                size: max_size,
+            })),
+        }
+    }
+
+    /// Shrinks this allocator, returning a guard that restores the
+    /// old constraints when it leaves scope.
+    fn shrink(&self, child_size: usize) -> ShrinkGuard<'_> {
```
yannham, line 227: I think the RAII pattern is cool, but I'm concerned that the `shrink_guard` is never used by anything. Can we thus guarantee that the compiler won't drop it before the end of the scope? Or even will just get rid of it entirely and never create the object as technically dead code?

I'm not really sure how to look at that beside looking at the generated ASM (but even that isn't a guarantee). Another approach could be to do the same as for a mutex: we could borrow `self`, even mutably - to make sure we don't use the original while a shrinked version is used - and then expose `self` as an immutable reference inside `ShrinkGuard`, so all usages of `shrinked` self should go through the guard. Exactly as a mutex guard, in the end.

core/src/pretty.rs
```diff
@@ -152,17 +153,197 @@ fn needs_parens_in_type_pos(typ: &Type) -> bool {
...
+        if let Some(bound) = self.bound.as_ref() {
+            let old = bound.get();
+            bound.set(SizeBound {
+                depth: old.depth - 1,
```
yannham, line 231: Should we defend over underflow ? (nothing prevents from instantiating with `depth=0` and then shrinking)

```suggestion
                depth: old.depth.saturating_sub(1),
```

yannham (commented):
I agree that I don't see a much better way to do it, without butchering the actual pretty-printing code. I also think that the `NickelAllocatorExt` becomes useless - it's was done only because we wanted to monkey-patch the external type `BoxAllocator`, but now we can just implement the methods directly on our own type.

core/src/pretty.rs
jneem: I had this at some point but was worried about the asymmetry between decrementing and then incrementing it back again. I think the underflowing one is actually ok, since it will panic in tests and then behave slightly strangely but deterministically (possibly missing some truncation) in release mode.

yannham: > I had this at some point but was worried about the asymmetry between decrementing and then incrementing it back again.

But the code is only decrementing it, isn't it? How would this "asymmetry between decrementing and then incrementing it" would manifest in practice, in terms of concrete effect? This isn't really important because we're talking about a parameter that controls pretty printing, so worst case we print too much, but in term of semantics, I find the difference in behavior between debug and release - the latter having a completely unexpected behavior - to be unsatisfactory.

core/src/pretty.rs
yannham: Good to know that there are some guarantees. However, I still find it really strange to create a guard that we never use and somehow to still rely on its scope, as a user of that API.

If we just want scoping, what about spawning a closure instead?

Like

```rust
allocator.shrinked(|| {
  do_stuff with shrinked self
)
```

We can pass `&self` as an argument as well, or not, it doesn't really matter. At least it makes it clear where the scope is delimited and that using any immutable reference to `self` doesn't make any difference (whether it comes from the external or not).

jneem: Ah, the closure API seems nicer! I'll give that a try

core/src/pretty.rs
jneem: The incrementing happens when dropping the guard. But I think this can be solved by storing the old value and then restoring it, like I'm doing with `size`.

yannham: I see, I thought both were saved and restored. In any case neither wrapping nor incrementing after a saturating sub are correct, but I would still err on the side of wrapping is more incorrect than anything else :stuck_out_tongue:

jneem:
Ok, the new version gets rid of the `NickelAllocatorExt` trait and a bunch of generics. It also uses a closure for scoping and avoids underflows.

I couldn't figure out the purpose of the `A` parameter. Even within `pretty`, it just seems to get passed around without doing anything. So I got rid of it also.

yannham (approved):

PR #2074, opened 2024-10-21
Switch to toml-edit for spanned deserialization

jneem:
This little experiment shows that we can use toml-edit for spanned parsing instead of the toml crate. It doesn't actually add any dependencies, since `toml` uses `toml_edit` for parsing anyway.


core/Cargo.toml
```diff
@@ -27,7 +27,6 @@ doc = ["dep:comrak"]
 format = ["dep:topiary-core", "dep:topiary-queries", "dep:tree-sitter-nickel"]
 metrics = ["dep:metrics"]
 nix-experimental = [ "dep:cxx", "dep:cxx-build", "dep:pkg-config" ]
-spanned-deser = ["dep:serde-untagged"]
```
yannham, line 30: Ah, nice, we can even get rid of the feature.

yannham (approved):
Purely stylistic nitpick: when I start to have 3 or more (xxx_to_yyy), I now don't hesitate to declare a local, private trait (it's a bit more line but it feels slightly less icky than having to call different combinations of methods based on the type). It's just a comment, it's also fine as it is.

Otherwise, I think a bit more doc would be nice. Not necessarily on conversion methods that are self-expalanatory, but at least at the module level: if I'm a new contributor, I wouldn't necessarily understand immediately why we would do toml deserialization differently and using this lower level API while we have the serde machinery.

jneem:
> Purely stylistic nitpick: when I start to have 3 or more (xxx_to_yyy), I now don't hesitate to declare a local, private trait

Yes, that turned out nicely (and also removed the asymmetry between _to_term and _to_rich_term)

PR #2077, opened 2024-10-23
Strict typechecking mode

### LINKED ISSUE #2049: Opt-in static typing globally
My team is considering using Nickel for our config language, but we *really* like static typing, and want to make sure it is enforced when people outside of our immediate team try to change stuff. We would like a way to opt-in to enforce static typing globally, so for example we can have CI type-check the config and "you shall not pass" people if it doesn't check out.

A simple solution is to just have a CLI flag to opt-in

Another idea that has been floated is to have a per-file pragma, but this might allow people to comment that out in a moment of frustration and still be able to merge.

**Additional context**
This idea was originally discussed on the Discord: https://discord.com/channels/1174731094726295632/1179430745727586407/1285771238618435605 (and following replies)

jneem:
This adds a `--strict-typechecking` flag that starts the typechecker in "checking" mode, equivalent to wrapping each file in `(...) : _`.

A couple of points I was unsure of:
- this PR does it for every file, but maybe we only want it at the top level?
- I noticed while testing that despite the "strict" in the name, it's sometimes more permissive than non-strict mode (because of type inference; see the added tests). Maybe "strict" mode should run both kinds of typechecking so it's actually strictly stricter?

Fixes #2049


yannham:
> I think it's fine as it is; it's the expected semantics (as you say, it behaves is if we put : _ every where at the top of each file), and while it's more permissive, it's not accepting dubious programs. I would say that it's smarter rather than just more permissive.

Ok but now maybe I see your point: it might be surprising that some files require `--strict-typechecking` to work. On the other hand it would be sad and wasteful to just run typechecking a second time in walk mode. I'm tempted to say that it's on the user. Maybe providing a way to do the same thing with a pragma located in the file instead would be more resilient to this kind of issues, as now the decision is taken by the source program and not the invoker of the CLI?

jneem:
I was mainly thinking from the point of view of someone wanting to run extra checks in CI. There, you don't really care if `nickel typecheck` is a little (or even a factor of 2) slower, but you'd be surprised if it misses out on some problems that are caught by a "normal" typecheck. I'm not sure that having pragmas in the file really solves this use case, though: I'm imagining that I have a normal nickel file that isn't supposed to need any special settings, and I just want to run some extra-strict checks on it.

yannham:
Hmm, indeed.

By the way, should we make `--strict-typechecking` an argument of `nickel typecheck` only? Maybe that would make it clear that this is some kind of separate check, and then it would be acceptable to run typechecking in both mode - we don't penalize normal execution, which is what matters probably. In that case I tend to agree that maybe the double run is what you want.

And if it starts to be too slow, we can easily make it so that the typechecker just ignores any expression with a type or a contract annotation on the second run (some kind of `shallow walk`,so to speak), to make sure it doesn't duplicate any work from the strict pass.

Another solution would be to rerun the typechecker in strict mode upon failure in normal execution: if the program is typeable with a bit more effort, then we accept it. But I'm not sure it's such a good idea: if people start to rely on this behavior, then the default path would waste work in the first attempt at each execution.

yannham:
> And if it starts to be too slow, we can easily make it so that the typechecker just ignores any expression with a type or a contract annotation on the second run (some kind of shallow walk,so to speak), to make sure it doesn't duplicate any work from the strict pass.

Wait, we can't do that, because that won't catch your example. Maybe the converse is fine though: first run the normal pass, and for the strict pass, only reprocess what was untyped before? Since inference is more powerful, it seems like if a typed block passes the normal phase, then it would necessarily pass the strict one. Would have to think a bit harder to make sure.

cli/src/typecheck.rs
```diff
@@ -9,11 +11,25 @@ use crate::{
...
+    /// Start the type-checker in strict mode, so that the entire input is treated as a typed block.
+    #[arg(long, global = true)]
+    pub strict_typechecking: bool,
 }
 
 impl TypecheckCommand {
     pub fn run(self, global: GlobalOptions) -> CliResult<()> {
+        if self.strict_typechecking {
+            // In strict mode we run *both* forms of typechecking, because in fact neither one
+            // is more strict than the other.
```
yannham, line 24: Nitpick: showing an example here (especially given that we have a very small one) could help a future reader here, otherwise this sentence could look a tad mystic.

PR #2079, opened 2024-10-24
[RFC007] New AST for types

yannham:
Depends on #2072

The previous PR #2072 introducing the new AST used the "old" type representation from mainline Nickel. Now that `typ::Type` has been made more generic in #2078, we can use a better representation that can be arena allocated as well and can contain contracts that are themselves represented in the new AST.

This PR integrates this new type representation and adds corresponding translation functions from the mainline implementation.

### DIFF
--- a/core/src/bytecode/ast/typ.rs
+++ b/core/src/bytecode/ast/typ.rs
@@ -1,0 +1,33 @@
+//! Representation of Nickel types in the AST.
+
+use super::{Ast, TermPos};
+use crate::typ as mainline_typ;
+pub use mainline_typ::{EnumRowF, EnumRowsF, RecordRowF, RecordRowsF, TypeF};
+
+/// The recursive unrolling of a type, that is when we "peel off" the top-level layer to find the actual
+/// structure represented by an instantiation of `TypeF`.
+pub type TypeUnr<'ast> = TypeF<&'ast Type<'ast>, RecordRows<'ast>, EnumRows<'ast>, Ast<'ast>>;
+
+/// The recursive unrolling of a enum rows.
+pub type EnumRowsUnr<'ast> = EnumRowsF<&'ast Type<'ast>, &'ast EnumRows<'ast>>;
+
+/// The recursive unrolling of record rows.
+pub type RecordRowsUnr<'ast> = RecordRowsF<&'ast Type<'ast>, &'ast RecordRows<'ast>>;
+
+/// Concrete, recursive definition for an enum row.
+pub type EnumRow<'ast> = EnumRowF<&'ast Type<'ast>>;
+/// Concrete, recursive definition for enum rows.
+#[derive(Clone, PartialEq, Debug)]
+pub struct EnumRows<'ast>(pub EnumRowsUnr<'ast>);
+/// Concrete, recursive definition for a record row.
+pub type RecordRow<'ast> = RecordRowF<&'ast Type<'ast>>;
+#[derive(Clone, PartialEq, Debug)]
+/// Concrete, recursive definition for record rows.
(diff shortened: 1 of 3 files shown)


core/src/bytecode/ast/typ.rs
```diff
@@ -0,0 +1,33 @@
...
+/// Concrete, recursive definition for a record row.
+pub type RecordRow<'ast> = RecordRowF<&'ast Type<'ast>>;
+#[derive(Clone, PartialEq, Debug)]
+/// Concrete, recursive definition for record rows.
+pub struct RecordRows<'ast>(pub RecordRowsUnr<'ast>);
+
+/// Concrete, recursive type for a Nickel type.
+#[derive(Clone, PartialEq, Debug)]
+pub struct Type<'ast> {
+    pub typ: TypeUnr<'ast>,
```
jneem, line 31: In `Ast` you have a reference to the enum, and then that enum takes `Ast`s by value. Here you take the enum by value, and then that enum has references to `Type`s. Is there a reason for the difference?

jneem (approved):

core/src/bytecode/ast/typ.rs
yannham: This is a good question. I think my initial scheme is (as documented in `ast/mod.rs`) to not use references inside structs (they better be packed), and use references inside enums to avoid the size bloat. So in that respect, `Type` seems to follow the original approach while `Ast` is the outlier.

I think I did things like that for `Ast` because:

1. This is how `RichTerm` works currently (it's an `Rc`, and the position isn't behind a reference, but it's still more or a less a "shared fat pointers" that can be cloned relatively cheaply). So kept the idea of cheaply clonable base type for terms (`Node` is the not-shared version, akin to `Term`).
2. Not a very good argument, but there's a lot of references to `Ast` in `Node` and they should all be references in some way. So it felt a bit better to just internalize this reference in `Ast` and then just use `Ast<'ast>` instead of `&'ast Ast<'ast>` everywhere. But this can be done by just defining `type SharedAst<'ast> = &'ast Ast<'ast>` or sth anyway.
3. Not fundamental either, but the position is filled once and then only ever accessed upon errors (more or less). In the happy path it seems like it's fine to not have it in cache and make the type a bit smaller by putting it behind a reference (maybe we can even benchmark if it's better to put them in a separate arena, but this is a too subtle optimization for the first version). Once again, not a definitive argument because we could very well but `pos` behind a reference but not `Node`.

So we have two solutions: remove the references from inside `Ast`, to match the current type and general idea that struct are better packed, or say that the base entry type for a whole category of nodes (`Ast` for expressions, `Type` for types, etc.) should be copy or at least cheap to clone and thus align `Type` with this idea by adding references inside.

What do you think?

jneem: I don't have a strong opinion about which is better; I was mostly just curious about the discrepancy. I guess if we don't expect `pos` to get much use then maybe it's better to have `TermPos` inline inside `Ast` and then pass around thin `&Ast<'ast>` pointers, as opposed to having `Ast` contain a reference and pass around fat `Ast<'ast>` "pointers." But I think compared to the arena allocation and the slimmer enums, it's a lower order thing anyway.

yannham: You still pay the price of `TermPos` in other places, typically for `[Ast<'ast>]` (where you probably don't want to store indirect references for cache locality). But we could even have a double indirection still (use thin `&Ast<'ast>` pointers _and_ have `pos: &'ast TermPos` inside `Ast`). But I think it kinda makes sense to stick to the original approach which is simple and uniform, and thus inline stuff within `Ast` and pass references around (even if we abstract it as `AstRef`, it has the advantage of making explicit the sharing points). I'll do that in a separate PR, though.

PR #2080, opened 2024-10-24
[RFC007] Conversion from new AST to mainline AST

yannham:
Depends on #2079.

Add a the inverse transformation to `bytecode::compat` (to go from the new AST to the mainline AST), following our new plan of replacing the left part of the pipeline (parsing, typechecking ,etc.) step by step to use the new AST including in mainline Nickel, and just translate to the old one for the passes that aren't yet upgraded, provided that we measure that the AST conversion is negligible.

### DIFF
--- a/core/src/bytecode/ast/compat.rs
+++ b/core/src/bytecode/ast/compat.rs
@@ -7,7 +7,7 @@
 use super::{primop::PrimOp, *};
 use crate::{
     label,
-    term::{self, pattern as mline_pat},
+    term::{self, pattern},
     typ as mline_type,
 };
 use smallvec::SmallVec;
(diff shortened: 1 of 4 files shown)


core/src/bytecode/ast/compat.rs
```diff
@@ -605,3 +601,599 @@ impl From<&term::NAryOp> for PrimOp {
...
+
+                let rterm = args.iter().fold(fun.to_mainline(), |result, arg| {
+                    // This case is a bit annoying: we need to extract the position of the sub
+                    // application to satisfy the old AST structure, but this information isn't
+                    // available directly.
+                    // What we do here is to fuse the span of the term being built and the one of
+                    // the current argument, which should be a reasonable approximation (if not
+                    // exactly the same thing).
+                    // unwrap(): the position of Ast should always be set (we might move to `RawSpan`
+                    // instead of `TermPos` soon)
```
jneem, line 1102: :+1:

jneem (approved):

PR #2081, opened 2024-10-28
Add basic runtime metrics for different phases

yannham:
The goal is to have some coarse metrics to decide if switching the parser to the new AST and then converting back to the old one (to replace the current pipeline step by step) is going to have any noticeable performance hit. And also because it's quite useful in general to see which phases are the most costly. 

As a first experiment, on the private benchmark of 24kLoC of Nickel (plus 10-15kLoc of JSON or YAML), it's already interesting to see that the evaluation cost trumps pretty much everything else measured (type checking, stdlib prep, transformation, etc.). Preparing (parse + import resolution + transformation + type checking) is < 2% of the eval time (not even the total time, which is probably also impacted by serialization, substitution and pretty printing). Preparing the stdlib is < 0.1%, typechecking the main file is < 0.5%, and parsing is in the same order of magnitude as those.

On a small example (GCC example), almost all operations are negligible anyway (a handful of ms for each stage). So there's a good chance that the conversion from the new ast to the old ast will be reasonable in both case (in small cases it should be fast, in large examples it should be trumped by evaluation). But we'll make a more thorough benchmarking anyway.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -4,6 +4,7 @@ use crate::closurize::Closurize as _;
 use crate::error::{Error, ImportError, ParseError, ParseErrors, TypecheckError};
 use crate::eval::cache::Cache as EvalCache;
 use crate::eval::Closure;
+use crate::metrics::measure_runtime;
 #[cfg(feature = "nix-experimental")]
 use crate::nix_ffi;
 use crate::parser::{lexer::Lexer, ErrorTolerantParser};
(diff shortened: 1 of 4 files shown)


jneem (approved):
It may be worth having a benchmark with a lot of generated contracts but a relatively small exported part that only uses a few of them. I think this would be reasonably common in, e.g. tf-ncl and it might have a different relationship between parse time and eval time.

PR #2082, opened 2024-10-28
[RFC007] Fix the representation of Ast: remove inner indirection 

yannham:
Depends on #2080.

`bytecode::ast::Ast` wasn't following the general approach that we set for designing other datastructures of the ast: use references within enums to avoid size bloat, but use owned data in structures to avoid indirection (since structures are packed).

Ast was doing something different, pushing the reference inside the individual fields to make a relatively thin pointer around `Node` and store that datastructure inline in the `Node` enum. While this essentially doesn't make a huge difference, we rather follow our simple general rule. Additionally, other data structures such as `bytecode::ast:typ::Type` follows the simple general rule, and we would rather do the same thing everywhere in a consistent manner.

Doing so, this commit also cleans a bit the interface of `AstAlloc` to reflect the ownership change and to get rid of some variations of methods in the sake of simplicity.

### DIFF
--- a/core/src/bytecode/ast/mod.rs
+++ b/core/src/bytecode/ast/mod.rs
@@ -250,7 +250,7 @@ impl AstAlloc {
         self.generic_arena.alloc(node)
     }
 
-    pub fn number<'ast>(&'ast self, number: Number) -> Node<'ast> {
+    pub fn number(&self, number: Number) -> Node<'_> {
         Node::Number(self.number_arena.alloc(number))
     }
 
--- a/core/src/bytecode/ast/mod.rs
+++ b/core/src/bytecode/ast/mod.rs
@@ -33,13 +33,12 @@ use typ::*;
 
 /// A Nickel AST. Contains a root node and a span.
 ///
-/// Both members are references so that `Ast` is cheap to clone - it's in fact `Copy` - and to
-/// store. Additionally, we don't expect to access the span much on the happy path, so the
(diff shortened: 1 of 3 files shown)


core/src/bytecode/ast/mod.rs
```diff
@@ -33,13 +33,12 @@ use typ::*;
 
 /// A Nickel AST. Contains a root node and a span.
 ///
-/// Both members are references so that `Ast` is cheap to clone - it's in fact `Copy` - and to
-/// store. Additionally, we don't expect to access the span much on the happy path, so the
-/// indirection should be ok.
-#[derive(Clone, Copy, Debug, PartialEq)]
+//TODO: we don't expect to access the span much on the happy path. Should we add an indirection
+//through a reference?
```
jneem, line 37: Given that we mostly work with `&Ast` instead of `Ast`, I suspect that having `TermPos` inline doesn't hurt.

jneem (approved):

core/src/bytecode/ast/mod.rs
yannham: Yeah, probably. Though it does take up space, for example in `[Ast]`. I mean, the positions have to leave somewhere anyway, but  more generally question is: if we don't need them in CPU cache, should we put them somewhere else? It could be in a dedicated, separate arena for example, so that reading and traversing ASTs don't need to bring them into cache.

PR #2084, opened 2024-10-31
Make files copy-on-write

jneem:
This introduces our own `Files` database, and uses it instead of the one in `codespan`. The main point of this change is that our `Files` is cheaply clonable and copy-on-write (using `Vector` under the hood). Also, the loading of the stdlib text has been moved from `Cache` to `Files`. Together, this means that errors needing files for reporting can take a cloned `Files` instead of a `Program`. (I haven't actually made this change to the errors yet.)

There are a lot of changes caused by the fact that we're moving from `codespan::{Files, FileId}` to `nickel_lang_core::files::{Files, FileId}` everywhere, but most of the interesting changes are to `core/src/cache.rs`.

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -3,9 +3,10 @@
 use nickel_lang_core::{
     error::{
         report::{ColorOpt, ErrorFormat},
-        Diagnostic, FileId, Files, IntoDiagnostics, ParseError,
+        Diagnostic, IntoDiagnostics, ParseError,
     },
     eval::cache::lazy::CBNCache,
+    files::{FileId, Files},
     program::{FieldOverride, FieldPath, Program},
 };
 
@@ -62,12 +63,8 @@ pub enum Error {
     FailedTests,
 }
 
-impl IntoDiagnostics<FileId> for CliUsageError {
-    fn into_diagnostics(
-        self,
-        files: &mut Files<String>,
-        stdlib_ids: Option<&Vec<FileId>>,
-    ) -> Vec<Diagnostic<FileId>> {
+impl IntoDiagnostics for CliUsageError {
+    fn into_diagnostics(self, files: &mut Files) -> Vec<Diagnostic<FileId>> {
         fn mk_unknown_diags<FileId>(
             data: UnknownFieldData,
             method: &str,
@@ -120,7 +117,7 @@ impl IntoDiagnostics<FileId> for CliUsageError {
                     ])]
             }
             CliUsageError::AssignmentParseError { error } => {
-                let mut diags = IntoDiagnostics::into_diagnostics(error, files, stdlib_ids);
+                let mut diags = IntoDiagnostics::into_diagnostics(error, files);
                 diags.push(
                     Diagnostic::note()
                         .with_message("when parsing a field assignment on the command line")
@@ -135,7 +132,7 @@ impl IntoDiagnostics<FileId> for CliUsageError {
                 diags
             }
             CliUsageError::FieldPathParseError { error } => {
-                let mut diags = IntoDiagnostics::into_diagnostics(error, files, stdlib_ids);
+                let mut diags = IntoDiagnostics::into_diagnostics(error, files);
                 diags.push(
                     Diagnostic::note()
                         .with_message("when parsing a field path on the command line")
@@ -164,12 +161,8 @@ pub enum Warning {
     EmptyQueryPath,
 }
 
-impl<FileId> IntoDiagnostics<FileId> for Warning {
-    fn into_diagnostics(
-        self,
-        _files: &mut Files<String>,
-        _stdlib_ids: Option<&Vec<FileId>>,
-    ) -> Vec<Diagnostic<FileId>> {
+impl IntoDiagnostics for Warning {
+    fn into_diagnostics(self, _files: &mut Files) -> Vec<Diagnostic<FileId>> {
         vec![Diagnostic::warning()
             .with_message("empty query path")
             .with_notes(vec![
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -1126,13 +1126,6 @@ impl Cache {
         &self.files
     }
 
-    /// Get a mutable reference to the underlying files. Required by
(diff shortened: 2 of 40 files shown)


yannham:
I haven't took a deep look yet, but the copy-on-write part made it scary at first: it would mean that we might have to copy the whole file database upon addition of a temporary snippet, which sounds costly. But in fact it's not copy-on-write, it's using persistent vectors, which is different. Is that correct?

jneem:
It is behind a persistent vector, yes, but the data in a vector can still be cloned when the vector is modified so I still think of it as "copy-on-write." The file data is in an `Rc<str>`, though, so these clones are cheap.

yannham:
I see. To me copy-on-write is the good old one, implemented with `Cow`, where you would copy the whole structure upon write. I think persistent vectors are sharing much more data, but I can see where you come from. Anyway, if the files are behind `Rc<str>`, we're fine either way.

cli/src/error.rs
```diff
@@ -135,7 +132,7 @@ impl IntoDiagnostics<FileId> for CliUsageError {
                 diags
             }
             CliUsageError::FieldPathParseError { error } => {
-                let mut diags = IntoDiagnostics::into_diagnostics(error, files, stdlib_ids);
+                let mut diags = IntoDiagnostics::into_diagnostics(error, files);
```
yannham, line 135: It's nice not to have to juggle around with `stdlib_ids` :+1:

core/src/cache.rs
```diff
@@ -1124,13 +1122,14 @@ impl Cache {
...
     /// the WASM REPL error reporting code and LSP functions.
-    pub fn files(&self) -> &Files<String> {
+    pub fn files(&self) -> &Files {
         &self.files
     }
 
     /// Get a mutable reference to the underlying files. Required by
     /// [crate::error::IntoDiagnostics::into_diagnostics].
-    pub fn files_mut(&mut self) -> &mut Files<String> {
+    /// TODO: maybe we can delete this now?
```
yannham, line 1131: One way to know is to delete it and see if that breaks :stuck_out_tongue: I'm in favor of deleting it if it's not useful anymore. It's a public item and honestly it's a bit ugly.

core/src/files.rs
```diff
@@ -0,0 +1,212 @@
+use std::{
```
yannham, line 1: Missing module documentation.

core/src/files.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Files {
+    files: Vector<File, 8>,
+    first_non_stdlib: usize,
+}
+
+impl Files {
```
yannham, line 62: Same, I think those methods could get documented, even with a one liner.

core/src/files.rs
```diff
@@ -0,0 +1,212 @@
...
+use codespan::ByteIndex;
+use codespan_reporting::files::Error;
+use nickel_lang_vector::Vector;
+
+use crate::{position::RawSpan, stdlib::StdlibModule};
+
+#[derive(
+    Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, serde::Serialize, serde::Deserialize,
+)]
+pub struct FileId(u32);
```
yannham, line 16: I think the original implementation used `NonZeroU32` (I don't remember the exact syntax) to enable some niche optimization (like `Option<FileId>` taking the same space as `FileId`). It probably doesn't matter on x64, and maybe we don't care at all; just FYI.

core/src/files.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[derive(Debug, Clone)]
+pub struct File {
+    /// The name of the file.
+    name: OsString,
+    /// The source code of the file.
+    source: Rc<str>,
+    /// The starting byte indices in the source code.
+    line_starts: Rc<[ByteIndex]>,
```
yannham, line 25: I'm not sure what does `line_starts` represent, even with the comment.

core/src/repl/wasm_frontend.rs
```diff
@@ -194,8 +194,7 @@ impl WasmInputResult {
     fn error(cache: &mut Cache, error: InputError) -> Self {
         let (msg, errors) = match error {
             InputError::NickelError(err) => {
-                let stdlib_ids = cache.get_all_stdlib_modules_file_id();
-                let diagnostics = err.into_diagnostics(cache.files_mut(), stdlib_ids.as_ref());
+                let diagnostics = err.into_diagnostics(cache.files_mut());
```
yannham, line 197: Do we still  need `files_mut()`?

vector/src/slice.rs
```diff
@@ -114,6 +114,20 @@ where
         self.start.checked_add(idx).and_then(|i| self.vec.get(i))
     }
 
+    /// Gets an element at a given index, panicking if `idx` is out-of-bounds.
```
yannham, line 117: The documentation is out of sync with the actual method (get vs set)

```suggestion
    /// Sets an element at a given index, panicking if `idx` is out-of-bounds.
```

yannham (approved):
While the general approach is a bit convoluted, I guess we can't do much better. It's also nice that it simplifies many APIs across the codebase by dropping those `stdlib_ids` and `<FileId>`.

core/src/files.rs
jneem: Right, I noticed that but ignored it because I think we don't store `Option<FileId>` anywhere (we have `TermPos::None` instead)

core/src/cache.rs
jneem: Ah, I was hoping to notice the "TODO" before making this pr :sweat_smile: I guess I'll need to find a more unique string...

And yes, we can delete it and clean up a few things.

core/src/repl/wasm_frontend.rs
jneem: Nope!

core/src/files.rs
jneem: Hopefully the changed one is clearer:
```
    /// The byte index of the start of each line. The first element of this array is always 0.
```

PR #2085, opened 2024-10-31
[RFC007] Add a builder module for the new AST

yannham:
This PR copies `term::make::builder` to `bytecode::ast::builder` and change it to make it support building fields and records for the new AST.

This PR has been split as an independent and reasonably-sized change from #2083.

### DIFF
--- a/core/src/bytecode/ast/builder.rs
+++ b/core/src/bytecode/ast/builder.rs
@@ -48,7 +48,7 @@ pub struct Complete<'ast>(Option<Ast<'ast>>);
 /// A Nickel record field being constructed
 #[derive(Debug)]
 pub struct Field<'ast, State> {
-    /// This field store different piece of information depending on the state of this field:
+    /// This field stores different piece of information depending on the state of this field:
     ///
     /// - It's empty for an incomplete field (`()`).
     /// - When a value (or `None`) has ben set, `State` becomes [Complete] and `state` store the
--- a/core/src/bytecode/ast/builder.rs
+++ b/core/src/bytecode/ast/builder.rs
@@ -51,7 +51,7 @@ pub struct Field<'ast, State> {
     /// This field stores different piece of information depending on the state of this field:
     ///
     /// - It's empty for an incomplete field (`()`).
-    /// - When a value (or `None`) has ben set, `State` becomes [Complete] and `state` store the
+    /// - When a value (or `None`) has been set, `State` becomes [Complete] and stores the
     ///   optional value of this field.
     /// - Finally, when the field has been attached to a record, `state` store the parent record.
     state: State,
--- a/core/src/bytecode/ast/builder.rs
+++ b/core/src/bytecode/ast/builder.rs
@@ -11,12 +11,12 @@
 //!     .field("foo")
 //!     .priority(MergePriority::Bottom)
 //!     .doc("foo?")
-//!     .not_exported()
-//!     .value(alloc.string("foo"));
+//!     .not_exported(true)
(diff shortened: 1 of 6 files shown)


core/src/bytecode/ast/builder.rs
```diff
@@ -0,0 +1,716 @@
...
+/// Typestate style tag for `Field`s that are not yet completely specified
+pub struct Incomplete();
+
+/// Typestate style tag for `Field`s that have been finalized
+pub struct Complete<'ast>(Option<Ast<'ast>>);
+
+/// A Nickel record field being constructed
+#[derive(Debug)]
+pub struct Field<'ast, State> {
+    /// This field store different piece of information depending on the state of this field:
```
yannham, line 51:
```suggestion
    /// This field stores different piece of information depending on the state of this field:
```

core/src/bytecode/ast/builder.rs
```diff
@@ -0,0 +1,716 @@
...
+/// Typestate style tag for `Field`s that have been finalized
+pub struct Complete<'ast>(Option<Ast<'ast>>);
+
+/// A Nickel record field being constructed
+#[derive(Debug)]
+pub struct Field<'ast, State> {
+    /// This field stores different piece of information depending on the state of this field:
+    ///
+    /// - It's empty for an incomplete field (`()`).
+    /// - When a value (or `None`) has ben set, `State` becomes [Complete] and `state` store the
```
yannham, line 54:
```suggestion
    /// - When a value (or `None`) has ben set, `State` becomes [Complete] and stores the
```

core/src/bytecode/ast/builder.rs
```diff
@@ -0,0 +1,716 @@
...
+
+/// A Nickel record field being constructed
+#[derive(Debug)]
+pub struct Field<'ast, State> {
+    /// This field stores different piece of information depending on the state of this field:
+    ///
+    /// - It's empty for an incomplete field (`()`).
+    /// - When a value (or `None`) has been set, `State` becomes [Complete] and stores the
+    ///   optional value of this field.
+    /// - Finally, when the field has been attached to a record, `state` store the parent record.
```
jneem, line 56: The "Finally" here confused me for a bit, because it made me thing that the state lifecycle goes incomplete -> complete -> parent record but this is not the case.

jneem (approved):

core/src/bytecode/ast/builder.rs
```diff
@@ -0,0 +1,716 @@
...
+        self
+    }
+
+    /// Mark the field as optional
+    pub fn optional(self) -> Self {
+        self.set_optional(true)
+    }
+
+    /// Mark the field as optional or not, depending on `opt`
+    pub fn set_optional(mut self, opt: bool) -> Self {
```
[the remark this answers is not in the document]
yannham: Fair enough, I took a look and clap's builder have a similar `require(self, bool)`.

core/src/bytecode/ast/builder.rs
jneem: I think my understanding is the same as what you wrote, which is that the state can go Incomplete -> Complete (and then gets consumed by the record, but it doesn't become a `Field<Record>`) or Record -> Complete. But the wording of the comment initially made me think that the lifecycle went `Field<Incomplete> -> Field<Complete> -> Field<Record>`. Maybe it isn't important, it just tripped me up a little while reading...

yannham: Fair enough, I'll try to reword the explanation to make it clearer :+1:

PR #2086, opened 2024-11-01
Support warnings

jneem:
This adds a bunch of infrastructure for warnings (and for reporting multiple errors, potentially), and also a warning for applying a naked function as a contract.

Most of the PR involves the threading of some extra global state through the program and the vm. I could see a few options:
1. pass the extra state as a parameter to lots of functions
2. pass in the extra state as something to be owned by the vm, and then allow it to be taken back out
3. add a lifetime parameter to the program and vm, so that a `&mut <global state>` can be passed in.

I went with the third option. The first option would require a lot of churn (and then potentially more churn if we want to change the type of the global state). The second option would require the global state to either be a concrete type (but I wanted different types in the cli and nls) or a type parameter. So then given that it looked like we'd want an additional generic parameter anyway, I went for the lifetime parameter of the third option.


core/src/files.rs
```diff
@@ -117,7 +117,7 @@ impl Files {
         // This implementation would be a little nicer if `Vector` supported mutable access.
         // unwrap: we're allowed to panic if file_id is invalid
         let mut old = self.get(file_id).unwrap().clone();
-        old.source = source.into();
+        old = File::new(old.name, source);
```
jneem, line 120: This fixes a recently-introduced but where the `line_starts` cache wasn't updated. Unlikely to trigger in `nickel` itself, but it caused some problems for nls.

yannham:
I haven't looked at it in details yet, but why won't something like `Vec<Box<dyn IntoDiagnostic>>` work for 2. ? Do you need to access the actual structure of the data? As a first impression, I'm not too keen on adding yet another lifetime  parameter to `Vm` and `Program` just for warnings if it can be avoided (and as long as it doesn't impact performance on the happy path, but this shouldn't).

jneem:
Fair enough, I'll explore option 2 a little. I didn't want to tie the implementation to a vec of diagnostics, because I want to allow for the possibility that some diagnostics handlers will want to issue diagnostics immediately while evaluation continues.

I do have another secret motivation for wanting to try out a lifetime parameter on `Vm`, which is to allow a `Cache` that isn't owned by the vm, so that it can be re-used.

yannham:
> Fair enough, I'll explore option 2 a little. I didn't want to tie the implementation to a vec of diagnostics, because I want to allow for the possibility that some diagnostics handlers will want to issue diagnostics immediately while evaluation continues.

By issuing you mean that you might want to print a diagnostic immediately, instead of accumulating and only at the end printing them? In that case I think you can still keep the `Renderer` idea. But as this isn't expected to be called a lot or to matter that much for performance, I would put it under a `Box<dyn _>` instead of carrying a lifetime parameter everywhere. In fact, it's a separate subject but that's probably what we should havedone as well for the `ImportResolver` parameter, because the number of times `resolve` is called doesn't justify the pollution of the `VirtualMachine` type with `R: ImportResolver` everywhere,  IMHO.

> I do have another secret motivation for wanting to try out a lifetime parameter on Vm, which is to allow a Cache that isn't owned by the vm, so that it can be re-used.

That's a fair motivation, but probably slightly orthogonal (at least in the `Box<dyn _>` vs lifetime parameter debate). Also, would those lifetime be necessarily  the same ? (I guess the lifetime is in fact a way of speaking about `'self` in some sense?)

jneem:
> That's a fair motivation, but probably slightly orthogonal (at least in the Box<dyn _> vs lifetime parameter debate). Also, would those lifetime be necessarily the same ? (I guess the lifetime is in fact a way of speaking about 'self in some sense?)

I think we could get by with a single lifetime parameter, but I guess that's something to look into later anyway.

> But as this isn't expected to be called a lot or to matter that much for performance, I would put it under a Box<dyn _> instead of carrying a lifetime parameter everywhere.

The previous iteration did in fact use a `Box<dyn Reporter>`, but it had a lifetime (i.e. `Box<dyn Reporter + 'ctx>`) because it was internally using a mutable reference to push errors into something owned by something longer-lived than the `Program`. The latest version removes the lifetime, and instead uses an mpsc channel to deliver the errors. It feels a little strange because I usually think of channels as involving threads, but it actually worked pretty well I think. (And it's encapsulated, so no one else has to think about the channels)

core/src/error/warning.rs
```diff
@@ -0,0 +1,40 @@
...
+                }
+
+                if let Some(func_span) = func_pos.into_opt() {
+                    labels.push(secondary(&func_span).with_message("this function"));
+                }
+
+                vec![Diagnostic::warning()
+                    .with_message("plain functions as contracts are deprecated")
+                    .with_labels(labels)
+                    .with_notes(vec!["use one of the constructors in `std.contract` instead, like `std.contract.custom`".to_owned()])
```
yannham, line 35: I feel like this is slightly more precise (making more explicit the connection between the mentioned `std.contract` and the function that caused the warning), but feel free to ignore if you don't agree. Also I think mentioning validators could help, because it might be the case that a lot of custom contracts can be expressed as validators, and it's better to do so whenever possible. The original message could make it sound like `std.contract.custom` is the default, go-to solution.

```suggestion
                    .with_notes(vec!["wrap this function using one of the constructors in `std.contract` instead, like `std.contract.from_validator` or `std.contract.custom`".to_owned()])
```

yannham (approved):
While I was initially a bit skeptical about the overall size of the diff, in the end there are many things I like in this PR, mainly:

- We get rid of a lot of juggling caused by the need to include `program` in CLI errors before; now we can just cheaply clone `files` and call it  a day, which looks much cleaner :+1: (we also got rid of unsatisfying `report_xxx` methods in `Program`)
- Using channels for reporting stuff is actually quite elegant, and corresponds well to the mental model of sending data to a reporter or a terminal.

TODOs that I see for the future:

- Add more warnings (another obvious one would be unused variable, which can be done currently in the free var transform, and in the future in the compiler)
- Various flags to control warnings (I suppose if you run this version of Nickel on a large existing code base that use naked contracts you'll get a full load of warnings). Like `-Werror` and to disable them at least?

core/src/program.rs
```diff
@@ -301,44 +303,57 @@ impl<EC: EvalCache> Program<EC> {
...
         I: IntoIterator<Item = (T, S)>,
         T: Read,
         S: Into<OsString>,
     {
         let inputs = sources.into_iter().map(|(s, n)| Input::Source(s, n));
-        Self::new_from_inputs(inputs, trace)
+        Self::new_from_inputs(inputs, trace, reporter)
+    }
+
+    pub fn with_reporter(self, reporter: impl Reporter<(Warning, Files)> + 'static) -> Program<EC> {
```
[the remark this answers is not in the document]
jneem: Looks like it was just left over from an earlier attempt. I removed it.

jneem:
Thanks! I'm glad the refactors made sense, because it was starting to feel like I lost track of the original goal...

> Various flags to control warnings (I suppose if you run this version of Nickel on a large existing code base that use naked contracts you'll get a full load of warnings). Like -Werror and to disable them at least?

Maybe it makes sense to bikeshed some sort of global nickel config? There's a long discussion [here](https://github.com/clap-rs/clap/discussions/2763) about how to integrate layered config into clap, but it appears there's no resolution yet. (I didn't read the whole thing, just scrolled to the end)

yannham:
> Maybe it makes sense to bikeshed some sort of global nickel config? There's a long discussion https://github.com/clap-rs/clap/discussions/2763 about how to integrate layered config into clap, but it appears there's no resolution yet. (I didn't read the whole thing, just scrolled to the end)

Yes, I agree we need a structured approach, with relations and priorities between env vars, CLI arguments and a potential config file, and a consistent option naming scheme, instead of just adding ad hoc flags here and there.

PR #2090, opened 2024-11-07
Move 'clap' crate dependency behind feature flag in core

akavel:
As a developer of an app embedding Nickel as a scripting language, I don't want to be forced to incur the overhead of a third-party commandline-parsing library (clap) with core Nickel.

The change keeps the new "clap" feature enabled by default in Cargo.toml, to minimize impact on any users.

See also: https://github.com/tweag/nickel/discussions/2088

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,7 +19,7 @@ links = "nix"
 bench = false
 
 [features]
-default = ["markdown", "repl", "doc", "format", "clap"]
+default = ["markdown", "repl", "doc", "format"]
 clap = ["dep:clap", "dep:colorchoice-clap"]
 markdown = ["dep:termimad"]
 repl = ["dep:rustyline", "dep:rustyline-derive", "dep:ansi_term"]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,7 +19,8 @@ links = "nix"
 bench = false
 
 [features]
-default = ["markdown", "repl", "doc", "format"]
+default = ["markdown", "repl", "doc", "format", "clap"]
+clap = ["dep:clap", "dep:colorchoice-clap"]
 markdown = ["dep:termimad"]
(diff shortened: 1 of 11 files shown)


core/src/error/report.rs
```diff
@@ -24,33 +25,41 @@ pub enum ErrorFormat {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct ColorOpt(pub(crate) clap::ColorChoice);
+pub struct ColorOpt(pub(crate) colorchoice::ColorChoice);
```
jneem, line 28: Do we even need this wrapper struct anymore? I think it was probably just there to avoid leaking `clap` into our public API, but maybe it's fine just to export `colorchoice::ColorChoice` since the crate is small and it's aligned exactly with what we want.

core/Cargo.toml
```diff
@@ -19,7 +19,8 @@ links = "nix"
 bench = false
 
 [features]
-default = ["markdown", "repl", "doc", "format"]
+default = ["markdown", "repl", "doc", "format", "clap"]
```
jneem, line 22: I think it's ok to drop clap from the defaults (we'll just have to remember to bump the crate version). Leaving clap out seems like the right choice absent backwards-compat constraints, and we don't care too much about API stability of the core crates.

core/src/error/report.rs
akavel: Looks like the consequence is that the orphan rule strikes and we then need to convert the From impls to helper functions instead. Unless you see some nicer way out? 🤔 If not, I can revert or keep as is, whichever you prefer, and make some further tweaks if you like (e.g. if you want me to do some changes to the function names or whatever).

jneem: I think either option is fine, let's not hold this PR up with bikeshedding.

jneem (approved):

PR #2094, opened 2024-11-15
Add support for packages to nickel-lang-core

jneem:
These are the changes to the core crate, picked out from the `npm` branch.

Essentially, this enables the `import pkg` syntax, and introduces `PackageMap` for figuring out how to interpret those import statements.

It also includes deserialization support for enum variants, because I ran into that while working on manifest files. That's unrelated to most of the changes here, so I could break it out as a separate PR if you want.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -5,7 +5,6 @@ use crate::error::{Error, ImportError, ParseError, ParseErrors, TypecheckError};
 use crate::eval::cache::Cache as EvalCache;
 use crate::eval::Closure;
 use crate::files::{FileId, Files};
-use crate::identifier::Ident;
 use crate::metrics::measure_runtime;
 #[cfg(feature = "nix-experimental")]
 use crate::nix_ffi;
@@ -15,7 +14,7 @@ use crate::position::TermPos;
 use crate::program::FieldPath;
 use crate::stdlib::{self as nickel_stdlib, StdlibModule};
 use crate::term::record::{Field, RecordData};
-use crate::term::{RichTerm, SharedTerm, Term};
+use crate::term::{Import, RichTerm, SharedTerm, Term};
 use crate::transform::import_resolution;
 use crate::typ::UnboundTypeVariableError;
 use crate::typecheck::{self, type_check, TypecheckMode, Wildcards};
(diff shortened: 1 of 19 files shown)


core/src/cache.rs
```diff
@@ -1545,6 +1598,43 @@ fn normalize_abs_path(path: &Path) -> PathBuf {
     ret
 }
 
+/// Normalize a relative path, removing mid-path `..`s.
+///
+/// Like [`normalize_abs_path`], this works only on the path itself
+/// (i.e. not the filesystem) and does not follow symlinks.
+pub fn normalize_rel_path(path: &Path) -> PathBuf {
```
jneem, line 1605: I needed this utility function in the upcoming `nickel_lang_package` crate. Putting it next to `normalize_abs_path` seemed reasonable, but it doesn't actually need to be in this crate.

core/src/term/mod.rs
```diff
@@ -209,6 +209,9 @@ pub enum Term {
     #[serde(skip)]
     Import { path: OsString, format: InputFormat },
 
+    #[serde(skip)]
+    ImportPkg(Ident),
```
yannham, line 213: Although we are getting rid of `Term`, adding a new variant impacts quite a few places with pattern matching where we don't care about imports. Should we just rename `PathOrPackage` to something like `Import`, taking an owned `OsString`, and use that type as the parameter of `Term::Import`? After all the'yre all imports. What do you think?

core/src/cache.rs
```diff
@@ -120,12 +122,19 @@ pub struct Cache {
     rev_imports: HashMap<FileId, HashSet<FileId>>,
     /// The table storing parsed terms corresponding to the entries of the file database.
     terms: HashMap<FileId, TermEntry>,
+    /// A table mapping FileIds to the package that they belong to.
+    ///
+    /// Path dependencies have already been canonicalized to absolute paths.
+    package: HashMap<FileId, PathBuf>,
```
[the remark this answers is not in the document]
jneem: I went with `packages` since `PackageMap` is used elsewhere.

core/src/term/mod.rs
jneem: Yep, that turned out nicer.

PR #2095, opened 2024-11-15
Add some git utilities

jneem:
This adds a mini-crate with some git-related utilities. It's split off from the main packaging PR, and doesn't interact with anything else.

### DIFF
--- a/git/src/lib.rs
+++ b/git/src/lib.rs
@@ -2,7 +2,14 @@
 //! that nickel uses.
 
 use anyhow::anyhow;
-use gix::{objs::Kind, ObjectId};
+use gix::{
+    interrupt::IS_INTERRUPTED,
+    objs::Kind,
+    progress::Discard,
+    remote::{self, fetch, Direction},
+    worktree::state::checkout,
+    ObjectId,
+};
 use std::{num::NonZero, path::Path};
 
 /// An error that occurred during a git operation.
@@ -105,16 +112,10 @@ impl std::fmt::Display for Target {
     }
 }
 
-impl Target {
-    fn refspec(&self) -> String {
-        self.to_string()
-    }
-}
-
-fn source_object_id(source: &gix::remote::fetch::Source) -> Result<ObjectId> {
+fn source_object_id(source: &fetch::Source) -> Result<ObjectId> {
     match source {
-        gix::remote::fetch::Source::ObjectId(id) => Ok(*id),
-        gix::remote::fetch::Source::Ref(r) => {
+        fetch::Source::ObjectId(id) => Ok(*id),
+        fetch::Source::Ref(r) => {
             let (_name, id, peeled) = r.unpack();
 
             Ok(peeled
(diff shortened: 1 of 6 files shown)


git/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    },
+
+    #[error("target `{0}` not found")]
+    TargetNotFound(Target),
+
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
yannham, line 24: Is this a default parameter? Meaning that if you write `Result<T>` it's equivalent as if you had defined `type Result<T> = std::result::Result<T, Error>`, but you can still use `Result<T, E'>` for `E' != Error` ? I didn't know about this!

git/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Target a specific tag.
+    Tag(String),
+    /// Target a specific commit.
+    ///
+    /// Currently, we only support a full commit: this needs to be a full hex-encoded
+    /// sha hash. We could try to support prefixes also, but it requires some work
+    /// and it *appears* to be inherently less efficient because there doesn't seem
+    /// to be a way to fetch it in one shot. At least, when `cargo` needs to fetch an
+    /// abbreviated hash it fetches everything and then looks for the hash among the
+    /// things that it finds.
```
yannham, line 93: Agreed :+1: a full hash is not too much to ask for now

git/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+            Target::Head => write!(f, "HEAD"),
+            Target::Branch(branch) => write!(f, "refs/heads/{branch}"),
+            Target::Tag(tag) => write!(f, "refs/tags/{tag}"),
+            Target::Commit(c) => write!(f, "{}", c),
+        }
+    }
+}
+
+impl Target {
+    fn refspec(&self) -> String {
```
yannham, line 109: Is it worth it to have an alias instead of just using `to_string()`? Or do you think it conveys the meaning more precisely?

git/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    let connection = remote.connect(gix::remote::Direction::Fetch).wrap_err()?;
+    let outcome = connection
+        .prepare_fetch(
+            &mut gix::progress::Discard,
+            gix::remote::ref_map::Options::default(),
+        )
+        .wrap_err()?
+        // For now, we always fetch shallow. Maybe for the index it's more efficient to
+        // keep a single repo around and update it? But that might be in another method.
+        .with_shallow(gix::remote::fetch::Shallow::DepthAtRemote(
```
yannham, line 166: Nitpick: there are a lot of `gix::remote::fetch::` prefixes. Maybe it would be worth importing at least `fetch`, or even a few symbols directly?

git/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+/// shallow fetch).
+///
+/// Only the contents of the git repository will be written to the given
+/// directory; the git directory itself will be discarded.
+pub fn fetch(spec: &Spec, dir: impl AsRef<Path>) -> Result<ObjectId> {
+    let dir = dir.as_ref();
+    std::fs::create_dir_all(dir).with_path(dir)?;
+
+    // Fetch the git directory somewhere temporary.
+    let git_tempdir = tempfile::tempdir().wrap_err()?;
```
yannham, line 142: Maybe a naive question, but why do we need a temporary directory first? Can't we do everything in `dir` directly?

yannham (approved):
Unrelated, but as the number of our crates grow, I wonder if we should put them under a common `crate` directory, because it's not very clear anymore what is a crate and what isn't, I guess

git/src/lib.rs
jneem: We need to put the `.git` stuff somewhere. There's probably no harm in putting it in `dir`, but I thought it was cleaner to just populate `dir` with the contents, and not the `.git` stuff.

git/src/lib.rs
jneem: Yeah, probably not worth it. I'll remove.

git/src/lib.rs
yannham: Oh, you can choose to put the git stuff separately, I see. That's fine by me :+1:

PR #2098, opened 2024-11-19
Fix spawning of too many LSP `TestHarness`es

rben01:
During benchmarking, each criterion run would spawn `TestHarness`, each with its own `lsp` process. They weren't cleaned up between criterion runs, so eventually they would exhaust the system’s process pool. Fixed this by explicitly killing and cleaning up the lsp at the end of a criterion run.


jneem (approved):

yannham:
Thanks for that @rben01 ! I've had LSP tests randomly hanging in the past but haven't had the drive to investigate. That must be the culprit.

PR #2099, opened 2024-11-19
Added `from_*` functions for easy deserialization for crates using nickel as a lib

### LINKED ISSUE #2097: Equivalents of `serde_json::from_*`
It would be nice if nickel had functions that would wrap the entire deserialization pipeline. For instance, serde_json has `from_str`, `from_reader`, etc. In nickel, however, you have to create a program, manually set its source name and stderr, then eval it, and then deserialize.

Something like this (but without the nested `Result`; it's necessary in this example because AFAICT `RustDeserializationError` isn't `Into<nickel_lang_core::error::Error>`).
```rust
fn from_reader<'de, R, T>(
	r: R,
) -> Result<Result<T, RustDeserializationError>, nickel_lang_core::error::Error>
where
	R: std::io::Read,
	T: serde::Deserialize<'de>,
{
	Ok(T::deserialize(
		Program::<CacheImpl>::new_from_source(r, "source", std::io::stderr())
			.map_err(IOError::from)?
			.eval_full_for_export()?,
	))
}
```

None

**Additional context**
Related: #1751

rben01:
Closes #2097

### DIFF
--- a/core/src/deserialize.rs
+++ b/core/src/deserialize.rs
@@ -10,7 +10,7 @@ use serde::de::{
     VariantAccess, Visitor,
 };
 
-use crate::error::{IOError, Sink};
+use crate::error::{self, NullReporter};
 use crate::eval::cache::CacheImpl;
 use crate::identifier::LocIdent;
 use crate::program::{Input, Program};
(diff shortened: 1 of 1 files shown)


core/src/deserialize.rs
```diff
@@ -30,6 +35,70 @@ macro_rules! deserialize_number {
...
+where
+    T: serde::Deserialize<'a>,
+    Content: std::io::Read,
+    Source: Into<OsString>,
+{
+    Ok(T::deserialize(
+        Program::<CacheImpl>::new_from_input(
+            input,
+            std::io::stderr(),
+            Sink::<(crate::error::Warning, crate::files::Files)>::default(),
```
jneem, line 66: Since you aren't actually emitting warnings, I'd use a `NullReporter` here instead of a `Sink`

core/src/deserialize.rs
```diff
@@ -30,6 +35,70 @@ macro_rules! deserialize_number {
     };
 }
 
+#[derive(Debug, PartialEq)]
+pub enum EvalOrDeserError {
+    Nickel(crate::error::Error),
```
jneem, line 40: Maybe a `Nickel { error: crate::error::Error, files: crate::files::Files }` would be better here, since it would allow the caller to print nice diagnostics if they want to.

jneem (commented):
Thanks for the PR, this is certainly a much nicer public API. I just have a couple of requests around error handling.

core/src/deserialize.rs
rben01: I'm having trouble figuring out the right way to turn an `Input` into a `Files`. When constructing `EvalOrDeserError` from a `crate::error::Error`, what should `files` look like? It seems like there is a bit of a chicken and egg problem, where constructing `Files` requires reading a file, which can fail, which can return the very error we're trying to construct...

jneem (approved):

PR #2102, opened 2024-11-21
[RFC007] Improve/simplify record representation in the new AST

yannham:
Instead of elaborating piecewise definitions (such as `{foo.bar = 1, foo.baz = 2}`) directly at the parsing stage, this commit makes the new AST closer to the source language by making record a list of field definitions, where the field "name" (left hand side of `=`) can be a sequence of identifiers and dynamic strings. This representation is used internally by the parser; we now make it the default in the new AST, such that the migration of the parser in #2083  won't have to do this elaboration at all. The elaboration is offloaded to the conversion to `RichTerm`, which happens in the `ast::compat` module.

This makes the AST closer to the source language.

The first motivation is that it'll be better for the LSP, where some open issues on the tracker are caused by the inability to trace what the LSP get back to the original piecewise definitions.

The second reason is that we can't actually elaborate a piecewise definition while staying in the new AST correctly as of today: the new AST only has one record variant, which is recursive by default, but this doesn't match the way recursion and scoping work for piecewise definition. For example, `{foo.bar = 1, baz.foo = foo + 1}` works fine in today's Nickel (evaluate to `{foo = {bar = 1}, baz {foo = 2}}`), but if we elaborate it in the new AST naively, we'll get an infinite recursion: `{foo = {bar = 1}, baz = {foo = foo + 1}}`.

Mailine Nickel currently uses a non recursive `Record` for that, but we don't want to introduce such "runtime dictionary" in the new AST as they can't be expressed in the source language. Instead, we rather keep records as piecewise defined without transformation and will do further elaboration when needed later in the pipeline, during typechecking, future compilation, or in the meantime when converting the new AST representation to mainline Nickel.

### DIFF
--- a/core/src/bytecode/ast/compat.rs
+++ b/core/src/bytecode/ast/compat.rs
@@ -5,8 +5,10 @@
 //! in [crate::bytecode::ast].
 
 use super::{primop::PrimOp, *};
-use crate::{label, term, typ as mline_type};
+use crate::{combine::Combine, label, position::RawSpan, term, typ as mline_type};
+use indexmap::IndexMap;
 use smallvec::SmallVec;
+use std::iter;
 
 /// Convert from the mainline Nickel representation to the new AST representation. This trait is
 /// mostly `From` with an additional argument for the allocator.
@@ -163,11 +165,22 @@ impl<'ast> FromMainline<'ast, term::TypeAnnotation> for Annotation<'ast> {
     }
 }
 
-impl<'ast> FromMainline<'ast, term::record::Field> for record::Field<'ast> {
-    fn from_mainline(alloc: &'ast AstAlloc, field: &term::record::Field) -> Self {
-        record::Field {
-            value: field.value.as_ref().map(|term| term.to_ast(alloc)),
-            metadata: field.metadata.to_ast(alloc),
+impl<'ast> FromMainline<'ast, (LocIdent, term::record::Field)> for record::FieldDef<'ast> {
+    fn from_mainline(
+        alloc: &'ast AstAlloc,
+        (id, content): &(LocIdent, term::record::Field),
+    ) -> Self {
+        let pos = content
+            .value
+            .as_ref()
+            .map(|value| id.pos.fuse(value.pos))
+            .unwrap_or(id.pos);
+
+        record::FieldDef {
+            path: alloc.alloc_iter(iter::once(record::FieldPathElem::Ident(*id))),
+            value: content.value.as_ref().map(|term| term.to_ast(alloc)),
+            metadata: content.metadata.to_ast(alloc),
+            pos,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


core/src/bytecode/ast/mod.rs
```diff
@@ -266,6 +266,28 @@ impl AstAlloc {
...
+    ///
+    /// See [Self::alloc].
+    pub fn alloc_iter<T, I>(&self, iter: I) -> &[T]
+    where
+        I: IntoIterator<Item = T>,
+        I::IntoIter: ExactSizeIterator,
+    {
+        self.generic_arena.alloc_slice_fill_iter(iter)
+    }
+
```
yannham, line 290: Those functions are a bit orthogonal to the matter, but they were introduced in the commit cherry picked here from #2083 and it was easier to keep them. After #2083 is merged I'll make a pass on the `AstAlloc` API to trim it down a bit anyway.

core/src/bytecode/ast/compat.rs
```diff
@@ -791,13 +831,111 @@ impl<'ast> FromAst<Annotation<'ast>> for term::TypeAnnotation {
...
+            record::FieldPathElem::Expr(chunks, pos) => {
+                let chunks = chunks.iter().map(ToMainline::to_mainline).collect();
+                FieldName::Expr(chunks, *pos)
+            }
+        }
+    }
+}
+
+impl<'ast> FromAst<record::FieldDef<'ast>> for (FieldName, term::record::Field) {
+    fn from_ast(field: &record::FieldDef<'ast>) -> Self {
```
yannham, line 863: This is mostly `parser::utils::build_record` adapted to take a new AST as an input.

yannham: Sorry, this is rather `elaborate_field_def`.

core/src/bytecode/ast/compat.rs
```diff
@@ -1252,3 +1366,178 @@ impl<'ast> FromAst<&'ast Ast<'ast>> for term::RichTerm {
...
+/// fields). If a field is defined several times, the different definitions are merged, statically
+/// if possible.
+
+impl<'ast> FromAst<Record<'ast>>
+    for (
+        term::record::RecordData,
+        Vec<(term::RichTerm, term::record::Field)>,
+    )
+{
+    fn from_ast(record: &Record<'ast>) -> Self {
```
yannham, line 1380: This is mostly `parser::utils::build_record` adapted to take a new AST as an input.

core/src/bytecode/ast/compat.rs
```diff
@@ -1252,3 +1366,178 @@ impl<'ast> FromAst<&'ast Ast<'ast>> for term::RichTerm {
...
+/// Merge two fields by performing the merge of both their value (dynamically if
+/// necessary, by introducing a merge operator) and their metadata (statically).
+///
+/// If the values of both fields are static records ([`term::Term::Record`]s), their merge is
+/// computed statically. This prevents building terms whose depth is linear in the number of fields
+/// if partial definitions are involved. This manifested in
+/// https://github.com/tweag/nickel/issues/1427.
+///
+/// This is a helper for the conversion of a record definition to mainline.
+fn merge_fields(
```
yannham, line 1459: This was taken and type-adapted from `parser::utils` as well.

core/src/bytecode/ast/mod.rs
yannham: That's why I talk about a marker trait that is yet to be written, but will exist in the future.

core/src/bytecode/ast/record.rs
```diff
@@ -1,18 +1,85 @@
...
+    /// one element which is a static identifier. Returns `None` otherwise.
+    pub fn path_as_ident(&self) -> Option<LocIdent> {
+        if self.path.len() > 1 {
+            return None;
+        }
+
+        self.path.first().and_then(|path_elem| match path_elem {
+            FieldPathElem::Expr(..) => None,
+            FieldPathElem::Ident(ident) => Some(*ident),
+        })
```
jneem, line 72: Does a slice pattern like
```
match &self.path {
  [FieldPathElem(ident)] => Some(*ident),
  _ => None
}
```

work?

jneem (approved):

core/src/bytecode/ast/record.rs
yannham: Ah, I always forget about slice patterns. Looks much better, thanks :+1:

PR #2103, opened 2024-11-21
Removed errant `proc.kill()`

rben01:
In #2098 I tried `kill`ing the process. It turns out that doing this before `wait`ing leads to out of order IO, leading to IO errors when Bincode tries to deserialize. It also turns out that killing isn't even necessary; waiting is sufficient.

So this PR removes that `kill` line, getting rid of the IO errors introduced in #2098.

Also Added `finish` to `benchmark_diagnostics` as well (should have been there all along). Thought: do we want to just `impl Drop for TestHarness`? This might make the panicking situation tricky (generally, you shouldn't panic in `drop`), whereas explicitly calling `harness.finish()`, while possible to forget, makes handling panics just work.

### DIFF
--- a/lsp/lsp-harness/src/jsonrpc.rs
+++ b/lsp/lsp-harness/src/jsonrpc.rs
@@ -133,8 +133,7 @@ impl Server {
     /// exhausting the PID pool when thousands of servers are created.
     pub fn die(mut self) -> Result<()> {
         self.shutdown()?;
-        self.proc.kill()?;
-        self.proc.wait()?; // force cleanup of process (kill alone is insufficient)
+        self.proc.wait()?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


yannham (approved):

jneem (approved):

jneem:
> Thought: do we want to just impl Drop for TestHarness?

I don't have a strong opinion. Since this is test code, accidentally aborting due to a panic in `drop` isn't a big deal. On the other hand, we don't construct `TestHarness` that often, so remembering to call `finish` is probably not too hard. So either way is really ok with me.

rben01:
Since I did in fact forget to call `finish` once, it's probably best to `impl Drop for Server`, bundling shutting down and waiting into it, so that they can't be forgotten. I don't expect the errors to actually ever be thrown, so panicking should be rare. And `shutdown` was never actually used anywhere, so I just moved it right into `drop`.

jneem (approved):

PR #2110, opened 2024-11-27
Do deep eval for doctests

jneem:
Previously, we weren't doing deep eval of doctests, which would fail to catch some errors.

I think *maybe* there are use-cases where you don't want full evaluation, but I'm pretty sure it should be the default. (And then we could support a test annotation to make it lazy if anyone actually needs that)


yannham:
Ah. Maybe it's a sign that it's time to implement freezing properly :sweat_smile: I can try to take a look at that after the ongoing PR that I have (a simple cleaning of the `AstAlloc` interface)

jneem:
It works!

yannham (approved):

PR #2112, opened 2024-11-27
[RFC007] Make function node multi-ary by default

yannham:
Depends on #2111.

In the new AST, the representation function application has been changed to be multi-ary by default (as is primop application as well). It just matches the syntax better and it avoids having to desugar a n-ary application to nested unary applications as before, saving space and time (for the time being we still have  this desugaring to do when converting to the mainline AST). The only cost is that unary applications are slighty larger in memory (basically one word for the size of the slice), but we expect the gain from n-ary applications for `n > 1` to dwarf this. It also avoids having to come up with dubious position for the generated intermediate nodes.

This commit applies the same treatment to function abstraction since the same arguments apply. Passing by, we also specialize macros like `fun` and `app` for one argument to avoid vec allocation.

### DIFF
--- a/core/src/bytecode/ast/compat.rs
+++ b/core/src/bytecode/ast/compat.rs
@@ -277,8 +277,33 @@ impl<'ast> FromMainline<'ast, term::Term> for Node<'ast> {
                     }
                 }))
             }
-            Term::Fun(id, body) => alloc.fun(Pattern::any(*id), body.to_ast(alloc)),
-            Term::FunPattern(pat, body) => alloc.fun(pat.to_ast(alloc), body.to_ast(alloc)),
+            t @ (Term::Fun(_, body) | Term::FunPattern(_, body)) => {
+                let fst_arg = match t {
+                    Term::Fun(id, _) => Pattern::any(*id),
+                    Term::FunPattern(pat, _) => pat.to_ast(alloc),
+                    // unreachable!(): we are in a match arm that matches either Fun or FunPattern
+                    _ => unreachable!(),
+                };
+
+                let mut args = vec![fst_arg];
+                let mut maybe_next_fun = body;
+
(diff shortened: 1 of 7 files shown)


core/src/bytecode/ast/compat.rs
```diff
@@ -1205,10 +1234,31 @@ impl<'ast> FromAst<Node<'ast>> for term::Term {
...
+                            _ => term::Term::FunPattern((*arg).to_mainline(), acc),
+                        };
+
+                        // [^nary-constructors-unrolling]: this case is a bit annoying: we need to
+                        // extract the position of the intermediate created functions to satisfy
+                        // the old AST structure, but this information isn't available directly.
+                        //
+                        // What we do here is to fuse the span of the term being built and the one
+                        // of the current argument, which should be a reasonable approximation (if
+                        // not exactly the same thing).
```
jneem, line 1256: I'm not sure if it's worth an extra fold, but I think a better position (and a closer approximation of the old behavior) would be to set each "inner" function to have the *entire* fused args+body as its position.

jneem (approved):

core/src/bytecode/ast/compat.rs
yannham: I don't have a strong opinion. I think fusing the arguments as done in this PR is a bit exotic at first sight, but has the advantage of being more precise (like for type errors and such, it show you which part of the multi-ary function precisely is at fault, while the old behavior would always show you the whole function for several different terms). On the other hand, matching the old behavior is maybe the most conservative approach.

yannham: After a second thought, we won't have this issue anymore (or at least not in the same form) once we have only one AST, so I'll proceed with the current version - it's temporary anyways.

Issue #1866, opened 2024-03-24
Confusing merge behaviour in `std.record.{update,remove}`


solson:
**Describe the bug**

The `std.record.update` and `std.record.remove` functions don't recalculate changes in fields that depend on the modified field, but those recalculations do take effect after a merge, even a trivial one (`& {}`).

**To Reproduce**

```
nickel> std.record.update "a" 2 { a = 1, b = a }
{ a = 2, b = 1, }

nickel> std.record.update "a" 2 { a = 1, b = a } & {}
{ a = 2, b = 2, }

nickel> std.record.remove "a" { a = 1, b = a }
{ b = 1, }

nickel> std.record.remove "a" { a = 1, b = a } & {}
error: unbound identifier `a`
  ┌─ <repl-input-261>:1:36
  │
1 │ std.record.remove "a" { a = 1, b = a } & {}
  │                                    ^ this identifier is unbound
```

**Expected behavior**

This might be a language design question, but the least surprising to me would be:

```
nickel> std.record.update "a" 2 { a = 1, b = a }
{ a = 2, b = 2, }

nickel> std.record.remove "a" { a = 1, b = a }
// Error if you try to use `.b`, so the REPL would error trying to print the result.
```

(And in general attempt to enforce that `x & {}` is always equivalent to `x` if `x` is a record, unless there is a subtle reason that shouldn't be the case.)

**Environment**
 - OS name + version: nixos-unstable
 - Version of the code: Nickel 1.5.0

**Additional context**

I ran into this while testing out some more complicated merging/overriding scenarios, and this behaviour was a lot more surprising in context before I tracked down what made "recalculation" of a field happen or not happen.

yannham:
Thanks for reporting. This is definitely a bug. I think the original design guideline was that only merge could impact recursive fields, and as soon as you use a record as a "normal" dictionary (`std.record.{update,remove,insert}`), you don't get the recursive behavior (otherwise, you should just use merge instead - although `remove` can't be emulated with merge, but this is on purpose).

Following this guideline, I believe the right result should be the following:

```
nickel> std.record.update "a" 2 { a = 1, b = a }
{ a = 2, b = 1, }

nickel> std.record.update "a" 2 { a = 1, b = a } & {}
{ a = 2, b = 1, }

nickel> std.record.remove "a" { a = 1, b = a }
{ b = 1, }

nickel> std.record.remove "a" { a = 1, b = a } & {}
{ b = 1, }
```

I see two ways to achieve that:

[...]

PR #2113, opened 2024-11-29
Add a freeze primop and stdlib wrapper

### LINKED ISSUE #1877: Proposal: freeze recursive records upon insert/remove/update/map
## Motivation

As stated in #1866, the current semantics of record operations (insertion, removal, update and map) with respect to merging is confusing, and in fact accidental.

The initial design guideline was that record operations shouldn't interact with merging, and shouldn't trigger recursive recomputations (as overriding through merging does): you either use the merging system to get those, but when you use dict-like operations, then the record is considered as a normal dictionary.

However, this guideline is a bit vague, as noticed in #1866. As a concrete example, when removing a field which has reverse dependencies, should we freeze the reverse dependencies only or the whole record? That is, what should be the result of `(std.record.remove "to_remove" {foo = bar + 1, bar = 0, to_remove = false, rev_dep = to_remove}) & {bar | force = 1, to_remove = true}`?

Freezing the whole record would give `{foo = 1, bar = 1, to_remove = true, rev_dep = false}`. Freezing only reverse dependencies gives `{foo = 2, bar = 1, to_remove = true, rev_dep = false}`. If `record.remove` would preserve recursivity entirely, we would get `{foo = 2, bar = 1, to_remove = true, rev_dep = true}`.

### Only freeze reverse deps

The argument for freezing only the reverse dependencies is the following: try to preserve the recursive structure of the record as much as possible (that is, as far as it's reasonable). When removing a field, we need to at least freeze the reverse dependencies, but the ones that don't interact with that field could very well stay recursive and overridable. This is more expressive, as even after a dictionary operation, one can still apply merge operations.

### Freeze the whole record

The argument for freezing the whole record is simplicity and consistency. It seems that in Nickel, a single object - records - is used to represent several things: (record) contracts, partial configurations (modules) and classical key-value dictionaries. Merging makes sense for contracts and partial configurations, but not much for dictionaries (one can just remove, insert and update there). There is thus the idea that those two different "modes" shouldn't be mixed: one can use the merging system to build up a record value, but once you start to use dictionary functions, it should be considered as a static dictionary and not as a recursive structure anymore. 

The motivation isn't only philosophical: while merging can already makes provenance tracking a bit more challenging, interleaving merging and dictionaries operations, where some dictionary operations might freeze some fields but not others can make it very hard to predict what happens when we override a value on the final result and how overrides propagates through a structure with potential "broken" links.

## Proposal

We propose to freeze the whole record. The rationale is that this seems like the more restrictive but more principled approach to keep the complexity of provenance tracking under control. This is also the simpler mental model for users and thus the more predictable one.

That is, prior to any record operations among `insert`, `remove`, `map` and `update`:

1. All the lazy contracts are applied to the values, and cleaned from the lazy contract fields
2. The value of recursive fields is fixed at this point, that is, future merges won't trigger the recomputation of any field (beside the one being added or overridden)

This semantics will be independently implemented as a `record.freeze` (the name is subject to bikeshedding) stdlib function as well.

## Challenges

It can happen that merging configurations is followed by a post-processing step which for example replaces enum variants with a proper serializable representation. Freezing the whole record could be a problem here, because it makes it impossible to override after-the-fact, for example from the CLI (customize mode).

However, we think that the right solution is to make it possible to perform such tasks (post-processing or custom serializing) within the merging system itself, rather than tweaking dict operations to do so. Typically, a proposal like #1170 would make post-processing part of the metadata system.

One might also want to override a value without having to know about its current priority, while still keeping all the other neat features of merging (keep contracts around and applying them lazily, recursive overriding, etc.). Same conclusion: this use-case mandates something like an asymmetric variant of the merge operator, whose only difference would be that one side would be selected independently from the priorities (and the priority of the other side is preserved), rather than trying to make this fit in a bloated `record.update` that would preserve recursive records.

In the end, note that if recursivity-preserving variants are really needed, we can always provide those under a different namespace in the stdlib - think `record.fix.{insert,remove,update,map}`. The question is really what should be the _default_ behavior of those functions.

yannham:
Following a confusing behavior around the interaction of recursive fields and dictionary operations reported in #1866 (insert, remove, update), it was proposed in #1877 to freeze recursive records before applying dictionary operations.

This PR adds a primitive operation to do so, which applies pending contracts, flushes them, and remove any dependency data, effectively freezing recursive fields to their current value.

A flag is also added to record attributes, in order to remember that some record has already been frozen. Indeed, freezing is linear in the number of fields and pending contracts: freezing at every dictionary operation thus could be costly. Instead, as for closurization, we do it once and then remember that it's been done in flag. It helps dictionary operations preserve frozenness. Only merging or lazy contract applications might require freezing again.

What's missing to close #1877 is to just add a call to `%record/freeze%` in preamble of every dictionary operations. This will be addressed in a follow-up PR.

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -645,9 +645,18 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                                 missing_field_err.into_eval_err(pos, pos_op)
                             })?;
 
+                        // By construction, mapping freezes the record. We set the frozen flag so
+                        // that operations that require the record to be frozen don't have to
+                        // perform the work again.
+                        let attrs = record.attrs.frozen();
+
                         Ok(Closure {
                             body: RichTerm::new(
-                                Term::Record(RecordData { fields, ..record }),
+                                Term::Record(RecordData {
+                                    fields,
+                                    attrs,
+                                    ..record
+                                }),
                                 pos_op_inh,
                             ),
                             env: Environment::new(),
@@ -1176,6 +1185,73 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                     empty.sealed_tail = r.sealed_tail;
                     Ok(Closure {
                         body: RichTerm::new(Term::Record(empty), pos_op.into_inherited()),
+                        env,
+                    })
+                }
+                _ => mk_type_error!("Record"),
+            }),
+            UnaryOp::RecordFreeze => match_sharedterm!(match (t) {
+                Term::Record(record) => {
+                    let mut record = record;
+
+                    if record.attrs.frozen {
+                        // A frozen record shouldn't have a polymorphic tail
+                        debug_assert!(record.sealed_tail.is_none());
+
+                        return Ok(Closure {
+                            body: RichTerm::new(Term::Record(record), pos),
+                            env,
+                        });
+                    }
+
+                    // It's not clear what the semantics of freezing a record with a sealed tail
+                    // would be, as their might be dependencies between the sealed part and the
+                    // unsealed part. Merging is disallowed on records with tail, so we disallow
+                    // freezing as well.
+                    if let Some(record::SealedTail { label, .. }) = record.sealed_tail {
+                        return Err(EvalError::IllegalPolymorphicTailAccess {
+                            action: IllegalPolymorphicTailAction::Freeze,
+                            evaluated_arg: label.get_evaluated_arg(&self.cache),
+                            label,
+                            call_stack: std::mem::take(&mut self.call_stack),
+                        });
+                    }
+
+                    let fields = record
+                        .fields
+                        .into_iter()
+                        .map(|(id, field)| {
+                            let value = field.value.map(|value| {
+                                let pos = value.pos;
+                                RuntimeContract::apply_all(
+                                    value,
+                                    field.pending_contracts.into_iter(),
+                                    pos,
+                                )
+                            });
+
+                            let field = Field {
+                                value,
+                                pending_contracts: Vec::new(),
+                                ..field
+                            }
+                            .closurize(&mut self.cache, env.clone());
+
+                            (id, field)
+                        })
+                        .collect();
+
+                    let attrs = record.attrs.frozen();
+
+                    Ok(Closure {
+                        body: RichTerm::new(
+                            Term::Record(RecordData {
+                                fields,
+                                attrs,
+                                sealed_tail: None,
+                            }),
+                            pos_op.into_inherited(),
+                        ),
                         env,
                     })
                 }
@@ -2250,6 +2326,7 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                                     ))
                                 }
                                 _ => Ok(Closure {
+                                    // Insertion preserves the frozenness
                                     body: Term::Record(RecordData { fields, ..record }).into(),
                                     env: env2,
                                 }),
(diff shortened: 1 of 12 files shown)


core/src/eval/operation.rs
```diff
@@ -1181,6 +1190,73 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
+                        return Ok(Closure {
+                            body: RichTerm::new(Term::Record(record), pos),
+                            env,
+                        });
+                    }
+
+                    // It's not clear what the semantics of freezing a record with a sealed tail
+                    // would be, as their might be dependencies between the sealed part and the
+                    // unsealed part. Merging is disallowed on records with tail, so we disallow
+                    // freezing as well.
```
jneem, line 1210: Could this raise a back-compat issue once the std.record functions are updated to freeze? I didn't succeed to construct an example that calls `std.record.remove` on a record with a sealed tail, but maybe it's possible?

jneem (approved):

core/src/eval/operation.rs
yannham: I think it's hard to build without referring to primops directly. On paper, you might say that stuff like `insert_foo | forall r. {; r} -> { foo : Number; r }` or `remove_bar | forall a r. {bar: a; r} -> {; r}` looks reasonable, but to write those using the stdlib you'd need to cast back the result of dictionary operations - since they operate on dictionary types - but you can't name `r`, so you can't fill the blank in `std.record.insert "foo" 5 record | {foo : Number; ????}`.

Granted, it's entirely accidental (due to the absence of something like scoped type variables), but the consequence is that you can't write today without resorting to `%record/insert%`, and the idea is to not touch the latter, just add freezing in the stdlib wrappers.

Even if we had scoped type variables, a second issue is that stdlib dictionary functions use dictionary types for their input and output. The corresponding contract maps over its argument, but mapping isn't permitted either on a record with a sealed tail:

```nickel
nickel> let f | forall r. {foo : Number; r} -> {_ : Number} = fun x => x
nickel> f {foo = 1, bar = 2}
       : cannot map over a record sealed by a polymorphic contract
  ┌─ <repl-input-0>:1:19
  │
1 │ let f | forall r. {foo : Number; r} -> {_ : Number} = fun x => x
  │                   ----------------- expected type of an argument of an inner call
  │
  ┌─ <repl-input-2>:1:3
  │
1 │ f {foo = 1, bar = 2}
  │   ------------------ evaluated to this expression
```

So right now, it seems that dictionaries are pretty incompatible with polymorphic tails anyway. That being said, we could ask if those maybe _should_ be well-typed / possible to write. Let's say we have scoped type variables or whatnot to make that possible. I think the question of the semantics remain: what should we do upon freezing a record with a sealed tail? 

1. We only freeze the non-sealed part. For the caller, it means that you call to some polymorphic function on records, and what you get back is that _some_ of your recursive dependencies are erased, but not all. I think that's really hard to reason about and error-prone.
2. We look under the tail and remove dependencies there as well. We might need to do that recursively, as tails can be nested. This is a bit more reasonable for the caller, but it means that the sacred tail isn't sacred anymore, as it can be touched indirectly by freezing.
3. We refuse to freeze polymorphic tails, as in the current PR.

Another idea that came from a discussion with @sjoerdvisscher is that maybe this idea of either being in dictionary mode or not should be embodied by types/contracts, and not chosen by functions individually. This would have the merit of making it clear from the interface of a function if freezing happens, and freezing would be exerted by the dictionary type `{_ : T}`. I haven't yet thought about this alternative enough to be sure, but in any case this concerns a follow-up PR. The present PR is about introducing the freezing operation itself; it's not yet used anywhere.

solson:
> What's missing to close https://github.com/tweag/nickel/issues/1877 is to [...]

@yannham I think GitHub saw `close #1877` and ironically took that as an instruction to auto-close it, which I think was not intended.

PR #2118, opened 2024-12-02
Fixed `Debug` impl of `EvalOrDeserError` printing entire source of files

rben01:
Now it just prints filenames, which is much more useful.


yannham (approved):
Thanks! The CI is failing for reason unrelated to this PR; this is being addressed in parallel. Might need a rebase then.

PR #2122, opened 2024-12-11
Disabled default benches throughout the workspace

rben01:
This allows running `cargo bench -- --criterion-specific-arg` without erroring due to `error: Unrecognized option: 'criterion-specific-arg'`.

https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options explains why this is necessary; I've copied it below for posterity.

> By default, Cargo implicitly adds a `libtest` benchmark harness to your crate when benchmarking, to handle any `#[bench]` functions, even if you have none. It compiles and runs this executable first, before any of the other benchmarks. Normally, this is fine - it detects that there are no `libtest`benchmarks to execute and exits, allowing Cargo to move on to the real benchmarks. Unfortunately, it checks the command-line arguments first, and panics when it finds one it doesn't understand. This causes Cargo to stop benchmarking early, and it never executes the Criterion.rs benchmarks.
>
> This will occur when running `cargo bench` with any argument that Criterion.rs supports but `libtest` does not. For example, `--verbose` and `--save-baseline` will cause this issue, while `--help` will not. There are two ways to work around this at present:
>
> You could run only your Criterion benchmark, like so:
>
> `cargo bench --bench my_benchmark -- --verbose`
>
> Note that `my_benchmark` here corresponds to the name of your benchmark in your `Cargo.toml` file.
>
> Another option is to disable benchmarks for your lib or app crate. For example, for library crates, you could add this to your `Cargo.toml` file:
>
> ```
> [lib] 
> bench = false
> ```
>
> If your crate produces one or more binaries as well as a library, you may need to add additional records to `Cargo.toml` like this:
>
> ```
> [[bin]]
> name = "my-binary" 
> path = "src/bin/my-binary.rs" 
> bench = false
> ```
>
> This is because Cargo automatically discovers some kinds of binaries and it will enable the default benchmark harness for these as well.
>
> Of course, this only works if you define all of your benchmarks in the `benches` directory.
>
> See [Rust Issue #47241](https://github.com/rust-lang/rust/issues/47241) for more details.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -113,9 +113,9 @@ harness = false
 name = "arrays"
 harness = false
 
-# [[bench]]
-# name = "records"
-# harness = false
+[[bench]]
+name = "records"
+harness = false
 
 [[bench]]
 name = "serialization"
--- a/git/Cargo.toml
+++ b/git/Cargo.toml
@@ -4,6 +4,9 @@ description = "Git utility functions for internal use in nickel"
 version = "0.1.0"
 edition = "2021"
 
+[lib]
+bench = false
+
 [dependencies]
 anyhow.workspace = true
 gix = { workspace = true, features = ["blocking-network-client", "serde"] }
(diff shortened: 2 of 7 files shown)


yannham (approved):
LGTM. Thanks!

Issue #2125, opened 2024-12-18
Completion can crash the LSP in 1.9 (index out of bounds in `Files`)


yannham:
**Describe the bug**

I get  occasional crashes since Nickel 1.9 when using the LSP. I can't pinpoint the exact minimal conditions, but I have a reproduction below.

**To Reproduce**

For this to work (well, work meaning "crash NLS" here) **you need to open a fresh file and type it**. For some reason even copy/pasting the MRE in a new file doesn't trigger the bug, with or without the last `.b`.

```
let Schema = {
  field | String,
}
in

let value | Schema = {
  field = "bar",
}
in

value.b
```

The trailing `.b` will trigger completion and crash NLS. The log shows the following:

```text
[2024-12-18T09:33:05Z DEBUG nls::server] handle completion
thread 'main' panicked at core/src/files.rs:238:12:
index out of bounds: the len is 10 but the index is 10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected behavior**

I expect this to not crash NLS.

**Environment**
 - OS name + version:
 - Version of the code: nickel-lang-lsp nls 1.9.0 (rev c58265a9-modified)
 
This is the version of the 1.9 release, just as installed by the release script (hence why the modified and the specific rev).

**Additional context**

As we shipped the new file implementation in 1.9, something must have gone wrong here. cc @jneem

jneem:
Could it be [this](https://github.com/tweag/nickel/pull/2086#discussion_r1827317882) issue maybe?

yannham:
Might be, I'll try out with latest master. If it is, I would be for a patch release if the fix can be extracted - I hit this several times since 1.9 release (maybe because I open a lot of scratch file, and people don't get to create as many new file otherwise?)

yannham:
Fixed in master an in 1.9.1

PR #2127, opened 2024-12-19
Backport nls crash fix

jneem:
In case this is the cause of #2125 here's a backport of the fix. I can't actually reproduce the exact issue, but it sounds like the same problem (especially the part about how it doesn't trigger if you copy-paste, because the issue is caused by a stale cache and so it might not get hit if you enter the input all at once). Also, I did double-check that the files refactor was in 1.9.0 but the warnings PR (which originally contained this fix) wasn't.

Probably targeting the 1.9.0-release branch isn't quite right; I guess we should branch out a 1.9.1-release first?

### DIFF
--- a/core/src/files.rs
+++ b/core/src/files.rs
@@ -117,7 +117,7 @@ impl Files {
         // This implementation would be a little nicer if `Vector` supported mutable access.
         // unwrap: we're allowed to panic if file_id is invalid
         let mut old = self.get(file_id).unwrap().clone();
-        old.source = source.into();
+        old = File::new(old.name, source);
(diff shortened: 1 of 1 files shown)


yannham:
I confirm that this fixes the issue on my hand, thanks! I think we don't need to merge that in fact. The update script takes care of creating the `xxx-release` branch, and just needs a source branch, but I can use `nls-crash-backport` directly. Once the CI is green and I've released the minor version, we can close this PR (for bystanders: because the fix is already in `master`).

yannham:
It's been released, so we can close this. Thanks!

PR #2130, opened 2024-12-23
[RFC007] Extract runtime contract equality out of typecheck

yannham:
This PR continues the preparation work to migrate the typechecker to the new AST. One current blocker is that  `typecheck::eq` is used both by the typechecker and at runtime for contract deduplication. We need to get rid of this module when switching the typechecker implementation, while still needing to perform contract deduplication at runtime (until the landing of the bytecode virtual machine).

This PR extracts the code from `typecheck::eq` for contract deduplication in a new module, so that the runtime isn't dependent on `typecheck` anymore. The code is specialized (as it doesn't need to handle static contract equality), and in order to eliminate the dependence on `UnifType`, we drop the ability to compare polymorphic types. The latter is the only reason `typecheck::eq` operated on `UnifType` instead of `Type` (which makes it dependent on the old typechecker), but the motivation is not very convincing: how many times do we encounter a type annotation, itself inside a contract annotation, such that annotation is a polymorphic type AND it happens to be a candidate for deduplication? Our intuition would be: zero.

This hypothesis has been validated on our working benchmarks, as `--metrics` show detailed data on contract deduplication - including the total count of deduplicated contracts - which doesn't change between this PR and master.

### DIFF
--- a/core/src/eval/contract_eq.rs
+++ b/core/src/eval/contract_eq.rs
@@ -1,7 +1,7 @@
 //! Computation of contract equality.
 //!
 //! This module is temporary, and has been extracted from the typechecker operating in the legacy
-//! AST (`RichTerm`). It's used by the current tree-walking virtual machine, but will be scraped
+//! AST (`RichTerm`). It's used by the current tree-walking virtual machine, but will be scrapped
 //! once the bytecode virtual machine (RFC007) is operational. At the time, we won't have the AST
 //! around at runtime anymore, and will use an entirely different technique for contract
 //! deduplication.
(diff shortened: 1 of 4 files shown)


core/src/eval/contract_eq.rs
```diff
@@ -0,0 +1,518 @@
+//! Computation of contract equality.
+//!
+//! This module is temporary, and has been extracted from the typechecker operating in the legacy
+//! AST (`RichTerm`). It's used by the current tree-walking virtual machine, but will be scraped
```
jneem, line 4:
```suggestion
//! AST (`RichTerm`). It's used by the current tree-walking virtual machine, but will be scrapped
```

jneem (approved):

PR #2131, opened 2024-12-23
Freeze records on insert/remove/update.

jneem:
This finishes off #1877, I think.

I didn't do freezing for `map`, because I couldn't find an example where it made a difference. Did I miss something?

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -3187,7 +3187,7 @@
           # => error
           ```
         "%%
-      = fun field content r => %record/insert% field r content,
+      = fun field content r => %record/insert% field (%record/freeze% r) content,
 
     insert_with_opts
       : forall a. String -> a -> { _ : a } -> { _ : a }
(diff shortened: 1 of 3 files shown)


yannham (approved):
`map` doesn't need freezing indeed (or, put differently, it's already been freezing its argument in practice, even if we didn't formulate it that way, because `map` needs to flush all pending lazy contracts or the result is quite quickly entirely wrong).

I initially delayed this follow-up in order to give a bit of thinking around the idea of having dictionary records freezing their argument (https://github.com/tweag/nickel/pull/2113#discussion_r1871003580). That being said, I think this PR is forward-compatible with this idea, shall we execute, as all those functions have a dictionary type in their signature (we would just revert the manual freezing that you just introduced). So I guess it's fine for now, as it fixes the original misbehavior and also unblocks you on the stdlib testing, I suppose.

jneem:
I like the idea of a sharper distinction between records and dictionaries. But yes, I'd like this to go in before we figure that out :smile:

PR #2132, opened 2024-12-24
pass thru feature `nix-experimental`

KiaraGrouwstra:
I found that the [`nickel` package in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ni/nickel/package.nix) had yet to add support for nickel's [`nix-experimental`](https://github.com/tweag/nickel/commit/91d63ed611f2f86e245b696b34b4cf09f637ee91) feature flag.

in an [in-progress attempt](https://github.com/NixOS/nixpkgs/compare/master...KiaraGrouwstra:nixpkgs:nickel-enable-nix-import?expand=1) to add this, i bumped into a few hurdles for this on the nickel side.
this PR tries to address these, on the Cargo side erring toward passing thru the feature for a few extra packages for consistency.

whereas i'm not sure this would be sufficient on the nickel side (yet having to get it to work at run-time), i felt maybe i could get a bit of feedback on this direction already.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -21,6 +21,7 @@ repl = ["nickel-lang-core/repl"]
 doc = ["nickel-lang-core/doc", "comrak"]
 format = ["nickel-lang-core/format", "dep:tempfile"]
 metrics = ["dep:metrics", "dep:metrics-util", "nickel-lang-core/metrics"]
+nix-experimental = ["nickel-lang-core/nix-experimental"]
 
 [dependencies]
 nickel-lang-core = { workspace = true, features = [ "markdown", "clap" ], default-features = false }
--- a/core/src/nix_ffi/cpp/nix.cc
+++ b/core/src/nix_ffi/cpp/nix.cc
@@ -11,10 +11,7 @@ using namespace nix;
 #include <nix/value-to-json.hh>
 #include <nix/command.hh>
 #include <nix/value.hh>
-// We will need this include when we update to latest Nix, but for now it's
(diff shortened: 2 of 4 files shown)


wasm-repl/Cargo.toml
```diff
@@ -14,4 +14,7 @@ nickel-lang-core = { workspace = true, default-features = false, features = ["re
 
 [lib]
 crate-type = ["cdylib", "rlib"]
-bench = false
\ No newline at end of file
+bench = false
+
+[features]
+nix-experimental = ["nickel-lang-core/nix-experimental"]
```
jneem, line 20: I understand the other ones, but will nix really build for wasm?

KiaraGrouwstra: [this seems to build, so i guess so?](https://github.com/NixOS/nixpkgs/compare/master...KiaraGrouwstra:nixpkgs:nickel-enable-nix-import-wasm?expand=1) 🤷‍♀️🤷‍♀️

jneem: Fair enough. It does fail, though, if I add `--target wasm32-unknown-unknown` and use that together with the "nix-experimental" feature.

KiaraGrouwstra: fair! i'll just remove it then

jneem (approved):

KiaraGrouwstra:
corresponding [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/384215)

PR #2134, opened 2024-12-28
[RFC007] Migration of the typechecker - part II

yannham:
WIP. This PR continues #2121. The goal is to scrap the old typecheck module, and update whatever needs to be updated in order to use the migrated `bytecode::typecheck` instead.

Doing so, I realize that NLS isn't the only obstacle: the migration of the source cache isn't entirely trivial, given the allocator approach of the new AST.


yannham:
This PR has been split in several parts, the last piece being https://github.com/tweag/nickel/pull/2182. It can be closed in its current state.

PR #2145, opened 2025-01-24
Dont check in the grammar, but generate it at release time

jneem:
This is an alternative to #2141, that avoids checking in a 22MB generated file. Here, we generate the grammar at `cargo publish` time. The main disadvantage of this one is that it's hard to test: although I've tested the individual commands, I haven't tested the release script as a whole because I didn't want to push changes to the `stable` branch.

I've simplified the build script so that it doesn't try to check the freshness of the generated grammar: it assumes that if there's a generated grammar already present then it's fresh. In principle the published crate could have an even simpler build process (with no build-dependency on lalrpop) but that would introduce more hard-to-test release stuff so I haven't gone that route.


scripts/release.sh
```diff
@@ -421,6 +421,18 @@ for crate in "${crates_to_publish[@]}"; do
...
+# have to deal with lalrpop being slow.
+# 
+# cargo check is the cheapest supported way to run just the build script
+# https://github.com/rust-lang/cargo/issues/7178
+# The output path is hard to predict (it contains a hash), so we find it
+# using a clean output directory and some globbing
+rm -rf __temp_target
+cargo check -p nickel-lang-core --target-dir=__temp_target
+cp __temp_target/debug/build/nickel-lang-core-*/out/parser/grammar.rs core/src/parser/grammar.rs
+git add core/src/parser/grammar.rs
```
yannham, line 434: should we remove the temporary directory? I guess its size might not be entirely negligible (and it will also pollute the git status).

core/build.rs
```diff
@@ -1,8 +1,33 @@
...
+        Ok(_) => {}
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            lalrpop::Configuration::new()
+                .use_cargo_dir_conventions()
+                .process_file("src/parser/grammar.lalrpop")
+                .unwrap();
+        }
+        Err(e) => panic!("{e}"),
+    }
+    println!("cargo:rerun-if-changed=src/parser/grammar.lalrpop");
```
yannham, line 30: If I understand correctly, we don't perform the sha3 check here, so we always assume the provided grammar file is up to date. It's not necessarily an issue, if we pack it up when packaging for crates.io, but I wonder if this extra check couldn't avoid some hard to track issues (maybe some other use cases around providing the generated parser will follow in the future, e.g. get a ready to work dev environment?)

scripts/release.sh
```diff
@@ -421,6 +421,18 @@ for crate in "${crates_to_publish[@]}"; do
...
 done
 
+# Generate the nickel grammar so that users of the published crate don't
+# have to deal with lalrpop being slow.
+# 
+# cargo check is the cheapest supported way to run just the build script
+# https://github.com/rust-lang/cargo/issues/7178
+# The output path is hard to predict (it contains a hash), so we find it
+# using a clean output directory and some globbing
+rm -rf __temp_target
```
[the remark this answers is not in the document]
jneem: Yes, that makes sense

scripts/release.sh
jneem: How do you feel about this now that we're using `mktemp`? I usually prefer to let my `/tmp/` get polluted because I have it configured to clean on boot (and not auto-removing things is sometimes useful for debugging).

core/build.rs
yannham: > Are you commenting specifically on the "cargo:rerun-if-changed" line, or the general philosophy of copying the generated grammar to the output directory if it's present without checking hashes?

I just used the last line but it was about the general philosophy, yeah. Maybe I'm just overthinking it, but I just have an impression that having a branching like this in the build script that can silently ignores the `.lalrpop` file has a very low but non null probability of biting someone, some day. However, you're right that right now for our release use-case it also makes sense - and it's simpler - to take a pre-generated parser if it's there, unconditionally.

Maybe a very cheap way to prevent any issue would be simply to output something during build if we're using the checked file? Like "Found a pre-generated LALRPOP parser, using it and ignoring the grammar" or something like that, so that it's visible in the log?

scripts/release.sh
yannham: It might be worse with `mktemp` because if we run 10 time the script now we have 10 times the space usage (since it'd be a different, random dir each time). Although it should get erased at reboot at least. I would still vouch for deleting the directory once we're done; we won't use it anymore, it's a one line in the release script and should prevent most problems. Maybe we need to do something like `rm -rf $temp_dir || true` to avoid making the whole script fail if for any reason we can't delete it, as it's not a big deal.

core/build.rs
jneem: Sure, I added a message in both branches.

scripts/release.sh
jneem: Sure, makes sense. I added it after the `cargo check` instead of putting it in `cleanup_actions`, so that if something fails exactly at that step then the output will be preserved.

yannham (approved):

Issue #2140, opened 2025-01-22
nickel startup overhead seems significant.


suimong:
**Is your feature request related to a problem? Please describe.**

This is not a "request" in any way, just a nice-to-have sort of issue.

I noticed that nickel's startup time is "slow": on my machine it took about 35ms to execute a minimal '{a = 1} | nickel eval'. For reference, a minimal python invocation `python -c '1 + 1'` took 15ms. 

In most circumstances this doesn't matter. But I noticed the lag because I was using nickel in a nushell module to validate a bunch of json files during import. Nickel binary is called 8 times, which costs ~400ms, leading to the noticeable lag for the nushell module import. 

I wonder if there are low hanging fruits that could improve the startup time. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

jneem:
I ran into something similar when benchmarking package manifest evaluation. It turns out that there's a substantial (~20ms on my machine) amount of time spent parsing the standard library. We embed the standard library source in the nickel binary, but parse it on every evaluation.

Possible mitigations:
- store some intermediate representation that's faster to load
- allow multiple evaluations on a single run. You'd still pay the 35ms, but not the 400ms from running 8 times.

suimong:
> It turns out that there's a substantial (~20ms on my machine) amount of time spent parsing the standard library.

That explains. Thanks!

> store some intermediate representation that's faster to load

I imagine if the parsed stdlib is somehow serializable (in a straightforward way a la serde) then this approach would be quite feasible? Otherwise, it sounds like lots of work for relatively little gain.

> allow multiple evaluations on a single run. You'd still pay the 35ms, but not the 400ms from running 8 times.

This is an interesting approach, and paying the startup overhead once is definitely acceptable, since that's the minimum cost no matter what. Though the CLI interface might be the tricky part.  On top of my head I can't think of a CLI program that enables such "multiple input, multiple output" paradigm. Also in my case, there are calls to both `nickel eval` and `nickel export` among the 8 times nickel gets called, so if we were to allow multiple evaluations per single run then it should also take different subcommands into consideration.

"lazy" parsing of stdlib is another approach, just to add to the ideas list.

yannham:
Given RFC007 is in the process of being implemented, I would be inclined to wait to see the result first. In particular once we have a bytecode VM, there's a natural, serializable format to store the stdlib already processed: bytecode.

[...]

PR #2146, opened 2025-01-24
Package management CLI, part 1

jneem:
This is the first part of the package manager CLI, hidden behind a "experimental-package" feature flag. It initially supports only git and path dependencies.

- [x] Allow the git repo realization to check for up-to-date git repos using an existing lock file.
- [x] Thoroughly test the manifest deserialization
- [x] Test the is-lock-file-up-to-date logic
- [x] Integration tests involving the cli
- [ ] Progress messages during git operations


jneem:
I haven't added a progress bar to the git downloads, but I think this is ready for a first look.

This does auto-updating of the lock-file, like cargo but unlike npm. I haven't yet implemented re-using the standard library across evaluation of multiple manifests, but I think it shouldn't be too hard (certainly easier than the general case of #2140 where you want to re-use work across nickel invocations).

cli/src/input.rs
```diff
@@ -2,6 +2,11 @@ use std::path::PathBuf;
 
 use nickel_lang_core::{eval::cache::lazy::CBNCache, program::Program};
 
+#[cfg(feature = "package-experimental")]
+use nickel_lang_package::config::Config as PackageConfig;
+#[cfg(feature = "package-experimental")]
+use nickel_lang_package::ManifestFile;
```
yannham, line 8:
```suggestion
#[cfg(feature = "package-experimental")]
use nickel_lang_package::{config::Config as PackageConfig, ManifestFile};
```

cli/src/input.rs
```diff
@@ -66,6 +86,40 @@ impl<C: clap::Args + Customize> Prepare for InputOptions<C> {
             program.add_import_paths(nickel_path.split(':'));
         }
 
+        #[cfg(feature = "package-experimental")]
+        {
+            let manifest_path = self.manifest_path.clone().or_else(|| {
+                // If the manifest path isn't given, where should we start looking
+                // for it? If there's only one file, we start with its parent directory.
```
yannham, line 93: When you say the parent directory, this means the directory where this file is located, not one the one above that, right?

cli/src/input.rs
```diff
@@ -66,6 +86,40 @@ impl<C: clap::Args + Customize> Prepare for InputOptions<C> {
...
+                // If there are no files, we take the current directory.
+                //
+                // If there are multiple files, finding a good heuristic
+                // is harder. For now, we take the parent directory if it's
+                // unique and otherwise we require an explicit manifest
+                // path.
```
yannham, line 99: One possibility would be to look for the common ancestor. But I can see how it might be wrong in some cases (if the files are far apart), so playing safe by explicitly requiring a manifest path sounds good :+1:

cli/src/package.rs
```diff
@@ -0,0 +1,156 @@
...
+#[derive(clap::Subcommand, Debug)]
+pub enum Command {
+    /// Generate a lock file for a package.
+    ///
+    /// The lock file contains exact information about a package's dependencies,
+    /// including recursive dependencies.
+    Lock {
+        /// The path at which to write the lock file.
+        ///
+        /// Defaults to the filename `package.ncl.lock` in the same directory
```
yannham, line 26: It's a detail, but I think usually package managers don't keep the `.ncl` equivalent in the lock file: `flake.nix` -> `flake.lock`, `Cargo.toml` -> `Cargo.lock`, `Pipefile` -> `Pipefile.lock`, `foo_bar.gemspec` -> `Gemfile.lock`, etc.

cli/src/package.rs
```diff
@@ -0,0 +1,156 @@
...
+
+                manifest.snapshot_dependencies(config)?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+fn print_package_map(map: &PackageMap) {
```
yannham, line 105: Would that make sense to make this a `Display` instance instead, or are there other considerations against this idea?

cli/tests/snapshot/main.rs
```diff
@@ -43,10 +43,24 @@ fn check_snapshots(path: &str) {
...
     // customize mode's delimiter `--`.
-    let invocation = NickelInvocation::new()
-        .file(&file)
-        .args(annotation.command)
-        .extra_args(annotation.extra_args);
+    let invocation = if annotation.is_package_manifest {
+        #[cfg(not(feature = "package-experimental"))]
+        return;
+
+        #[allow(unreachable_code)]
```
yannham, line 50: This annotation is a bit surprising, semantically speaking. I wonder if using something like `#[cfg(feature = "package-experimental")]` would better convey the meaning.

core/stdlib/std.ncl
```diff
@@ -3252,6 +3252,31 @@
...
+            # =>  "0c0a82aa4a05cd84ba089bdba2e6a1048058f41b"
+            ```
+          "%
+          =
+            let hex_chars = std.string.is_match "^[a-f0-9]*$" in
+            let value_validator =
+              std.contract.from_validator (fun s =>
+                if (std.string.length s) != 40 then
+                  let len = std.string.length s |> std.to_string in
+                  'Error { message = m%"expected 40 characters, got %{len}"%, notes = ["use the full commit id; short ids are not supported"] }
```
yannham, line 3272:
```suggestion
                  'Error { message = "expected 40 characters, got %{len}", notes = ["use the full commit id; short ids are not supported"] }
```

package/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "nickel-lang-package"
+description = "The Nickel Package Manager (npm)"
+version = "0.1.0"
```
yannham, line 4: It's  not very important for this PR, but I wonder if we shouldn't make it target either `nickel-lang-core` or the Nickel version (`nickel-lang-cli`) (probably the latter, since it does provide new CLI commands), to avoid the proliferation of uncorrelated versions. On the other hand, this removes the possibility of updating those crates independently.

package/src/error.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl std::error::Error for Error {}
+
+impl std::fmt::Debug for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(self, f)
+    }
+}
+
+impl std::fmt::Display for Error {
```
yannham, line 66: I haven't tried, but I suppose this means that package management errors are displayed in a way that doesn't follow the other errors look-and-feel? It doesn't have to be done now, but would implementing `IntoDiagnostics` instead and re-use the Nickel error reporting infrastructure work as well?

package/src/error.rs
```diff
@@ -0,0 +1,188 @@
...
+                )
+            }
+            Error::Git(e) => e.fmt(f),
+            Error::InvalidUrl { msg, .. } => {
+                write!(f, "{}", msg)
+            }
+            Error::InternalManifestError { path, msg } => {
+                write!(
+                    f,
+                    "internal error reading the manifest at {}; this is a bug in nickel: {msg}",
```
yannham, line 107: TODO: reuse the same internal error message that is used in Nickel-lang core. I think there is also a note saying to report the error to the issue tracker.

package/src/error.rs
```diff
@@ -0,0 +1,188 @@
...
+                    path.display()
+                )
+            }
+            Error::NoPackageRoot { path } => write!(
+                f,
+                "tried to import a relative path ({}), but we have no reference",
```
yannham, line 113: I'm not sure I would understand this error message without context.

package/src/error.rs
```diff
@@ -0,0 +1,188 @@
...
+            Error::NoProjectDir => {
+                write!(
+                    f,
+                    "failed to find a location for the nickel cache directory"
+                )
+            }
+            Error::LockFileDeserialization { path, error } => {
+                write!(f, "lock file {} is invalid: {error}", path.display())
+            }
+            Error::NoManifestParent => {
```
yannham, line 127: What does this  error correspond to? Is that possible for a regular file to have no parent?

yannham (commented):
Just a partial review for now. Note to self: next item on the stack is `package/src/lib.rs`.

cli/src/package.rs
jneem: Right, but I was worried about colliding with npm's (the other npm :wink:) `package.lock`. Maybe it's better to avoid a collision by having a more nickel-specific name than "package", though...

package/src/error.rs
yannham: Addendum: I think "reference" wasn't clear initially. I find "root" better, or "base" maybe, for what it's worth

cli/src/input.rs
jneem: Right. I updated the comment to be clearer.

cli/src/package.rs
jneem: Not that I know of, I've made it `Display` instead.

package/Cargo.toml
jneem: I think it makes some sense to have the version in sync with `nickel-lang-cli`. It does require a little bit of manual intervention to keep them in sync: I can put `version.workspace = true` in `package/Cargo.toml`, but I need to repeat the "1.9.0" in the workspace's dependencies list.

package/src/error.rs
jneem: I don't have a strong opinion about this, but I did it this way because most of the package errors don't use the reporting machinery. In `cli/src/error.rs` you'll see that the `Display` implementation is used in conjunction with `report_with_msg` for most variants, and `ManifestEval` gets special handling. So the errors should have the same look-and-feel as the others.

package/src/error.rs
jneem: I've gone with root for now. I have a TODO to enrich these sorts of errors with dependency chains, which I think will make them easier to understand.

package/src/error.rs
jneem: I think only empty paths have no parent, so I've dropped this error and just added a panic in the manifest constructor.

package/src/error.rs
yannham: Ah, if it gets reported with the usual machinery, it's fine then. No need to import that here if you don't use the notes, labels or the like.

cli/tests/snapshot/main.rs
yannham: @jneem now you should be able to get rid of the `[allow(unreachable_code]]`, don't you?

cli/src/package.rs
yannham: I see. Yes, maybe having a more specific name than "package" could be helpful in general, if only there are projects with both an NPM manifest an a Nickel manifest (or even imagine that you handle everything with Nickel, so you would like to write your "package.json" as a "package.ncl" instead and generate it).

package/Cargo.toml
jneem: Right, but none of the other non-"independent" crates are a dependency of other crates in the workspace. Naively adding `package` to the list of crates gives

```
 -- Updating cross-dependencies...
  -- Patching cross-dependency lsp-harness in ./Cargo.toml to version 0.1.0
  -- Patching cross-dependency nickel-lang-core in ./Cargo.toml to version 0.11.0
  -- Patching cross-dependency nickel-lang-package in ./Cargo.toml to version {
  "workspace": true
}
sed: -e expression #1, char 87: unterminated `s' command
++ Unexpected exit. Cleaning up...
```

(Also, I guess we should add `git` and `vector`)

package/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn deserialize<'de, D: serde::Deserializer<'de>>(de: D) -> Result<gix::Url, D::Error> {
+        let s = String::deserialize(de)?;
+        gix::Url::try_from(s).map_err(|e| D::Error::custom(e.to_string()))
+    }
+}
+
+/// A precise package version, in a format suitable for putting into a lockfile.
+#[serde_with::serde_as]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, PartialOrd, Ord)]
+pub enum Precise {
```
yannham, line 109: Nitpick, feel free to ignore: `Precise` is quite general. Is it a `PreciseDep` ? `PreciseGitDep` ?

package/src/lock.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub name: String,
+    pub id: u32,
+}
+
+impl std::fmt::Display for EntryName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.id == 0 {
+            write!(f, "{}", self.name)
+        } else {
+            write!(f, "{} {}", self.name, self.id)
```
yannham, line 45: Naive question: is it possible that a package name has a space in the name?

package/src/lock.rs
```diff
@@ -0,0 +1,267 @@
...
+impl LockFileNamer {
+    fn name(&mut self, name: &str, pkg: &Precise) -> EntryName {
+        if let Some((old_name, old_num)) = self.assigned.get(pkg) {
+            EntryName {
+                name: old_name.to_owned(),
+                id: *old_num,
+            }
+        } else {
+            let count = self.counts.get(name).copied().map_or(0u32, |i| i + 1);
+            self.counts.insert(name.to_owned(), count);
```
yannham, line 83: I suspect this can be done in a one liner (and with only one hash lookup) with the `Entry` API.

package/src/lock.rs
```diff
@@ -0,0 +1,267 @@
...
+            Precise::Git { url, id, path } => LockPrecise::Git { url, id, path },
+            Precise::Path { .. } => LockPrecise::Path,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct LockFileDep {
+    pub name: EntryName,
+
```
yannham, line 252:
```suggestion
```

package/src/lock.rs
```diff
@@ -0,0 +1,267 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub spec: Option<GitDependency>,
+}
+
+/// The dependencies of a single package.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct LockFileEntry {
+    pub precise: LockPrecise,
+
```
yannham, line 265:
```suggestion
```

package/src/manifest.rs
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+}
+
+/// A package manifest file.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ManifestFile {
+    // The directory containing the manifest file.
+    //
+    // This is used to resolve relative path dependencies.
```
yannham, line 110:
```suggestion
    /// The directory containing the manifest file.
    ///
    /// This is used to resolve relative path dependencies.
```

package/src/manifest.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        let manifest_term = prog.eval_full().map_err(|e| Error::ManifestEval {
+            package: None,
+            files: prog.files(),
+            error: e,
+        })?;
+        ManifestFile::from_term(path, &manifest_term)
+    }
+
+    /// Returns the location of the
```
yannham, line 164: Missing end of sentence

package/src/snapshot.rs
```diff
@@ -0,0 +1,326 @@
...
+    config::Config,
+    error::{Error, IoResultExt as _},
+    lock::{LockFile, LockFileDep, LockPrecise},
+    repo_root, Dependency, GitDependency, ManifestFile, Precise,
+};
+
+/// Collects and locks all the path and git dependencies in the dependency tree.
+///
+/// The manifest file(s) can ask for git dependencies without being too specific:
+/// it can as for a branch, for example, without knowing which commit is the head
```
yannham, line 24:
```suggestion
/// it can ask for a branch, for example, without knowing which commit is the head
```

package/src/snapshot.rs
```diff
@@ -0,0 +1,326 @@
...
+/// of the branch. The "snapshot" process fetches all the git dependencies in
+/// the dependency tree and figures out exactly which commit should be used for
+/// each one.
+///
+/// Snapshotting is different from version resolution in that we don't consider
+/// multiple possible versions of any dependency: we just fetch git branches
+/// or tags or whatever, and see what version we get.
+#[derive(Clone, Debug)]
+pub struct Snapshot {
+    /// The nickel package configured used for building this snapshot.
```
yannham, line 34:
```suggestion
    /// The Nickel package configured used for building this snapshot.
```

package/src/version.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/// Our most-widely-used version type.
+///
+/// This drops the build metadata part (which we allow during parsing but
+/// ignore for all version-resolution purposes).
+///
+/// Possible optimizations:
+/// - shrink the numbers to `u32`
+/// - intern the prerelease tag. This needs to be done in a way that preserves
```
yannham, line 29: Do you think it's worth the trouble, given that `pre` is probably rather small? Another solution is to use `SmolString` or the like.

package/tests/integration.rs
```diff
@@ -0,0 +1,197 @@
...
+            );
+        };
+
+        let run_in_dir = |cmd: &mut Command| {
+            run(cmd.current_dir(&dir_path));
+        };
+
+        // The rust stdlib doesn't have anything for recursively copying a directory. There are
+        // some crates for that, but it's easier just to shell out.
+        run(Command::new("cp")
```
yannham, line 115: Wouldn't that break the CI on Windows?

yannham (approved):
I can't pretend I followed all the details, but it looks good after a pass. The extensive tests are also great and reassuring. I just wonder if all the tests using `git` commands explicitly wouldn't break tests on Windows ?

package/src/lib.rs
jneem: I went with `PrecisePkg`. (I think "package" is more than "dependency" here because this can be used to precisely specify a package in general, not only when we're talking about a package that's a dependency of another package.)

package/src/lock.rs
jneem: rustfmt disagrees :smile: But sure, I think the one-statement-over-five-lines version is clearer.

package/src/version.rs
jneem: Probably not worth it, I just observed that the `semver` crate puts some effort into an optimized layout so I just wanted to record the possibility. Actually, I was more concerned about `pre` bloating the size of `SemVer` than about the size of the string: `Ident` is only 4 bytes, vs 24 for `String`.

package/src/lock.rs
jneem: You can't use such a package because the import syntax for packages requires the package name to be an identifier ("identifier" in the sense of the nickel lexer). But it's a good point, `nickel_lang_package` should check this.

package/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+/// A source includes the place to fetch a package from (e.g. git or a registry),
+/// along with possibly some narrowing-down of the allowed versions (e.g. a range
+/// of versions, or a git commit id).
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Deserialize)]
+pub enum Dependency {
+    Git(GitDependency),
+    Path { path: PathBuf },
```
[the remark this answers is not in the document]
jneem: Yeah, I think what happened is that I started with named fields in both variants, and then decided to extract the git one to a separate struct. I'll make them both unnamed, and I've removed the `Deserialize` because apparently it isn't needed.

cli/src/package.rs
jneem: Ok, I realized while fiddling with the website that actually npm uses "package-lock.json", not "package.lock". So there would be no conflict to go with "package.lock", but maybe a more distinctive name is better anyway. What do you think about `Nickel.ncl`/`Nickel.lock`? Or `Nickelfile.ncl`/`Nickelfile.lock`?

package/src/lock.rs
jneem: I've added this check to `std.package.Manifest`

jneem:
Ok, I think the main outstanding thing is to settle on a name for the manifest and lock files.

I'm pretty sure the windows CI is working ok with `git` and `cp` -- the CI already caught various small differences between git on windows and linux (e.g. it doesn't auto-detect the username on windows, so you need to provide it), so I'm pretty sure that it's running and passing in CI.

core/stdlib/std.ncl
```diff
@@ -2969,8 +2969,23 @@
         partial_semver_re = m%"^(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?$"%,
         # An exact version constraint. This one is required to have minor and patch versions, and it's allowed to have a prerelease.
         semver_equals_req_re = m%"^=(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?$"%,
+        # The regex for nickel identifiers.
+        ident_re = "^_*[a-zA-Z][_a-zA-Z0-9-']*$",
+      in
+      let
+        semver_req_re = "(%{partial_semver_re})|(%{semver_equals_req_re})",
+        IdentKeys =
```
yannham, line 2977: Should we expose `IdentKeys` as well? While it feels like a small self-explanatory contract, exposing it makes it discoverable, re-usable, and queri-able (if we add documentation). Though it also makes it part of the backward-compatibility. So I don't have a strong opinion either way.

yannham:
> I'm pretty sure the windows CI is working ok with git and cp -- the CI already caught various small differences between git on windows and linux (e.g. it doesn't auto-detect the username on windows, so you need to provide it), so I'm pretty sure that it's running and passing in CI.

I see, this makes our life simpler (it was just a random thought - I haven't used Windows in ages so I wondered if the CLI argument syntax was different, I remember using `/`).

As per the bikeshedding for the file name:

- `nickel.ncl`/`nickel.lock`: clear,  no risk of conflict with other tools, small name. The only disadvantage is that we tied it to the  name of the language, so it might not be 100% clear for a newcomer that this is related to package management? For example if we ever implement file-based configuration for the interpreter itself, maybe it could also be `nickel.ncl`?
- `nickelfile.ncl` / `nickelfile.lock`: I'm not sure why Pip and Gem adds the `file` prefix, because it doesn't seem to add much meaning.

Maybe something with both `nickel` and another word that indicates it's related  to package management would work as well? What do yo think about e.g. `nickel-pkg{.ncl,.lock}`, `nickel-manifest{.ncl,.lock}` or something along those lines? Also, we can post-pone the bikeshedding and merge this first: it's an experimental feature, and 1.10 has just been released, so we have time.

core/stdlib/std.ncl
jneem: I've been playing with json schema lately, and it seems pretty common to have a dict with a regex for the key names. What if we exposed, say, `std.record.KeysMatch` which could get used like

```nickel
{
  foo = 1
} | std.record.KeysMatch "^[a-z]+*"
```

(basically, I'm worried that `IdentKeys` is a bit too specific to be worth making public)

jneem:
`nickel-pkg.ncl` sounds good to me, but I wonder if we should capitalize it? It seems to be common, and at least [cargo](https://doc.rust-lang.org/cargo/faq.html?highlight=cross-compile#why-cargotoml) explicitly explains their rationale.

yannham:
I initially didn't like the inconsistency of upper casing this specific file but I think their rationale makes sense. Let's start with `Nickel-pkg.%{extension}` and review later if needed?

core/stdlib/std.ncl
yannham: Yes, I even think CUE provide built-in syntax for that. Nitpick: should we use `FieldMatches` instead to follow Nickel's jargon? Although the issue with field is that it designates both the content and the name. Could use `FieldNameMatches`, but it's longer. That being said, we have `std.record.fields` which return field names, so `FieldMatches` might be reasonable.

Fair for `IdentKeys`, we can keep it local for now.

PR #2148, opened 2025-01-31
bump flake.lock

jneem:
This bumps nixpkgs and the git hooks.

The automatic flake updater is stuck because bumping nix breaks the build. At some point we should really switch to the stable C API...


jneem:
Meh. The tests passed on my machine...

PR #2149, opened 2025-02-03
Try again to bump the flake

jneem:
nixos-unstable now has a nix that seems to be new enough for us, so switch to that instead of using nix straight from the flake.

There was also some changes to wasm-bindgen in nixpkgs.

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -48,7 +48,8 @@ fn main() {
 
         cxx_build::bridge("src/nix_ffi/mod.rs")
             .file("src/nix_ffi/cpp/nix.cc")
-            .flag_if_supported("-std=c++20")
+            .std("c++20")
+            .cpp(true)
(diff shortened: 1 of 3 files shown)


jneem:
Ok, this one is failing because the C++ compiler that the `cxx_build` crate is pulling in on MacOS doesn't seem to like the operator name "<=>".

I looked into using the nix C api, and it seems like it will be possible but take some work: it doesn't provide a simple "convert this nix to json" function, so we'd have to write our own by recursing over the evaluated values. Can we just drop the nix-experimental feature for now? We could re-do it using the C api if anyone needs it. @yannham

yannham:
It looks like the MacOS compiler isn't following the `.flag_if_supported("-std=c++20")` for some reason, given that this is a construct that has been introduced in C++20 ([three way comparison](https://en.cppreference.com/w/cpp/language/operator_comparison#Three-way_comparison)).

Let me try something quick

yannham (approved):
Seems to work now!

PR #2154, opened 2025-02-06
Fix nls crash when typechecking parse errors

### LINKED ISSUE #2153: crash in nls
The following input crashes nls, with a panic in the pretty-printer.

```nickel
{
  multipleOf
    : Number -> Dyn -> [| 'Ok, 'Error { message | String, notes] } |]
    = 1
}
```

There's a syntax error, but that isn't the point: it shouldn't cause a crash.

jneem:
Fixes #2153.

Nls attempts to typecheck the input even if parsing fails (as long as the failure is non-fatal). The problem is that the pretty-printing code (for diagnosing a type error) assumes that all dynamic fields of all records are `Term::StrChunks`. This may be true for successful parses, but if the parse fails then you can also have a `Term::ParseError` as a dynamic field.

### DIFF
--- a/lsp/nls/tests/inputs/no-crash-on-pretty-print.ncl
+++ b/lsp/nls/tests/inputs/no-crash-on-pretty-print.ncl
@@ -1,0 +1,7 @@
+### /main.ncl
+{
+  multipleOf
+    : Number -> Dyn -> [| 'Ok, 'Error { message | String, notes] } |]
+    = 1
+}
+### diagnostic = ["file:///main.ncl"]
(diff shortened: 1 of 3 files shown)


lsp/nls/tests/inputs/no-crash-on-pretty-print.ncl
```diff
@@ -0,0 +1,7 @@
+### /main.ncl
+{
+  multipleOf
```
yannham, line 3: Snake_case police here :police_officer: 

```suggestion
  multiple_of
```

yannham (approved):

PR #2160, opened 2025-02-10
[Infra] Get rid of the Terraform/AWS, fix missing arm64 and docker image artifacts

### LINKED ISSUE #2156: The release build for ARM64 is broken
**Describe the bug**

For the last 3 releases or so, the workflow building the ARM64 executable has failed, unable to get an Amazon runner to complete. This means that the releases don't have an ARM64 Linux executable,  and also the docker image isn't available or published, since it's multi-platform and requires the ARM64 and x86 jobs to succeed.

**To Reproduce**
Make a new release, I guess ? 🤷 

**Expected behavior**

I expect the workflow to succeed and to provide Linux ARM64 and Docker artifacts.

yannham:
Fixes #2156.

We previously used a bespoke AWS EC2 instance in order to build statically linked ARM64 linux release artifacts. However, it hasn't worked in the last 3 releases, and GitHub now provides ARM64 runners out of the box. It's much simpler to just use them, as we do for X86_64, and scrap the whole AWS part.


yannham:
I ran the workflows manually and they've been going on for hours, which is a good sign. I will consider this to work as far as we can tell if they fail on the final step when they won't found the release tag.

I'll also drop the last commit, which is just for testing purpose (avoid the failure of the workflows at the very beginning)

.github/workflows/release-artifacts.yaml
```diff
@@ -13,85 +13,20 @@ permissions:
...
             architecture: arm64
     runs-on: ${{ matrix.os.runs-on }}
-    needs:
-      - start-runner
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
+        # with:
+          # ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
```
jneem, line 29: Reminder to put this back

.github/workflows/release-artifacts.yaml
```diff
@@ -149,15 +84,15 @@ jobs:
...
 
   release-artifacts-macos:
     name: "Build MacOS Nickel binaries"
     runs-on: [self-hosted, macOS, ARM64]
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
+        # with:
+          # ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
```
jneem, line 95: Reminder to put this back

jneem, line 95: Reminder to put this back

.github/workflows/release-artifacts.yaml
```diff
@@ -178,8 +113,8 @@ jobs:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
+        # with:
+          # ref: ${{ github.event_name == 'release' && '' || github.event.inputs.release_tag }}
```
jneem, line 117: Reminder to put this back

jneem:
```
2025-02-10T17:43:10.3699979Z +++ command cargo build --release --message-format json-render-diagnostics --frozen --offline --package nickel-lang-cli
2025-02-10T17:43:14.8108768Z    Compiling cfg-if v1.0.0
2025-02-10T17:43:14.8109096Z    Compiling serde v1.0.197
2025-02-10T17:43:14.8129641Z    Compiling libc v0.2.162
2025-02-10T17:43:14.8160707Z    Compiling once_cell v1.19.0
2025-02-10T17:43:15.6847272Z error: rustc interrupted by SIGSEGV, printing backtrace
2025-02-10T17:43:15.6850544Z 
2025-02-10T17:43:15.6851345Z /nix/store/bkscnhmwafli28sxzjhyx5yv5q57b3w5-rust-minimal-1.84.1/lib/librustc_driver-6483a246a57da5ec.so(+0xbf8430) [0xffffee9f8430]
2025-02-10T17:43:15.6852125Z linux-vdso.so.1(__kernel_rt_sigreturn+0x0) [0xfffff7ffb7e0]
2025-02-10T17:43:15.6852414Z 
2025-02-10T17:43:15.6852686Z note: we would appreciate a report at https://github.com/rust-lang/rust
2025-02-10T17:43:15.6853352Z help: you can increase rustc's stack size by setting RUST_MIN_STACK=16777216
```

:thinking: 

Maybe this is unrelated, but it seems like the long build time is because we're trying to rebuild a bunch of nix arm stuff. Maybe switching from nixos-unstable to nixos-24.11 would make us hit the cache more?

yannham:
> Maybe this is unrelated, but it seems like the long build time is because we're trying to rebuild a bunch of nix arm stuff. Maybe switching from nixos-unstable to nixos-24.11 would make us hit the cache more?

Officially, Hydra provides ARM64 binaries for the stable and the unstable channel. But if it's moving fast I guess you can miss some of them in unstable? In general this is a good question to ask. I think we still need unstable for some dependencies, typically Topiary - if we want to take it from Nixpkgs instead of directly from the flake. We could probably use stable for the Rust toolchain but I wonder if that would be a win, as having two different nixpkgs versions comes with a risk of losing sharing of core libraries and binaries.

Or take everything "cutting-edge" from github directly and use an `follows` to use the latest stable version of NixOS, but that also comes with a risk (that it just doesn't work - I think a user had this problem with Topiary recently, which wasn't backward compatible with older Nixpkgs).

yannham:
https://github.com/rust-lang/rust/issues/135867: it seems it's a known bug and switching to ubuntu-22 images instead of 24 is a work-around for now.

jneem (approved):

PR #2166, opened 2025-02-12
Support pull diagnostics and use them for testing

jneem:
Push diagnostics (where the LSP decides when to publish, and sends notifications) are hard to test reliably, because the harness doesn't know when and how many diagnostics to expect.

This PR adds support for "pull" diagnostics, where there's a diagnostic request and the server responds to it. This simplifies the testing, and hopefully stops it from timing out at random.

This will probably conflict with the typechecker changes. I'm happy to wait and rebase.

### DIFF
--- a/lsp/nls/src/server.rs
+++ b/lsp/nls/src/server.rs
@@ -1,15 +1,20 @@
+use std::collections::HashMap;
+
 use anyhow::Result;
 use crossbeam::select;
 use log::{debug, trace, warn};
-use lsp_server::{Connection, ErrorCode, Message, Notification, RequestId, Response};
+use lsp_server::{
+    Connection, ErrorCode, Message, Notification, RequestId, Response, ResponseError,
+};
 use lsp_types::{
-    notification::Notification as _,
-    notification::{DidChangeTextDocument, DidOpenTextDocument},
+    notification::{DidChangeTextDocument, DidOpenTextDocument, Notification as _},
     request::{Request as RequestTrait, *},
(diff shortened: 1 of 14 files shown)


lsp/nls/src/server.rs
```diff
@@ -120,6 +138,9 @@ impl Server {
     pub fn run(&mut self) -> Result<()> {
         trace!("Running...");
         loop {
+            let never = crossbeam::channel::never();
+            let bg = self.background_jobs.receiver().unwrap_or(&never);
```
yannham, line 142: Nitpick: given that `never` is used exactly once and is not very long, I would inline it. Feel free to ignore.

yannham (commented):
If I understand correctly, this adds a new capability to NLS - send current document diagnostics - and use that in the test harness instead of looping over the answers.

The format is slightly less nice for snapshot test, but 1. it's not very important and 2. it's limited to eval diagnostics, if I understand correctly. Let's try this PR on top of one of the failing PR and see if it seems to solve the problem.

lsp/nls/src/server.rs
jneem: Right, but it needs somewhere to live. Putting it inline makes it complain about borrowing a temporary.

yannham: Ah, fair enough.

yannham (approved):
Let's try this way. It seems to work in https://github.com/tweag/nickel/pull/2173, although the failures aren't always deterministic. This still adds a new capability to NLS, and the price is minimal (beside added LoC, the snapshot for eval diagnostics are slightly less nice, but it's not user-facing)

PR #2170, opened 2025-02-14
[RFC007] Migration of the typechecker part II - focus on `bytecode::ast::typecheck`

yannham:
Sorry about the name, it is not very inspired. This PR is a split of the behemoth of #2134 that has become unmanageable, leading to unexpected and mandatory changes to many parts of Nickel. The current PR only keeps the changes to `bytecode::ast`, and doesn't actually get rid of the old typechecker, but instead only modifies `bytecode::ast::typecheck`.

The changes that landed here comes from actually testing the new typechecker, making all the tests of `nickel-lang-core` pass. It also includes various helpers additions to the `bytecode::ast` module.

### DIFF
--- a/core/src/bytecode/ast/typ.rs
+++ b/core/src/bytecode/ast/typ.rs
@@ -229,33 +229,25 @@ impl<'ast> RecordRows<'ast> {
     /// - path: `["a", "b"]`
     /// - result: `Some(Number)`
     pub fn find_path<'a>(&'a self, path: &[Ident]) -> Option<&'a RecordRow<'ast>> {
-        if path.is_empty() {
-            return None;
-        }
+        let mut curr_rrows = self;
 
-        // While going through the record rows, we use this helper for recursion instead of
-        // `find_path`, to avoid cloning a lot of intermediate rows, and rather only clone the
-        // final one to return.
-        fn find_path_ref<'a, 'ast>(
-            rrows: &'a RecordRows<'ast>,
-            path: &[Ident],
-        ) -> Option<&'a RecordRow<'ast>> {
-            let next = rrows.iter().find_map(|item| match item {
-                RecordRowsItem::Row(row) if row.id.ident() == path[0] => Some(row),
+        for (idx, id) in path.iter().enumerate() {
+            let next_rrows = curr_rrows.iter().find_map(|item| match item {
+                RecordRowsItem::Row(row) if row.id.ident() == *id => Some(row),
                 _ => None,
             });
 
-            if path.len() == 1 {
-                next
-            } else {
-                match next.map(|row| &row.typ.typ) {
-                    Some(TypeF::Record(rrows)) => find_path_ref(rrows, &path[1..]),
-                    _ => None,
-                }
+            if idx == path.len() - 1 {
+                return next_rrows;
+            }
+
+            match next_rrows.map(|row| &row.typ.typ) {
+                Some(TypeF::Record(rrows)) => curr_rrows = rrows,
+                _ => return None,
             }
         }
 
-        find_path_ref(self, path)
+        None
     }
 
     /// Find the row with the given identifier in the record type. Return `None` if there is no such


core/src/bytecode/typecheck/subtyping.rs
```diff
@@ -155,42 +161,67 @@ impl<'ast> SubsumedBy<'ast> for UnifRecordRows<'ast> {
...
-                    let urrows2 = UnifRecordRows::Concrete {
-                        rrows: rrows2,
-                        var_levels_data: levels2,
+                // We reverse the order of the arguments when compared to `unify`. It shouldn't
+                // really matter, but experience shows that the inferred type - here on the left -
+                // is often much bigger in a structural type system, than the expected type (coming
+                // from constraints or annotations).
+                //
+                // For example, when checking `array.fold_left` where `array` is `std.array`. The
+                // expected type (right hand side of `subsume`j) is `{ fold_left : ?a; ?b }`. The
```
jneem, line 187:
```suggestion
                // expected type (right hand side of `subsume`) is `{ fold_left : ?a; ?b }`. The
```

core/src/traverse.rs
```diff
@@ -94,15 +94,21 @@ pub trait TraverseAlloc<'ast, T>: Sized {
         F: FnMut(T) -> Result<T, E>;
 
     /// Same as [Traverse::traverse_ref], but takes an additional AST allocator.
+    ///
+    /// There is as small difference though: this function guarantees that the lifetime of the
+    /// references is bound to the lifetime of the AST allocator, which the signature in
+    /// [Traverse::traverse_ref] does not. This is useful e.g. in the LSP to extract references and
```
jneem, line 100: Ok, maybe this answers my question about about the lifetimes in `traverse_ref`, but doesn't this also contradict the changes to `Traverse::traverse_ref` (because after this PR they will be bound there too)?

core/src/bytecode/typecheck/mod.rs
```diff
@@ -97,6 +97,27 @@ use self::subtyping::SubsumedBy;
 /// The max depth parameter used to limit the work performed when inferring the type of the stdlib.
 const INFER_RECORD_MAX_DEPTH: u8 = 4;
 
+//TODO[RFC007]: once we switch to this typechecker, replace this trait definition by an import of
+//`cache::AstImportResolver`.
+pub trait AstImportResolver<'ast> {
+    /// Resolve an import.
+    ///
+    /// Read and store the content of an import, put it in the file cache (or get it from there if
+    /// it is cached), then parse it and return the corresponding term and file id.
```
jneem, line 106: This just returns the term, right? Not the file id

core/src/bytecode/typecheck/mod.rs
```diff
@@ -112,6 +133,7 @@ pub type TypeEnv<'ast> = Environment<Ident, UnifType<'ast>>;
 /// A term environment defined as a mapping from identifiers to a tuple of a term and an
 /// environment (i.e. a closure). Used to compute contract equality.
 #[derive(PartialEq, Clone, Debug)]
+//TODO: we should just store `&'ast Ast<'ast>` references here, instead of an owned version
```
jneem, line 136: Why? Isn't `Ast` pretty small?

core/src/bytecode/typecheck/mod.rs
```diff
@@ -1765,46 +1793,60 @@ impl<'ast> Walk<'ast> for FieldDef<'ast> {
...
-                    if let TypeF::Arrow(mut domain, _) = &c.typ {
+                let domains = annots.contracts().find_map(|c| {
+                    if let TypeF::Arrow(domain, mut codomain) = &c.typ {
                         let mut domains =
                             vec![UnifType::from_type((*domain).clone(), &ctxt.term_env)];
 
-                        while let TypeF::Arrow(next_domain, _) = &domain.typ {
-                            domains.push(UnifType::from_type((*domain).clone(), &ctxt.term_env));
-                            domain = next_domain;
+                        while let TypeF::Arrow(next_domain, next_codomain) = &codomain.typ {
```
jneem, line 1846: I don't follow the details here, but this looks like an actual change and not a refactor. Is this fixing some bug, and if so should there be a test for it?

jneem (approved):

core/src/bytecode/ast/builder.rs
```diff
@@ -56,6 +54,12 @@ pub struct Field<'ast, State> {
     state: State,
     path: StaticPath,
     metadata: FieldMetadata<'ast>,
+    /// We need to store the documentation separately, because the metadata only accepts a string
+    /// that has been allocated in the AST allocator. We could require to thread the allocator for
+    /// calls to [Self::doc] or [Self::some_doc], but it's more ergonomic to keep the phases of
+    /// building a field and finalizing it separate. We thus store the documentation here
+    /// temporarily.
+    doc: Option<String>,
```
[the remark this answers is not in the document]
yannham: Currently drowning in `'ast` lifetimes everywhere, I'm not too keen on adding a new one lightly :stuck_out_tongue: also, I think the interface of metadata requires an owned string, because we allocate it in the arena (and `alloc` takes owned data in order to move it into the arena indeed). I'm not sure we can avoid it.

core/src/bytecode/typecheck/mod.rs
yannham: One answer is because we can :slightly_smiling_face: 

Storing a ref isn't much harder than the owned `Ast`. While `Ast` is significantly smaller than `Term`, it is bigger than `RichTerm` and significantly bigger than `&'ast Ast<'ast>`: contrary to the former, we don't put the content behind an indirection (`Rc<Term>` for `RichTerm`), so it stores both the position and the node inline, taking up 56 bytes of storage currently vs 8 for the pointer alone. The node is much smaller than before (than `Term`), but it's still a large enum. In some sense, the right equivalent for `RichTerm` or `SharedTerm` is `&'ast Ast<'ast>` rather than `Ast<'ast>`.

core/src/bytecode/typecheck/mod.rs
yannham: Yes, the previous code was wrong: if you have a multi-parameter function with a contract `Foo -> Bar -> Baz`, to get type hints for the parameter, you need to look into the domain of the  contract but then iterate on the _codomain_. But the code iterated on the domain. I believe this wasn't used in statically typed code enough to create a typechecking bug, but it's visible on multi-parameter function with contracts in the LSP, where the hovering is off. I'll try to reconstruct a simple example and add it to the tests, yeah.

yannham: In fact I'm dumb, I  can't add a test in this PR or it will be failing (since we still use the old typechecker). It'll need to land in a subsequent PR.

core/src/bytecode/typecheck/mod.rs
yannham: Right, it's the old documentation that has been copy-pasted. Updated in latest commits.

core/src/traverse.rs
yannham: I don't think there's been any change to `Traverse::traverse_ref` (which doesn't even have a lifetime parameter), only to `TraverseAlloc::traverse_ref`. The very first item you commented on is a specific implementation of `TraverseAlloc::traverse_ref` - maybe you though it was the definition of the trait `Traverse` instead?

To answer your first original question, I think the general idea is that `traverse_ref` can provide stronger guarantees - that the visited ref live as long as the allocator - provided that the root is in the allocator, which is a rather weak assumption (if we have an allocator, I guess it's simpler to allocate everything there and use `&'ast reference` instead of having owned `Ast<'ast>` for some ASTs, for some reason). And it was indeed needed in the LSP, where we can use `&'ast _` refs pretty much everywhere.

core/src/bytecode/ast/mod.rs
```diff
@@ -442,8 +439,8 @@ impl<'ast> TraverseAlloc<'ast, Ast<'ast>> for Ast<'ast> {
     }
 
     fn traverse_ref<S, U>(
-        &self,
-        f: &mut dyn FnMut(&Ast<'ast>, &S) -> TraverseControl<S, U>,
+        &'ast self,
+        f: &mut dyn FnMut(&'ast Ast<'ast>, &S) -> TraverseControl<S, U>,
```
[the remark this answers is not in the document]
yannham: Treated in a later comment (in the definition of the `TraverseAlloc` trait)

core/src/bytecode/ast/typ.rs
```diff
@@ -215,3 +218,160 @@ impl<'ast> TraverseAlloc<'ast, Type<'ast>> for RecordRows<'ast> {
...
+    /// - self: `{a : {b : Number }}`
+    /// - path: `["a", "b"]`
+    /// - result: `Some(Number)`
+    pub fn find_path<'a>(&'a self, path: &[Ident]) -> Option<&'a RecordRow<'ast>> {
+        if path.is_empty() {
+            return None;
+        }
+
+        // While going through the record rows, we use this helper for recursion instead of
+        // `find_path`, to avoid cloning a lot of intermediate rows, and rather only clone the
```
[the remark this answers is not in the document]
yannham: Right, the documentation was copied verbatim. Let me update it.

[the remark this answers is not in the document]
yannham: In the end I implemented an iterative version, which maybe is more direct in this case. I suspect I needed a recursive version for properly nesting lifetimes on the old AST, but here we can just take references everywhere.

PR #2175, opened 2025-02-19
Support for index packages

jneem:
This adds support for a package index, but requires that all the index packages have exact version constraints. As a result, we don't yet need to do version resolution.

The main points of this PR are
- to define the index format
- to decide when/how to refresh the index, and how to download index packages
- to add primitive index-manipulation tooling, so that we can start hosting (and using) index packages


Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "package",
     "utils",
     "wasm-repl",
-    "py-nickel",
+    "py-nickel", "flock",
```
yannham, line 12:
```suggestion
    "py-nickel",
    "flock",
```

cli/src/package.rs
```diff
@@ -32,6 +37,32 @@ pub enum Command {
...
+    /// 2. Clone your fork onto your local machine.
+    /// 3. The package you want to publish must be at the root of a git repository.
+    ///    `cd` into that repository (or supply `--manifest-path` in the next step).
+    /// 4. Run `nickel publish-to-local-index --index <directory-of-your-clone> --package-id github/you/your-package`
+    /// 5. You should see that your local machine's index was modified. Commit that modification.
+    /// 6. Push your package to the `you/your-package` repository on github. These
+    ///    names *must* match the package id in the index, and you must ensure that
+    ///    the version you push to github matches the SHA-1 hash in the index.
+    /// 7. Open a pull request to `github.com/nickel-lang/nickel-mind` to make your index
+    ///    modifications public.
```
yannham, line 55: I wonder if the whole description of the process really belongs to the CLI help message. What do you think about putting it somewhere (manual maybe?) and linking to the page here?

cli/src/package.rs
```diff
@@ -80,8 +111,10 @@ impl PackageCommand {
             Command::DebugResolution => {
                 let path = self.find_manifest()?;
                 let manifest = ManifestFile::from_path(path.clone())?;
-                let snap = manifest.snapshot_dependencies(config)?;
-                let package_map = snap.package_map(&manifest)?;
+                let snap = manifest.snapshot_dependencies(config.clone())?;
+                let index = PackageIndex::shared(config.clone())?;
+                let resolution = resolve::resolve(&manifest, snap, index, config)?;
+                let package_map = resolution.package_map(&manifest)?;
                 eprintln!("{package_map}");
```
yannham, line 118: left-over debug statement?

cli/src/package.rs
```diff
@@ -92,7 +125,23 @@ impl PackageCommand {
                     ..config
                 };
 
-                manifest.snapshot_dependencies(config)?;
+                manifest.snapshot_dependencies(config.clone())?;
```
yannham, line 128: Out of curiosity, why do we need to take a config as an owned value?

cli/src/package.rs
```diff
@@ -32,6 +37,32 @@ pub enum Command {
...
+    /// 3. The package you want to publish must be at the root of a git repository.
+    ///    `cd` into that repository (or supply `--manifest-path` in the next step).
+    /// 4. Run `nickel publish-to-local-index --index <directory-of-your-clone> --package-id github/you/your-package`
+    /// 5. You should see that your local machine's index was modified. Commit that modification.
+    /// 6. Push your package to the `you/your-package` repository on github. These
+    ///    names *must* match the package id in the index, and you must ensure that
+    ///    the version you push to github matches the SHA-1 hash in the index.
+    /// 7. Open a pull request to `github.com/nickel-lang/nickel-mind` to make your index
+    ///    modifications public.
+    PublishToLocalIndex {
```
yannham, line 56: Nitpick: this sounds a bit long for a command name.

flock/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
yannham, line 1: Should we add a description?

git/src/lib.rs
```diff
@@ -21,8 +20,8 @@ pub enum Error {
         path: std::path::PathBuf,
     },
 
-    #[error("target `{0}` not found")]
-    TargetNotFound(Target),
+    #[error("target `{1}` not found in `{0}`")]
+    TargetNotFound(Box<gix::Url>, Target),
```
yannham, line 24: Nitpick: should we use named fields/anonymous record, as is the case with e.g. `Self::Io`?

package/src/error.rs
```diff
@@ -133,7 +165,49 @@ impl std::fmt::Display for Error {
...
+                    f,
+                    "package {id} depends on path package {}, so it cannot be put in the index",
+                    path.display()
+                ),
+            },
+            Error::DuplicateIndexPackageVersion { id, version } => {
+                write!(f, "package {id}@{version} is already present in the index")
+            }
+            Error::PackageIndexSerialization { error, pkg } => {
+                write!(f, "error serializing package. Failed package {pkg:?}, caused by {error}\n{INTERNAL_ERROR_MSG}")
```
yannham, line 197: There is some duplication in the error message. Maybe it's a mix of two different versions?

package/src/index/lock.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Fetch an updated package index from github and save it to our cache directory.
+    pub fn download(&self) -> Result<(), Error> {
+        let config = &self.config;
+        let parent_dir = config.index_dir.parent().unwrap();
+        std::fs::create_dir_all(parent_dir).with_path(parent_dir)?;
+
+        eprint!("Fetching an updated package index...");
+        let tree_path = tempdir_in(parent_dir).with_path(parent_dir)?;
+        let _id = nickel_lang_git::fetch(&Spec::head(config.index_url.clone()), tree_path.path())?;
```
yannham, line 74:
```suggestion
        let _ = nickel_lang_git::fetch(&Spec::head(config.index_url.clone()), tree_path.path())?;
```

yannham (commented):
Partial review. Stopped just before `package/index/mod.rs`, will continue later.

cli/src/package.rs
jneem: Sure, I've added a manual page about package management in general and included this there.

cli/src/package.rs
jneem: No, printing the package map is the main job of this command. (It maybe be time to remove the command, though. It was mainly just to help debug things while implementing...)

cli/src/package.rs
jneem: Hm, I think maybe I can get rid of this one. In some other places it's just convenient to store copies of the config so that we can figure out where various files go.

cli/src/package.rs
jneem: Renamed to "publish".

flock/Cargo.toml
jneem: Yes!

package/src/index/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+            IdParseError::UnknownIndex { index } => write!(
+                f,
+                "unknown index `{index}`, the only valid value is `github`"
+            ),
+            IdParseError::InvalidId { id } => write!(f, "invalid identifier `{id}`"),
+        }
+    }
+}
+
+static ID_REGEX: LazyLock<Regex> =
```
yannham, line 376: I didn't know something like `LazyLock` existed in the standard library. I wonder if this deprecates the usage we make of `lazy_static` elsewhere in Nickel.

package/src/index/scrape.rs
```diff
@@ -0,0 +1,69 @@
...
+use super::{Id, Package, PreciseId};
+
+/// Fetch a package from the specified place, and figure out what its index
+/// entry should look like.
+///
+/// Question: should we allow index packages to come from a _subdirectory_ of
+/// a git repo, or should they always need to be at the root? Currently we only
+/// allow them at the root, the motivation being that if the git repo backing
+/// the index package only contains the package itself, then it will be a
+/// smaller download. On the other hand, we allow git dependencies to live in
```
yannham, line 22: I'm not sure to understand the motivation: if we allow subdirs, we can still allow being it at the root, without paying additional price, right? Or put differently: is allowing subdirs affecting negatively the optimistic workflow of "package is the whole repo"? Or is more around forcing people to do it to ensure the index is more performant? I think we can leave that for future work, but the mono repo case surely sounds plausible.

package/src/index/scrape.rs
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    Package::from_manifest_and_id(&manifest, &id)
+}
+
+// Check if the manifest file lives in a git directory, and check that the directory is clean.
+// Then find the id of head and construct a package from it
+pub fn read_from_manifest(id: &Id, manifest: &ManifestFile) -> Result<Package, Error> {
+    let repo = gix::open(&manifest.parent_dir)?;
+    if !matches!(repo.is_dirty(), Ok(false)) {
```
yannham, line 53: I'm not really sure it's better, so feel free to ignore. But I have a slight tendency to find it more basic that using a macro, although `match!` is pretty core as well.

```suggestion
    if let Ok(false) = repo.is_dirty() {
```

package/src/index/serialize.rs
```diff
@@ -0,0 +1,119 @@
...
+    index,
+    version::{SemVer, VersionReq},
+    IndexDependency,
+};
+
+use super::{Package, PreciseId};
+
+/// Defines the serialization format for `IndexDependency` in the package index.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct IndexDependencyFormat {
```
yannham, line 18: I suspect you have good reasons to not (de)serialize `IndexDependency` directly and use an indirection instead, but out of curiosity, what are they?

package/src/index/serialize.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl From<IdFormat> for index::Id {
+    fn from(i: IdFormat) -> Self {
+        match i {
+            IdFormat::Github { org, name } => index::Id::Github { org, name },
+        }
+    }
+}
```
yannham, line 64:
```suggestion
}

```

package/src/lib.rs
```diff
@@ -62,23 +62,76 @@ impl GitDependency {
...
+    pub fn as_unversioned(self) -> Option<UnversionedDependency> {
+        match self {
+            Dependency::Index(_i) => None,
+            Dependency::Git(g) => Some(UnversionedDependency::Git(g)),
+            Dependency::Path(p) => Some(UnversionedDependency::Path(p)),
+        }
+    }
+}
+
+/// The same as [`Dependency`], but only for the packages that have fixed, unresolvable, versions.
```
yannham, line 123: What does "unresolvable" means in this context? Just like we don't get to pick a different one, like rigid version constraints, or something like a precise git ref?

package/src/manifest.rs
```diff
@@ -54,13 +66,35 @@ struct ManifestFileFormat {
...
+    Index(IndexDependencyFormat),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Deserialize)]
+/// A dependency that comes from the global package index.
+///
+/// This is currently identical to `IndexDependency`, but we keep
+/// them separate so it's clear which things are part of our public
+/// serialization API.
+struct IndexDependencyFormat {
```
yannham, line 78: isn't this struct defined two times, here in `manifest` but also in `index::serialize`?

package/src/manifest.rs
```diff
@@ -230,15 +298,35 @@ impl ManifestFile {
...
+                    eprintln!("failed to refresh package index ({e}), trying with the old one");
+                    PackageIndex::shared(config.clone())?
+                }
+            }
+        } else {
+            // TODO: we could avoid instantiating the index (which triggers a download if it doesn't exist)
+            // if the dependency tree has no index packages.
+            PackageIndex::shared(config.clone())?
+        };
+        // TODO: maybe refresh the index, if the snapshot has index dependencies?
```
yannham, line 319: Aren't we doing that above?

package/src/manifest.rs
```diff
@@ -373,12 +477,18 @@ mod tests {
             // Invalid dependency names
             r#"{name = "foo", version = "1.0.0", minimal_nickel_version = "1.9.0", authors = [], dependencies = { "42" = 'Path "dep" }}"#.as_bytes(),
             r#"{name = "foo", version = "1.0.0", minimal_nickel_version = "1.9.0", authors = [], dependencies = { "has space" = 'Path "dep" }}"#.as_bytes(),
-            // TODO: add index dependencies, once they're supported
+
+            r#"{name = "foo", version = "1.0.0", minimal_nickel_version = "1.9.0", authors = [], dependencies = { dep = 'Index { package = "codeberg/example/example", version = "=1.2.0" }}}"#.as_bytes(),
+            // This should become successful once we support version resolution
+            r#"{name = "foo", version = "1.0.0", minimal_nickel_version = "1.9.0", authors = [], dependencies = { dep = 'Index { package = "github/example/example", version = "1.2.0" }}}"#.as_bytes(),
```
yannham, line 483: Is it ok to keep a failing test? Won't that make the CI red?

yannham (commented):
Second chunk: I stopped just before `resolve.rs`.

package/src/index/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+            id: id.clone(),
+            version: v,
+        };
+        let target_dir = precise.local_path(&self.cache.borrow().config);
+        self.ensure_downloaded_to(&package.id, &target_dir)
+    }
+
+    fn ensure_downloaded_to(&self, index_id: &PreciseId, target_dir: &Path) -> Result<(), Error> {
+        let PreciseId::Github { org, name, commit } = index_id;
+        let url = format!("https://github.com/{org}/{name}.git");
```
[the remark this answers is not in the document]
jneem: The manifest format ensures that `org` and `name` are nickel idents (in the lexer sense, not the "interned string" sense). I'll add a comment.

package/src/index/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+        match self {
+            Id::Github { org, name } => Ok(format!("https://github.com/{org}/{name}").try_into()?),
+        }
+    }
+}
+
+impl std::fmt::Display for Id {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Id::Github { org, name } => write!(f, "github/{org}/{name}"),
```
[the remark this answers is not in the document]
jneem: Sure, makes sense. I'll have to also change the validation in the stdlib.

package/src/index/mod.rs
jneem: Yes, I think so. It's only been in the stdlib since 1.80

package/src/index/mod.rs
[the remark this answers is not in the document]
jneem: Actually, I take it back: the `FromStr` implementation checks the names, but the manifest doesn't go through that. I'll fix this...

package/src/index/scrape.rs
jneem: It was mainly about avoiding the performance surprise of depending on an index package and then downloading a big monorepo.

But also, when a rust crate is published from a monorepo the packaging step extracts just the crate that you're publishing. So I thought we could have some similar tooling, where when you package from a monorepo the tooling extracts just the part you want to publish into a smaller repo. (Or maybe even into a compressed tarball)

package/src/lib.rs
jneem: It means that the resolver doesn't care about (or get to pick) the version. Or in other words, a path or git dependency.

Git deps don't need precise versions to be "unresolvable": if a git dep points at HEAD, then we just fetch HEAD and that's the package we depend on. The resolver won't consider whether it should take a different revision.

On the other hand, rigid version constraints on an index package are still "resolvable" in this sense, because their presence affects the resolver's decisions.

package/src/manifest.rs
jneem: Right, but one of them defines how to deserialize the manifest (going through the `RichTerm` deserializer) and the other defines how to serialize/deserialize the index through `serde_json`. The current differences are just cosmetic for now (for some reason I decided to have "github" lower case in the index, but it needs to be upper case for the manifest to match the `'Github` enum), but I think there's some value in allowing them to be different. For example, if we add support for other forges then an old nickel version reading `'Codeberg` from a manifest is allowed to error out immediately (because it won't know how to fetch that dependency), but an old nickel version shouldn't fail to deserialize an index that contains a package hosted on codeberg. (It can fail if the dependency tree contains such a package, though)

package/src/manifest.rs
jneem: Indeed, I guess I forgot to remove the TODO

package/src/manifest.rs
jneem: Ah, but this case is in the "manifests that should fail" test, so the test passes. By "become successful" I meant that it will get moved to the "manifests that should succeed" test.

package/src/index/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+        tmp.persist(&out_path)?;
+        Ok(())
+    }
+
+    /// Creates a temporary file that's in the same directory as the place that `id`'s
+    /// index file would go.
+    fn tmp_file(&self, id: &Id) -> NamedTempFile {
+        let path = self.path(id);
+        // unwrap: the `path` function always outputs a non-empty path
+        let parent = path.parent().unwrap();
```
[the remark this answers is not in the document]
jneem: Ah, this one didn't work because it borrows the temporary returned by `parent()`

package/src/index/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+            )
+            .with_path(lock_path)?;
+
+            // Now that we hold the download lock, check for existence again.
+            if target_dir.exists() {
+                eprintln!("Package {org}/{name}@{commit} already exists");
+                return Ok(());
+            }
+
+            eprintln!(
```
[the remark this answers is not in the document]
jneem: I added `info` and `warn` macros that just wrap `eprintln` for now.

package/src/index/serialize.rs
jneem: I may have gone overboard with this pattern, but I originally had a bunch of trouble caused by `ManifestFile` defining both the deserialization of manifests from the nickel terms and the representation for working with manifests internally. Separating out `ManifestFileFormat` made that all better, so I've been imitating the same pattern elsewhere.

cli/src/package.rs
jneem: Indeed, we now no longer store `Config` in `Snapshot`.

package/src/index/scrape.rs
yannham: I see. So the answer is that you'd prefer to do it properly, rather than blindly downloading the monorepo, and that is possible but just require some work?

package/src/lib.rs
yannham: I see. Maybe it would be clearer to mention it explicitly ("git ref or path"), because "version" is overloaded quite a bit in our context (I initially though version would only correspond to version bounds, which is why "unresolvable version" sounded a bit paradoxical).

package/src/index/scrape.rs
jneem: I mean, it's less work on the package consumption side (one less case to handle, no need to think about maybe deduplicating the monorepos...) but a bit more work on the package creation side.

package/src/resolve.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// The snapshot (of path and git packages) that was used to construct
+    /// this resolution. Note that path and git packages are not "resolved";
+    /// they have fixed versions. The snapshot is only used to collect index
+    /// dependencies of git and path packages.
+    pub snapshot: Snapshot,
+    pub index: PackageIndex<Shared>,
+    /// All the index packages in the dependency tree.
+    ///
+    /// Each package id can resolve to multiple versions, but those versions should all fall
+    /// into disjoint semantic-version buckets.
```
yannham, line 28: Is it true even when using precise version bounds? What happens if two dependencies define two different but compatible (in the same bucket) exact version bounds of a third dependency? Or maybe exact version bounds are just not recorded here in `index_packages`?

package/src/snapshot.rs
```diff
@@ -53,44 +50,51 @@ impl Snapshot {
...
     fn snapshot_recursive(
         &mut self,
+        config: &Config,
         root_path: &Path,
         lock: &LockFile,
         lock_entry: Option<&LockFileDep>,
         dep: &Dependency,
-        relative_to: Option<&PrecisePkg>,
+        // TODO: maybe this is simpler if we make this non-optional, and have the root denoted by PrecisePkg with an empty path?
+        // (Or maybe with an absolute path to the root package?)
```
yannham, line 88: Is that comment obsolete? It looks like you've settled down on the root being a `PrecisePkg` with an empty path indeed?

package/tests/up_to_date.rs
```diff
@@ -57,10 +58,48 @@ fn path_dep_up_to_date() {
...
+    let (lock, resolution) = manifest.lock(config.clone()).unwrap();
 
-    // Path deps are always up to date.
-    assert!(manifest.is_lock_file_up_to_date(&lock));
-    std::fs::write(dep_dir.path().join("Nickel-pkg.ncl"), "hi").unwrap();
-    assert!(manifest.is_lock_file_up_to_date(&lock));
+    assert!(manifest.is_lock_file_up_to_date(&resolution.snapshot, &lock));
+
+    // Changes to a path package that don't affect dependencies don't affect
+    // up-to-dateness.
```
yannham, line 66: This comment is a bit scrambled.

yannham (approved):

package/src/snapshot.rs
jneem: Indeed...

package/src/resolve.rs
jneem: In this PR, two different but compatible version bounds are allowed (because we basically don't check anything). The next PR will disallow it -- it will ensure when populating `index_packages` that there are no conflicts.

yannham: So if we have two different dependencies A and B that defined compatible but different exact version bounds on the C dependencies, say "1.1.0" and "1.2.1", we can't just use A and B together? Is that expected/part of the cargo semantics as well?

PR #2179, opened 2025-02-27
Add missing package tests

jneem:
I somehow managed to leave these out of the last package-management PR (even though I included the snapshots of expected output!)

### DIFF
--- a/package/tests/integration/inputs/path/duplicate-lock-name/Nickel-pkg.ncl
+++ b/package/tests/integration/inputs/path/duplicate-lock-name/Nickel-pkg.ncl
@@ -1,0 +1,10 @@
+{
+  name = "duplicate-lock-name",
+  description = "A package having two different identically-named path packages in its dep graph",
+  version = "0.1.0",
+  authors = ["Joe"],
+  minimal_nickel_version = "1.9.0",
+  dependencies = {
+    leaf = 'Path "../single-path-dep"
+  },
+} | std.package.Manifest
--- a/package/tests/integration/inputs/path/duplicate-path/Nickel-pkg.ncl
+++ b/package/tests/integration/inputs/path/duplicate-path/Nickel-pkg.ncl
@@ -1,0 +1,11 @@
+{
+  name = "duplicate-path",
+  description = "A package having two differently named dependency paths to the same path package",
+  version = "0.1.0",
+  authors = ["Joe"],
+  minimal_nickel_version = "1.9.0",
+  dependencies = {
+    dep = 'Path "../single-path-dep",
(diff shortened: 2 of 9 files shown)


package/tests/util.rs
```diff
@@ -104,7 +104,7 @@ impl ManifestBuilder {
 /// so that you can use it as a git dependency.
 pub fn init_git() -> TempDir {
     let dir = TempDir::new().unwrap();
-    run!(dir.path(), "git init");
+    run!(dir.path(), "git init --initial-branch=master");
```
jneem, line 107: Without this, tests fail if your global git config defaults to "main"

jneem:
The "out of space" error on macOS has been happening a lot lately. Is there something we can do about it?

package/tests/integration/inputs/path/duplicate-path/Nickel-pkg.ncl
```diff
@@ -0,0 +1,11 @@
+{
+  name = "duplicate-path",
+  description = "A package having two differently named dependency paths to the same path package",
+  version = "0.1.0",
+  authors = ["Joe"],
+  minimal_nickel_version = "1.9.0",
+  dependencies = {
+    dep = 'Path "../single-path-dep",
+    leafy = 'Path "../leaf",
```
yannham, line 9: Shouldn't the paths be the same here? That what's I thought from the description.

package/tests/integration/inputs/path/duplicate-lock-name/Nickel-pkg.ncl
```diff
@@ -0,0 +1,10 @@
+{
+  name = "duplicate-lock-name",
+  description = "A package having two different identically-named path packages in its dep graph",
+  version = "0.1.0",
+  authors = ["Joe"],
+  minimal_nickel_version = "1.9.0",
+  dependencies = {
+    leaf = 'Path "../single-path-dep"
```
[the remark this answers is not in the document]
jneem: This package has a dependency that it calls "leaf", which has a dependency that it calls "leaf". We're testing that the lock file will deduplicate them to "leaf" and "leaf 1".

package/tests/integration/inputs/path/duplicate-path/Nickel-pkg.ncl
jneem: "dep" also has `'Path "../leaf"` as a dependency. I changed the description to "A package having two different routes in the dependency graph to the same path package". Is that clearer?

yannham (approved):

PR #2184, opened 2025-03-04
Add std.cast

jneem:
This adds a `std.cast` function that's like the payload-carrying version of `std.typeof`. The motivation is to make it easier to write statically typed code that checks dynamic types.

The immediate motivation was for the contracts in json-schema-to-nickel, where statically typing the `max_properties` contract looks like

```
{
max_properties: Number -> Dyn = fun max => std.constract.custom (fun label value =>
  if std.is_record value then
    if std.record.length (value | {_: Dyn}) <= max then ... else ...
#                                 ^^^ annoying contract annotation, because value lacks a static type
  else
    'Error ...
)
}
```

With this addition, you can write it as
```
{
max_properties: Number -> Dyn = fun max => std.constract.custom (fun label value => std.cast value
  |> match {
    'Record r => if std.record.length r <= max then ... else ...
    _ => 'Error ...
  })
}
```


yannham (commented):
I can see the motivation for such an operation. As preliminary questions:

- What is the motivation to make this a primop? Is it to avoid undue contract evaluation? Because otherwise I guess you could implement it in user-code as `x |> %typeof% |> match { 'Number => 'Number (x | Number), ...etc}`.
- I think it's a nice solution with the current features of Nickel. In the long term, though, I wonder if we won't want to implement some form of occurrence typing (or flow-sensitive typing), which would make the original example work without contract annotation (_à la TypeScript_). In this case I'm wondering if adding a different way of doing that now is a good idea on the long-term. On the other hand, it's filling a need, and it's not awful if at some point we have `cast` and `typeof` overlap; we can deprecate `cast`, which is still quite readable/understandable.

So I guess there are three possible solutions:

- This PR
- This PR but without the primop
- Move `cast` to an external lib for the time being (or let people implement it if they need it)

jneem:
I guess there isn't good evidence that this is needed in the standard library. The main reason I went for the primop was to avoid the contract application for possible (but unmeasured) performance reasons.

Regarding occurrence/flow-sensitive typing, I'm probably biased by using rust as my main language for many years now, but what benefit does it have over ADTs + pattern matching?

yannham:
> Regarding occurrence/flow-sensitive typing, I'm probably biased by using rust as my main language for many years now, but what benefit does it have over ADTs + pattern matching?

flow-sensitive typing makes sense for gradually typed language, precisely for this use-case: you get some untyped data in a typed context, but thanks to some blessed test, the typechecker can refine the type of the values in each branch. TypeScript uses this to emulate ADTs, but I think it's useful in general. Since this kind of decision tree `if typeof foo == bar` is common in such languages, flow-sensitive typing makes it possible to write statically typed code while keeping such dynamic tests in an idiomatic/natural way.

This technique doesn't really make sense as it is for strongly, statically, nominally-typed language like Rust, Haskell, OCaml, where you can't really "cast" or refine values to begin with. Although you could argue that refinement types, dependent types and GADTs do have require vaguely similar mechanisms.

However, I also realize that all the simple examples I can think of could probably be covered by the "produce an ADT" approach as well. So maybe `cast` is covering 90% of the examples and we shouldn't bother with occurrence typing. Let's sleep on it and discuss that on the next weekly?

jneem:
Sure, I can also just open an issue and we can let it sit for a while. I don't think there's any urgency for this

core/stdlib/std.ncl
```diff
@@ -4959,6 +4959,46 @@
...
+      'ForeignId Dyn,
+      'CustomContract Dyn,
+      'Type Dyn,
+      'Other Dyn
+    |]
+    | doc m%"
+      Returns the value, wrapped in a type discriminator.
+
+      Mainly useful in statically typed code, because it allows you to extract a
+      statically typed value in a match statement.
```
yannham, line 4981: I would refer to `typeof` here, and to `cast` in the documentation of `typeof`.

yannham (approved):
After more thinking, I think whatever operation that is blessed by flow-sensitive typing (like `typeof`) can be mechanically turned into a `cast`-like interface. For example, if in `if std.blessed_test x == y then ... else ... ` gives you the refinement that `x` is of type `T` in the `then`-branch, I guess you can always write a corresponding `blessed_cast : Dyn -> [| 'Y T, ... |]` with tags being the discrete and hopefully finite possible values of `blessed_test`. This doesn't require any extension to the typechecker whatsoever, at the cost of using a maybe slightly less dynamic-typing-idiomatic version, but it's not so much of an issue in Nickel where there's a clear separation between statically typed code (where you shouldn't really use `typeof` anyway) and dynamically code, which doesn't care about flow-sensitive typing.

I'm a bit worried about the inflation of primitive operations. I guess a `%unsafe_cast%` much like Rust's `transmute` could make it possible to implement this (and hopefully other currently primop) as pure user-code in the stdlib. But we'll need to review primops for the bytecode virtual machine anyway, so let's proceed this way for now.

Issue #2050, opened 2024-09-24
`std.contract.from_validator` is hard to use in statically typed code


jneem:
`std.contract.from_validator` takes an argument of type `Dyn -> [| 'Ok, 'Error { message | String | optional, notes | Array String | optional, } |]`, but it's annoying to produce a literal of that type. For example, consider

```nickel
let
  IsZero = std.contract.from_validator (fun x => if x == 0 then 'Ok else 'Error { message = "blah" }
in
0 | IsZero
```

This runs ok, but as soon as I wrap it in `(...) : _` it errors with

```
2 │   IsZero = std.contract.from_validator (fun x => if x == 0 then 'Ok else 'Error { message = "blah" })
  │                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this expression
  │
  = Expected an expression of an enum type with the enum row `'Error { message | String | optional, notes | Array String | optional, }`
  = Found an expression of an enum type with the enum row `'Error { message : String }`
  = Could not match the two declarations of `Error`
```

Maybe this is another situation where we can use subtyping?

A work-around is to annotate the record type, like `{message = "blah"} | { message, notes | optional }` but that's a bit verbose. Also, I'm a little surprised that this annotation passes the typechecker despite not supplying a type for `notes`.

yannham:
The rationale was that I expected people would use `from_validator`  more in untyped code than in typed code, and thus that the well-typed alternative without optional values would be more cumbersome most of the time (that is, you have to always specify both `message` and `notes`, and it's also not forward-compatible with adding more fields in the future).

However, I do agree that it's annoying to use in statically typed code. The issue is that `{message | String | optional, notes | Array String | optional}` is currently **not** representable in the type system. That would be the first step: add optional fields. Then, it would indeed not be very hard to extend subtyping - especially now that we have the base implementation - such that `{message : String} <: {message ?: String, notes ?: Array String}`, and thus we could probably just change the type annotation of `from_validator` without breaking backward compatibility (or only in specific, easy-to-fix cases). Or implement something that is described in the original "merge type and term syntax RFC", which is that we could extract a static type from a contract when they have a specific form - in this case without even changing the contract annotation, it would be statically understood as `{message ?: String, notes ?: Array String}`. But that's a lot of non trivial steps to get there.

[...]

PR #2185, opened 2025-03-04
Optional field types

jneem:
This was an attempt at #2050 

- makes fields optional in record types
- adds a unification rule for optional record fields
- adds automatic UniTerm -> Type conversions for records with "simple enough" contract annotations

I think the last one was a mistake, though. It's not backwards compatible (e.g. it breaks one test), and it leads to some weird special cases. I'm going to try doing the contract -> type conversion later (like maybe just in the typechecker itself) and see if that's better.

I also didn't figure out how to trigger subtype checking. I tried some variants of the existing subtype tests but I only ever hit the unification code.

### DIFF
--- a/core/src/bytecode/ast/compat.rs
+++ b/core/src/bytecode/ast/compat.rs
@@ -1042,6 +1042,7 @@ impl<'ast> FromAst<RecordRow<'ast>> for mline_type::RecordRow {
     fn from_ast(rrow: &RecordRow<'ast>) -> Self {
         mline_type::RecordRowF {
             id: rrow.id,
+            opt: false,
             typ: Box::new(rrow.typ.to_mainline()),
         }
     }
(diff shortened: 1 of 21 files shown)


yannham (commented):
I haven't looked at it yet, but my first reaction is maybe that a PR is a bit premature. A type system (especially with type inference) is a complicated and intricate system and each addition or changes must be principled and if possible backed by existing literature, IMHO. Sometimes seemingly harmless additions might interact unexpectedly with other parts of the system (often with polymorphism).

I currently don't know a lot about optional fields in structural typing, maybe it's a rather simple extension. I don't know yet if this requires an RFC, but I would advocate for at the very least an issue proposal where we can debate the design, the syntax, the motivations, and the interaction with the existing components - in particular how does it behave with respect to (row) polymorphism, to subtyping, to unification, to inference, to contracts, etc.

jneem:
Oh, a PR is definitely premature :sweat_smile: I mostly just uploaded this to record the not-too-successful experiment. I'll open an issue for further discussion.

yannham:
Ah, my bad! A draft PR for experimenting is rarely too premature indeed :slightly_smiling_face: I just wanted to avoid putting too much effort into this before careful design.

PR #2194, opened 2025-03-25
[Chore] Pretty printing for the new AST

yannham:
Until now, the new AST `bytecode::ast::Ast` was pretty printed by first converting it through the legacy `RichTerm` and then pretty printing the result. In addition to being wasteful, the legacy AST won't be there forever; this PR bites the bullet and implement `Pretty` for the new AST.

The code was copy pasted from the current `pretty` module and adapted. There is duplication between the two modules, as some helper functions are copied unchanged. However, since we're going to get rid of the current `pretty` at some point, I'm not too inclined to spend time deduplicating the code or introducing coupling between the two. My hope is that that ripping the bandaid will be painless. But happy to be convinced otherwise.

On a positive note, the migrated implementation has less dark corner-case and ad-hoc logic (such that unfolding applications `(f a) b` to `f a b`, difference between `Let`/`Fun` and their `LetPattern`/`FunPattern` counterpart, etc.), since the new AST preserves the original structure quite faithfully (we also preserve field paths for example).

### DIFF
--- a/core/src/bytecode/ast/primop.rs
+++ b/core/src/bytecode/ast/primop.rs
@@ -933,7 +933,7 @@ pub enum PrimOp {
     ArraySlice,
 }
 
-/// Syntactic positionning of a primitive operator.
+/// Syntactic positioning of a primitive operator.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum OpPos {
     Infix,
--- a/core/src/bytecode/ast/primop.rs
+++ b/core/src/bytecode/ast/primop.rs
@@ -933,6 +933,14 @@ pub enum PrimOp {
     ArraySlice,
 }
 
+/// Syntactic positionning of a primitive operator.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum OpPos {
+    Infix,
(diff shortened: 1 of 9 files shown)


core/src/bytecode/ast/primop.rs
```diff
@@ -933,6 +933,14 @@ pub enum PrimOp {
     ArraySlice,
 }
 
+/// Syntactic positionning of a primitive operator.
```
jneem, line 936:
```suggestion
/// Syntactic positioning of a primitive operator.
```

lsp/nls/src/requests/completion.rs
```diff
@@ -154,11 +156,9 @@ fn sanitize_record_path_for_completion<'ast>(ast: &Ast<'ast>) -> Option<&'ast As
 }
 
 fn to_short_string(typ: &Type<'_>) -> String {
-    use compat::FromAst as _;
-
     let alloc = Allocator::bounded(DEPTH_BOUND, SIZE_BOUND);
     //TODO[RFC007]: Implement Pretty for the new AST
```
jneem, line 160: This is done now, right?

PR #2195, opened 2025-03-26
Automatically convert interpolated values to string

yannham:
This PR implements a feature request (although I can't find the issue for it, but I believe it was requested several times on other channels at least, or maybe as a comment on a different issue) to avoid having to convert manually values to string before interpolating them (at least for primitive values). That is, string interpolation now accepts not only strings, but also numbers, booleans, enum tags and `null`.

This raises the question of what to do about statically typed code. I adopted the most restrictive approach for now, which is to make the typechecker still infer interpolated values to be string; so in a statically typed block, the user has to use `to_string`. This is properly documented in the manual in consequence.

On some-place-I-still-don't-remember (or maybe oral discussion in a weekly), we also discussed a potential follow-up which would be to somehow hardcode typing rules for `std.string.Stringable`: the introduction rule would simply be defined by subtyping, that is `String <: Stringable`, `Number <: Stringable`, `Bool <: Stringable`, etc. This is a bit ad-hoc, but might be ok for something as fundamental as string interpolation. This evolution is compatible with the restrictive approach taken in this PR.

### DIFF
--- a/core/src/term/mod.rs
+++ b/core/src/term/mod.rs
@@ -52,8 +52,8 @@ pub use malachite::{
 
 use serde::{Deserialize, Serialize, Serializer};
 
-// Because we use `IndexMap` for recors, consumer of Nickel (as a library) might have to
-// manipulate values of this type, so we re-export this type.
+// Because we use `IndexMap` for recorsd, consumer of Nickel (as a library) might have to
+// manipulate values of this type, so we re-export it.
 pub use indexmap::IndexMap;
 
 use std::{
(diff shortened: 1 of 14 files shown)


core/src/term/mod.rs
```diff
@@ -52,8 +52,8 @@ pub use malachite::{
 
 use serde::{Deserialize, Serialize, Serializer};
 
-// Because we use `IndexMap` for recors, consumer of Nickel (as a library) might have to
-// manipulate values of this type, so we re-export this type.
+// Because we use `IndexMap` for recorsd, consumer of Nickel (as a library) might have to
```
jneem, line 55:
```suggestion
// Because we use `IndexMap` for records, consumer of Nickel (as a library) might have to
```

jneem (approved):

core/src/term/mod.rs
yannham: When you "fix" things :smile:

PR #2197, opened 2025-03-26
Add regression test for #1971

### LINKED ISSUE #1971: Hovering on a record field defined several times will only show the information for the first definition
**Describe the bug**

In the LSP, hovering over a record field will show some information on that field if there's any:

image

However, if the field is defined more than once, only the first one will get a useful information on hover:

image

image

**To Reproduce**

On the following file:

```nickel
{ foo, foo } | { foo | doc "The field Foo" }
```

Hover on the second `foo`

**Expected behavior**

Both instances of the field should display the full information on hover

**Environment**
 - OS name + version:
 - Version of the code: f5b5876

**Additional context**

yannham:
Closes #1971. The latter was automagically fixed by the migration of NLS to the new AST, which is more faithful. This PR just adds a regression test.

### DIFF
--- a/lsp/nls/tests/inputs/hover-double-def.ncl
+++ b/lsp/nls/tests/inputs/hover-double-def.ncl
@@ -1,0 +1,10 @@
+### /main.ncl
+{ foo, foo } | { foo | doc "The field Foo" }
+### [[request]]
+### type = "Hover"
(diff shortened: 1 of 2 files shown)


yannham:
@jneem just in case you missed this small one :slightly_smiling_face:

jneem (approved):

PR #2200, opened 2025-03-28
Build the workspace in the flake

jneem:
This changes the nix packages so that they build the whole workspace instead of going package-by-package. Because of the way feature unification works in cargo, this should reduce the number of times we need to build dependencies.

As a consequence, this also runs more tests in CI: `nix flake check` now runs the tests for all crates, not just the few that we selected.

It may be possible to get back to package-by-package builds in the future, since cargo would like to improve this on their end: https://rust-lang.github.io/rfcs/3692-feature-unification.html

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -397,6 +397,7 @@
           # A version of the Nickel CLI with the experimental package management
           # feature enabled.
           nickel-lang-pkg = fixupGitRevision (buildWorkspace {
+            pnameSuffix = "-pkg";
             extraBuildArgs = "--features package-experimental";
             extraArgs = {
               inherit env;
@@ -649,6 +650,7 @@
       packages = {
         inherit (mkCraneArtifacts { })
           nickel-lang
+          nickel-lang-pkg
           benchmarks
           cargoArtifacts;
         default = packages.nickel-lang;
--- a/flake.nix
+++ b/flake.nix
@@ -637,6 +637,17 @@
         ''
           ${pkgs.lib.getExe self.packages."${system}".default} test ${./core/stdlib/std.ncl} && mkdir $out
         '';
+
+      # This flake used to build nickel-lang-cli and nickel-lang-lsp packages with separate
+      # crane invocations. We switched to building the full workspace for better cargo caching,
+      # but we also support the old packages for backwards-compatibility.
+      compatPackage = { nickel-lang, binary, name }: pkgs.runCommand name
+        {
+          meta.mainProgram = binary;
+        } ''
+        mkdir -p $out/bin
+        cp -r "${nickel-lang}/bin/${binary}" $out/bin/
+      '';
     in
     rec {
       packages = {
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -371,39 +357,38 @@
...
+              # package.build gets set to a dummy file. reset it to use local build.rs
+              # tomlq -i broken (https://github.com/kislyuk/yq/issues/130 not in nixpkgs yet)
+              ''
+                ${pkgs.yq}/bin/tomlq -t 'del(.package.build)' $out/core/Cargo.toml > tmp
+                mv tmp $out/core/Cargo.toml
+              '';
             };
+
+            cargoArtifacts = cargoArtifactsDeps;
+            cargoExtraArgs = "--package nickel-lang-core";
```
yannham, line 381: Why is this preferred over the previous `cargoPackage = "nickel-lang-core";`? This seems lower-level

flake.nix
```diff
@@ -338,16 +318,22 @@
...
               CARGO_PROFILE = profile;
             } // extraArgs);
 
           # To build Nickel and its dependencies statically we use the musl
           # libc and clang with libc++ to build C and C++ dependencies. We
           # tried building with libstdc++ but without success.
-          buildStaticPackage = { pnameSuffix, cargoPackage, extraBuildArgs ? "", extraArgs ? { } }:
-            (buildPackage {
-              inherit pnameSuffix cargoPackage;
+          buildStaticWorkspace = { pnameSuffix, extraBuildArgs ? "", extraArgs ? { } }:
```
yannham, line 334: It's a detail but the ability to decouple `pnameSuffix` from the original cargo package name had a small upside, which is that the derivations would exhibit more specific names (`nickel-lang-cli-pkg` for example, instead of just `nickel-lang-cli`). It's not very important, but was reason for sometimes having a `pnameSuffix` that doesn't match precisely the package name.

yannham (commented):
That looks good. Do we have any way of measuring, quantitatively or qualitatively the effect of this PR? I always find it a bit hard to know if this has any positive or negative effects, given that the CI run time is also impacted by previous caching. We still seem to build a lot of things in the action, but is it less or more than before? :man_shrugging:

flake.nix
jneem: I'm not sure I understand: we don't use cargo package names anymore, so why would we need to decouple them from suffices? We can still have derivations with more specific names (like `nickel-lang-pkg` vs `nickel-lang`)

jneem:
> is it less or more than before?

Good question, I'll see if I can measure this somehow

flake.nix
yannham: Sure, my point I guess is that as it is, this PR will show `nickel-lang-cli>` in the logs for both `nickel-lang-cli` and `nickel-lang-cli-pkg`, while it didn't use to be the case, is that correct? I agree that we don't really care about the `pnameSuffix` mechanism in particular.

jneem:
I can confirm that this approach reduces the size of the cached cargo dependencies. Previously, I had
```
3.7M /nix/store/1pbz3gqq1vfykwjs3h408s404655i2yr-nickel-lang-core-lalrpop-1.11.0/target.tar.zst
851M /nix/store/1pbz3gqq1vfykwjs3h408s404655i2yr-nickel-lang-core-lalrpop-1.11.0/target.tar.zst.prev
```
After this patch, I have
```
151M /nix/store/9g01s3srgj6zd0fhv1k0zl38041ilv83-nickel-lang-core-lalrpop-1.11.0/target.tar.zst
429M /nix/store/9g01s3srgj6zd0fhv1k0zl38041ilv83-nickel-lang-core-lalrpop-1.11.0/target.tar.zst.prev
```

So, 850M -> 580M.

flake.nix
jneem: Good point. I think the latest version is using `pnameSuffix` correctly now.

Issue #1968, opened 2024-06-19
Forward the metadata when re-exporting a field


thufschmitt:
**Is your feature request related to a problem? Please describe.**

Record fields can have an attached metadata. Since this metadata is attached to the field itself and not the value, it isn't forwarded when the value is used. It's sometimes what we want, but not always. In particular, it's very natural to re-export a field from another record, but that means losing the documentation, type annotation, and so on.

For instance, in:

```nickel
let r = {
  foo | doc "Some doc" = 1,
}
in
{
  foo = r.foo,
}
```

the exported documentation (or the LSP) won't show the documentation for `foo`.

**Describe the solution you'd like**

Have the metadata forwarded when re-exporting a field from a different record.

**Describe alternatives you've considered**

This is bound to be somewhat heuristic and fragile (what about `{ foo = let bar = r.foo in bar }`?). Maybe that would be an argument for an `inherit` keyword that would not just forward the value, but also the metadata. That way setting `{ foo = r.foo }` would keep the current semantics, but the “re-export” case could be handled by `{ inherit foo from r }` (or whatever the syntax would be).

Related: https://github.com/tweag/nickel/issues/747

**Additional context**

This shows-up in a few places in Organist. For instance, [`import_nix`](https://github.com/nickel-lang/organist/blob/7b3a104a2cab71ccfabd0ebe85c82bd6c9491539/lib/nix-interop/builtins.ncl#L20-L33) has a documentation and a schema, but that's lost when it's re-exported ([here](https://github.com/nickel-lang/organist/blob/7b3a104a2cab71ccfabd0ebe85c82bd6c9491539/lib/nix-interop/nix.ncl#L29) and [here](https://github.com/nickel-lang/organist/blob/7b3a104a2cab71ccfabd0ebe85c82bd6c9491539/lib/organist.ncl#L6)).

yannham:
> This is bound to be somewhat heuristic and fragile

I agree. Also, there are some cases when the semantics of propagating metadata is really not what you want (in a bad way). This question of propagation has been extensively scrutinized in RFC005, see for example [this section](https://github.com/tweag/nickel/blob/master/rfcs/005-metadata-rework.md#propagating-metadata). For balance, propagation is acknowledged as desirable though in some situations [here](https://github.com/tweag/nickel/blob/master/rfcs/005-metadata-rework.md#representation), so it was a known tradeoff when siding for static metadata that can only appear at field location.

[...]

Issue #1931, opened 2024-05-29
A way to assign a variable to a field with the same name


norpadon:
**Is your feature request related to a problem? Please describe.**
Here is an example config:
```
let learning_rate = 1e-3
in {
    optimizer = {
        learning_rate = learning_rate
    }
}
```
If I try to compile it, I get an Error:
```
error: infinite recursion
  ┌─ /Users/norpadon/Work/ticksim-ml/nuggets/configs/demo.ncl:4:25
  │
4 │         learning_rate = learning_rate
  │                         ^^^^^^^^^^^^^ recursive reference
```

As I understand, the only way to fix it is to rename the outer `learning_rate` variable, which is inconvinient.

**Describe the solution you'd like**
I see two possible solutions:
1) Introduce some king of special `outer` record which always contains bindings in the outer scope, so that I can write `{learning_rate = outer.learning_rate}`
2) Disabling recursion for quoted field names, e.g. {"learning_rate" = learning_rate}

yannham:
Hi,

This is indeed missing. For the record, I think this is a duplicate of https://github.com/tweag/nickel/issues/747, which already has some proposals. But we can maybe keep this issue open as well, because the original one isn't as searchable.

yannham:
Since #747 has been closed. `include` expressions now allow to do that.

Issue #747, opened 2022-06-23
Field punning (inherit)


yannham:
# Inherit

Nix has an `inherit` keyword that is used to define a field with the same name and value as a a variable in scope:

```nix
let foo = 1; in
{ inherit foo; }

# is equivalent to
{ foo = foo; }
```

This kind of syntactic sugar is also present in other languages (Haskell, Rust, OCaml, etc.) sometimes under the name record /field punning.

Nix `inherit` also supports multiple fields, and the specification of another record to take the fields from:

```nix
let record = {foo = 1; bar = 2;}; in
let baz = 3; blarg = 4; in

{
  inherit (record) foo bar;
  inherit baz blarg;
}

# is equivalent to
{
  foo = record.foo;
  bar = record.bar;
  baz = baz;
  blarg = blarg;
}
```

## Recursive records in Nickel

While this is a nice shorthand, `inherit` is at most a nicety. However, in Nickel, records are recursive by default. This means that the expanded version above will just cause an infinite loop:

```nickel
{foo = foo}
```

This make defining fields with the same name and content as an identifier in scope a pain: one either has to change the variable from the beginning:

```nickel
let foo_ = 1 in
let bar_ = 2 in

{
  foo = foo_,
  bar = bar_,
}
```

But sometimes that is not desirable. Variables can be defined far away, and can be used in a lot of other places where it makes sense to keep the original name. The only solution left is to bind those variables to new ones locally, but this must be done before the definition of the record:

```nickel
let foo = 1 in
let bar = 2 in

{
  sub_record =
    let foo = foo_ in
    let bar = bar_ in
    {
       foo = foo_,
       bar = bar_,
     }
}
```

Another case that I encountered in practice is when you need to recursively refer to a field defined in a parent record, but with the same name. There, you can't rename the original identifier either:

```nickel
{
  meta = "metadata",
  sub = let meta_ = meta in {
    meta = meta_
  }
}
```

This starts to be really painful. Because records are recursive by default, I think having field punning is even more important for Nickel than for other languages.

**Describe the solution you'd like**

Have a syntactic construct, which can be `inherit` or something else, to define fields with the same name and value as a list of identifiers in scope. 

Note that the common punning syntax for several languages (Rust, OCaml, etc.) is to just write a field without a definition. For example in OCaml:

```ocaml
type r = {foo : int}
let foo = 1 in {foo;}
```

Or in Rust:

```rust
struct Foo { bar: i32 }

fn mk_foo() -> Foo {
[...]

PR #2204, opened 2025-04-02
[Feat] Field punning (aka `include` aka `inherit`)

yannham:
This is a first implementation of #747 (and #1931, related #1968, although it doesn't support field yet). For now, this doesn't support extracting a field from a record (which is a bit more complicated dynamically), just from variables.

There are a number of design questions that rose up during the implementation. I reported them in #747.


yannham:
Closed in favor of a smaller scoped PR #2241. I keep this branch for the follow-up that will include metadata annotations.

PR #2207, opened 2025-04-04
Add package-management enabled flake outputs and release binaries

yannham:
As package management is landing as an experimental feature, I suspect people will want to try out in the next release. But currently it's not enabled by default, which means you have to fetch this repo and manually build from source with the proper cargo command, requiring specialized knowledge. We probably don't want to include it by default in e.g. the crates.io packages, but I'd still like to have a simpler way of distributing it.

This PR finds a middle ground for common architectures by providing package-management-enabled binaries flake outputs and upload them as part of the release workflow for the usual platforms we support. If people use an exotic platform, they can still build from source, but the majority of users should be able to just use the distributed binaries.

There are a few cosmetic changes / clean up in the flake unrelated to package management.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -70,6 +70,9 @@
         ];
       };
 
+      # Provide a `wasm-bindgen-cli` package which has the exact same version as
+      # the Rust dependency `wasm-bindgen` locked in the Cargo.lock file. This
+      # is needed to properly build the WASM REPL.
       wasm-bindgen-cli =
         let
           wasmBindgenCargoVersions = builtins.map ({ version, ... }: version) (builtins.filter ({ name, ... }: name == "wasm-bindgen") cargoLock.package);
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -245,10 +248,12 @@
...
             outBin="$out/bin/$(basename $srcBin)"
-            # [dirty] must have 7 characters to match dummyRev (hard coded in
-            # nickel-lang-cli and nickel-lang-lsp)
-            # we have to pad them out to the same length as dummyRev so they fit
-            # in the same spot in the binary
+        '' +
+        # [dirty] must have 7 characters to match dummyRev (hard coded in
+        # nickel-lang-cli and nickel-lang-lsp)
+        # we have to pad them out to the same length as dummyRev so they fit
+        # in the same spot in the binary
```
yannham, line 255: This avoid having comments be part of the output, and thus causing a rebuild if we ever change them.

flake.nix
```diff
@@ -415,16 +432,23 @@
             };
           });
 
-          # Static building isn't really possible on MacOS because the system call ABIs aren't stable.
+          # Static building isn't really possible on MacOS because the system
+          # call ABIs aren't stable. This output shouldn't be used on MacOS.
           nickel-static =
-            if pkgs.stdenv.hostPlatform.isMacOS
-            then nickel-lang-cli
```
yannham, line 421: the if-else is useless because there is already a guard at the output level, so `nickel-static` is not even available as an output on Darwin to begin with.

jneem (approved):

PR #2214, opened 2025-04-09
Traverse enum variants

### LINKED ISSUE #2180: nls doesn't pick up types of nested fields in `std.package.Manifest`
Consider the input

```nickel
{
  name = "single-git-dep",
  description = "A package with a git dependency",
  version = "0.1.0",
  authors = ["Joe"],
  minimal_nickel_version = "1.9.0",
  dependencies = {
    leaf = 'Git { url = 3 }
  },
} | std.package.Manifest
```

Hovering over "dependencies" gives me the right type, docs, etc. But hovering over "url" does not.

I've only observed this on a branch so far, just recording it here so I don't forget to test it on master.

jneem:
This fixes an issue where contracts inside enum variants wouldn't get turned into runtime contracts. This PR also modifies nls's permissive evaluator to recurse into enum variants. (I was trying to fix #2180 but got distracted...)

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -1047,6 +1047,9 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
                             }
                         }
                     }
+                    Term::EnumVariant { arg, .. } => {
+                        inner(slf, acc, arg.clone(), recursion_limit.saturating_sub(1));
+                    }
                     _ => {}
                 },
             }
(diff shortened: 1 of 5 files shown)


yannham (approved):
It's surprising that the missing case for `Traverse` hasn't caused more issues, but in fact we mostly use it for wildcard substitution (I thought we used it for type substitution but we resort to `map` instead).

Another test that could have found this is thus:

```nickel
let f : [| 'Foo _ |] -> Number = match { 'Foo n => n} in f ('Foo "a")
```

In your version, this should ideally fail with a contract error (wildcards are substituted for their inferred type and this acts as a contract). But indeed in 1.10 I get no error at all. Maybe we could add this test as well?

jneem:
Sure, added that test too. I guess bugs where we incorrectly allow things are just harder to catch than bugs giving wrong answers...

PR #2217, opened 2025-04-10
ADT and package management improvements in nls

jneem:
This PR combines two things that together improve the experience of editing package manifests (and other things too, probably).

The first improvement allows enum tags in parent/child paths. This means that the two "foo"s in `'Ok { foo } | [| 'Ok { foo } |]` are considered "cousin defs", which allows completions and metadata to propagate between them.

The second improvement fixes the resolution (in nls's analysis) of the name "std" from within the standard library. Previously it would fail to resolve, breaking the various parts of `std.package.Manifest` that refer to other things in `std`.

### DIFF
--- a/lsp/nls/src/requests/completion.rs
+++ b/lsp/nls/src/requests/completion.rs
@@ -265,6 +265,7 @@ fn field_completion<'ast>(
     world: &'ast World,
     path: &[Ident],
 ) -> Vec<CompletionItem<'ast>> {
+    log::info!("field completion for {ast}");
     let resolver = FieldResolver::new(world);
     let mut records = resolver.resolve_record(ast);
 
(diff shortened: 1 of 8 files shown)


lsp/nls/src/requests/completion.rs
```diff
@@ -265,6 +265,7 @@ fn field_completion<'ast>(
     world: &'ast World,
     path: &[Ident],
 ) -> Vec<CompletionItem<'ast>> {
+    log::info!("field completion for {ast}");
```
yannham, line 268: Are those logs supposed to stay, or were there just a temporary debug thing?

yannham (approved):

lsp/nls/src/requests/completion.rs
yannham: No, it's fine, I just wanted to make sure - I know I added a ton of them for debugging and had a hard time hunting them. If they're useful in general it's totally fine keeping them

PR #2219, opened 2025-04-10
Describe customize mode in the CLI chapter of the manual

### LINKED ISSUE #2215: Add documentation about the customize mode of the CLI to the manual
The customize mode of the CLI is still officially experimental, but has been somehow stable for several minor versions now. Whether we break it or not, like package management, it'd be good to have some user-facing documentation in the manual, because it's not very discoverable currently.

yannham:
Closes #2215. Fixed some capitalization for consistency as well.

### DIFF
--- a/doc/manual/cli.md
+++ b/doc/manual/cli.md
@@ -106,14 +106,21 @@ help`.
 The value part of an assignment supports an extended syntax called sigil
 expressions, of the form `@<selector>[/attribute]:<argument>`. Currently, the
 only supported selector is `env`, which fetches an environment variable and puts
-its value in the corresponding field as an expression:
+its value as a string in the corresponding field as an expression:
 
 ```console
 nickel export myconfig.ncl -- environment.path=@env:PATH enviornment.classpath=@env:CLASSPATH
 ```
 
-`@env` doesn't support any attribute and fails if the environment variable isn't
-set.
+Assigning `value=@env:VAR` achieves the same as `"value=\"$VAR\""` with two key
+differences:
+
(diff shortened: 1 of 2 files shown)


doc/manual/cli.md
```diff
@@ -34,12 +34,96 @@ Arguments:
...
+```console
+$ nickel export main.ncl -- params.environment=\'dev params.mock_db=true
+{
+  "test_string": "env=dev, mock_db=true"
+}
+~
+```
+
+The customize mode only allow to set fields that are not already set in the
+configuration (or only have a defaut value). You can override existing values as
```
jneem, line 72:
```suggestion
configuration (or only have a default value). You can override existing values as
```

doc/manual/cli.md
```diff
@@ -34,12 +34,96 @@ Arguments:
...
+expressions, of the form `@<selector>[/attribute]:<argument>`. Currently, the
+only supported selector is `env`, which fetches an environment variable and puts
+its value in the corresponding field as an expression:
+
+```console
+nickel export myconfig.ncl -- environment.path=@env:PATH enviornment.classpath=@env:CLASSPATH
+```
+
+`@env` doesn't support any attribute and fails if the environment variable isn't
+set.
```
jneem, line 116: Maybe it's worth pointing out that this is sort of the same as `environment.path=$PATH` except with sane semantics regarding escaping and undefined variables

jneem (approved):

yannham:
I took the liberty to add a small subsection about escape sequences in the syntax chapter,  because I realized that it was entirely missing.

PR #2234, opened 2025-04-18
Improved error message when subtyping an inhomogeneous record as a dict

### LINKED ISSUE #2231: Better error message when failing record -> dict subtyping because of non-homogeneity
Homogeneous (i.e. all values have the same type) records are subtypes of dicts, so that you can pass a `{ foo: Number }` to a function of type `{ _ : Number } -> Dyn`. This obviously doesn't apply to inhomogeneous records: `{ foo: Number, bar: String }` is not a subtype of any dict type. However, the typechecking error message you get when trying to cast `{ foo: Number, bar: String }` as a dict is pretty mysterious: it complains about an expected type `Number` but an actual type `String`.

The reason for this is that you're trying to subtype `{ foo: Number, bar: String }` as `{ _ : a }` so first `a` gets unified with `Number`, and then it tries to get unified again with `String`, which fails. See #2230 for an example of this kind of error "in the wild".

We should have an error message here making it clear that the issue is records vs dicts.

jneem:
Fixes #2231.

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -2654,7 +2654,6 @@ impl IntoDiagnostics for TypecheckError {
                         format!("Found a record with a field of type {expected} and a field of type {inferred}"),
                         "Records are compatible with dicts only if all their fields have the same type".into(),
                     ])]
-                //Expected a dict type (i.e. { _ : a })")
             }
             TypecheckError::OrPatternVarsMismatch { var, pos } => {
                 let mut labels = vec![primary_alt(var.pos.into_opt(), var.into_label(), files)
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -2650,7 +2650,7 @@ impl IntoDiagnostics for TypecheckError {
                     .with_message("incompatible types")
                     .with_labels(mk_expr_label(&pos))
                     .with_notes(vec![
-                        "Expected a dict type (like `{ _ : a }`)".into(),
+                        "Expected a dictionary type".into(),
                         format!("Found a record with a fields of type {expected} and {inferred}"),
                         "Records are compatible with dicts only if all their fields have the same type".into(),
                     ])]
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -451,6 +451,15 @@ pub enum TypecheckError {
         /// The position of the expression that was being typechecked as `type_var`.
         pos: TermPos,
     },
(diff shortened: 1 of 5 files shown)


core/src/typecheck/subtyping.rs
```diff
@@ -74,12 +74,32 @@ impl<'ast> SubsumedBy<'ast> for UnifType<'ast> {
...
-                            a.typ
+                            let row_result =
+                                a.typ
                                 .clone()
-                                .subsumed_by(*type_fields.clone(), state, ctxt.clone())?
+                                .subsumed_by(*type_fields.clone(), state, ctxt.clone());
+
+                            // One way that this row can fail to typecheck is if the record has rows with
+                            // differing types. In that case, the dict's field type will unify with the
+                            // first row's type, and then it will trigger a type mismatch with it tries
```
yannham, line 88:
```suggestion
                            // first row's type, and then it will trigger a type mismatch when it tries
```

core/src/error/mod.rs
```diff
@@ -2632,6 +2641,21 @@ impl IntoDiagnostics for TypecheckError {
...
+            } => {
+                vec![Diagnostic::error()
+                    .with_message("incompatible types")
+                    .with_labels(mk_expr_label(&pos))
+                    .with_notes(vec![
+                        "Expected a dictionary type".into(),
+                        format!("Found a record with a fields of type {expected} and {inferred}"),
+                        "Records are compatible with dicts only if all their fields have the same type".into(),
+                    ])]
+                //Expected a dict type (i.e. { _ : a })")
```
yannham, line 2657: Is this comment a left-over? It feels a bit out of place

yannham (approved):

core/src/error/mod.rs
```diff
@@ -2632,6 +2641,21 @@ impl IntoDiagnostics for TypecheckError {
...
+            } => {
+                vec![Diagnostic::error()
+                    .with_message("incompatible types")
+                    .with_labels(mk_expr_label(&pos))
+                    .with_notes(vec![
+                        "Expected a dictionary type".into(),
+                        format!("Found a record with a field of type {expected} and a field of type {inferred}"),
+                        "Records are compatible with dicts only if all their fields have the same type".into(),
+                    ])]
+                //Expected a dict type (i.e. { _ : a })")
```
jneem, line 2657:
```suggestion
                //Expected a dict type (i.e. { _ : a })")
```
```suggestion
```

PR #2235, opened 2025-04-21
Use ouroboros in core/src/cache.rs

jneem:
It turns out that the feature we thought was missing (the ability to simultaneously mutably borrow disjoint fields) was actually present in ouroboros all along, under the `with_mut` method.

This PR updates the core cache; nls is not yet converted. I did run into an issue with how `AstResolver` borrows the ast map, and it looked like there was an outdated comment describing a similar issue. I'm not super happy with the current solution, because it's pretty easy to forget copying over the new asts. But I do think it beats having lots of unsafe code...


yannham (changes requested):
That's great, less code and less unsafe. Thanks for noticing this. At the time I was looking at NLS as well, and I remember I thought it wouldn't work there for different reasons. But I can't remember for sure now :sweat: maybe because some of the safety relied on additional invariants that were specific to NLS and enforced by us. But it's worth trying though, I don't see why at least each `Analysis` registry can't be done using ouroboros.

One point that I thing might be missing is that in the code of the resolver, we probably want to also look up inside `new_asts` when resolving an import: typically the same file can be imported several times from the same source, and we don't want to do all of the work twice. Though I wonder if you'll then run into the borrowing issue you had with `asts` in the first place...

jneem:
Ok, revisiting this it turns out that `new_asts` wasn't needed after all. My previous try I'd run into borrowing errors with the ast cache, but it turns out that the problem was that `let ast = slf.asts.get(...)` returns an `Option<&'this &'ast Ast>` and after dereferencing away the `&'this` borrow everything worked fine.

yannham (approved):
Well, this is even better.

PR #2244, opened 2025-05-03
Add context to error opening trace file in nls

L0r3m1p5um:
This improves the error logging when nls is started with an invalid trace option. Right now the output of nls when starting it with a path that doesn't exist is:
```
$ nls --trace MissingDir/trace
Error: No such file or directory (os error 2) 
```
I ran into an issue where I had accidentally entered some input into the trace setting in the VSCode extension and it was pretty difficult to figure out what was going on from that message.

This PR improves the logging in a couple of ways:
- ```file.canonicalize()``` uses filesystem access to resolve symlinks, meaning that nls crashes there without ever printing the debug statement. I've changed that to path::absolute which does not access the filesystem.
- If nls fails to open the trace file, it will return the attempted file path along with the error.

### DIFF
--- a/lsp/nls/src/main.rs
+++ b/lsp/nls/src/main.rs
@@ -1,4 +1,8 @@
-use std::{fs, io, path::PathBuf, thread};
+use std::{
+    fs, io,
+    path::{self, PathBuf},
+    thread,
+};
 
 use anyhow::Result;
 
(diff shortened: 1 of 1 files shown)


jneem (approved):
Thanks!

PR #2247, opened 2025-05-07
Refactor Environment.

jneem:
Removes unsafe code, and also makes Environment covariant.

When trying to use ouroboros in nls, I ran into the problem that Environment isn't covariant, because it uses a RefCell internally to do copy-on-write. I don't really understand why though: it seems more natural to do the copy-on-writing during `insert`, where we have a mutable reference already and don't need interior mutability.

So this PR makes that change, and also removes some unsafe code that doesn't seem necessary. Local benchmarks suggest a small speed-up, if we can believe them.

### DIFF
--- a/core/src/environment.rs
+++ b/core/src/environment.rs
@@ -16,14 +16,12 @@ use crate::metrics::{increment, sample};
 /// keys, and `V` are their value.
 ///
 /// The linked list is composed of the current layer and the previous layers.
-/// The current layer is stored as an `Rc<Hashmap>`. It is inserted in the previous layers
-/// by cloning.
-/// The insertion is made by trying to get the current layer as mutable using
-/// [`Rc::get_mut`]. If it can, it means it is the only owner of this layer
-/// in the environment, allowing it to mutate it. If it cannot, it means that
-/// the current has been cloned and inserted in the previous environment already,
-/// so it can safely be reset as a new hashmap.
-/// The previous layers are set in order from the most recent one to the oldest.
+/// The current layer is stored as an `Rc<Hashmap>`. It is inserted in the
+/// previous layers by cloning.
+///
+/// Insertions are made by copying-on-write: if the current layer has no other
+/// references, it is mutated. If it has other references, the current layer is
+/// pushed down as the new "previous" layer, and a new layer is started.
 #[derive(Debug, PartialEq)]
 pub struct Environment<K: Hash + Eq, V: PartialEq> {
     current: Rc<HashMap<K, V>>,
@@ -299,12 +297,19 @@ mod tests {
         assert_eq!(env_base.iter_layers().count(), 1);
         env_base.insert(2, 'b');
         assert_eq!(env_base.iter_layers().count(), 1);
-        // TODO: make this test case better
-        let _ = env_base.clone();
+
+        let _env2 = env_base.clone();
+        env_base.insert(3, 'c');
+        assert_eq!(env_base.iter_layers().count(), 2);
         let mut iter = env_base.iter_layers();
         let map1 = iter.next().unwrap();
-        assert_eq!(map1.get(&2), Some(&'b'));
-        assert_eq!(map1.get(&1), Some(&'a'));
+        assert_eq!(map1.get(&3), Some(&'c'));
+        assert_eq!(map1.get(&2), None);
+        assert_eq!(map1.get(&1), None);
+        let map2 = iter.next().unwrap();
+        assert_eq!(map2.get(&3), None);
+        assert_eq!(map2.get(&2), Some(&'b'));
+        assert_eq!(map2.get(&1), Some(&'a'));
         assert!(iter.next().is_none());
     }
 
--- a/core/src/environment.rs
+++ b/core/src/environment.rs
@@ -1,5 +1,5 @@
 //! An environment for storing variables with scopes.
-use std::collections::{hash_map, HashMap};
+use std::collections::{HashMap, HashSet};
 use std::hash::Hash;
 use std::iter::FromIterator;
 use std::rc::Rc;
@@ -16,10 +16,9 @@ use crate::metrics::{increment, sample};
 /// keys, and `V` are their value.
 ///
 /// The linked list is composed of the current layer and the previous layers.
-/// The current layer is stored as an `Rc<Hashmap>`. It is inserted in the
-/// previous layers by cloning.
+/// The current layer is stored as an `Rc<Hashmap>`.
 ///
-/// Insertions are made by copying-on-write: if the current layer has no other
+/// Insertions are made by "split-on-write": if the current layer has no other
 /// references, it is mutated. If it has other references, the current layer is
 /// pushed down as the new "previous" layer, and a new layer is started.
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


core/src/environment.rs
```diff
@@ -19,43 +16,36 @@ use crate::metrics::{increment, sample};
...
-/// The insertion is made by trying to get the current layer as mutable using
-/// [`Rc::get_mut`]. If it can, it means it is the only owner of this layer
-/// in the environment, allowing it to mutate it. If it cannot, it means that
-/// the current has been cloned and inserted in the previous environment already,
-/// so it can safely be reset as a new hashmap.
-/// The previous layers are set in order from the most recent one to the oldest.
+/// The current layer is stored as an `Rc<Hashmap>`. It is inserted in the
+/// previous layers by cloning.
+///
+/// Insertions are made by copying-on-write: if the current layer has no other
```
yannham, line 22: Unless I'm misunderstanding (I'll probably know at the end of the review), I don't think copy-on-write is a really good name, because the whole point of the environment is to never copy anything (beyond  `Rc`s maybe). It's more like "share-on-write"?

yannham (changes requested):
I haven't read all of the diff, but I believe there's a confusion: the old environment is *not* copy-on-write. The current layer is never deeply cloned: it is appended to a shared linked list (prev) upon cloning, and shared among all clones. The reason why we need a refcell is that to implement this requirement, we need to mutate the environment upon cloning, but cloning is taking an immutable reference. I think we can consider current to be a uniquely owned `HashMap`, but we use `Rc` because it's easier to put in `prev` when needed and because it keeps track of previous clones for us instead of having to maintain a separate boolean flag.

This change, if I understand correctly, through `get_mut`, does copy the current layer upon writing after sharing. In fact, I believe the very first environment to be created (not from a clone) will never put anything in its previous layer, so it will be just one very large hashmap with a current layer.

It might end up being as a better approach in the end performance-wise, but this changes the complexity profile of the environment, and the impact isn't obvious to me. I wouldn't trust our simplistic benchmarks too much for this, unfortunately. I can try on larger benches locally, and see how it goes.

yannham:
Ah, sorry, I confused `get_mut` and `make_mut`, plus the "copy-on-write" terminology, made me making the wrong assumptions. I see that you're just delaying the mutation to the next insertion, which is fair.

jneem:
In fact, my first attempt used `make_mut`, but it failed because of the missing `Clone` bound :sweat_smile:

core/src/environment.rs
```diff
@@ -101,35 +103,18 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
...
-            // SAFETY: Rc::as_ptr never returnes null
-            .map(|hmap| unsafe { NonNull::new_unchecked(Rc::as_ptr(hmap) as *mut _) })
-            .collect();
-        // SAFETY: by design, env cannot be empty, and coming from an Rc, it is well aligned and initialized
-        let current_map = unsafe { env.pop().unwrap().as_ref() }.iter();
-        EnvElemIter { env, current_map }
+    pub fn iter_elems(
+        &self,
+    ) -> std::iter::Flatten<std::iter::Rev<std::vec::IntoIter<&HashMap<K, V>>>> {
+        let env: Vec<&HashMap<K, V>> = self.iter_layers().map(|hmap| &**hmap).collect();
```
yannham, line 116: Do we need to `collect` and then call `into_iter` right away for borrowing reasons (need to consume the first iterator)?

core/src/environment.rs
```diff
@@ -348,15 +296,20 @@ mod tests {
...
-        assert_eq!(map1.get(&2), Some(&'b'));
+        assert_eq!(map1.get(&3), Some(&'c'));
+        assert_eq!(map1.get(&2), None);
         assert_eq!(map1.get(&1), None);
         let map2 = iter.next().unwrap();
+        assert_eq!(map2.get(&3), None);
+        assert_eq!(map2.get(&2), Some(&'b'));
         assert_eq!(map2.get(&1), Some(&'a'));
-        assert_eq!(map2.get(&2), None);
         assert!(iter.next().is_none());
```
yannham, line 313: I find a bit counter intuitive that the next layer is empty. Where has `a` gone?

yannham (approved):

core/src/environment.rs
jneem: I don't understand: `c` is in the top layer (`map1`) and `a` and `b` are in the previous layer `map2`.

core/src/environment.rs
```diff
@@ -101,35 +103,18 @@ impl<K: Hash + Eq, V: PartialEq> Environment<K, V> {
...
-            .iter_layers()
-            // SAFETY: Rc::as_ptr never returnes null
-            .map(|hmap| unsafe { NonNull::new_unchecked(Rc::as_ptr(hmap) as *mut _) })
-            .collect();
-        // SAFETY: by design, env cannot be empty, and coming from an Rc, it is well aligned and initialized
-        let current_map = unsafe { env.pop().unwrap().as_ref() }.iter();
-        EnvElemIter { env, current_map }
+    pub fn iter_elems(
+        &self,
+    ) -> std::iter::Flatten<std::iter::Rev<std::vec::IntoIter<&HashMap<K, V>>>> {
```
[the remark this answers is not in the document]
jneem: It changes the behavior slightly, because with `impl Iterator` rust doesn't know if there's a `Drop` impl. I'll go with the `impl Iterator` though, it just requires one extra explicit `drop` in one test.

core/src/environment.rs
jneem: I think it's not so much for borrowing reasons as because we want to visit in reverse order. But the collecting was in the previous version also.

I wonder if it would be faster to implement `iter_elems` by walking the environment from newest to oldest, and keeping a `HashSet` to avoid repeating any keys. That seems maybe faster than the current approach that requires two collects. I'll try it.

core/src/environment.rs
yannham: My bad, I had trouble reading the diff, but looking at the final version alone it makes sense. I thought `a` disappeared from the iterator.

core/src/environment.rs
```diff
@@ -19,43 +16,36 @@ use crate::metrics::{increment, sample};
...
-        if !self.current.is_empty() && !self.was_cloned() {
-            sample!(
-                "Environment.curr_layer_size_at_clone",
-                self.current.len() as f64
-            );
-            self.previous.replace_with(|old| {
-                Some(Rc::new(Environment {
+        if self.current.is_empty() {
+            Self {
+                current: self.current.clone(),
```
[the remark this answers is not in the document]
jneem: Ok, I've measured this now and cloning the empty layer is actually faster: since the current layer is behind an `Rc`, making a fresh one requires an allocation which I guess is slower than the refcount stuff. I did simplify `clone` a little further though. Now all the emptiness checking is done in `set_current`.

core/src/environment.rs
jneem: Ok, I tried optimizing `iter` but it turns out that `iter` isn't actually used anywhere so I deleted it instead.

Before deleting the new `iter`, I tried using it to replace the existing uses of `iter_elems`, but it didn't help. Either way there's an allocation, and I guess the `Vec` for going backwards over the hashmaps beats the `HashSet` for tracking repeated elements.

PR #2254, opened 2025-05-15
Ouroboros nls

jneem:
Removes the unsafe from `nls`, with the help of ouroboros.

`PackedAnalysis` grows a lifetime parameter, allowing it to borrow external `Ast`s with that lifetime. The stdlib's analysis is a `PackedAnalysis<'static>` because it doesn't borrow external `Ast`s, and every other analysis is a `PackedAnalysis<'std>`, where `'std` is the lifetime of the stdlib's analysis. The constructor of `AnalysisRegistry` is in charge of tying these lifetimes together.

### DIFF
--- a/lsp/nls/src/analysis.rs
+++ b/lsp/nls/src/analysis.rs
@@ -1,3 +1,9 @@
+// This is needed to avoid warnings in the #[self_referencing] annotations. Somehow,
+// the way that proc macro works doesn't allow us to restrict this lint to just the
+// self-referencing struct. This has been fixed upstream but not released:
+// https://github.com/someguynamedjosh/ouroboros/pull/121
+#![allow(clippy::too_many_arguments)]
+
 use std::collections::HashMap;
 
 use lsp_server::{ErrorCode, ResponseError};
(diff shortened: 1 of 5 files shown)


lsp/nls/src/analysis.rs
```diff
@@ -549,230 +545,243 @@ impl PackedAnalysis {
...
-/// any allocator that will live as long as `_alloc`).
-unsafe fn borrow_analysis_mut<'a, 'ast>(
-    _alloc: &'ast AstAlloc,
-    analysis: &'a mut Analysis<'static>,
-) -> &'a mut Analysis<'ast> where {
-    std::mem::transmute::<&'a mut Analysis<'static>, &'a mut Analysis<'ast>>(analysis)
+    /// If this is the packed analysis of the stdlib, returns its type context.
+    ///
+    /// If this is any other packed analysis, panics.
+    pub(crate) fn stdlib_type_context(&self) -> &TypeContext {
```
yannham, line 657: There's probably a good reason for it, but the specification of this method sounds slightly odd. Is there no simpler way to get the stdlib context from the whole registry, first fetching the stdlib analysis and then its type context without being at risk of panicking?

lsp/nls/src/analysis.rs
```diff
@@ -549,230 +545,243 @@ impl PackedAnalysis {
...
+    ///
+    /// `callback` receives a reference to the whole analysis registry *without*
+    /// the analysis being modified, along with a mutable reference to the analysis
+    /// to modify.
+    pub fn modify<T, F>(&mut self, file_id: FileId, callback: F) -> Option<T>
+    where
+        F: for<'std> FnOnce(AnalysisRegistryRef<'_, 'std>, &mut PackedAnalysis<'std>) -> T,
+    {
+        self.modify_and_insert(file_id, |reg, analysis| {
+            Ok::<_, ()>((Vec::new(), callback(reg, analysis)))
```
yannham, line 730: If this can never fail, I believe you can be even more precise using [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) instead of `()`.

PR #2257, opened 2025-05-18
Update behavior of debugLog flag in VSCode extension

### LINKED ISSUE #2233: VSCode Extension seems to start in debug mode by default
**Describe the bug**
I have observed this behavior for quite a while, but only get around to give it a deeper look recently.

the nickel vscode extension seems to always launch `nls` with `RUST_LOG=trace`.
This results in lots of messages printed to vscode's "output" tab during editing (especially when there's ), so much so that it causes noticeable latency on keypress.

Maybe a separate issue, but sometimes even when the bottom panel is closed, the output tab of "Nickel Language Server" will automatically reopen the bottom panel and show me the trace log. This behavior doesn't feel like the fault of nickel's extension intuitively, but I never observe such behavior with other LSP extensions or extensions in general.

I skimmed through the extension code, and it seems `RUST_LOG=trace` is only set when debugging is enabled. But I never touched the configuration.

**To Reproduce**
I can reproduce this on my workstation running NixOS 24.11 and on my laptop running Pop!OS 22.04.

 
```shell
cd $(mktemp -d)
# create a trivial nickel file
'{}' > foo.ncl

# Start a fresh vscode instance with default setting
code --user-data-dir . foo.ncl

# Find nls PID
pgrep nls

# check the environment variables
sudo cat /proc/<pid_of_nls>/environ | tr '\0' '\n' | grep RUST_LOG
````

**Expected behavior**
Not printing trace log unless explicitly configured.

L0r3m1p5um:
Fixes #2233.

Since options.env is set as the target of Objects.assign, RUST_LOG=trace is being set on options.env, and not just debugOptions.env. This results in nls starting up with trace logging, regardless of the value of nls.server.debugLog. This is fixed to set trace logging only when debugLog is true as originally intended.

Additionally, I've set a value of RUST_LOG for the case in which debugLog is false. Users might have the RUST_LOG variable set for any number of reasons when running VSCode, and it's unlikely they intended for it to affect the behavior of the language server, but since the environment is passed through, nls would use that log level. It seemed to me that the info level logs would not be useful to most users so I've set the log level to warn in this case.

### DIFF
--- a/lsp/vscode-extension/src/extension.ts
+++ b/lsp/vscode-extension/src/extension.ts
@@ -24,8 +24,8 @@ export async function activate(context: ExtensionContext) {
 	console.log(serverPath);
 
 	// Pass the process.env to nls. Fixes direnv integration
-	const options = { env: process.env };
-	const debugOptions = { env: Object.assign(options.env, { "RUST_LOG": "trace" }) };
+	const options = { env: {...process.env, "RUST_LOG": "warn" } };
+	const debugOptions = { env: {...options.env, "RUST_LOG": "trace" } };
 
 	// If the extension is launched in debug mode then the debug server options are used
 	// Otherwise the run options are used


jneem (approved):

jneem:
Thanks!

PR #2260, opened 2025-05-19
Bigger stack on windows

### LINKED ISSUE #2237: Stack overflow with a large configuration
Panda on discord reported a stack overflow with a large configuration. I was able to reproduce the issue, but only after shrinking my stack size from 8MB to 2MB. The stack overflow was caused by a 400-ish deep recursion in `RichTerm::traverse`. 

Maybe we should revisit #1614? The new AST might make it easier. [stacker](https://github.com/rust-lang/stacker) might be another option.

As a short-term hack, we could just add linker flags to increase the default stack size.

jneem:
Fixes #2237, hopefully (I don't have a windows machine available to test on).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,13 @@
+# Set the windows stack size to 8MB, because the traversal functions are recursive
+# and can use up a substantial amount of stack.
+# https://github.com/tweag/nickel/issues/2237
+
+[target.x86_64-pc-windows-msvc]
+rustflags = [
+	"-C", "link-arg=/STACK:8000000"
+]
+
+[target.x86_64-pc-windows-gnu]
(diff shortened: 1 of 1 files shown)


yannham (commented):
I don't either. We could add the original example as a test, but IIRC we already modify the size of the stack for tests, so I'm not sure it would catch the difference. But it's worth trying (seeing if the orignal repro does fail with stack overflow on the Windows CI without this patch)

jneem:
The original repro is giant -- I think it's the same giant config you were using for benchmarking. I don't think it belongs in CI. I did finally manage to cross-compile a windows binary and successfully run the repro using wine, so I think this fix works.

yannham (approved):

PR #2263, opened 2025-05-20
Filter out LSP diagnostics for labels with no message

### LINKED ISSUE #2259: Syntax errors aren't displayed in VSCode using the LSP extension
**Describe the bug**
When using the Nickel extension for VSCode, error messages and locations do not get displayed for parsing errors.

**To Reproduce**
This seems to occur any time that a file has a syntax error as far as I can tell. Other types of errors, like type errors, import errors, etc. are displayed fine. So for example, the record below is syntactically invalid because of a missing comma:
```
{
  a = 1
  b = 2
}
```
However, VSCode does not display an error here.
Image

**Expected behavior**
VSCode normally displays a red underline at the location that the error was detected and displays an error message when mousing over it.

**Environment**
 - OS name + version: Darwin arm64 24.4.0
 - Version of the code: Version 0.4.0 of the Nickel Language Support plugin on VSCode 1.100.2

**Additional context**
Based on the language server logs in VSCode, nls does seem to sending publishDiagnostics with this errror:
```
[2025-05-19T00:13:21Z TRACE nls::server] Sending notification: Notification {
        method: "textDocument/publishDiagnostics",
        params: Object {
            "diagnostics": Array [
                Object {
                    "message": String("unexpected token"),
                    "range": Object {
                        "end": Object {
                            "character": Number(5),
                            "line": Number(2),
                        },
                        "start": Object {
                            "character": Number(4),
                            "line": Number(2),
                        },
                    },
                    "relatedInformation": Array [],
                    "severity": Number(1),
                },
                Object {
                    "message": String(""),
                    "range": Object {
                        "end": Object {
                            "character": Number(5),
                            "line": Number(2),
                        },
                        "start": Object {
                            "character": Number(4),
                            "line": Number(2),
                        },
                    },
                    "severity": Number(4),
                },
            ],
            "uri": String("file:///home/michael/nickel/test.ncl"),
        },
    }
[2025-05-19T00:13:21Z DEBUG lsp_server::msg] > {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"diagnostics":[{"message":"unexpected token","range":{"end":{"character":5,"line":2},"start":{"character":4,"line":2}},"relatedInformation":[],"severity":1},{"message":"","range":{"end":{"character":5,"line":2},"start":{"character":4,"line":2}},"severity":4}],"uri":"file:///home/michael/nickel/test.ncl"}}
```

The issue appears to be that this message is sending an array of two diagnostics, the second of which has an empty message. VSCode seems to reject an empty string as a valid message. This is from the Window logs of the VSCode output:
```
2025-05-18 20:13:21.893 [error] [Window] [Extension Host] Notification handler 'textDocument/publishDiagnostics' failed with message: message must be set
2025-05-18 20:13:21.920 [error] [Window] [Extension Host] Notification handler 'textDocument/publishDiagnostics' failed with message: message must be set
```

Apart from the empty message, it seems odd to me that two different diagnostics are being issued for the same location here. I wonder if something's getting duplicated. I don't have much familiarity with LSP though so I don't know if that's a normal thing to happen though.

jneem:
Fixes #2259.

### DIFF
--- a/lsp/nls/src/diagnostic.rs
+++ b/lsp/nls/src/diagnostic.rs
@@ -221,6 +221,11 @@ impl DiagnosticCompat for SerializableDiagnostic {
         diagnostics.extend(within_file_labels.filter_map(|label| {
             let range = lsp_types::Range::from_codespan(&label.file_id, &label.range, files)?;
 
+            // There's no point emitting a diagnostic without a message.
+            if label.message.clone().is_empty() {
+                return None;
+            }
+
             Some(SerializableDiagnostic {
                 range: OrdRange(range),
                 message: label.message.clone(),
--- a/lsp/nls/tests/snapshots/main__lsp__nls__tests__inputs__no-crash-on-pretty-print.ncl.snap
+++ b/lsp/nls/tests/snapshots/main__lsp__nls__tests__inputs__no-crash-on-pretty-print.ncl.snap
@@ -2,4 +2,4 @@
 source: lsp/nls/tests/main.rs
 expression: output
 ---
-[2:63-2:64: "unexpected token", 2:63-2:64: "", 3:6-3:7: "incompatible types\nExpected an expression of type `Number -> Dyn -> [| 'Ok, 'Error { message | String, <parse error>, } |]`\nFound an expression of type `Number`\nThese types are not compatible", 3:6-3:7: "this expression"]
+[2:63-2:64: "unexpected token", 3:6-3:7: "incompatible types\nExpected an expression of type `Number -> Dyn -> [| 'Ok, 'Error { message | String, <parse error>, } |]`\nFound an expression of type `Number`\nThese types are not compatible", 3:6-3:7: "this expression"]


yannham (approved):
I'm a bit surprised that we have labels with no messages. When does that happen? Is this errors coming from LALRPOP? Or additional "notes"-severity messages? Do we have to fix something for the CLI as well?

yannham:
(Ok, reading the original issue, I understand that we're talking about labels here, not diagnostics). Follow-up question: doesn't this just erase important position information from the diagnostics? If the LSP really needs a message for an associated label, can't we provide one instead (re-using the one from the diagnostic e.g.)?

jneem:
Ok, I may have already answered this in the original issue, but we're already grabbing the primary label's location and attaching it to the main diagnostic so I think it's unnecessary to also grab the main message and attach it to the primary label.

yannham:
Ok, sounds good then :+1:

PR #2273, opened 2025-06-06
Less test generation

jneem:
This switches our path-based tests from `test_generator`'s compile-time generation to runtime generation using `libtest_mimic`.

This has a couple of benefits:
- new test files are automatically picked up without having to trigger a rebuild some other way (for example, in `cargo test; echo '{' > cli/tests/snapshot/inputs/pretty/asdf.ncl; cargo test` the second test succeeds because the new snapshot doesn't get picked up
- modifying existing test files doesn't trigger any recompilation

It also has a couple of downsides:
- it's different from the usual test registration story, and the custom runner doesn't automatically recognize `#[test]`-annotated tests.
- it doesn't capture stdout by default. That's apparently [unfixable](https://github.com/LukasKalbertodt/libtest-mimic/issues/9) because of some black magic in the rust standard library. Workarounds include `cargo nextest`, or just living with messy test output.

While making these changes, I discovered some core integration tests that weren't actually being run, so I moved them to a place where they'd be picked up automatically.

### DIFF
--- a/core/src/files.rs
+++ b/core/src/files.rs
@@ -192,7 +192,7 @@ impl Files {
         self.files.get(id.0 as usize).ok_or(Error::FileMissing)
     }
 
-    pub(crate) fn filenames(&self) -> impl Iterator<Item = &OsStr> {
+    pub fn filenames(&self) -> impl Iterator<Item = &OsStr> {
         self.files.iter().map(|f| &*f.name)
     }
 }
--- a/core/tests/manual/main.rs
+++ b/core/tests/manual/main.rs
@@ -158,7 +158,7 @@ fn check_repl(content: String) {
                 (Err(e), ReplResult::Error(expected)) => {
                     let mut error = NoColor::new(Vec::<u8>::new());
                     let mut files = repl.cache_mut().sources.files().clone();
-                    dbg!(&files);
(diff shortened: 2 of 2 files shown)


yannham:
> it's different from the usual test registration story, and the custom runner doesn't automatically recognize #[test]-annotated tests.

What does this mean in practice?

jneem:
So for example in `lsp/nls/tests/main.rs` we were running a bunch of file-based snapshot tests but also a couple of special-cased tests. After switching to `libtest-mimic`, those special-cased tests were no longer registered by the `#[test]` macro, so I had to either register them manually or move them to another test file that uses the default harness (which is what I did).

yannham:
I see. This is a potential footgun, so I suppose it must be properly documented that one should not mix custom-harness-based tests and normal tests. The stdout part isn't a showstopper, but it is a degradation of the quality of life.

jneem:
Yeah, overall I'm not totally sold on this approach. It looks like there's some [interest](https://github.com/LukasKalbertodt/libtest-mimic/issues/54) [upstream](https://github.com/crate-ci/libtest2) in improving the testing situation, so maybe we can just wait for that.

jneem:
I'll close this for now and revisit if the built-in test situation improves.

PR #2274, opened 2025-06-06
Fix formatting of doc/manual/package-management.md

yannham:
This fixes the rendering of the package management chapter of the manual which is broken on the Nickel website.

image


jneem (approved):

yannham:
Wait, while we're at it, the command given is wrong: it's `nickel package publish`, not `nickel publish`...

PR #2275, opened 2025-06-06
Fix missing package field in package manual example

yannham:


jneem (approved):
Shouldn't this sort of thing get picked up by the manual snippet tests :thinking:

yannham:
> Shouldn't this sort of thing get picked up by the manual snippet tests 🤔

After digging quickly, it seems the default mode for evaluation of manual snippet is normal eval, and not deep eval. I suppose sometimes we don't want deep-eval (when writing examples of modules, which would have missing field errors). But maybe we want to do a bit better, like having an annotation for deep-eval, or even set deep eval to be the default and have an annotation for shallow evaluation? A last possibility would be to perform LSP-like evaluation, ignoring missing field definition errors⋅

yannham:
Note by the magic of Nickel, the previous version was parsed successfully: it defined a record with a field without definition named `github:nickel-lang/github-workflow` :slightly_smiling_face:

PR #2276, opened 2025-06-07
Fix cache miss when pulling source file by a relative path

L0r3m1p5um:
get_or_add_file tries to pull a file from SourceCache using the non-normalized path, but normalized it before inserting it into the cache if there's a cache miss. This means that if it's called with a relative path, it will always miss the cached file and reinsert it into the cache with a new file ID. I think this has been causing some performance issues I've been experiencing with the LSP. This fixes that problem by using the normalized path to pull from the cache instead.


jneem (approved):
Good catch, thanks!

PR #2282, opened 2025-06-11
[RFC007] Runtime value representation, take 1

yannham:
This is moving forward with RFC007. This PR aims implement the memory-packed representation of values described in the RFC. As we don't have a bytecode virtual machine yet, the _code_ part of the values will be using the old AST representation. But we'll trim a good part of the old AST representation, as all "values" (that is, weak head normal forms) will be represented using the new implementation, which should trim memory usage quite drastically, hopefully.

## Random thoughts (WIP)

- We can't use `[u8]` to represent the variable part of a value block, because then `Rc<ValueBlock>` is a fat pointer, and we can't cast it from and to a `usize` anymore. What's more, we don't actually need the size of `[u8]` - it's just the only way in Rust to represent a DST which is an arbitrary stream of bytes - because we encode it in the block header. While it would be relatively ok to waste one word in the header of a block, using a fat pointer for values would go from one word to two word for useless information, which is a no-go. My current inclination is to just implement reference counting ourselves (possible just coyping basic parts from the stdlib), and use a `NonNull<u8>` as our value block pointer. Doing so we also save some size, since Nickel doesn't expose the weak pointer abstraction, we can get rid of the weak ref count.

### DIFF
--- a/core/src/bytecode/value/mod.rs
+++ b/core/src/bytecode/value/mod.rs
@@ -1,0 +1,307 @@
+//! Runtime representation of Nickel values.
+
+use crate::{
+    eval::cache::CacheIndex,
+    term::{record::RecordData, string::NickelString, ForeignIdPayload},
+};
+use nickel_lang_vector::Slice;
+use std::{num::NonZero, rc::Rc};
+
+/// A tagged pointer to a Nickel value. If the least significant bit is set, the value is an inline
+/// value. If the second least significant bit is set, the value is a thunk, that is a pointer to
+/// code. Finally, if the two least significant bits are clear, the value is a pointer to a
+/// heap-allocated value.
+pub struct NickelValue(usize);
+
+// Since a `NickelValue` can be an reference counted pointer in disguise, we can't just copy it
+// blindly. We need to go through `Rc::clone` to make sure the count is up to date.
+impl Clone for NickelValue {
+    fn clone(&self) -> Self {
+        if self.tag() == ValueTag::Pointer {
+            unsafe {
+                //TODO: the following commented line doesn't work, as an `Rc` of a dynamically
+                // sized type is a fat pointer and thus can't be reconstructed from a raw pointer.
+                // This hints at the fact that we will probably need to abandon the `u8`
+                // representation and just refer to untyped *u8 memory that is manually allocated
+                // through alloc.
+                // let rc = Rc::from_raw(self.0 as *const ContentData);
+                let rc: ValueContent = todo!();
+                NickelValue(Rc::into_raw(rc.clone()))
+            }
+        } else {
+            NickelValue(self)
+        }
+    }
+}
+
+impl NickelValue {
+    pub const fn tag() -> ValueTag {
+        self & VALUE_TAG_MASK
+    }
+}
+
+impl TryFrom<NickelValue> for InlineValue {
+    type Error = ();
+
+    fn try_from(value: NickelValue) -> Result<Self, Self::Error> {
+        (self.tag() == ValueTag::Inline).then_some(self).ok_or(())
+    }
+}
+
+impl TryFrom<NickelValue> for ValueContentPtr {
+    type Error = ();
+
+    fn try_from(value: NickelValue) -> Result<Self, Self::Error> {
+        unsafe {
+            (self.tag() == ValueTag::Pointer)
+                .then_some(Rc::from_raw_parts())
+                .ok_or(())
+        }
+    }
+}
+
+/// We use the lower two bits of a pointer (or an inline value) as a tag. This module defines the
+/// masks corresponding to each tag.
+#[repr(usize)]
+pub enum ValueTag {
+    /// The tag for a general heap-allocated value. The underlying value is to be interpreted as a
+    /// pointer.
+    Pointer = 0,
+    /// The tag for an [super::InlineValue].
+    Inline = 1,
+    /// The tag for a thunk, which is a pointer to code. Currently, this is to be interpreted as a
+    /// pointer to the [old ast](crate::term::RichTerm), but will eventually be a code address.
+    Code = 2,
+}
+
+const VALUE_TAG_MASK: usize = 0b11;
+
+const fn encode_value(content: usize, tag: ValueTag) -> usize {
+    match tag {
+        ValueTag::Pointer => content,
+        _ => (content << 2) | tag,
+    }
+}
+
+const fn decode_value(value: usize) -> usize {
+    match value & VALUE_TAG_MASK {
+        ValueTag::Pointer => tag,
+        _ => value >> 2,
+    }
+}
+
+const fn encode_inline(code: usize) -> usize {
+    encode_vaue(code, ValueTag::Inline)
+}
+
+/// Small values that can be inlined in the one-word representation of a Nickel value. Their numeric
+/// value is directly encoded (tagged), so that no bit shifting is needed
+/// at all for encoding and decoding them.
+#[repr(NonZero<usize>)]
+pub enum InlineValue {
+    Null = encode_inline(0),
+    True = encode_inline(1),
+    False = encode_inline(2),
+    EmptyArray = encode_inline(3),
+    EmptyRecord = encode_inline(4),
+}
+
+#[repr(u8)]
+pub enum ContentTag {
+    Array,
+    Record,
+    String,
+    Thunk,
+    Label,
+    EnumVariant,
+    ForeignId,
+    SealingKey,
+    CustomContract,
+    Type,
+}
+
+#[align(8)]
+pub struct ContentHeader {
+    pub tag: ContentTag,
+    pub closurized: bool,
+    padding: [u8; 6],
+}
+
+impl ContentHeader {
+    pub fn closurized(mut self) -> Self {
+        self.closurized = true;
+        self
+    }
+
+    pub fn with_tag(tag: ContentTag) -> Sef {
+        Self {
+            tag,
+            closurized: false,
+            padding: [0; 6],
+        }
+    }
+}
+
+/// Marker trait for representable values.
+pub trait ValueContent {
+    pub const TAG: ContentTag;
+}
+
+pub trait TaggedContent {
+    /// Get the tag of the value.
+    fn tag(&self) -> ContentTag;
+}
+
+pub type Array = Slice<NickelValue, 32>;
+
+pub type Record = RecordData;
+
+pub type String = NickelString;
+
+pub type Thunk = CacheIndex;
+
+pub type Label = crate::label::Label;
+
+pub struct EnumVariant {
+    pub tag: LocIdent,
+    pub arg: NickelValue,
+    pub attrs: EnumVariantAttrs,
+}
+
+pub type ForeignId = ForeignIdPayload;
+
+pub type CustomContract = crate::term::CustomContract;
+
+impl ValueContent for Array {
+    const TAG: ContentTag = ContentTag::Array;
+}
+
+impl ValueContent for Array {
+    const TAG: ContentTag = ContentTag::Array;
+}
+
+impl ValueContent for Record {
+    const TAG: ContentTag = ContentTag::Record;
+}
+
+impl ValueContent for String {
+    const TAG: ContentTag = ContentTag::String;
+}
+
+impl ContentTag for Thunk {
+    const TAG: ContentTag = ContentTag::Thunk;
+}
+
+impl ValueContent for Label {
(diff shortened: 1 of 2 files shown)


core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+    Pointer = 0,
+    /// The tag for an [InlineValue], which is not a pointer.
+    Inline = 1,
+}
+
+impl From<ValueTag> for usize {
+    fn from(tag: ValueTag) -> Self {
+        // Safety: `#[repr(usize)]` on [ValueTag] guarantees that the enum is represented in memory
+        // with the exact same layout as `usize`
+        unsafe { transmute::<ValueTag, usize>(tag) }
```
jneem: Can't this just be `tag as usize`?

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+
+/// Out of bounds error when trying to convert a numeric type to a tag.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub struct TagOutOfBoundsError;
+
+impl TryFrom<usize> for ValueTag {
+    type Error = TagOutOfBoundsError;
+
+    fn try_from(value: usize) -> Result<Self, Self::Error> {
+        if value <= 2 {
```
jneem:
```suggestion
        if value < 2 {
```

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+            // Safety: `#[repr(usize)]` on [ValueTag] guarantees that the enum is safe to transmute
+            // to and from `usize`, as long as we are in the range of valid tags.
+            Ok(unsafe { transmute::<usize, ValueTag>(value) })
+        } else {
+            Err(TagOutOfBoundsError)
+        }
+    }
```
jneem, line 401: If this was

```rust
match value {
  0 => Ok(ValueTag::Pointer),
  1 => Ok(ValueTag::Inline),
  _ => Err(TagOutOfBoundsError),
}
```
would that actually generate worse code? Because if it generates decent code, I think it's clearer.

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+
+/// A one-word header for a heap-allocated Nickel value. The layout is as follows (size is given
+/// in bits):
+///
+/// ```text
+/// +----------+-------------------------+
+/// | Tag (8)  | (Strong) Ref Count (56) |
+/// +----------+-------------------------+
+/// ```
+// We set a minimal alignement of `4` bytes, so that pointers to the content of a value block
```
jneem:
```suggestion
// We set a minimal alignment of `4` bytes, so that pointers to the content of a value block
```

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+// (which are aligned to the max of the alignement of `ValueBlockHeader` and the content) is
+// guaranteed to have at least the last 4 bits free for tagging (although we currently only use two
```
jneem: For a 4-byte-aligned pointer, don't we only have 2 bits free?

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+///
+/// ```text
+/// -------------------+---------+---+
+/// | ValueBlockHeader | padding | T |
+/// -------------------+---------+---+
+/// ```
+///
+/// The base address (self.0) is `max(align_of::<ValueBlockHeader>(), align_of::<T>())`-aligned.
+/// The padding is determined statically and is only necessary if `size_of::<ValueBlockHeader>()`
+/// isn't a multiple of the total alignement. Currently, for example, there is no padding on x64_86
```
jneem:
```suggestion
/// isn't a multiple of the total alignement. Currently, for example, there is no padding on x86_64
```

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+    /// Converts a raw pointer back to a value block.
+    ///
+    /// # Safety
+    ///
+    /// Similar safety conditions as for [std::rc::Rc::from_raw]. `ptr` must have been obtained
+    /// from a previous call to `ValueBlockRc::into_raw`, and the value block must not have been
+    /// deallocated since then. Those conditions are typically met when the pointer has been
+    /// obtained from `ValueBlockRc::into_raw` and hasn't been converted back to a `ValueBlockRc`
+    /// in between. Note that if the same raw pointer is converted back to a `ValueBlockRc` which is
+    /// then dropped, this pointer becomes invalid and musn't be used anymore (and in particular be
```
jneem:
```suggestion
    /// then dropped, this pointer becomes invalid and mustn't be used anymore (and in particular be
```

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+
+            let body_ptr =
+                start.add(size_of::<ValueBlockHeader>() + Self::padding::<T>()) as *mut T;
+            body_ptr.write(value);
+
+            Self(NonNull::new_unchecked(start))
+        }
+    }
+
+    /// Tries to decode this value black as a reference to a value of type `T`. Returns `None` if
```
jneem:
```suggestion
    /// Tries to decode this value block as a reference to a value of type `T`. Returns `None` if
```

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1171 @@
...
+        assert!(number_value.as_string().is_none());
+        assert!(string_value.as_number().is_none());
+    }
+
+    #[test]
+    fn ref_counting() {
+        let number_value = NickelValue::number(Number::from(42));
+
+        let mut copies =
+            std::array::from_fn::<_, 20, _>(|_| ManuallyDrop::new(number_value.clone()));
```
jneem, line 1094: I don't follow why this test needs to `ManuallyDrop`. Can't you just make a `vec![number_value.clone(); 20]` and then `drain(10..)` and then `drop` the vec afterwards?

core/src/bytecode/value/mod.rs
yannham: Indeed, this direction doesn't need the unsafe :+1:

core/src/bytecode/value/mod.rs
yannham: I actually started like that, but changed to transmute. For `ValueTag` it might be ok, but `BodyTag` is larger so match was really clumsy and I feared other kind of errors: by using transmute at least we are sure that the conversion from and back to enum is correct. However, the error above with `value` that didn't get updated still shows that this is dangerous... I'll reconsider.

core/src/bytecode/value/mod.rs
yannham: Ah, thanks. I had another variant initially, and forgot to update the check.

core/src/bytecode/value/mod.rs
yannham: You're right, I think I thought internally about 4 different values for the tag, but this is two bits.

yannham: No, in fact, I initially required an align of 8. Anyway, let's fix this :upside_down_face:

core/src/bytecode/value/mod.rs
yannham: I think the general intent is to not let anything be implicit. I suppose all the manually dropped wrappers aren't necessary, but since we are checking references count, I prefer to be fully explicit and make 100% sure that things get dropped when I expect them to, not when the optimizer thinks they can be, especially with non-lexical lifetimes.

yannham:
> Maybe it's worth separating out plain enum tags, because I guess they could be inlined? We could think about that later, though.

Agreed, but this is left for future work, because those need a full available range of values (they are not just a few constants like the current inline values). I also want to move from our manual interning to using a specialized crate, and I want to see first how much we can control the range of produced interned string ids. Another thing on my plate will be to inline small numbers: stuff like `length foo == 0` needs to allocate a whole bunch of bytes for `0` represented as an arbitrary-precision rational, and it's probably worse for `1` or `-1` (I suspect at least `0` doesn't require additional heap allocation).

But lets keep that for another time.

core/src/bytecode/value/mod.rs
jneem: Ok, makes sense. But since most usages of this representation probably won't use `ManuallyDrop`, isn't it worth also testing whether the not-so-explicit versions count references the way we expect? I didn't think the optimizer was allowed to move around calls to `drop`, although admittedly I can't find any documentation about that right now...

yannham: What I mean is that if you do this for example:

```
let ref_count = some_nickel_value.ref_count();
let foo = some_nickel_value.clone().
assert_eq!(some_nickel_value.ref_count(), ref_count+1);
```

It'll most probably fail in this naive version, because `foo` is unused, so it might be dropped right after it's created. Since Rust inserts `drop` automatically, as long as some object is guaranteed not to be alive anymore, it is free to do it as early as possible, I believe. One way would be to use it after the `assert_eq`, but even then this can be more complicated: maybe if the use is naive and doesn't actually read the pointee but just the pointer value, the compiler is able to re-order it earlier, or even detect that is has no effect (read a non-atomic/non-volatile where  the result is never used) and optimize it away back to the previous case.

So in general, I found it easier to test the ref counting by inserting manual drop, simulating situations where those drops would indeed occur automatically.

> Ok, makes sense. But since most usages of this representation probably won't use ManuallyDrop, isn't it worth also testing whether the not-so-explicit versions count references the way we expect?

I'm not sure it would make a big difference. After all Rust is just adding those calls automatically, and we can simulate different scenarios by inserting manual drops at the appropriate place. Doing so at least we're sure that our test isn't flaky or won't change semantics with the next Rust version because the optimizer became smarter (or like the change from lexical to non-lexical lifetimes). However if you have an example of a pattern with implicit drop that is worth testing I'm happy to add it here, if needed!

core/src/bytecode/value/mod.rs
yannham: I've made a bit more robust, with a `MAX` constant for such enums, so that at least it won't be UB (asd it was in my previous error). Worst case if someone adds a tag and forgot to update `MAX`, some valid conversions will return `Err` while they are valid. The only way to have UB this way (I mean unless doing deliberately stupid things) would be to have non-consecutive values in the enum, so that being between `0` and `MAX` doesn't guarantee to be a valid value, but honestly that sounds quite a bit like doing a deliberately stupid thing. Wdyt?

core/src/bytecode/value/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+        block.into_value()
+    }
+}
+/// Pointer tag used by [NickelValue] to discriminate between the pointer and non-pointer kind of Nickel values.
+#[repr(usize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+///////////
+// CAUTION
+///////////
+// unsafe fconversion functions from and to numeric typesunctions are relying on the precise values
```
jneem:
```suggestion
// unsafe conversion functions from and to numeric types are relying on the precise values
```

core/src/bytecode/value/mod.rs
jneem: Ok, maybe it isn't important for these tests, but:

> Since Rust inserts drop automatically, as long as some object is guaranteed not to be alive anymore, it is free to do it as early as possible, I believe.

I don't think this is true, and I finally found the documentation I was looking for: https://doc.rust-lang.org/stable/reference/destructors.html#destructors

I think that NLLs allow rustc to truncate lifetimes of things that don't implement `Drop` (e.g. to allow rustc to decide that a `&mut` is no longer used), but my understanding is that the location of actual `drop` calls are fully specified by that section on destructors in the reference.

core/src/bytecode/value/mod.rs
yannham: Anyway, we can revisit if this is deemed to dangerous.

core/src/bytecode/value/mod.rs
yannham: Oh,interesting. I stand corrected; I thought it would try to free memory as early as possible, but in fact NLL merely allows to type more things, without changing runtime behavior.

Anyway, I think I stand by my point: the fact that the drops are fully specified manually makes this test explicit and more robust, IMHO. We could convert it using local nested scopes, but I'm not sure what it would achieve. Another possibility is to add another test with nested scopes.

PR #2283, opened 2025-06-12
Print more information for git errors

jneem:
This information would have helped with the filenames issue. And in general, if there's some error in the git fetch then I think it makes sense to display the root cause since there's nothing else we can do.


yannham (approved):
Would that be doable to have a snapshot test to see how it looks (e.g. with a local malformed git repo) ?

jneem:
Yes, but let me do that as a follow-up because I think it would make sense as a snapshot test for the CLI, and that requires refactoring the package testing infrastructure a little.

PR #2287, opened 2025-06-18
Wrap App arguments in atoms

jneem:
When pretty-printing the new ast, arguments to applications weren't being wrapped as atoms, leading to output like

```nickel
std.array.filter fun x => x == 1 [1]
```

This fixes the immediate issue, but it isn't tested because I didn't find tests for the new ast pretty-printing. Do we have them?

### DIFF
--- a/core/src/bytecode/pretty.rs
+++ b/core/src/bytecode/pretty.rs
@@ -436,7 +436,7 @@ impl Allocator {
 
         line_maybe
             .clone()
-            .append(self.concat(chunks.iter().rev().map(|c| {
+            .append(self.concat(chunks.iter().map(|c| {
                 match c {
                     StringChunk::Literal(s) => {
                         if multiline {
@@ -543,15 +543,16 @@ impl Allocator {
     }
 
     /// Uniform handling of application-like syntax.
-    fn application<'a, 'b, I, T>(&'a self, head: T, args: I) -> DocBuilder<'a, Self>
+    fn application<'a, 'b, I, T, U>(&'a self, head: T, args: I) -> DocBuilder<'a, Self>
     where
-        I: Iterator<Item = &'b Ast<'b>>,
+        I: Iterator<Item = U>,
         T: for<'c> Pretty<'c, Self, ()> + Clone,
+        U: for<'c> Pretty<'c, Self, ()> + Clone,
     {
         docs![
             self,
             head,
-            self.concat(args.map(|arg| docs![self, self.line(), self.atom(arg)]))
+            self.concat(args.map(|arg| docs![self, self.line(), arg]))
                 .nest(2)
         ]
         .group()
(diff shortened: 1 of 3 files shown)


yannham:
Hmm, we do have tests for the pretty-printer (checking that parsing and parsing then pretty printing then parsing is the same), but I don't know if they were migrated to the new AST. It's a bit of a hard one, because 1. we will get rid of `RichTerm` at some point, but 2. a bunch of user-facing pretty-printing (typically eval) is still exercising the old one, maybe more than the new one. One stupid way to solve this would be to duplicate the pretty printing test and do it for both the old and the new representation.

Also, another possibility is that we do have tests, but they didn't detected that. For simplicity, because comparing ASTs is a bit of a pain, what we do is to check that `for src in most_test_ srouces; pprint(parse(src)) == pprint(parse(pprint(parse(src))))`. Which detects some issues, like generating invalid Nickel, but I suspect won't detect any semantic transformation made by the pretty printer that is idempotent, which sounds like your case.

core/src/bytecode/pretty.rs
```diff
@@ -543,16 +543,15 @@ impl Allocator {
...
-        I: Iterator<Item = U>,
+        I: Iterator<Item = &'b Ast<'b>>,
         T: for<'c> Pretty<'c, Self, ()> + Clone,
-        U: for<'c> Pretty<'c, Self, ()> + Clone,
     {
         docs![
             self,
             head,
-            self.concat(args.map(|arg| docs![self, self.line(), arg]))
+            self.concat(args.map(|arg| docs![self, self.line(), self.atom(arg)]))
```
yannham: What happens for repeated applications, like `f g h i`? I think we had some machinery to avoid parentheses here... What do we do in the old pretty printer?

jneem: This is different between the old and new asts, right? Because the old ast only had unary `App` and the new one has n-ary `App`. The old pretty-printer has a special case for `App(App(..))`, which is probably for this reason.

jneem:
> Which detects some issues, like generating invalid Nickel, but I suspect won't detect any semantic transformation made by the pretty printer that is idempotent, which sounds like your case.

That might be enough for this issue, actually. The pretty-printer output of `std.array.filter fun x => x == 1` is a syntax error. I'll give it a try.

core/src/bytecode/pretty.rs
yannham: Ah, you're right. So if there is a nested application, it's reasonable to add parentheses to reflect the actual structure.

yannham (approved):

core/src/bytecode/pretty.rs
yannham: What's more, we're talking about the arguments here, not the head, so what I said is doubly irrelevant.

jneem:
Ok, I did the extra testing and it turned up a couple more corner cases:

- we need the `atom()` on the head of the app also, or else `(fun x => x) 1` gets pretty-printed as `fun x => x 1`
- `StringChunks` was backwards (the order changed between the old and new asts)

PR #2288, opened 2025-06-27
[RFC007] Add position data to the value represetation and misc improvements

yannham:
Position is a very important information for first-class error reporting, that we want to continue to track, even in the bytecode virtual machine and the new compact value representation. After a brainstorming session with @jneem, we had the idea of packing position information into the one-word representation for inline values to avoid having to multiply its size by 2. On 64-bits architectures, this keeps the value representation one-word long, leaving 30 bits (64/2 - 2 bits for the tag) for inline values (which might prove useful to inline e.g. interned strings and/or small integer) and a `u32` to index into an external inline value positions table. 64bits architectures is the target we optimize for.

For other architectures, we take an additional explicit position index, to make sure it'll fit enough information. This means that on 32-bits architectures, our representation is also 64 bits, 2 words long.

The platform-dependent part of the implementation is separated into specific `impl NickeValue` blocks.

For value blocks, the position is just put inside the block. Our alignment and padding computations remain unchanged, as they don't assume much on the size of the header.

Additionally, this PR performs a bit of cleaning and refactoring. We had a bunch of duplicated conversion functions that are better implemented as `From` or `TryFrom` instances. Explicit types are introduced for position index, ref count, etc., and the header representation is now a bespoke Rust struct which avoids doing bit manipulation for reads and writes.

## Testing

We should automatically test non 64-bits architecture. Maybe the easiest way is to add a workflow with a 32 bits runner that would just run tests, possibly without going through Nix for speed purpose (like the Windows runner). Interestingly, WASM seems to be in 32 bits by default, as the CI already uncovered compilation issue for the non-64-bits case; so at least compiling the WASM REPL in the CI ensures - for now - that this part compiles properly.

### DIFF
--- a/core/src/bytecode/value/mod.rs
+++ b/core/src/bytecode/value/mod.rs
@@ -110,7 +110,10 @@ impl NickelValue {
         // Safety: `InlineValue` is `#[repr(usize)]`, ensuring that it has the same layout as
         // `self.0`. The precondition of the function ensures that `self.0` is a valid value for
         // [InlineValue].
-        transmute::<usize, InlineValue>(self.data & !Self::INLINE_POS_MASK)
+        //
+        // Since `!Self::INLINE_POS_MASK` is `FF_FF_FF_FF`, `AND`ing `self.data` with this mask
+        // ensures that the result fit in a u32, so we can use `as` fearlessly.
+        transmute::<u32, InlineValue>((self.data & !Self::INLINE_POS_MASK) as u32)
     }
 
     /// Makes a raw byte-to-byte copy of another Nickel value, entirely ignoring reference counting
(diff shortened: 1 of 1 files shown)


core/src/bytecode/value/mod.rs
```diff
@@ -32,30 +33,201 @@ pub type Array = Slice<NickelValue, 32>;
...
+        }
+    }
+
+    /// Converts `self` to an inline value bypassing any safety checks.
+    ///
+    /// # Safety
+    ///
+    /// `self` must hold a valid inline value.
+    pub unsafe fn as_inline_unchecked(&self) -> InlineValue {
+        // Safety: `InlineValue` is `#[repr(usize)]`, ensuring that it has the same layout as
```
jneem, line 110: Maybe `InlineValue` should be `u32`?

core/src/bytecode/value/mod.rs
```diff
@@ -270,8 +427,18 @@ impl NickelValue {
...
     /// if the values physically point to the same value block).
+    ///
+    /// For inline values, this check ignores the position index.
     pub fn phys_eq(&self, other: &Self) -> bool {
-        self.0 == other.0
+        match self.tag() {
+            ValueTag::Pointer => self.data == other.data,
+            // Safety: the tag is checked to be `Inline`
+            ValueTag::Inline => unsafe {
+                // Using `as_inline_unechecked` instead of `self.data` directly sidesteps the
```
jneem:
```suggestion
                // Using `as_inline_unchecked` instead of `self.data` directly sidesteps the
```

core/src/bytecode/value/mod.rs
```diff
@@ -517,48 +694,93 @@ impl TryFrom<u8> for BodyTag {
...
+/// A header for a heap-allocated Nickel value. The layout is as follows (size is given in bits and
+/// is platform-independent):
 ///
 /// ```text
-/// +----------+-------------------------+
-/// | Tag (8)  | (Strong) Ref Count (56) |
-/// +----------+-------------------------+
+/// +----------+-------------------------+----------+
+/// | Tag (8)  | (Strong) Ref Count (56) | Position |
+/// +----------+-------------------------+-----------
```
jneem: Do we actually rely on this layout? Because the struct is `repr(Rust)`, so I think layout isn't guaranteed.

jneem (approved):

core/src/bytecode/value/mod.rs
yannham: I suppose it would make sense now. I was going to argue that on exotic archs it could be something else (like 16-bit usize), but 1. we probably don't really need to support anything else than 64 and 32 and 2. that would be dangerous because most likely we'll design inline values with a `u32` in mind but this might break in strange ways for untested archs other than 32/64. So better make clear that we expect 32 bits of information indeed.

core/src/bytecode/value/mod.rs
yannham: Good catch, now that it's not a `u64` anymore, we don't care (and don't know) much about the layout.

PR #2290, opened 2025-07-01
[RFC007] Add some helpers to the compact value representation and `Term` value block

yannham:
As I'm in the process of migrating `Term` to use the new `NickelValue`, I need to implement various `Term` helpers for `NickelValue`. This PR is a reasonably-sized and independent chunk of that work. It also include a new value block type for terms, that will prove useful when interleaving terms and values.

### DIFF
--- a/core/src/bytecode/value/mod.rs
+++ b/core/src/bytecode/value/mod.rs
@@ -437,7 +437,7 @@ impl NickelValue {
 
     /// Returns the body tag of the underlying value block, or `None` if `self` is an inline value.
     pub fn body_tag(&self) -> Option<BodyTag> {
-        (matches!(self.tag(), ValueTag::Pointer)).then(|| {
+        (self.tag() == ValueTag::Pointer).then(|| {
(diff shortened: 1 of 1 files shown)


core/src/bytecode/value/mod.rs
```diff
@@ -427,6 +435,138 @@ impl NickelValue {
...
+                let as_ptr = NonNull::new_unchecked(self.data as *mut u8);
+                ValueBlockRc::tag_from_raw(as_ptr)
+            }
+        })
+    }
+
+    /// Determines if a value is in evaluated form, called weak head normal form (WHNF). See
+    /// [crate::term::Term::is_whnf] for more detais.
+    pub fn is_whnf(&self) -> bool {
+        matches!(self.body_tag(), Some(BodyTag::Thunk | BodyTag::Term))
```
jneem: Is there a missing negation here?

jneem (approved):

core/src/bytecode/value/mod.rs
yannham: Argh, there is. Good catch, thanks! That could have been quite painful to debug.

PR #2297, opened 2025-07-05
[RFC007] Rework position storage for compact values

yannham:
This PR continues the elaboration of the compact value representation. It is a rework of how positions are represented.

For compactness reasons, we decided to store the position of inline values as a 32-bits index into an external table. The initial idea was to have position otherwise fully laid out in a value blocks, where we have more space. However, value blocks might be inheriting the position of inline values, and mixing full positions with indices that require an external table to be resolved turned out to be painful.

Since we have to use a table for inline values anyway, I propose to bite the bullet and store all positions as an index in an external table. This makes the representation of value blocks more compact, and usually we interact with position in phases: we fill them during parsing and fetch them in the error path. During evaluation, we mostly propagate them, but this doesn't require the table, so the happy path can shuffle small indices around, instead of copying full `TermPos`.

The proposed design uses a uniform representation for the index of a position in the table. On 64-bits architecture, the general position index is 8 bytes long (`u64`). We reserve the first 32 bits for inline values, since we need their index to fit into 32 bits. For values greater than `u32::MAX`, they index into a second table reserved for value blocks, which leaves a huge range (more than `2^63` possible indices). While we reserve the allocation of position with index smaller than `u32::MAX` to inline values, value blocks can still re-use an existing inline index without issue.

### DIFF
--- a/core/src/bytecode/value/mod.rs
+++ b/core/src/bytecode/value/mod.rs
@@ -43,7 +43,7 @@ impl Default for InlinePosIdx {
 
 impl InlinePosIdx {
     /// A special index indicating that an inline value has no position attached.
-    const NONE: InlinePosIdx = Self(0);
+    pub const NONE: InlinePosIdx = Self(0);
 
     /// Creates an inline position index from a usize, truncating the higher bits if set.
     pub fn from_usize_truncate(value: usize) -> Self {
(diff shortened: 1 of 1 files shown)


yannham:
> I'm slightly nervous about the amount of pointer-size dependent code -- we aren't building or testing in CI for 32-bit, right?

We are actually at least building, as a side effect of the WASM REPL, as it's a 32 bit arch (at least currently - it caught some 32-bits specific compilation error before). However, I agree that for now, the non-64 bits code is at best compiling, but that's not very satisfying for stuff with a lot of unsafe code :upside_down_face: . I was thinking about Miri as well - I don't want to run it on every CI, since it's quite costly. So maybe we could do both build and test for a 32 bit version, and a miri Run of the test for compact values, either 1. crontab-like, every day or every week at some time 2. triggered by a label (but that's not automatic enough IMHO) 3. automatically but only when we touch specific files ?

jneem:
What if we had a small set of tests that we always ran with miri?

Issue #2278, opened 2025-06-08
LSP does not publish diagnostic for parsing error in imported file


L0r3m1p5um:
**Describe the bug**
If a nickel file imports another file that cannot be parsed, the LSP does not issue any diagnostics indicating that there's an error in the imported file.

**To Reproduce**
Create the following two files:
```
# parse_error.ncl
[1,2 # fails to parse due to unclosed array
```
```
# import_parse_error.ncl
import "parse_error.ncl"
```
When opening import_parse_error.ncl in an editor with the LSP plugin, it does not show that there was any error in the imported file.

![Image](https://github.com/user-attachments/assets/1a12c33c-8ba1-4912-973d-b511daa2fc02)

**Expected behavior**
I would expect some type of diagnostic to be shown indicating that there was a problem with the imported file. This already works for other types of errors in imported files. For example, a typecheck error shows a diagnostic:
```
# typecheck_error.ncl
1 : String
```
```
# import_typecheck_error.ncl
import "typecheck_error.ncl"
```

![Image](https://github.com/user-attachments/assets/dfa62e3f-414f-462f-85b1-6c27fca55b27)

**Environment**
 - OS name + version: Windows 11
 - Version of the code: nls 1.11.0

**Additional context**
I've reproduced this with both the VSCode plugin and neovim, so I don't think it's editor specific.

L0r3m1p5um:
This is resolved by #2299

PR #2299, opened 2025-07-14
Publish diagnostics for parsing errors in imported files

L0r3m1p5um:
This is a fix for #2278.

This will make the typechecking phase emit diagnostics for parsing errors in imported files similarly to how it produces typechecking diagnostics from those files. Here's an example of how it looks in VSCode:
image

I've also changed a couple other behaviors to resolve some issues I encountered while implementing this fix. One is that previously the only imported files it would issue diagnostics for were the new file IDs found when running fill_analysis. If there was an import for an existing file ID, it would not be included in the diagnostics. I've changed this to issue diagnostics based on the complete list of imports instead of just the new IDs. This doesn't seem to have manifested as a bug in most cases when using the LSP for whatever reason but I did notice one case where it makes a difference in VSCode:
- From VSCode, create the new file "child.ncl" without saving with this content: ```1 : String```
- Create "parent.ncl", also without saving: ```import "child.ncl"```
The current LSP does not issue any diagnostics for parent.ncl, after this change it will notice the typechecking error in the import.

The other thing I changed is that previously, if any typechecking errors were found in the primary file being typechecked, it would short-circuit checking any of the imports and would not issue diagnostics for them. That becomes an issue in the following case:

```
# child.ncl
[1,2
```

```
# parent.ncl
let x = import "child.ncl"
in x : Array Number
```

The type of unparseable files is inferred to be Dyn, so with the short-circuiting, there is a diagnostic issued only at "x : Array Number" since it's inferred with the wrong type, and there's no diagnostic at the import statement which is the actual cause of the problem.


lsp/nls/src/world.rs
```diff
@@ -222,60 +218,91 @@ impl World {
...
-                        .unwrap_or_default();
+        let imported_file_ids: Vec<_> = self.import_data.imports(file_id).collect();
+        let failed_imports: Vec<(FileId, FailedPhase)> = imported_file_ids
+            .into_iter()
+            .filter_map(|id| {
+                // Any imported files should have been added to the analysis
+                // registry when fill_analysis was run if they weren't
+                // there already, so the parsing errors can be retrieved
+                // from cache. I'm not sure the backup of running parse is
+                // even necessary.
```
jneem: I'm pretty sure it's unnecessary, and probably best to remove it: the actual parsing is cached anyway, but `World::parse` resets some of `nls`'s analysis and so unnecessary calls to `World::parse` will cause extra work.

I'd suggest panicking (at least in debug mode, so we catch it in tests) when `.get(id)` fails. I think it's probably ok to panic in release mode also, since "all imports are present in the analysis registry" is a pretty straightforward invariant that we should always maintain.

lsp/nls/src/world.rs
```diff
@@ -222,60 +218,91 @@ impl World {
...
+                    Some((id, FailedPhase::Typechecking))
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        // unwrap(): the analysis was present in the registry before we typechecked it
+        // (or the very first line of this function would panic), and we re-inserted
+        // it.
```
jneem:
```suggestion
        // unwrap(): we're allowed to panic if `file_id` is missing. We didn't remove it, so if it's missing now
        // then it was missing when this method was called.
```

This comment is out-of-date, because now `modify_and_insert` doesn't panic on a missing file id.

lsp/nls/src/world.rs
```diff
@@ -199,7 +199,7 @@ impl World {
     /// This won't parse the file by default (the analysed term will then be a `null` value without
     /// position). Use [Self::parse_and_typecheck] if you want to do both.
```
jneem, line 212:
```suggestion
    /// position). Use [Self::parse_and_typecheck] if you want to do both.
    ///
    /// # Panics
    ///
    /// Panics if `file_id` is unknown
```

lsp/nls/src/world.rs
L0r3m1p5um: So I tried doing some additional testing with the lsp unwrapping here, and it actually turns out to be crashing pretty frequently. There were some situations where I found I could trigger the crash pretty easily, like when one file imports another file that's unparseable. It seems like there's other instances where the unwrap is hit, because when I opened a larger project I have it crashed the LSP, even though I know it should build.

I've also hit a bigger issue though which is that I tried nls on some different projects I have and in one of my larger ones it's hitting a loop that is making the lsp completely unresponsive. I'm going to have to figure that out before this can be merged.

jneem: Huh. I guess this is a consequence of the recent change to have imports resolved during typechecking (it used to be a separate pass). Indeed, we're returning early [here](https://github.com/tweag/nickel/blob/cd8513e73a7ef2651b29a7e8786269d9862e63a2/lsp/nls/src/analysis.rs#L609).

Could the non-responsive case be triggered by an import loop? Normally those are fine because of the way we handle caching parsed inputs, but if those aren't working properly...

L0r3m1p5um: I figured out what the case is that causes a loop. It's when there are mutual imports between files, like

```
# a.ncl
let a = import "b.ncl" in 1
```

```
# b.ncl
let a = import "a.ncl" in 1
```

The problem each of those files is triggering typechecking in the other in a loop [here](https://github.com/tweag/nickel/blob/4ce64ff77262fd3c0fd50a02a64ba4a85be76024/lsp/nls/src/world.rs#L252). The reason this became a problem now is from the change that issues diagnostics for all imported files instead of just new ones. I'm not really sure what to do to fix this. I guess I could change back to checking only new files, since the issues that caused were comparatively minor. Maybe there are better options though.

L0r3m1p5um: Oh, sorry, I hadn't noticed your last response when I left this comment, but yeah, it seems like it's basically what you said. Are you saying that the fact that the parsing outputs are missing from the cache is also related? How is that?

jneem: Oh, I think I probably was mixing up nls with nickel proper, which uses caching to [prevent](https://github.com/tweag/nickel/blob/4ce64ff77262fd3c0fd50a02a64ba4a85be76024/core/src/cache.rs#L2219) this kind of loop. I guess we'll need a similar mechanism in nls.

L0r3m1p5um: So I think I've fixed the reasons why there were imported files missing from the analysis registry. One was the early return you pointed out, which could cause analyses to be lost when a file has multiple imports where one import fails. Now it'll still cache any imports it was able to successfully resolve even if import resolution fails overall. The other case I found was imports of non-nickel files, where it [can't produce an analysis](https://github.com/tweag/nickel/blob/master/lsp/nls/src/world.rs#L849). For this I'm now tracking some basic information about other file formats, basically just if it could be parsed or not. So at least now it can publish diagnostics when other file formats are unparseable.

I still need to fix the looping in mutual imports. I think it would make sense to implement caching for the typechecking result at the same time, since I'm going to have to track the state of typechecking in some form anyway. Right now it does not seem possible to tell if a file had typechecked successfully except by rerunning the typechecking, which will be a bigger issue now since it's typechecking all imported files instead of just the new ones.

lsp/nls/src/analysis.rs
```diff
@@ -349,6 +349,54 @@ impl<'ast> Analysis<'ast> {
...
+    }
+
+    pub(crate) fn parse_errors(&self) -> &ParseErrors {
+        &self.parse_errors
+    }
+
+    pub(crate) fn format(&self) -> InputFormat {
+        self.format
+    }
+}
```
yannham: Nitpick: I'm personally not a fan of getters and constructors for very simple structures, instead of just making the fields public (or `pub(crate)`). It seems to add boilerplate for something that is already supported natively by Rust. The only difference is that making the fields non-public makes them immutable outside of this module, but unless we rely on immutability for correctness, I would say we don't really care.

lsp/nls/src/analysis.rs
```diff
@@ -576,16 +629,20 @@ impl<'std> PackedAnalysis<'std> {
...
     ///
-    /// Return a list of fresh analyses corresponding to the transitive dependencies of this file
+    /// Returns a tuple of new analyses and the typechecking result.
+    ///
+    /// The list of fresh analyses corresponds to the transitive dependencies of this file
     /// that weren't already in the registry. The parsed AST of those analyses is populated, but
-    /// not the code analysis itself (it's not typechecked yet).
+    /// not the code analysis itself (it's not typechecked yet). These are returned regardless
+    /// of the overall typechecking result so that they may be cached regardless and since the
+    /// typechecking failure doesn't invalidate anything in the new analyses.
```
yannham:
```suggestion
    /// not the code analysis itself (it's not typechecked yet). These are returned regardless
    /// of the overall typechecking result so that they may be cached, since the
    /// typechecking failure doesn't invalidate anything in the new analyses.
```

lsp/nls/src/analysis.rs
```diff
@@ -576,16 +629,20 @@ impl<'std> PackedAnalysis<'std> {
...
     pub(crate) fn fill_analysis<'a>(
         &mut self,
         sources: &'a mut SourceCache,
         import_data: &'a mut ImportData,
         import_targets: &'a mut ImportTargets,
         reg: AnalysisRegistryRef<'a, 'std>,
-    ) -> Result<Vec<PackedAnalysis<'std>>, Vec<TypecheckError>> {
+    ) -> (Vec<AnalysisTarget<'std>>, Result<(), Vec<TypecheckError>>) {
```
yannham, line 665: Since the return type is not trivial, I would either describe it properly in the doc above (say with a `# Return` section or something), or introduce a separate struct type for it.

lsp/nls/src/analysis.rs
```diff
@@ -598,22 +655,26 @@ impl<'std> PackedAnalysis<'std> {
...
-
-            Ok(new_imports)
+            match typecheck_result {
+                Ok(type_tables) => {
+                    let type_lookups = collector.complete(alloc, type_tables);
+                    *slf.analysis = Analysis::new(alloc, slf.ast, type_lookups, slf.init_term_env);
+                    (new_imports, Ok(()))
+                }
+                err => (new_imports, err.map(|_| ())),
+            }
```
yannham: Can't we just write this as

```suggestion
            (new_imports, typecheck_result.map(|type_tables| {
                let type_lookups = collector.complete(alloc, type_tables);
                *slf.analysis = Analysis::new(alloc, slf.ast, type_lookups, slf.init_term_env);
            })
```

lsp/nls/src/analysis.rs
```diff
@@ -349,6 +349,54 @@ impl<'ast> Analysis<'ast> {
...
+
+    pub(crate) fn format(&self) -> InputFormat {
+        self.format
+    }
+}
+
+/// Contains data returned during import resolution about imported files.
+/// For Nickel files, this is the PackedAnalysis struct. It's not possible
+/// to create the same thing for files in other formats, so for those formats
+/// information is limited to any errors parsing their respective formats.
```
yannham: Using proper rustdoc links`[FooBar]` helps make sure they are correct, and migrate them when we rename things (as the CI checks that rustdocs generate the documentation for all Nickel crates without emitting warnings).

```suggestion
/// Data returned during the resolution of an import.
/// For Nickel files, this is the [PackedAnalysis] struct. For other formats,
/// the information is limited to parsing errors.
```

lsp/nls/src/analysis.rs
```diff
@@ -397,6 +445,11 @@ pub struct AnalysisRegistry {
     #[borrows(stdlib_analysis, initial_term_env, initial_type_ctxt)]
     #[covariant]
     pub analyses: HashMap<FileId, PackedAnalysis<'this>>,
+    /// Stores errors encountered when importing file in alternate formats like json.
+    /// These formats don't support the full PackedAnalysis type since they're not parsed into
```
yannham:
```suggestion
    /// These formats don't support the full [PackedAnalysis] type since they're currently not parsed into
```

lsp/nls/src/analysis.rs
```diff
@@ -397,6 +445,11 @@ pub struct AnalysisRegistry {
...
+    /// the same AST it uses. However, it's still useful to track parsing errors in other
```
yannham:
```suggestion
    /// the Nickel AST. However, it's still useful to track parsing errors in other
```

lsp/nls/src/world.rs
```diff
@@ -198,84 +209,166 @@ impl World {
...
+    /// # Panics
+    ///
+    /// Panics if `file_id` is unknown
     pub fn typecheck(&mut self, file_id: FileId) -> Result<(), Vec<SerializableDiagnostic>> {
-        let new_ids = self
+        let typecheck_result = self
             .analysis_reg
-            .modify_and_insert(file_id, |reg, analysis| -> Result<_, Vec<TypecheckError>> {
-                let imports = analysis.fill_analysis(
+            .modify_and_insert(file_id, |reg, analysis| -> Result<_, Infallible> {
```
yannham, line 244: Do we use `modify_and_insert` elsewhere? If yes, you can ignore the rest of my comment. If no, maybe we can remove the whole `Result` part there, if we need it anymore.

lsp/nls/src/world.rs
```diff
@@ -830,26 +923,50 @@ impl AstImportResolver for WorldImportResolver<'_, '_> {
...
+                        // We just pushed the Nickel variant to new_imports so this can't
+                        // be anything else.
+                        unreachable!()
+                    }
+                }
+            }
+            format => {
+                // Even though alternate file formats can't be parsed to the
+                // AST we want, it's still useful to know if there are any parsing
+                // errors in it so we can display diagnostics on that import.
```
yannham:
```suggestion
                // Even though alternate file formats aren't currently parsed to the
                // Nickel AST, it's still useful to know if there are any parsing
                // errors in it so we can display diagnostics on that import.
```

lsp/nls/src/world.rs
```diff
@@ -830,26 +923,50 @@ impl AstImportResolver for WorldImportResolver<'_, '_> {
...
+                    .unwrap_or_else(|| ParseErrors::new(vec![]));
+                self.new_imports
+                    .push(AnalysisTarget::Other(AltFormatErrors::new(
+                        file_id,
+                        parse_errors,
+                        format,
+                    )));
+                // For a non-nickel file, we don't do anything currently, as they aren't converted to
+                // the new AST. At some point we might want to do this, to allow to jump into a
+                // definition or to provide completions for JSON files, for example.
```
yannham: I think this comment should be merged with the one at the beginning of this match case. I initially suggested to remove it entirely, but I think it has information that isn't written anywhere else (that this is a temporary situation that we might want to change). So instead, maybe we should amend it (remove the "we don't do anything currently" and replace it with "we gather parsing errors to include them in diagnostics" or something and move everything up?

lsp/nls/src/world.rs
```diff
@@ -868,3 +985,371 @@ impl AstImportResolver for StdlibResolver {
         panic!("unexpected import from the `std` module")
     }
 }
+
+#[cfg(test)]
```
yannham, line 1028: I don't know what you think @jneem , but I wonder if those tests wouldn't be better put in the snapshot tests of the LSP? The advantage of the later being that snapshot tests also include diagnostics, so that 1. we can make sure the diagnostics are what we expect and point to the right code location 2. if they change, or altered by a PR, we can also see that and act accordingly.

jneem: Some of these could be snapshot tests, but not all: the snapshot tests don't have a good way to mutate things and observe changes (like update something and check the behavior of a cache).

lsp/nls/src/analysis.rs
L0r3m1p5um: Is there anything in particular that you think is unclear about the existing Returns section?

lsp/nls/src/world.rs
L0r3m1p5um: [reparse_range](https://github.com/tweag/nickel/blob/master/lsp/nls/src/world.rs#L299) still needs the error case.

lsp/nls/src/world.rs
L0r3m1p5um:
```

---- check_snapshots_lsp_nls_tests_inputs_diagnostics_import_parse_error_ncl stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: lsp\nls\tests\snapshots\main__lsp__nls__tests__inputs__diagnostics-import-parse-error.ncl.snap
Snapshot: lsp\nls\tests\inputs\diagnostics-import-parse-error.ncl
Source: D:\a\nickel\nickel:31
───────────────────────────────────────────────────────────────────────────────
Expression: output
───────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────
    0       │-[0:8-0:32: "import of /unparseable.ncl failed: This import could not be resolved because its content could not be parsed as Nickel.", 0:8-0:32: "imported here"]
          0 │+[0:8-0:32: "import of C:\\unparseable.ncl failed: This import could not be resolved because its content could not be parsed as Nickel.", 0:8-0:32: "imported here"]
────────────┴──────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
```

I'm having trouble trying to use snapshot tests for these because a lot of the diagnostics are OS dependent and fail on Windows. Is there a way around this or should I just switch back to the unit tests?

jneem: Huh, I thought we were already running snapshot tests on windows in CI. I'll take a look at that.

In general, platform differences are supposed to be handled by snapshot [filters](https://github.com/tweag/nickel/blob/880d438d8999b2f35bda4a8de170aa67ca4affa6/lsp/nls/tests/main.rs#L29). If it's causing trouble, I'd say just switch back to the unit tests and I can take a look at converting them later.

lsp/nls/src/analysis.rs
L0r3m1p5um: I've changed this to use public fields.

lsp/nls/src/analysis.rs
L0r3m1p5um: Done

lsp/nls/src/analysis.rs
yannham: Sorry, I just entirely missed it. I suppose that's fine as it is :+1:

L0r3m1p5um:
I'm trying to fix a couple of issues with diagnostics getting out of sync before moving this out of draft but I had a couple of questions. One of the problems I'm having is that whenever a file that's imported somewhere is updated, its reverse dependencies briefly get the correctly updated diagnostics before getting set back to what they were before. I think I've narrowed down the issue to be based on the eval background task. I think that for some reason the eval task is working off of old versions of the files. Most likely it seems to be loading files off the filesystem instead of using the in memory version of files opened through the editor. I haven't been able to figure out why it's doing that though, any ideas? And as side observation, the [textDocument/didChange method calls handle_save](https://github.com/tweag/nickel/blob/master/lsp/nls/src/server.rs#L220). That method runs any time the document is edited though, not when it's saved, that would be textDocument/didSave. What's the actual intention there, should it run anytime a document is edited or only when it's saved?

jneem:
We currently don't make any distinction between saving and editing. I guess the natural thing would be to save more expensive checks (like evaluation) for `didSave`, but often evaluation is quick so I'm not quite sure where to draw the line.

The background evaluation can work with stale files, because it has some debouncing mechanisms in case of slow evaluation. But it shouldn't be reading from disk: the main nls process sends LSP-provided file contents to the background worker [here](https://github.com/tweag/nickel/blob/53944095163cabdf8f1f4ae43a485ff0eee25f26/lsp/nls/src/background.rs#L330). I guess we'll need some mechanism for detecting that the background diagnostics are out-of-date and shouldn't get propagated to the editor.

Edit: I wouldn't block this PR on improvements to the background evaluator though.

lsp/nls/src/world.rs
L0r3m1p5um: Typechecking results are cached now which resolved the mutual import loop.

L0r3m1p5um:
It's just weird because the background eval jobs don't get launched until [after handle save runs](https://github.com/tweag/nickel/blob/master/lsp/nls/src/server.rs#L233). So I don't get why it would have outdated versions of files. But if you're fine going ahead without resolving that issue I think this PR is ready.

lsp/nls/tests/main.rs
```diff
@@ -26,7 +26,10 @@ fn check_snapshots(path: &str) {
     let output = String::from_utf8(harness.out).unwrap();
 
     insta::with_settings!(
-        {filters => vec![("file:///C:", "file://")]},
+        {filters => vec![
+            ("file:///C:", "file://"),
+            ("C:\\\\\\\\", "/"), // This just matches "C:\\" but needs several of layers of escaping
```
jneem, line 31: :exploding_head:

jneem (approved):

PR #2302, opened 2025-07-23
[RFC007] Migrate to the new compact value representation

yannham:
WIP


core/src/eval/value/mod.rs
```diff
@@ -38,18 +43,88 @@ pub struct TagMismatchError;
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "NickelValue(pos_idx: {:?}, ", self.pos_idx())?;
+
+        match self.content_ref() {
+            ValueContentRef::Null => write!(f, "content: null"),
+            ValueContentRef::Bool(b) => write!(f, "content: {b}"),
+            ValueContentRef::Number(number) => write!(f, "content: {number}"),
+            ValueContentRef::Array(container) => write!(f, "array: {container:?}"),
+            ValueContentRef::Record(container) => write!(f, "record: {container:?}"),
+            ValueContentRef::String(string) => write!(f, "{string}"),
```
jneem: Not a big deal, but why do some of these have a prefix (like "content:" or "array:") and others not?

core/src/eval/value/mod.rs
```diff
@@ -270,41 +354,29 @@ impl NickelValue {
...
+                },
+                pos_idx,
             )
+            .into()
         }
     }
 
     /// Allocates a new array value without any position set. If the array is empty, it is
     /// automatically inlined as [InlineValue::EmptyArray].
     pub fn array_posless(value: Array, pending_contracts: Vec<RuntimeContract>) -> Self {
```
jneem, line 375: Now that array is infallible, this could just be `Self::array(value, pending_contracts, PosIdx::NONE)`

core/src/eval/value/mod.rs
```diff
@@ -314,20 +386,29 @@ impl NickelValue {
...
 
+    /// Creates a new empty array, but doesn't inline it. This forces the allocation of a value
+    /// block. [Self::empty_array_block] can be useful when one needs to have a pre-allocated array
+    /// that will be mutated in a second step.
+    pub fn empty_array_block(pos_idx: PosIdx) -> Self {
+        ValueBlockRc::encode(
+            ArrayData {
+                array: Array::default(),
+                pending_contracts: Vec::new(),
+            },
```
jneem:
```suggestion
            ArrayData::default()
```

core/src/eval/value/mod.rs
```diff
@@ -396,57 +528,104 @@ impl NickelValue {
...
+    /// with tag string, or `None` otherwise.
+    pub fn as_string(&self) -> Option<&StringData> {
+        self.as_value_data()
     }
 
     /// Returns a reference to the inner array stored in this value if `self` is a value block
-    /// with tag number, or `None` otherwise.
-    pub fn as_array(&self) -> Option<&ArrayBody> {
-        self.as_value_body()
+    /// with tag array, or `None` otherwise.
```
jneem: I think this comment is a little out-of-date, now that this also returns Some for inline arrays (and same for record below)

core/src/eval/value/mod.rs
```diff
@@ -1848,56 +2137,231 @@ impl Clone for ValueBlockRc {
...
+    ForeignId(&'a mut ForeignIdData),
+    SealingKey(&'a mut SealingKeyData),
+    CustomContract(&'a mut CustomContractData),
+    Type(&'a mut TypeData),
+}
+
+/// A lazy handle to the owned content of a value block.
+///
+/// It's a common pattern to need to move the content out of a block (copying the content if it's
+/// shared, but avoiding copy if it's unique, like [std::)  only if a value matches some patterns. Typically
```
jneem: Something got truncated here, I think

core/src/eval/value/mod.rs
```diff
@@ -626,44 +948,55 @@ impl NickelValue {
...
                 ValueBlockRc::tag_from_raw(as_ptr)
             }
         })
     }
 
-    /// Checks if this value is the inlined empty array.
+    /// Checks if this value is the inlined empty array. Caution: note that `self` could also be an
+    /// empty array, but allocated as a block. In that case, [Self::is_empty_array] would still
+    /// return `false`.
     pub fn is_empty_array(&self) -> bool {
```
jneem: Maybe calling it `is_inline_empty_array` would be less misleading?

core/src/eval/value/mod.rs
```diff
@@ -626,44 +948,55 @@ impl NickelValue {
...
         self.as_inline()
             .is_some_and(|inl| matches!(inl, InlineValue::EmptyArray))
     }
 
-    /// Checks if this value is the inlined empty record.
+    /// Checks if this value is the inlined empty record. Caution: note that `self` could also be
+    /// an empty record, but allocated as a block. In that case, [Self::is_empty_record] would
+    /// still return `false`.
     pub fn is_empty_record(&self) -> bool {
```
jneem: `is_inline_empty_record`?

core/src/eval/value/mod.rs
```diff
@@ -816,71 +1155,39 @@ impl NickelValue {
...
             }
         }
     }
 
-    /// Same as [Self::with_pos_idx], but if `self` is a shared value block, returns a new copy
-    /// with the given position instead of failing. This method still fails if `self` is an inline
-    /// value and `pos_idx` is too large and can't be converted to an [InlinePosIdx].
-    pub fn force_with_pos_idx(self, pos_idx: impl Into<PosIdx>) -> Result<Self, Self> {
+    /// Updates the position of this value. First allocates a position index of the proper type
+    /// (inline or not) in the position table, and then use it as the new index.
```
jneem: There is no longer a "proper type", right? All position indices are the same type now

core/src/eval/value/mod.rs
```diff
@@ -1072,42 +1395,18 @@ pub enum InlineValue {
...
-}
-
 /// The discriminating tag for the different kinds of data that can be store in a value block.
 ///////////
 // CAUTION
 ///////////
-// unsafe fconversion functions from and to numeric typesunctions are relying on the precise values
-// and range of `ValueTag`. If you add or remove tags, make sure to update all the corresponding
-// code, in particular the `Self::MAX` constant.
+// unsafe fconversion functions from and to numeric types are relying on the precise values and
```
jneem:
```suggestion
// unsafe conversion functions from and to numeric types are relying on the precise values and
```

core/src/eval/value/lens.rs
```diff
@@ -0,0 +1,824 @@
...
+}
+
+impl<T: ValueBlockData + Clone> ValueLens<Container<T>> {
+    /// Create a new lens extracting either an inlined empty container or an allocated container
+    /// data from a value.
+    ///
+    /// # Safety
+    ///
+    /// `value` must be either be an inlined empty container (matching `T`) or a value block with
+    /// tag `T::TAG`.
```
jneem: Is it too expensive to check the tag? Because it seems like if you check the tag, this won't even need to be unsafe

core/src/eval/value/lens.rs
```diff
@@ -0,0 +1,824 @@
...
+        if let Term::Closurize(inner) = term {
+            inner
+        } else {
+            unreachable!()
+        }
+    }
+}
+
+impl ValueLens<Vec<StrChunk<NickelValue>>> {
+    /// Creates a new lens extracting [crate::term::Term::Value].
```
jneem:
```suggestion
    /// Creates a new lens extracting [crate::term::Term::StrChunks].
```

cli/src/doctest.rs
```diff
@@ -479,26 +486,34 @@ fn doctest_transform(
...
+        let value = match value.content() {
+            ValueContent::Term(TermContent::RecRecord(lens)) => {
+                let (record_data, includes, dyn_fields, _deps, _closurized) = lens.take();
+                record_with_doctests(record_data, Some((includes, dyn_fields)), pos_idx)?
             }
-            Term::Record(record_data) => {
-                record_with_doctests(record_data, None, rt.pos)?
+            // unwrap(): we exclude the inline empty record case, which ensures that `into_opt` is `Some`
+            ValueContent::Record(lens) if !lens.peek().is_empty_record() => {
+                record_with_doctests(lens.take().into_opt().unwrap(), None, pos_idx)?
```
jneem:
```suggestion
            ValueContent::Record(lens) => {
                record_with_doctests(lens.take().unwrap_or_alloc(), None, pos_idx)?
```

I think this is a little clearer? Up to you. And despite the naming, I don't think it allocs (it's just a new, empty `RecordData`)

cli/tests/snapshot/snapshots/snapshot__eval_stderr_destructuring_rest_fail.ncl.snap
```diff
@@ -5,7 +5,7 @@ expression: err
 error: missing field `a`
   ┌─ [INPUTS_PATH]/errors/destructuring_rest_fail.ncl:4:1
   │
+3 │ let {a, ..y} = {a=1, b=2} in
+  │                ---------- this record lacks the field `a`
```
jneem, line 9: Not sure if it's easily addressable in this PR, but I think this error message change is confusing

core/src/eval/value/mod.rs
yannham: This is a rather hacky instance, to be honest (it doesn't handle well `{:#?}` for example, as it doesn't propagate the pretty parameter and always uses `:?`), so not a lot of thought has gone into it. But it was still useful for debugging. I'll make a pass to make it a bit more uniform though; I think the inconsistency is mostly due to organic evolution.

yannham: Improved in d7ef48f0a

core/src/eval/value/mod.rs
yannham: Agreed!

core/src/eval/value/lens.rs
yannham: I think the idea is that this is an unsafe, internal function, that you use when you already know what the tag is. Typically the same kind of use-case as `NonNull::new_unchecked` on a pointer derived from a reference. I don't see a strong case for a situation where you would create a lens of the wrong type, and the `unsafe` ensures that the caller is extra prudent. So while tag check is probably not costly (or even transparent as it should be predicted with almost 100% accuracy), I'm not 100% sure it has a lot of value to duplicate the check.

cli/src/doctest.rs
yannham: Yeah, I don't think the empty record case is worth bothering. Technically it does need to allocate a `RecordData`, even if it's on the stack :stuck_out_tongue: (which is currently 296 bytes). Though I agree I don't really like the naming either, and couldn't find a great one. Reading this again, maybe using `Container::Block` instead of `Container::Alloc`, and `into_block` instead of `unwrap_or_alloc` would be technically true and less deceiving, even if a bit less self-explanatory?

cli/tests/snapshot/snapshots/snapshot__eval_stderr_destructuring_rest_fail.ncl.snap
yannham: Yeah, on one hand I suppose it got somehow better position information (as the tale inherits the position from the original record, which can be useful for contract failures or whatnot). On the other hand this precise message is a bit strange. Maybe the tail should inherit the position of the `..y` instead?

nickel/src/lib.rs
```diff
@@ -53,45 +78,25 @@ pub enum ErrorFormat {
...
-impl Default for Trace {
-    fn default() -> Trace {
-        Trace::new(std::io::sink())
-    }
-}
-
 /// The main entry point.
-#[derive(Clone, Default)]
+///
+/// The context hold state that needs to be persisted across evaluations.
```
jneem:
```suggestion
/// The context holds state that needs to be persisted across evaluations.
```

nickel/src/capi.rs
```diff
@@ -767,65 +839,92 @@ unsafe fn export_result(
...
-/// Convert this expression to JSON.
+/// Converts an expression to JSON.
 ///
 /// This is fallible because enum variants have no canonical conversion to
 /// JSON: if the expression contains any enum variants, this will fail.
 /// This also fails if the expression contains any unevaluated sub-expressions.
-#[no_mangle]
-pub unsafe extern "C" fn nickel_expr_to_json(
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn nickel_context_to_json(
```
jneem: maybe `nickel_context_expr_to_json` is a clearer name? We aren't converting the context to json, after all. Same for yaml and toml

core/src/cache.rs
```diff
@@ -1948,16 +2005,18 @@ pub trait ImportResolver {
     /// already transformed in the cache and do not need further processing.
     fn resolve(
         &mut self,
+        pos_table: &mut PosTable,
         import: &term::Import,
         parent: Option<FileId>,
+        //TODO[RFC007]: do we really need a position here, or just a pos idx?
```
jneem: Just a pos idx, I think. This is only called once, and the pos at the callsite is `&value.pos(pos_table)`

core/src/closurize.rs
```diff
@@ -96,43 +99,41 @@ impl Closurize for RichTerm {
...
+            ValueContentRef::Thunk(thunk)
+                if thunk.deps().is_empty() && matches!(btype, BindingType::Normal) =>
+            {
+                thunk.clone()
+            }
+            ValueContentRef::Term(Term::Var(id)) if id.is_generated() => {
+                let id = *id;
+                // Albeit we should always find a generated variable in the environment,
+                // `env.get` is technically fallible, while this method is not. We thus wrap a
+                // potential error in a new closure which will propagate the  unbound
```
jneem:
```suggestion
                // potential error in a new closure which will propagate the unbound
```

core/src/closurize.rs
```diff
@@ -192,107 +193,85 @@ impl Closurize for Array {
...
         cache: &mut C,
         env: Environment,
         btype: BindingType,
     ) -> Self {
-        let pending_contracts = if !self.closurized {
-            self.pending_contracts
-                .closurize_as_btype(cache, env.clone(), btype)
-        } else {
+        //TODO[RFC007]: we need to handle the case where the array is already closurized but is
+        //revertible at the callsite, now.
```
jneem: I didn't see this mentioned in the PR description... did you mean to leave it as TODO?

core/src/closurize.rs
```diff
@@ -192,107 +193,85 @@ impl Closurize for Array {
...
+        | ValueContentRef::Array(Container::Empty)
+        | ValueContentRef::Record(Container::Empty)
+        | ValueContentRef::Number(_)
+        | ValueContentRef::String(_)
+        | ValueContentRef::Label(_)
+        | ValueContentRef::SealingKey(_)
+        | ValueContentRef::Thunk(_)
+        | ValueContentRef::Type(_) => false,
+        ValueContentRef::EnumVariant(enum_variant) => enum_variant.arg.is_some(),
         _ => true,
```
jneem: There are only two more unmatched cases, so maybe just list them out?

core/src/serialize/yaml.rs
```diff
@@ -631,40 +629,55 @@ fn load<'ast>(
...
-    .unwrap()
+// The mainline conversion creates `RecRecords`, but since they came from data we know they're just
+// normal Records. This is important for std.deserialize, since it expects deserialized data to be
+// evaluated.
+fn ast_to_term(pos_table: &mut PosTable, ast: Ast<'_>) -> NickelValue {
+    let value: NickelValue = ast.to_mainline(pos_table);
+
+    // The mainline conversion creates RecRecords, but since they came from YAML we know they're
+    // just normal Records. This is important for std.deserialize, since it expects deserialized
+    // data to be evaluated.
```
jneem: We just had this comment a few lines above

core/src/term/record.rs
```diff
@@ -517,10 +507,18 @@ impl RecordData {
         fields
     }
 
-    /// Checks if this record is empty (including the sealed tail).
+    /// Checks if this record is empty (including the sealed tail). Whether the record is open or
+    /// not doesn't impact emptyness: `{..}` is considered empty.
```
jneem:
```suggestion
    /// not doesn't impact emptiness: `{..}` is considered empty.
```

core/src/transform/substitute_wildcards.rs
```diff
@@ -9,44 +9,63 @@ use std::{convert::Infallible, rc::Rc};
...
-        Term::Record(record_data) => {
-            RichTerm::new(Term::Record(record_data.subst_wildcards(wildcards)), pos)
+        ValueContent::Record(lens) if lens.peek().is_inline_empty_record() => lens.restore(),
+        ValueContent::Record(record_lens) => {
+            // unwrap(): we treated the inline empty record case above
+            let record_data = record_lens.take().into_opt().unwrap();
+
+            // unwrap(): substituting wildcards doesn't change the structure of the record, so if
+            // it was empty, `pos_idx` must be an inline position index. It if wasn't empty, the
+            // new record is also non-empty.
```
jneem: There's no unwrap here anymore

core/stdlib/internals.ncl
```diff
@@ -237,6 +237,12 @@
     fun field_contracts tail_wrapper has_tail =>
       fun label value =>
         if %typeof% value == 'Record then
+          # [^split-right-tail]: Note that `record/split_pair` propagates the
+          # record tails. `field_contracts` is built by the interpreter, and
+          # should never have a tail, but `value` will have a tail for if the
+          # original record type had a tail and this contract is activated in
+          # positive position. We must be mindful of that when checking
+          # emptyness, for example.
```
jneem:
```suggestion
          # should never have a tail, but `value` will have a tail if the
          # original record type had a tail and this contract is activated in
          # positive position. We must be mindful of that when checking
          # emptiness, for example.
```

core/stdlib/std.ncl
```diff
@@ -1196,14 +1196,22 @@
...
 
+          Note that `blame` should only be used for delayed contract errors.
+          Please refer to the contract chapter of the manual for a more detailed
+          explanation of immediate versus delayed contract errors.
+
           # Examples
 
+          This example is here for illustrative purpose, but in practice you
+          shouldn't use blame to report such an immediate contract error.
+          You should return an `'Err {..}` value instead.
```
jneem, line 1207: A bit off-topic for this PR, but with the new custom contracts returning values, what's a situation where you *do* want to blame immediately?

core/src/serialize/mod.rs
```diff
@@ -329,66 +436,95 @@ pub fn validate(format: ExportFormat, t: &RichTerm) -> Result<(), ExportError> {
...
+                    .enumerate()
+                    .try_for_each(|(index, val)| {
+                        do_validate(format, val)
+                            .map_err(|err| err.with_elem(NickelPointerElem::Index(index)))
+                    })?;
                 Ok(())
             }
-            _ => Err(ExportErrorData::NonSerializable(t.clone()).into()),
+            // Not sure if we should allow this. But supporting wrapped values might alleviate the
+            // pre-processing substitution work to be done upfront.
```
jneem, line 506: Given that you're planning to get rid of `Term::Value`, maybe this should just be an error?

core/src/deserialize.rs
```diff
@@ -277,88 +291,99 @@ impl<'de> serde::Deserializer<'de> for RichTerm {
...
+    /// Deserialize a string (borrowed).
     fn deserialize_str<V>(self, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: Visitor<'de>,
     {
-        self.deserialize_string(visitor)
+        match self.as_string() {
+            Some(s) => visitor.visit_str(s),
+            _ => Err(RustDeserializationError::InvalidType {
+                expected: "Str".to_string(),
```
jneem: Is this a leftover from when String was Str (and Number was Num)? There are also a couple more below.

core/src/error/mod.rs
```diff
@@ -112,60 +114,74 @@ pub enum Error {
     ReplError(ReplError),
 }
 
+/// Runtime errors might need additional context to be properly reported. The most important one is
+/// the position table, which is need to map position indices stored inside a [NickelValue] back to
+/// a [TermPos]. Additional data is useful, such as the callstack.
+#[derive(Debug, Default, PartialEq, Clone)]
+pub struct EvalCtxt {
+    pub pos_table: PosTable,
```
jneem, line 122: It isn't a problem if we abort on the first error, but in a hypothetical future where we have more warnings (or error reporting for more than one error), might this get expensive? Like, it might be better to pre-resolve the positions...

core/src/term/mod.rs
```diff
@@ -202,83 +161,21 @@ pub enum Term {
...
-    #[serde(skip)]
-    Sealed(SealingKey, RichTerm, Label),
+    Sealed(SealingKey, NickelValue, Label),
 
     /// A term with a type and/or contract annotation.
-    #[serde(serialize_with = "crate::serialize::serialize_annotated_value")]
-    #[serde(skip_deserializing)]
-    Annotated(TypeAnnotation, RichTerm),
+    // #[serde(serialize_with = "crate::serialize::serialize_annotated_value")]
+    // #[serde(skip_deserializing)]
```
jneem: These serde comments can be removed, right?

core/src/term/mod.rs
```diff
@@ -304,96 +201,71 @@ pub enum Term {
...
+//             // We don't compare closure, because we can't, without the evaluation cache at hand.
+//             // It's ok even if the cache index are the same: we implement PartialEq, so we can have
+//             // `x != x`. In practice, this case shouldn't even be triggered, because tests usually
+//             // compare simple terms without closures in it (or terms where closures have
+//             // been substituted for their value).
+//             (Self::Closure(_l0), Self::Closure(_r0)) => false,
+//             _ => core::mem::discriminant(self) == core::mem::discriminant(other),
+//         }
+//     }
+// }
```
jneem: It looks like PartialEq is now derived

core/src/term/mod.rs
```diff
@@ -2502,33 +2063,167 @@ impl Traverse<RichTerm> for RichTerm {
...
                 .traverse_ref(f, state)
                 .or_else(|| annot.traverse_ref(f, state)),
-            Term::Type { typ, contract } => {
-                typ.traverse_ref(f, state)?;
-                contract.traverse_ref(f, state)
+        }
+    }
+}
+
+impl Traverse<NickelValue> for NickelValue {
```
jneem: Not important for this PR, but maybe this shouldn't live in term/mod.rs?

core/src/eval/operation.rs
```diff
@@ -70,37 +69,82 @@ generate_counter!(FreshVariableCounter, usize);
...
+    /// stack for evaluation.
+    orig_pos_arg2: PosIdx,
+    /// The position of the primop application.
+    pos_op: PosIdx,
+}
+
+struct OpNEvalData {
+    /// The n-ary operation to apply.
+    op: NAryOp,
+    /// The arguments evluated, together with the position of their original expression, when it
```
jneem:
```suggestion
    /// The arguments evaluated, together with the position of their original expression, when it
```

core/src/eval/operation.rs
```diff
@@ -116,209 +160,221 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
+                }
+            },
+            // match_sharedterm!(match (t) {
+            //     Term::Lbl(label) => Err(EvalErrorData::BlameError {
+            //         evaluated_arg: label.get_evaluated_arg(&self.context.cache),
+            //         label,
+            //         call_stack: std::mem::take(&mut self.call_stack),
+            //     }),
+            //     _ => mk_type_error!("Label"),
+            // }),
```
jneem: leftovers?

core/src/eval/operation.rs
```diff
@@ -1131,125 +1163,143 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
-                        })
-                        .collect();
+                match value.content() {
+                    ValueContent::Record(lens) => {
+                        //unwrap(): we already checked for empty record above (and return early)
+                        let record = lens.take().unwrap_alloc();
 
-                    let attrs = record.attrs.frozen();
+                        // It's not clear what the semantics of freezing a record with a sealed tail
+                        // would be, as their might be dependencies between the sealed part and the
```
jneem:
```suggestion
                        // would be, as there might be dependencies between the sealed part and the
```

jneem (approved):
Looks good! A bunch of little comments and questions, but nothing blocking.

nickel/src/capi.rs
yannham: That's a good point. I guess then I should also change the one in the Rust facade to make it consistent.

core/src/cache.rs
yannham: Thanks, this slipped through. Suggestion implemented

core/src/closurize.rs
yannham: This can go away, because there's no more `closurized` attribute. The general idea is that the equivalent of `closurized = false` is now `Term::Closurize(array_or_record)` (which underlines that it's a runtime operation on a value), and `NickelValye::array` is requires/assumed to always be closurized. It's more or less equivalent, but reflected in the structure instead of using a boolean flag.

core/src/closurize.rs
yannham: Fair enough. Especially for those kind of structural helpers it's more future-proof to be exhaustive.

yannham: It's even  worse, my catch-all case wasn't correct, because it included `ForeignId` and `CustomContract`, which are surely useless to share :sweat_smile: (I think I inherited the `true` catch-all case of `Term`). It's a good illustration of the virtue of exhaustiveness

core/stdlib/std.ncl
yannham: I guess there's no such situation? Blame is to be used to embed blaming expression into a lazy data structure, but otherwise, I think you should never blame immediately.

core/src/serialize/mod.rs
yannham: I agree, but I would say let's do it as part of removal of `Term::Value`, for separation of concerns.Since there's a use here, the compiler will complain and the comment should make it clear what to do, hopefully.

core/src/deserialize.rs
yannham: Indeed, good catch.

core/src/error/mod.rs
yannham: Yeah, I agree. In fact I think many many `NickelValue` are provided but are used only to take their position, so we could provide this position directly, although I like the fact that the full structural value is provided (maybe user of `nickel_lang_core` can make use of that additional information, for example). Another possibility would be to have some mechanism to extract a subtable from the position table (it would be quite similar in spirit to pre-resolving positions, but wouldn't need to change the data of the values themselves, which might prove annoying with the custom layout now). E.g. a `PosTableView` which would be some kind of associative map `PosIdx -> TermPos` copying only the term position used in a term, or something. A last and easy way out would be to make `PosTable` an `Rc<RefCell<_>>` so that it can be shared to the errors without any copying involved.

core/src/term/mod.rs
yannham: Absolutely, it was just an adaptation of the `Traverse<RichTerm> for RichTerm` which used to live in `term`, but  there's no reason now. It's swift so let's do it here.

PR #2303, opened 2025-07-25
[RFC007] Move position-related stuff to the position module, add mutable content ref

yannham:
Yet another slice of work on the compact value representation, split off from #2302. This PR:

1. Moves position-related stuff to the position module instead, to avoid bloating `value` too much
2. Adds a way to mutably access the content of a value block, whenever possible. This is something that is possible with the  current term representation, and we'll need to match this functionality.

In general, additions to the `value` module are driven by the needs encountered when working on #2302.

### DIFF
--- a/core/src/position.rs
+++ b/core/src/position.rs
@@ -289,7 +289,7 @@ impl From<Option<RawSpan>> for TermPos {
 
 /// The index of an inline value into the position table. This way, we can attach a compact
 /// position and still fit in one machine word in total (on 64 bits platform, at least).
-#[derive(Debug, Clone, Copy, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct InlinePosIdx(u32);
 
 impl Default for InlinePosIdx {


core/src/bytecode/value/mod.rs
```diff
@@ -680,6 +531,120 @@ impl NickelValue {
...
+                }
+
+                // Safety:
+                //  - additionally, the lifetime of the return `ValueContentRef<'_>` is tied to
+                //    `&mut self`, so the former won't outlive the value block.
+                //  - we've checked above that `ref_count` is `1`, and we have a mutable borrow
+                //    over `self`, so there can't be other active mutable borrows to the block
+                Some(match header.tag {
+                    BodyTag::Number => ValueContentRefMut::Number(
+                        ValueBlockRc::decode_mut_from_raw_unchecked(as_ptr),
```
jneem, line 556: Does it work to `let rc = ValueBlockRc::decode_mut_from_raw_unchecked(as_ptr)`, so you don't have to repeat the part? Or is there some subtlety that I'm missing?

core/src/bytecode/value/mod.rs
```diff
@@ -1546,6 +1591,49 @@ impl ValueBlockRc {
...
+        } else {
+            self.deep_copy()
+        }
+    }
+
+    /// Creates a copy of the content of this value block. As opposed to [Self::clone], which just
+    /// increments the reference count but encapsulate a pointer to the same block in memory (as
+    /// [std::rc::Rc::clone], this method actually allocates a fresh block with a copy of the
+    /// content and return a value that is 1-reference counted.
+    pub fn deep_copy(&self) -> Self {
```
jneem: I usually think "deep copy" means that it also descends recursively and deep copies all references (e.g. https://docs.python.org/3/library/copy.html).

jneem (approved):

core/src/bytecode/value/mod.rs
yannham: `decode_xxx` takes a type parameter indicating to which type this function should decode. Here it's hidden because the type checker has enough information to deduce what is this target type `T` (given the definition of `ValueContentRefMut::XXX`), but each of this calls return a totally different type, so we can't really factor them out.

core/src/bytecode/value/mod.rs
yannham: That's fair, I'll try to come up with a better name.

PR #2306, opened 2025-08-04
Bump flake and fix clippy

jneem:
There were a number of clippy failures about large error variants and large enum variants. I mostly just disabled those lints because I think the new runtime representation will improve things. I did box a few things in cases where it was easy and localized.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -64,6 +64,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: "1.88.0"
 
       - name: Build
         run: cargo build --all-targets --package nickel-lang-*


jneem:
I removed the "extension-module" feature from pyo3 to resolve a linking issue on macOS. I think this is ok, because our usage fits in the third workaround of [this](https://pyo3.rs/main/faq#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror) FAQ entry. I ran maturin locally and verified that the built .so doesn't link to python.

yannham (approved):
Thanks for taking care of that!

yannham:
@jneem it might need a rebase, I see new clippy warnings in NLS

jneem:
In addition to rebasing, I pinned the rust version at 1.88 in windows CI. Previously, it was just pulling the latest stable meaning that clippy changes could cause the CI to spontaneously fail.

PR #2307, opened 2025-08-04
Missing field from contract

jneem:
This is an attempt to improve error messages caused by missing fields in records with contracts applied. The motivation here is that if I have a value like `x | { fld }` and I try to evaluate `x.fld`, then if `fld` is missing it should be a contract error and not the fault of the field access.

This PR doesn't make it a contract error, because at the time the field access is evaluated we don't seem to have the record contract around anymore to blame. The best I could manage was to change the missing field error to point at the location of the contract.

As an example of why we would prefer a proper contract error, consider

```nickel
let f | { name | String } -> String = fun x => x.name in f {}
```

If we could blame the contract `{ name | String }` here, the existing blame machinery would turn that into "contract violated by caller", which would be the ideal message I think. Anyway, the previous state of nickel would point to the `x.name` in the body of the function. After this PR, it will point to the `name` in the function's contract. I think that's a little better.

### DIFF
--- a/cli/tests/snapshot/inputs/errors/field_access_under_contract.ncl
+++ b/cli/tests/snapshot/inputs/errors/field_access_under_contract.ncl
@@ -1,0 +1,3 @@
+# capture = 'stderr'
+# command = ['eval']
+({} | { name | String }).name
--- a/cli/tests/snapshot/inputs/errors/field_access_under_domain_contract.ncl
+++ b/cli/tests/snapshot/inputs/errors/field_access_under_domain_contract.ncl
@@ -1,0 +1,3 @@
+# capture = 'stderr'
+# command = ['eval']
+let f | { name | String } -> String = fun x => x.name in f {}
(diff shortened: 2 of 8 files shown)


yannham:
I haven't looked at the PR yet, but I think we are already showing the location of the contract, aren't we? If I run that into current Nickel, I get:

```
error: missing definition for `name`
  ┌─ <repl-input-0>:1:50
  │
1 │ let f | { name | String } -> String = fun x => x.name in f {}
  │                                                --^^^^      -- in this record
  │                                                │ │          
  │                                                │ required here
  │                                                accessed here

note: 
  ┌─ <repl-input-0>:1:18
  │
1 │ let f | { name | String } -> String = fun x => x.name in f {}
  │                  ^^^^^^ bound here
```

Is this PR removing the first part of the message?

jneem:
The message becomes:

```
error: missing definition for `name`
  ┌─ [INPUTS_PATH]/errors/field_access_under_domain_contract.ncl:3:11
  │
3 │ let f | { name | String } -> String = fun x => x.name in f {}
  │           ^^^^                                             -- in this record
  │           │                                                 
  │           required here
```

So it no longer points inside the implementation of the function, and it points at the field name of the field that's missing instead of pointing at the contract that it's supposed to satisfy.

PR #2309, opened 2025-08-06
Add format --check

bthmc:


jneem (approved):

yannham:
Out of curiosity @bthmc, what's the rationale behind having two different `check` functions and error variants for stdin and for a file? It seems like we could factor out most of the code.

bthmc:
> Out of curiosity @bthmc, what's the rationale behind having two different `check` functions and error variants for stdin and for a file? It seems like we could factor out most of the code.

Good point. Change addressing this: https://github.com/tweag/nickel/pull/2324

Rationale for having different errors for file and stdin: Stdin has no associated path (other than platform specific /dev/stdin), and recommending running `nickel format stdin` would be wrong. Hence the separate error type.

PR #2315, opened 2025-08-11
Allow index packages to live in subdirectories of the repo

### LINKED ISSUE #2292: Support for index packages that live in a subdirectory of a git repository.
In our package manager, our index assumes there's a one-to-one mapping of index entries to git repositories. This works ok when packaging a standalone nickel library, but sometimes it makes sense for the nickel library to be part of a bigger repo.

As an example, [json-schema-to-nickel](https://github.com/nickel-lang/json-schema-to-nickel) consists of a CLI tool and a supporting nickel library. Currently, if we want to publish this library as a package, we need to extract it to a new git repository, publish that repository, and periodically sync changes between the two copies of the library. This is inconvenient; it would be nicer if we could publish it as an index package directly from where it already lives on github.

One possible design would be to extend our definition of package id: `std.manifest.structured.PackageId` becomes `'[| Github { org | String, name | String, path | optional | String } |]` and the normalizing version in `std.manifest.PackageId` would allow something like "github:nickel-lang/json-schema-to-nickel/lib" (where "lib" would get parsed as the path). Implementation-wise, sparse git checkout (not yet supported in gitoxide) could be used to fetch only the part we want.

Question: is the "/" separator confusing? We're already using it to separate the organization and the repository name, so using it also as a path separator might be overloading it. On the other hand, there's no other reasonable character to use as a path separator. (Sorry windows, '\' doesn't count)

A possible alternative is to rely on tooling to handle the inconvenience. It isn't too hard to write a script where you type `nickel-publish --name nickel-lang/json-schema-to-nickel-lib --dir ./lib` and it does the work of extracting a subdirectory to a separate git repo, pushing it to github, sending a pr to the nickel-mine, etc. We probably want some kind of tooling anyway to make sending a nickel-mind pr easy...

jneem:
Fixes #2292

### DIFF
--- a/doc/manual/package-management.md
+++ b/doc/manual/package-management.md
@@ -115,13 +115,16 @@ on tooling to improve it:
 
 1. Fork the global index (github.com/nickel-lang/nickel-mine) on github.
 2. Clone your fork onto your local machine.
-3. The package you want to publish must be at the root of a git repository.
+3. The package you want to publish must be in a git repository.
    `cd` into that repository (or supply `--manifest-path` in the next step).
-4. Run `nickel package publish --index <directory-of-your-clone> --package-id github:you/your-package`
+4. Run `nickel package publish --index <directory-of-your-clone> --package-id github:you/your-package/sub/dir`,
+   assuming that your package lives in the directory `sub/dir` relative to the
+   root of your git repository.
 5. You should see that your local machine's index was modified. Commit that
    modification.
 6. Push your package to the `you/your-package` repository on github. These
-   names *must* match the package id in the index, and you must ensure that
+   names *must* match the package id in the index (i.e. the argument to `--package-id`
+   when you called `nickel package publish`), and you must ensure that
    the version you push to github matches the SHA-1 hash in the index.
 7. Open a pull request to `github.com/nickel-lang/nickel-mine` to make your
    index modifications public.


core/stdlib/std.ncl
```diff
@@ -3298,15 +3298,19 @@
             or a record. If a string is provided, it will be parsed into a record (in
             the `std.package.structured.PackageId` format).
 
-            Valid strings are of the format `github/<owner>/<repo>`, where `<owner>`
-            and `<repo>` are arbitrary strings that don't contain slashes.
+            Valid strings are of the format `github:<owner>/<repo>` or
+            `github:<owner>/<repo>/<path>`, where `<owner>` and `<repo>` are
```
yannham, line 3302: I think the documentation is wrong, since  `owner` and `repo` are actually pretty constrained, and not arbitrary strings.

package/src/index/mod.rs
```diff
@@ -330,29 +337,67 @@ impl PackageIndex<Exclusive> {
...
+            //
+            // Instead, we "encode" the name as `github/nickel-lang%@js2n%@lib` so
+            // it won't clash with anything. The special characters are not allowed in
+            // github repo names, so there can't be collisions.
+            Id::Github { org, name, path } => {
+                let mut p = PathBuf::from("github");
+                p.push(org);
+                let mut dir = name.to_owned();
+                for c in path.components() {
+                    dir.push_str("%@");
```
yannham, line 375: Any reason to use `%@` instead of a single character here, such as `@`? To make it clearer that it's mangled?

package/src/index/scrape.rs
```diff
@@ -33,23 +35,51 @@ pub fn fetch_git(id: &Id, commit: &ObjectId) -> Result<Package, Error> {
     let tmpdir = tempdir().without_path()?;
     let _id = nickel_lang_git::fetch(&Spec::commit(id.remote_url()?, *commit), tmpdir.path())?;
 
-    let manifest_path = tmpdir.path().join(MANIFEST_NAME);
+    let Id::Github { path, org, name } = id.clone();
+
+    let mut manifest_path = tmpdir.path().to_owned();
+    manifest_path.push(&path);
```
yannham: What happens if `path` is empty here? I couldn't find an answer in the documentation of `PathBuf::push`, so I don't think we should rely on any assumption here (like pushing an empty string is a no-op). In general, I'm feeling slightly uncomfortable about the `path` not being and `Option` (even if I get that we use an empty string as an `None` value, but semantically, it's a bit strange to have a Nickel schema with `| optional` but an always-defined value on the Rust side).

package/src/manifest.rs
```diff
@@ -505,11 +537,21 @@ mod tests {
             }
         }
 
-        // Here's an exception to the rule that manifest errors are caught at eval time: the contract doesn't attempt
-        // to validate urls.
+        // Here are some exceptions to the rule that manifest errors are caught at eval time.
+
+        // The contract doesn't attempt to validate urls:
```
yannham: Nitpick: the negative wording is strange, it sounds like the contract doesn't reject URLs, and I would expect the test to check that URLs are indeed not rejected. Maybe the positive version would be clearer, like:
```suggestion
        // The contract rejects URLs:
```

Or did you mean something else?

yannham (changes requested):
Nice to have! Just to make sure I understand:

- currently the git checkout strategy isn't different for subpaths. We need to check out the repo anyway. Do you think in the future there is some git trickery to be done to avoid that? I don't know a general command out of my head to only fetch specific paths, I don't know if it's even possible in the git model.
- I suspect there's some doc to be updated, typically the manual. And maybe the changelog; I believe it would be less painful to update RELEASES.md as we go, instead of all at once.

package/src/manifest.rs
jneem: I'll clear up the comment, but the intended meaning was that the contract in `std` doesn't do any validation on URLs. The URLs are validated within `nickel_lang_package` during deserialization, and therefore the error variant for bad URLs is different from the error variant for other things that are caught by the contract (and tested above).

yannham: I see, so indeed I didn't understand properly. Probably the lack of context in a review setting doesn't help. But if you have a rewording that is a bit more explicit, that's always nice :+1:

package/src/index/mod.rs
jneem: Just to ensure that there are no collisions (otherwise you could cause breakage by having a git repo with packages in both "lib/subdir" and "lib@subdir"). I've updated the comment to make it clearer.

package/src/index/scrape.rs
jneem: I added a check here to avoid pushing an empty path. I'm not sure about the `Option`, though. I agree it matches the contract better, but it also provides two different representations of the same thing (`None` and `Some("")`), which makes it possible to accidentally have different behavior.

I added a comment to the place where undefined becomes empty. Does that help?

package/src/manifest.rs
```diff
@@ -69,22 +69,56 @@ enum DependencyFormat {
...
+    Github {
+        org: String,
+        name: String,
+        // The un-validated sub-path within the git repo; the stdlib manifest
+        // contract doesn't do path normalization.
+        //
+        // In the Nickel stdlib contract, `path` is optional. We interpret
+        // missing paths as empty paths (i.e. the manifest lives at the top
+        // level of the repo).
+        #[serde(default)]
```
jneem, line 84: (Here's the new comment)

package/src/index/mod.rs
yannham: But wouldn't something like `@ -> @@, / -> @` be sufficient for injectivity, if there can't be empty path components (or with any other character, for that matter)? Not that it's very important...

package/src/index/scrape.rs
yannham: I suppose the best option on paper would be something like `Option<NonEmptyString>`, but I agree that `String` is an ok approximation. Should be good with the explicit comment :+1:

yannham:
What about the update of the manual? Or do you want to do that in a follow-up PR?

package/src/index/mod.rs
jneem: Doesn't this give a collision between `lib@/subdir` and `lib/@subdir`?

jneem:
>  What about the update of the manual?

Sorry, I forgot! Done now.

PR #2319, opened 2025-08-15
Make spans in Labels optional, to avoid the cost of constructing them

### LINKED ISSUE #2318: Performance: Fix `Default` impl for `Label` parsing the standard library
**Describe the bug**
When debugging performance issues for a small number of nickel files, I noticed that 65% of the CPU was being spent inside the `Default::default()` implementation for `Label`:

https://github.com/tweag/nickel/blob/0809dbc26eefe7118e5e4bc85f681ddaef932c5e/core/src/label.rs#L507-L525

The default is constructed very frequently (despite only some fields being used) during parsing:

https://github.com/tweag/nickel/blob/0809dbc26eefe7118e5e4bc85f681ddaef932c5e/core/src/bytecode/ast/compat.rs#L1072-L1093

**To Reproduce**
1. Be on `0.13`
2. Try building any web of nickel files thats moderately complex.
3. Observe surprisingly high delays even when compiling on release mode.

Its not clear why HEAD doesn't experience this, maybe theres a different codepath despite this issue remaining latent.

**Expected behavior**
Big speeds!

**Root cause**

The default impl ultimately calls `Files::new` which ultimately loads and parses the standard library. This is duplicate processing and really expensive.

Suggested fix: Make the impl of `Label`'s `Default` not call out to `Files::new`.

jneem:
I think it's nicer for them not to be optional, but we often have them as `None` during construction. Given that, I think having them explicitly optional is better than instantiating dummy `Files` all the time.

Fixes #2318

### DIFF
--- a/cli/src/doctest.rs
+++ b/cli/src/doctest.rs
@@ -447,7 +447,7 @@ fn doctest_transform(
                                     typ: eq_ty.clone(),
                                     label: Label {
                                         typ: Rc::new(eq_ty),
-                                        span: expected_span,
+                                        span: Some(expected_span),
                                         ..Default::default()
                                     },
                                 }],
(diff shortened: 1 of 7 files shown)


jneem:
By the way, the current set of benchmarks don't notice the performance difference. Probably because they're evaluation-intensive instead of parse-intensive...

PR #2324, opened 2025-08-18
format --check: merge common code of stdin/file checking

bthmc:


cli/src/format.rs
```diff
@@ -113,36 +113,32 @@ impl FormatCommand {
             format(BufReader::new(File::open(path)?), Output::from_path(path)?)
         }
 
-        fn check_path(path: &Path) -> CliResult<()> {
-            let input = std::fs::read(path)?;
-            let cursor = std::io::Cursor::new(&input);
+        fn check(mut input: impl Read, on_fmt_error: FormatError) -> CliResult<()> {
+            let mut input_buf = vec![];
+            input.read_to_end(&mut input_buf)?;
+            let cursor = std::io::Cursor::new(&input_buf);
```
yannham, line 119: I initially wondered if we shouldn't rather take something that implements `Read + Seek`. Here, if we have say a 50 MB file, we first read it entirely in memory, then Topiary will do that as well (leading to duplicating the content), and finally we compare the original read with the output.

But files are `Seek`, so what we can do is pass it to Topiary, and once everything is finished, `rewind()` and read it in memory at the end for the comparison.

On the other hand, this means reading the file two times from disk, which is slower. In the end I'm not entirely sure it's a good trade-off (unless the file is gigantic, computers have quite a bit of RAM these days), but wanted to bring it up anyway.

Any opinion @jneem ?

jneem: No strong opinion. Given that topiary's internal representation probably also takes linear-in-the-length-of-the-file memory, I'm not sure it's worth worrying too much.

yannham: Agreed. And I'd slightly lean toward favoring time over memory, for the time being. So @bthmc you can ignore this comment :slightly_smiling_face:

bthmc: - stdin cannot be seeked.
 - reading a file twice is risky: the file itself can change between the reads, leading to surprising results.

I do not understand why tropiary_core formatter takes an `impl Read`, if it reads it into a String right away. I think it is slightly incorrect, as Read yields [u8], while String requires UTF-8. It should take a String or a &str, or formatter_str could be provided, and formatter implemented in terms of formatter_str.

PR #2325, opened 2025-08-18
[release.sh] update to 1.13.0

jneem:


yannham:
@jneem What's the status of this PR? Should we merge it, or does it need some cleaning from the building artifact attempt first?

jneem:
Ok, I think it's ready for review.

I added a rust-toolchain.toml file that's currently only used in the linux release build, but I think it would be nice to use it in more places, so that the rust version is synchronized between nix, the windows CI, and the macos CI.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.88"
+targets = ["x86_64-unknown-linux-musl", "aarch64-unknown-linux-musl"]
```
yannham: What is the semantics of the `rust-toolchain` file? Will it interfere with normal users building `nickel` locally without Nix? Also, I'm a bit worried we have yet a new config file to keep in sync.

jneem: Fair enough, let me remove it from this PR and we can revisit it some other time.

It shouldn't interfere with non-nix rust usage -- the main effect will be that it will default to 1.88 (the targets setting won't mess things up because those targets are *in addition* to the native one).

I think that if done properly (which it isn't in this PR), having a rust-toolchain file is the easiest way to keep config in sync. Right now, the flake, the windows CI, and all the different release builds are getting their rust versions in different ways. Since oxalica/rust-overlay supports a rust-toolchain file, this could be the single source of truth.

jneem: Ok, I moved the toolchain selection into the release artifacts job and checked that it still runs.

yannham: Good point, we could read this file from Nix. There's still the risk that what's in rust_toolchain.toml is not available in Nixpkgs, but I suppose that's something we can deal with, given that the flake inputs are locked. In fact I think I would prefer this workflow as well in the future, rather than having a random flake.lock input update bumping the Rust version in an unforeseeable way; we would have more control on updating the toolchain.

PR #2328, opened 2025-08-22
Factor out the parser into a separate crate

### LINKED ISSUE #2265: Move the parser into its own crate
Crates are the unit of compilation for Rust. They are also the unit of a re-usable piece of code for `cargo`. Since the parser and the grammar don't change that often, and changes there are costly (in build time at least), it could make sense to move the parser into its own crate.

Now that we have a proper parse-time AST, we can move it there as well, and have a self-container `nickel-lang-parser` crate.

Another benefit would be that third parties could reuse it to implement tooling or to integrate Nickel with other tools.

Have a separate `nickel-lang-parser` crate that contains the grammar, the lexer, the parser, and the type definitions for the AST

jneem:
Factors out the parser into a separate crate, `nickel-lang-parser`. This is mostly just moving things around, but there were a few other changes.

- The implementations of `CloneTo` for some type aliases in `nickel_lang_core::typecheck` (like `UnifEnumRow`) broke the orphan rule. I worked around these by adding some more generic implementations of `CloneTo` in the parser, so that those type aliases already implement `CloneTo` from the implementations in `nickel_lang_parser`
- Some of the `Display` impls in `nickel_lang_core::typ` (like `EnumRow`) broke the orphan rule. I couldn't re-use the same trick from the last point, because of differences in the two different `Allocator` structs. So I used a newtype wrapper instead.
- I ported the parser tests to the new ast.

I didn't pay too much attention to exactly where the API boundary is. The main goal was to get it working, and then we can refine it later.

Fixes #2265


parser/src/environment.rs
```diff
@@ -0,0 +1,322 @@
+//! An environment for storing variables with scopes.
```
yannham, line 1: Why do we need to have environment here? It's a bit strange. Maybe another  solution would be, middle term, to separate it entirely into its own helper crate.

yannham:
> Some of the Display impls in nickel_lang_core::typ (like EnumRow) broke the orphan rule. I couldn't re-use the same trick from the last point, because of differences in the two different Allocator structs. So I used a newtype wrapper instead.

Where is this code/newtype located? I'm trying to focus the review a bit.

core/src/typ.rs
```diff
@@ -313,12 +41,14 @@ pub enum TypeF<Ty, RRows, ERows, Te> {
 // be boxed. Hence, we don't need to additionally box `RecordRow`.
 
 /// Concrete, recursive definition for an enum row.
-pub type EnumRow = EnumRowF<Box<Type>>;
+#[derive(Clone, PartialEq, Debug)]
+pub struct EnumRow(pub EnumRowF<Box<Type>>);
```
jneem, line 82: Here (and RecordRow just below) is the newtype wrapper to allow us to implement Display.

parser/src/ast/alloc.rs
```diff
@@ -645,63 +673,75 @@ impl CloneTo for Type<'_> {
     }
 }
 
-impl CloneTo for typ::EnumRows<'_> {
-    type Data<'ast> = typ::EnumRows<'ast>;
+impl<Ty: CloneTo, ERows: CloneTo> CloneTo for typ::EnumRowsF<Ty, ERows> {
+    type Data<'ast> = typ::EnumRowsF<Ty::Data<'ast>, ERows::Data<'ast>>;
```
jneem, line 677: Here's the more generic impl of `CloneTo` for `EnumRowsF` so that it covers the type alias in `nickel_lang_core`

core/src/typ.rs
yannham: Ah, I feared this would be quite painful, but it doesn't impact the typechecker, only the runtime `Type` representation. Maybe we should add a comment as to why this one only get the newtype treatment. Another possibility would be to use only the newtype wrapper when we actually need to pretty-print/display, and keep the naked type alias in the actual definition. Though if  the newtype wrapper didn't cause much diff,maybe that's not very useful...

parser/src/ast/alloc.rs
yannham: Thanks!

parser/src/environment.rs
jneem: It's used in `fix_type_vars`  in the `uniterm` module.

jneem:
@yannham this is now rebased on a recent master. I reduced the renaming pain by adding a bunch of `pub use`s to preserve the old paths. These are marked with `TODO(parser migration)` and I'll follow up with a PR that does nothing but rework the import paths.

core/src/parser.rs
```diff
@@ -0,0 +1,145 @@
+use crate::{
```
yannham, line 1: Nitpick: maybe we could add a comment to explain why we still need a `parser` module although there is a parser crate (compatibility with/parsing directly into the runtime repr). In fact I wonder if that shouldn't move to `compat`, or to a submodule of `compat`. But let's keep that question for another PR.

parser/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    let out_parser_dir = out_dir.join("parser");
+    std::fs::create_dir_all(&out_parser_dir).expect("failed to create $OUT_DIR/parser");
+    // Running lalrpop can be expensive. When building from git, we generate the parser
+    // in this build script, but when publishing the crate we add the generated
+    // parser to the published crate at `src/parser/grammar.rs`.
+    //
+    // In order to have this build script work for both the published crate and the git
+    // version, we try to copy `src/parser/grammar.rs` into the same location in $OUT_DIR
+    // that lalrpop would generate the grammar. If that copy fails because `src/parser/grammar.rs`
+    // doesn't exist, we're probably building from git and so we generate the grammar.
```
yannham, line 17: By the way, is all this gymnastic still required, now that `nickel-lang-parser` is in its own crate? I don't remember exactly what were the cases we wanted to cover with the pre-generated grammar, but I expect that `nickel-lang-parser` wouldn't change much and thus the grammar won't be recompiled often. On the other hand, there might still be cases where pre-generating the grammar can avoid a run of LALRPOP from scratch, so I don't know...let's keep it in this PR anyway

yannham (approved):

yannham:
I also forgot to add: I wonder of some of the shim shouldn't stay. Might be useful to re-export parsing errors `nickel_lang_core::error`, for example. I've seen a few other examples where that could be the case.

parser/build.rs
jneem: This was originally prompted by #2123, which I think still applies: users depending on `nickel-lang-core` from `crates.io` shouldn't have to generate the grammar. But I do need to update the paths in that comment...

PR #2331, opened 2025-08-29
Use rust-toolchain.toml to synchronize the rust version

jneem:
This removes the beta devShell. Was anyone using it?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -101,8 +101,6 @@
         in
         { rustProfile ? "minimal"
         , rustExtensions ? [
-            "rust-src"
-            "rust-analyzer"
             "rustfmt"
             "clippy"
           ]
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -109,25 +102,17 @@
         { rustProfile ? "minimal"
         , rustExtensions ? [
             "rust-src"
-            "rust-analysis"
+            "rust-analyzer"
```
yannham: I wonder if rust-analyzer should be here, and propagated to all builds. Shouldn't we only include it in the mkShell? Or is it better for caching purposes to just have one global setting and use the same setup everywhere?

yannham (approved):

flake.nix
jneem: Good point, I moved it (and rust-src) to the dev shell. It shouldn't affect caching, as the dev shell already has a slightly different toolchain (e.g. with a wasm32 target).

PR #2332, opened 2025-08-30
Convert to saphyr and make tests pass

jneem:


core/src/error/mod.rs
```diff
@@ -1495,6 +1503,20 @@ impl IntoDiagnostics for EvalError {
                     .with_message(format!("{format} parse error: {msg}"))
                     .with_labels(labels)]
             }
+            EvalError::DeserializationErrorWithInner { format, inner, pos } => {
+                let mut diags = inner.into_diagnostics(files);
+                if let Some(diag) = diags.first_mut() {
+                    if let Some(span) = pos.as_opt_ref() {
+                        *diag = diag
```
yannham: I believe the  `labels` field is public, so you should be able to just push to it mutably (or overwrite), without needing to clone it.

core/src/error/mod.rs
```diff
@@ -1495,6 +1503,20 @@ impl IntoDiagnostics for EvalError {
...
+                            .clone()
+                            .with_label(secondary(span).with_message("deserialized here"));
+                    }
+                    *diag = diag
+                        .clone()
```
yannham: ditto avoidable clone()

core/src/eval/operation.rs
```diff
@@ -2739,12 +2740,22 @@ impl<R: ImportResolver, C: Cache> VirtualMachine<R, C> {
...
-                            "Yaml" => serde_yaml::from_str(s).map_err(|err| {
-                                EvalError::DeserializationError(
-                                    String::from("yaml"),
-                                    format!("{err}"),
-                                    pos_op,
-                                )
+                            // TODO: we could try to generate better error positions here,
+                            // but it will be some work.
+                            //
+                            // We pass `None` to `load_yaml` even if we have a position for `s`,
```
yannham: Nitpick: it's hard to know what the second argument of `load_yaml` is, without going to the definition. Maybe recalling its nature here would help understand the tradeoff that is mode by passing `None`?

core/src/yaml.rs
```diff
@@ -0,0 +1,144 @@
+use codespan::ByteIndex;
```
yannham: I wonder if  this shouldn't be a submodule of serialize or something... It's peculiar that we have a free-standing `yaml` module at the root of the hierarchy. Nitpick: add top-level documentation for  the module.

core/src/yaml.rs
```diff
@@ -0,0 +1,144 @@
...
+                src_id,
+                start: ByteIndex::from(start as u32),
+                end: ByteIndex::from(end as u32),
+            })
+        })
+        .unwrap_or_default()
+}
+
+/// Convert saphyr's yaml representation to a `RichTerm`.
+fn yaml_to_rich_term(
```
yannham: I'm slightly concerned with the recursive nature of this function, even if we probably do that for `toml` already. It could be interesting to try to have even a rough idea of the threshold size from which the stack starts to blow up.

Relatedly, I wonder what is the performance impact of this change. We didn't gate the "spanned TOML deserialization" with a feature, neither did we properly check the difference, but I'm still curious of the order of magnitude. I reckon conversion to `RichTerm` and allocation will probably be demanding as well, so I'm not sure how noticeable it is.

jneem:
I haven't measured the performance yet, but I did convert to the lower-level methods in saphyr-parser. The code is a little longer, but I think it isn't too bad. It's also non-recursive, and has better support for YAML aliases (which are now allowed to be recursive, and generate nickel `Term::Var`s instead of effectively copy-pasting chunks of input).

core/src/serialize/yaml.rs
```diff
@@ -0,0 +1,437 @@
...
+
+/// Keep track of anchors.
+///
+/// We turn anchor references into Nickel variable references: every node
+/// with an anchor gets inserted into a big top-level `let` block, and every
+/// reference to an anchor gets turned into a `Term::Var` node.
+///
+/// This is different from saphyr's approach, where they clone the referent
+/// every time it's referenced. In addition to avoiding unnecessary clones,
+/// our version supports recursive anchor references.
```
yannham, line 103: Nitpick: a little example of the translation of (recursive) anchors could help the reader here.

core/src/serialize/yaml.rs
```diff
@@ -0,0 +1,437 @@
...
+            return;
+        }
+
+        use saphyr_parser::Event::*;
+        let pos = mk_pos(self.file_id, span.start.index(), span.end.index());
+        match ev {
+            DocumentStart(_) | Nothing | StreamStart | StreamEnd => {}
+            DocumentEnd => {
+                let doc = match self.doc_stack.len() {
+                    0 => RichTerm::new(Term::Null, pos),
```
yannham, line 319: Was that the previous behavior, to return `null` upon empty document? I'm asking out of curiosity, as I don't really see what else can we do...

core/src/serialize/yaml.rs
```diff
@@ -0,0 +1,437 @@
...
+                let node = Node::Map {
+                    map: IndexMap::default(),
+                    cur_key: None,
+                    pos,
+                };
+                self.doc_stack.push((node, anchor_id));
+            }
+            SequenceEnd | MappingEnd => {
+                // TODO: we could handle recursive anchor ids (saphyr doesn't) by building
+                // a recursive let block...
```
yannham: Is this a left-over? It seems to contradict the documentation of `AnchorMap`, which leads to think that recursive anchors are in fact supported.

core/src/serialize/yaml.rs
```diff
@@ -0,0 +1,437 @@
...
+                .or_insert((Ident::fresh(), Term::Null.into()));
+            *slot = rt;
+            Term::Var((*ident).into()).into()
+        } else {
+            rt
+        }
+    }
+}
+
+#[derive(Default)]
```
yannham, line 168: Nitpick: the fields of this structure could be documented (the two stacks, and how we use them, succinctly).

core/src/serialize/yaml.rs
jneem: Yes, that was the behavior of `serde_yaml`

core/src/serialize/yaml.rs
jneem: Oops, that was indeed a left-over

PR #2334, opened 2025-09-04
Initial nickel-lang API proposal

jneem:
This is a proposal for the initial public API of the `nickel-lang` crate.

It's a little fancier than py-nickel in that it allows for exploration of in-memory structures (not just writing to strings).

I'll write some documentation, I promise. I just wanted to check if the general shape of this looks acceptable first.

### DIFF
--- a/nickel/Cargo.toml
+++ b/nickel/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "nickel-lang"
+description = "Stable library interface to the Nickel language"
+version = "0.1.0"
+authors.workspace = true
+license.workspace = true
+edition.workspace = true
+keywords.workspace = true
+repository.workspace = true
+homepage.workspace = true
+readme.workspace = true
+
+[dependencies]
+codespan-reporting.workspace = true
+indexmap.workspace = true
+malachite.workspace = true
+nickel-lang-core.workspace = true
+nickel-lang-vector.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+serde_yaml.workspace = true
+toml.workspace = true
--- a/nickel/src/lib.rs
+++ b/nickel/src/lib.rs
@@ -1,0 +1,727 @@
+//! A stable library interface to the Nickel language.
+//!
+//! Most of the Nickel language is implemented in `nickel-lang-core`, but
+//! `nickel-lang-core` exposes implementation details that frequently change.
+//! This library implements a more basic interface, with fewer features. Its
+//! goals are:
+//!
+//! - Simplicity: it should be easy to do basic things, like evaluating
+//!   some Nickel code into a structured representation.
+//! - Stability: we aim not to break public APIs. This doesn't mean
+//!   we'll *never* break them, just that stability is a priority.
+//! - Embeddability: our interfaces are designed to be usable from
+//!   other languages.
+
+use std::{
+    cell::RefCell,
+    ffi::OsString,
+    io::{Cursor, Write},
+    rc::Rc,
+};
+
+use codespan_reporting::term::termcolor::{Ansi, NoColor, WriteColor};
+use malachite::base::{num::conversion::traits::RoundingFrom, rounding_modes::RoundingMode};
+
+use nickel_lang_core::{
+    error::{report::DiagnosticsWrapper, IntoDiagnostics, NullReporter},
+    eval::{cache::CacheImpl, Closure, Environment},
+    files::Files,
+    identifier::{Ident, LocIdent},
+    program::Program,
+    serialize::{to_string, validate, ExportFormat},
+    term::{
+        self,
+        record::{Field, RecordData},
+        RichTerm, Term,
+    },
+};
+
+pub use nickel_lang_core::error::report::ErrorFormat;
+
+/// Provides a destination for the output of `std.trace`.
+#[derive(Clone)]
+pub struct Trace {
+    // This is a little annoying, in that VirtualMachine already takes a generic
+    // trace and boxes it. Since we don't want a generic parameter on Context,
+    // it means we need to double-box it.
+    inner: Rc<RefCell<dyn Write>>,
+}
+
+impl Trace {
+    pub fn new<W: Write + 'static>(write: W) -> Self {
+        Trace {
+            inner: Rc::new(RefCell::new(write)),
+        }
+    }
+}
+
+impl Write for Trace {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.inner.borrow_mut().write(buf)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.inner.borrow_mut().flush()
+    }
+}
+
+impl Default for Trace {
+    fn default() -> Trace {
+        Trace::new(std::io::sink())
+    }
+}
+
+/// The main entry point.
+#[derive(Clone, Default)]
+pub struct Context {
+    import_paths: Vec<OsString>,
+    trace: Trace,
+    name: Option<String>,
+}
+
+/// Controls the usage of ANSI escape codes in error messages.
+pub enum Color {
+    Ansi,
+    Off,
+}
+
+/// A Nickel evaluation error.
+pub struct Error {
+    error: Box<nickel_lang_core::error::Error>,
+    files: nickel_lang_core::files::Files,
+}
+
+impl std::fmt::Debug for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.error.fmt(f)
+    }
+}
+
+impl Error {
+    pub fn format<W: Write>(
+        mut self,
+        write: &mut W,
+        format: ErrorFormat,
+        color: Color,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let diagnostics = DiagnosticsWrapper::from(self.error.into_diagnostics(&mut self.files));
+        match format {
+            ErrorFormat::Text => {
+                let mut ansi;
+                let mut no_color;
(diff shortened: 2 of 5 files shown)


jneem:
Sure, I can write a quick RFC for this.

I'd also add

- control evaluation eagerness/laziness, e.g. by evaluating to WHNF and then controlling which record fields to evaluate further

to the list. And it's probably also worth looking at the Nix C API for more inspiration.

yannham:
> Sure, I can write a quick RFC for this.

Sounds great; it can be very short, but it's always nice to write things down and have them on record.

jneem:
Not everything's in there, but I think it's in a reasonable state for a look.

I added metadata and shallow evaluation.

I haven't yet added
- C API
- rust iterators
- deserialization to rust types

nickel/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "nickel-lang"
+description = "Stable library interface to the Nickel language"
+version.workspace = true
```
yannham: Versioning this is going to be tricky. On one hand, as opposed to the other crates using the `version.workspace`, this isn't a binary, but a library. And stability is one of the main goal, I think we can't use the same version as the workspace: if we want to make a breaking change to the API, we can't bump the whole Nickel language to `2.0`, so we'd be blocked (and conversely, the language could do a `2.0` upgrade while this API stays backward-compatible). We'll probably have to version this library independently from the rest.

On the other hand, beside the version "as a Rust library", there is the question of which version of the Nickel language is supported. We could imagine that the same version of the Rust API could in fact support different Nickel version/backend. Though I'm not sure how useful it is; my first intuition would be to version `nickel-lang` independently, but such that each version of `nickel-lang` supports a fixed version of the Nickel language. In practice `nickel-lang` should upgrade strictly slower than the workspace, that is `upgrade minor(nickel-lang) => upgrade minor or major (workspace)` and `upgrade major(nickel-lang) => upgrade major(workspace)`.

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+                    }
+                    Color::Off => {
+                        no_color = NoColor::new(write);
+                        &mut no_color
+                    }
+                };
+                diagnostics.diagnostics.iter().try_for_each(|d| {
+                    Ok(codespan_reporting::term::emit(
+                        writer,
+                        &codespan_reporting::term::Config::default(),
```
yannham: this is probably optimized away, but maybe we could pull the creation of the config out of the loop

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+    /// This is used to format error messages. If you read the main input
+    /// program from a file, its path is a good choice.
+    pub fn with_source_name(self, name: String) -> Self {
+        Context {
+            name: Some(name),
+            ..self
+        }
+    }
+
+    /// Evaluate a Nickel program deeply, returning the resulting expression.
```
yannham, line 194: As a public API, it might be helpful to detail just quickly what does "deep evaluation" mean. Stating that by default we don't recurse into containers (lazy evaluation), and give an example with a record and a failing contract that is not triggered by lazy evaluation, but is by deep evaluation, or something like that?

yannham: Ok, I saw `eval_shallow` afterwards. So there are enough details there, but I still think we should define idiosyncratic jargon like "deep". If only that as opposed to `eval_shallow`, this recurses into records and arrays.

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+/// but might have unevaluated thunks (if you got it from [`Context::eval_shallow`]).
+#[derive(Clone)]
+pub struct Expr {
+    rt: RichTerm,
+}
+
+/// A Nickel record.
+///
+/// Fields can be accessed either by name or by index.
+pub struct Record<'a> {
```
yannham, line 311: What are those lifetimes `'a`? Since the internal fields are not visible in the documentation, we should probably explain roughly what they represent.

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+            _ => None,
+        }
+    }
+
+    /// Is this expression an array?
+    pub fn is_array(&self) -> bool {
+        self.as_array().is_some()
+    }
+
+    pub fn is_thunk(&self) -> bool {
```
yannham: I think the naming here is confusing. Internally we use "thunk" to mean something else; for example a bare "1 + 1" is not a thunk at runtime but is indeed not an effective weak head normal form.

Maybe exposing `is_whnf` could be closer to the Nickel model, with the same definition (still using the `is_eff_whnf`). In fact the latter (`is_eff_whnf`) is a bit of an artifact of the current representation, but it will be gone with #2302, and there will only be one notion of WHNF, so we might get away with not exposing the `eff_whnf/whnf` subtlety to the user at all.

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+        !self.rt.as_ref().is_eff_whnf()
+    }
+}
+
+impl Record<'_> {
```
yannham, line 515: Should we provide iterator for `Record`?

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+        self.data.fields.get_index(idx).map(|(key, fld)| {
+            (
+                key.label(),
+                fld.value.as_ref().map(|rt| Expr { rt: rt.clone() }),
+            )
+        })
+    }
+}
+
+impl Array<'_> {
```
yannham, line 579: Ditto about the iterator.

nickel/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+
+pub use nickel_lang_core::error::report::ErrorFormat;
+
+/// Provides a destination for the output of `std.trace`.
+#[derive(Clone)]
+pub struct Trace {
+    // This is a little annoying, in that VirtualMachine already takes a generic
+    // trace and boxes it. Since we don't want a generic parameter on Context,
+    // it means we need to double-box it.
+    inner: Rc<RefCell<dyn Write>>,
```
yannham, line 47: I suppose one solution would be to pass a `Box<dyn Write>` to the VM, instead of having it do the boxing itself. However here you use `Rc` instead. Is there a reason why we would want the context to be cloneable (other than "it might be a good idea, UX-wise, in general")? Another solution (in addition to the change to `VirtualMachine` and `Program` to NOT do the boxing themselves) would be to separate the trace from the rest of the context, even if it makes the whole idea of having only one big context around a tad less appealing.

yannham (commented):
It's a good start and step ahead! I can already think of many extensions, but we'll have to think about the most useful steps. At some point I think it'd very useful to pull `nickel_lang_core::deserialize` in one way or another here (maybe even moving it entirely here, instead of providing just a facade). But first things first...

nickel/src/lib.rs
jneem: I added a one-sentence explanation to each one. Is it enough?

nickel/src/lib.rs
jneem: It had to be clone because `Program::new_from_source` takes the tracer by value. So either we need a cloneable `Trace` or else `Context::eval_deep_for_export` needs to take `self` by value.

On the one hand, it sort of makes sense to take `Trace` by value, because if you accidentally reuse it you might be surprised by the interleaved trace messages. On the other hand, I suspect most users don't care too much about handling tracing, so then it's awkward that you can't just make a single context and evaluate multiple nickel expressions...

nickel/Cargo.toml
jneem: Makes sense. Shall we start at 0.1.0 then? I think we should try to get to 1.0 fairly quickly, like maybe once all the obvious omissions are addressed?

nickel/src/lib.rs
jneem: I was trying to avoid the term "whnf" because I thought it might be scary. Fair point about "thunk" having a different meaning internally, though. What about "lazy"?

yannham: Another possibility, although it could be confusing as well (it depends on the naming elsewhere in this API in particular, it can be ok as long as we don't introduce `NickelValue` for ex), would be to use `is_value`.

After all a weak head normal form is kind of an extension of the usual meaning of _value_ in non-lazy languages, where a value is something inert, a terminal computation that doesn't evaluate further. Although people don't use the term "value" for WHNF in the literature, it should be fine for the surface API here (especially if we want to avoid literature jargon).

nickel/src/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        })?;
+
+        Ok(Expr { rt })
+    }
+}
+
+/// A Nickel expression.
+///
+/// This might be fully evaluated (for example, if you got it from [`Context::eval_deep`])
+/// but might have unevaluated thunks (if you got it from [`Context::eval_shallow`]).
```
yannham:
```suggestion
/// or might have unevaluated thunks (if you got it from [`Context::eval_shallow`]).
```

nickel/src/lib.rs
yannham: Yup, looks good

nickel/src/lib.rs
yannham: Ok, well, I suppose we can start like this. Double boxing isn't great but in a normal setting the trace shouldn't really be overflowed to a point where it really matters.

nickel/Cargo.toml
yannham: Sounds about right. We will need to publish at least at `1.0.2` because `1.0.1` exists and is yanked: https://crates.io/crates/nickel-lang

Issue #2333, opened 2025-09-04
How should I handle recursive records when driving the Rust API?


twitchyliquid64:
Consider this nickel:

```ncl
let rec
  b1 = {
    name = "build 1",
    inputs = [b2],
  },
  b2 = {
    name = "build 2",
    inputs = [b1],
  },

in
b1
```

Note that `b1` is listed as `b2`'s input, and vice-versa, making this recursive.

I get into an infinite loop with growing memory as soon as I build a `Program` out of this and call `eval_full_for_export()`. This makes sense, given that method is for exporting to JSON or whatever.

My question is, how can I eval the `RichTerm` tree in a manner that doesn't explode when there are recursive definitions, and handle circular references like this myself?

Thanks,
Tom

yannham:
Hello!

This is indeed something we have to account for in several places in the Nickel pipeline. It's fair that an export loops, as you mention. However, `nickel doc` must be able to handle those kind of cyclic structures (it happens in the wild that there are co-recursive default values, such as `{foo | default = derive_from_bar bar, bar | default = derive_from_foo foo}` so that you only have to set one). The second case that comes to mind is the background evaluator of the LSP (there's background evaluation to report contract errors live).

The general idea is to rely on the _lazy evaluation strategy_ of Nickel. The default evaluation strategy for Nickel, if you're using `eval_closure` for example, is to stop as soon as it sees a data structure - here, the record `b1` - without recursing inside. The fields are left unevaluated. `eval_full_for_export()` forces the result, but if you used `b1.name` instead as the last expression, that would work because `inputs` is never used and isn't part of the final result (including with `eval_full_for_export`).

So I would say it depends on what you want to do with the data, and how deep you want to explore the result.

## The `nickel doc` way

As an example, `nickel doc` uses the lazy evaluation strategy to recurse step by step in a value in the following way:

- lazily evaluate the top-level expression to a literal, a data structure or a function
- if it's a record, recurse into each field:
  - evaluate (lazily) the content of the field. If there is an error, abort. If it's a record, rinse, repeat.
  - there's an additional locking mechanism to avoid infinite recursion, so that if we go through a field while we were evaluating said field to begin with, leave it unevaluated and proceed with its next sibling
  
The result is what we call the _record spine_. It's implemented by `program::Program::eval_record_spine`, with the real logic in the freestanding function `maybe_closurized_eval_record_spine`:

[...]

PR #2337, opened 2025-09-05
Initial public api rfc

jneem:


rfcs/008-public-api.md
```diff
@@ -0,0 +1,90 @@
...
+Currently, the only stable interface to Nickel is the CLI tool; anyone
+wanting to embed the Nickel interpreter needs to deal with the
+`nickel_lang_core` crate, which changes frequently and isn't
+particularly user-friendly.
+
+This RFC proposes a simple Nickel API. Its main goals are
+
+- Simplicity: it should be easy to do basic things, like evaluating
+  some Nickel code into a structured representation.
+- Stability: we aim not to break public APIs.
```
yannham: I wonder if we shouldn't tame this down a bit. Do we really want to promise from now that the API will stay stable forever? It does sound a bit strong to me. However, we could say that the goal is to make breaking changes as rare as possible?

rfcs/008-public-api.md
```diff
@@ -0,0 +1,90 @@
...
+enums, records, and arrays. We also want to allow for some control
+over evaluation, so we should allow some kind of thunk.
+
+We will define an opaque `Expr` struct representing one of the value
+types above, with methods for querying the kind and extracting values.
+Our initial version will omit thunks.
+
+Access to arrays will be done by querying the length and indexing.
+(The Rust interface could also have an iterator, but the C FFI will
+need the simpler interface.) Access to records can be done by name,
```
yannham, line 38:
```suggestion
need the simpler interface). Access to records can be done by name,
```

rfcs/008-public-api.md
```diff
@@ -0,0 +1,90 @@
...
+or by querying the length and indexing. (Again, the Rust interface
+could also have an iterator.) Records can also support access to
```
yannham:
```suggestion
or by querying the length and indexing (again, the Rust interface
could also have an iterator). Records can also support access to
```

rfcs/008-public-api.md
```diff
@@ -0,0 +1,90 @@
...
+In principle, querying structured data from expressions is all the
+user needs. In practice, Nickel already has some more convenient
+ways to extract data and so we should give access to them.
+
+- We should allow conversion from `Expr` to JSON, YAML, and TOML.
+  These will return a single, large string. In the future, there
+  could also be a lazier version with a `std::io::Write`-like
+  interface.
+- We should allow conversion from `Expr` to serde-compatible
+  rust types (and maybe facet-compatible?).
```
yannham, line 69: Should we also allow to convert from `Expr` to a structured C value? It might be cheaper for an external consumer to convert the `Expr` all at once to a C datastructure isomorphic to `JsonValue` (even maybe directly language-specific representation, e.g. Java or Go), and move it to the external world once and for all.

rfcs/008-public-api.md
```diff
@@ -0,0 +1,90 @@
...
+this part should be as configurable as the Nickel CLI. For example,
+it should allow for configuration of input paths and variable overrides.
+
+## Evaluation
+
+We should have a function for full evaluation, where the resulting
+expression will have no thunks. We should have a function (but maybe
+not in the initial version) for evaluation to WHNF. We should have
+a function for typechecking input without evaluating it. *Maybe* we should
+have functions for more specialized kinds of evaluation.
```
yannham: I agree with the last setence, in particular in light of e.g. https://github.com/tweag/nickel/issues/2333, where someone would be keen on using `eval_record_spine` or similar laxer and/or error-tolerant deep evaluation strategies.

rfcs/008-public-api.md
jneem: I'd lean against this, because I think for most use-cases where you want a structured representation, you want a *custom* structured representation. In rust, serde derives generate the glue for this. But I don't know of a C equivalent, so our only option would be to generate our own C representation, which would end up a bit like `serde_json::Value`: an intermediate representation between the original and the custom one that the user actually wants.

For languages with reflection, it would make sense to have a nice way to export from Nickel to custom types, but I think the mechanism would need to be language-specific and live on the far side of the `nickel <-> C <-> other lang` boundary.

rfcs/008-public-api.md
jneem: Right, so I think evaluating to WHNF *allows* for anyone to implement their own strategy, and the main question in my mind is whether we want to provide more prebuilt strategies.

yannham: Well, I suppose as long as the strategy is reasonably defined, and thus it is expected that it will be needed in the future and is not some kind of temporary thing (like the record spine), I guess why not. But I suppose this isn't too important for now. We should clearly keep that for future evolution. Parsimony shall be the rule, if we want to live up the stability expectation.

rfcs/008-public-api.md
yannham: > I'd lean against this, because I think for most use-cases where you want a structured representation, you want a custom structured representation

Well, maybe not for a pure C/C++ usage. I'm just thinking that back-and-forth between Rust and C callbacks might be more complex (and potentially costly) than moving everything to the C (or any other external world) at once, in a C-like `serde_json::Value`. Even more if this is a back and forth like Rust <-> C <-> High level, GCed language.

That being said, I agree that in the case of a "high level, GCed language" L, most of the time this needs to be transformed again to a L-specific representation.

At any rate, it's fair to leave this out of the RFC for now.

rfcs/008-public-api.md
jneem: I think this one is ok, at least in American English (https://english.stackexchange.com/a/11129)

rfcs/008-public-api.md
jneem: I reworded this a little.

rfcs/008-public-api.md
yannham: Ah, TIL.

PR #2338, opened 2025-09-07
Pass sources from World into background eval

### LINKED ISSUE #2322: LSP cache for background jobs gets out of sync with main cache when files are imported from the filesystem
**Describe the bug**
The SupervisorState used by background jobs keeps its own cache of file contents and dependencies, which must be kept in sync with World. In the case that a file is added to World by an import from the filesystem, that change is not being propagated to the background eval cache. The most noticeable effect of this it can cause the background eval job to reload files from the filesystem instead of using the in memory version of the file, giving incorrect diagnostics.

**To Reproduce**
Create the following files:
```
# a.ncl
import "b.ncl"
```
```
# b.ncl
import "c.ncl"
```
```
# c.ncl
1 : String
```
Open a.ncl and c.ncl in VSCode or probably any other editor (I've only tested with VSCode). Make sure b.ncl has never been opened during that editing session. Both a.ncl and c.ncl should have diagnostics due to the type error in c.ncl. Fixing the type error in c.ncl by changing it to ```1 : Number``` will fix the diagnostics in that file but a.ncl will still show an error.

**Expected behavior**
The diagnostic in a.ncl should be resolved based on the correction to c.ncl

**Environment**
 - Version of the code: Tested on the main branch as of commit 0809dbc26eefe7118e5e4bc85f681ddaef932c5e

**Additional context**
I was thinking of working on this myself but I wanted to check in on something first. What's the reason that that the background thread maintains its own version of the cache in the first place? If it used the same instance of World instead this sort of thing would not be a problem, but I assume there's some reason it's set up the way it is.

L0r3m1p5um:
fixes #2322.

This is the most straightforward way I could find to share the state of world with the background eval task and prevent synchronization issues. The downside is that the contents of the file to evaluate and its dependencies have to get copied into the background thread on every evaluation. I figured I'd check if this was ok despite that since the alternatives I can think of would be significantly more complicated. I did try just using a mutex to share SourceCache but it's not thread safe due to Rc being used pretty frequently throughout core.

### DIFF
--- a/lsp/nls/src/background.rs
+++ b/lsp/nls/src/background.rs
@@ -16,11 +16,6 @@ use crate::{config, diagnostic::SerializableDiagnostic, files::uri_to_path, worl
 /// Environment variable used to pass the recursion limit value to the child worker
 const RECURSION_LIMIT_ENV_VAR_NAME: &str = "NICKEL_NLS_RECURSION_LIMIT";
 
-#[derive(Debug, Serialize, Deserialize)]
-enum Command {
-    EvalFile(Eval),
-}
-
 /// The evaluation data that gets sent to the background worker.
 #[derive(Debug, Serialize, Deserialize)]
 struct Eval {
@@ -39,7 +34,7 @@ pub struct Diagnostics {
 
 pub struct BackgroundJobs {
     receiver: Option<Receiver<Diagnostics>>,
-    sender: Sender<Command>,
+    sender: Sender<Eval>,
 }
 
 fn run_with_timeout<T: Send + 'static, F: FnOnce() -> T + Send + 'static>(
@@ -91,7 +86,7 @@ pub fn worker_main() -> anyhow::Result<()> {
 }
 
 struct SupervisorState {
-    cmd_rx: Receiver<Command>,
+    eval_rx: Receiver<Eval>,
     response_tx: Sender<Diagnostics>,
 
     // A stack of files we want to evaluate, which we do in LIFO order.
@@ -106,12 +101,12 @@ struct SupervisorState {
 
 impl SupervisorState {
     fn new(
-        cmd_rx: Receiver<Command>,
+        eval_rx: Receiver<Eval>,
         response_tx: Sender<Diagnostics>,
         config: config::LspEvalConfig,
     ) -> anyhow::Result<Self> {
         Ok(Self {
-            cmd_rx,
+            eval_rx,
             response_tx,
             banned_files: HashMap::new(),
             eval_stack: Vec::new(),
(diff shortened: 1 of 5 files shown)


jneem:
It isn't ideal but I think it's acceptable: correctness is more important than efficiency, and it shouldn't be a massive hit to efficiency anyway given that we have no incremental parsing/typechecking/evaluation yet. I'll take a closer look tomorrow.

lsp/nls/src/background.rs
```diff
@@ -21,21 +18,7 @@ const RECURSION_LIMIT_ENV_VAR_NAME: &str = "NICKEL_NLS_RECURSION_LIMIT";
...
-        uri: Url,
-        deps: Vec<Url>,
-    },
-    EvalFile {
-        uri: Url,
-    },
-    DeleteFile {
-        uri: Url,
-    },
+    EvalFile(Eval),
```
jneem: Maybe just get rid of the `Command` altogether, since there's only a single variant now?

L0r3m1p5um: Good point, that's done now.

yannham:
@jneem @L0r3m1p5um  I'm trying to get a picture of how much more we spend on sending files over the wire here. I suppose we already had to send data over in the previous version. In fact, if an evaluation is requested, there is a chance that the file has been edited since last time, so for the main file, unless we are in a case like "modify reverse dependencies", this shouldn't change much.

However, does this means that we have to reparse the dependencies of the file we want to handle on _each and every evaluation_? I can see that having noticeable performance impact, if you have a small file important large contracts (which is probably expected, typically e.g. when importing contracts generated from json-schema-to-nickel).

L0r3m1p5um:
The evaluation does have to reparse all of the dependencies, but this PR didn't change anything in that regard, that was already happening. Basically there's a main thread, a background thread that manages evaluation jobs, and child processes spawned by the background thread that perform the actual evaluation. This PR doesn't affect how the background thread interacts with the child process, either way only the file sources are passed in through stdin, and the child process has to reparse them anyway. What this changes is how the main thread interacts with the background job thread. Before, the background thread maintained its a separate cache of source files, and the main thread would have to propagate any changes to those files to the background thread. Now the separate background cache has been removed, and instead the necessary dependencies are copied from the main thread into the background job thread. They're copied because the main thread source cache is not thread safe so we can't share the sources directly.

yannham:
@L0r3m1p5um I see, thanks for the explanation, I didn't recall all the details. When you say that we can't share expressions, you're talking about sources, `RichTerm`, or both? For sources, we could easily turn the `Rc` into an `Arc` in `Files`. Using `Rc` in the first place is kinda of a work-around for error reporting, and I only expect very few copies to be alive during normal workload.

L0r3m1p5um:
I'm only talking about the sources. I do think that if Files were made thread safe, we'd be able to avoid the copying into the background thread. Files uses Vector from nickel-lang-vector though, which also uses Rc in its implementation. Vector is also used elsewhere so changing that to Arc would impact more than just Files.

yannham:
>  Files uses Vector from nickel-lang-vector though, which also uses Rc in its implementation. Vector is also used elsewhere so changing that to Arc would impact more than just Files.

I checked the definition of `File` with `Rc<str>`, and I didn't remember that we use `Vector` in `Files`. Honestly, `Files` is most of the time uniquely owned, and we made it copiable because of some error reporting/borrowing shenanigans, but in the happy path they shouldn't ever be copied (and even then, it's only a few times). A project with hundred of source files is already huge, which would only make `Files` a few kilobytes.

I wonder if it wouldn't just make more sense to use a proper `Vec` here, that we would clone if needed, since content sharing already happens at the level of individual `File`. Then we could slap an `Arc` onto `File`, avoid copying source contents in the background evaluator, and call it a day.

What do you think @jneem ? Am I missing some LSP corner cases that make any of those assumptions wrong?

jneem:
A `Vec<Arc<str>>` sounds fine to me.

Performance-wise, a background eval job should be roughly equivalent to shelling out to `nickel eval`, in that there's nothing incremental going on. There is some benefit to avoiding copying to the background thread: the actual eval jobs get throttled/debounced while the submissions to the background thread don't. So there could be many more background thread submissions than eval jobs.

PR #2339, opened 2025-09-07
Fix performance of UsageLookup

jneem:
When visiting a contract inside a type, we should `fill` it but not continue the traversal (because `fill` already recurses).


jneem:
(I guess we should also have continuous benchmarking for nls, but that isn't the kind of thing I do on weekends :smile:)

yannham (approved):
Nice diff :sunglasses: 

I'm going to force merge as the Linux CI is passing, the MacOS builder is just out of space and there's absolutely zero chance that this PR is platform-specific.

PR #2341, opened 2025-09-10
Clean up Ident conversions to reduce accidental allocations

jneem:
The original goal of this PR was to delete our custom interner, but that didn't work out.

I found three widely-used crates for string interning:
- `lasso`, which appears unmaintained and apparently doesn't build on recent versions of rust. There's a fork (lasso2), but it isn't widely used
- `string-interner`, which seems pretty similar to our existing approach. For our usage, the main downside is that it doesn't build in the notion of a process-wide interner. That means we need to put the `StringInterner` object in a global `Mutex`, and then we can't support our current usage of `Ident::label`, because the returned `&str` can only live as long as the temporary mutex lock. `string-interner` does have a variant with stable address, so in principle we could transmute the references. But then we haven't succeeded in delegating our unsafe code...
- `internment`, which builds in an easy-to-use global intern table. Unfortunately, it caused some substantial slowdowns in our benchmarks (about 10% across the board -- I ran it a few times to confirm). I'm not totally sure, but this might be because it increases `Ident`'s size from 4 to 8 bytes (because it basically becomes `&'static str`).

So after all that, our interning code is mostly untouched. I tweaked the conversions a bit to avoid allocations (for example, previously our conversion from `&str` to `Ident` was passing through a temporary `String`). Benchmarks improved by about 2-6%.

I also noticed that our generated identifiers aren't guaranteed unique, because there are ways besides `Ident::fresh` to create identifiers like `"%1"`. I left this as a FIXME, because I wasn't sure how to address it. One possibility is to reserve a bit in `Ident` to mark generated identifiers. This would guarantee that they always compare not-equal to non-generated identifiers, but it does cut the identifier space in half, supporting only 2B generated ids and 2B non-generated ids.

### DIFF
--- a/core/src/identifier.rs
+++ b/core/src/identifier.rs
@@ -17,7 +17,7 @@ static INTERNER: Lazy<interner::Interner> = Lazy::new(interner::Interner::new);
 // Implementation-wise, this is just a wrapper around interner::Symbol that uses a hard-coded,
 // static `Interner`.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
-#[serde(into = "String", from = "String")]
+#[serde(into = "&'static str", from = "String")]
 pub struct Ident(interner::Symbol);
 
 impl Ident {
@@ -26,7 +26,7 @@ impl Ident {
     }
 
     /// Return the string representation of this identifier.
-    pub fn label(&self) -> &str {
+    pub fn label(&self) -> &'static str {
         INTERNER.lookup(self.0)
     }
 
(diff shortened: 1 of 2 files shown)


yannham (approved):
One possibility would be to make `LocIdent` a `u64`, but this might make `LocIdent` bigger. Though with RFC007, we could maybe change the representation to use `PosIdx` instead, and get back to something similar overall, at least for the runtime repr (which is the only one where generated loc ident are used).

Another possibility would be to store this bit of information in the pointee, e.g. in a separate bitmap in the interner. Currently we need to retrieve the label to check if an identifier is generated anyway, so it probably won't change much, performance-wise.

By the way, I wonder if now that we pull ouroboros in for the new AST, we could use it to get rid of the unsafe part of the interner.

jneem:
Good points, I'll address them in a follow-up.

Just to confirm the semantics of generated ids, I think we want to guarantee that an ident generated with `Ident::new()` never collides with an `Ident::fresh()`, right? Even if they have the same string representation?

yannham:
> Just to confirm the semantics of generated ids, I think we want to guarantee that an ident generated with Ident::new() never collides with an Ident::fresh(), right? Even if they have the same string representation?

Right. So even beyond a bitmap, maybe what we want is to have two separate hashmaps mapping allocated strings to ids. That should simulate two separate spaces, but the index generation or hashing doesn't have to care about nor know about it: if we allocate a new "%1" that is not generated, it will create a new index, which will be hashed differently, and not be equal to, any potential previous generated "%1". If we don't use a bitmap, checking if an ident is generated is a bit more costly - we need to look into one of the hashmap at least - but it's still constant time. I don't know if that's something we do that often, but if we don't, that might be acceptable.

The advantage would be that we don't have to split the space in two, but basically the two disjoint spaces can grow dynamically in the same allocator.

PR #2346, opened 2025-09-12
Plumb custom_transform into EntryState to minimize traversal

### LINKED ISSUE #2344: `nickel_lang_core::traverse::Traverse<RichTerm>` gets exponentially slower
**Describe the bug**
As my codebase of Nickel gets larger and imports more things from all over the place, use of `nickel_lang_core::traverse::Traverse` gets exponentially slower to the point where it's taking minutes to run.

Image

**To Reproduce**
1. Have a dense web of imports
2. Call `traverse` on the RichTerm tree

The issue appears to be that calls to the callback pass the RichTerm by value and expect it returned by value, which leads to a ton of allocations and overhead even if you only want to mutate a small fraction of the tree.

https://github.com/tweag/nickel/blob/18ac25e66bd0a3a434934aaf15abfdd7c8656a83/core/src/term/mod.rs#L2233-L2251

**Suggestions**

1. Have a way to return `None` or something to signal no mutations, and have a code path which avoids `collect()` and all the heavy jazz
2. (Specific to my use-case) - All I'm trying to do is assign a unique ID to all `RecRecord` and `Record` terms - if we could just add a field to `RecordData` thats populated from an atomic, I could avoid traversal entirely.

(<2> I'm happy to implement if youre open to it, <1> i would need some guidance to implement)

twitchyliquid64:
Fixes #2344

This change addresses the underlying cause of performance issues in https://github.com/tweag/nickel/issues/2344, which was that a dense web of imports will result in lots of duplicated invocations when using `custom_transform`.

This change introduces custom transformations to the `EntryState` state machine, which allows iteration over terms to avoid terms which have already been applied to the custom transformation.

I can also confirm this addresses the issue upstream!

Thanks @jneem for coaching!!

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -184,7 +184,9 @@ impl TermCache {
 
                 Ok(CacheOp::Done(()))
             }
-            _ => Err(CacheError::NotParsed),
+            _ => Err(CacheError::IncompatibleState {
+                want: EntryState::Parsed,
+            }),
         }
     }
 
@@ -237,7 +239,9 @@ impl TermCache {
 
                 Ok(CacheOp::Done(()))
             }
-            _ => Err(CacheError::NotParsed),
+            _ => Err(CacheError::IncompatibleState {
+                want: EntryState::Parsed,
+            }),
         }
     }
 
@@ -1037,40 +1041,50 @@ impl CacheHub {
         Ok(())
     }
 
-    /// Applies a custom transform to an input and its imports, leaving them in the same state as
-    /// before. Requires that the input has been parsed. In order for the transform to apply to
-    /// imports, they need to have been resolved.
+    /// Applies a custom transform to an input and its imports. [CacheError::IncompatibleState] is returned
+    /// if the file has not yet been typechecked.
+    ///
+    /// If multiple invocations of `custom_transform` are needed, you must supply `transform_id` with
+    /// with a number higher than that of all previous invocations.
     pub fn custom_transform<E>(
         &mut self,
         file_id: FileId,
+        transform_id: usize,
         f: &mut impl FnMut(&mut CacheHub, RichTerm) -> Result<RichTerm, E>,
     ) -> Result<(), CacheError<E>> {
         match self.terms.entry_state(file_id) {
-            Some(state) if state >= EntryState::Parsed => {
-                if state < EntryState::Transforming {
+            None => Err(CacheError::IncompatibleState {
+                want: EntryState::Parsed,
+            }),
+            Some(state) if state < EntryState::Typechecked => Err(CacheError::IncompatibleState {
+                want: EntryState::Typechecked,
+            }),
+            Some(state) => {
+                if state.needs_custom_transform(transform_id) {
                     let cached_term = self.terms.terms.remove(&file_id).unwrap();
                     let term = f(self, cached_term.term)?;
                     self.terms.insert(
                         file_id,
                         TermEntry {
                             term,
-                            state: EntryState::Transforming,
+                            state: EntryState::CustomTransforming,
                             ..cached_term
                         },
                     );
 
                     if let Some(imports) = self.import_data.imports.get(&file_id).cloned() {
                         for file_id in imports.into_iter() {
-                            self.custom_transform(file_id, f)?;
+                            self.custom_transform(file_id, transform_id, f)?;
                         }
                     }
-                    // TODO: We're setting the state back to whatever it was.
                     // unwrap(): we inserted the term just above
-                    let _ = self.terms.update_state(file_id, state).unwrap();
+                    let _ = self
+                        .terms
+                        .update_state(file_id, EntryState::CustomTransformed { transform_id })
+                        .unwrap();
                 }
                 Ok(())
             }
-            _ => Err(CacheError::NotParsed),
         }
     }
 
@@ -1172,7 +1186,9 @@ impl CacheHub {
             }) => Ok(CacheOp::Done(Vec::new())),
             // >= EntryState::ImportsResolved
             Some(_) => Ok(CacheOp::Cached(Vec::new())),
-            None => Err(CacheError::NotParsed),
+            None => Err(CacheError::IncompatibleState {
+                want: EntryState::Parsed,
+            }),
         }
     }
 
@@ -1447,6 +1463,10 @@ pub enum EntryState {
     Typechecking,
     /// The entry and its transitive imports have been typechecked.
     Typechecked,
+    /// A custom transformation of the entry (through `Program::custom_transform`) is underway.
+    CustomTransforming,
+    /// This entry has completed custom transformations of this ID and lower.
+    CustomTransformed { transform_id: usize },
     /// The imports of the entry have been resolved, and the imports of its (transitive) imports are
     /// being resolved.
     ImportsResolving,
@@ -1458,6 +1478,19 @@ pub enum EntryState {
     Transformed,
     /// The entry has been closurized.
     Closurized,
+}
+
+impl EntryState {
+    fn needs_custom_transform(&self, transform_id: usize) -> bool {
+        if let EntryState::CustomTransformed {
+            transform_id: done_transform_id,
+        } = self
+        {
+            transform_id > *done_transform_id
+        } else {
+            self < &EntryState::CustomTransforming
+        }
+    }
 }
 
 /// The result of a cache operation, such as parsing, typechecking, etc. which can either have
@@ -1482,7 +1515,7 @@ impl<T> CacheOp<T> {
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub enum CacheError<E> {
     Error(E),
-    NotParsed,
+    IncompatibleState { want: EntryState },
 }
 
 impl<E> From<E> for CacheError<E> {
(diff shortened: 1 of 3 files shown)


core/src/program.rs
```diff
@@ -538,21 +538,22 @@ impl<EC: EvalCache> Program<EC> {
             .expect("File parsed and then immediately accessed doesn't exist"))
     }
 
-    /// Applies a custom transformation to the main term, assuming that it has been parsed but not
-    /// yet transformed.
+    /// Applies a custom transformation to the main term. Typechecking but not evaluation must have
+    /// been performed.
```
yannham: I think this change loses some information, which is that custom transformations happen before built-in program transformations:

> assuming that it has been parsed but not yet transformed.

core/src/cache.rs
```diff
@@ -1460,6 +1468,20 @@ pub enum EntryState {
     Closurized,
 }
 
+impl EntryState {
+    fn needs_custom_transform(&self, transform_id: usize) -> bool {
+        use EntryState::*;
```
yannham: Style nitpick: we tend to not import enum variants but use fully qualified names (although it has happened sometimes) unless there are really many variants, like in `traverse` or something. IMHO it helps readability.

core/src/cache.rs
```diff
@@ -1037,36 +1037,40 @@ impl CacheHub {
...
         &mut self,
         file_id: FileId,
+        transform_id: usize,
         f: &mut impl FnMut(&mut CacheHub, RichTerm) -> Result<RichTerm, E>,
     ) -> Result<(), CacheError<E>> {
         match self.terms.entry_state(file_id) {
-            Some(state) if state >= EntryState::Parsed => {
-                if state < EntryState::Transforming {
+            Some(state) if state >= EntryState::Typechecked => {
+                if state.needs_custom_transform(transform_id) {
```
jneem, line 1063: I think this can just be `state < EntryState::CustomTransformed { transform_id }`. That way it looks like all our other state comparisons, where we're using the order to figure out what's been done.

core/src/cache.rs
twitchyliquid64: Happy to! reverted to use the fully-qualified name.

core/src/program.rs
twitchyliquid64: Done! (reverted to previous description of state)

core/src/cache.rs
twitchyliquid64: Done!

twitchyliquid64: I'll note that `state < EntryState::CustomTransformed { transform_id }` didn't remove the need for `needs_custom_transform()`, afaict the derived `Ord/PartialOrd` implementations just order based on the enum variants alone, and do not try to create an ordering based on the field values in an enum (struct) variant for the case `cmp(Enum::A(...), Enum::A(..))`.

yannham: Ah, it's a gotcha I didn't know about, good catch @twitchyliquid64.

yannham (approved):

yannham:
It's missing a `cargo fmt`

core/src/cache.rs
jneem: It isn't critical for this PR, but I'm pretty sure the derived `PartialOrd` looks at the fields (or if not, I've written a lot of wrong code...). Here's a playground example:  https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5eeac881dd9232167b88de4fa38e757c

core/src/cache.rs
```diff
@@ -1037,36 +1037,40 @@ impl CacheHub {
...
     pub fn custom_transform<E>(
         &mut self,
         file_id: FileId,
+        transform_id: usize,
         f: &mut impl FnMut(&mut CacheHub, RichTerm) -> Result<RichTerm, E>,
     ) -> Result<(), CacheError<E>> {
         match self.terms.entry_state(file_id) {
-            Some(state) if state >= EntryState::Parsed => {
-                if state < EntryState::Transforming {
+            Some(state) if state <= EntryState::CustomTransformed{ transform_id: transform_id } => {
```
jneem: I think this needs to go back to the older version: we need to error if we haven't reached `EntryState::Parsed` yet.

twitchyliquid64: `EntryState::Parsed` is the first state so I thought thats always true - is it possible to have an entry in the terms cache that isnt parsed?

jneem: Oh, indeed, sorry. I wonder why we have so many other comparisons to `EntryState::Parsed` then. Maybe there used to be other states?

twitchyliquid64: That would be my guess! It works regardless as I imagine unparsed meant there was no entry in the term cache, and `_ => Err(CacheError::NotParsed),` would match a `None`.

twitchyliquid64:
I've run `cargo fmt`, so this should be good for a fresh CI run

jneem:
Ok, I just started CI. I'm still not following the match block, though, because if the state is `EntryState::Transformed`, for example, then won't the match fail and we'll report `NotParsed`?

Ah, CI is now complaining about clippy: https://github.com/tweag/nickel/actions/runs/17773742935/job/50521612752?pr=2346#step:5:986

twitchyliquid64:
> I'm still not following the match block, though, because if the state is EntryState::Transformed, for example, then won't the match fail and we'll report NotParsed?

I think so - this is why I originally had the language in the docs that `custom_transform` can be used on a Program which has been typechecked but not yet transformed. All uses of `custom_transform` I could find follow the pattern of calling it after typechecking.

jneem:
Right, but if you're in an `EntryState::Transformed` state then you've already been parsed, so it doesn't make sense to return a `NotParsed` error.

The pattern we follow elsewhere is that the state advances through a list of predefined stages, and if you call a function to process a stage that's already been processed (like, calling `typecheck` even though the state is past `EntryState::Typechecked`) then it's a no-op. This PR breaks that pattern, by erroring instead of a no-op if the state is already too advanced.

I'm not fanatically dedicated to keeping the existing pattern -- `custom_transform` is a bit special, and so I'm happy to discuss having special behavior -- but if it's going to be an error to call `custom_transform` when the state is too advanced, it shouldn't be a `NotParsed` error.

twitchyliquid64:
The former implementation, because it didnt actually participate in the state machine flow, just took a copy of the current state and restored it when it was done, allowing it to work both post-parse-pre-transform and post-transform.

Now that we are trying to make `custom_transform` fit the state machine, i think it needs to have a well-defined place. Based on my usage and the only other usage of this function I could find (doctests), I think custom transforms should run after typechecking.

Automagically progressing the state machine if possible sounds ideal. Agree that we need a better error than `NotParsed`.

jneem:
Automagically progressing the state machine would be nice, but I don't think we need it for this PR. My only blocking concern is that if we've progressed the state *past* the custom transform then we should no-op instead of erroring.

twitchyliquid64:
Understood - I've added a case to go lame-duck and return `Ok(())` if state has progressed passed the custom transform.

twitchyliquid64:
Should be good for another CI run / pass!

core/src/cache.rs
```diff
@@ -1037,36 +1037,48 @@ impl CacheHub {
...
+            Some(state)
+                if state
+                    > EntryState::CustomTransformed {
+                        transform_id: usize::MAX,
+                    } =>
+            {
+                Ok(())
+            }
+            Some(state) if state <= EntryState::CustomTransformed { transform_id } => {
+                if state.needs_custom_transform(transform_id) {
```
jneem: Ok, but now I get a `NotParsed` error if state is `EntryState::CustomTransformed { x }` for any `x > transform_id`.

For example, this test case errors out the second time I try to `custom_transform(4)` it:

```rust
        let mut p: Program<CacheImpl> = Program::new_from_source(
            std::io::Cursor::new("{}"),
            "<test>",
            std::io::sink(),
            NullReporter {},
        )
        .unwrap();
        p.typecheck(TypecheckMode::Walk).unwrap();
        p.custom_transform(4, |_ch, rt| Ok::<_, std::convert::Infallible>(rt))
            .unwrap();
        p.custom_transform(5, |_ch, rt| Ok::<_, std::convert::Infallible>(rt))
            .unwrap();
        // This one fails:
        p.custom_transform(4, |_ch, rt| Ok::<_, std::convert::Infallible>(rt))
            .unwrap();
```

I think this layout of the match block is simpler, and gets all the cases.

```rust
            None => Err(CacheError::NotParsed),
            Some(state) if state <= EntryState::Typechecked => Err(CacheError::NotParsed),
            Some(state) => {
                if state.needs_custom_transform(transform_id) {
                    // do the transform
                }
                Ok(())
            }
```

yannham: I don't think there's a precedent for returning `NotParsed` when a file is actually parsed (`state <= EntryState::Typechecked`. I find this behavior a bit misleading. We should either return a richer error, or I believe some other cache method just automatically apply the missing steps to reach the desired one.

twitchyliquid64: Automagically doing typechecking would be hard because custom transforms can return an error, which is wrapped as a `CacheError<E>`, whereas typechecking returns a different error `CacheError<..TypecheckError>`.

Is it okay to add `CacheError::NotTypechecked` ? or do we want to decompose into a more general `CacheError::IncompatibleState(state)` or some jazz now that we have two NotBlah cases?

yannham: I suppose having `IncompatibleState` would be sensible and lightweight, yes :+1:

twitchyliquid64:
Refactored `NotParsed` and `NotTypechecked` into a single `CacheError::InvalidState{ want: EntryState }` CI again plz :pleading_face: :pray:

yannham (approved):

yannham:
There's just a dangling reference to `NotTypechecked` in the rustdoc in a comment, otherwise LGTM :+1:

twitchyliquid64:
Fixed! :pray:

PR #2347, opened 2025-09-13
Avoid copying file contents in LSP eval

L0r3m1p5um:
Shares file contents with an Arc when sending it to the BackgroundJobs thread rather than copying the string. It's not exactly the approach we discussed but this has a more minimal impact to core.

### DIFF
--- a/lsp/nls/src/background.rs
+++ b/lsp/nls/src/background.rs
@@ -17,45 +17,13 @@ use crate::{config, diagnostic::SerializableDiagnostic, files::uri_to_path, worl
 /// Environment variable used to pass the recursion limit value to the child worker
 const RECURSION_LIMIT_ENV_VAR_NAME: &str = "NICKEL_NLS_RECURSION_LIMIT";
 
-/// The url and source contents to evaluate. This is what the background process will deserialize
-/// the request into so all data is owned.
-#[derive(Debug, Deserialize)]
+/// The evaluation data that gets sent to the background worker.
+#[derive(Debug, Serialize, Deserialize)]
 struct Eval {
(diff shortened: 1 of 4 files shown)


lsp/nls/src/background.rs
```diff
@@ -147,7 +179,11 @@ impl SupervisorState {
         let mut tx = tx.ok_or_else(|| anyhow!("failed to get worker stdin"))?;
         let mut rx = rx.ok_or_else(|| anyhow!("failed to get worker stdout"))?;
 
-        bincode::serde::encode_into_std_write(eval, &mut tx, bincode::config::standard())?;
+        bincode::serde::encode_into_std_write(
+            eval.eval_ref(),
```
yannham: Is `eval_ref` used to serialize exactly as `EvalCommand` but without requiring ownership?

L0r3m1p5um: Yeah, it's basically because Serialize can't be derived for an Arc by default. Although I can enable the rc feature flag for serde if you prefer, then it wouldn't need a separate type.

yannham: Hmm, there are a few gotchas (which is why it's feature-gated), but on the other hand, I don't really know what semantics one could expect from serializing/deserializing (a)rcs, other that the one implemented by `serde`'s `rc` feature. I am tempted to say yes and to get rid of the `EvalRef` boilerplate.

L0r3m1p5um: No problem, I should be able to update the PR sometime tonight. For that matter then I wouldn't need Eval and EvalCommand to be separate either, since Eval is also the same thing, but fully owned for deserialization.

PR #2351, opened 2025-09-16
Remove one unsafe and a fictitious lifetime in the interner

jneem:
I think this is a slight improvement. All lifetimes are `'static'`, which is a little less flexible than the hypothetical version using ouroborous (which could support strings with lifetime tied to the interner). But I don't think we need that flexibility.

There's one extra pointer indirection when allocating a new identifier, because the easiest way to know that the arena doesn't get deallocated (without cheating by knowing that our only usage of `Interner` is as a static) is to allocate and leak it. I can't measure any performance difference.

### DIFF
--- a/core/src/identifier.rs
+++ b/core/src/identifier.rs
@@ -287,7 +287,7 @@ mod interner {
     impl Interner {
         /// Creates an empty [Interner].
         pub(crate) fn new() -> Self {
-            Self(RwLock::new(InnerInterner::new()))
+            Self(RwLock::new(InnerInterner::empty()))
         }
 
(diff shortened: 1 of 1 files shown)


jneem:
Ok, here's another try that doesn't leak. I went with ouroboros to keep the string lifetimes more accurate, although it doesn't actually reduce the unsafety as we still need to extend the lifetimes through the locks. But since we already have it as a dependency, I thought it was ok...

yannham:
(I agree that `'this` is clearer than a phony `'static`, so I do prefer the ouroboros version as well, if only because it makes the intent clear)

PR #2353, opened 2025-09-17
Convert the yaml loader to Ast

jneem:
This PR does two things (in two commits, in case you want to review separately):
- use our own scalar-parsing code instead of saphyr's. This (1) removes our dependency on the higher-level `saphyr` crate (but not `saphyr-parser`) and (2) avoids an extra detour through `saphyr::Scalar`
- makes the yaml loading convert to `Ast` first instead of straight to `RichTerm`. This is going to be useful in the lsp.


core/src/serialize/yaml.rs
```diff
@@ -232,24 +231,95 @@ impl YamlLoader {
...
-                    )
-                })?,
-            ),
-            Scalar::String(s) => Term::Str(s.into_owned().into()),
+        // Returns Ok(Some(f)) if it succeeds, Ok(None) if the argument wasn't a float, or
+        // Err it it was a float that Nickel doesn't support (i.e. inf or nan).
+        //
+        // The motivation here is that we want an error on ".inf" instead of silently treating
+        // it as a string.
+        let parse_float = |v: &str| -> Result<Option<Term>, ParseError> {
```
yannham: Nitpick: this closure is rather non-trivial. Should it be a proper function instead? Does it capture a lot of parameters, beside `pos`? (it can still be a subfunction of the current function)

core/src/serialize/yaml.rs
```diff
@@ -232,24 +231,95 @@ impl YamlLoader {
...
+                    let f = v.parse::<f64>().ok();
+                    // unwrap: we've already checked for inf and nan
+                    Ok(f.map(|f| Term::Num(Number::try_from_float_simplest(f).unwrap())))
+                }
+                _ => Ok(None),
+            }
+        };
+
+        // Parse a YAML scalar, inferring the type from the value itself.
+        let parse = |v: Cow<'_, str>| -> Result<Term, ParseError> {
```
yannham: Ditto closure vs function

core/src/serialize/yaml.rs
```diff
@@ -232,24 +231,95 @@ impl YamlLoader {
...
+                    }
+                }
+            }
         };
+
+        let t = if style != ScalarStyle::Plain {
+            // Any quoted scalar is a string.
+            Term::Str(value.into())
+        } else if let Some(tag) = tag.map(Cow::as_ref) {
+            if tag.is_yaml_core_schema() {
```
yannham: By the way, if we switch to edition 2024 (which I did anyway on the RFC007 PR for that very reason), we have let-chains stabilized and I _believe_ we can rewrite this to avoid the duplicated `else { parse(value)? }` and the double nesting:

```suggestion
        } else if let Some(tag) = tag.map(Cow::as_ref) && tag.is_yaml_core_schema() {
```

core/src/serialize/yaml.rs
```diff
@@ -127,47 +127,48 @@ impl Node {
...
 
-    fn var(&self, anchor_id: NonZeroUsize) -> RichTerm {
-        Term::Var(self.map[&anchor_id].0.into()).into()
+    fn var(&self, anchor_id: NonZeroUsize) -> Ast<'ast> {
+        ast::Node::Var(self.map[&anchor_id].0.into()).into()
     }
 
     /// If a `RichTerm` has an anchor id, replace it with a `Term::Var`.
     ///
     /// (The original `RichTerm` will get hoisted up to the top-level recursive let block.)
```
yannham: The comments might need a pass to be updated.

core/src/serialize/yaml.rs
jneem: I tried switching to edition 2024, but it caused a bunch of errors because of the [match ergonomics changes](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/match-ergonomics.html). I guess you've already fixed these in your PR, so in order to avoid merge conflicts I'll just mark this as TODO for now.

PR #2355, opened 2025-09-18
Eliminate generated ident collisions

jneem:
This changes ident generation so that idents generated with `Ident::fresh` will never clash with idents generated with `Ident::new`. I think this was the originally intended behavior, but as discussed in the now-deleted docs for `Ident::fresh`, it didn't work.

There was a problem with pattern compilation, though, which used to generate idents with `Ident::fresh`, then turn those idents into strings to pass to `%record/insert%`. These strings would get turned back into idents via `Ident::new` and then they'd no longer match the original fresh idents.

I originally tried to solve this by having a special-case `Ident::fake_fresh` that was only used in pattern compilation to revert to the old behavior. It didn't work, because pattern compilation sometimes consumes fresh identifiers coming from other places.

So this PR takes another approach, by allowing record primops to take either an ident (represented as an enum tag) or a string. I'm not sure this is the right approach, and it probably has conflicts with #2302... As discussed somewhere (I can't find it ATM) the better approach might be to revisit how pattern compilation collects its bindings.

### DIFF
--- a/parser/src/identifier.rs
+++ b/parser/src/identifier.rs
@@ -296,6 +296,8 @@ mod interner {
 
     use typed_arena::Arena;
 
+    use super::Bitmap;
+
     /// A symbol is a correspondence between an [Ident](super::Ident) and its string representation
     /// stored in the [Interner].
     #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
@@ -372,7 +374,7 @@ mod interner {
         vec: Vec<&'this str>,
 
         /// Allows checking whether an identifier was generated.
-        generated: Vec<bool>,
+        generated: Bitmap,
     }
 
     impl InnerInterner {
@@ -382,7 +384,7 @@ mod interner {
                 Mutex::new(Arena::new()),
                 |_arena| HashMap::new(),
                 |_arena| Vec::new(),
-                Vec::new(),
+                Bitmap::default(),
             )
         }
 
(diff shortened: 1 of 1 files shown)


jneem:
As discussed in the weekly, maybe instead of having a special path for pattern compilation/idents, pattern compilation should use arrays instead of records.

There's a more general design decision about whether we should be interning as much as we do right now. Maybe not everything that's currently Ident should be interned? (Like, maybe only intern at parsing time?)

jneem:
With the latest pattern compilation changes, this still runs into issues with the fresh identifiers coming from generated contracts. I think the more thorough pattern compilation re-work is still a prerequisite for this.

jneem:
With the new pattern compilation, I think this is finally ok. I tested it by adding a panic whenever `Ident::new` is called with a string starting with `%`, and it only failed on `core/tests/integration/inputs/records/record_defs.ncl`, which explicitly creates record fields starting with `%`. So I'm pretty sure we don't have any more accidental conversions from strings to generated idents.

parser/src/identifier.rs
```diff
@@ -317,6 +370,9 @@ mod interner {
         #[borrows(arena)]
         #[covariant]
         vec: Vec<&'this str>,
+
+        /// Allows checking whether an identifier was generated.
+        generated: Vec<bool>,
```
yannham: Should we use pointer tagging :stuck_out_tongue: ? More seriously though, I'm not sure it has a measurable impact, but a vec of `bool` is represented as bytes, which means we lose 7/8th of the space. Well it means 1MB for 1M ids, which is kinda ok, but I wonder if there some easy, packed, and non-const-generic-based bitmaps that are drop-in replacement for `Vec<bool>`.

jneem: I could do a quick wrapper around `Vec<u64>`? I think it's simple enough to not be worth pulling in an external dep.

PR #2356, opened 2025-09-19
Add local configs to nls

jneem:
This adds a system of "local" configurations to nls, where you can scatter `nls.ncl` files in your directory tree to control how nls acts on certain files. The design of this hasn't been discussed at all yet, so this PR is also a little bit of a mini-RFC.

The basic idea is that your `nls.ncl` looks like

```nickel
{
  "**/shallow.ncl" = { eval_config.eval_limits.recursion_limit = 2 }
}
```

and any file matching the glob "**/shallow.ncl" will get a smaller recursion limit.

When looking for an `nls.ncl` file, each file starts in its current directory and searches up. If it finds a config file with a matching glob, it stops looking.

The main goal for this configuration system is to make it possible to specify nickel contracts to apply to non-nickel files, but I separated it out into its own PR so that we could discuss the design.


yannham:
I'm curious about prior art on the subject. I vaguely remember PHP having some kind of LSP config file. Maybe some other languages have had the same requirements of providing the LSP with more context in order to do a better job.

jneem:
I had a look around, and found a little bit of prior art:

- in the [python-lsp-server docs](https://github.com/python-lsp/python-lsp-server?tab=readme-ov-file#configuration) they point out (as we know) that most configuration comes from the editor's lsp initialization. But since python-lsp-server integrates a lot of existing tools, it also gets config from the global config of those tools.
- the haskell language server configuration docs are also mostly about editor options, but they have a section on [build configuration](https://haskell-language-server.readthedocs.io/en/latest/configuration.html#configuring-your-project-build) where they point out that the language server needs to build stuff, and it tries to read build configuration from the filesystem
- I didn't find explicit documentation for this, but rust-analyzer also gets most of its config from the editor, but when it invokes `cargo check` (for example) it picks up `.cargo/config`

Maybe the message here is that this is the wrong approach? Any nickel-contract-for-yaml configuration should be *Nickel* configuration, not nls configuration. And nls should just pick it up.

yannham:
> Maybe the message here is that this is the wrong approach? Any nickel-contract-for-yaml configuration should be Nickel configuration, not nls configuration. And nls should just pick it up.

I wonder, though. Mapping YAML files to contract do seem, on the surface, like something that is only LSP-related (you'd probably use `--contract` on the CLI, and I don't think applying a contract based on the project's config automatically is a great idea, in term of the principle of least surprise).

That being said, we could still have LSP-specific info in the central manifest, to avoid scattering the config.

jneem:
> you'd probably use --contract on the CLI, and I don't think applying a contract based on the project's config automatically is a great idea, in term of the principle of least surprise

I agree that too much implicit behavior can be a problem, but if you're `nickel eval`ing a YAML file and you don't supply `--contract`, is it really too surprising to look for some configuration for which contract to apply? There's probably only one contract per YAML file that makes sense, and it's probably only ever makes sense to use the same one for `nickel eval` and `nls`...

yannham:
>  is it really too surprising to look for some configuration for which contract to apply? There's probably only one contract per YAML file that makes sense, and it's probably only ever makes sense to use the same one for nickel eval and nls...

For a YAML file, probably. But if we want to extend this capability to Nickel files (as in organist, where the contract is applied by a consumer of the file), then I wonder if it can lead to double applications: what is the behavior with respect to the different possible invocations? Does it also apply if you do something like `nickel eval file1.ncl module.ncl`, and there is a glob `file1.ncl -> contract.ncl` ? If module is defined as something like `{ mod.inputs | Contract, outputs = {..} }` and `file1.ncl` is `{ mod.inputs = {..}}`, that this will be a double application. Maybe it can be elided, maybe not.

I'm not sure it's really useful useful, but what if one wants to explicitly export the file without applying the contract?

I don't have a strong opinion currently but I do have a feeling that this is a bit ad-hoc (this is not a general mechanism for merging or whatnot, it's specific to contract application to files) and implicit. I fear this can lead to surprising results for occasional developers, such as where this damned contract application is coming from.

PR #2359, opened 2025-09-23
Use Ast instead of RichTerm for typechecking error

yannham:
When migrating the typechecker to the new AST, the boundary between `typecheck::error` errors and the main `error::TypecheckError` had been a point of friction. Indeed, the typechecker internal errors reference to allocator-borrowing ASTs, but in the error path, those structure must escape their orignal allocator.

To make things simpler, we've somehow hackily converted arena-allocated ASTs and AST-based types to the old/runtime representation (`RichTerm`) at this boundary. `RichTerm`s and `RichTerm`-based types are heap-allocated and don't require an allocator.

However, this is artificial: during typechecking, there's no reason to refer to such a runtime representation. It came back to bite me in the compact value work, where the new runtime representation requires extra data structures (e.g. the position table), which we don't have at hand during typechecking and are annoying to provide or fake.

This commit/PR bites the bullet and implement the clean solution which is to:

1. Endow typechecking errors with their own allocator
2. Copy any required arena-allocated data into the error allocator, so that they can outlive the function and the typechecker

This is reasonably easy to do, thanks to our `CloneTo` trait and `ouroboros`.

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -2822,7 +2822,7 @@ impl<'ast> IntoDiagnostics for &'_ TypecheckErrorData<'ast> {
             }
             // clone() here is unfortunate, but I haven't found a better way to interface typecheck
             // errors - which can generate a diagnostic by reference - from other errors (import
-            // errrors can themselves hide parsing errors), where `into_diagnostics` consume `self`
+            // errors can themselves hide parsing errors), where `into_diagnostics` consume `self`
             // in the current implementation. Maybe we should migrate from `into_diagnostics` to
             // `to_diagnostic`, taking the error by reference, but this might cause more copying.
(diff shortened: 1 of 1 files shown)


core/src/error/mod.rs
```diff
@@ -2775,7 +2820,12 @@ impl IntoDiagnostics for TypecheckError {
                             .into(),
                     ])]
             }
-            TypecheckError::ImportError(err) => err.into_diagnostics(files),
+            // clone() here is unfortunate, but I haven't found a better way to interface typecheck
+            // errors - which can generate a diagnostic by reference - from other errors (import
+            // errrors can themselves hide parsing errors), where `into_diagnostics` consume `self`
```
jneem:
```suggestion
            // errors can themselves hide parsing errors), where `into_diagnostics` consume `self`
```

PR #2360, opened 2025-09-23
Start on the C api

jneem:
This adds C bindings to the public API. I think the most important basics are covered (inspecting values, deep/shallow eval, printing errors).

I've also added some very basic tests, included in `nix flake check`. However, I haven't updated the release workflows to actually distribute anything. I'm not quite sure what (if anything) makes sense to distribute, actually. Maybe `libnickel_lang.a` and `nickel_lang.h`?

### DIFF
--- a/nickel/src/capi.rs
+++ b/nickel/src/capi.rs
@@ -15,9 +15,6 @@ use nickel_lang_core::term::Term;
 
 use crate::{Array, Context, Error, ErrorFormat, Expr, Number, Record, Trace, VirtualMachine};
 
-// TODO: serialization functions
-// TODO: error reporting
-
 /// The main entry point.
 pub struct nickel_context {}
 
@@ -202,6 +199,18 @@ pub enum nickel_error_format {
     NICKEL_ERROR_FORMAT_YAML = 3,
     /// Format an error as TOML.
     NICKEL_ERROR_FORMAT_TOML = 4,
+}
+
+impl From<nickel_error_format> for ErrorFormat {
+    fn from(e: nickel_error_format) -> Self {
+        match e {
+            nickel_error_format::NICKEL_ERROR_FORMAT_TEXT => ErrorFormat::Text,
+            nickel_error_format::NICKEL_ERROR_FORMAT_ANSI_TEXT => ErrorFormat::AnsiText,
+            nickel_error_format::NICKEL_ERROR_FORMAT_JSON => ErrorFormat::Json,
+            nickel_error_format::NICKEL_ERROR_FORMAT_YAML => ErrorFormat::Yaml,
+            nickel_error_format::NICKEL_ERROR_FORMAT_TOML => ErrorFormat::Toml,
+        }
+    }
 }
 
 struct CTrace {
@@ -651,6 +660,69 @@ pub unsafe extern "C" fn nickel_expr_as_array(expr: *const nickel_expr) -> *cons
     nickel_expr::as_rust(&expr).as_array().unwrap().into()
 }
 
+unsafe fn export_result(
+    result: Result<String, Error>,
+    out_string: *mut nickel_string,
+    out_err: *mut nickel_error,
+) -> nickel_result {
+    match result {
+        Ok(s) => {
+            if !out_string.is_null() {
+                (*out_string).inner = s;
+            }
+            nickel_result::NICKEL_RESULT_OK
+        }
+        Err(e) => {
+            if !out_err.is_null() {
+                (*out_err).inner = Some(e);
+            }
+            nickel_result::NICKEL_RESULT_ERR
+        }
+    }
+}
+
+/// Convert this expression to JSON.
+///
+/// This is fallible because enum variants have no canonical conversion to
+/// JSON: if the expression contains any enum variants, this will fail.
+/// This also fails if the expression contains any unevaluated sub-expressions.
+#[no_mangle]
+pub unsafe extern "C" fn nickel_expr_to_json(
+    expr: *const nickel_expr,
+    out_string: *mut nickel_string,
+    out_err: *mut nickel_error,
+) -> nickel_result {
+    export_result(nickel_expr::as_rust(&expr).to_json(), out_string, out_err)
+}
+
+/// Convert this expression to YAML.
+///
+/// This is fallible because enum variants have no canonical conversion to
+/// YAML: if the expression contains any enum variants, this will fail.
+/// This also fails if the expression contains any unevaluated sub-expressions.
+#[no_mangle]
+pub unsafe extern "C" fn nickel_expr_to_yaml(
+    expr: *const nickel_expr,
+    out_string: *mut nickel_string,
+    out_err: *mut nickel_error,
+) -> nickel_result {
+    export_result(nickel_expr::as_rust(&expr).to_yaml(), out_string, out_err)
+}
+
+/// Convert this expression to TOML.
+///
+/// This is fallible because enum variants have no canonical conversion to
+/// TOML: if the expression contains any enum variants, this will fail.
+/// This also fails if the expression contains any unevaluated sub-expressions.
+#[no_mangle]
+pub unsafe extern "C" fn nickel_expr_to_toml(
+    expr: *const nickel_expr,
+    out_string: *mut nickel_string,
+    out_err: *mut nickel_error,
+) -> nickel_result {
+    export_result(nickel_expr::as_rust(&expr).to_toml(), out_string, out_err)
+}
+
 /// Is this number an integer within the range of an `int64_t`?
 #[no_mangle]
 pub unsafe extern "C" fn nickel_number_is_i64(num: *const nickel_number) -> c_int {
@@ -897,14 +969,6 @@ pub unsafe extern "C" fn nickel_error_format(
     write_payload: *const c_void,
     format: nickel_error_format,
 ) -> nickel_result {
-    let format = match format {
-        nickel_error_format::NICKEL_ERROR_FORMAT_TEXT => ErrorFormat::Text,
-        nickel_error_format::NICKEL_ERROR_FORMAT_ANSI_TEXT => ErrorFormat::AnsiText,
-        nickel_error_format::NICKEL_ERROR_FORMAT_JSON => ErrorFormat::Json,
-        nickel_error_format::NICKEL_ERROR_FORMAT_YAML => ErrorFormat::Yaml,
-        nickel_error_format::NICKEL_ERROR_FORMAT_TOML => ErrorFormat::Toml,
-    };
-
     let err = err
         .as_ref()
         .unwrap()
@@ -916,9 +980,36 @@ pub unsafe extern "C" fn nickel_error_format(
         flush: None,
         context: write_payload,
     };
-    if err.format(&mut write, format).is_err() {
+    if err.format(&mut write, format.into()).is_err() {
         nickel_result::NICKEL_RESULT_ERR
     } else {
+        nickel_result::NICKEL_RESULT_OK
+    }
+}
+
+/// Write out an error as a user- or machine-readable diagnostic.
+///
+/// This is like `nickel_error_format`, but writes the error to a string instead
+/// of via a callback function.
+pub unsafe extern "C" fn nickel_error_format_as_string(
+    err: *const nickel_error,
+    out_string: *mut nickel_string,
+    format: nickel_error_format,
+) -> nickel_result {
+    let err = err
+        .as_ref()
+        .unwrap()
+        .inner
+        .as_ref()
+        .expect("uninitialized error");
+    let mut out = Vec::new();
+    if err.format(&mut out, format.into()).is_err() {
+        nickel_result::NICKEL_RESULT_ERR
+    } else {
+        if !out_string.is_null() {
+            // unwrap: Nickel's formatting always produces UTF-8.
+            (*out_string).inner = String::from_utf8(out).unwrap();
+        }
         nickel_result::NICKEL_RESULT_OK
     }
 }
--- a/nickel/src/capi.rs
+++ b/nickel/src/capi.rs
@@ -1,6 +1,41 @@
 //! C bindings for the Nickel language
 //!
-//! TODO: example
+//! These bindings cover
+//! - evaluation (including lazy/partial evaluation),
+//! - inspection of the results as structured values,
+//! - serialization to JSON, TOML, and YAML,
+//! - error reporting.
+//!
+//! Here is a basic example that evaluates a Nickel expression
+//! and inspects the result. For more examples, see the C api
+//! tests in the Nickel repository.
+//!
+//! ```c
+//! nickel_context *ctx = nickel_context_alloc();
+//! nickel_expr *expr = nickel_expr_alloc();
+//!
+//! /* Passing a null pointer as the "error" param is the easy option, but we won't get
+//!    nice diagnostics. */
+//! nickel_result result = nickel_context_eval_deep(ctx, "{ foo | Number = 1 }", expr, NULL);
+//! assert(result == NICKEL_RESULT_OK);
+//!
+//! /* Now the result of evaluation is in `expr`. It's a record with one element */
+//! assert(nickel_expr_is_record(expr));
+//! nickel_record const *rec = nickel_expr_as_record(expr);
+//! assert(nickel_record_len(rec) == 1);
+//!
+//! /* The single element of the record has key "foo" and value `1` */
+//! char const* key;
+//! uintptr_t len;
+//! nickel_expr *val = nickel_expr_alloc();
+//! nickel_record_key_value_by_index(rec, 0, &key, &len, val);
+//!
+//! /* Careful: the strings coming from the nickel API are not null-terminated! */
+//! assert(len == 3);
+//! assert(!strncmp(key, "foo", 3));
+//! assert(nickel_expr_is_number(val));
+//! assert(nickel_expr_as_i64(val) == 1);
+//! ```
 
 #![allow(clippy::missing_safety_doc)]
 #![allow(non_camel_case_types)]
@@ -275,13 +310,16 @@ pub unsafe extern "C" fn nickel_context_set_source_name(
 }
 
 /// Perform some sort of evaluation, and return the error appropriately.
-unsafe fn do_eval<F: FnOnce(&mut crate::Context, &str) -> Result<Expr, Error>>(
+unsafe fn do_eval<F>(
     f: F,
     mut ctx: *mut nickel_context,
     src: *const c_char,
     mut out_expr: *mut nickel_expr,
     out_error: *mut nickel_error,
-) -> nickel_result {
+) -> nickel_result
+where
+    F: FnOnce(&mut crate::Context, &str) -> Result<Expr, Error>,
+{
     let src = CStr::from_ptr(src).to_str().unwrap();
     match f(nickel_context::as_rust_mut(&mut ctx), src) {
         Ok(expr) => {
@@ -992,6 +1030,7 @@ pub unsafe extern "C" fn nickel_error_display(
 ///
 /// This is like `nickel_error_format`, but writes the error to a string instead
 /// of via a callback function.
+#[no_mangle]
 pub unsafe extern "C" fn nickel_error_format_as_string(
     err: *const nickel_error,
     out_string: *mut nickel_string,
--- a/nickel/src/capi.rs
+++ b/nickel/src/capi.rs
@@ -172,7 +172,7 @@ pub unsafe extern "C" fn nickel_context_free(ctx: *mut nickel_context) {
 // a nullable function pointer (since rust fns are never null).
 // https://github.com/mozilla/cbindgen/issues/326#issuecomment-584288686
 pub type nickel_write_callback =
-    Option<extern "C" fn(context: *const c_void, buf: *const u8, len: usize) -> usize>;
+    Option<extern "C" fn(context: *mut c_void, buf: *const u8, len: usize) -> usize>;
 
 /// A callback function for flushing data that was written by a write callback.
 pub type nickel_flush_callback = Option<extern "C" fn(context: *const c_void)>;
@@ -216,7 +216,7 @@ impl From<nickel_error_format> for ErrorFormat {
 struct CTrace {
     write: nickel_write_callback,
     flush: nickel_flush_callback,
-    context: *const c_void,
+    context: *mut c_void,
 }
 
 impl Write for CTrace {
@@ -248,7 +248,7 @@ pub unsafe extern "C" fn nickel_context_set_trace_callback(
     mut ctx: *mut nickel_context,
     write: nickel_write_callback,
     flush: nickel_flush_callback,
-    user_data: *const c_void,
+    user_data: *mut c_void,
 ) {
     let trace = Trace::new(CTrace {
         write,
@@ -462,9 +462,9 @@ pub unsafe extern "C" fn nickel_context_eval_shallow(
 /// nickel_expr *field = nickel_expr_alloc();
 /// nickel_record_value_by_name(rec, "foo", field);
 ///
-/// /* Now `rec` points to data owned by `expr`, but `field`
-///    owns its own data. The following deallocation invalidates
-///    `rec`, but not `field`. */
+/// // Now `rec` points to data owned by `expr`, but `field`
+/// // owns its own data. The following deallocation invalidates
+/// // `rec`, but not `field`.
 /// nickel_expr_free(expr);
 /// printf("number: %d\n", nickel_expr_is_number(field));
 /// ```
@@ -963,10 +963,11 @@ pub unsafe extern "C" fn nickel_error_free(err: *mut nickel_error) {
 /// - `write` is a callback function that will be invoked with UTF-8 encoded data.
 /// - `write_payload` is optional extra data to pass to `write`
 /// - `format` selects the error-rendering format.
-pub unsafe extern "C" fn nickel_error_format(
+#[no_mangle]
+pub unsafe extern "C" fn nickel_error_display(
     err: *const nickel_error,
     write: nickel_write_callback,
-    write_payload: *const c_void,
+    write_payload: *mut c_void,
     format: nickel_error_format,
 ) -> nickel_result {
     let err = err
--- a/nickel/src/capi.rs
+++ b/nickel/src/capi.rs
@@ -1,0 +1,838 @@
+//! C bindings for the Nickel language
+//!
+//! TODO: example
+
+#![allow(clippy::missing_safety_doc)]
+#![allow(non_camel_case_types)]
+#![deny(missing_docs)]
+
+use std::{
+    ffi::{c_char, c_int, c_void, CStr},
+    io::Write,
+};
+
+use nickel_lang_core::term::Term;
+
+use crate::{Array, Context, Error, Expr, Number, Record, Trace, VirtualMachine};
+
+// TODO: put wrapper structs around Error
+// just so that we can have C-API-specific documentation for it
+// TODO: serialization functions
+// TODO: error reporting
+
+/// The main entry point.
+pub struct nickel_context {}
+
+impl nickel_context {
+    unsafe fn as_rust_mut(this: &mut *mut Self) -> &mut Context {
+        (*this as *mut Context).as_mut().unwrap()
+    }
+}
+
+/// A Nickel expression.
+///
+/// This might be fully evaluated (for example, if you got it from [`nickel_context_eval_deep`])
+/// or might have unevaluated sub-expressions (if you got it from [`nickel_context_eval_shallow`]).
+// This is only used to hide `Expr`, and to attach C-API-specific documentation to.
+// We only ever use pointers to `nickel_expr`, which are secretly pointers to `Expr`.
+pub struct nickel_expr {}
+
+impl nickel_expr {
+    /// Convert a *nickel_expr to a &Expr.
+    ///
+    /// # Safety
+    ///
+    /// Assumes that `this` was originally a valid pointer to Expr.
+    unsafe fn as_rust(this: &*const Self) -> &Expr {
+        (*this as *const Expr).as_ref().unwrap()
+    }
+
+    /// Convert a *mut nickel_expr to a &mut Expr.
+    ///
+    /// # Safety
+    ///
+    /// Assumes that `this` was originally a valid pointer to Expr.
+    unsafe fn as_rust_mut(this: &mut *mut Self) -> &mut Expr {
+        (*this as *mut Expr).as_mut().unwrap()
+    }
+}
+
+/// A Nickel array.
+///
+/// See [`nickel_expr_is_array`] and [`nickel_expr_as_array`].
+// This is not really an empty struct, it's just an opaque type to
+// hide the implementation. The C API only ever uses `*const Array`,
+// and they are secretly pointers to `nickel_lang_core::term::array::Array`.
+pub struct nickel_array {}
+
+impl nickel_array {
+    unsafe fn as_rust(this: &*const Self) -> Array<'_> {
+        Array {
+            array: (*this as *const nickel_lang_core::term::array::Array)
+                .as_ref()
+                .unwrap(),
+        }
+    }
+}
+
+impl<'a> From<Array<'a>> for *const nickel_array {
+    fn from(arr: Array<'a>) -> Self {
+        arr.array as *const _ as *const nickel_array
+    }
+}
+
+/// A Nickel record.
+///
+/// See [`nickel_expr_is_record`] and [`nickel_expr_as_record`].
+// This is not really an empty struct, it's just an opaque type to
+// hide the implementation. The C API only ever uses `*const Record`,
+// and they are secretly pointers to `nickel_lang_core::term::record::RecordData`.
+pub struct nickel_record {}
+
+impl nickel_record {
+    unsafe fn as_rust(this: &*const Self) -> Record<'_> {
+        Record {
+            data: (*this as *const nickel_lang_core::term::record::RecordData)
+                .as_ref()
+                .unwrap(),
+        }
+    }
+}
+
+impl<'a> From<Record<'a>> for *const nickel_record {
+    fn from(rec: Record<'a>) -> Self {
+        rec.data as *const _ as *const nickel_record
+    }
+}
+
+/// A Nickel number.
+///
+/// See [`nickel_expr_is_number`] and [`nickel_expr_as_number`].
+// This is not really an empty struct, it's just an opaque type to
+// hide the implementation. The C API only ever uses `*const Number`,
+// and they are secretly pointers to `nickel_lang_core::term::Number`.
+pub struct nickel_number {}
+
+impl nickel_number {
+    unsafe fn as_rust(this: &*const Self) -> Number<'_> {
+        Number {
+            num: (*this as *const nickel_lang_core::term::Number)
+                .as_ref()
+                .unwrap(),
+        }
+    }
+}
+
+impl<'a> From<Number<'a>> for *const nickel_number {
+    fn from(n: Number<'a>) -> Self {
+        n.num as *const _ as *const nickel_number
+    }
+}
+
+/// A Nickel virtual machine, which can be used for evaluating partially-evaluated expressions.
+pub struct nickel_virtual_machine {
+    inner: Option<VirtualMachine>,
+}
+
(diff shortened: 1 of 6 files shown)


nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+    inner: Option<Error>,
+}
+
+/// A Nickel expression.
+///
+/// This might be fully evaluated (for example, if you got it from [`nickel_context_eval_deep`])
+/// or might have unevaluated sub-expressions (if you got it from [`nickel_context_eval_shallow`]).
+// This is only used to hide `Expr`, and to attach C-API-specific documentation to.
+// We only ever use pointers to `nickel_expr`, which are secretly pointers to `Expr`.
+pub struct nickel_expr {}
```
yannham: Is this a common pattern in for C APIs in Rust (to use a lower_case empty struct as a wrapper)? For example, what benefit does it have over an actual wrapper `NickelExpr { inner: ... }`? I also see on some other project that you can use a renaming feature in  `cbindgen.toml` (https://github.com/Tencent/tquic/blob/b62863afb8c59828b55393d5ea09c24f79e8efd6/cbindgen.toml#L23-L58).

jneem: I'm not sure if this is a common pattern, but it was the only way I found that:
- has custom docs for the C code (the renaming feature just copies the rust docs over, which led to confusing references to things like `Context::eval_deep` instead of `nickel_context_eval_deep`)
- allows casting from reference-to-the-C-thing to reference-to-the-rust-thing. (For some types this wasn't needed, like `nickel_error`, and so there I just used a struct with an inner member.)
- hides the inner type. I tried solving the previous point with `#[repr(transparent)]`, but cbindgen would then leak the inner name (like "Context")

Basically, this pattern gave me the generated header that I wanted, which I just realized isn't easily visible in this PR because it doesn't get checked in. I'll attach it in the main conversation thread.

jneem:
Here's the generated header:

```c
// SPDX-License-Identifier: MIT

#ifndef NICKEL_LANG_H
#define NICKEL_LANG_H

#include <stdint.h>

/**
 * For functions that can fail, these are the interpretations of the return value.
 */
typedef enum {
    /**
     * Format an error as human-readable text.
     */
    NICKEL_ERROR_FORMAT_TEXT = 0,
    /**
     * Format an error as human-readable text, with ANSI color codes.
     */
    NICKEL_ERROR_FORMAT_ANSI_TEXT = 1,
    /**
     * Format an error as JSON.
     */
    NICKEL_ERROR_FORMAT_JSON = 2,
    /**
     * Format an error as YAML.
     */
    NICKEL_ERROR_FORMAT_YAML = 3,
    /**
     * Format an error as TOML.
     */
    NICKEL_ERROR_FORMAT_TOML = 4,
} nickel_error_format;

/**
 * For functions that can fail, these are the interpretations of the return value.
 */
typedef enum {
    /**
     * A successful result.
     */
    NICKEL_RESULT_OK = 0,
    /**
     * A bad result.
     */
    NICKEL_RESULT_ERR = 1,
} nickel_result;

/**
 * A Nickel array.
 *
 * See [`nickel_expr_is_array`] and [`nickel_expr_as_array`].
 */
typedef struct nickel_array nickel_array;

/**
 * The main entry point.
 */
typedef struct nickel_context nickel_context;

/**
 * A Nickel error.
 *
 * If you want to collect an error message from a fallible function
 * (like `nickel_context_eval_deep`), first allocate an error using
 * `nickel_error_alloc`, and then pass the resulting pointer to your fallible
 * function. If that function fails, it will save the error data in your
 * `nickel_error`.
 */
typedef struct nickel_error nickel_error;

/**
 * A Nickel expression.
 *
 * This might be fully evaluated (for example, if you got it from [`nickel_context_eval_deep`])
 * or might have unevaluated sub-expressions (if you got it from [`nickel_context_eval_shallow`]).
 */
typedef struct nickel_expr nickel_expr;

/**
 * A Nickel number.
 *
 * See [`nickel_expr_is_number`] and [`nickel_expr_as_number`].
 */
typedef struct nickel_number nickel_number;

/**
 * A Nickel record.
 *
 * See [`nickel_expr_is_record`] and [`nickel_expr_as_record`].
 */
typedef struct nickel_record nickel_record;

/**
 * A Nickel string.
 */
typedef struct nickel_string nickel_string;

/**
 * A Nickel virtual machine, which can be used for evaluating partially-evaluated expressions.
 */
typedef struct nickel_virtual_machine nickel_virtual_machine;

/**
 * A callback function for writing data.
 *
 * This function will be called with a buffer (`buf`) of data, having length
 * `len`. It need not consume the entire buffer, and should return the number
 * of bytes consumed.
 */
typedef uintptr_t (*nickel_write_callback)(void *context, const uint8_t *buf, uintptr_t len);

/**
 * A callback function for flushing data that was written by a write callback.
 */
typedef void (*nickel_flush_callback)(const void *context);

#ifdef __cplusplus
extern "C" {
#endif // __cplusplus

/**
 * Allocate a new [`nickel_context`], which can be used to evaluate Nickel expressions.
 *
 * Returns a newly-allocated [`nickel_context`] that can be freed with [`nickel_context_free`].
 */
nickel_context *nickel_context_alloc(void);

/**
 * Free a [`nickel_context`] that was created with [`nickel_context_alloc`].
 */
void nickel_context_free(nickel_context *ctx);

/**
 * Provide a callback that will be called when evaluating Nickel
 * code that uses `std.trace`.
 */
void nickel_context_set_trace_callback(nickel_context *ctx,
                                       nickel_write_callback write,
                                       nickel_flush_callback flush,
                                       void *user_data);

/**
 * Provide a name for the main input program.
 *
 * This is used to format error messages. If you read the main input
 * program from a file, its path is a good choice.
 *
 * `name` should be a UTF-8-encoded, nul-terminated string. It is only
 * borrowed temporarily; the pointer need not remain valid.
 */
void nickel_context_set_source_name(nickel_context *ctx, const char *name);

/**
 * Evaluate a Nickel program deeply.
 *
 * "Deeply" means that we recursively evaluate records and arrays. For
 * an alternative, see [`nickel_context_eval_shallow`].
 *
 * - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
 * - `out_expr` either NULL or something that was created with [`nickel_expr_alloc`]
 * - `out_error` can be NULL if you aren't interested in getting detailed
 *   error messages
 *
 * If evaluation is successful, returns `NICKEL_RESULT_OK` and replaces
 * the value at `out_expr` (if non-NULL) with the newly-evaluated Nickel expression.
 *
 * If evaluation fails, returns `NICKEL_RESULT_ERR` and replaces the
 * value at `out_error` (if non-NULL) by a pointer to a newly-allocated Nickel error.
 * That error should be freed with `nickel_error_free` when you are
 * done with it.
 */
nickel_result nickel_context_eval_deep(nickel_context *ctx,
                                       const char *src,
                                       nickel_expr *out_expr,
                                       nickel_error *out_error);

/**
 * Evaluate a Nickel program deeply.
 *
 * This differs from [`nickel_context_eval_deep`] in that it ignores
 * fields marked as `not_exported`.
 *
 * - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
 * - `out_expr` either NULL or something that was created with [`nickel_expr_alloc`]
 * - `out_error` can be NULL if you aren't interested in getting detailed
 *   error messages
 *
 * If evaluation is successful, returns `NICKEL_RESULT_OK` and replaces
 * the value at `out_expr` (if non-NULL) with the newly-evaluated Nickel expression.
 *
 * If evaluation fails, returns `NICKEL_RESULT_ERR` and replaces the
 * value at `out_error` (if non-NULL) by a pointer to a newly-allocated Nickel error.
 * That error should be freed with `nickel_error_free` when you are
 * done with it.
 */
nickel_result nickel_context_eval_deep_for_export(nickel_context *ctx,
                                                  const char *src,
                                                  nickel_expr *out_expr,
                                                  nickel_error *out_error);

/**
 * Evaluate a Nickel program to weak head normal form (WHNF).
 *
 * The result of this evaluation is a null, bool, number, string,
 * enum, record, or array. In case it's a record, array, or enum
 * variant, the payload (record values, array elements, or enum
 * payloads) will be left unevaluated.
 *
 * Together with the expression, this returns a Nickel virtual machine that
 * can be used to further evaluate unevaluated sub-expressions.
 *
 * - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
 * - `out_expr` is either NULL or something that was created with [`nickel_expr_alloc`]
 * - `out_error` can be NULL if you aren't interested in getting detailed
 *   error messages
 * - `out_virtual_machine` is either NULL or something that was created
 *   with [`nickel_virtual_machine_alloc`]
 *
 * If evaluation is successful, returns `NICKEL_RESULT_OK` and replaces
 * the value at `out_expr` (if non-NULL) with the newly-evaluated Nickel expression,
 * and the value at `out_virtual_machine` (if non-NULL) with a virtual machine
 * that can be used for further evaluation.
 *
 * If evaluation fails, returns `NICKEL_RESULT_ERR` and replaces the
 * value at `out_error` (if non-NULL) by a pointer to a newly-allocated Nickel error.
 * That error should be freed with `nickel_error_free` when you are
 * done with it.
 */
nickel_result nickel_context_eval_shallow(nickel_context *ctx,
                                          const char *src,
                                          nickel_expr *out_expr,
                                          nickel_virtual_machine *out_virtual_machine,
                                          nickel_error *out_error);

/**
 * Allocate a new Nickel expression.
 *
 * The returned expression pointer can be used to store the results of
 * evaluation, for example by passing it as the `out_expr` location of
 * `nickel_context_eval_deep`.
 *
 * Each call to `nickel_expr_alloc` should be paired with a call to
 * `nickel_expr_free`. The various functions (like `nickel_context_eval_deep`)
 * that take an `out_expr` parameter overwrite the existing expression
 * contents, and do not affect the pairing of `nickel_expr_alloc` and
 * `nickel_expr_free`.
 *
 * For example:
 *
 * ```c
 * nickel_context *ctx = nickel_context_alloc();
 * nickel_context *expr = nickel_expr_alloc();
 *
 * nickel_context_eval_deep(ctx, "{ foo = 1 }", expr, NULL);
 *
 * // now expr is a record
 * printf("record: %d\n", nickel_expr_is_record(expr));
 *
 * nickel_context_eval_deep(ctx, "[1, 2, 3]", expr, NULL);
 *
 * // now expr is an array
 * printf("array: %d\n", nickel_expr_is_array(expr));
 *
 * // the calls to nickel_context_eval_deep haven't created any new exprs:
 * // we only need to free it once
 * nickel_expr_free(expr);
 * nickel_context_free(ctx);
 * ```
 *
 * An `Expr` owns its data. There are various ways to get a reference to
 * data owned by an expression, which are then invalidated when the expression
 * is freed (by `nickel_expr_free`) or overwritten (for example, by
 * `nickel_context_deep_eval`).
 *
 * ```c
 * nickel_context *ctx = nickel_context_alloc();
 * nickel_expr *expr = nickel_expr_alloc();
 *
 * nickel_context_eval_deep(ctx, "{ foo = 1 }", expr, NULL);
 *
 * nickel_record *rec = nickel_expr_as_record(expr);
 * nickel_expr *field = nickel_expr_alloc();
 * nickel_record_value_by_name(rec, "foo", field);
 *
 * // Now `rec` points to data owned by `expr`, but `field`
 * // owns its own data. The following deallocation invalidates
 * // `rec`, but not `field`.
 * nickel_expr_free(expr);
 * printf("number: %d\n", nickel_expr_is_number(field));
 * ```
 */
nickel_expr *nickel_expr_alloc(void);

/**
 * Free a Nickel expression.
 *
 * See [`nickel_expr_alloc`].
 */
void nickel_expr_free(nickel_expr *expr);

/**
 * Is this expression a boolean?
 */
int nickel_expr_is_bool(const nickel_expr *expr);

/**
 * Is this expression a number?
 */
int nickel_expr_is_number(const nickel_expr *expr);

/**
 * Is this expression a string?
 */
int nickel_expr_is_str(const nickel_expr *expr);

/**
 * Is this expression an enum tag?
 */
int nickel_expr_is_enum_tag(const nickel_expr *expr);

/**
 * Is this expression an enum variant?
 */
int nickel_expr_is_enum_variant(const nickel_expr *expr);

/**
 * Is this expression a record?
 */
int nickel_expr_is_record(const nickel_expr *expr);

/**
 * Is this expression an array?
 */
int nickel_expr_is_array(const nickel_expr *expr);

/**
 * Has this expression been evaluated?
 *
 * An evaluated expression is either null, or it's a number, bool, string,
 * record, array, or enum. If this expression is not a value, you probably
 * got it from looking inside the result of [`nickel_context_eval_shallow`],
 * and you can use the [`VirtualMachine`] you got from
 * `nickel_context_eval_shallow` to evaluate this expression further.
 */
int nickel_expr_is_value(const nickel_expr *expr);

/**
 * Is this expression null?
 */
int nickel_expr_is_null(const nickel_expr *expr);

/**
 * If this expression is a boolean, returns that boolean.
 *
 * # Panics
 *
 * Panics if `expr` is not a boolean.
 */
int nickel_expr_as_bool(const nickel_expr *expr);

/**
 * If this expression is a string, returns that string.
 *
 * A pointer to the string contents, which are UTF-8 encoded, is returned in
 * `out_str`. These contents are *not* nul-terminated. The return value of this
 * function is the length of these contents.
 *
 * The returned string contents are owned by this `Expr`, and will be invalidated
 * when the `Expr` is freed with [`nickel_expr_free`].
 *
 * # Panics
 *
 * Panics if `expr` is not a string.
 */
uintptr_t nickel_expr_as_str(const nickel_expr *expr, const char **out_str);

/**
 * If this expression is a number, returns the number.
 *
 * The returned number pointer borrows from `expr`, and will be invalidated
 * when `expr` is overwritten or freed.
 *
 * # Panics
 *
 * Panics if `expr` is not an number.
 */
const nickel_number *nickel_expr_as_number(const nickel_expr *expr);

/**
 * If this expression is an enum tag, returns its string value.
 *
 * A pointer to the string contents, which are UTF-8 encoded, is returned in
 * `out_str`. These contents are *not* nul-terminated. The return value of this
 * function is the length of these contents.
 *
 * The returned string contents point to an interned string and will never be
 * invalidated.
 *
 * # Panics
 *
 * Panics if `expr` is null or is not an enum tag.
 */
uintptr_t nickel_expr_as_enum_tag(const nickel_expr *expr, const char **out_str);

/**
 * If this expression is an enum variant, returns its string value and its payload.
 *
 * A pointer to the string contents, which are UTF-8 encoded, is returned in
 * `out_str`. These contents are *not* nul-terminated. The return value of this
 * function is the length of these contents.
 *
 * The returned string contents point to an interned string and will never be
 * invalidated.
 *
 * # Panics
 *
 * Panics if `expr` is not an enum tag.
 */
uintptr_t nickel_expr_as_enum_variant(const nickel_expr *expr,
                                      const char **out_str,
                                      nickel_expr *out_expr);

/**
 * If this expression is a record, returns the record.
 *
 * The returned record pointer borrows from `expr`, and will be invalidated
 * when `expr` is overwritten or freed.
 *
 * # Panics
 *
 * Panics if `expr` is not an record.
 */
const nickel_record *nickel_expr_as_record(const nickel_expr *expr);

/**
 * If this expression is an array, returns the array.
 *
 * The returned array pointer borrows from `expr`, and will be invalidated
 * when `expr` is overwritten or freed.
 *
 * # Panics
 *
 * Panics if `expr` is not an array.
 */
const nickel_array *nickel_expr_as_array(const nickel_expr *expr);

/**
 * Convert this expression to JSON.
 *
 * This is fallible because enum variants have no canonical conversion to
 * JSON: if the expression contains any enum variants, this will fail.
 * This also fails if the expression contains any unevaluated sub-expressions.
 */
nickel_result nickel_expr_to_json(const nickel_expr *expr,
                                  nickel_string *out_string,
                                  nickel_error *out_err);

/**
 * Convert this expression to YAML.
 *
 * This is fallible because enum variants have no canonical conversion to
 * YAML: if the expression contains any enum variants, this will fail.
 * This also fails if the expression contains any unevaluated sub-expressions.
 */
nickel_result nickel_expr_to_yaml(const nickel_expr *expr,
                                  nickel_string *out_string,
                                  nickel_error *out_err);

/**
 * Convert this expression to TOML.
 *
 * This is fallible because enum variants have no canonical conversion to
 * TOML: if the expression contains any enum variants, this will fail.
 * This also fails if the expression contains any unevaluated sub-expressions.
 */
nickel_result nickel_expr_to_toml(const nickel_expr *expr,
                                  nickel_string *out_string,
                                  nickel_error *out_err);

/**
 * Is this number an integer within the range of an `int64_t`?
 */
int nickel_number_is_i64(const nickel_number *num);

/**
 * If this number is an integer within the range of an `int64_t`, returns it.
 *
 * # Panics
 *
 * Panics if this number is not an integer in the appropriate range (you should
 * check with [`nickel_number_is_i64`] first).
 */
int64_t nickel_number_as_i64(const nickel_number *num);

/**
 * The value of this number, rounded to the nearest `double`.
 */
double nickel_number_as_f64(const nickel_number *num);

/**
 * The value of this number, as an exact rational number.
 *
 * - `out_numerator` must have been allocated with [`nickel_string_alloc`]. It
 *   will be overwritten with the numerator, as a decimal string.
 * - `out_denominator` must have been allocated with [`nickel_string_alloc`].
 *   It will be overwritten with the denominator, as a decimal string.
 */
void nickel_number_as_rational(const nickel_number *num,
                               nickel_string *out_numerator,
                               nickel_string *out_denominator);

/**
 * The number of elements of this Nickel array.
 */
uintptr_t nickel_array_len(const nickel_array *arr);

/**
 * Retrieve the element at the given array index.
 *
 * The retrieved element will be written to `out_expr`, which must have been allocated with
 * [`nickel_expr_alloc`].
 *
 * # Panics
 *
 * Panics if the given index is out of bounds.
 */
void nickel_array_get(const nickel_array *arr, uintptr_t idx, nickel_expr *out_expr);

/**
 * The number of keys in this Nickel record.
 */
uintptr_t nickel_record_len(const nickel_record *rec);

/**
 * Retrieve the key and value at the given index.
 *
 * If this record was deeply evaluated, every key will come with a value.
 * However, shallowly evaluated records may have fields with no value.
 *
 * Returns 1 if the key came with a value, and 0 if it didn't. The value
 * will be written to `out_expr` if it is non-NULL.
 *
 * # Panics
 *
 * Panics if `idx` is out of range.
 */
int nickel_record_key_value_by_index(const nickel_record *rec,
                                     uintptr_t idx,
                                     const char **out_key,
                                     uintptr_t *out_key_len,
                                     nickel_expr *out_expr);

/**
 * Look up a key in this record and return its value, if there is one.
 *
 * Returns 1 if the key has a value, and 0 if it didn't. The value is
 * written to `out_expr` if it is non-NULL.
 */
int nickel_record_value_by_name(const nickel_record *rec, const char *key, nickel_expr *out_expr);

/**
 * Allocates a new string.
 *
 * The lifecycle management of a string is much like that of an expression
 * (see `nickel_expr_alloc`). It gets allocated here, modified by various other
 * functions, and finally is freed by a call to `nickel_string_free`.
 */
nickel_string *nickel_string_alloc(void);

/**
 * Frees a string.
 */
void nickel_string_free(nickel_string *s);

/**
 * Retrieve the data inside a string.
 *
 * A pointer to the string contents, which are UTF-8 encoded, is written to
 * `data`. These contents are *not* nul-terminated, but their length (in bytes)
 * is written to `len`. The string contents will be invalidated when `s` is
 * freed or overwritten.
 */
void nickel_string_data(const nickel_string *s, const char **data, uintptr_t *len);

/**
 * Allocate space for a virtual machine.
 *
 * The virtual machine can be initialized by `nickel_context_eval_shallow`.
 */
nickel_virtual_machine *nickel_virtual_machine_alloc(void);

/**
 * Free a virtual machine.
 */
void nickel_virtual_machine_free(nickel_virtual_machine *vm);

/**
 * Evaluate an expression to weak head normal form (WHNF).
 *
 * This has no effect if the expression is already evaluated (see
 * [`nickel_expr_is_value`]).
 *
 * The result of this evaluation is a null, bool, number, string,
 * enum, record, or array. In case it's a record, array, or enum
 * variant, the payload (record values, array elements, or enum
 * payloads) will be left unevaluated.
 */
nickel_result nickel_virtual_machine_eval_shallow(nickel_virtual_machine *vm,
                                                  const nickel_expr *expr,
                                                  nickel_expr *out_expr,
                                                  nickel_error *out_error);

/**
 * Allocate a new `nickel_error`.
 */
nickel_error *nickel_error_alloc(void);

/**
 * Frees a `nickel_error`.
 */
void nickel_error_free(nickel_error *err);

/**
 * Write out an error as a user- or machine-readable diagnostic.
 *
 * - `err` must have been allocated by `nickel_error_alloc` and initialized by some failing
 *   function (like `nickel_context_eval_deep`).
 * - `write` is a callback function that will be invoked with UTF-8 encoded data.
 * - `write_payload` is optional extra data to pass to `write`
 * - `format` selects the error-rendering format.
 */
nickel_result nickel_error_display(const nickel_error *err,
                                   nickel_write_callback write,
                                   void *write_payload,
                                   nickel_error_format format);

#ifdef __cplusplus
}  // extern "C"
#endif  // __cplusplus

#endif  /* NICKEL_LANG_H */

```

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
+//! C bindings for the Nickel language
+//!
+//! TODO: example
```
yannham: Should this be fixed for this PR?

nickel/src/capi.rs
yannham: I wonder if cbindgen's `no-export` annotation on the wrapper types would work (instead of `#[repr(transparent)]`), thus avoiding the manual pointer/ref conversion fonctions. But otherwise I can see why you did this, and if `no-export` doesn't work, this looks like a sensible approach.

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+pub struct nickel_expr {}
+
+impl nickel_expr {
+    /// Convert a *nickel_expr to a &Expr.
+    ///
+    /// # Safety
+    ///
+    /// Assumes that `this` was originally a valid pointer to Expr.
+    unsafe fn as_rust(this: &*const Self) -> &Expr {
+        (*this as *const Expr).as_ref().unwrap()
```
yannham: How do we  ensure that `Expr` never moves? I'm not really sure `Pin` would help here, because it's mostly FFI code, but do you think we'd have to wrap all underlying ` nickel-lang` types in `Box` wrapper to ensure no FFI Rust function could ever lend a pointer to the C world that could be then invalidated?

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+    }
+
+    /// Convert a *mut nickel_expr to a &mut Expr.
+    ///
+    /// # Safety
+    ///
+    /// Assumes that `this` was originally a valid pointer to Expr.
+    unsafe fn as_rust_mut(this: &mut *mut Self) -> &mut Expr {
```
yannham, line 125: I feel like a bunch of those implementations could be factored as a macro. I know that some of them later are a bit different, so I'm not sure it'd would be worth it overall, but it does feel like mechanical boilerplate. What do you think?

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+pub unsafe extern "C" fn nickel_context_set_source_name(
+    mut ctx: *mut nickel_context,
+    name: *const c_char,
+) {
+    let name = String::from_utf8_lossy(CStr::from_ptr(name).to_bytes()).into_owned();
+    nickel_context::as_rust_mut(&mut ctx).name = Some(name);
+}
+
+/// Perform some sort of evaluation, and return the error appropriately.
+unsafe fn do_eval<F: FnOnce(&mut crate::Context, &str) -> Result<Expr, Error>>(
```
yannham: nitpick: when the trait constraints are large, I find it more readable to put them in a `where` clause.

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+) -> nickel_result {
+    do_eval(|ctx, src| ctx.eval_deep(src), ctx, src, out_expr, out_error)
+}
+
+/// Evaluate a Nickel program deeply.
+///
+/// This differs from [`nickel_context_eval_deep`] in that it ignores
+/// fields marked as `not_exported`.
+///
+/// - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
```
yannham:
```suggestion
/// - `src` is a null-terminated string containing UTF-8-encoded Nickel source.
```

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+        }
+    }
+}
+
+/// Evaluate a Nickel program deeply.
+///
+/// "Deeply" means that we recursively evaluate records and arrays. For
+/// an alternative, see [`nickel_context_eval_shallow`].
+///
+/// - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
```
yannham:
```suggestion
/// - `src` is a null-terminated string containing UTF-8-encoded Nickel source.
```

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+///
+/// The result of this evaluation is a null, bool, number, string,
+/// enum, record, or array. In case it's a record, array, or enum
+/// variant, the payload (record values, array elements, or enum
+/// payloads) will be left unevaluated.
+///
+/// Together with the expression, this returns a Nickel virtual machine that
+/// can be used to further evaluate unevaluated sub-expressions.
+///
+/// - `src` is a nul-terminated string containing UTF-8-encoded Nickel source.
```
yannham:
```suggestion
/// - `src` is a null-terminated string containing UTF-8-encoded Nickel source.
```

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+/// A Nickel array.
+///
+/// See [`nickel_expr_is_array`] and [`nickel_expr_as_array`].
+// This is not really an empty struct, it's just an opaque type to
+// hide the implementation. The C API only ever uses `*const Array`,
+// and they are secretly pointers to `nickel_lang_core::term::array::Array`.
+pub struct nickel_array {}
+
+impl nickel_array {
+    unsafe fn as_rust(this: &*const Self) -> Array<'_> {
```
yannham, line 144: Why do  those `as_rust` and `as_rust_mut` need to take a double pointer (a reference or a mutable reference to a pointer), as opposed to the a raw pointer like `*const Self` directly?

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+/// nickel_record_value_by_name(rec, "foo", field);
+///
+/// // Now `rec` points to data owned by `expr`, but `field`
+/// // owns its own data. The following deallocation invalidates
+/// // `rec`, but not `field`.
+/// nickel_expr_free(expr);
+/// printf("number: %d\n", nickel_expr_is_number(field));
+/// ```
+#[no_mangle]
+pub unsafe extern "C" fn nickel_expr_alloc() -> *mut nickel_expr {
```
yannham, line 563: Ok, so maybe my pinning question is answered somehow. It's fine if we never ever return opaque pointers, but only fill them by destination, which forces the caller to allocate them, and we use `Box` which guarantees that the content doesn't move. Is that more or less correct? If yes, maybe that invariant should be made clear somewhere in the doc, so that future dev would  know "never ever return `nickel-lang` types to C that weren't provided by the C world, or derived from something provided in the C world".

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn nickel_expr_is_null(expr: *const nickel_expr) -> c_int {
+    nickel_expr::as_rust(&expr).is_null() as c_int
+}
+
+/// If this expression is a boolean, returns that boolean.
+///
+/// # Panics
+///
+/// Panics if `expr` is not a boolean.
```
yannham, line 643: I wonder if panicking in a FFI is a good idea. Not so long ago, it was actually UB (under the unwinding default behavior), because you could start to unwind a stack with C stuff on it, which is not supported. My understanding is that it's bit changed in a bit of a ad-hoc way just for `extern "C"`, so this should be fine. Still, I guess it's a bit dangerous, as an interface. Do you think it would be possible to use idiomatic `C` mechanisms to signal `None` basically, like returning `-1` or other impossible values for booleans, or setting `out` to `NULL` for more compound expressions?

On the other hand, you could argue that there are other functions to check for the error case...

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+    }
+}
+
+/// A Nickel string.
+// It would be nice to put `repr(transparent)` here, but (1) we don't
+// actually need to cast `String` pointers to `nickel_string` pointers,
+// and (2) adding `repr(transparent)` makes cbindgen expose `String`
+// even though it's private.
+pub struct nickel_string {
+    inner: String,
```
yannham, line 194: by the way, do you know why you don't have the "expose inner" issue with `String`?

nickel/src/capi.rs
```diff
@@ -0,0 +1,1016 @@
...
+
+use nickel_lang_core::term::Term;
+
+use crate::{Array, Context, Error, ErrorFormat, Expr, Number, Record, Trace, VirtualMachine};
+
+/// The main entry point.
+pub struct nickel_context {}
+
+impl nickel_context {
+    unsafe fn as_rust_mut(this: &mut *mut Self) -> &mut Context {
```
yannham, line 86: Even if it's a bit of copy-pasting and sounds trivial, I think any non-`extern "C"` unsafe function should be documented with the safety pre-conditions, as we would do for any `unsafe` block, whether it is exposed or not.

yannham (commented):
I left a bunch of comments, but they are more open questions than requests, so feel free to challenge them.

nickel/src/capi.rs
yannham: (NB: you should probably not answer this comment, there is an updated one later that revisits the issue)

nickel/src/capi.rs
yannham: Argh, I wrote a long follow-up to this question but  it was lost somehow. I understand that it's to tie the lifetime of the return `Array` to _something_. I wondered if that was really useful: it's better than nothing, but it doesn't really prevent any UB, because the lifetime of the pointer and of the pointee are entirely uncorrelated. And it could give a sense of false safety. I wondered if there were particular cases, that we would like to avoid, and that this API could prevent. Otherwise, I wonder if we shouldn't just bite the bullet, as we're in unsafe anyway, and return `forall a. Array<'a>` basically, making the API of this method simpler.

nickel/src/capi.rs
jneem: I'm not totally sure this is a good idea, but returning sentinal values has a few issues, I think. For booleans specifically, my C knowledge is a bit out-of-date (apparently `_Bool` and `bool` have been around for a while, but I don't think they were widely used back in the day...). But when using the old-style ints as bools, it's traditional to treat any non-zero value as true. And apparently `_Bool` only ever represents 0 or 1.

There's also some annoyance with the reference-based types, like `nickel_expr_as_record`. Because it returns a pointer in the success case, any non-pointer sentinal that it returns in the failure case could lead to UB if they aren't using the API properly. So for those reference types, I think an instant panic is better than potentially delayed UB. And then I figured it made sense to use the same pattern for everything.

Another option (which I think is a reasonably common way of encoding an optional return type) is something like

```rust
nickel_result nickel_expr_as_bool(nickel_expr *expr, bool *out_bool);
```

But I don't think it's much easier to use than checking in advance with `nickel_expr_is_bool`. And I think that checking followed by casting is common enough in C (with with lacking tagged unions, for example), so I thought that it would be a reasonable pattern.

I guess what I'm trying to say is that C's typesystem is such that it's impossible to have an API for dynamic type casting that makes both panics and UB impossible, so it's mostly a question of which API is less suprising...

nickel/src/capi.rs
jneem: I think cbindgen only leaks members that are `pub`, or that have `repr(transparent)`.

nickel/src/capi.rs
jneem: I agree it doesn't add much safety, but the pointer is basically always a local variable and so `as_rust` generates an `Array` with a short lifetime. Basically, `as_rust` makes it convenient to write a function that takes a `arr: *const nickel_array` and returns something with a lifetime *not* derived from `arr`.

jneem:
For some reason, github's UI won't let me respond directly to this comment:

> I wonder if cbindgen's no-export annotation on the wrapper types would work (instead of #[repr(transparent)]), thus avoiding the manual pointer/ref conversion fonctions. But otherwise I can see why you did this, and if no-export doesn't work, this looks like a sensible approach.

I tried adding

```rust
// (in lib.rs)
/// cbindgen:no-export
pub struct Expr { ... }

// (in capi.rs)
#[repr(transparent)]
pub struct nickel_expr { inner: Expr }
```

The result was that the generated header no longer defined a `struct Expr`, but it did still have the `typedef Expr nickel_expr`, line (along with a warning that it couldn't find the `Expr` type). I also tried setting `export.exclude` in `cbindgen.toml`, but to no effect.

nickel/src/capi.rs
jneem: I had a go at this, but it only worked for nickel_expr and nickel_context. The other empty-struct opaque types (nickel_array, nickel_record, nickel_number) were different enough that I couldn't easily cover them with the macro.

nickel/src/capi.rs
yannham: Hmm, I can see your point that injecting `null` in client code might be more dangerous than panicking. I still think that returning `null` is kinda idiomatic in C (ignoring the question of booleans). But your point about panic vs UB is indeed convincing (or we would have to protect against `null` in the API for all inputs).

nickel/src/capi.rs
yannham: Sorry, I think I'm probably just missing something. But why again can't we use that to implement opaque pointers to e.g. `Expr`? What would be wrong with:

```
// (in capi.rs)
pub struct nickel_expr { inner: Expr }
```

jneem: Ok, for the owned types (expr and context), I think you're right. I'll change those.

For the borrowed types (record, array, number), I still don't see how to avoid the empty struct: you want `*const nickel_array` to be convertible to `&nickel_lang_core::term::array::Array`, and the only two ways I know to do that are `repr(transparent)` or pointer casting to some placeholder type.

yannham: Is the lifetime the issue? I guess you could do `struct nickel_array { inner: Array<'static> }` (in some sense, we do stop tracking the lifetime of the inner `Array` as soon as it crosses the FFI boundary). It would require a transmute at allocation time, but no (unsafe) cast at usage. Honestly, I'm not entirely sure it's worth the trouble - it's just an idea that came up but I'll let be the judge of what's best.

yannham (approved):
Besides the `inner` question for borrowed types, I think you answered all the design questions, thanks! I'm excited about the C FFI - I also wonder if at some point we can scrap the whole Py03/maturin machinery and have a simple binding layer over the C API. More importantly, we can start to look at bindings to other languages.

nickel/src/capi.rs
jneem: It isn't really the lifetime, it's that I wanted the type on the C side to be a pointer: `nickel_expr_as_array` should return `*const nickel_array` and not `nickel_array`. I think this way it's clearer that it's a reference, as otherwise it might be surprising that freeing the `nickel_expr` could invalidate a `nickel_array`.

So now since `Array<'_>` is already effectively a pointer, I don't want to return a pointer to that pointer, as that would require more allocation and lifetime management. I really just want to cast the `Array<'_>` to some opaque pointer type and return that.

jneem:
Is PyO3/maturin that painful? I was thinking of rewriting (and extending) py-nickel using the public rust API, but I hadn't thought of going through C...

yannham:
It's quite easy to use, but I have strong suspicions (but unsubstantiated to this day) that this blows up the build, and causes undue rebuild and/or Nix cache misses, for example. I often see my rust-analyzer looping over PyO3-related buid steps upon small modifications. So it's more about the tooling. But the usage, on the other hand, is quite nice.

nickel/src/capi.rs
yannham: Ah, I see now. Thanks for clarifying.

nickel/src/capi.rs
yannham: Documentation has been added.

PR #2361, opened 2025-09-24
[RFC007] Refactor: delay the AST to RichTerm conversion

yannham:


core/src/cache.rs
```diff
@@ -1266,10 +1346,103 @@ impl CacheHub {
...
+    /// the AST cache, or [CacheError::NotParsed] is returned. However, for non-Nickel
+    /// dependencies, they are instead parsed directly into the term cache,
+    ///
+    /// "Compile" is anticipating a bit on RFC007, although it is a lowering of the AST
+    /// representation to the runtime representation.
+    ///
+    /// Compilation doesn't have a proper state associated, and thus should always be coupled with
+    /// program transformations through [Self::compile_and_transform]. It should preferably not be
+    /// observable as an atomic transition, although as far as I can tell, this shouldn't cause
+    /// major troubles to do so.
```
yannham, line 1374: Note: I might revise the absence of a state. Initially `compile` wasn't public, and was just a helper of `compile_and_transform`, but in the end it needed to be made public, so maybe it should have a proper state associated. On the other hand, the default state for `TermEntryState`, which is currently `Populated`, is maybe sufficient - or the mere presence of a term in the term cache.

core/src/program.rs
```diff
@@ -843,6 +848,23 @@ impl<EC: EvalCache> Program<EC> {
         Ok(())
     }
 
+    /// Parse and compile the stdlid and the program to the runtime representation. This is usually
```
jneem:
```suggestion
    /// Parse and compile the stdlib and the program to the runtime representation. This is usually
```

core/src/cache.rs
```diff
@@ -821,32 +835,66 @@ impl CacheHub {
...
+        if self.terms.contains(file_id) {
+            return Ok(CacheOp::Cached(()));
+        }
+
+        let term = if let InputFormat::Nickel = format {
+            match self.compile(file_id) {
+                Ok(cache_op) => return Ok(cache_op),
+                Err(_) => {
+                    let alloc = AstAlloc::new();
+                    self.sources.parse_nickel(&alloc, file_id)?.to_mainline()
```
jneem, line 885: Why fall back here if `compile` fails?

core/src/cache.rs
```diff
@@ -1016,15 +1094,17 @@ impl CacheHub {
         asts.typecheck_stdlib(slice)
     }
 
-    /// Loads, parses, and applies program transformations to the standard library. We don't
-    /// typecheck for performance reasons: this is done in the test suite.
+    /// Loads, parses, and compile the standard library. We don't typecheck for performance
```
jneem:
```suggestion
    /// Loads, parses, and compiles the standard library. We don't typecheck for performance
```

core/src/cache.rs
```diff
@@ -1266,10 +1345,103 @@ impl CacheHub {
         asts.add_type_bindings(slice, term)
     }
 
-    /// Applies both import resolution and other program transformations on the given term.
-    fn apply_all_transforms(&mut self, file_id: FileId) -> Result<CacheOp<()>, CacheError<Error>> {
+    /// Converts an AST and all of its transitive dependencies to the runtime representation,
+    /// populating the term cache. `file_id` and any of its Nickel dependenncies must be present in
```
jneem:
```suggestion
    /// populating the term cache. `file_id` and any of its Nickel dependencies must be present in
```

jneem (approved):

core/src/cache.rs
yannham: It's just the semantics of this method, I guess. It can be useful if you don't want to perform typechecking and just parse something directly to a term, even if it's Nickel (while `compile` fails if there is no entry in the AST cache). It's used e.g. in doctest, to prepare the samples. The rationale is that `compile` is to be used in the normal progression, as one step out of many others, while `parse_to_term` is something that can be used in a stand-alone fashion.

PR #2364, opened 2025-09-24
Declare MSRV as 1.85

brandonweeks:
MSRV was validated with `cargo-msrv`.


jneem (approved):
This sounds good to me, but with the caveat that we currently only aim for stability with the CLI interface: we don't aim to keep the rust interface stable, and we're pretty happy to bump MSRV whenever we feel like it.

This will probably be changing in the near-ish future with the `nickel-lang` crate (which seems to be the only one not touched by this PR...), at which point we might start doing things like checking MSRV in CI.

yannham (approved):
Thanks!

PR #2365, opened 2025-09-25
Fix building py-nickel on macOS

brandonweeks:
https://pyo3.rs/v0.26.0/building-and-distribution.html?highlight=darwin#macos


jneem:
Hm, that link suggests the those arguments should already be provided by maturin, which is what we're using in CI to build py-nickel. Does that not work?

brandonweeks:
Oh I was reading the outdated version of the documentation linked off of Google (https://pyo3.rs/v0.4.1/) that didn't have this language. 🤦 

Applogies!

PR #2366, opened 2025-09-25
Narrow serde_with features

brandonweeks:
The serde_with macros never appear to be used with the types from the supported crates.

https://docs.rs/serde_with/3.14.1/serde_with/guide/feature_flags/index.html#features


yannham (approved):
It does clean up the Cargo.lock file a bit. Thank you.

PR #2373, opened 2025-10-01
Restore the static nix release jobs

jneem:
I finally figured out how to statically link nickel in nix without pulling in an entire musl stdenv (which was causing binary cache misses and ridiculous build times). The cost is that the static build doesn't work with `--features=nix-experimental`. I don't think this is actually a regression, since we didn't expose that package in the flake anyway. I checked that the non-static build with `--features=nix-experimental` still works.

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -14,95 +14,70 @@ permissions:
 
 jobs:
   release-artifacts:
-    name: "Build Nickel binary"
+    name: "Build Nickel binary and Docker image"
     strategy:
       matrix:
         os:
           - runs-on: ubuntu-24.04
             architecture: x86_64
-            rust_architecture: x86_64
           - runs-on: ubuntu-24.04-arm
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -329,25 +329,25 @@
           # libc and clang with libc++ to build C and C++ dependencies. We
           # tried building with libstdc++ but without success.
           buildStaticWorkspace = { pnameSuffix, extraNickelFeatures ? [ ], extraBuildArgs ? "", extraArgs ? { } }:
+            let
+              env_tgt = builtins.replaceStrings [ "-" ] [ "_" ] pkgs.pkgsMusl.stdenv.hostPlatform.rust.rustcTarget;
```
yannham, line 338: Why do we need to do that? Maybe a small comment could help future reading

yannham (approved):

flake.nix
jneem: It's the env variable that the cc crate looks for. I've added a comment.

PR #2378, opened 2025-10-06
Add serde compatiblity to the public API

jneem:
I looked into *moving* the whole `deserialize` module into `nickel-lang`, but that would be a bit more disruptive. This was easy, and doesn't break anyone that's currently using `deserialize` in `nickel-lang-core`


jneem:
That was exactly my reasoning :smile:

PR #2379, opened 2025-10-08
Publish to PyPI if the action was triggered by a release

### LINKED ISSUE #2374: Nickel 1.13 (py-nickel) isn't publised on PyPi
**Describe the bug**
PyPi https://pypi.org/project/py-nickel/ is still on 1.10

**To Reproduce**
-

**Expected behavior**
PyPi has the latest 1.13 :)

jneem:
This might fix #2374. We'll have to check after the next release.


jneem:
Because `dry_run` is only an input if it's triggered by `workflow_dispatch`

PR #2380, opened 2025-10-09
Fix failure of `Program::new_from_inputs` with non-path sources

### LINKED ISSUE #2362: Program::new_from_inputs() dysfunctional on non-path sources
```rust
fn main() -> Result<()> {
    let source = Record::new().build().pretty_print_cap(1000);
    let sink = Sink::default();
    let mut prog: Program<CBNCache> = Program::new_from_inputs(
        [Input::Source(source.as_bytes(), &String::from("asdf"))],
        std::io::stderr(),
        sink,
    )?;
    let res = prog
        .eval_full_for_export()
        .map_err(|e| Error::msg(report_as_str(&mut prog.files(), e, ColorOpt::Auto)))?;
    println!("{}", res.pretty_print_cap(1000));
    Ok(())
}
```

results in:

```
Error: error: import of asdf failed: could not find import (looked in [])
  ┌─ <generated main>:1:1
  │
1 │ import "asdf" as 'Nickel
  │ ------------------------ imported here
```

whereas  using `new_from_input()` (singular,) successfully evaluates to an empty record.

---

The difference between the two is the generated import statement, and things don't change if there's more than one input passed to the plural version, Looking at the code the source does seem to be added to the eval cache, maybe the issue is in import resolution? `Path` inputs work fine.

---

That all aside, I'm kinda reporting an XY bug here, what I *actually* want to do is construct a program from a user-specified file and a generated `RichTerm`, to parametrise the config with locally available information. Overrides aren't really flexible enough, from all I can see they only support stings, and contracts don't get merged but applied. Not really interested in adding primitive functions, extending the stdlib or something like that though I guess some people might be, it's just about sending in plain structured data.

yannham:
Closes #2362. Applies the hotfix suggestion. I finally went with the ad-hoc prefix suggestion, because:

1. It preserves source names in error message (instead of non-sensical normalized paths derived from those sources), as opposed to the solution of pre-normalizing source names
2. It reduces the risk of shadowing/conflicts, as opposed to the solution of falling back to searching the non-normalized path if we don't find the normalized one. For example, with the latter solution, any file named `builtin` in the current  working  directory could interfere (in both directions) with Topiary's builtin config, which is injected as a source named "builtin". It's not a very rare file name. On the other hand, it's virtually impossible to have a file named `%inmem_src%:builtin` by accident.
3. It's still a very small patch/solution.

All in all, it's  not too dissimilar to the idea of having proper non-file imports like `import <builtin>`, but encode it as an ugly and undocumented prefix to avoid making it public or stable.

During debugging I needed a `Display` instance for `InputFormat`, which I decided to keep afterwards because it's useful elsewhere, even if it's in the end unrelated to this PR.

I intend to release a minor version of `nickel-lang-core` afterwards to unblock Topiary's release.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -289,9 +289,11 @@ impl TermCache {
     }
 }
 
-/// This is a temporary fix for https://github.com/tweag/nickel/issues/2362. File paths prefixed
-/// with this are treated specially: they can refer to in-memory source.
-pub(crate) const IN_MEMORY_SOURCE_PATH_PREFIX: &str = "%inmem_src%:";
+/// This is a temporary fix for [#2362](https://github.com/tweag/nickel/issues/2362). File paths
+/// prefixed with this are treated specially: they can refer to in-memory source. To build an
+/// import expression that refers to an in-memory source, append the source name to this prefix and
+/// use it as the path: `format!({IN_MEMORY_SOURCE_PATH_PREFIX}{src_name})`.
+pub const IN_MEMORY_SOURCE_PATH_PREFIX: &str = "%inmem_src%:";
 
 /// The source cache handles reading textual data from the file system or other souces and storing
 /// it in a [Files] instance.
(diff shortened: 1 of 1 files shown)


core/src/cache.rs
```diff
@@ -378,13 +390,43 @@ impl SourceCache {
...
+    ///
+    /// As a temporary fix for [#2362](https://github.com/tweag/nickel/issues/2362), if a file path
+    /// starts with [IN_MEMORY_SOURCE_PATH_PREFIX], the suffix is looked up un-normalized value
+    /// first, which makes it possible to hit in-memory only sources by importing a path
+    /// `"{SOURCE_PATH_PREFX}{src_name}"`. If it can't be found, it is looked up normally, so that
+    /// it doesn't break strange file names that happen to contain the source path prefix.
+    ///
+    /// It is theoretically possible that if both the source "abc" and the file
+    /// "{IN_MEMORY_SOURCE_PATH_PREFIX}abc" exist, the source is imported instead of the intended
+    /// file. However, given the prefix, it just can't be accidental. As we want to give access to
```
ksf, line 404: Might want to add a cheeky link to the definition of [portable POSIX filenames](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282), which outlaw `%` and `:`.

core/src/cache.rs
```diff
@@ -378,13 +390,43 @@ impl SourceCache {
...
+    /// in-memory sources in any case, although this can be surprising, I don't see any obvious
+    /// attack scenario here. This fix is also intended to be temporary. If you still need to make
```
ksf, line 406: There might be a situation where it's desirable to hide internal sources from users, or prevent users from overriding internal sources, especially when writing something like nix (which might run suid).

OTOH meh, it's a quick fix, and in the end it's likely a better idea to sandbox the whole runtime when you're doing that kind of thing, anyway. It's not like current nix gets suid to root, it gets suid to nixbld with severely lowered permissions so the worst possible outcome is GIGO.

yannham: Right, and here the scenario is that you might get a source instead of a file, but not the converse: users can't override a source, because if it has the prefix, it's first tried as a source.

I do dislike this shadowing thing, and feel like it could, one day, in a specific situation, either by a footgun or even a vulnerability, but I came to the conclusion that 1. it was already the case before the path norm PR 2. it's temporary (tm) 3. the attack isn't clear at all at first. So it _should_ be ok as a quick fix. But it must not stay this way.

ksf: If the implementation has a source `<secrets>` that's used by `<api>` then being able to access `<secrets>` directly might allow a user to exfiltrate data, is all I can think of right now from that attack angle. But then noone who knows how to protect secrets would expect a random language runtime to provide exfiltration safety in the first place so it's a non-issue.

core/src/program.rs
```diff
@@ -337,13 +337,19 @@ impl<EC: EvalCache> Program<EC> {
...
                 }
                 Input::Source(source, name) => {
-                    let path = PathBuf::from(name.into());
+                    let name = name.into();
+                    let mut import_path = OsString::new();
+                    // See https://github.com/tweag/nickel/issues/2362 and the documentation of
+                    // IN_MEMORY_SOURCE_PATH_PREFIX
+                    import_path.push(IN_MEMORY_SOURCE_PATH_PREFIX);
+                    import_path.push(name.clone());
+
```
blindFS, line 346: Maybe it's by design, yet I'd like to remind that there's a similar `new_from_input` pub function that may need this change as well.

PR #2381, opened 2025-10-09
Refactor: split external state from VM

yannham:
The virtual machine used to own state that is likely to outlive it. This makes interfaces awkward, as in the current implementation of e.g. Program, we need to go through the VM to do a lot of cache operations just because the VM requires ownership of the cache, while those operations have nothing to do with the VM. It also requires to keep the VM alive as soon as we put the cache inside, as long as we need the cache.

This is even worse in the context of #2302, as we need to add the position table to this state.

This refactor split this state as a separate context, that the VM borrows from. The cost is to add a lifetime parameter to `VirtualMachine`, but in return, it makes it much simpler to spin local instances of a VM, instead of threading one VM through many operation mostly just because we need the cache to continue to exist.

This commit changes a bit the way the stack is unwinded (`VirtualMachine::reset`). Since we spin a lot more local/temporary instances, I want to make sure the stack is properly unwinded, which is easy to forget (by dropping the VM silently), leaving thunks in blackholed state upon errors. Instead of reseting a VM before the next eval, this PR incorporates stack unwinding as part of a `Drop` implementation for the VM, guaranteeing to leave thunks in a consistent state. The drawback is that we could unwind more than previously (say, upon a single evaluation that aborts). However, in the happy path (success), the stack is empty, so the unwinding is virtually free, which sounds acceptable overall. The PR also includes a non-unwinding wrapper based on `ManuallyDrop`, that is used e.g. in tests.

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -122,7 +122,7 @@ impl AsRef<Vec<StackElem>> for CallStack {
     }
 }
 
-/// The context of the Nickel virtual machine. The context stores external state might need to
+/// The context of the Nickel virtual machine. The context stores external state that might need to
 /// outlive one VM instance. The virtual machine typically borrows the context mutably.
 pub struct VmContext<R: ImportResolver, C: Cache> {
     /// The interface used to resolve and fetch imports.


core/src/eval/mod.rs
```diff
@@ -118,88 +122,130 @@ impl AsRef<Vec<StackElem>> for CallStack {
...
-    call_stack: CallStack,
-    // The interface used to fetch imports.
-    import_resolver: R,
-    // The evaluation cache.
-    pub cache: C,
-    // The initial environment containing stdlib and builtin functions accessible from anywhere
-    initial_env: Environment,
-    // The stream for writing trace output.
-    trace: Box<dyn Write>,
+/// The context of the Nickel virtual machine. The context stores external state might need to
```
jneem:
```suggestion
/// The context of the Nickel virtual machine. The context stores external state that might need to
```

core/src/repl/mod.rs
```diff
@@ -86,73 +87,81 @@ pub trait Repl {
 
 /// Standard implementation of the REPL backend.
 pub struct ReplImpl<EC: EvalCache> {
+    /// The virtual machine context.
+    vm_ctxt: VmContext<CacheHub, EC>,
     /// The current eval environment, including the stdlib and top-level lets.
     eval_env: eval::Environment,
-    /// The state of the Nickel virtual machine, holding a cache of loaded files and parsed terms.
-    vm: VirtualMachine<CacheHub, EC>,
+    phantom: PhantomData<EC>,
```
jneem, line 94: Why do we need this one? EC is already a parameter of VmContext...

yannham: Thanks, it's probably a left-over. I'll get rid of it in #2302 directly (because rebasing has been painful already :sweat_smile: ).

PR #2382, opened 2025-10-10
Add locations to json

jneem:
It occurred to me as I was doing this that we don't want to *just* parse JSON as YAML, because that would be too permissive. So I'm using serde_json to first make sure it's valid json, and only then parsing using saphyr.

This is quite a hack, but at least the performance doesn't seem too bad: the serialization benchmark in nickel-lang-core doesn't seem to change.

As I was looking into better solutions, I noticed that facet-json appears to have positioning support, so that might be worth looking into more. There's also a facet-yaml (with a parser based on yaml-rust2, by the same authors as saphyr) and a facet-toml, so we'd be able to use a similar interface for all three...

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -133,5 +133,9 @@ harness = false
 name = "typecheck-nixpkgs-lib"
 harness = false
 
+[[bench]]
+name = "deserialization"
+harness = false
+
 [lints.clippy]
 new_without_default = "allow"
--- a/core/benches/deserialization.rs
+++ b/core/benches/deserialization.rs
@@ -1,0 +1,35 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use nickel_lang_core::{bytecode::ast::AstAlloc, serialize::yaml::load_json};
+
+pub fn json_ast_construction(c: &mut Criterion) {
+    fn generate_json(depth: usize, branching: usize) -> serde_json::Value {
+        if depth == 0 {
+            serde_json::Value::String("hi".into())
+        } else {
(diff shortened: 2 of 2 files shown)


jneem:
I looked into the performance a bit more. I added a benchmark (in this PR now) that tests parsing and loading to `Ast`, using documents with 10k and 100k nodes.

The current state of this PR takes 104ms and 1.08s for the two documents.

Switching to oxidized-json-checker reduces this to 89ms or 1.05s, but comes at the cost of losing error locations.

Removing JSON validation altogether reduces this to 82ms or 0.98s.

I'm inclined to say the speed improvement isn't worth the extra dependency and loss of error locations, what do you think?

I took a further look at saphyr, and the parser doesn't publicly expose anything that lets you reject invalid json. One nice solution (and maybe not even too difficult) would be to write a JSON parser that uses saphyr's `SpannedEventReceiver` interface. Then it would be very easy to switch between json and yaml...

yannham:
> I'm inclined to say the speed improvement isn't worth the extra dependency and loss of error locations, what do you think?

I tend to agree, in particular since we bench only the parsing and ast allocation part, which is in reality probably a small part of the execution of an importing program. Given the really small difference,do you have any idea where most of the time is spent? Allocating the AST? Or is Saphyr really much slower than a JSON parser?

jneem:
> Or is Saphyr really much slower than a JSON parser?

It seems that way, yes. In this benchmark, saphyr -> AST is about 6-7x slower than serde_json -> serde_json::Value. I'm not sure how much of that is the inherent complexity of YAML and how much is saphyr not being as optimized.

PR #2383, opened 2025-10-10
Allow nls to report contract errors in yaml

jneem:
This adds support for `Nls-contracts.ncl` files, which are read by nls to determine whether plain data files should have Nickel contracts applied to them. If so, nls reads and applies the contract, and reports diagnostics.

The configuration file (Nls-contracts.ncl) looks like

```nickel
[
 { glob = "**/*-foo.yaml", contract_path = "foo.ncl" },
 { glob =  "**/*-bar.yaml", contract_path = "bar.ncl" }
]
```

which tells nls that any file matching the glob "**/*-foo.yaml" (relative to the location of the `Nls-contracts.ncl` file) will have the contract "foo.ncl" (also relative to the location of the `Nls-contracts.ncl` file) applied.

This PR also starts the work of saving an analysis of the yaml file, which could eventually help support things like hover queries. That part isn't done yet, though.


lsp/nls/src/analysis.rs
```diff
@@ -367,14 +367,14 @@ pub(crate) struct AltFormatErrors {
     reason = "the large variant is the common case"
 )]
 pub(crate) enum AnalysisTarget<'std> {
-    Nickel(PackedAnalysis<'std>),
+    Analysis(PackedAnalysis<'std>),
```
yannham, line 372: We should probably update the doc above, and add a documentation for this filed as well, to explain what it means. `AnalysTarget` is `Analysis` might be a bit confusing for an external reader.

lsp/nls/src/codespan_lsp.rs
```diff
@@ -13,15 +13,13 @@ use std::ops::Range;
...
 ) -> Result<LspPosition, Error> {
     if column > line_str.len() {
-        let max = line_str.len();
-        let given = column;
-
-        Err(Error::ColumnTooLarge { given, max })
-    } else if !line_str.is_char_boundary(column) {
+        column = line_str.len();
+    }
+    if !line_str.is_char_boundary(column) {
```
yannham, line 24: Why was this code changed? Did it have to be adapted for YAML?

lsp/nls/src/contracts.rs
```diff
@@ -0,0 +1,290 @@
...
+        ContractConfig {
+            contract_path: parent_path.join(self.contract_path),
+        }
+    }
+}
+
+/// A collection of file-local nls configurations, provided by a single
+/// config file.
+///
+/// When we find a config file (called `nls.ncl`) in the filesystem,
```
yannham: Should `nls.ncl` be updated here? Or is it a separate config file, in addition to the contract configuration?

lsp/nls/src/contracts.rs
```diff
@@ -0,0 +1,290 @@
...
+/// {
+///   "*o.ncl" = { eval_config.eval_limits.timeout = "10 seconds" }
+/// }
+/// ```
+///
+/// then the background eval job for `foo.ncl` will have an evalution
+/// timeout of 10 seconds.
+#[derive(Debug, Default, PartialEq)]
+pub struct LocalConfig {
+    globs: IndexMap<glob::Pattern, ContractConfig>,
```
yannham: I find it a bit strange to have an index map here. I think it ties to a more general question: why do we use a record of globs instead of a list, in the contract mapping Nickel file? Though on one hand, I can understand for constant file names like `foo/bar.ncl`. But do we really ever index into `glob` with a given glob?

lsp/nls/src/contracts.rs
```diff
@@ -0,0 +1,290 @@
...
+        local_config.into_local_config(parent_path)
+    }
+}
+
+pub struct LocalConfigs {
+    configs: HashMap<PathBuf, LocalConfig>,
+}
+
+/// A self-updating collection of [`LocalConfig`]s.
+pub struct ContractConfigsWatcher {
```
yannham, line 94: Is the goal to be able to change the configuration dynamically without having to restart the LSP? Now that it's here, why not, but I would assume it's not unexpected to have to reload the LSP when you change the config, in general.

lsp/nls/src/world.rs
```diff
@@ -216,18 +237,43 @@ impl World {
...
-                let analysis = PackedAnalysis::parsed(
-                    file_id,
-                    &self.sources,
-                    init_term_env.clone(),
-                    init_type_ctxt.clone(),
-                );
-                errs = analysis.parse_errors().clone();
-                analysis
-            });
+        let errs = match self.file_format(file_id).unwrap_or_default() {
```
yannham: `file_format` is failing if we can't find the source id in the cache. I find it a bit strange to use the default value (`InputFormat::Nickel`) when that happens. Shouldn't we throw an error, panic, return empty diagnostics or something like that?

lsp/nls/src/world.rs
```diff
@@ -172,9 +194,8 @@ impl World {
     pub fn close_file(&mut self, uri: Url) -> anyhow::Result<(Option<FileId>, Vec<FileId>)> {
         let path = uri_to_path(&uri)?;
 
-        let result = self
-            .sources
-            .close_in_memory_file(path, InputFormat::Nickel)?;
+        let format = InputFormat::from_path(&path).unwrap_or_default();
```
yannham, line 197: (Same remark as below about `unwrap_or_default`).

yannham (commented):
It's been reflected in comments already, but do you think it makes more sense to have a record of glob-indexed values, or just an array? It doesn't sound like globs as keys make a lot of sense (on the other hand, it's more amenable to merging, so I'm not sure).

Additionally, it could be nice to test the new functionality, with a YAML file failing a contract.

lsp/nls/src/analysis.rs
jneem: Sure, I updated the comment and also changed `AnalysisTarget` to `AnalysisResult`

lsp/nls/src/codespan_lsp.rs
jneem: It was actually because of the serde_json parse errors, which can sometimes point 1 past the end of the file. I've added a comment.

lsp/nls/src/contracts.rs
jneem: No, that was just left over from the last attempt...

lsp/nls/src/contracts.rs
jneem: Yes, I think you're right -- I'll try making it a list instead. I think I was just initially attracted by the fact that the `{ "*.yaml" = "c.ncl" }` syntax is a little lighter.

lsp/nls/src/contracts.rs
jneem: It isn't too unexpected, sure. But I've been trying out the zed editor a little, and it does a very nice job of autoloading things...

lsp/nls/src/world.rs
jneem: Sure, I've turned this one into a panic (since we panic in a bunch of places on invalid file ids)

lsp/nls/src/world.rs
jneem: I think this one is ok defaulting to Nickel, because there are several places where we default to nickel format if the path is ambiguous (like `Program::new_from_input`, for example)

yannham: Ah, sorry, you're right - this is just a `from_path` call, which has nothing to do with missing file ID.

yannham:
I think it's still missing tests, and there's still an `IndexMap` of globs, instead of a list. Is the latest force push missing some recent changes?

jneem:
Sorry, I was too eager on the "request re-review" button... I tried to remove it, but I guess the notification was already sent. The changes should be visible now, along with a test.

yannham (approved):

PR #2388, opened 2025-10-21
Use json_scanner for parsing json with positions

jneem:
Maybe not the *best* use of my time, but I was so annoyed by the state of json parsing with positions that I wrote my own (`json_scanner`, which I added to the `nickel-lang` org). It offers a fairly similar low-level interface as the saphyr parser, so we can reuse the saphyr loader, but with a json parser instead of a yaml parser.

This is a pretty significant performance win, with the 100k-entry example going from around 100ms to 17ms, and the 1M-entry example going from 1s to 310ms.

While testing, I noticed that we weren't handling char vs. byte indexing correctly in the saphyr loader, because saphyr reports char indices, not byte indices. So I fixed that too.


cli/tests/snapshot/snapshots/snapshot__eval_stderr_json_but_its_yaml.ncl.snap
```diff
@@ -2,11 +2,11 @@
 source: cli/tests/snapshot/main.rs
 expression: err
 ---
-error: json parse error: expected ident at line 1 column 2
-  ┌─ [IMPORTS_PATH]/map.yaml:1:2
+error: json parse error: expected `false`
```
yannham, line 5: What this error message is supposed to mean?

core/src/serialize/yaml.rs
```diff
@@ -182,14 +184,25 @@ struct YamlLoader<'ast> {
...
     /// doesn't provide a way for the loader to exit on error. We have a few
     /// error cases that aren't caught by the yaml parser (for example, because
     /// Nickel records only allow string keys). Because there's no way to
     /// early-exit, we keep track of those errors by stashing them here.
     err: Option<ParseError>,
+    /// Saphyr reports positions in char indices, so we need to convert
+    /// them back to bytes. Our little JSON shim, on the other hand, reports
+    /// positions in bytes directly.
+    ///
+    /// So, if this vector is non-empty the it's a mapping of char indices to
```
yannham:
```suggestion
    /// So, if this vector is non-empty then it's a mapping of char indices to
```

core/src/serialize/yaml.rs
```diff
@@ -486,11 +516,58 @@ pub fn load_json<'ast>(
...
     location: Option<(FileId, &Files)>,
 ) -> Result<Ast<'ast>, ParseError> {
-    // JSON is a subset of YAML (or at least, YAML 1.2, which is what saphyr implements).
-    // We can't *just* use saphyr to parse JSON because it would be overly permissive.
-    let _val: serde_json::Value =
-        serde_json::from_str(s).map_err(|err| ParseError::from_serde_json(err, location))?;
-    load(alloc, s, location.map(|x| x.0), "json")
+    let mut parser = json_scanner::Parser::new(s.as_bytes());
+    let file_id = location.map(|(id, _)| id);
+    let mut loader = YamlLoader {
```
yannham: Nitpick: maybe it shouldn't be named `YamlLoader` anymore

yannham (approved):
Well, I did find the previous situation unfortunate as well, and the results look good. We have control over `json_scanner` and it probably won't be too hard to switch back to the previous situation, if something goes wrong. We now have position information for all three major config format :tada:

cli/tests/snapshot/snapshots/snapshot__eval_stderr_json_but_its_yaml.ncl.snap
jneem: It expected to see the token `false` (because it saw an `f` in a position where the only valid follow-up would be `alse`)

PR #2391, opened 2025-10-22
Adds support for any_of, all_of and Sequence to nls

### LINKED ISSUE #2343: Make the LSP understand `any_of` and `all_of`
LSP auto-completion has some advanced rules, which makes it able to fetch information from the context, type and contract annotations, merge sequences, etc. to accurately try to provide a completion of the fields available on a record.

Unfortunately, it doesn't understand the recent addition of `std.contract.any_of` and `std.contract.one_of`. For example, in the following source file:

```nickel
let Foo = {foo | String} in
let Bar = {bar | Number} in

let f | std.contract.any_of [Foo, Bar] -> Dyn = fun x => x.[cursor] in
0
```

No completion is proposed for either `foo` or `bar`. Changing the argument contract of the function to either `Foo` or `Bar` alone does work again, providing the available field.

The LSP should provide completion information based on contracts built from `any_of` and `all_of`. For both, we should probably give the union of the available fields on each contract. Although it's an over-approximation for `all_of`, this sounds like a reasonable first step, and it's not clear how we could do better in the general case.

**Additional context**
Technically, we should try to cover cases like:

```
let {contract, ..} = std in contract.any_of [Foo, Bar] in
```

Maybe we could resolve the definition of the head of any function application appearing in a contract, and see if it points to the definition of `any_of`, to make it more robust.

jneem:
Fixes #2343

### DIFF
--- a/lsp/nls/src/field_walker.rs
+++ b/lsp/nls/src/field_walker.rs
@@ -720,6 +720,7 @@ impl<'ast> FieldResolver<'ast> {
             }
         }
 
+        // unwraps: the stdlib should have these things defined.
         let std = world.analysis_reg.stdlib_analysis().ast();
         let any_of = at_path(std, &[Ident::new("contract"), Ident::new("any_of")]).unwrap();
         let all_of = at_path(std, &[Ident::new("contract"), Ident::new("all_of")]).unwrap();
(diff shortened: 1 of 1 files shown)


lsp/nls/src/field_walker.rs
```diff
@@ -876,6 +1001,37 @@ impl<'ast> FieldResolver<'ast> {
...
+            Node::App { head, args } => {
+                let heads = self.resolve_term(head);
+                if heads
+                    .iter()
+                    .any(|&a| a == self.any_of || a == self.all_of || a == self.sequence)
+                {
+                    // `all_of`, `any_of` and `Sequence` expect a single arg, and that arg
+                    // should be an array (but it might not be a literal array, so
+                    // we try to resolve it first).
+                    // TODO: can try let chains after edition 2024
```
yannham, line 1013: FWIW, I've migrated to 2024 as part of #2302.

yannham (approved):

PR #2398, opened 2025-10-27
Bump serde_with to version 3.15.1

brandonweeks:
Eliminates an extra copy of darling from the dependency tree.


yannham:
Is it on purpose that no change to `Cargo.toml` was committed here? Is it a pure lockfile update?

PR #2401, opened 2025-10-28
Add a build step that shrinks the static archive

jneem:
This shrinks libnickel_lang.a from about 150MB to about 18MB. It's motivated by the go bindings, because go wants to have the compiled library in the git repo, and it likes static linking, but a 150MB archive was just too big.

### DIFF
--- a/scripts/shrink_archive.sh
+++ b/scripts/shrink_archive.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+
+# This script shrinks the libnickel_lang.a static library, which is otherwise
+# a shocking 150MB. We achieve this by unpacking the archive and re-linking
+# most of the object files into one big object, while removing unreachable code.
+#
+# One way to do this on Linux is with `ld -r --gc-sections`, but `--gc-sections`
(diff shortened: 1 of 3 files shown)


scripts/shrink_archive.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bash
+
+# This script shrinks the libnickel_lang.a static library, which is otherwise
+# a shocking 150MB. We achieve this by unpacking the archive and re-linking
+# most of the object files into one big object, while removing unreachable code.
+#
+# One way to do this on Linux is with `ld -r --gc-sections`, but `--gc-sections`
+# isn't supported by the MacOS linker, and the analogous `-dead_strip` doesn't
+# work with `-r`. So what we actually do is extract the llvm bitcode from the
+# object files that have it (which is most of them), link them into one bit
```
yannham:
```suggestion
# object files that have it (which is most of them), link them into one big
```

yannham (approved):
That's a very interesting script outside of Nickel, I believe :thinking: 

I'm not an expert of the LLVM CLI or pipeline, but from what I can say, it looks good to me.

PR #2404, opened 2025-10-29
Ignore closurization in Expr::is_value

jneem:
This fixes another bug in the public API, where `is_value` would return false on deeply-evaluated records.

### DIFF
--- a/nickel/src/lib.rs
+++ b/nickel/src/lib.rs
@@ -526,7 +526,7 @@ impl Expr {
     pub fn is_value(&self) -> bool {
         // The term here should always be closurized, because the only public way to
         // construct an Expr is via evaluation. Therefore we do not check is_eff_whnf,
-        // which doesn't do what we with with `eval_deep` because `subst` resets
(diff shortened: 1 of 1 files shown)


nickel/src/lib.rs
```diff
@@ -524,7 +524,11 @@ impl Expr {
     /// and you can use the [`VirtualMachine`] you got from `eval_shallow` to
     /// evaluate this expression further.
     pub fn is_value(&self) -> bool {
-        self.rt.as_ref().is_eff_whnf()
+        // The term here should always be closurized, because the only public way to
+        // construct an Expr is via evaluation. Therefore we do not check is_eff_whnf,
+        // which doesn't do what we with with `eval_deep` because `subst` resets
```
yannham:
```suggestion
        // which doesn't do what we want with with `eval_deep` because `subst` resets
```

yannham (approved):
This is also more in line with upcoming compact values, since `is_eff_whnf` doesn't even exist anymore (it doesn't make sense) on `NickelValue`.

PR #2412, opened 2025-11-06
Make decisions about the parser re-exports

jneem:
This just moves things around (and deletes a duplicated `combine` module).

A bunch of the parser crate is re-exported in `core`, mostly at the root (so, for example, we have `nickel_lang_core::ast` as a re-export of `nickel_lang_parser::ast`)

### DIFF
--- a/core/src/bytecode/ast/mod.rs
+++ b/core/src/bytecode/ast/mod.rs
@@ -1,1 +1,0 @@
-pub mod compat;
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -1,6 +1,8 @@
 use crate::{
-    ast::{Ast, AstAlloc},
-    bytecode::ast::compat::{FromAst, ToMainline},
+    ast::{
+        Ast, AstAlloc,
+        compat::{FromAst, ToMainline},
+    },
     eval::value::NickelValue,
     position::PosTable,
 };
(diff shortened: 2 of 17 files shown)


core/src/parser.rs
```diff
@@ -1,25 +1,25 @@
...
-        CliFieldAssignmentParser, ExtendedTermParser, FixedTypeParser, StaticFieldPathParser,
-        TermParser,
-    },
     identifier::LocIdent,
     metrics,
     position::RawSpan,
 };
 
-// TODO(parser migration): compatibility shim
 pub use nickel_lang_parser::{
```
yannham, line 18: I'm not entirely sure about this one. It sounds like it would be fair to use `nickel_lang_parser` directly instead of `nickel_lang_core::parser`. The only remaining code here is about compatibility, so maybe it should move to compat? But it's just, like, my opinion. In any case we can leave this for follow up work.

core/src/bytecode/ast/mod.rs
```diff
@@ -1,5 +1 @@
 pub mod compat;
```
yannham: While we're at it, we should get rid of the `bytecode` prefix for `ast`. Maybe put compat in its own module  `ast_compat`, or move it to `eval::value`, or move it to `ast` in addition to re-exporting `nickel_lang_parser::ast::*`, whatever; I find it confusing to have now `nickel_lang_core::ast` but `nickel_lang_core::bytecode::ast::compat`...

yannham (approved):

core/src/parser.rs
jneem: My main motivation here was to make it easier for dependent crates, so they don't need to take an extra dependency on `nickel_lang_parser` and then make sure the versions stay in sync.

PR #2417, opened 2025-11-07
Fix large error types in `core`

yannham:
This PR removes the `#[allow(clippy::result_large_err)]` out of jail free card from `nickel_lang_core`, and fix the warning by introducing the required intermediate error types and boxing everywhere.

To keep the PR reasonable, I haven't touched the same issue in other crates of the workspace (typically `cli` or `package`).

### DIFF
--- a/core/src/eval/tests.rs
+++ b/core/src/eval/tests.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use crate::{
     cache::resolvers::{DummyResolver, SimpleResolver},
-    error::{ImportErrorKind, NullReporter},
+    error::{ImportError, ImportErrorKind, NullReporter},
     files::Files,
     label::Label,
     parser::{ErrorTolerantParserCompat, grammar, lexer},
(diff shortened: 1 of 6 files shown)


core/src/term/mod.rs
```diff
@@ -255,7 +255,7 @@ pub enum Term {
     /// missing field definition error. Thus, we need to bind `bar` to a term wich, if ever
     /// evaluated, will raise a proper missing field definition error. This is precisely the
     /// behavior of `RuntimeError` behaves.
```
jneem:
```suggestion
    /// behavior of `RuntimeError`.
```

core/src/eval/tests.rs
```diff
@@ -188,7 +188,7 @@ fn imports() {
         var: &str,
         import: &str,
         body: NickelValue,
-    ) -> Result<NickelValue, ImportError>
+    ) -> Result<NickelValue, ImportErrorKind>
```
jneem: Why not box up this error variant (similar to EvalErrorKind)?

yannham: No good reason, it probably just went under the radar because it's in tests. Thanks!

yannham: In fact I suspect the actual reason is that `ImportErrorKind` is 64 bytes and rustc doesn't consider that big enough to fire the warning (since I followed the compile warnings to fix all the places). I still made the change though, 64 bytes is _something_...

PR #2420, opened 2025-11-09
Add note to error when arrays are merged piecewise in a record

L0r3m1p5um:
This adds some additional context to the error when merging arrays piecewise in a record, like this for example:
```nickel
{
  a = [1, 2],
  a = [2, 3],
}
```
As best as I can tell, there is not any reason anyone would intentionally do this, and if there are a lot of fields defined between the two instances of the redefined field, it can be hard to spot. The current error you would see isn't particularly helpful in this case, because if you don't already know that the field has been redefined in this way, it's hard to figure out why an array merge is even occurring. Here's what the output looks like in the above case. If arrays are merged for a different reason, like a normal record merge, this new note is not displayed.

```
$ nickel eval unmergeable_arrays.ncl
       cannot merge unequal arrays
  │
1 │ contract.Equal some_array
  │ ------------------------- expected type
  │
  ┌─ /nickel_test_files/unmergeable_arrays.ncl:3:10
  │
3 │   a = [2,3],
  │          - evaluated to this expression
  │
  = This equality contract was auto-generated from a merge operation on two arrays. Arrays can only be merged if they are equal.
  = The arrays were merged because they were assigned to the same record field piecewise. This is likely to have been a mistake. Check for duplicate definitions of the same record field.
```

I would have preferred to add a secondary label here pointing to the other location that the same field is defined, similar to how other unmergeable types do, but I didn't find a way to do it. Array merges aren't the same as those other cases since they just apply a contract, instead of explicitly returning an error when merging, and I don't think I can create a contract that would create secondary labels like that. As an alternative I checked if I could add an additional diagnostic note pointing to the other location where the same field was defined, and while I can get the RawSpan of that other location I couldn't figure out how to resolve it to a human readable line/column for display. With a Files instance it should be possible but I don't think I have access to it in that context.


L0r3m1p5um:
I've added the snapshot tests but the build is failing due to a random CI failure it looks like.

PR #2424, opened 2025-11-10
refactor: generate generic `ValueLens` impls with a macro

yannham:
Depends on #2423

Now that all term variants only have one argument, it's possible to factor out all the `ValueLens<_>` implementations in a macro, which shaves a off a few hundred lines of code.

### DIFF
--- a/core/src/eval/value/lens.rs
+++ b/core/src/eval/value/lens.rs
@@ -300,8 +300,11 @@ macro_rules! impl_term_lens {
             //
             // # Safety
             //
-            // `value` must be a value block with tag `DataTag::Term`, and the inner term must
-            // match `Term::$term_cons`.
+            // `value` must be a value block with tag `DataTag::Term`.
+            //
+            // # Panics
+            //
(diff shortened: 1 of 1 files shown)


core/src/eval/value/lens.rs
```diff
@@ -283,6 +283,50 @@ impl TermContent {
...
+/// the function documentation within the macro definition just below.
+macro_rules! impl_term_lens {
+    ( $lens_cons:ident, $lens_extrct:ident, $term_cons:ident, $type:ty ) => {
+        impl ValueLens<$type> {
+            // Creates a new lens extracting `Term::$term_cons`.
+            //
+            // # Safety
+            //
+            // `value` must be a value block with tag `DataTag::Term`, and the inner term must
+            // match `Term::$term_cons`.
```
jneem: The inner term part isn't actually a safety issue, right? Because `unreachable!` isn't unsafe, it's just a panic...

yannham: Good catch, will fix in upcoming committ

PR #2429, opened 2025-11-13
perf: do not use lenses in eval

yannham:
After some measurement, it turns out it's almost never the case that we can take advantage of 1-rc values in the main eval loop. Indeed, any value that ends up in the environment will be kept there as a copy, and so will never be 1-rc (we need to keep it in the environment). Consequently, if the root is cloned, no children will be 1-rc either, etc.. 

Lenses are still useful for program transformation, but in the main eval loop, it seems that it's mostly adding a small overhead. This PR goes back to taking everything by ref, and cloning on demand. Consequently, we also implement a version of `merge::split` that works on references, as it's mostly used that way in the eval loop.

Benches show a small but steady improvements in most benches, around a few percents, although it's within potential noise (criterion is supposed to filter noise already, but I still see a lot of variance from run to run). On the OPL benchmark though, we do seem to get a consistent improvement, still rather small (5%~ish).

### DIFF
--- a/core/benches/records.rs
+++ b/core/benches/records.rs
@@ -12,7 +12,7 @@ ncl_bench_group! {
         }, {
             name = "merge",
             path = "records/merge",
-            args = (500, 50),
+            args = (50, 10),
             eval_mode = EvalMode::DeepSeq,
     }
 }


core/benches/records.rs
```diff
@@ -12,7 +12,7 @@ ncl_bench_group! {
         }, {
             name = "merge",
             path = "records/merge",
-            args = (500, 50),
+            args = (50, 10),
```
jneem, line 15: Did you intend to push this?

jneem: Never mind, I tried running them and I understand now...

PR #2430, opened 2025-11-13
Move the inherited bit to PosIdx

jneem:
This PR moves the "inherited" bit from the position table to the position index, so that the evaluator doesn't need to be constantly adding to the position table. As a consequence, the number of allowed positions has gone down to about 2 billion. (But we aren't using more indices for inherited positions, which cancels things out a little)

This doesn't completely remove the need for mutating the pos table during eval: as noted in [this TODO](https://github.com/tweag/nickel/blob/cd67dc5b100f6f8e896c12470f05296319eba20e/core/src/eval/mod.rs#L495) records have `TermPos` inside their fields and so require pos table modifications at runtime. I think that's the only remaining reason (in a few places) that the pos table gets modified during eval.

I measured a 1-2% improvement on most benchmarks.

### DIFF
--- a/core/src/position.rs
+++ b/core/src/position.rs
@@ -18,21 +18,30 @@ impl From<PosIdx> for usize {
 }
 
 impl PosIdx {
-    /// A special value indicating that an value doesn't have a position defined. This is the first
-    /// available position index for values.
-    pub const NONE: PosIdx = Self(0);
+    /// A special value indicating that an value doesn't have a position defined.
+    pub const NONE: PosIdx = Self(u32::MAX);
+    /// We use the highest-order bit to indicate that the position is inherited.
+    const INHERITED_BIT: u32 = (1 << 31);
+    /// Excluding the highest-order bit gives us a mask turning a `PosIdx` into
+    /// and index into a `PosTable`.
+    const IDX_MASK: u32 = PosIdx::INHERITED_BIT - 1;
+    /// Don't allow `0b0111...1` as an index, because making it inherited would
+    /// give us `PosIdx::NONE`. The biggest allowed index is `0b0111...110`.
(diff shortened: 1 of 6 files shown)


utils/src/bench.rs
```diff
@@ -172,7 +172,7 @@ macro_rules! ncl_bench_group {
                             let id = cache.sources.add_file(bench.path(), InputFormat::Nickel).unwrap();
 
                             if matches!(bench.eval_mode, $crate::bench::EvalMode::TypeCheck) {
-                                cache.parse_to_term(&mut pos_table, id, nickel_lang_core::cache::InputFormat::Nickel).unwrap();
+                                cache.parse_to_ast(id).unwrap();
```
jneem, line 175: I was getting failures on the `typecheck-nixpkgs-lib` benchmark because typechecking wanted an Ast but `parse_to_term` doesn't cache one.

yannham: Ah, yeah, I got the same error, but didn't bother yet because I didn't care about typechecking but evaluation. Still, good thing that you fixed this in the meantime.

core/src/position.rs
```diff
@@ -59,29 +68,54 @@ impl Default for PosIdx {
...
+                } else {
+                    PosIdx(idx)
+                }
+            }
+            TermPos::None => PosIdx::NONE,
+        }
     }
 
     /// Returns the position at index `idx`.
     pub fn get(&self, idx: impl Into<PosIdx>) -> TermPos {
```
yannham: Unrelated, but I think we can get rid of this. This is a left-over from when we had `InlinePosIdx` which was convertible to `PosIdx`.

```suggestion
    pub fn get(&self, idx: PosIdx) -> TermPos {
```

PR #2432, opened 2025-11-14
[DON'T MERGE] See benchr results with generalized use of lenses

yannham:
I'm investigating the perf degradation on `generate unchecked` bench. Just running a version in the benchr env to see if that makes a difference.


yannham:
This wasn't the culprit. Closing for now.

PR #2433, opened 2025-11-14
Convert pattern compilation to arrays

jneem:
This reworks record compilation so that the "rest" term is passed in an array instead of as a field in the bindings record. This removes a few record operations (for adding and removing the "rest" term), but the main purpose was to get rid of the bit where we put generated ids as record fields.

I didn't measure any performance difference.

### DIFF
--- a/core/src/term/pattern/compile.rs
+++ b/core/src/term/pattern/compile.rs
@@ -60,38 +60,16 @@ fn insert_binding(id: LocIdent, value_id: LocIdent, bindings_id: LocIdent) -> Ni
     )
 }
 
-/// Generate a Nickel expression which update the `REST_FIELD` field of the working bindings by
-/// remove the `field` from it.
+/// Generate a Nickel expression which removes `field` from `rest`.
 ///
 /// ```nickel
-/// %record/insert% "<REST_FIELD>"
-///   (%record/remove% "<REST_FIELD>" bindings_id)
-///   (%record/remove% "<field>"
-///     (%static_access(REST_FIELD) bindings_id)
-///   )
+/// %record/remove% "<field>" <rest>
 /// ```
-fn remove_from_rest(rest_field: LocIdent, field: LocIdent, bindings_id: LocIdent) -> NickelValue {
-    let rest = make::op1(UnaryOp::RecordAccess(rest_field), Term::Var(bindings_id));
-
-    let rest_shrinked = make::op2(
+fn remove_from_rest(rest: NickelValue, field: LocIdent) -> NickelValue {
+    make::op2(
         BinaryOp::RecordRemove(RecordOpKind::ConsiderAllFields),
         NickelValue::string_posless(field.label()),
         rest,
-    );
-
-    let bindings_shrinked = make::op2(
-        BinaryOp::RecordRemove(RecordOpKind::ConsiderAllFields),
-        NickelValue::string_posless(rest_field),
-        Term::Var(bindings_id),
-    );
-
-    mk_app!(
-        make::op2(
-            record_insert(),
-            NickelValue::string_posless(rest_field),
-            bindings_shrinked,
-        ),
-        rest_shrinked
     )
 }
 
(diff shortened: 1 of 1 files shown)


core/src/term/pattern/compile.rs
```diff
@@ -70,28 +70,11 @@ fn insert_binding(id: LocIdent, value_id: LocIdent, bindings_id: LocIdent) -> Ni
 ///     (%static_access(REST_FIELD) bindings_id)
 ///   )
 /// ```
-fn remove_from_rest(rest_field: LocIdent, field: LocIdent, bindings_id: LocIdent) -> NickelValue {
-    let rest = make::op1(UnaryOp::RecordAccess(rest_field), Term::Var(bindings_id));
-
-    let rest_shrinked = make::op2(
+fn remove_from_rest(rest: NickelValue, field: LocIdent) -> NickelValue {
```
yannham, line 68: Should the documentation of this  function be updated?

yannham:
Ah, wait, I think I get it now. The problem is that we need to _return_ a pair, not take it as an argument. I guess that answers my question

core/src/term/pattern/compile.rs
```diff
@@ -602,79 +596,73 @@ impl CompilePart for RecordPattern {
...
-                        Term::Var(final_bindings_id),
-                    ),
-                    make::op1(
-                        UnaryOp::RecordAccess(rest_field),
-                        Term::Var(final_bindings_id)
-                    )
+            // (%record/insert% <rest>
+            //   final_bindings_id
+            //   rest_id
+            // )
```
yannham: I think the arguments order doesn't in the comment doesn't correspond to the actual generated code. Probably already an issue in the previous version of the comment, that has been propagated.

jneem:
> I think I was rather thinking about the bindings themselves (we know statically the variables captured by the pattern, so we can order them a build the bindings as an array).

I tried this a little, but in the end we need the names to build the environment of the match arm, right? So unless I missed something, we'd need an array of name/value pairs anyway. (When we have bytecode compilation, this objection goes away because we can use some more direct way of resolving variables.)

yannham:
> I tried this a little, but in the end we need the names to build the environment of the match arm, right? So unless I missed something, we'd need an array of name/value pairs anyway. (When we have bytecode compilation, this objection goes away because we can use some more direct way of resolving variables.)

Yes, but those names are known statically, so we don't need to use a hashtable at runtime, I think is my point. For example if we have a lower-level `%bind(<name>)% value continuation`, which is parametrized by an identifier and bind the corresponding value in the environment then proceed with the continuation, we could generate a sequence of nested `bind` from a binding array for a given pattern. Or alternatively just have an array as an associative `[(id1, value1), ..., (idn, valuen)]`, in a slightly less static way. I suppose the general observation is that:

1. The number of bindings is statically known
2. The corresponding identifiers are statically known

Thus we can probably optimize things further than using a dynamic record for that, which is not small, and is probably copied over at each insertion (this is a more general problem for Nickel). For example, we could typically have a way to allocate a fixed-size mutable array of `NickelValue`s directly from Nickel code, and insert into it, although this is maybe a bit  _ad hoc_.

core/src/term/pattern/compile.rs
jneem: I'm always confused by the argument order for `%record/insert%`, but I think this is correct? `<rest>` here stands for the user-provided name that we're binding to the rest of the fields, and `rest_id` is a generated id that evaluates to the rest of the fields. So we're binding that value to the name `<rest>` in the final bindings.

yannham: I don't event pretend to have tried to recall what happens here, but `%primop% a b c` for binary primop is parsed as `make_app!(make::op2(primop, a, b), c)`. Whatever are the semantics or argument order of the primop. Here we have `mk_app!( op2(.., rest_id, final_bindings_id), rest)` which should syntactically correspond to `%record_insert% rest_id final_bindings_id rest`, following dumbly the syntax. This is not what's written in the comment. Or am I missing something?

jneem: Right, sorry, I mixed up `<rest>` and `rest_id` in my comment above. Fixing the doc...

PR #2435, opened 2025-11-15
Allow the input format of stdin to be specified in the CLI

### LINKED ISSUE #2426: Allow non-nickel formats to be passed in through stdin
The Nickel cli can read alternative formats from a file based on the file extension, but all input through stdin is assumed to be Nickel. I think it'd be useful to be able to specify that stdin is in some other format, like to convert other formats to Nickel, or for use with --apply-contract to validate inputs in other formats.

A command line flag that allows you to specify that stdin is in some other format, for example:
```$ echo "{\"a\": 1, \"b\": 2}" | nickel eval --stdin-format json
{ a = 1, b = 2, }
```

This can be done in two steps by first writing the input to a file with the right extension, and then running nickel with the file as input instead of stdin.
```
$ echo "{\"a\": 1, \"b\": 2}" > value.json
$ nickel eval value.json
{ a = 1, b = 2, }
```
This works fine but it's not as convenient.

L0r3m1p5um:
Closes #2426.

As discussed on that issue, I opted to make the flag only valid on inputs from stdin, and made its usage when providing positional file arguments an error. I used the flag name ```--stdin-format``` to make it clear that it's not applicable to other types of inputs.

Here's an example usage:
```
$ echo "{\"a\": 1, \"b\": 2 }" | nickel eval --stdin-format json
{ a = 1, b = 2, }
```


cli/src/input.rs
```diff
@@ -141,3 +148,29 @@ impl<C: clap::Args + Customize> Prepare for InputOptions<C> {
         self.customize_mode.customize(program)
     }
 }
+
+trait InputFormatOptions {
```
yannham, line 157: Nitpick: this trait and the following items could have a short documentation that explain why they exists, and to which arguments they correspond.

cli/tests/integration/main.rs
```diff
@@ -129,3 +129,118 @@ fn merge_mixed_formats() {
...
+    let output = run_input_from_stdin(
+        vec!["export", "--stdin-format", "json", "-f", "toml"],
+        "{ \"foo\": \"hello\", \"bar\": 123 }",
+    );
+    assert_eq!(output.stderr, "");
+    assert_eq!(output.stdout, "bar = 123\nfoo = \"hello\"\n");
+}
+
+#[test]
+fn query_supports_stdin_format_flag() {
```
yannham: I think this test would be more robust as a snapshot test? (no need to encode the precise output, and if it changes, it's easier to update). Also, I wonder if we shouldn't move the tests you just added to a separate module, to avoid creep in `main`.

L0r3m1p5um: I had considered that but I would have to change the snapshot test runner to allow inputs to be passed in through stdin, as well as to allow input formats other than nickel. For the number of tests here it didn't really seem worth it to me. I also wouldn't expect these outputs to change very frequently. I've moved the tests to their own module though.

yannham: Fair enough!

yannham (approved):

PR #2437, opened 2025-11-20
Adds a convert command

### LINKED ISSUE #2213: Add a command to convert/translate from existing configuration to Nickel
Nickel can currently import from and export to different static configuration formats, such as JSON, YAML, and TOML. However, this doesn't cover the use-case when one want to migrate an existing configuration to Nickel: there, you probably want everything to be translated to Nickel, and then work on this code from now on.

Add e.g. a `nickel convert` subcommand that would take a file in an importable format (JSON, TOML or YAML) and output the same configuration as valid, well-formatted Nickel.

People are currently using `nickel eval file.json | nickel format > file.ncl` for that, which is surprisingly rather robust. However, it's not really made for that purpose. Once RFC007 is fully implemented, `nickel eval` won't be able to pretty print proper ASTs anymore (as we'll only have bytecode and values), so this won't work forever.

Additionally, this is slightly wasteful, as we read some JSON, convert it to Nickel, prepare for evaluation, evaluate it (although it doesn't need to - it's already evaluated), and then pretty print it.

Finally, this is not very discoverable.

jneem:
Fixes #2213.

This adds some code for parsing TOML into `Ast`, which feels a little redundant because we already have code to go from TOML to `NickelValue`. But I guess we'll want it anyway, for example in the lsp.


cli/src/convert.rs
```diff
@@ -0,0 +1,120 @@
...
+                std::io::stdin().read_to_string(&mut buf)?;
+                (stdin_format, buf)
+            }
+            (Some(path), None) => {
+                let format = InputFormat::from_path(path).ok_or_else(|| Error::CliUsage {
+                    files: Files::empty(),
+                    error: CliUsageError::CantDetectFormat { path: path.clone() },
+                })?;
+                let data = std::fs::read_to_string(path)?;
+                (format, data)
```
yannham: Nitpick, feel free to ignore (I think small linear bindings don't add much to readability)

```suggestion
                (format, std::fs::read_to_string(path)?)
```

cli/src/convert.rs
```diff
@@ -0,0 +1,120 @@
...
+            }
+        };
+
+        let mut files = Files::empty();
+        let file_id = files.add("todo", data.as_str());
```
yannham: Is it a left-over todo? I think we use a specific name for `stdin` elsewhere in the CLI, don't we? Probably just `<stdin>` or something like that.

cli/src/convert.rs
```diff
@@ -0,0 +1,120 @@
...
+            InputFormat::Toml => {
+                nickel_lang_core::serialize::toml_deser::ast_from_str(&alloc, &data, file_id)
+                    .map_err(|e| ParseError::from_toml(e, file_id).into())
+            }
+            InputFormat::Text => {
+                // We convert text to Nickel by wrapping it in a string.
+                Ok(Node::String(alloc.alloc_str(&data)).into())
+            }
+            #[cfg(feature = "nix-experimental")]
+            InputFormat::Nix => {
```
yannham, line 104: I wonder if we shouldn't print an error here instead. If `nickel convert foo.nix` worked, I would expect that it actually transpiles Nix to Nickel, not that it tries to evaluate it first and then convert the result (there's no evaluation happening for other cases of conversion).

cli/src/convert.rs
jneem: A left-over, sorry. I'm using "<stdin>" for stdin now, and the path otherwise.

cli/src/convert.rs
jneem: Makes sense, I've updated this to error on Nix.

Issue #2418, opened 2025-11-08
Overlapping diagnostics issued by nls


L0r3m1p5um:
**Describe the bug**
I've noticed a case where nls can issue multiple overlapping diagnostics for parsing and typechecking which leads to some misleading errors being shown.

**To Reproduce**
The following file reproduces the issue:
```nickel
{
  x = 1,, # extra comma here
  y = 2,
} : { x : Number, y : Number }
```

There are two diagnostics issued for this file, a parsing error where the extra comma is, and a typechecking error covering the entire record. While both are published, users are likely to miss the parsing error and see only the typechecking error, which looks like this in VSCode.

<img width="563" height="202" alt="Image" src="https://github.com/user-attachments/assets/9f49a55b-96f7-4711-afb3-f5e67a8c26b2" />

**Expected behavior**
In this case the typechecking error is a direct result of the parsing error, so it doesn't make sense to publish the typechecking diagnostic at all here.

**Environment**
 - Version of the code: nls 1.14.0

jneem:
I wonder if this can be addressed by better recovery from the parse error -- I think `{ x = 1,, y = 2}` could trigger a parse error and also be inferred to have type `{ x : Number, y : Number }`.

In general, I'm reluctant to just bail on the first error because I think sometimes it's better to issue more than one diagnostic. And while in this case it seems clear that the parse error caused the typecheck error, in general that seems hard to know. Maybe one reasonable heuristic is not to issue an error if its position contains another error?

yannham:
I agree with @jneem ; in this case it sounds more like the typechecker isn't good at handling this specific parse tree with an error, but we should probably not silence typechecking errors in general if there's been a parse error before: I suspect there are many situations where both are relevant, and can be unrelated. In particular I think "this typecheck error is caused by this parse error" might be pretty hard to decide in general.

For more context, what happens here is that the parse entree end up as a record with a static partt `{x = 1, y  = 2}`, and a dynamic field, probably defined as `"%{<parse error>}" = <parse error>"`, or maybe just `"%{parse error}"` as a field without definition. Since there's a dynamic field, the typechecker considers the record as a dictionary (it can't really do better in that case), hence the `{_ : _a}` type expected. I suppose one solution would be to just entirely filter out/ignore dynamic fields that are parse errors in the typechecker, making it just `{x =  1, y = 1}`. This is more precise, and is given a  subtype of a dictionary type, so it seems to be strictly better.

[...]

PR #2439, opened 2025-11-21
Change record typechecking to ignore fields with unparseable names.

L0r3m1p5um:
This relates to #2418. It solves the specific instance of the issue by changing how typechecking works for records with parsing errors rather than solving the general problem of overlapping diagnostics, so I don't know if you'd want to consider the issue closed or not.

This changes record typechecking so that if the field name is unparseable, that field is ignored. So in the particular instance described in the issue, the record is inferred to have type { x : Number, y : Number }, so type checking succeeds and only the parse error diagnostic is published.
image

I have not done the same when the value of a field is unparseable. So for example, this input will still issue a typechecking diagnostic:
```nickel
{
  x = 1,
  y = 2,
  z = 
} : { x : Number, y : Number}
```
image

I figured I'd leave this in because this would be a common situation when in the middle of editing a record. In that case, the feedback that there's a field being defined that's not allowed by the type is probably at least as important as knowing that there's a parsing error there, which might be just because the input isn't complete.

### DIFF
--- a/core/src/typecheck/record.rs
+++ b/core/src/typecheck/record.rs
@@ -75,12 +75,34 @@ pub(super) struct ShallowRecord<'ast> {
     /// The static fields of the record.
     pub stat_fields: IndexMap<LocIdent, ResolvedField<'ast>>,
     /// The dynamic fields of the record.
-    pub dyn_fields: Vec<(&'ast Ast<'ast>, ResolvedField<'ast>)>,
+    ///
+    /// This should not include any fields where the field name is a parser error, as those
+    /// fields should be ignored for typechecking purposes. Instances created through the
+    /// [ShallowRecord::new] constructor will filter out such fields.
+    dyn_fields: Vec<(&'ast Ast<'ast>, ResolvedField<'ast>)>,
     /// The position of the resolved record.
     pub pos: TermPos,
 }
 
 impl<'ast> ShallowRecord<'ast> {
+    /// Creates an instance of [ShallowRecord].
+    ///
+    /// This constructor will exclude fields where the field name is a parser error, so that
+    /// those fields will be ignored for typechecking purposes.
+    pub fn new(
+        stat_fields: IndexMap<LocIdent, ResolvedField<'ast>>,
+        dyn_fields: Vec<(&'ast Ast<'ast>, ResolvedField<'ast>)>,
+        pos: TermPos,
+    ) -> Self {
+        ShallowRecord {
+            stat_fields,
+            dyn_fields: dyn_fields
+                .into_iter()
+                .filter(|(key, _)| !matches!(key.node, Node::ParseError(_)))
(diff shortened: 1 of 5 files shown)


L0r3m1p5um:
Looks like there was a runner failure with macos. Could someone retry that?

core/src/typecheck/record.rs
```diff
@@ -75,12 +75,34 @@ pub(super) struct ShallowRecord<'ast> {
...
+    pub fn new(
+        stat_fields: IndexMap<LocIdent, ResolvedField<'ast>>,
+        dyn_fields: Vec<(&'ast Ast<'ast>, ResolvedField<'ast>)>,
+        pos: TermPos,
+    ) -> Self {
+        ShallowRecord {
+            stat_fields,
+            dyn_fields: dyn_fields
+                .into_iter()
+                .filter(|(key, _)| !matches!(key.node, Node::ParseError(_)))
```
yannham, line 101: I wonder if we shouldn't also requires that the rest of the field is empty, just to be safe/specific. I'm not sure how NLS parses something like `{foo = 1, "%{blarg" = 2}` (note the missing closing `}`), or something like `{foo = 1, | Contract | default = 2}`, but we probably shouldn't ignore those?

L0r3m1p5um: Your first example behaves the same way either way, you'll get an unexpected end of file parsing error, and no typechecking diagnostics are issued. Here's what the NLS shows with that input modified to try to get it to typecheck anything.
image

The second one does change. Given this input:
```nickel
{
  foo = 1,
  | Contract | default = 2
} : { foo : Number }
```
The current behavior shows a typechecking error here. It's also issuing a parsing error but that's not visible unless you're mousing over exactly the right spot. 
image

The new behavior ignores this field for typechecking and displays only the parser error that happens at that location
image

The new behavior isn't too much of a problem here imo, since you'd have to go back and fix the syntax error anyway, and immediately get the typechecking error upon fixing it. Given how the diagnostics are currently displayed it's maybe a bit preferable because the parser error isn't very visible when both are published. I made a suggestion on the issue that maybe we could dynamically change the severities of one of the diagnostics when overlapping diagnostics are published so that they'd display differently and be easier to distinguish. Like in this case we could set the typechecking error to warn instead and leave the parsing diagnostic as error when we detect an overlap. If something like that would work it would be more worthwhile to preserve the existing behavior I think.

yannham: Ok, I see your point that typechecking parse errors only matters in the LSP, and that in general it's not very interesting to be too "precise", since there will be a parse error anyway. In fact we mostly tolerate parse errors in the typechecker so that the LSP analysis can run, more than being able to typecheck partial programs _per se_, which seconds what you wrote. Sounds good as it is, then :+1:

PR #2440, opened 2025-11-24
Add a paragraph on nickel convert to the manual

jneem:


doc/manual/cli.md
```diff
@@ -389,3 +389,18 @@ For example, the following tests will succeed:
     = 1,
 }
 ````
+
+## `nickel convert`: convert data to Nickel format
+
+The `nickel convert` command allows you to convert plain data (JSON, YAML, TOML, or text)
+into Nickel format. For example, if the file `data.json` contains `{ "foo": 1 }` then
+`nickel convert data.json` will output `{ foo = 1 }`. Text files will converted into
```
yannham:
```suggestion
`nickel convert data.json` will output `{ foo = 1 }`. Text files will be converted into
```

doc/manual/cli.md
```diff
@@ -389,3 +389,18 @@ For example, the following tests will succeed:
     = 1,
 }
 ````
+
+## `nickel convert`: convert data to Nickel format
+
+The `nickel convert` command allows you to convert plain data (JSON, YAML, TOML, or text)
+into Nickel format. For example, if the file `data.json` contains `{ "foo": 1 }` then
```
yannham: You're using `into` many times. I was wondering, is there any subtle semantics difference or specific usage when compared to `to` (assuming we're talking about converting stuff)?

yannham (approved):

doc/manual/cli.md
jneem: Not that I'm aware of. I found a few pages on the internet with opinions on the matter, but not from any particularly trusted source. I changed one "into" to "to" just for some variety...

PR #2441, opened 2025-11-24
Avoid generated variables in yaml

jneem:
When testing `nickel convert` I realized that converting YAML could generate invalid Nickel because of generated identifiers. This PR changes the identifiers used for anchors in YAML so that they're syntactically valid.

The generated names are called `_anchor0`, `_anchor1`, etc. It would be a nicer user experience if they had some relation to the YAML anchor name, but saphyr doesn't expose that right now.

### DIFF
--- a/core/src/serialize/yaml.rs
+++ b/core/src/serialize/yaml.rs
@@ -6,9 +6,15 @@
 //! without passing through `saphyr`'s in-memory `Yaml` structure.
 //! (Also, this lets us support recursive aliases.)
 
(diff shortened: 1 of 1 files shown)


yannham (approved):
I was a bit sad that we need to perform two more traversals on YAML imports, but I guess it's OK if that only happens for documents with anchors. Though we do inflict the penalty for users that only import YAML, without needing to convert it, right? And it makes me think, we are crucially lacking good benchmarks for serialization and de-serialization to be able to measure those kind of effects.

jneem:
Right, we could (but don't right now) avoid the penalty during import. I'll add a benchmark as a follow-up.

PR #2442, opened 2025-11-25
perf: reduce memory consumption of the stack

yannham:
# Compact stack

This PR continues the work to use more compact data structures (getting rude of large Rust enums), tackling the stack. Previously, a stack item was 136 bytes large, while arguments should only take a few bytes.

I first tried to make the enum small again, a bit like for `Term`, but it's hard to get back to reasonable size because we currently need to store various non-trivial data structures on the stack (it will be less the case with a full bytecode VM). It's also probably not a good idea, performance wise, to box everything on the stack, which has strong locality. Instead, the stack is turned into an untyped `u8` buffer, where we push raw data followed by a marker, allowing to decode the raw data. The marker is one byte, which means we have a very few internal fragmentation, if any. Since we never materialize a reference into the stack, we can ignore any kind of alignment constraints: morally, we _move_ owned values onto the stack, but we always materialize them on the Rust stack before doing anything with the data. For `rustc`, the stack is just a raw stream of bytes.

This PR also pre-allocate one page for the stack, which should make small programs a bit faster, by avoiding the first few resizes of the stack.

## Impact

The impact is rather limited in run time but positive and measurable (a small but consistent improvements on benchmarks, visible on benchr plots, and on the opl bench simillarly). It does reduce the total allocated memory quite a bit: on the OPL small benchmark, the peak of live memory goes down from 439MB to 371MB.

### DIFF
--- a/core/src/eval/stack.rs
+++ b/core/src/eval/stack.rs
@@ -678,8 +678,8 @@ impl<C: Cache> Stack<C> {
     }
 
     /// Helper for an ad-hoc check that needs to be performed by the evaluation of a sealed term,
-    /// to decide what to do next. It used to be implemented using a generic `peek_op_cont` methodo
-    /// n, but such a generic method is now impossible to implement with the stack as a byte stream
+    /// to decide what to do next. It used to be implemented using a generic `peek_op_cont` method,
+    /// but such a generic method is now impossible to implement with the stack as a byte stream
     /// implementation. Instead, we moved this check inside stack.
     ///
     /// This method look at the first element of the stack and return a corresponding [SealedCont].


core/src/eval/stack.rs
```diff
@@ -2,210 +2,624 @@
...
+/// We used to store a single type `Marker` which was a big enum of everything that could land in
+/// the stack, having the stack itself just being a `Vec<Marker>`. Given how enums are represented
+/// in Rust, this consumed an huge amount of space for any stack element (more than 100 bytes). As
+/// mentioned above, it's really hard to bring this enum back to a reasonable size; unless we would
+/// allocate a lot of things on the heap, but that would have poor performance and somehow defeats
+/// the purpose of having a fast-to-access stack.
+///
+/// Instead, we now implement the stack as an untyped sequence of bytes (`Vec<u8>`). When we push
+/// an arbitrary data structure to the stack, we first push the data, and then use the last word to
+/// store an [ItemKind] marker. When popping, we look at thefirst word which lets us know how to
```
jneem:
```suggestion
/// store an [ItemKind] marker. When popping, we look at the first word which lets us know how to
```

core/src/eval/stack.rs
```diff
@@ -2,210 +2,624 @@
...
+/// interpret the rest of the data at the top of the stack, and how much bytes we need to pop.
```
jneem:
```suggestion
/// interpret the rest of the data at the top of the stack, and how many bytes we need to pop.
```

core/src/eval/stack.rs
```diff
@@ -215,132 +629,171 @@ impl<C: Cache> Stack<C> {
...
-    /// Try to pop the a string accumulator from the stack. If `None` is returned, the top element
-    /// was not a string accumulator and the stack is left unchanged.
-    pub fn pop_str_acc(&mut self) -> Option<StrAccData> {
-        if self.0.last().map(Marker::is_str_acc).unwrap_or(false) {
-            match self.0.pop() {
-                Some(Marker::StrAcc(str_acc_data)) => Some(str_acc_data),
-                _ => unreachable!(),
+    /// Helper for an ad-hoc check that needs to be performed by the evaluation of a sealed term,
+    /// to decide what to do next. It used to be implemented using a generic `peek_op_cont` methodo
+    /// n, but such a generic method is now impossible to implement with the stack as a byte stream
```
jneem:
```suggestion
    /// to decide what to do next. It used to be implemented using a generic `peek_op_cont` method,
    /// but such a generic method is now impossible to implement with the stack as a byte stream
```

PR #2443, opened 2025-11-25
More static pattern compilation

jneem:
This redoes pattern compilation to be more static, removing the `%pattern_branch%` primop in favor of let bindings. Pattern compilation has been re-written in more of a continuation-passing style, where each pattern compilation step takes in success and failure continuations, and essentially expands to "if this pattern matches, let-bind all the things and <match_cont>. Otherwise, <fail_cont>"

I measured substantial (20%-ish) performance improvements for the arrays/sum benchmark, and not much change for anything else.

~~Marking as draft because I just realized it has a problem with identifier leakage... I think it just needs one more indirection via fresh idents.~~

### DIFF
--- a/core/src/term/pattern/compile.rs
+++ b/core/src/term/pattern/compile.rs
@@ -171,10 +171,10 @@ pub trait CompilePart {
     ///
     /// `match_cont` and `fail_cont` are continuations: `match_cont` is the
     /// one to use if this pattern matches, while `fail_cont` is the one to
-    /// use if this pattern fails to match. Both `match_cont` and `fail_cont`
-    /// are allowed to contain -- as free variables -- identifiers bound by
+    /// use if this pattern fails to match. Both `match_cont` is allowed
+    /// to contain -- as free variables -- identifiers bound by
     /// this match expression. That is, `compile_part` is responsible for
-    /// putting `match_cont` and `fail_cont` into environments containing these
+    /// putting `match_cont` into an environment containing these
     /// identifiers.
     ///
     /// Because of the free variables involved, pattern compilation *must not*
@@ -182,13 +182,13 @@ pub trait CompilePart {
     /// Since a sub-pattern compilation might contain a `fail_cont`, this means
     /// you must avoid generating code like:
     ///
-    /// ```ignore
+    /// ```nickel
     /// let pattern_var = ... in <sub_pattern.compile_part(match_cont, fail_cont)>
     /// ```
     ///
     /// Instead, put the bindings on the inside:
     ///
-    /// ```ignore
+    /// ```nickel
     /// <sub_pattern.compile_part(let pattern_var = ... in match_cont, fail_cont)>
     /// ```
     ///
@@ -426,7 +426,7 @@ impl CompilePart for RecordPattern {
                 make::op1(
                     UnaryOp::ArrayLength,
                     make::op1(
-                        UnaryOp::RecordFields(RecordOpKind::ConsiderAllFields),
+                        UnaryOp::RecordFields(RecordOpKind::IgnoreEmptyOpt),
                         Term::Var(local_value_id),
                     ),
                 ),
--- a/core/src/transform/desugar_destructuring.rs
+++ b/core/src/transform/desugar_destructuring.rs
@@ -73,7 +73,7 @@ pub fn desugar_fun(FunPatternData { mut pattern, body }: FunPatternData) -> Term
 /// in body
 /// ```
 ///
-/// is desugared to
+/// in which `<pat1>` binds `foo` and `<pat2>` binds `bar` is desugared to
 ///
 /// ```text
 /// let
--- a/core/tests/integration/inputs/pattern-matching/ignore_empty_opt.ncl
+++ b/core/tests/integration/inputs/pattern-matching/ignore_empty_opt.ncl
@@ -1,0 +1,6 @@
+# test.type = 'pass'
+{ foo = 1, bar | optional }
+|> match {
+  { foo } => true
+}
+  | std.test.Assert


core/src/pretty.rs
```diff
@@ -289,7 +289,11 @@ impl Allocator {
     ) -> DocBuilder<'a, Self> {
         let size_per_child =
             self.size_constraint() / (record_data.fields.len() + dyn_fields.len()).max(1);
-        if record_data.fields.is_empty() && dyn_fields.is_empty() && !record_data.attrs.open {
+        if record_data.fields.is_empty()
+            && dyn_fields.is_empty()
+            && includes.is_empty()
```
jneem, line 294: Unrelated bug that I ran into while pprint-ast-ing: `{include a}` was being pretty-printed as `{}`.

core/src/transform/desugar_destructuring.rs
```diff
@@ -77,18 +79,16 @@ pub fn desugar_fun(FunPatternData { mut pattern, body }: FunPatternData) -> Term
 /// let
 ///   %b1 = <bound1>,
 ///   %b2 = <bound2>,
-///   %empty_record = {},
 /// in
 /// let
-///   %r1 = <pat1.compile_part(%b1, %empty_record)>,
-///   %r2 = <pat2.compile_part(%b2, %empty_record)>,
+///   %r1 = <pat1.compile_part(%b1, { include foo }, <error>)>,
```
yannham, line 84: What are `foo` and `bar` here? They don't appear previously, unless I missed them.

core/src/term/pattern/compile.rs
```diff
@@ -188,62 +164,78 @@ fn update_with_merge(
...
-    ///    matched expression (`bindings_id`)
+    /// Because of the free variables involved, pattern compilation *must not*
+    /// put `fail_cont` into any let block that binds a non-fresh variable.
+    /// Since a sub-pattern compilation might contain a `fail_cont`, this means
+    /// you must avoid generating code like:
     ///
-    /// The compiled expression must return either `null` if the pattern doesn't match, or a
-    /// dictionary mapping pattern variables to the corresponding sub-expressions of the
-    /// matched value if the pattern matched with success.
+    /// ```ignore
```
yannham: I think `ignores` is, according to the rust doc book, "almost never what you want"[^1]. I guess it assumes it is Rust, but that you should ignore it. I think we should rather use `text` or `nickel` here (even if it's pseudo-nickel).

[^1]: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

core/src/term/pattern/compile.rs
```diff
@@ -327,340 +324,245 @@ impl CompilePart for ConstantPatternData {
...
-    //
-    //  # if one of the previous patterns already matched, we just stop here and return the
-    //  # resulting updated bindings. Otherwise, we try the current one
-    //  if prev_bindings != null then
-    //    prev_bindings
-    //  else
-    //    <pattern.compile(value_id, bindings_id)>
-    //  <end fold>
+    // We clone `match_cont` for each alternative. This looks wasteful
+    // when pretty-printing, but `NickelValue` is cheaply clonable so it's
```
yannham, line 333: I think I don't get what's related to pretty-printing here?

core/src/term/pattern/compile.rs
```diff
@@ -327,340 +324,245 @@ impl CompilePart for ConstantPatternData {
...
+        // (%record/fields_with_opts% local_value_id |> %array/length%) == <self.patterns.len()>
+        let has_right_length = if self.is_open() {
+            NickelValue::bool_true()
+        } else {
+            make::op2(
+                BinaryOp::Eq,
+                make::op1(
+                    UnaryOp::ArrayLength,
+                    make::op1(
+                        UnaryOp::RecordFields(RecordOpKind::ConsiderAllFields),
```
yannham: Did we use to consider all fields before this PR? Since empty optionals should be "transparent" for most operations, I would expect pattern matching to be oblivious to them as well. But maybe we decided against this for good reason in the past, I can't remember.

core/src/term/pattern/compile.rs
```diff
@@ -327,340 +324,245 @@ impl CompilePart for ConstantPatternData {
...
+                NickelValue::number_posless(self.patterns.len()),
             )
-        }
+        };
 
-        let init_bindings = NickelValue::term_posless(Term::Var(bindings_id));
-        let init = array2(init_bindings, value_id);
+        let matches_shallow = and(has_fields, has_right_length);
+
+        let match_cont = if let TailPattern::Capture(tail) = self.tail {
```
yannham, line 439: This means that for patterns that don't capture the tail, we don't have to repeatedly remove fields from the original record. We only pay this priice when there is a tail capturing pattern. I like this better!

yannham (approved):
Not only it is more efficient, but I find the compilation scheme simpler and clearer (as demonstrated by the diff). This eliminates a lot of ceremony around `bindings_id`, null checks, tails, and whatnot. Looks good!

core/src/transform/desugar_destructuring.rs
jneem: Some arbitrary names bound by `<pat1>` and `<pat2>`. I tried to clarify it.

core/src/term/pattern/compile.rs
jneem: When pretty-printing, you get duplicated expressions even though internally it's only a single `NickelValue` that's referenced twice. For example, if a match arm has an or pattern and a large body, it bloats the pretty-printing output.

core/src/term/pattern/compile.rs
jneem: Good point, I changed it to `IgnoreEmptyOpt` and added a test. The previous version tested whether the "rest" record was equal to `{}`, which does ignore empty optionals.

PR #2444, opened 2025-11-26
Add base64 encoding and decoding functions to stdlib

### LINKED ISSUE #1654: base64 encoding in stdlib
In #1498 I need to base64 encode some values for inclusion into a kubernetes secret.

Currently I'm doing that with the coreutils `base64` command [at nix eval time](https://gitlab.uniworx.de/uniworx/machines/srv01/-/blob/9e0066e47240c003f82860c73eafcfd551da60cd/srv02/kubernetes/external-dns/default.nix#L66).

I would prefer to express the fact that values get base64 encoded directly in the nickel template.  
For that a function in std to base64 encode a string would be ideal.

It would be great if the entire space of possibilities (standard charset, urlsafe charset, padded, unpadded) was available for future purposes.

L0r3m1p5um:
Closes #1654.

This adds two new functions to the standard library for base64 encoding and decoding strings, including variants of base64 encoding:
```
std.string.base64_encode
type:  [| 'Standard, 'UrlSafe, 'NoPad, 'UrlSafeNoPad |] -> String -> String
```
```
std.string.base64_decode
type:  [| 'Standard, 'UrlSafe, 'NoPad, 'UrlSafeNoPad |] -> String -> String
```

This isn't quite finished, as I'm still working on the error diagnostics. I did have a couple of questions though. Right now, if there is a decoding error, the diagnostics blame the location in the stdlib where the builtin is called. That's not particularly useful, and I would prefer to blame the location in user code where the stdlib function was called. Is that possible?
image

The other thing I wasn't sure about is what documentation needs to be updated. I've added docs to std.ncl. I assume that the docs at https://nickel-lang.org/stdlib are just generated from that? Or is there somewhere else I'm missing?


jneem:
I think the way to get better error locations is to use a contract: most of the implementation goes in a (private) custom contract. I'm not sure it matters too much in this case, but it's always nice for contracts to be idempotent. So maybe the custom contract should transform a `String` input to an `'Encoded String`, and leave any non-`String` input unchanged. Then the public function just calls `std.contract.check` on the contract and drops the `'Encoded` tag.

Docs in `std.ncl` are enough. When we update the website to point at a new Nickel release, it grabs those docs and renders them.

yannham:
> This isn't quite finished, as I'm still working on the error diagnostics. I did have a couple of questions though. Right now, if there is a decoding error, the diagnostics blame the location in the stdlib where the builtin is called. That's not particularly useful, and I would prefer to blame the location in user code where the stdlib function was called. Is that possible?

Another possibility is to show the callstack. But in general, the best we have is indeed to go through a contract, where we can reuse error reporting mechanisms here. To get a label with nice error reporting, we need to somehow hook this contract in the annotation of the user-facing function. So I guess, something like:

- implement the primop as a contract, or at least as a fallible function that can be wrapped as a contract in an internal function of the stdlib. As @jneem said, contracts must be idempotent, 
- wrap the primop in a stdlib public contract, like `Base64String` or something
- use it in the signature: `base64_decode: [| 'Std, ..etc |] -> Base64String -> String`
- then `base64_decode` unwraps the result if we did something like returning an enum.

Since the contract must be public facing in order to put it in the signature of the function and get proper error reporting, it's not just nice to have it idempotent: we _must_ make it idempotent, because it's an assumption of the optimizer that all contracts are idempotent, so it could break user code if we don't. In practice, this means `Base64String` must probably be something like `std.contract.any_of [ActualStringDecodingContract, [| 'Decoded String |]]`

yannham:
It's also entirely possible to leave nice error reporting for a follow-up pull request :slightly_smiling_face:

L0r3m1p5um:
I've implemented it as a contract now and it did improve the error output:
image

I didn't make the contract public though, it's only created and applied within the ```std.string.base64_decode``` function. The main reason is that the contract would have to be parameterized by base64 encoding variant. I could do that, but then there's a problem with what the type signature of the function should be: ```base64_decode: [| 'Standard, ..etc |] -> Base64String ??? -> String```. The type of the second argument would depend on the first. In any case, I wouldn't really want this contract to be public if it didn't need to, since its behavior is not what is expected from a contract and I don't think it'd be useful outside of the ```base64_decode``` implementation anyway. I'm not sure if idempotence still matters since it can't actually be applied more than one time but I made it idempotent anyway.

core/stdlib/std.ncl
```diff
@@ -4713,6 +4713,96 @@
...
+              }
+            )
+              | [|
+                'Ok Dyn,
+                'Error {
+                  message | String | optional,
+                  notes | Array String | optional,
+                  blame_location | Dyn | optional
+                }
+              |]
```
jneem: Personally, I find the static typing of custom contracts annoying enough that I usually just do `let Base64String | Dyn = ...` to turn it off.

jneem (approved):
Looks good to me, but maybe let's let Yann have a look too.

Whether to make the contract public is a bit tricky.
- if it's public, you can have `base64_decode : Base64String -> String`, which improves the error messages some more
- but this normalizing contract is a bit strange and doesn't sound like it should be public
- and if you make the contract more "normal", there's a performance cost because you need to check the string twice.

Edit: actually, never mind about the public contract. The contract depends on the base64 variant so it can't easily go in the function signature.

yannham:
> The main reason is that the contract would have to be parameterized by base64 encoding variant. I could do that, but then there's a problem with what the type signature of the function should be:

Note that we have done this before in the stdlib. It mostly goes as:

- what we want is a dependent contract. There's some vague plan about making that a first-class syntax in the future, but for now, there's the `std.contract.unstable.DependentFun`.
- wraps the dependent part as a bespoke opaque contract in `std.contract.unstable`. Examples are `ArraySliceFun` and `IndexedArrayFun`.
- use a less precise type annotation for the function (e.g `Format -> String -> String`), and a second layer with the dependent contract for nice error reporting purpose only.

I don't know if this is the best approach, but we use that for many array-related functions for bounds check that give nice error message (`at`, `split_at`, `slice`, etc.). The contracts are "public", but in an unstable and clearly documented as "don't  use it" module. The advantage is that you don't show in the error the somehow internal part `let decoded = str | Base64String`, and I assume it might be in general more robust with respect to showing the proper offending value when there is more indirection.

On the other hand, the non-public contract doesn't sound too bad either, so I guess both approaches are fine.

L0r3m1p5um:
I tried out `DependentFun` but it didn't make the error particularly more understandable, so I decided to leave it as is. It also caused some problems with the static typing of the function although I didn't spend much time trying to figure out a way around that. I did add a little bit more information to the docs though so that there's a bit more context when seeing the internals of the function in the error report.

yannham:
Ok, fair enough. We can always revisit later.

core/src/eval/operation.rs
```diff
@@ -2900,6 +2901,94 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
+                        .map_err(|err| format!("Base64 decoding failed: {err}"))
+                        .and_then(|decoded_bytes| {
+                            String::from_utf8(decoded_bytes).map_err(|err| {
+                                format!("Could not convert decoded value to string: {err}")
+                            })
+                        });
+                    match decode_result {
+                        Ok(decoded_string) => Ok(NickelValue::enum_variant(
+                            "Ok",
+                            Some(NickelValue::string_posless(decoded_string)),
```
yannham: We might want to give the position of the operation to the decoded string. Otherwise, if said string is part of a later, unrelated contract failure, it won't have any location.

core/src/typecheck/operation.rs
```diff
@@ -447,6 +447,25 @@ impl PrimOpType for PrimOp {
                 vec![mk_uniftype::str(), mk_uniftype::str()],
                 mk_uniftype::array(TypeF::String),
             ),
+            // < Standard, UrlSafe, NoPad, UrlSafeNoPad > -> Str -> Str
```
yannham: The `< ... >` enum type syntax is an ancient relic. We'll have to make a pass in this file, but for newly added code, you should use the current syntax, `[| ... |]` instead.

core/stdlib/std.ncl
```diff
@@ -4713,6 +4713,91 @@
       "%
       = fun start end s => %string/substr% s start end,
 
+    base64_encode
+      : [| 'Standard, 'UrlSafe, 'NoPad, 'UrlSafeNoPad |] -> String -> String
+      | doc m%"
+        `base64_encode variant str` base64 encodes the value of a string using the given variant
```
```suggestion
        `base64_encode variant str` base64 encodes a string using the given variant
```

core/stdlib/std.ncl
```diff
@@ -4713,6 +4713,91 @@
...
+         of the encoding.
+
+         The encoding options are:
+         - 'Standard: Use the standard base64 alphabet and include padding.
+         - 'UrlSafe: Use the url-safe variant of the base64 alphabet, with padding.
+         - 'NoPad: Use the standard alphabet with no padding.
+         - 'UrlSafeNoPad: Use the url-safe alphabet without padding.
```
```suggestion
         - `'Standard`: Use the standard base64 alphabet and include padding.
         - `'UrlSafe`: Use the url-safe variant of the base64 alphabet, with padding.
         - `'NoPad`: Use the standard alphabet with no padding.
         - `'UrlSafeNoPad`: Use the url-safe alphabet without padding.
```

core/stdlib/std.ncl
```diff
@@ -4713,6 +4713,91 @@
...
+        std.string.base64_encode 'UrlSafeNoPad "😎"
+        # => "8J-Yjg"
+        ```
+      "%
+      = fun variant s => %string/base64_encode% variant s,
+
+    base64_decode
+      : [| 'Standard, 'UrlSafe, 'NoPad, 'UrlSafeNoPad |] -> String -> String
+      | doc m%"
+        `base64_decode variant str` decodes the value of the string using the given variant of
```
```suggestion
        `base64_decode variant str` decodes a string using the given variant of
```

core/stdlib/std.ncl
```diff
@@ -4713,6 +4713,91 @@
...
+         base64 encoding.
+
+         The encoding options are:
+         - 'Standard: Use the standard base64 alphabet and include padding.
+         - 'UrlSafe: Use the url-safe variant of the base64 alphabet, with padding.
+         - 'NoPad: Use the standard alphabet with no padding.
+         - 'UrlSafeNoPad: Use the url-safe alphabet without padding.
```
```suggestion
         - `'Standard`: Use the standard base64 alphabet and include padding.
         - `'UrlSafe`: Use the url-safe variant of the base64 alphabet, with padding.
         - `'NoPad`: Use the standard alphabet with no padding.
         - `'UrlSafeNoPad`: Use the url-safe alphabet without padding.
```

core/tests/integration/inputs/stdlib/string_base64_encoding.ncl
```diff
@@ -0,0 +1,18 @@
+# test.type = 'pass'
```
yannham: Note that the snippet `nickel multiline` of the stdlib should be tested in the CI via `nickel test` already. So if this test file doesn't have more content than the `base64_xxx` functions' documentation, it can be either removed (or filled with different content).

L0r3m1p5um: The tests are the same, I've removed it.

yannham:
Perfect. Thanks again for tackling this!

Issue #2186, opened 2025-03-05
Optional fields in record types


jneem:
It's pretty common to have a record with several fields, not all of which are always "interesting." Usually there's some sort of "default" interpretation if these fields are not specified. Some examples:

- in JSON schema, "properties" are not required by default. (JSON schema is only about structure and not semantics; it doesn't specify the interpretation of missing properties)
- in react, components can have [dozens](https://pursuit.purescript.org/packages/purescript-react-basic-dom/6.1.0/docs/React.Basic.DOM.Generated#v:label) of optional properties. This was [apparently](https://hgiasac.github.io/posts/2018-11-18-Record-Row-Type-and-Row-Polymorphism.html) one of the motivations for the introduction of row types in purescript
- in nickel, custom contracts are based around a `{ message, notes, blame_location }` error datum, where all of the fields are optional.

In all of these cases, a primary motivation seems to be to reduce the inconvenience of needing to specify uninteresting data. It might also be worth mentioning rust, which doesn't have optional fields as such, but does support syntax like `Struct { a: 1, ..Struct::default() }` and is [considering](https://rust-lang.github.io/rfcs/3681-default-field-values.html) more flexible extensions.

In nickel, the current situation is that our record *contracts* support optional fields but our record *types* don't. The consequence of this is that
- functions that accept or produce records with optional fields cannot be statically typed
- users of such functions must either forego static typing or apply contracts (and these contracts can be verbose because of the optional fields).

What should we do about this? Some possibilities:

# Nothing

We've already come to terms (e.g. #1228) with the fact that contracts and types are different and shouldn't be treated identically. Maybe optional fields could be an instance of this: if you need them, use contracts.

This feels a bit unsatisfying to me because optional fields of a record can be checked eagerly, so the problems discussed in #1228 don't apply.

# Allow record rows to be optional

Just to make up a syntax, suppose we allow record types like `{ foo : Number, bar ?: Number }`, where the "?:" denotes an optional field. We would need to flesh out how this affects the typesystem, but a few starting points

- We add a subtyping rule that says "deleting optional fields results in a subtype." I think this basically amounts to adding the rule that `{}` is a subtype of `{ foo ?: T }`. 
[...]

PR #2445, opened 2025-11-26
Add a YamlDocuments export format

### LINKED ISSUE #2408: Export an array as multiple documents in one YAML file
It is common in YAML to pack multiple documents in one file. In fact, when importing a YAML file, Nickel already handles this: a bunch of YAML documents packed together will be deserialized as an array of values. However, the other direction isn't covered: it's currently not possible to export a list of values as YAML documents. If one exports an array with format `Yaml`, then it will be just be exported as a YAML array.

I suppose the most logical route is to have a specialized export format, e.g. `yaml-docs` or `yaml-multi` or `yaml-documents`, etc. (to be bikeshedded) that one could use with `nickel export`, which would expect the result to be an array and produce a single file with one YAML document per element. 

- Currently, you can probably hack something together in pure Nickel, by manually creating a string from an array of values and `std.deserialize 'Yaml`, but this is quite tedious and manual.
- Another possibility is to export the array first, or multiple files, and then use a post-processing step to merge them together, but that requires an additional tool on top of Nickel, for something rather simple.

jneem:
Fixes #2408.

This doesn't add the requested `'YamlDocuments` variant to `std.serialize` because that would be a breaking change (as `std.serialize` is statically typed). I'm not sure how to navigate that...

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -2876,7 +2876,7 @@ impl IntoDiagnostics for ExportErrorData {
             ExportErrorKind::ExpectedArray { value } => {
                 vec![
                     Diagnostic::error()
-                        .with_message("yaml-documents export requires an array")
+                        .with_message("yaml-documents export expects an array")
                         .with_labels(vec![primary_term(&pos_table, &value, files)])
                         .with_notes(notes),
                 ]
--- a/cli/src/convert.rs
+++ b/cli/src/convert.rs
@@ -9,6 +9,7 @@ use nickel_lang_core::{
     error::{ParseError, Reporter as _},
     files::Files,
     parser::{self, ErrorTolerantParser, lexer::Lexer},
+    serialize::yaml::Listify,
 };
 
 use crate::{
@@ -88,10 +89,13 @@ impl ConvertCommand {
                 nickel_lang_core::serialize::yaml::load_json(&alloc, &data, Some(file_id))
                     .map_err(|e| e.into())
             }
-            InputFormat::Yaml => {
-                nickel_lang_core::serialize::yaml::load_yaml(&alloc, &data, Some(file_id))
-                    .map_err(|e| e.into())
-            }
+            InputFormat::Yaml => nickel_lang_core::serialize::yaml::load_yaml(
+                &alloc,
+                &data,
+                Some(file_id),
+                Listify::Auto,
+            )
+            .map_err(|e| e.into()),
             InputFormat::Toml => {
                 nickel_lang_core::serialize::toml_deser::ast_from_str(&alloc, &data, file_id)
                     .map_err(|e| ParseError::from_toml(e, file_id).into())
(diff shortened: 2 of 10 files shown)


yannham:
I would be inclined to just do the breaking change, although it's not great, rather than having  inconsistencies between the CLI and the stdlib's `serialize`.

As it's a stdlib function parameter, using a supertype shouldn't break any call site. The only thing that could break are typed wrappers around `serialize`, like `my_serialize : [| 'Json, 'Yaml, 'Toml, 'Text |] -> Dyn -> String = std.serialize`. We can't exclude someone did that in the wild, but it's much less common, and the type error is rather good (I just tried locally).

Another argument would be that the typechecker might accept old code without even making this change breaking in the future, if we have better handling of enum embeddings (typically adding subtyping rules for enums types). Though we don't have that right  now, so in practice, if we include the change, it would be breaking in the foreseeable future.

What do you think?

jneem:
Yeah, I agree that it's unlikely to cause breakage in practice. I'm just a bit nervous that if we get in the habit of changes like this then we'll cause breakage eventually.

I agree that subtyping for enums would be a nice way to avoid this. I guess it's a bit related to #2186, where we broke js2n with an optional record field...

Anyway, I'll update this PR to include `std.serialize`

jneem:
Ok, I added `std.serialize` support and found another little issue: I had originally planned *not* to support `'YamlDocuments` in `std.deserialize`, but then the round-trip behavior of `std.serialize` and `std.deserialize` becomes weird: if I serialize a one-element array with `'YamlDocuments` and then deserialize with `'Yaml`, it drops the array.

For that reason, the PR in its current state adds `'YamlDocuments` support for `std.deserialize`, and the implementation unconditionally produces an array even if there's only one document. I'm not confident that this is the right solution...

yannham:
> For that reason, the PR in its current state adds 'YamlDocuments support for std.deserialize, and the implementation unconditionally produces an array even if there's only one document. I'm not confident that this is the right solution...

I don't find that shocking. You can see one YAML document as a special instance of a singleton of YAML documents. With that point of view, IMHO it makes sense to deserialize it as a singleton. The types match, it's not absurd, and I don't see what else you could do.

core/src/error/mod.rs
```diff
@@ -2869,6 +2873,14 @@ impl IntoDiagnostics for ExportErrorData {
                         .with_notes(notes),
                 ]
             }
+            ExportErrorKind::ExpectedArray { value } => {
+                vec![
+                    Diagnostic::error()
+                        .with_message("yaml-documents export requires an array")
```
yannham: This matches the error name and sounds a tad more natural, IMHO.

```suggestion
                        .with_message("yaml-documents export expects an array")
```

test.ncl
```diff
@@ -0,0 +1,9 @@
+# test.type = 'pass'
```
yannham: left-over? This is a test file at the root of the repo.

PR #2447, opened 2025-11-27
refactor: re-use `make_unique` and count (avoided) clones

yannham:
This is a small PR split off from some performance investigation. This add metrics for when we avoided clones in various methods of `NickelValue` or associated lenses. Doing so, I realized that there was  some code duplication, so this PR includes a small refactoring that re-uses `make_unique` and its variant instead of re-implementing the same logic many times (which also requires us to use `increment` in less places).

### DIFF
--- a/core/src/eval/value/lens.rs
+++ b/core/src/eval/value/lens.rs
@@ -8,6 +8,7 @@ use crate::{
     error::{EvalErrorKind, ParseError},
     files::FileId,
     identifier::LocIdent,
+    metrics::increment,
     term::{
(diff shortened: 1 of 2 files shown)


yannham:
I realized by looking at other perf reports that for some reason this PR is increasing the memory usage quite a bit, and in particular leaks almost one third memory more than `master` (and Nickel is already leaking quite a bit). I currently have absolutely no idea why.

yannham:
Ok, the leak is fixed in the latest version. Sneaky low-level assignment `value.data = some_ptr` didn't properly decreased the refcount in a refactoring, causing data to be leaked.

jneem (approved):

PR #2451, opened 2025-12-01
perf: optimize term and metadata representation

yannham:
# Motivation

The PR that minimized the size of term was rather blunt: we fixed a threshold, and tried to box everything that didn't fit. This isn't necessarily the optimum, because boxing has a cost: it requiries allocation and adds an indirection. This PR tries to inline a bit more data, at the cost of making `Term` slightly bigger, while still keeping an upper bound of 64 bytes max (which should be a cache line on most machines).

From there, continuing the work of previous pull requests, this PR examines the `RecordData` structure, and tries to make  it more performant (either smaller and/or cheaper to clone). First, `FieldMetata` is `392` bytes large. It is stored inlined in each `Field`, which is itself in `RecordData`, meaning that a record such as `{foo, bar, baz}` is already bigger than `1KB` because of metadata alone, which are all empty. Not only that, but any copy of this value in the VM will need to move as many bytes around.

Another remark is that `doc` is stored as `Option<String>`, while it's basically never modified, but is at risk of being copied over many times (values inherit  the metadata from their contract, typically). Let's try to do better.

## Content

- `Term`: we "uninline" mostly primops operations, which are more common at runtime (many other things are evaluated away, like recursive records, errors, etc.). The new size of `Term` is 56 bytes. We also "moved" `Rc` up in `Annotated`, because a `TypeAnnotation` is likely to never change, but the rest of the data might be moved around.
- We make `doc` shareable, by putting it behind a `Rc<str>`.
- We make `FieldMetadata` shareabe, by putting it behind an `Rc`. Cloning `RecordData` should become noticeably cheaper.
- We put the rc behind an `Option`. This means that empty metadata are zero-cost (zero-alloc, at least), which is the case for many local small records.

## Benchmarks

I've mostly benchmarked on the OPL real world codebase. Let's see the plot on benchr in this PR for micro benchmarks. I've tried to benchmark things incrementally, to make sure each step was noticeably better:

- the un-inline `Term` gains a small percentage in time, but lose (as expected) a bit in memory. All in all, the impact, albeit measurable, is rather small (order of a percent).
- making `doc` and `FieldMetadata` shareable makes a consistent gain in leaked memory (around 25MB less, which, on the small OPL benchmark, is around 11.3%). The max live memory gain is smallish (for some reason). It also makes for a noticeable time gain in the medium OPL benchmark, around 6%
- Additionally making `Rc<FieldMetadata>` optional to optimize for empty metadata wins additional 5MB in leaked memory, and around 8MB of peak live memory (2.2%), and a few more points in time on the medium sized OPL benchmark.

All in all, this PR has a noticeable impact on the OPL benchmark (bigger than previous changes like label layout optimization), with around 13.6% leaked memory saved in the small OPL benchmark (~30MB), around 2.2% max live memory less (~7MB), and in the 7-8% ballpark of time improvement for the medium OPL benchmark.

### DIFF
--- a/core/src/ast/compat.rs
+++ b/core/src/ast/compat.rs
@@ -1858,7 +1858,7 @@ fn merge_fields(
         (Some(t1), Some(t2)) if metadata1.priority == metadata2.priority => {
             (Some(merge_values(id_span, t1, t2)), metadata1.priority)
         }
-        (Some(t), _) if metadata1.priority > metadata2.priority => (Some(t), metadata2.priority),
+        (Some(t), _) if metadata1.priority > metadata2.priority => (Some(t), metadata1.priority),
         (_, Some(t)) if metadata1.priority < metadata2.priority => (Some(t), metadata2.priority),
         (Some(t), None) => (Some(t), metadata1.priority),
         (None, Some(t)) => (Some(t), metadata2.priority),


cli/src/doctest.rs
```diff
@@ -430,7 +430,7 @@ fn doctest_transform(
         let mut doc_fields: Vec<(Ident, NickelValue)> = Vec::new();
 
         for (id, field) in &record_data.fields {
-            if let Some(doc) = &field.metadata.doc {
+            if let Some(doc) = &field.metadata.doc() {
```
jneem:
```suggestion
            if let Some(doc) = field.metadata.doc() {
```

core/src/ast/compat.rs
```diff
@@ -1835,19 +1851,17 @@ fn merge_fields(
...
-            (Some(t), field1.metadata.priority)
-        }
-        (_, Some(t)) if field1.metadata.priority < field2.metadata.priority => {
-            (Some(t), field2.metadata.priority)
+        (Some(t1), Some(t2)) if metadata1.priority == metadata2.priority => {
+            (Some(merge_values(id_span, t1, t2)), metadata1.priority)
         }
-        (Some(t), None) => (Some(t), field1.metadata.priority),
-        (None, Some(t)) => (Some(t), field2.metadata.priority),
+        (Some(t), _) if metadata1.priority > metadata2.priority => (Some(t), metadata2.priority),
```
jneem:
```suggestion
        (Some(t), _) if metadata1.priority > metadata2.priority => (Some(t), metadata1.priority),
```

jneem (approved):

core/src/ast/compat.rs
yannham: Ooch, good catch, this one might have been a headache :sweat_smile:. And it's surprising that no test caught that; I shall add one for multiple overriding.

cli/src/doctest.rs
yannham: It's strange that clippy didn't complain about this one.

yannham:
Most benchmarks show either nice gains or noise, but the `generate normal` is taking an abnormally big hit (almost x2 on the biggest one). Will investigate before merging.

yannham:
It's strange, I can't reproduce locally. I don't get any difference with `master` for `generate x normal` for `x` in `{50, 100}`. Going to take a look at the memory consumption, but otherwise I'm wondering if it's just benchr being flaky. Let's rerun benchr to make sure.

yannham:
I can reproduce locally with `x` set to `200` (which is interesting, that the difference doesn't show for `100`). I isolated the changes to `Term` (that is, the unlining, without changing `FieldMetadata`), and this shows the same as master, so the culprit is around metadata. Of course adding an `Rc` can be a pessimization for the cache if we do have metadata and do look at them very often, but I can't see how it is the case in practice (once the contracts are generated, they should be in `pending_contracts`, which is NOT part of metadata). Will try to investigate more and understand what changed before merging.

yannham:
After retrying, I actually can't reproduce it locally. I suspect there were some version shenanigan (I don't know which master `benchr` targeted, nor am I sure about the initial repro), but I did again extensive profiling using both `cargo bench`, and a standalone `runner.ncl` that calls the same Nickel code, with various sizes, looking at both memory consumption, cachegrind metrics, actual running time, etc. and this PR's version is invariably better on all metrics for the `generate normal` bench than master.

PR #2452, opened 2025-12-01
Fix evaluation order of record pattern matches, and related ident leakage

### LINKED ISSUE #2450: Breaking change in match statement evaluation
**Describe the bug**
There's a breaking change in how match statements are evaluated in the master branch as of 2eea9b5c752c9cfb3fbab1c48a6a6a5faf296685
Seems like evaluation on some fields is being forced during pattern matching that weren't evaluated before.

**To Reproduce**
The following input works in 1.14.0 but not in master.
```nickel
let x = { has_number = false } in
{ has_number = x.has_number, test = x.number == 1 } |> match {
  { has_number = true, test = true } => "number == 1",
  { has_number = true, test = false } => "number != 1",
  { has_number = false, .. } => "N/A",
}
```
In 1.14.0, this evaluates to "N/A". In master, it fails because it tries to evaluate ```x.number == 1``` and tries to access a missing field.

**Expected behavior**
I would expect this to evaluate to "N/A" as it does in 1.14.0

jneem:
The issue was essentially that we were checking patterns in record fields in the wrong order. But fixing the order exposed another bug with bindings leaking into the wrong match arm (which fortunately I was already testing for).

Basically, I had thought that binding leakage could be avoided if we make sure never to shadow anything in the failure continuation. But while folding record (and array) pattern matching, the failure continuation becomes part of the success continuation. So in fact you need to avoid shadowing anything ever. This PR avoids shadowing by only using fresh vars during `CompilePart::compile_part`, and doing the real binding just before the match arm (or the guard, if there is one).

Fixes #2450.

### DIFF
--- a/core/src/term/pattern/compile.rs
+++ b/core/src/term/pattern/compile.rs
@@ -33,7 +33,7 @@ use crate::{
     position::PosTable,
     term::{
         BinaryOp, MatchBranch, MatchData, NAryOp, RecordExtKind, RecordOpKind, Term, UnaryOp, make,
-        record::FieldMetadata,
+        pattern::bindings::Bindings, record::FieldMetadata,
     },
 };
 
@@ -171,26 +171,13 @@ pub trait CompilePart {
     ///
     /// `match_cont` and `fail_cont` are continuations: `match_cont` is the
     /// one to use if this pattern matches, while `fail_cont` is the one to
-    /// use if this pattern fails to match. Both `match_cont` is allowed
-    /// to contain -- as free variables -- identifiers bound by
-    /// this match expression. That is, `compile_part` is responsible for
-    /// putting `match_cont` into an environment containing these
-    /// identifiers.
+    /// use if this pattern fails to match. Both `match_cont` and `fail_cont`
+    /// are allowed to contain -- as free variables -- identifiers bound by this
+    /// match expression.
     ///
-    /// Because of the free variables involved, pattern compilation *must not*
-    /// put `fail_cont` into any let block that binds a non-fresh variable.
-    /// Since a sub-pattern compilation might contain a `fail_cont`, this means
-    /// you must avoid generating code like:
-    ///
-    /// ```nickel
-    /// let pattern_var = ... in <sub_pattern.compile_part(match_cont, fail_cont)>
-    /// ```
-    ///
-    /// Instead, put the bindings on the inside:
-    ///
-    /// ```nickel
-    /// <sub_pattern.compile_part(let pattern_var = ... in match_cont, fail_cont)>
-    /// ```
+    /// Pattern compilation must avoid shadowing any variables. In particular,
+    /// `compile_part` is not responsible for binding any of the free variables
+    /// in `match_cont` or `fail_cont`.
     ///
     /// As a motivating example, consider a match block like
     ///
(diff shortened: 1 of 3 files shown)


core/src/term/pattern/compile.rs
```diff
@@ -207,35 +195,44 @@ pub trait CompilePart {
...
+        bindings: &HashMap<LocIdent, LocIdent>,
     ) -> NickelValue;
 }
 
 impl CompilePart for Pattern {
     // Compilation of the top-level pattern wrapper. If there's an alias,
     // compiles to
     //
-    // pattern_data.compile(value_id, let <alias> = value_id in <match_cont>, <fail_cont>)
+    // pattern_data.compile(value_id, let %alias_id% = value_id in <match_cont>, <fail_cont>)
```
yannham: Nitpick: I don't think we use `%...%` syntax elsewhere in this file, as a meta-syntax. Reading like this, it looks like a Nickel operator, but it's not. I would suggest to either use `alias_id` like `value_id` after it, or `<alias_id>` if you think it better fits as a meta variable.

core/src/term/pattern/compile.rs
```diff
@@ -438,7 +446,7 @@ impl CompilePart for RecordPattern {
 
         let match_cont = if let TailPattern::Capture(tail) = self.tail {
             // let
-            //   <tail> = (%record/remove_with_opts% field0 (%record/remove_with_opts% field1 local_value_id) ... )
+            //   %tail_id% = (%record/remove_with_opts% field0 (%record/remove_with_opts% field1 local_value_id) ... )
```
yannham: Same remark about meta syntax.

core/src/term/pattern/compile.rs
```diff
@@ -836,9 +858,18 @@ impl Compile for MatchData {
...
+        // to fresh variables (e.g. `{ foo => %1, bar => %2 }`) if `foo` and
+        // `bar` are the variables bound by the current pattern. We pass
+        // this map to the pattern compilation so that it will bind `%1` to
+        // the expression that we'll eventually bind `foo` to.
+        // See `CompilePart::compile_part` for why we do this.
+        //
         // <for branch in branches.rev()
         //  - fail_cont is the accumulator
         // >
+        //    let foo = %1, bar = %2, <...other bindings> in
```
yannham: I think the let-biniding isn't actually put in the enclosing scope, but rather attached in `match_cont`, and _then_ compiled, right?

core/src/transform/desugar_destructuring.rs
```diff
@@ -129,29 +130,35 @@ pub fn desugar_let(
...
         }
 
-        let bindings_record = NickelValue::term_posless(Term::rec_record(
-            Default::default(),
-            includes,
-            Vec::new(),
-            None,
-            false,
-        ));
+        let bindings_record = NickelValue::record_posless(RecordData {
```
yannham, line 145: The documentation above still mentions `include`. It should be updated, I suppose.

core/src/term/pattern/compile.rs
jneem: Fair enough. I was trying to hint that `%alias_id%` was a generated identifier and not the actual pattern variable. I've given up on hinting, and now write that out explicitly.

core/src/term/pattern/compile.rs
jneem: Good point, fixed.

PR #2453, opened 2025-12-02
Try compiling matches at Ast -> NickelValue time

jneem:
This moves pattern match compilation forward: instead of compiling them during eval, it compiles them during `Ast -> NickelValue` conversion.

This causes some regressions in pretty-printing `NickelValue`, but I think we're probably moving in that direction anyway: once the runtime representation uses bytecode, we really won't be able to pretty-print it nicely.

- This could be a transform instead of happening during AST lowering. But then we don't get to delete `Term::Match`
- Conversely, we could extend this approach to the other places that pattern compilation happens. Then we could get rid of `Term::FunPattern` and `Term::LetPattern`.

Performance-wise, I see ~10% improvements to a few benchmarks (sums, and the foldl/foldr ones with strings).

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -1129,8 +1129,6 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                 ValueContentRef::Term(Term::Fun(FunData { arg, body })) if !has_cont_on_stack => {
                     if let Some((idx, pos_app)) = self.stack.pop_arg_as_idx(&mut self.context.cache)
                     {
-                        // FIXME: back when Match was compiled at eval time, it didn't modify the
-                        // call stack. Should we replicate that?
                         self.call_stack.enter_fun(&self.context.pos_table, pos_app);
                         env.insert(arg.ident(), idx);
                         Closure {


core/src/eval/mod.rs
```diff
@@ -1130,6 +1129,8 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                 ValueContentRef::Term(Term::Fun(FunData { arg, body })) if !has_cont_on_stack => {
                     if let Some((idx, pos_app)) = self.stack.pop_arg_as_idx(&mut self.context.cache)
                     {
+                        // FIXME: back when Match was compiled at eval time, it didn't modify the
+                        // call stack. Should we replicate that?
```
jneem: I was wondering about this, because it's a change in behavior but I didn't see any test failures. Is it important that a function application modifies the call stack but a match doesn't? If so, I can just add a `bool` flag to `FunData` to say whether it needs to touch the call stack.

jneem:
Ok, I converted let patterns and fun patterns also. Now we have no more patterns in the runtime representation, so I was able to get rid of `term::pattern::Pattern` altogether. Since the `term::pattern` module is now pretty much empty, I was planning to move everything in `term::pattern::compile` up one level to `term::pattern`. I haven't done it yet because I think it makes the github diff view confusing.

core/src/eval/mod.rs
yannham: I don't think this has any issue. In general the callstack is only used for error reporting, and even then, given how lazy evaluation work (in a kinda of CPS/tail-call way), it's best-effort heuristics anyway. It could be argued that `match`, acting like a function, should induce a callstack entry indeed.

PR #2455, opened 2025-12-04
perf: store thunk data directly in value blocks

yannham:
This PR fixes a useless indirection when storing a closure/thunk as a value (or, before that, as a term). We would handle this case by storing a `Thunk`, which is itself of the form `Rc<RefCell<ThunkData>>`. However, value blocks are already reference-counted, so it doesn't make much sense to store an rc-ed pointer into another rc-ed pointer for no obvious benefit.

This commit implements the usual layout for thunks, which is to store directly the mutable thunk data into the value block, removing an indirection and an allocation.

There is the question of how to define `Thunk`, now that it is the same as a `NickelValue` or a `ValueBlockRc`, just one specific case. I started by performing operations on `value::ThunkData := RefCell<lazy::ThunkData>` first, but some operations do require not only the content, but the original `Rc` pointer (typically, for cloning or for testing physical equality). In the end, I made `Thunk` a smart constructor/transparent wrapper around `NickelValue`, which encodes at the type level that we know that this value is a block of type thunk.

### DIFF
--- a/core/src/eval/cache/lazy.rs
+++ b/core/src/eval/cache/lazy.rs
@@ -134,6 +134,7 @@ impl ThunkData {
 
     /// Create new revertible thunk data.
     pub fn new_rev(orig: Closure, deps: FieldDeps) -> Self {
+        increment!("Thunk::new_rev");
         ThunkData {
             inner: InnerThunkData::Revertible {
                 orig: Rc::new(orig),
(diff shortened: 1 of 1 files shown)


core/src/eval/cache/lazy.rs
```diff
@@ -628,64 +645,45 @@ impl Thunk {
...
+        NickelValue::phys_eq(&this.0, &that.0)
     }
+}
 
-    /// Return a unique identifier for this thunk used for fast equality checking or other
-    /// optimizations. If two thunks have the same UID, they store the same expression. The
-    /// converse isn't true: two thunks can be equal as expressions but have different UID.
-    ///
-    /// In practice, the UID is currently the underlying `Rc` pointer value.
-    pub fn uid(&self) -> usize {
```
yannham, line 645: `uid()` seemed to not be used anymore.

yannham:
This wins a noticeable chunk of memory usage, but is around 2% slower on the OPL benchmark. Investigating a bit. I can see with cachegrind that the instruction cache has different stats (otherwise there seems to be slightly less branches, slightly less mispredictions, and slightly less data access/data cache misses), so maybe it's just noise from the change of layout of the binary :man_shrugging: Another possibility is that the improvement is nullified by our less optimal implementation of `Rc` than in the stdlib, as `clone/drop` for  `NickelValue` is probably stressed a lot.

yannham:
Ok, after a round of optimizing a bit the reference counting part of `NickelValue` (mostly copying the stdlib), I got the 2% back, and even 2% more. I'll push the optimization in a follow-up PR, but let's move forward with this one independently first.

jneem (approved):

PR #2456, opened 2025-12-05
perf: optimize refcounting

yannham:
Depends on  #2455.

This commit mostly copies the style of the stdlib in `std::rc::Rc`, in hope of  gaining the same kind of optimizations. It does nullify the loss of #2455, and wins overall 5-7% on the OPL benchmark in runtime when compared to master. After a bit of experimentation, it seems that the `[u8; 7]` representation had to go for those optimizations to fire properly.

### DIFF
--- a/core/src/eval/value/mod.rs
+++ b/core/src/eval/value/mod.rs
@@ -1858,11 +1858,10 @@ impl ValueBlockRc {
     ///
     /// # Safety
     ///
-    /// Same conditions as for [ValueBlockRc::from_raw].
+    /// `ptr` must be a valid, non-null pointer to a value block.
     #[inline]
     pub unsafe fn from_raw_unchecked(ptr: *mut u8) -> Self {
-        // Safety: if the pointer is coming from a call to `into_raw` (precondition), it can't be
-        // nulll (invariant of `ValueBlockRc`)
+        // Safety: function pre-condition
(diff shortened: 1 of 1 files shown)


core/src/eval/value/mod.rs
```diff
@@ -1580,96 +1644,98 @@ impl TryFrom<u8> for DataTag {
...
+        unsafe {
+            std::hint::assert_unchecked(count != 0);
+        }
+
+        self.set_ref_count(count + 1);
     }
 
-    fn decr_ref_count(&mut self) {
+    #[inline]
+    fn dec_ref_count(&self) {
```
jneem, line 1738: Is it not worthwhile to assert here that `count != 0`? (Naively, I'd expect this one to be more valuable than the other one, because it means the subtraction can't overflow)

yannham: To be honest I mostly bluntly copied the stdlib, which doesn't have this assertion. But I suppose it doesn't cost much to add it, as it's surely verified.

yannham: Reading the documentation for `assert_unchecked`, I think I'll pass for now, at least as long as we can't measure the effect properly.

>  Any use should come with a repeatable benchmark to show the value, with the expectation to drop it later should the optimizer get smarter and no longer need it.

PR #2458, opened 2025-12-06
Fix false negative in contract equality

jneem:
This fixes a performance regression caused by a false negative in contract equality: a `ValueContentRef::Term(Term::Var(_))` was comparing false against a `ValueContentRef::CustomContract`, instead of resolving the variable in its environment and comparing again.


jneem:
It's definitely unreachable, thanks for catching that. But there is one extra case that it could catch (if it weren't unreachable): when you have two identical ids that aren't in the environment. I just checked and that case is triggered by our test suite, so maybe it's worth keeping it in?

yannham:
Ah, you're right. I think it can be useful in open environments; I don't remember exactly when that can happen, but I think it does happen. Maybe for symbols in the stdlib, or something like that.

jneem:
Ok, I moved up the var/var case so it isn't dead code, but otherwise left it the same.

PR #2461, opened 2025-12-09
fix: missing enum closurization

### LINKED ISSUE #2460: Extra primops after new value representation
Before the new value representation, exporting the following program triggered a single `+` primop. In current master, it triggers 2. (The only other primops are 5 `force`s and 4 `seq`s. Those numbers didn't change.)

```
{
  a = b,
  b = 'Enum (1 + 1),
}
```

(I minimized this from a much larger benchmark in which the number of `+` ops went from 200k to 400k, causing a measurable performance difference.)

yannham:
Closes #2460. We were not closurizing enum variants during AST conversion/compilation, leading to `force` evaluating two times the argument of enums.

This indeed divides the number of `(+)` primops by 2 in the OPL benchmark v3 (2M -> 1M ops), and reduces various other metrics.

### DIFF
--- a/core/src/eval/operation.rs
+++ b/core/src/eval/operation.rs
@@ -1152,10 +1152,7 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                                 env,
                             })
                         } else {
-                            Ok(Closure {
-                                value: NickelValue::enum_tag(tag, pos_op_inh),
-                                env,
-                            })
+                            Ok(Closure { value, env })
                         }
                     }
                     _ => Ok(Closure { value, env }),


core/src/eval/operation.rs
```diff
@@ -1156,10 +1152,7 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                                 env,
                             })
                         } else {
-                            Ok(Closure {
-                                value: NickelValue::enum_tag(tag, pos_op_inh),
-                                env,
-                            })
+                            Ok(Closure { value, env })
```
yannham, line 1155: This is unrelated but avoid allocation when forcing a simple enum tag (the previous code would always allocate a new copy).

core/src/eval/operation.rs
```diff
@@ -1131,8 +1129,6 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
 
                         Ok(seq_terms(terms, pos_op, cont).into())
                     }
-                    // For an enum variant, `force x` is simply equivalent to `deep_seq x x`, as
-                    // there's no lazy pending contract to apply.
```
yannham, line 1135: I believe the comment is just not true: `force 'Foo x` is `'Foo (force x)`, which is different from `deep_seq`ing if for example `x` contains further records or arrays with lazy contracts.

PR #2463, opened 2025-12-09
chore: use the strict provenance API for `NickelValue`

yannham:
This PR uses the strict provenance API introduced in the recent versions of Rust for the implementation of `NickelValue`, instead of doing naked integer to pointer conversions. This supposedly enables more optimizations, make it easier to ensure the code isn't UB (and won't be in the future), and will be the right way to do things ™️ going forward.

## MSRV

Strict provenance API has been stabilized quite recently. This PR requires bumping the MSRV to 1.89. `1.89` dates back from August 2025, which is relatively recent. I personally have no issue with this (we shouldn't tie our hands), but it's open for debate.

Any diff outside of `eval::value` is mostly clippy noise (appeasing the new "break nested if" lint).

### DIFF
--- a/core/src/eval/value/mod.rs
+++ b/core/src/eval/value/mod.rs
@@ -50,7 +50,7 @@ pub struct TagMismatchError;
 /// details.
 #[derive(Eq)]
 pub struct NickelValue {
-    // According to [String provenance
+    // According to [Strict provenance
     // rules](https://doc.rust-lang.org/std/ptr/index.html#using-strict-provenance), as `data` can
     // contain a pointer and thus require provenance information, it must either be itself a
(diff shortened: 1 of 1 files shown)


core/src/eval/value/mod.rs
```diff
@@ -49,7 +50,25 @@ pub struct TagMismatchError;
 /// details.
 #[derive(Eq)]
 pub struct NickelValue {
-    data: usize,
+    // According to [String provenance
```
jneem: Strict provenance, maybe?

yannham:
```suggestion
    // According to [Strict provenance
```

PR #2465, opened 2025-12-09
Try out the gh runners again

jneem:
This switches us back to the free github runners for CI.

The free space thing is a little bit puzzling and I haven't looked into it too deeply. The CI doesn't seem to actually use much space: in [this run](https://github.com/tweag/nickel/actions/runs/20307736030/job/58329821525), `/nix` only uses about 5GB. Maybe the reason we were running out of space is just that `/nix` gets put by default on a partition with very little free space?

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -41,7 +41,7 @@ jobs:
       with:
         hatchet-protocol: 'holster'
       # Currently broken on macos:
-      # https://github.com/haskell/hackage-server/issues/1425
+      # https://github.com/wimpysworld/nothing-but-nix/issues/39
       if: runner.os == 'Linux'
 
     - name: Install Nix


yannham:
The CI is suspiciously fast, and I don't see any `cargo test` output. Or did it just hit the cache perfectly? I would still expect tests to be run

jneem:
It's definitely a little fishy, but I do see some store paths being copied over from tweag.cachix. And I think we set up the self-hosted CI to also push to the same cachix account. So it isn't completely unreasonable that it's just totally cached. I'll make a silly small change to test, though.

jneem:
`Unhandled exception. System.IO.IOException: No space left on device` :disappointed:

yannham:
Yes, I think this is why we started migrating some of the jobs to self-hosted runners (plus the speed). I think we used something like https://github.com/easimon/maximize-build-space, but I don't know if it was sufficient in the end. If it's not, I suppose we'd have to find a different solution. I feel like `crane` isn't serving us entirely well. Maybe we should take a look at one-derivation-per-crate tools again, if that might help, or try to break the tests and run them on different runners? Though I fear most of the space is common derivation like the Rust toolchain, which is hard to avoid in the general case...

.github/workflows/continuous-integration.yml
```diff
@@ -34,27 +34,36 @@ jobs:
...
-        cachix_path="$(nix build --print-out-paths --inputs-from . nixpkgs#cachix.bin)"
-        echo "cachixBin=${cachix_path}/bin/cachix" >> "$GITHUB_OUTPUT"
+    - name: Free disk space
+      # This hash is a recent main version, because v6 (the latest release as of
+      # this comment) doesn't work with the cachix install action.
+      uses: wimpysworld/nothing-but-nix@10c936d9e46521bf923f75458e0cbd4fa309300d
+      with:
+        hatchet-protocol: 'holster'
+      # Currently broken on macos:
+      # https://github.com/haskell/hackage-server/issues/1425
```
yannham: (upstream instead of downstream issue)

```suggestion
      # https://github.com/wimpysworld/nothing-but-nix/issues/39
```

yannham (approved):

PR #2466, opened 2025-12-09
Buffer output in `nickel export`

jneem:
While benchmarking, I noticed that `nickel export` was writing to an unbuffered file. Adding buffering gives a substantial improvement for large outputs: on my system,

```shell
echo "std.array.generate std.function.id 1000000" | nickel export -o test.json
```

went from 4.5 seconds before this change to 800ms after. Obviously, most real programs will have a different ratio of computation to i/o...

### DIFF
--- a/cli/src/export.rs
+++ b/cli/src/export.rs
@@ -1,8 +1,8 @@
-use std::{fs, io::Write, path::PathBuf};
+use std::{fs, path::PathBuf};
 
 use nickel_lang_core::{
     error::{Error, IOError},
-    eval::cache::lazy::CBNCache,
+    eval::{cache::lazy::CBNCache, value::NickelValue},
     program::Program,
     serialize::{self, ExportFormat},
 };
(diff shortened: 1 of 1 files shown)


cli/src/export.rs
```diff
@@ -34,30 +34,37 @@ impl ExportCommand {
...
+            let out = std::io::BufWriter::new(fs::File::create(file).map_err(IOError::from)?);
+            self.export_to(program, &rt, out)?;
         } else {
-            serialize::to_writer(std::io::stdout(), self.format, &rt)
-                .map_err(|error| error.with_pos_table(program.pos_table().clone()))?;
-
-            if trailing_newline {
-                println!();
-            }
+            let out = std::io::BufWriter::new(std::io::stdout().lock());
```
yannham, line 44: `std::io::stdout()` is already buffered, although it's only line-buffered. I have no idea if it's beneficial or not to add a second level of buffering?

jneem: For me, it was less beneficial than the file i/o but still noticeable: I measured 2 seconds before and 0.8 seconds after.

PR #2474, opened 2025-12-15
Fix windows release artifacts by using right LLVM version

jneem:
The recent rust version bump broke the windows release artifacts because there's a step that needs a matching LLVM toolchain. I'm not sure how to best automate this...

Tested here: https://github.com/tweag/nickel/actions/runs/20252047769

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -7,6 +7,11 @@ on:
       release_tag:
         description: "The release tag to target"
 
+env:
+  # We need to download the right LLVM tools on windows, and the version needs to match the one
+  # that rust uses. I don't know how to automate this...
+  LLVM_LINK: https://github.com/llvm/llvm-project/releases/download/llvmorg-21.1.2/clang+llvm-21.1.2-x86_64-pc-windows-msvc.tar.xz
+
 permissions:
   id-token: write
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/release-artifacts.yaml
```diff
@@ -7,6 +7,11 @@ on:
       release_tag:
         description: "The release tag to target"
 
+env:
+  # We need to download the right LLVM tools on windows, and the version needs to match the one
+  # that rust uses. I don't know how to automate this...
+  LLVM_LINK: https://github.com/llvm/llvm-project/releases/download/llvmorg-20.1.8/clang+llvm-20.1.8-x86_64-pc-windows-msvc.tar.xz -o llvm.tar.xz
```
yannham: It seems the version of LLVM is the same as it was before in the workflow, or am I missing something?

jneem: Sorry, not sure what went wrong with the version there. I also forgot that for testing I needed to move the `test-release` tag to the commit that I actually wanted to test. So I was testing with a commit that was before bumping the rust version, which is why it was succeeding without the LLVM version bump. [This](https://github.com/tweag/nickel/actions/runs/20275390671) one is the right job now, and I checked that linux artifact hashes built by it match the released ones.

yannham (approved):

PR #2475, opened 2025-12-16
chore: improve and update README (wording and roadmap)

yannham:
This is a pass on the README, mixing both cosmetic changes and information update.

## Content

- There's been a wordplay around a nickel and "the cheap configuration language" for a long time. I've always found that bothering because "cheap" has a clear negative connotation as well. This PR changes the catchphrase to something else (partly taken from the website).
- The very first description of Nickel's important traits is a bit heavy and convoluted, IMHO. I've tried to reword those parts to make them simpler, clearer and more efficient.
- Added a bunch of missing links to satellite projects (vim and emacs plugins, e.g.).
- Make using the `nixpkgs` version of Nickel the default (using stable instead of latest `master`, plus the fact that the build from `nixpkgs` should be much faster, if not cached)
- Updated the roadmap. For now there's no Nix interop work in the pipe, and 1.15.1 has done quite a bit of general performance work. Personally I'd like to tackle incremental evaluation next, so I put that top of the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,7 @@ Nickel's salient traits are:
 - **Design by contract**: *contracts* are a principled approach to validation.
     In a Nickel configuration, contracts act like *schemas*. You can write your
     own, mix them with existing contracts, and enforce that your configuration
-    is correct by sprinkling simple type-like annotations. The Nickel
-    interpreter also uses contracts under the hood to provide meaningful error
-    messages when statically typed functions are misused from within dynamically
-    typed code.
+    is correct by sprinkling simple type-like annotations.
 
 The motto guiding Nickel's design is:
 > Great defaults, design for extensibility


README.md
```diff
@@ -4,7 +4,8 @@
 [![Website](https://img.shields.io/website-up-down-green-red/http/cv.lbesson.qc.to.svg)](https://nickel-lang.org)
 [![Discord](https://img.shields.io/badge/Discord-100000?style=flat&logo=Discord&logoColor=C3C3C3&labelColor=4179DA&color=010101)][discord]
 
-Nickel is the cheap configuration language.
+Nickel is the universal configuration language. Nickel is configuration
+templating, but done right. Modular, correct and boilerplate-free.
```
aspiwack: Boilerplate-free is a big claim. Maybe some would argue that Nickel has more boilerplate than Yaml and Toml. That would be fair, wouldn't it?

README.md
```diff
@@ -16,34 +17,37 @@ Nickel's salient traits are:
...
+    Nickel features a *(sound) gradual type system*: it has types, but you get
+    to choose when you want to use them or not. You can statically check complex
+    functions, but use more flexible runtime validation for configuration data.
+- **Design by contract**: *contracts* are a principled approach to validation.
+    In a Nickel configuration, contracts act like *schemas*. You can write your
+    own, mix them with existing contracts, and enforce that your configuration
+    is correct by sprinkling simple type-like annotations. The Nickel
+    interpreter also uses contracts under the hood to provide meaningful error
+    messages when statically typed functions are misused from within dynamically
+    typed code.
```
aspiwack: My feeling is that this last paragraph isn't very clear.

README.md
yannham: I guess it depends on  the POV. To me boiler-plate free is vaguely related to expressiveness, so if you need to copy paste large chunks of nearly-identical code in YAML or TOML you could argue that it's boilerplate. But maybe this is not the usual definition. So maybe "expressive", "succinct", or something to mean the converse of verbose?

README.md
yannham: Point taken. I kept it from the previous version, but maybe it just doesn't fit in this very high-level, first-page description.

README.md
aspiwack: I don't think this quite counts as boilerplate. Well, it depends what you mean. The boilerplate is the things you have to write which aren't relevant to your application (`public static void main (String[] args)`, for instance, is boilerplate; or in Nickel, the `{`/`}` and the `,` feel boilerplaty compared to Yaml). So if what you mean is that you have power of abstraction and you don't have to duplicate business logic, I'm not sure it qualifies as boilerplate (I don't think, at any rate, that the word “boilerplate” will be taken as meaning that).

But maybe you actually mean boilerplate (like `jobs: { matrix: { vars: {EXE: …`…)). I don't actually know.

README.md
aspiwack: It does seem rather different in style than the other paragraphs, indeed.

README.md
yannham: Fair enough. I went with "concise", mostly because I find that it sounds much better with three adjectives, so I have to find another one to replace boilerplate-free :stuck_out_tongue: another candidate was "reusable", but I feel that "modular" already covers this aspect

README.md
yannham: I've got rid of it.

aspiwack:
No more complaint on my end.

PR #2484, opened 2026-01-07
feat(experimental): skeleton of an incremental evaluator

yannham:
This PR is an ongoing experiment for the implementation of a basic incremental interpreter. I have a pretty good idea of the general design; the goal here is to experiment and go into the technical details around hashing, storage, persistence, etc.

## Motivation

Incremental evaluation is an avenue we have wanted to explore for a long time in order to improve evaluation time. If done right, it can be a game changer on large and very large code bases. The fundamental observation is that often the changes between two versions of a codebase are very small (compared to the size of the codebase), and thus that the vast majority of the program doesn't change. Incremental evaluation aims at reusing the results from the previous evaluation to avoid recomputing everything from scratch.

## Content

This PR doesn't implements a incremental evaluator yet. It just wires up all the basic parts needed to do so. All the content is feature-gated (because the current approach requires to increase the size of thunks, which might penalize non-incremental runs a bit), plus an additional toggle flag to actually enable it. Not enabling the incremental evaluator by default makes it much simpler to handle tests, CI, etc: even with the feature enabled (which happens in CI because of `--all-features`), the tests still run the normal evaluator. We can add incremental-specific tests step by step.

Content:

- add a few fields to thunks in order to store semantic hashes
- implement the computation of the semantic hash of a thunk with dependencies
- implement a new instance of `crate::eval::cache::Cache`, under `incremental_ng`, which mostly acts like the normal cache but try to fetch or store specifically thunks that have a semantic hash attached
- makes the VM attach semantic hashes to thunks of interest, when storing a bound expression (let-binding), evaluating a recursive record, or closurizing an array or a record. Those should be almost all of the places where we allocate thunks for the first time, coming straight from the original source.
- add the `--incremental` flag to the CLI which toggles incremental evaluation.

The most important part for the actual incremental evaluator are left unimplemented(), so that running `nickel eval --incremental` will panic as of today. This idea is to experiment with different schemes from this PR:

- the part that decides which thunks are worth memoizing/fetching from last evaluation
- the computation of the semantic hash
- the persistence of the cache on-disk


core/src/eval/cache/lazy.rs
```diff
@@ -706,8 +690,100 @@ impl ThunkUpdateFrame {
...
+        /// The hash of this thunk can't be computed, and will never be, at least during the current
+        /// run.
+        ///
+        /// This happens for thunks that aren't deemed worth of being versioned by the incremental
+        /// evaluator, and aren't simple enough to compute their hash structurally (the hash of
+        /// constants is typically always defined). This hash value is contagious: computing the hash
+        /// of any (direct or indirect) reverse dependency of this thunk will lead to an undefined hash
+        /// as well.
+        Undefined,
+        /// The hash of this thunk hasn't been computed yet. Note that an [Self::Unknwon] hash might
```
jneem:
```suggestion
        /// The hash of this thunk hasn't been computed yet. Note that an [Self::Unknown] hash might
```

core/src/eval/mod.rs
```diff
@@ -140,6 +145,11 @@ pub struct VmContext<R: ImportResolver, C: Cache> {
     pub cache: C,
     /// The position table, mapping position indices to spans.
     pub pos_table: PosTable,
+    /// Wether incremental evaluation is enabled or not. Incremental evaluation is currently
```
jneem:
```suggestion
    /// Whether incremental evaluation is enabled or not. Incremental evaluation is currently
```

jneem (approved):
Is the plan that we'll have multiple impls of `Cache`, but the value representation will be the same for all of them? So in principle you could evaluate something with a `Program<CBNCache>` and then take the result and put it into something that you then evaluate with a `Program<IncCache>?`

yannham:
> Is the plan that we'll have multiple impls of Cache, but the value representation will be the same for all of them? So in principle you could evaluate something with a Program<CBNCache> and then take the result and put it into something that you then evaluate with a Program<IncCache>?

Yeah, for now this is the direction. Though to be fair this is just for ease of experimentation, it doesn't have to be this way moving forward: we could also make two different versions of `Thunk`, one for each cache, to keep non-incremental evaluation performant. The current situation is acceptable IMHO because: 1. nobody will use `--features incremental-experimental` beside the CI, and 2. even for the CI it adds a few words to thunk data. Thunk data _is_ quite perf sensitive, but I can't see it having a dramatic effect either.

core/src/eval/cache/incremental_ng.rs
```diff
@@ -0,0 +1,184 @@
...
+    lazy::{CBNCache, Thunk, ThunkState},
+};
+
+use crate::eval::semantic_hash::SemanticHash;
+use std::{
+    collections::{HashMap, hash_map::Entry},
+    io,
+};
+
+/// A thunk coming from a preivous evaluation. Can be lazily loaded as a normal thunk if needed.
```
yannham:
```suggestion
/// A thunk coming from a previous evaluation. Can be lazily loaded as a normal thunk if needed.
```

core/src/eval/cache/mod.rs
```diff
@@ -90,4 +86,15 @@ pub trait Cache: Clone {
...
         idx: &mut CacheIndex,
     ) -> Result<Self::UpdateIndex, BlackholedError>;
+
+    /// Attach a [CUI][crate::eval::cui] to the given index, thereby marking it as being of
+    /// interest for incremental evaluation.
+    ///
+    /// The default implementation does nothing, which is the behavior for caches that don't
+    /// support incremental evaluation, such as the default CBN cache. In practice we never expect
+    /// to actually call [Self::attach_cui] on such caches: a VM with incremental evaluation
+    /// enabled should always use a cache that supports incremental evaluation in practice, but we
```
yannham:
```suggestion
    /// enabled should always use a cache that supports incremental evaluation, but we
```

core/src/eval/mod.rs
```diff
@@ -140,6 +145,11 @@ pub struct VmContext<R: ImportResolver, C: Cache> {
     pub cache: C,
     /// The position table, mapping position indices to spans.
     pub pos_table: PosTable,
+    /// Whether incremental evaluation is enabled or not. Incremental evaluation is currently
+    /// feature-gated, but even when compiled by the feature, it needs to be explicitly enabled
```
yannham:
```suggestion
    /// feature-gated, but even when compiled with the feature, it needs to be explicitly enabled
```

PR #2486, opened 2026-01-08
Keep the closure environment in eval_permissive

### LINKED ISSUE #2483: Missing LSP diagnostic for contract violation inside an enum
The following input doesn't cause any LSP diagnostics, even though it errors on eval. Removing the enums (i.e., just using plain records) causes the error to appear.

```nickel
'Carnitas { pineapple = "1" } | [| 'Carnitas { pineapple | Number }, 'Fish { avocado | Number } |]
```

jneem:
When recursing into a record, array, or enum, `eval_permissive` was losing the closure's environment. I'm a little surprised this worked at all, but it especially broke enum contracts, which rely on generated variables in their environment.

Fixes #2483.

The reason this caused #2483 was because the enum contract was failing on an undefined generated variable instead of with a contract violation. This didn't generate any LSP diagnostics at all because we only report LSP diagnostics if they have a location.

### DIFF
--- a/core/src/eval/value/mod.rs
+++ b/core/src/eval/value/mod.rs
@@ -94,7 +94,9 @@ impl PartialEq for NickelValue {
             (ValueContentRef::String(string1), ValueContentRef::String(string2)) => {
                 string1 == string2
             }
-            (ValueContentRef::Thunk(thunk1), ValueContentRef::Thunk(thunk2)) => thunk1 == thunk2,
+            (ValueContentRef::Thunk(thunk1), ValueContentRef::Thunk(thunk2)) => {
+                *thunk1.borrow() == *thunk2.borrow()
+            }
             (ValueContentRef::Term(term1), ValueContentRef::Term(term2)) => term1 == term2,
             (ValueContentRef::Label(label1), ValueContentRef::Label(label2)) => label1 == label2,
             (
@@ -129,7 +131,7 @@ impl fmt::Debug for NickelValue {
             ValueContentRef::Array(container) => write!(f, "Array({container:?})"),
             ValueContentRef::Record(container) => write!(f, "Record({container:?})"),
             ValueContentRef::String(string) => write!(f, "String({string})"),
-            ValueContentRef::Thunk(thunk) => write!(f, "Thunk({thunk:?})"),
+            ValueContentRef::Thunk(thunk) => write!(f, "Thunk({:?})", thunk.borrow()),
             ValueContentRef::Term(term) => write!(f, "Term({term:?})"),
             ValueContentRef::Label(label) => write!(f, "Label({label:?})"),
             ValueContentRef::EnumVariant(enum_variant_data) => {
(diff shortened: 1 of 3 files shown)


core/src/eval/value/mod.rs
```diff
@@ -129,7 +131,7 @@ impl fmt::Debug for NickelValue {
             ValueContentRef::Array(container) => write!(f, "Array({container:?})"),
             ValueContentRef::Record(container) => write!(f, "Record({container:?})"),
             ValueContentRef::String(string) => write!(f, "String({string})"),
-            ValueContentRef::Thunk(thunk) => write!(f, "Thunk({thunk:?})"),
+            ValueContentRef::Thunk(thunk) => write!(f, "Thunk({:?})", thunk.borrow()),
```
jneem, line 134: Unrelated, but while debugging I noticed that debug-printing a thunk was recursing indefinitely. I guess this was a side-effect of removing the extra thunk indirection. I think there's a similar issue in `PartialEq`, although I didn't actually run into it.

yannham (commented):
I suppose it's not been an issue so far because containers such as arrays and records, are closurized, so evaluating their components recursively never needs the outer environment. I wonder if this was the meaning of the comment L1217:

>  // After eval_closure, all the array elements  are
>  // closurized already, so we don't need to do any tracking
>  // of the env.

In fact, enum variants are also containers and should be closurized...at least they are when they get out of `compat`. I'm not sure what to do here:

1. Consider the bug is that some enum variant arg, somewhere, is produced unclosurized (must be clearly `force` or `deep_seq` or maybe contract application).  Currently the only WHNFs that need the environment are functions and function-like expressions, I believe.
2. Consider that we should always propagate the environment. It doesn't hurt, at least correctness-wise, but it might keep more values alive than necessary.

jneem:
I dug into this a bit more. In

```
'Foo { bar = 1 } | [| 'Foo { bar | String } |]
```

the top-level term is a closure containing a `Term::Annotated`. When evaluating the `Term::Annotated`, the evaluator generates an enum contract. and applies it with a `%contract/apply%`. That generated contract looks like

```
($enum
    (fun
      %214
      %213
      =>
      let %216 = %213 in
      if (%enum/is_variant%  %216) && ((%enum/get_tag%  %216) == 'Foo) then
          let %218 = %enum/get_arg%  %216 in
          let %217 = %218 in
            let %215 = %217 in
              'Ok ('Foo ((%contract/apply% { bar | String, } %214) %215))
        else
          let  in $enum_fail %214))
```

So it's the `'Foo (..)` inside of the runtime-generated contract that's missing it's closure.

The latest version of this PR adds `Term::Closurize` wrappers in the enum pattern compilation. Maybe it also makes sense to remove `Term::Annotated` and handle it during Ast lowering? I actually didn't realize that we were still doing pattern compilation at runtime...

yannham:
(Another possibility would be to pre-compile types as contracts, and store it alongside a `Type` value. I think we already cache the generated contract; we could just pre-populate it)

PR #2490, opened 2026-01-16
Improve the CI, part 1

jneem:
This improves the CI, and hopefully stops it from timing out.

First of all, for some reason we were sometimes running for 6 hours and timing out. I don't really know the cause of this -- the job would just seem to get stuck right at the end. I _suspect_ it had something to do with the nothing-but-nix action we were using to free up some space, but I don't really have evidence. Anyway, I replaced that action with a much simpler scheme: put /nix in /mnt, where there's more free space.

I noticed that `cargoArtifactsDeps` wasn't really working, because apparently crane's `buildDepsOnly` doesn't support `cargoArtifacts`: you can't "chain" multiple runs of `buildDepsOnly`. I replaced this with a single `buildDepsOnly` that has a custom build step.

I noticed that the cachix pushes were no longer working, even though we still have the secret. Maybe it only works from within the tweag organization? Anyway, I replaced it with `nix-community/cache-nix-action`. Also, Determinate Systems' magic-nix-cache has been revived. But I tried it, and it didn't cache the things I wanted it to...

While looking at what `nix flake check` was spending time on, I found that it was doing release builds of nickel. It turns out that both the stdlib tests and the nickel formatter were building a release build. I changed the stdlib tests to use a debug build, and I wrote a bare-bones formatter that's relatively fast to build. This also makes the dev shell a bit nicer to use, because it was also spending time rebuilding nickel a lot.

There's still plenty that could be improved. For example, a lot of the stuff (like checking `nickel-lang` and `nickel-lang-pkg`) could be done in parallel. But I think this is a step in the right direction.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -57,27 +57,28 @@ jobs:
       with:
         primary-key: ci-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
         restore-prefixes-first-match: ci-${{ runner.os }}-
-        # TODO: how to prevent the most recent thing from getting gc-ed?
-        #gc-max-store-size-linux: 10G
-        # TODO: we want to purge old caches, but this purges the
-        # cache we just created...
         purge: true
+        # This was the only way I could figure out to stop cache-nix-action from
+        # purging the cache that it just created. The "purge: true" above makes
+        # it purge the old cache just before trying to save a new one. This line
+        # makes it not purge anything after saving.
         purge-prefixes: non-existent-prefix-
 
 
     - name: Run all checks
       run: |
+        # This is basically `nix flake check`, except that it creates extra GC
+        # roots. Then we run a GC to get rid of anything that's unlikely to be
+        # useful to cache.
         SYSTEM=$(nix eval --impure --raw --expr builtins.currentSystem)
         nix eval .#checks.${SYSTEM} --apply builtins.attrNames --json \
           | jq -r '.[]' \
           | xargs -I {} nix build .#checks.${SYSTEM}.{} --out-link .check-{} --print-build-logs
-        ls /nix/store/*nickel-lang*
         nix store gc
-        ls /nix/store/*nickel-lang*
 
-    - name: Check disk space
-      run: df -h
-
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -34,15 +31,16 @@ jobs:
...
+    # Last I checked, the ubuntu runners have a root filesystem with 23G free
+    # and a /mnt with 66G free. Putting /nix in /mnt seems to stop us from
+    # running out of disk space. (There are also actions available that do
+    # more drastic things.)
+    - name: Loop mount
       if: runner.os == 'Linux'
+      run: |
+        sudo mkdir /mnt/nix
+        sudo mkdir /nix
+        sudo mount --bind /mnt/nix /nix
```
yannham, line 43: That is...satisfyingly simple :upside_down_face:

flake.nix
```diff
@@ -270,9 +269,16 @@
           cargoBuildExtraArgs = "--frozen --offline";
 
           # Build *just* the cargo dependencies, so we can reuse all of that work (e.g. via cachix) when running in CI
-          mkCargoArtifactsDeps = { cargoExtraArgs ? "", prevArtifacts ? false }: craneLib.buildDepsOnly ({
+          cargoArtifactsDeps = craneLib.buildDepsOnly {
```
yannham, line 272: I think this was already the philosophy of the previous attempt (although it didn't work), but if I understand correctly, this means that building only the Nickel package from this flake, without any special feature, will require to build Nickel with default features, with all features, and with `package-experimental`? (I'm not saying this is a problem - after all one can always use `nixpkgs` or build from `cargo` if they want speed)

jneem: That's true. It's a trade-off, though, because there's quite a bit of overlap in the dependencies. I don't have the exact numbers handy, but a good fraction of the target dir is shared. Like, having three separate target dirs takes 2x the space, or something like that.

I do want to experiment with separating them, though, because one strategy for improving CI times is to do more in parallel. Clippy and docs could be in parallel, and nickel-lang-pkg and nickel-lang could be in parallel. But we do need to be a little careful of the total cache size, because at some point github starts to prune the cache...

PR #2493, opened 2026-01-17
Allow priority to be set for background evaluation in the LSP

L0r3m1p5um:
This improves the LSP background evaluation by making sure that the most relevant evaluations are done first. In particular, whatever file has just been edited or opened will be the first to have its evaluation diagnostics published. It does this by creating two separate stacks for the background evaluator, the normal stack and a new high priority stack. Whenever it selects a file to evaluate it will select from the high priority stack first and only select from the normal stack if no high priority items are available. I have a project where in the worst case, it can take over 90 seconds from when I edit a file to when I see the eval diagnostics for that file. With this change I see the diagnostics for the current file in a few seconds in the same case.

### DIFF
--- a/lsp/nls/src/background.rs
+++ b/lsp/nls/src/background.rs
@@ -29,6 +29,8 @@ struct Eval {
     contents: Vec<(Url, Arc<str>)>,
     /// The url of the file to evaluate.
     eval: Url,
+    /// Mark the evaluation as high priority.
+    high_priority: bool,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


lsp/nls/src/background.rs
```diff
@@ -195,7 +206,14 @@ impl SupervisorState {
...
 
-            if let Some(eval) = self.eval_stack.pop() {
+            // If there are any priority items, they should be done first.
+            let stack = if !self.priority_eval_stack.is_empty() {
+                &mut self.priority_eval_stack
+            } else {
+                &mut self.eval_stack
+            };
+
+            if let Some(eval) = stack.pop() {
```
yannham, line 216: Couldn't we avoid the conditional assignment, and just do something like `if let Some(eval) = self.priority_eval_stack.pop().or(self.eval_stack.pop())`? It's a bit long, so could also be assigned before like `let next = self...`. I'm just thinking that the conditional assignment and mutable borrow is probably not required. But it's not very important.

yannham:
Just to be sure to understand, what is not going to the high priority eval stack? Because for edition and open, it seems like the high priority is just like the previous stack. So I wonder what other events clutter the stack in between e.g. editions?

jneem:
> Just to be sure to understand, what is not going to the high priority eval stack?

All the files that got invalidated by the file that got changed.

yannham:
Thanks, that makes sense indeed now.

PR #2498, opened 2026-01-27
Recursively closurize the overrides

### LINKED ISSUE #2491: debug_asserts firing on simple program + overrides
Hi, 

consider the following simple program:

```rust
use nickel_lang_core::{
    error::Sink, eval::value::lazy::CBNCache, program::Program, term::MergePriority,
};
use std::io::Cursor;

fn main() {
    let source = r#"
{
    inputs = {
        user | String | default = "defaultuser",
    },
    result = inputs.user,
}
"#;

    let mut program: Program<CBNCache> = Program::new_from_source(
        Cursor::new(source.as_bytes()),
        "foobar",
        std::io::stderr(),
        Sink::default(),
    )
    .unwrap();

    let ovd = program
        .parse_override(
            r#"inputs.user="overridden""#.to_string(),
            MergePriority::Top,
        )
        .unwrap();

    program.add_overrides(vec![ovd]);

    let x = program.eval_full().unwrap();

    println!("{x}");
}

```

In debug mode this panics with:

```
thread 'main' (49337432) panicked at /<USER>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nickel-lang-core-0.16.1/src/eval/fixpoint.rs:56:21:
assertion failed: value.is_constant()
```

In release mode, this passes with:

```nickel
{ inputs = { user | String | force = "overridden", }, result = "overridden", }
```

jneem:
Although we were closurizing the records created by the cli customize mode, we weren't doing so recursively, so nested records weren't handled correctly.

This PR removes the automatic closurization from `Record::build`, because it's the only part of that module that does closurization. Instead, I made it a separate utility function.

Fixes #2491

### DIFF
--- a/cli/tests/snapshot/inputs/customize-mode/closurization.ncl
+++ b/cli/tests/snapshot/inputs/customize-mode/closurization.ncl
@@ -1,0 +1,12 @@
+# capture = 'stderr'
+# command = ['export']
+# extra_args = [
+#  '--',
+#  'inputs.user="me"',
+# ]
+{
+  inputs = {
+    user | String | default = "defaultuser",
+  },
+  result = inputs.user,
+}
--- a/cli/tests/snapshot/snapshots/snapshot__export_stderr_closurization.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__export_stderr_closurization.ncl.snap
@@ -1,0 +1,5 @@
+---
+source: cli/tests/snapshot/main.rs
+expression: err
+---
+
--- a/core/src/term/make/builder.rs
+++ b/core/src/term/make/builder.rs
@@ -230,10 +230,12 @@ where
     let fst = it.next().unwrap();
 
     let content = it.rev().fold(content, |acc, id| {
-        record::Field::from(NickelValue::record_posless(RecordData {
-            fields: [(LocIdent::from(id), acc)].into(),
-            ..Default::default()
-        }))
+        record::Field::from(NickelValue::term_posless(Term::Closurize(
+            NickelValue::record_posless(RecordData {
+                fields: [(LocIdent::from(id), acc)].into(),
+                ..Default::default()
+            }),
+        )))
     });
 
     (fst.into(), content)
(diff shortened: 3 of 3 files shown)


jneem:
I agree that adding closurizations while elaborating field paths would fix #2491, but I think that's just because only records are affected -- anyone building values with `NickelValue::xxx` constructors would still need to manually closurize arrays and enum variants, right?

Right now, there are two paths to build values: Ast -> NickelValue conversions, and manual NickelValue building (I'm including the `NickelValue::xxx` constructors and also the the `make::builder` module in this group). The Ast -> NickelValue conversions are automatically closurized, and so the idea of this PR was to make it so manual NickelValue building is *never* automatically closurized. That way, you can add explicit recursive closurization at the end without accidental double-closurization... but I guess if you combine Ast -> NickelValue conversions with manual NickelValue building then it could be a problem...

yannham:
> but I guess if you combine Ast -> NickelValue conversions with manual NickelValue building then it could be a problem...

Indeed, and this is what happens here in the customize mod, since the value of the overridden field is parsed, so it comes out of `compat` and is already closurized (you don't even have a choice). Not only we double closurize, but also you have to traverse (since we don't bail out) and thus destruct and re-allocate the whole resulting term unconditionally, including the arbitrarily complex values in the fields. On the other hand, adding the missing closurization in `build` only adds the `Closurize` introduced by the record builder, and leaves the responsibility of closurizing the fields' values to the caller (maybe they are already coming from `compat`, maybe they don't if they were built manually).

>  anyone building values with NickelValue::xxx constructors would still need to manually closurize arrays and enum variants, right?

True, but we likely agree that `NickelValue::xxx` are low-level constructors which must mirror exactly the value block you get, and thus shouldn't auto-closurize or do anything fancy. If you need to closurize such values, it's trivial: just slap a `Closurize` in front of it, which is constant time. I don't think the same applies to `build`. One middle-ground would be to add two variants, `build_closurize` and `build_unclosurized`. I believe that you never want `build_unclosurized` in practice, because you never want an unclosurized record and re-closurizing it after the fact is always going to be more expensive that doing it on the fly, potentially linear in the size of the whole combined record, as in the case of overriding here. So even if we had both variants, we should use `build_closurize` here, for pure perf reasons.

I think this entails getting rid of `build_unclosurized` and just rename `build_closurized` to `build` for simplicity, which is the auto-closurize proposal.

jneem:
Sure, that makes sense. I'll make `make::builder` always be closurized.

> the value of the overridden field is parsed, so it comes out of compat and is already closurized

Yes, but it doesn't go into the record: it's parsed into a new input and so the record only contains `ResolvedImport`s.

yannham:
> Yes, but it doesn't go into the record: it's parsed into a new input and so the record only contains ResolvedImports.

Ah, I missed that part, so the traversal and double closurization indeed don't apply in this specific case.

yannham (approved):

PR #2502, opened 2026-02-02
Format nickel converted from other formats

L0r3m1p5um:
Pretty much any time I've used `nickel convert` it has been to store the result to a file, so I think it would make sense to have the output formatted already. This PR implements that.

### DIFF
--- a/cli/src/convert.rs
+++ b/cli/src/convert.rs
@@ -113,7 +113,18 @@ impl ConvertCommand {
 
         let ast = ast.map_err(|error| Error::Program { files, error })?;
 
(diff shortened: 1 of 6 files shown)


aspiwack:
Maybe this is worth a flag `--no-formatting` or `--formatted=no` or such for those who really need to prioritise speed over style. I don't why anybody would end up in this situation, but it pays to be plan for this sort of things.

yannham:
I think @aspiwack's remark is fair, but it can be done in a follow-up step, so let's merge in the meantime.

PR #2505, opened 2026-02-03
Better localization of `std.fail_with` errors

yannham:
As noted in https://github.com/nickel-lang/nickel/discussions/2291#discussioncomment-15527124, the error message produced by calling to `fail_with` can't be traced back to the `fail_with` call, which makes them quite hard to diagnose (best luck is to grep for the error message, if unique). Since we're using the contract system, we can actually re-use the reporting infrastructure.

This PR implements an imperfect, but simple trick to get more localized error message, by blaming the provided message as if it was the value making the contract fail. The error message isn't entirely to the point, but I think it's a better situation than no position at all, and this PR adds a note to help interpret it.

This is compatible with a more involved version in the future, for example with a specific core error for `fail_with` (instead of using contracts), or enriching blame errors with `fail_with`-specific data.

cc @L0r3m1p5um


core/stdlib/std.ncl
```diff
@@ -5323,5 +5323,19 @@
...
       ```
     "%
-    = fun msg => null | FailWith msg,
+    =
+      # Blaming the message argument is a cheap way to get `fail_with` errors to
+      # show a meaningful location, instead of pointing into the stdlib only
+      let Blame =
+        %contract/custom% (fun _label msg =>
+          'Error {
+            message = msg,
```
yannham, line 5332: Interestingly, using `fun _label message => ... { include message, ...}` led to bad error messages (basically, the contract error would print `msg` - so the one below in the function - as the "evaluated" version of the error message, which would thus point into the stdlib and not at the call site). This is unexpected but is another investigation for another time.

jneem: Independently of this, I've noticed that the "evaluated" part of the error message often feels redundant. I bet we can have a decent heuristic for hiding it.

yannham: I think we currently only check if the span is different from the original span of the argument. The issue is that evaluation often loses the position information, even if there's not much evaluation going on. A simple improvement that I can think of would be to avoid showing it if we know that the original argument is already a WHNF; it's likely that the evaluated part won't add much information in this case.

PR #2510, opened 2026-02-11
Performance improvements for language server

L0r3m1p5um:
This makes some performance improvements to `nls`. In particular:
- Diagnostics are no longer published synchronously to document change. Instead tasks will be added to run diagnostics when the main thread is available.
- The main loop prioritizes work better. In particular requests and document changes are prioritized above issuing diagnostics, and diagnostics for open files are prioritized over diagnostics for other files.
- The server will do less obsolete work, in particular with respect to diagnostics. If there are multiple document changes queued, they can all be applied before the server begins publishing new diagnostics.
- When publishing diagnostics, the main loop will handle one file at a time. So it spends less time blocked on long operations, and can be more responsive if higher priority requests come in.

Most of this is accomplished by adding the `task_queue` module. Any work that the LSP main loop will need to do is dumped into `TaskQueue` and it will handle prioritizing and deduplicating work.

Additionally this should make it pretty easy to add request cancellation, at least in the case where a request is still queued. I plan to implement that as a follow up.

### DIFF
--- a/lsp/nls/src/task_queue.rs
+++ b/lsp/nls/src/task_queue.rs
@@ -65,7 +65,7 @@ pub struct TaskQueue {
     /// for open files.
     open_files: HashSet<Url>,
     /// The set of files that need updated diagnostics published for them.
-    diagnostics: HashSet<Url>,
+    stale_diagnostics: HashSet<Url>,
 }
 
 impl TaskQueue {
@@ -73,7 +73,7 @@ impl TaskQueue {
         TaskQueue {
             request_or_sync: VecDeque::new(),
             open_files: HashSet::new(),
-            diagnostics: HashSet::new(),
+            stale_diagnostics: HashSet::new(),
         }
     }
 
@@ -97,7 +97,7 @@ impl TaskQueue {
 
     /// Add a task to publish updated diagnostics on a file.
     pub fn add_diagnostics_task(&mut self, uri: Url) {
-        self.diagnostics.insert(uri);
+        self.stale_diagnostics.insert(uri);
     }
 
     /// Takes an LSP notification and adds a task if needed.
@@ -133,16 +133,16 @@ impl TaskQueue {
         if let Some(uri) = self
             .open_files
             .iter()
-            .find(|file| self.diagnostics.contains(file))
+            .find(|file| self.stale_diagnostics.contains(file))
             .cloned()
         {
-            self.diagnostics.remove(&uri);
+            self.stale_diagnostics.remove(&uri);
             Some(Task::Diagnostics(DiagnosticsRequest {
                 uri,
                 priority: Priority::High,
             }))
-        } else if let Some(uri) = self.diagnostics.iter().next().cloned() {
-            self.diagnostics.remove(&uri);
+        } else if let Some(uri) = self.stale_diagnostics.iter().next().cloned() {
+            self.stale_diagnostics.remove(&uri);
             Some(Task::Diagnostics(DiagnosticsRequest {
                 uri,
                 priority: Priority::Normal,
@@ -198,7 +198,7 @@ impl TaskQueue {
                         // The eval command publishes the new diagnostics for a file
                         // synchronously, so if there were a pending update to the diagnostics for a
                         // file, we can remove it.
-                        self.diagnostics.remove(&uri);
+                        self.stale_diagnostics.remove(&uri);
                         Task::HandleRequest(req)
                     }
                     RequestCategory::TextDocumentUri(uri) => {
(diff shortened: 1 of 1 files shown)


lsp/nls/src/task_queue.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// may often be the case that if the document changes after a request is made then the
+    /// response to the request may no longer be useful, but the LSP spec recommends that the
+    /// server should not make the decision to cancel a request in most cases, instead the client
+    /// should send a cancellation notification if the response is no longer needed.
+    request_or_sync: VecDeque<ReqOrSync>,
+    /// The set of files that are currently open in memory. This is used to prioritize tasks
+    /// for open files.
+    open_files: HashSet<Url>,
+    /// The set of files that need updated diagnostics published for them.
+    diagnostics: HashSet<Url>,
```
jneem: nit: maybe `need_diagnostics` or `stale_diagnostics` is clearer?

lsp/nls/src/task_queue.rs
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    /// Get the next request to handle if there's one pending.
+    fn next_request_or_sync(&mut self) -> Option<Task> {
+        // There's a few types of requests that the queue will handle differently. These categories
+        // will determine that handling.
+        enum RequestCategory {
+            /// ExecuteCommand request for the "eval" command
+            EvalCommand(Url),
+            /// Any request extending extending TextDocumentPositionParams in the LSP spec.
```
jneem:
```suggestion
            /// Any request extending TextDocumentPositionParams in the LSP spec.
```

jneem (approved):
Looks good, thanks!

It can be left for the future, but I'm slightly bothered by the potential re-ordering of document updates and other requests. You've preserved the order within a file but not between files, and so couldn't it be the case that we get a change notification for `dep.ncl` and then a diagnostics request for `main.ncl` (which imports `dep.ncl`), and the diagnostics request goes first and runs on an old version?

It seems a bit tricky, though, as strictly following the LS spec's no-cancellation advice would make it difficult to ever process things out of order. Based on [this](https://github.com/neovim/neovim/issues/30985), it seems like maybe rust-analyzer is applying file changes quickly and cancelling diagnostics requests that are out-of-date?

yannham:
Sorry, I wanted to look at this earlier but haven't find the time. I still have to dive a bit more into the implementation, but something that comes immediately to mind at a first glance, beside re-ordering that @jneem mentions: is there a risk of starvation in practice for some type of requests, since we're unconditionally prioritizing some?

L0r3m1p5um:
> Looks good, thanks!
> 
> It can be left for the future, but I'm slightly bothered by the potential re-ordering of document updates and other requests. You've preserved the order within a file but not between files, and so couldn't it be the case that we get a change notification for `dep.ncl` and then a diagnostics request for `main.ncl` (which imports `dep.ncl`), and the diagnostics request goes first and runs on an old version?
> 
> It seems a bit tricky, though, as strictly following the LS spec's no-cancellation advice would make it difficult to ever process things out of order. Based on [this](https://github.com/neovim/neovim/issues/30985), it seems like maybe rust-analyzer is applying file changes quickly and cancelling diagnostics requests that are out-of-date?

I don't think that any of the reordering that happens should be a problem. Documents updates and actual requests expecting a response go into the same queue, and are processed in the order that they're received. So requests should always be processed on the correct version of the file based on when it was received. The difference is that document changes and the publication of new diagnostics based on the change no longer happen at the same time, and it's only the publication of diagnostics that can be reordered. So the document synchronization task only involves changing the file contents in memory and invalidating the analyses of that file and its reverse dependencies.

The scenario you describe should still publish the correct diagnostics for `main.ncl`. The change notification for `dep.ncl` will change that files content in memory, and then invalidate its own parsing and typechecking as well that for reverse dependencies, including `main.ncl`. So when the diagnostics for `main.ncl` are requested, the in memory contents of `dep.ncl` will already be up to date, and it won't use a cached typechecking result for `dep.ncl` since it's already been invalidated. So even though the diagnostics for `dep.ncl` haven't been published yet it shouldn't be a problem.

I did look into what rust-analyzer is doing, and it's more or less what you've described. Document changes are fast because it's just changing the file content in memory, and it can detect what analyses currently running are now out of date and cancel those. I think it would be a lot more difficult to go that route than what I did here though. It would require the language server to be multithreaded, because even if there are analyses ongoing the main loop still needs to be able to detect that a document change was made in order to cancel the other analyses. It also requires analyses to be cancellable in the first place. I don't think it'd be that big of a problem to make  `nls` multithreaded, but making everything cancellable would be more difficult.

L0r3m1p5um:
> Sorry, I wanted to look at this earlier but haven't find the time. I still have to dive a bit more into the implementation, but something that comes immediately to mind at a first glance, beside re-ordering that @jneem mentions: is there a risk of starvation in practice for some type of requests, since we're unconditionally prioritizing some?

The only thing that can be pushed back in the queue is publication of diagnostics based on change notifications, rather than specific requests for diagnostics. The most likely way I could see any sort of starvation happening is if there are a lot of change notifications coming in, which might delay the publication of new diagnostics, especially for closed files, since open files are prioritized. But I think that's a reasonable way to behave in that scenario.

jneem:
Oh right, I missed the fact that it's only the stale diagnostics being re-ordered, and not the other requests.

> It also requires analyses to be cancellable in the first place. I don't think it'd be that big of a problem to make nls multithreaded, but making everything cancellable would be more difficult.

One way to make this easier is to interpret "cancel" as just "remove it from the queue if it hasn't started yet".

PR #2518, opened 2026-02-17
Apply pending contracts in the public API

### LINKED ISSUE #2514: Missing pending contracts when using the stable API
Containers (arrays and records) lazily accumulate contracts instead of mapping them right away onto elements, for both performance and the merge semantics related reasons. When evaluating an expression deeply, this isn't an issue: those contracts are explicitly generated and applied.

However, the stable API - and thus the C API as well - exposes lazy evaluation, such that you can first evaluate something to a weak head normal form, and then explore selected parts of it, without evaluating everything upfront. When doing so, pending container contracts are simply dropped, which is not what you expect. Worse, given the API is high-level, it doesn't expose any way of fixing this behavior. The pending contracts are invisible to the stable API.

We should automatically apply pending contracts whenever a value is extracted from a container. Since those methods return `Expr`, which isn't borrowed, we can return a new expression generated on the fly without having to break the API.

jneem:
I'm not totally sure about the position handling for contract application in arrays: it seemed to mostly be the case that the contract application for array elements is assigned to the position of the array itself. But there was one place where it was given the position of the array element instead. This PR goes with the former option, since it was the most common.

Fixes #2514

### DIFF
--- a/core/src/eval/mod.rs
+++ b/core/src/eval/mod.rs
@@ -425,16 +425,9 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
 
         // unwrap(): by definition, extract_field_impl(_, _, true) ensure that
         // `field.value.is_some()`
-        let value = field.value.unwrap();
-        let pos_idx = value.pos_idx();
+        let value = field.value_with_pending_contracts().unwrap();
 
-        let value_with_ctr =
-            RuntimeContract::apply_all(value, field.pending_contracts.iter().cloned(), pos_idx);
-
-        let Closure { value, env } = self.eval_closure(Closure {
-            value: value_with_ctr,
-            env,
-        })?;
+        let Closure { value, env } = self.eval_closure(Closure { value, env })?;
 
         Ok(Closure { value, env })
     }
@@ -456,7 +449,7 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
         };
 
         for id in path {
-            let Some(current_value) = field.value else {
+            let Some(current_value) = field.value_with_pending_contracts() else {
                 return self.throw_with_ctxt(EvalErrorKind::MissingFieldDef {
                     id: *prev_id,
                     metadata: field.metadata.into_inner(),
@@ -464,17 +457,12 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                     pos_access: PosIdx::NONE,
                 });
             };
+            prev_pos_idx = current_value.pos_idx();
 
             // We evaluate the fields' value, either to handle the next ident of the
             // path, or to show the value if we are treating the last ident of the path
-
-            prev_pos_idx = current_value.pos_idx();
-
-            let curr_value_with_ctr =
-                RuntimeContract::apply_all(current_value, field.pending_contracts, prev_pos_idx);
-
             let current_evaled = self.eval_closure(Closure {
-                value: curr_value_with_ctr,
+                value: current_value,
                 env,
             })?;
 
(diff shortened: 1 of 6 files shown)


core/src/eval/value/mod.rs
```diff
@@ -1781,6 +1781,18 @@ pub struct ArrayData {
     pub pending_contracts: Vec<RuntimeContract>,
 }
 
+impl ArrayData {
+    /// Iterates over all values in the array, with pending contracts applied.
+    pub fn iter_with_pending_contracts(
+        &self,
+        pos: PosIdx,
+    ) -> impl Iterator<Item = NickelValue> + use<'_> {
+        self.array.iter().cloned().map(move |v| {
```
yannham: nitpick: if you clone `v` inside the closure, would that get rid of the `.cloned()` and the `move` at once? Arguably it doesn't make a huge diff, but it's a bit less ceremony

core/src/eval/value/mod.rs
```diff
@@ -2352,6 +2364,13 @@ impl Container<&ArrayData> {
             .flat_map(|array_data| array_data.array.iter())
     }
 
+    /// Iterates over the elements of the array, with pending contracts applied.
+    pub fn iter_with_pending_contracts(&self, pos: PosIdx) -> impl Iterator<Item = NickelValue> {
+        self.into_opt()
+            .into_iter()
+            .flat_map(move |array_data| array_data.iter_with_pending_contracts(pos))
```
yannham, line 2371: Why do we need the `move` here? Isn't `PosIdx: Copy`?

core/src/eval/mod.rs
```diff
@@ -456,25 +449,20 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
-            let Some(current_value) = field.value else {
+            let Some(current_value) = field.value_with_pending_contracts() else {
                 return self.throw_with_ctxt(EvalErrorKind::MissingFieldDef {
                     id: *prev_id,
                     metadata: field.metadata.into_inner(),
                     pos_record: prev_pos_idx,
                     pos_access: PosIdx::NONE,
                 });
             };
+            prev_pos_idx = current_value.pos_idx();
```
yannham, line 460: I think it doesn't make a difference, but just to be sure: `with_pending_contracts` does preserve the position of the value, right? Since now we're taking the position _after_ contract application.

nickel/src/lib.rs
```diff
@@ -611,21 +621,32 @@ impl<'a> Iterator for RecordIter<'a> {
...
+        self.array().len()
     }
 
     /// Is this array empty?
     pub fn is_empty(&self) -> bool {
-        self.array.is_empty()
+        self.array().is_empty()
     }
 
     /// Returns the element at the requested index, if the index is in-bounds.
```
yannham, line 659: Should we mention now that `get` also applies pending contracts?

nickel/src/lib.rs
```diff
@@ -640,8 +661,15 @@ impl<'a> IntoIterator for Array<'a> {
     type IntoIter = ArrayIter<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
+        let arr = self.array();
+        let pending_contracts = match arr {
+            Container::Empty => &[][..],
```
yannham: Why is the `[..]` needed here?

yannham (approved):

jneem:
Thanks for the review! I didn't manage to get to this yet, and I'll be gone for a week and a half, but I'll finish up and merge when I get back.

core/src/eval/mod.rs
jneem: Right, `field.value_with_pending_contracts()` has the same position as `field.value`

nickel/src/lib.rs
jneem: Because `&[]` gets inferred to be of type `&[_; 0]`, which causes an error about the match arms having different types.  I think it's often not needed because arrays coerce to slices, but I guess in this context type inference can't figure it out. I changed it to `[].as_slice()` because I think it looks less cryptic.

nickel/src/lib.rs
jneem: I've made an attempt at explaining it. I tried to explain in terms of evaluated vs non-evaluated expressions, because I hope that the public API is high-level enough that they don't really need to know how pending contracts are represented...

core/src/eval/value/mod.rs
jneem: Yes, but apparently if I leave off the `move` it still gets captured by reference...

core/src/eval/value/mod.rs
jneem: It turns out I still need the `move` to prevent `pos` from being captured by reference. But I did move the `clone` inside.

yannham: Isn't `PosIdx: Copy`? Interesting, I thought it would capture copy values as `move` by default...

PR #2520, opened 2026-02-19
Allow unicode escape sequences in string literals

### LINKED ISSUE #2516: Support Unicode escape sequences in string literals
I'm using Nickel to manage my dotfiles. When migrating strings that contain special Unicode code points (e.g. [Starship](https://starship.rs/config/) prompt symbols, Powerline glyphs), I have to embed the literal character directly:

```nickel
{
          rust = { symbol = "" }, # \ue7a8 (Nerd Font nf-dev-rust)
}
```

It makes the content unreadable without specialized fonts. It would be nice if Nickel supported Unicode escape sequences in strings.

_Originally posted by @frantic1048 in https://github.com/nickel-lang/nickel/discussions/2511_

L0r3m1p5um:
Closes #2516

This adds support for unicode escape sequences in string literals using a braced syntax like `\u{e7a8}`. It allows for variable length sequences up to 6 characters (anything longer is above the unicode character range).

I also made a quick fix to the documentation where it says that ascii code sequences can match the pattern `\x[a-zA-Z0-9]{2}`, when only hex characters are allowed, so it should be `\x[a-fA-F0-9]{2}`.

One thing I was wondering is if I should make the lexer less strict in what it matches as an escape code so that it can display a more specific error. Right now most errors would just show a generic `invalid escape sequence` message. Or maybe it'd be better for the invalid escape sequence message to show usage information whenever an escape code is known but isn't matched because it's being used incorrectly.

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -1943,6 +1943,9 @@ impl IntoDiagnostics for ParseError {
             ParseError::InvalidAsciiEscapeCode(span) => Diagnostic::error()
                 .with_message("invalid ascii escape code")
                 .with_labels(vec![primary(&span)]),
+            ParseError::InvalidUnicodeEscapeCode(span) => Diagnostic::error()
+                .with_message("invalid unicode escape code")
+                .with_labels(vec![primary(&span)]),
             ParseError::StringDelimiterMismatch {
                 opening_delimiter,
                 closing_delimiter,
--- a/doc/manual/syntax.md
+++ b/doc/manual/syntax.md
@@ -248,15 +248,16 @@ ending.
 A backslash `` \ `` followed by one or more characters is interpreted as an
 escape sequence. The following escape sequences are supported:
 
-|    Escape sequence | Description                                    |
-|:------------------:|:-----------------------------------------------|
-| `\\`               | Backslash `` \ ``                              |
-| `\"`               | Double quote `"`                               |
-| `\n`               | New line                                       |
-| `\t`               | Tab                                            |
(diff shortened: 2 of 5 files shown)


jneem (approved):
Looks good to me, but maybe let's give Yann a chance to take a look.

What sort of improved error messages were you thinking of? I haven't really felt the need for better lexer error messages, because they always seem to be pretty easy to figure out...

L0r3m1p5um:
I just mean that a lot of errors end up just getting a generic "invalid escape sequence" message:
```
error: invalid escape sequence
  ┌─ <stdin>:1:3
  │
1 │ "\u{1066x}"
  │  
  ```
  
  Same thing happens with ascii escape codes:
  ```
  error: invalid escape sequence
  ┌─ <stdin>:1:3
  │
1 │ "\xAZ"
  │   ^
  ```
  
It might be a bit more useful if the messages mentioned that those escape codes exist, but are being used improperly in some way. I don't think it's a big deal though.

yannham (approved):
LGTM!

yannham:
I think we can discuss error message in a follow-up. I tend to agree with @jneem that while not optimal, I haven't found it to be an issue in practice because you still have enough information to figure things out, but that being said that shouldn't be a reason to not make things better, if you are motivated to do so @L0r3m1p5um.

By the way, I've also wanted for some time - now that we have the new AST - to avoid escaping things on the fly during parsing, but do it later (maybe during "compilation"?). Escaping on the fly requires to always defensively allocate a `String`, while we could probably borrow most of strings appearing in a source. But that's another question for another time.

PR #2523, opened 2026-02-21
Request cancellation for language server

L0r3m1p5um:
This implements client cancellation of language server requests via the `$/cancelRequest` notification. There's a limitation on what requests can be canceled. The request has to be waiting in the queue. If handling of the request has already begun, that request will be completed normally even if a cancellation notification is received. The server will never proactively cancel a request, this only cancels requests when requested by the client.

### DIFF
--- a/lsp/nls/src/server.rs
+++ b/lsp/nls/src/server.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, time::Duration};
+use std::collections::HashMap;
 
 use anyhow::Result;
 use crossbeam::select;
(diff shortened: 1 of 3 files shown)


lsp/nls/src/server.rs
```diff
@@ -139,28 +139,60 @@ impl Server {
...
         trace!("Running...");
+        // This tracks if the last loop took a message out of the receiver for the LSP connection.
+        let mut received_msg_last_loop = false;
         loop {
             let never = crossbeam::channel::never();
             let bg = self.background_jobs.receiver().unwrap_or(&never);
 
-            if let Ok(msg) = self.connection.receiver.try_recv() {
+            // `recv_timeout` is used here to avoid a race condition that made request cancellation
+            // ineffective.
```
jneem: Would this issue go away if the channel were slightly bigger than zero-sized? (Like, 1 or 2?) If so, that seems like a nicer solution to me.

And I realize that `lsp-server` doesn't expose a capacity knob, but we could spin up an extra thread that listens to the unbuffered channel and pushes on a bounded-but-buffered channel.

lsp/nls/src/task_queue.rs
```diff
@@ -120,6 +126,14 @@ impl TaskQueue {
                     serde_json::from_value::<DidChangeTextDocumentParams>(notification.params)?;
                 self.add_sync_task(DocumentSync::Change(params));
             }
+            Cancel::METHOD => {
+                let params = serde_json::from_value::<CancelParams>(notification.params)?;
+                let id = match params.id {
+                    NumberOrString::Number(id) => id.into(),
+                    NumberOrString::String(id) => id.into(),
+                };
+                self.cancelled_requests.insert(id);
```
jneem: Did you consider doing the cancellation immediately (i.e. removing the cancelled task from the queue and sending the reply straight away)? I guess that would increase the API surface of `TaskQueue`, but it would also reduce the amount of state that we track...

L0r3m1p5um: Mostly I didn't do that because it didn't fit how the API works very well, since the `queue_message` method just adds a task and doesn't require the caller to do anything specific after that. If we wanted cancellation to be handled right away, that method would have to optionally something that the caller would have to handle immediately. I didn't like that as much as having `next_task` be the only method that has any expectations for the caller to do something after calling it. I guess what I could do is that when a cancellation notification is received, check the queue for matching requests and swap them out with cancellation tasks. Then we wouldn't have to separately track the state of which requests have been cancelled, but the API could stay how it is.

lsp/nls/src/server.rs
L0r3m1p5um: I like that a lot better than what I'm doing here. I'll make that change.

lsp/nls/src/task_queue.rs
L0r3m1p5um: Cancellation is not immediate still but I got rid of the need to maintain a separate `cancelled_requests`

PR #2529, opened 2026-02-28
Add --no-format flag to nickel convert

### LINKED ISSUE #2504: Add an option to eschew formatting in `nickel convert`
Auto-formatting has been added to the output of `nickel convert`. It's been argued that maybe, in some specific cases, one wants to avoid this cost (e.g. temporary batch conversions that aren't read by a human). We could do so by adding an argument to control auto-format, see https://github.com/nickel-lang/nickel/pull/2502#issuecomment-3833084719.

costajohnt:
Closes #2504

Adds a `--no-format` flag to `nickel convert` that skips the auto-formatting step. This is useful for batch conversions or cases where formatted output isn't needed.

## Changes

- Added `--no-format` boolean flag to `ConvertCommand` (clap `#[arg(long)]`)
- When set, the converted AST output is returned as-is without running through the formatter
- Added snapshot tests for `--no-format` covering all input formats (JSON, YAML, TOML, text, Nickel, unknown extension)

## Test plan

- [x] `cargo test -p nickel-cli` passes (all existing + new snapshot tests)
- [x] Manual verification: `nickel convert --no-format` produces unformatted output
- [x] Verified snapshot diffs show the expected difference (e.g. `[1, 2]` formatted vs `[ 1, 2 ]` unformatted)

### DIFF
--- a/cli/src/convert.rs
+++ b/cli/src/convert.rs
@@ -30,6 +30,11 @@ pub struct ConvertCommand {
     /// Output file. Standard output by default
     #[arg(short, long)]
     pub output: Option<PathBuf>,
+
+    /// Skip formatting the output. Useful for batch conversions or when
+    /// formatted output isn't needed.
+    #[arg(long = "no-formatting")]
+    pub no_formatting: bool,
 }
 
 impl ConvertCommand {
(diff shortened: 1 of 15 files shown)


yannham:
Hello, and thanks for your contribution! I think you didn't properly added the inputs for the snapshot tests, or maybe I'm missing something?

L0r3m1p5um:
I think it's using the same inputs as the `convert` tests, but running them once with and once without the new flag.

PR #2601, opened 2026-04-20
Cargo.toml: update wasm-bindgen to 0.2.108, still need to update `flake.nix`

twitchyliquid64:
Context: `wasm-bindgen` version is incompatible with selection of major crates like `reqwest`. See: 

 * https://github.com/topiary/topiary/pull/1214
 * https://github.com/topiary/topiary/issues/1128

Could someone who has a working `nix` + flakes setup run the flake to get the values for `pkgs.fetchCrate { pname = "wasm-bindgen-cli" }.hash` and `pkgs.rustPlatform.fetchCargoVendor { ... }.hash` ? I can update the PR with these.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ lsp-harness = { version = "0.1.0", path = "./lsp/lsp-harness" }
 # if this version changes without the nix output hashes being updated. It's okay
 # to bump the version (for example if not doing so prevents some dependency from
 # building) but flake.nix needs to be kept in sync.
-wasm-bindgen = "=0.2.100"
+wasm-bindgen = "=0.2.108"
 
 anstyle = "1.0.13"
 anyhow = "1.0.97"


jneem:
Unfortunately this will be tricky to coordinate, because we depend (transitively) on `topiary-tree-sitter-facade`, which also pins `wasm-bindgen`...

I think we can break the cycle by unpinning it in `Cargo.toml`, but pinning it in `flake.nix` with something like
```nix
      wasm-bindgen-cli =
        let
          wasmBindgenCargoVersions = ...
          wasmBindgenVersion = ...
          expectedWasmBindgenVersion = "0.2.100";
        in
        if wasmBindgenVersion != expectedWasmBindgenVersion
        then
          abort "The wasm-bindgen version in Cargo.lock isn't ${expectedWasmBindgenVersion}. You'll need to update the hashes in flake.nix"
        else
        pkgs.buildWasmBindgenCli rec {...}
```

That way anyone who consumes nickel via cargo can use an unpinned version, but our CI will check that we're pulling in the `wasm-bindgen-cli` that we expect.

L0r3m1p5um:
Does topiary actually use the `format` feature of Nickel? If it doesn't, then I think building Nickel without that feature would also break the cycle.

L0r3m1p5um:
Never mind, I guess topiary has default features disabled anyway. I didn't realize that cargo would still use Nickel's pinned version in that case.

twitchyliquid64:
Could I request some help getting this over the line? I dont have a working nix setup so might be hard for me to do the flakes surgery. I used to be a big nix guy, but its been a number of years so i expect im quite rusty (ha), also flakes is like a new thing

jneem:
Closing, as this has been finished up in #2603

PR #2605, opened 2026-04-29
Create `Analysis` even when typechecking fails in nls

L0r3m1p5um:
This fixes an issue where an `Analysis` was not getting created for a file in the case that typechecking failed. This meant that certain features would break in the presence of typechecking errors even if they don't depend on the typechecking result, like completions and usage lookups.


yannham (approved):
LGTM. Thanks!

Issue #2600, opened 2026-04-15
FR: Way to inject nickel values to the top-level variable scope when driving `nickel_lang_core::program::Program`


twitchyliquid64:
Heya! In my use of Nickel as a config language, I need to inject constants/values that the nickel can work with.

Some suggestions of possible approaches:

1. Have a method on `Program` that lets me provide `(String, NickelValue)` pairs that will be injected into the environment
2. Lean into `%inmem_src%`, give mutable access to `Files` so you can inject some nickel that can be accesses via an import ([discussed in discord](https://discord.com/channels/1174731094726295632/1179430745727586407/1473385799918686310))

### Current workaround

You have to commit war crimes to make this work today: by writing out a file to disk and then adding it to the import path:

```rust
let generated_lib_dir = tempfile::TempDir::new()?;
std::fs::write(generated_lib_dir.path().join("__injected_config__.ncl"), {
    let mut out = Vec::with_capacity(128);
    out.extend(b"{target = {os = ");
    out.extend(opts.for_target().os().as_nickel_literal());
    out.extend(b",arch = ");
    out.extend(opts.for_target().arch().as_nickel_literal());
    out.extend(b"}}");
    out
})?;

let mut program = Program::new_from_sources(
    [(io::Cursor::new(src.into()), "toplevel")],
    std::io::stderr(),
    NullReporter {},
)?;

program.add_import_paths([generated_lib_dir.path()].iter());
// I can now access in Nickel using `import "__injected_config__.ncl".target`
```

I'm happy to send PRs to take this over the line, if I can get a bit of guidance on the right approach for the project.

jneem:
For some reason I couldn't find it when we discussed on discord, but you can modify the `Files` by imitating how the [repl does it](https://github.com/nickel-lang/nickel/blob/d2a66b23eb8563213da8892384dda944e88dfeb4/core/src/repl/mod.rs#L125). That still leaves the question as to how to *reference* the new file (i.e. some stabilized version of `%inmem_src%).

I think independently of that, it does make sense to allow embedders to customize the initial environment. I'd probably have the API take `Ident` instead of `String`, because I can imagine use-cases where you'd want to generate code with unique (`Ident::fresh`) identifiers.

yannham:
[...]

PR #2606, opened 2026-04-29
Refactor Program to use the builder pattern for construction

twitchyliquid64:
Relates: #2600 

Creates a `ProgramBuilder` type to configure a `Program`, removing the ~11 different new_* methods that existed before.

Rough flow is:

```rust
let mut prog: Program<CacheImpl> = ProgramBuilder::new()
            .add_source(Cursor::new(s.to_owned()), "<test>")
            .with_trace(std::io::stderr())
            .build()
            .unwrap();
```

Spicy's:

 * I opted to erase the `T: Read` on the current `Input<T,S>` with `Box<dyn Read>`, so we didnt have to do the whole turbofish syntax to specify a type for an input that didnt need it.
 * Bikeshedding: should we name it `Builder` so its `program::Builder`? 

Future work:

 * new builder methods/fields to control the initial environmenrt as discussed in #2600

### DIFF
--- a/core/src/program.rs
+++ b/core/src/program.rs
@@ -25,7 +25,7 @@ use crate::{
     cache::*,
     closurize::Closurize as _,
     error::{
-        Error, EvalError, EvalErrorKind, IOError, ParseError, ParseErrors, Reporter,
+        Error, EvalError, EvalErrorKind, IOError, NullReporter, ParseError, ParseErrors, Reporter,
         warning::Warning,
     },
     eval::{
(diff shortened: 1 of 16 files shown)


core/src/program.rs
```diff
@@ -1204,6 +950,371 @@ impl<EC: EvalCache> Program<EC> {
...
+/// - `EC: EvalCache` — same as on [`Program`].
+/// - `R` — the warning reporter type, defaulting to [`NullReporter`].
+/// - `W` — the trace writer type, defaulting to [`io::Sink`].
+///
+/// Defaults can be replaced with [`Self::with_reporter`] and [`Self::with_trace`]; both consume
+/// `self` and return a builder with the new generic parameter.
+///
+/// # Example
+///
+/// ```ignore
```
yannham: Nitpick, but why ignore this example? Shouldn't this compile?

core/src/program.rs
```diff
@@ -1204,6 +950,371 @@ impl<EC: EvalCache> Program<EC> {
...
+    field: FieldPath,
+    contracts: Vec<ProgramContract>,
+    contract_paths: Vec<OsString>,
+    import_paths: Vec<PathBuf>,
+    package_map: Option<PackageMap>,
+    trace: W,
+    reporter: R,
+    // `fn() -> EC` makes the builder invariant in `EC` without requiring `EC: 'static` and
+    // without imposing variance constraints from a phantom owned `EC`.
+    _ec: PhantomData<fn() -> EC>,
```
yannham: Why do we need the builder to be invariant in `EC` (and will really a function make it invariant? I would expect this to still be covariant in general in languages with subtyping, but I don't know for Rust in particular)

core/src/program.rs
```diff
@@ -1204,6 +950,371 @@ impl<EC: EvalCache> Program<EC> {
...
+}
+
+impl<EC: EvalCache> Default for ProgramBuilder<EC> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<EC: EvalCache> ProgramBuilder<EC> {
```
yannham: I wonder if it's really useful use `EC` as a parameter of program builder. It seems nothing actually depends on `EC`? Since we have to use a type annotation anyway because of this free generic parameter, I don't think making `build<EC>` generic instead will change anything for callers - they need a type annotation anyway - but this would probably unclutter the program builder implementation a bit.

core/src/program.rs
```diff
@@ -1204,6 +950,371 @@ impl<EC: EvalCache> Program<EC> {
...
+            contract_paths,
+            import_paths,
+            package_map,
+            trace,
+            reporter,
+            _ec,
+        } = self;
+
+        if inputs.is_empty() {
+            return Err(Error::IOError(IOError(
```
yannham: I'm tempted to introduce a `BuilderError`, since the set of errors that can happen at "build time" is rather small, or so it seems. We can always implement a `From<BuilderError> for Error` and piggy back on `IOError` or `Other` for now, but this error in particular feel a bit different (it's a logical error from the calling code, not a runtime error caused by the environment or an external input).

core/src/program.rs
twitchyliquid64: I wanted to show an example of actually using this, but getting this to run as a doctest would mean i would need to get `config.ncl` stamped on disk and all this other scaffolding so I set it as ignore. My hope was to rely on the explicit tests for correctness.

package/src/manifest.rs
```diff
@@ -190,20 +189,31 @@ impl ManifestFile {
     /// Panics if the path has no parent.
     pub fn from_path(path: impl AsRef<Path>) -> Result<Self, Error> {
         let path = path.as_ref();
-        let prog =
-            Program::new_from_file(path, std::io::stderr(), NullReporter {}).with_path(path)?;
+        // Read the file ourselves so that an open/read failure surfaces as a structured
```
yannham: This is a bit fishy. Does this mean that the behavior also changes for `nickel-lang-core` such that the error doesn't carry the path anymore, or doesn't translate to the same error as before? Ideally we could just do the same as before and directly use the builder's file-based helper.

package/src/manifest.rs
```diff
@@ -190,20 +189,31 @@ impl ManifestFile {
...
-        let prog = Program::new_from_source(
-            std::io::Cursor::new(data),
-            "<in-memory manifest>",
-            std::io::stderr(),
-            NullReporter {},
-        )
-        .without_path()?;
+        let prog = ProgramBuilder::new()
+            .add_source(
+                std::io::Cursor::new(data.to_owned()),
```
yannham, line 203: Seeing all those cursors in tests and in other callsite, I wonder if we shouldn't add a `add_string_source` or similar helper to avoid having to turn a `String` into a `Read` each time.

yannham (commented):
Thanks, this mostly looks much better than before! I left a few questions/remarks

core/src/program.rs
twitchyliquid64: Will give this a go and get back to you!

core/src/program.rs
twitchyliquid64: Hey man you tell me what you need! My imperative was getting those 11 new methods into one builder and minimizing other changes, but if you think a `BuilderError` is a better pattern here than infallible methods lets do it!

yannham: Just to clarify, having only `build` throwing error is totally fine (not sure if that what's you're referring to with "infallible methods"?). Here I'm just talking of using a newly introduced, and more specific error type in the return type of this function, instead of generic `Error`.

core/src/program.rs
yannham: Ah, then you can use `norun` instead of `ignore`, and rustdoc will at least try to compile it. For some reason I thought this was the default (compile and not run), but indeed the default is to actually run the code as well.

core/src/program.rs
twitchyliquid64: Ah, now I understand! Ive made a new `BuilderError` in the same module, with `From` impls for conversion into wrapping error types.

package/src/manifest.rs
twitchyliquid64: Yeah 100%, we can do this properly now we have `BuilderError`! I've updated this to use `add_path()`.

core/src/program.rs
twitchyliquid64: Ripped it out (except the `EC` generic on `build()`) and its working a treat!

package/src/manifest.rs
twitchyliquid64: Good idea! gimme 10 and this should be ready for another review pass

twitchyliquid64:
@yannham This is good for another pass! Ive created the new `BuilderError` type which ended up fixing some of the fishy refactoring, as well as removing `EC` and adding the `add_source_string` method.

core/src/program.rs
```diff
@@ -1204,6 +949,411 @@ impl<EC: EvalCache> Program<EC> {
...
+/// ```no_run
+/// use nickel_lang_core::{eval::cache::CacheImpl, program::{Program, ProgramBuilder}};
+///
+/// let program: Program<CacheImpl> = ProgramBuilder::new()
+///     .add_path("config.ncl")
+///     .add_import_paths(["./vendor"])
+///     .build()
+///     .unwrap();
+/// ```
+pub struct ProgramBuilder<R = NullReporter, W = io::Sink> {
```
yannham, line 1020: TIL that you can have default types for generic parameters!

core/src/program.rs
```diff
@@ -1204,6 +949,411 @@ impl<EC: EvalCache> Program<EC> {
...
+
+        for file in contract_paths {
+            let file_str = file.to_string_lossy().into_owned();
+            let file_id = cache
+                .sources
+                .add_file(file, InputFormat::Nickel)
+                .map_err(|err| BuilderError::Io {
+                    path: Some(file_str.into()),
+                    error: err,
+                })?;
```
yannham: I think `PathBuf` implements `From<OsString>` as well, so there's no need for string conversion here.

```suggestion
            let file_id = cache
                .sources
                .add_file(file.clone(), InputFormat::Nickel)
                .map_err(|err| BuilderError::Io {
                    path: Some(file_str.into()),
                    error: err,
                })?;
```

core/src/program.rs
```diff
@@ -1534,37 +1684,22 @@ mod doc {
...
 mod tests {
     use super::*;
-    use crate::{error::NullReporter, eval::cache::CacheImpl};
+    use crate::eval::cache::CacheImpl;
     use assert_matches::assert_matches;
-    use std::io::Cursor;
+
+    fn build_test_program(s: &str) -> Result<Program<CacheImpl>, BuilderError> {
+        ProgramBuilder::new()
+            .add_source_string(s.to_owned(), "<test>")
```
yannham: I think you nicely made `add_source_string` general enough for that, no?

```suggestion
            .add_source_string(s, "<test>")
```

core/src/program.rs
```diff
@@ -1607,4 +1742,39 @@ mod tests {
             Err(Error::ParseErrors(_))
         );
     }
+
+    #[test]
+    fn builder_evaluates_single_source() {
+        let mut prog: Program<CacheImpl> = ProgramBuilder::new()
+            .add_source_string("1 + 2".to_owned(), "<builder-test>")
```
yannham:
```suggestion
            .add_source_string("1 + 2", "<builder-test>")
```

core/src/program.rs
```diff
@@ -1607,4 +1742,39 @@ mod tests {
...
+            .unwrap();
+        let v = prog.eval_full().unwrap();
+        assert_eq!(v.without_pos(), crate::term::make::integer(3).without_pos());
+    }
+
+    #[test]
+    fn builder_merges_multiple_sources() {
+        let mut prog: Program<CacheImpl> = ProgramBuilder::new()
+            .add_source_string("{ a = 1 }".to_owned(), "<a.ncl>")
+            .add_source_string("{ b = 2 }".to_owned(), "<b.ncl>")
```
yannham:
```suggestion
            .add_source_string("{ a = 1 }", "<a.ncl>")
            .add_source_string("{ b = 2 }", "<b.ncl>")
```

core/tests/integration/main.rs
```diff
@@ -70,16 +71,13 @@ fn run_test(test_case: TestCase<Test>, path: String) {
...
-        )
-        .expect("");
+        let mut builder = ProgramBuilder::new()
+            .add_source_string(program.clone(), path.as_str())
+            .with_trace(std::io::stderr());
         if let Some(imports) = &test_case.annotation.nickel_path {
-            p.add_import_paths(imports.iter());
+            builder = builder.add_import_paths(imports.iter().cloned());
         }
+        let p: TestProgram = builder.build().expect("");
```
yannham:
```suggestion
        let p: TestProgram = builder.build().unwrap();
```

core/tests/examples/main.rs
```diff
@@ -14,12 +17,11 @@ fn check_example_file(path: &str) {
...
 
     // `test_resources` uses paths relative to the workspace manifesty
-    let mut p = TestProgram::new_from_file(
-        project_root().join(path),
-        std::io::stderr(),
-        NullReporter {},
-    )
-    .expect("Failed to load program from file");
+    let mut p: TestProgram = ProgramBuilder::new()
+        .add_path(project_root().join(path).into_os_string())
```
yannham: Same, the method is nicely generic, and `PathBuf <-> OsString` with `From` instances

```suggestion
        .add_path(project_root().join(path))
```

core/tests/integration/main.rs
```diff
@@ -70,16 +71,13 @@ fn run_test(test_case: TestCase<Test>, path: String) {
...
     for _ in 0..repeat {
-        let mut p = TestProgram::new_from_source(
-            Cursor::new(program.clone()),
-            path.as_str(),
-            std::io::stderr(),
-            NullReporter {},
-        )
-        .expect("");
+        let mut builder = ProgramBuilder::new()
+            .add_source_string(program.clone(), path.as_str())
```
yannham:
```suggestion
            .add_source_string(program.clone(), path)
```

core/tests/integration/program_api.rs
```diff
@@ -1,31 +1,21 @@
...
-                InputFormat::Nickel,
-            ),
-        ],
-        std::io::stderr(),
-        sink,
-    )
-    .unwrap();
+    let mut prog: Program<CBNCache> = ProgramBuilder::new()
+        .add_source(Cursor::new(b"{}".to_vec()), "fst")
+        .add_source(Cursor::new(b"{} & {}".to_vec()), "snd")
```
yannham:
```suggestion
        .add_source_string("{}", "fst")
        .add_source_string("{} & {}", "snd")
```

core/tests/integration/query.rs
```diff
@@ -18,18 +17,20 @@ impl ProgramExt for Program<CBNCache> {
     }
 }
 
+fn program_from(src: &'static str) -> TestProgram {
+    ProgramBuilder::new()
+        .add_source(src.as_bytes(), "regr_tests")
```
yannham:
```suggestion
        .add_source_string(src, "regr_tests")
```

package/src/manifest.rs
```diff
@@ -190,20 +189,26 @@ impl ManifestFile {
...
+                "<in-memory manifest>",
+            )
```
yannham, line 205:
```suggestion
            .add_source_string(data, "<in-memory manifest>")
```

yannham (approved):
Left a bunch of minor suggestions or simplifications, but looks good overall! I'll let you have a quick look and deny/accept and merge then.

package/src/manifest.rs
twitchyliquid64: Unfortunately it doesnt seem to be about to work with a &[u8]:

```
error[E0277]: the trait bound `std::string::String: From<&[u8]>` is not satisfied
    --> package/src/manifest.rs:202:32
     |
 202 |             .add_source_string(data, "<in-memory manifest>")
     |              ----------------- ^^^^ the trait `From<&[u8]>` is not implemented for `std::string::String`
```

yannham: Ah, I guess we could still convert to a String and keep using `add_source_string` to avoid the `Cursor` but then we'd have to decide what to do on UTF8 errors. I guess it's fine as it is.

PR #2607, opened 2026-05-01
ci(macos,bin): upload `nickel-pkg-arm64-macos` variant

mkatychev:
`nickel-pkg-arm64-macos` was never being uploaded to the github releases page. 
Added `gh release upload --clobber $RELEASE_TAG nickel-pkg-arm64-macos` to the macos release job.

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -135,10 +135,11 @@ jobs:
           GH_TOKEN: ${{ github.token }}
           RELEASE_TAG: ${{ github.event_name == 'release' && github.event.release.tag_name || github.event.inputs.release_tag }}
         run: |
-          gh release upload --clobber $RELEASE_TAG nickel-arm64-macos
-          gh release upload --clobber $RELEASE_TAG nls-arm64-macos
           gh release upload --clobber $RELEASE_TAG libnickel_lang-arm64-macos.a
           gh release upload --clobber $RELEASE_TAG libnickel_lang-arm64-macos.dylib
+          gh release upload --clobber $RELEASE_TAG nickel-arm64-macos
+          gh release upload --clobber $RELEASE_TAG nickel-pkg-arm64-macos
+          gh release upload --clobber $RELEASE_TAG nls-arm64-macos
 
   release-artifacts-windows:
     name: "Build Windows Nickel binaries"


jneem (approved):

mkatychev:
@jneem can you confirm that nickel is not intended to be linked dynamically linked against nix packages for macos?

```nushell
(http get https://github.com/nickel-lang/nickel/releases/download/1.16.0/nickel-arm64-macos 
  | save nickel-gh-release
  chmod +x ./nickel-gh-release
  ./nickel-gh-release
)
```

```sh-session
$ ./nickel-gh-release
dyld[81836]: Library not loaded: /nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib
  Referenced from: <5930CB07-A159-3A59-BAB0-A5112B9FFB83> /Users/mkatychev/nickel-gh-release
  Reason: tried: '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file)
zsh: abort      nickel
```

I can confirm there's a dylib path reference inside of the binary:
```
$ rg --text 'store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' ./nickel-gh-release
2:�4__cstring__TEXT�Y?n'�Y?__unwind_info__TEXT�@ ��@__eh [... 1 more match]
```

Looks like we're maybe not statically linking here?
https://github.com/nickel-lang/nickel/blob/6cef5d8d4d3971ac02691247185cc3a55c87bc24/flake.nix#L88-L93

jneem:
Yeah, that definitely isn't intentional. And although I don't think any of us uses macos regularly, I'm pretty sure it was tested and working at some point...

There is a comment in the flake about static linking not working on macos:

```
          # Static building isn't really possible on MacOS because the system
          # call ABIs aren't stable. This output shouldn't be used on MacOS.
```

but maybe there's a way to link everything statically except libc?

mkatychev:
> Yeah, that definitely isn't intentional. And although I don't think any of us uses macos regularly, I'm pretty sure it was tested and working at some point...
> 
> There is a comment in the flake about static linking not working on macos:
> 
> ```
>           # Static building isn't really possible on MacOS because the system
>           # call ABIs aren't stable. This output shouldn't be used on MacOS.
> ```
> 
> but maybe there's a way to link everything statically except libc?

I have https://github.com/nickel-lang/nickel/pull/2609 as a fix ~but I'm not sure how I can test the build pipeline to confirm the releases work.~

mkatychev:
Don't have the bandwidth for this issue currently, moving to draft

PR #2609, opened 2026-05-01
fix(build,darwin,nix): reference static version of libiconv for macOS build

### LINKED ISSUE #2608: macOS libiconv is being dynamically linked
@jneem can you confirm that nickel is not intended to be linked dynamically linked against nix packages for macos?

```nushell
(http get https://github.com/nickel-lang/nickel/releases/download/1.16.0/nickel-arm64-macos 
  | save nickel-gh-release
  chmod +x ./nickel-gh-release
  ./nickel-gh-release
)
```

```sh-session
$ ./nickel-gh-release
dyld[81836]: Library not loaded: /nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib
  Referenced from: <5930CB07-A159-3A59-BAB0-A5112B9FFB83> /Users/mkatychev/nickel-gh-release
  Reason: tried: '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file)
zsh: abort      nickel
```

I can confirm there's a dylib path reference inside of the binary:
```
$ rg --text 'store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' ./nickel-gh-release
2:�4__cstring__TEXT�Y?n'�Y?__unwind_info__TEXT�@ ��@__eh [... 1 more match]
```

Looks like we're maybe not statically linking here?
https://github.com/nickel-lang/nickel/blob/6cef5d8d4d3971ac02691247185cc3a55c87bc24/flake.nix#L88-L93

_Originally posted by @mkatychev in https://github.com/nickel-lang/nickel/issues/2607#issuecomment-4361041201_

mkatychev:
macOS users who do not have nix installed get a linker error when pulling the github release build artifact for macOS:
```
dyld[15436]: Library not loaded: /nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib
  Referenced from: <5930CB07-A159-3A59-BAB0-A5112B9FFB83> /path/to/nickel
  Reason: tried: '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file), '/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib' (no such file)
```

use `pkgs.darwin.pkgsStatic.libiconv` for macos building

resolves https://github.com/nickel-lang/nickel/issues/2608

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -93,7 +93,7 @@
       # Additional packages required to build Nickel on Darwin
       systemSpecificPkgs =
         if pkgs.stdenv.isDarwin then
-          [ pkgs.darwin.pkgsStatic.libiconv ]
+          [ pkgs.pkgsStatic.darwin.libiconv ]
         else
           [ ];
 
--- a/flake.nix
+++ b/flake.nix
@@ -309,31 +309,53 @@
               featuresArg = pkgs.lib.optionalString
                 (extraNickelFeatures != [ ])
                 ("--features " + pkgs.lib.strings.concatStringsSep " " extraNickelFeatures);
+              # Darwin libiconv is pulled in by the `libc` crate's build
+              # script (`cargo:rustc-link-lib=iconv`).
+              # including Darwin stdenv will prepend `pkgs.darwin.libiconv` to the link path,
+              # so without intervention the resulting binary pulls in
+              # `/nix/store/...libiconv.dylib`.
+              # Prepend the static archive's lib directory to rustc's native search path
+              # so `-liconv` resolves to `libiconv.a` first under Apple ld's
+              # `-search_paths_first` default.
+              darwinStaticIconv = pkgs.pkgsStatic.darwin.libiconv;
+              darwinIconvRustflags =
+                "-L native=${darwinStaticIconv.dev}/lib";
+              baseEnv = {
+                PYO3_PYTHON = "${pkgs.python3}/bin/python";
+              } // pkgs.lib.optionalAttrs pkgs.stdenv.isDarwin {
+                RUSTFLAGS = darwinIconvRustflags;
+              };
+              base = {
+                inherit
+                  pname
+                  pnameSuffix
+                  src
+                  version
+                  cargoArtifacts;
+
+                # pyo3 needs a Python interpreter in the build environment
+                # https://pyo3.rs/v0.17.3/building_and_distribution#configuring-the-python-version
+                nativeBuildInputs = with pkgs; [ pkg-config python3 ];
+                # A git binary is needed for some of the tests
+                buildInputs = with pkgs;
+                  [ git ]
+                    ++ pkgs.lib.optionals
+                    (builtins.elem "nix-experimental" extraNickelFeatures)
+                    [ boost nix ]
+                    ++ pkgs.lib.optional pkgs.stdenv.isDarwin darwinStaticIconv;
+
+                cargoExtraArgs = "${cargoBuildExtraArgs} ${featuresArg} ${extraBuildArgs} --workspace";
+                CARGO_PROFILE = profile;
+                env = baseEnv;
+              };
             in
             # The `cargoArtifacts` are the dependencies of the workspace, so we
-              # can reuse them in the build.
-            craneLib.buildPackage ({
-              inherit
-                pname
-                pnameSuffix
-                src
-                version
-                cargoArtifacts;
-
-              # pyo3 needs a Python interpreter in the build environment
-              # https://pyo3.rs/v0.17.3/building_and_distribution#configuring-the-python-version
-              nativeBuildInputs = with pkgs; [ pkg-config python3 ];
-              # A git binary is needed for some of the tests
-              buildInputs = with pkgs;
-                [ git ]
-                  ++ pkgs.lib.optionals
-                  (builtins.elem "nix-experimental" extraNickelFeatures)
-                  [ boost nix ];
-
-              cargoExtraArgs = "${cargoBuildExtraArgs} ${featuresArg} ${extraBuildArgs} --workspace";
-              CARGO_PROFILE = profile;
-              env.PYO3_PYTHON = "${pkgs.python3}/bin/python";
-            } // extraArgs);
+            # can reuse them in the build. Merge `env` from `extraArgs` rather
+            # than replace it so callers that pass an `env` (e.g. for
+            # `NICKEL_NIX_BUILD_REV`) don't clobber the darwin RUSTFLAGS.
+            craneLib.buildPackage (base // extraArgs // {
+              env = baseEnv // (extraArgs.env or { });
+            });
 
           # To build Nickel and its dependencies statically we use the musl
           # libc and clang with libc++ to build C and C++ dependencies. We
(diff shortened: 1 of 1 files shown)


jneem (approved):
I'm not in a position to test this, but it seems fine to me.

yannham (commented):
I wonder if we should unconditionally do that, or do it only for the static build. What I mean is that if you use Nix and don't build a static executable, you suddenly have a bigger binary than before on MacOS, right?

mkatychev:
> I wonder if we should unconditionally do that, or do it only for the static build. What I mean is that if you use Nix and don't build a static executable, you suddenly have a bigger binary than before on MacOS, right?

I agree, but I think it is to be expected to have a statically linked binary show up in the github releases if it is going to be a platform agnostic distributable.

I'm not sure what the most idiomatic nix way to express both.

mkatychev:
This still hasn't been confirmed to work so I'm putting it in draft (I need to find a macOS install without nix) and I don't want to uninstall it from my system 😆.

mkatychev:
Getting this error in CI which I didn't hit locally 🤔:
```
    error: attribute 'pkgsStatic' missing
       at /Users/runner/work/nickel/nickel/flake.nix:96:13:
           95|         if pkgs.stdenv.isDarwin then
           96|           [ pkgs.darwin.pkgsStatic.libiconv ]
             |             ^
           97|         else
```

There's some reference to an `isStatic` flag upstream but I have no idea how to toggle it:
https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/os-specific/darwin/apple-source-releases/libiconv/package.nix#L66

yannham:
> Getting this error in CI which I didn't hit locally 🤔:

An LLM suggested that `pkgsStatic` should be in front of darwin, as in `pkgs.pkgsStatic.darwin` instead of `pkgs.darwin.pkgsStatic`, which I can confirm in the Nix REPL by looking at the attrs set structure (though I'm not on darwin, but can still see the available attributes).

But the fact that it works locally for you is strange, then.

mkatychev:
> > Getting this error in CI which I didn't hit locally 🤔:
> 
> An LLM suggested that `pkgsStatic` should be in front of darwin, as in `pkgs.pkgsStatic.darwin` instead of `pkgs.darwin.pkgsStatic`, which I can confirm in the Nix REPL by looking at the attrs set structure (though I'm not on darwin, but can still see the available attributes).
> 
> But the fact that it works locally for you is strange, then.

I've gone ahead and created a test case for this issue to hopefully catch these kinds of regressions on CI, keep in mind this has to be done without nix installed:
https://github.com/mkatychev/nickel/blob/test/macos-libiconf/.github/workflows/continuous-integration.yml

...the issue isn't solved but at least it is testable now.

flake.nix
```diff
@@ -93,7 +93,7 @@
       # Additional packages required to build Nickel on Darwin
       systemSpecificPkgs =
         if pkgs.stdenv.isDarwin then
-          [ pkgs.darwin.libiconv ]
+          [ pkgs.darwin.pkgsStatic.libiconv ]
```
yannham: I'm trying this for the CI.

```suggestion
          [ pkgs.pkgsStatic.darwin.libiconv ]
```

jneem:
@mkatychev I triggered a release action with this change. Could you check if the macos binary [here](https://github.com/nickel-lang/nickel/releases/tag/untagged-6e044db3bf2856b0e497) is as you expect?

mkatychev:
@jneem @yannham 
I'm still getting the same error, and grep is returning 2 dylib strings in the binary:

```sh-session
$ rg --text ".dylib" /nix/store/4kxxg8w48qfrkq8kkd1a9zhficijsk9a-nickel-lang-rev-fixup-1.16.0/bin/nickel
2:h�4__cstring__TEXTHY?n'HY?__unwind_info__TEXT��@���@__e [... 2 more matches]
```

https://github.com/mkatychev/nickel/actions/runs/25330806565/job/74263729514?pr=1

EDIT: the other `.dylib` match is `/usr/lib/libSystem.B.dylib` which is to be expected:

```sh-session
$ otool -L /nix/store/4kxxg8w48qfrkq8kkd1a9zhficijsk9a-nickel-lang-rev-fixup-1.16.0/bin/nickel
/nix/store/4kxxg8w48qfrkq8kkd1a9zhficijsk9a-nickel-lang-rev-fixup-1.16.0/bin/nickel:
	/nix/store/9kffgbvhza212ishsam4p8wklh92ih9v-libiconv-109.100.2/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.100.2)
```

mkatychev:
@jneem I have a fix that is confirmed to be working:
https://github.com/mkatychev/nickel/actions/runs/25333078826/job/74271364940?pr=1

...but the solution is a bit ugly because of how stdenv prepends to link paths: https://github.com/nickel-lang/nickel/pull/2609/commits/3140d7f5bc47118d40518bade4ea7290f1987007

yannham:
Hmm. I wonder if another simple solution would just be to avoid the `patchElf` phase for the "static" Darwin executable. After all, this seems to be the real offender, messing with the dynlib paths in the binary? Is libiconv available by default on MacOS?

To do so, we have to drop `dontPatchElf = true;` at the right level of a derivation, but that shouldn't be too hard.

mkatychev:
> `dontPatchElf`
should that be set in a fixup or build phase? 

> Is libiconv available by default on MacOS?

does not look like it

EDIT the cargo installed version of it does point to a `/usr/lib` path:
```
$ otool -L /Users/mkatychev/.cargo/bin/nickel
/Users/mkatychev/.cargo/bin/nickel:
	/System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1351.101.1)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 3423.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

PR #2611, opened 2026-05-04
Implement ProgramBuilder::extend_initial_env and plumbing

### LINKED ISSUE #2600: FR: Way to inject nickel values to the top-level variable scope when driving `nickel_lang_core::program::Program`

twitchyliquid64:
Fixes #2600 

This PR implements the ability for a user of `Program` to define additional values in the initial environment.

The wiggly-est bit was trying to make this work with the typecheck logic + the cache types. I kinda short-circuited on just registering everything as a `Dyn` - let me know if this is inappropriate or you have suggestions of how we should integrate with the type check path.

I also made `Environment` be the type where additional bindings are communicated, but given you have to mint it with a `CacheImpl` im not sure that was the right choice.

### DIFF
--- a/core/src/cache.rs
+++ b/core/src/cache.rs
@@ -1377,6 +1377,16 @@ impl CacheHub {
         asts.add_type_bindings(pos_table, slice, term)
     }
 
+    /// Registers each given identifier in the typing environment as `Dyn`. Used to expose
+    /// host-provided top-level names to the typechecker. See
+    /// [AstCache::add_dyn_type_bindings] for details.
+    pub fn add_dyn_type_bindings<I>(&mut self, idents: I)
+    where
+        I: IntoIterator<Item = crate::identifier::Ident>,
+    {
+        self.asts.add_dyn_type_bindings(&self.sources, idents);
+    }
+
     /// Converts an AST and all of its transitive dependencies to the runtime representation,
     /// populating the term cache. `file_id` and any of its Nickel dependencies must be present in
     /// the AST cache, or [CacheError::IncompatibleState] is returned. However, for non-Nickel
@@ -2851,6 +2861,26 @@ mod ast_cache {
                     .collect();
 
                 *slf.type_ctxt = typecheck::mk_initial_ctxt(slf.alloc, stdlib_terms_vec).unwrap();
+            });
+        }
+
+        /// Adds bindings to the typing environment with type [`ast::typ::TypeF::Dyn`], one per
+        /// identifier in `idents`. Used to expose host-provided extras to the typechecker without
+        /// committing to a more precise type.
+        ///
+        /// Ensures that the initial typing context is populated first, so the registrations are
+        /// layered on top of stdlib and win on collision.
+        pub fn add_dyn_type_bindings<I>(&mut self, sources: &SourceCache, idents: I)
+        where
+            I: IntoIterator<Item = crate::identifier::Ident>,
+        {
+            self.populate_type_ctxt(sources);
+            self.with_mut(|slf| {
+                for ident in idents {
+                    slf.type_ctxt
+                        .type_env
+                        .insert(ident, typecheck::UnifType::concrete(ast::typ::TypeF::Dyn));
+                }
             });
         }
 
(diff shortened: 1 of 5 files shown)


twitchyliquid64:
@yannham this is for closing out #2600, just wanted to get it on your radar!

core/src/cache.rs
```diff
@@ -2854,6 +2864,26 @@ mod ast_cache {
             });
         }
 
+        /// Adds bindings to the typing environment with type [`ast::typ::TypeF::Dyn`], one per
+        /// identifier in `idents`. Used to expose host-provided extras to the typechecker without
+        /// committing to a more precise type — values flow through dynamic typing.
```
yannham:
```suggestion
        /// committing to a more precise type.
```

core/src/eval/mod.rs
```diff
@@ -140,6 +140,10 @@ pub struct VmContext<R: ImportResolver, C: Cache> {
     pub cache: C,
     /// The position table, mapping position indices to spans.
     pub pos_table: PosTable,
+    /// Bindings to layer on top of the stdlib in the initial environment when a [VirtualMachine]
+    /// is created via [VirtualMachine::new]. On collision, the extension wins.
+    /// Configured via [Self::with_extend_env]; defaults to empty.
+    extend_env: Environment,
```
yannham: I wonder if that should live in the vm context. It's never used after VM creation, so I feel like we could just get rid of it and add an `extend_initial_env` or something like that directly to `VirtualMachine`?

yannham: Ok, in fact you need it to extend the type environment, independently.

yannham (changes requested):
Thanks for the contribution, the overall approach looks reasonable.

For typechecking, I think it's ok to start like this, but we can re-use later some of the machinery used for the standard library: there's a very limited way to perform type inference there, which could help for std-like extra environment.

However, I don't think `extra_env` should be an `Environment`. `Environment` is a custom data-structure optimized for evaluation, but `extra_env` is never used as it is but eventually inserted in the environment. So it could be a hashmap, or in fact even better, a `Vec<(Ident, NickelValue)>`.

Notice that I use `NickelValue` instead of `Cache::Id`: since we provide the extra env to the `VmContext`, and the latter already has a cache at hand, the `VmContext` can take care of eval cache business (basically substituting `value` for `cache.add(value)`) and free the consumer from having to think about eval cache at all. In fact, the statement that the eval cache is stateless is currently true, but isn't guaranteed and might change in the future (incremental evaluation). We must use the same cache as the one in the context.

twitchyliquid64:
Thank you for the review! I will change the type carrying the extra env (currently `Environment`) to `Vec<(Ident, NickelValue)>` and address the other comments (probably tomorrow) and get back to you.

yannham (approved):

PR #2614, opened 2026-05-21
Bump nickel-lang

jneem:
This also adds it to the release script, so it should be less likely that we'll forget in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3550,7 +3550,7 @@ dependencies = [
 
 [[package]]
 name = "nickel-lang"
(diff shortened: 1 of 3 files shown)


yannham:
Is it a duplicate of #2615?

jneem:
It's the same diff, but this one targets the 1.16.0-release branch

edit: (this is the one that I used to actually `cargo publish` the `nickel-lang` crate)

yannham (approved):
Ah, my bad 👍

PR #2619, opened 2026-06-05
chore: update cargo deps

twitchyliquid64:
Went through cargo deps and tried to bump them to more recent versions!

The LOC diff is misleading, most of that is the autogenerated `Cargo.lock`.

Everything applied cleanly except for the parser, but there were just two smol changes to get them to dance to the (updated) tune.


twitchyliquid64:
Worth calling out in the PR, but i updated `sha-1` from RustCrypto to `sha1` from the same author, as the whole crate is deprecated. The intent is coalescing the digest internal crates that are duplicated due to the age of the `sha-1` crate.

Cargo.toml
```diff
@@ -47,101 +47,101 @@ lsp-harness = { version = "0.1.0", path = "./lsp/lsp-harness" }
 # have to care about our nix build.
 wasm-bindgen = "0.2.100"
 
-anstyle = "1.0.13"
-anyhow = "1.0.97"
-assert_cmd = "2.0.11"
+anstyle = "1.0.14"
```
jneem: Where we don't have a specific reason (e.g. a fix for a bug that we care about), can we just leave out the semver-compatible bumps? That's just more flexible, I think, and shouldn't result in any duplicate dependencies in the tree.

I have no objection to the semver-changing version bumps, and no objection to bumping anything in the lock file.

twitchyliquid64: Ooo on that note, would you be okay with us eliminating unncessary patch versions entirely, and relying on the lock file for consistency?

That might also help downstream in terms of version selection.

twitchyliquid64:
A little controversial, but what do you think of this approach? `0.x` for non 1.0 deps, and a major version for the rest, with the following exceptions:

  - wasm-bindgen "0.2.100" - given the coupling we have with the topiary crate and my earlier attempt of a PR on that one
  - js-sys "0.3.77" - samezies, paired with wasm-bindgen
  - lsp-types "0.95" - Url removed in 0.96+
  - pubgrub "0.3" - 0.4 requires rustc 1.92, MSRV is 1.91
  - criterion "0.5" - pprof pins this

All the tests and such pass.

jneem:
That makes sense to me. Any objections, @yannham?

yannham:
Sounds good to me as well. We can always special case some dependencies if for some reason we need to impose a higher lower bound.

twitchyliquid64:
Awesomesauce! Lmk if you want any changes but i think what i have now fits the bill.

core/src/eval/operation.rs
```diff
@@ -3716,6 +3716,15 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
     }
 }
 
+fn hex_encode(bytes: &[u8]) -> String {
```
yannham, line 3719: Why use this helper? It looks longer, and not necessarily more performant (manual looping).

lsp/nls/src/contracts.rs
```diff
@@ -78,7 +78,7 @@ impl LocalConfig {
     fn from_file(path: &Path) -> Result<Self, anyhow::Error> {
         let local_config: LocalConfigFormat = nickel_lang_core::deserialize::from_path(path)
             // Eagerly format the error, because anyhow requires `Send` errors and ours aren't
-            .map_err(|e| anyhow::anyhow!("{}", e.to_string()))?;
+            .map_err(|e| anyhow::anyhow!("{}", e))?;
```
yannham: I guess `anyhow` support the same syntax as `format!` more or less?

```suggestion
            .map_err(|e| anyhow::anyhow!("{e}"))?;
```

parser/src/grammar.lalrpop
```diff
@@ -968,14 +968,14 @@ OrPatternBranch: Pattern<'ast> = {
     // A variant pattern of the form `'<Tag> or`. The trick is to instead
     // consider it as the enum tag pattern `'<Tag>` followed by the `or`
     // contextual keyword after-the-fact.
-    EnumVariantOrPattern => {
-        let pos = <>.pos;
+    <evp: EnumVariantOrPattern> => {
```
yannham, line 971: This looks unrelated, and the rest of the grammar uses the convention that for single rules like this we use `<>`.

core/src/eval/operation.rs
twitchyliquid64:
```shell
error[E0277]: the trait bound `md5::digest::hybrid_array::Array<u8, UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>>: LowerHex` is not satisfied
    --> core/src/eval/operation.rs:2582:41
     |
2582 |                         format!("{:x}", hasher.finalize())
     |                                  ----   ^^^^^^^^^^^^^^^^^ the trait `LowerHex` is not implemented for `md5::digest::hybrid_array::Array<u8, UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>>`
     |                                  |
     |                                  required by this formatting parameter
```

The new digest crates all coalesced on using typenum-based array generics, and i couldnt figure out a conversion that had short/easy ergonomics.

Obvious stuff doesnt work because `[u8]` doesn't implement `LowerHex` either :(

```shell
2582 |                         format!("{:x}", &hasher.finalize().as_slice())
     |                                  ----   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `LowerHex` is not implemented for `[u8]`
```

lsp/nls/src/contracts.rs
twitchyliquid64: Done!

parser/src/grammar.lalrpop
twitchyliquid64: The former text is no longer accepted it seems:

```shell
  processing file `src/grammar.lalrpop`
  src/grammar.lalrpop:971:5: 971:24 error: When there are multiple `<>` in the action, there must be the same number of sources for the `<>`s. Found 2 `<`>`s and 1 anonymous sources.

  --- stderr
  Generating a fresh LALRPOP grammar
        EnumVariantOrPattern => {
        ~~~~~~~~~~~~~~~~~~~~
```

Iiuc, `<>` is only allowed for a single anonymous value and not for two?

jneem (approved):

yannham (approved):

PR #2628, opened 2026-06-26
fix(stdlib): avoid panic in string.find on a match at the end of the string

### LINKED ISSUE #2627: std.string.find panics when matching empty string
Consider the following session:

```
❯ nickel repl
Nickel 1.17.0
Type ":help" for help, or ":exit" to exit
nickel> std.string.find "^$" ""

thread 'main' (8251) panicked at core/src/term/string.rs:342:18:
We already know that `first_match.start()` occurs on a cluster boundary.
```

Expected:

```
nickel> std.string.find "^$" ""
{ groups = [  ], index = 0, matched = "", }
```

SAY-5:
Fixes #2627.

`std.string.find` panics when the match ends at the end of the string, such as `std.string.find "^$" ""`. The code translates the regex byte offset to a grapheme cluster index by scanning `grapheme_indices`, but no cluster starts at the very end of the string (and the empty string has no clusters at all), so the `expect` on the lookup fires.

The byte offset for a match at the end of the string equals the string's byte length, which maps to the total cluster count. I append that pair to the scan so the lookup resolves instead of panicking, and added the empty-match case from the issue to the existing `string_find` fixture.

### DIFF
--- a/core/src/term/string.rs
+++ b/core/src/term/string.rs
@@ -329,10 +329,18 @@ impl NickelString {
             // The indices returned by the `regex` crate are byte offsets into
             // the string, but we need to return the index into the Nickel string,
             // i.e., the grapheme cluster index which starts at this byte offset.
+            //
+            // A match can start at the very end of the string (for example an
+            // empty match like `^$` on the empty string), in which case no
+            // grapheme cluster starts at that byte offset. We append the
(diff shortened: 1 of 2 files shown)


yannham (approved):
LGTM. Thanks for the fix!

PR #2632, opened 2026-07-06
Fix parse error when using negative number literals in match patterns

### LINKED ISSUE #2629: Negative number literals in patterns
`-1` is a valid expression, but it's not a valid pattern in a match clause:

```
nickel> -1 |> match {-1 => true}
error: unexpected token
  ┌─ <repl-input-19>:1:14
  │
1 │ -1 |> match {-1 => true}
  │              ^
```

Adding parentheses does not help either:

```
nickel> -1 |> match {(-1) => true}
error: unexpected token
  ┌─ <repl-input-20>:1:15
  │
1 │ -1 |> match {(-1) => true}
  │               ^
```

It should be possible to express negative numbers in the patterns of match clauses.

L0r3m1p5um:
Fixes #2629

The main change here is to match a `SignedNumLiteral` instead of a `NumberLiteral` in constant patterns so negative values can be matched. The rest is just to get `without_pos` to work for patterns so I can write a test for it. The `without_pos` implementation isn't fully tested and is likely to have some issues but it's only used in `#[cfg(test)]` so I hope that's all right.

### DIFF
--- a/parser/src/ast/pattern/mod.rs
+++ b/parser/src/ast/pattern/mod.rs
@@ -3,6 +3,8 @@ use std::collections::{HashMap, hash_map::Entry};
 
 use super::{Annotation, Ast, Number};
 
+#[cfg(test)]
+use crate::ast::AstAlloc;
 use crate::{
     error::ParseError, identifier::LocIdent, impl_display_from_bytecode_pretty, position::TermPos,
     traverse::*,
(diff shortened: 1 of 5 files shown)


L0r3m1p5um:
Could someone check on CI? There seems to be a runner issue.

jneem:
I can't even load the logs... I restarted the job, and we'll see what happens.

parser/src/ast/pattern/mod.rs
```diff
@@ -155,6 +157,112 @@ impl Pattern<'_> {
     }
 }
 
+impl<'ast> Pattern<'ast> {
```
jneem, line 158: I guess if we ever want this outside of tests (or for anything more general), this should be replaced by an `impl Traverse`. It seems fine for now, though

jneem (approved):

parser/src/ast/pattern/mod.rs
```diff
@@ -3,6 +3,8 @@ use std::collections::{HashMap, hash_map::Entry};
 
 use super::{Annotation, Ast, Number};
 
+#[cfg(test)]
+use crate::ast::AstAlloc;
```
yannham: Nit: we could move that in `without_pos` to avoid the additional cfg guard

yannham (approved):

yannham:
The build is failing because of a clippy error in nickel-lang-parser I believe

L0r3m1p5um:
> The build is failing because of a clippy error in nickel-lang-parser I believe

Where are you seeing that? The logs in CI look incomplete and when I run clippy myself I don't see any warnings.

yannham:
```
nickel-lang-clippy> error: this `.into_iter()` call is equivalent to `.iter()` and will not consume the `slice`
nickel-lang-clippy>    --> parser/src/ast/pattern/mod.rs:177:26
nickel-lang-clippy>     |
nickel-lang-clippy> 177 |                         .into_iter()
nickel-lang-clippy>     |                          ^^^^^^^^^ help: call directly: `iter`
nickel-lang-clippy>     |
nickel-lang-clippy>     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#into_iter_on_ref
nickel-lang-clippy>     = note: `-D clippy::into-iter-on-ref` implied by `-D warnings`
nickel-lang-clippy>     = help: to override `-D warnings` add `#[allow(clippy::into_iter_on_ref)]`
nickel-lang-clippy> 
nickel-lang-clippy> error: could not compile `nickel-lang-parser` (lib test) due to 1 previous error
```

Maybe it's a feature thing? But yeah the issue with Nix is that the different dependencies are built in a row and are very verbose, so by the time the overall build fails, the cause has been drowned into other unrelated, more recent logs.

PR #2634, opened 2026-07-11
Update vscode package metadata

jneem:
For a start, this updates the repo url in the vscode package metadata, which I think should be uncontroversial.

It also updates the publisher, which is a little more annoying and will cause some churn, but I think this is more accurate? And because vscode marketplace insists on having unique `name` and `displayName` (even across different publishers) it tweaks those fields a little.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -684,14 +684,14 @@
 
       # Build the Nickel VSCode extension
       vscodeExtension = pkgs.mkYarnPackage {
-        pname = "vscode-nickel";
+        pname = "vscode-nickel-lang";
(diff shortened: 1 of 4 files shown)


L0r3m1p5um:
We'll need to add a deprecation notice on the extension under the `tweag` publisher so that users will know they need to move to the new publisher. https://code.visualstudio.com/api/working-with-extensions/publishing-extension#deprecating-extensions

Otherwise, makes sense to me.

L0r3m1p5um (approved):

yannham (approved):

yannham:
I probably own the tweag publisher on the vscode market place, so feel free to ping me when we do need to deprecate.

PR #2638, opened 2026-07-17
Add std.array.group

mboes:
`std.array.group` is a generalization of `std.array.partition` to an
arbitrary number of keys instead of just `right` and `wrong`. Data
munging in Nickel is often structured as chains of calls to
`std.array.flat_map`, `std.array.filter` and the like. This is as in
C#'s [LINQ][ms-query-ops] that calls out 5 basic data operations,
which we only partially support today:

- ✅ filter data with `std.array.filter`
- ✅ order data with `std.array.sort`
- ❌ group data
- ❌ join data (except in a limited form with `std.array.zip`)
- ✅ project data with `std.array.map` and `std.array.flat_map`

Here we implement the "group data" operation.

[ms-query-ops]: https://learn.microsoft.com/en-us/dotnet/csharp/linq/standard-query-operators/

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -524,6 +524,8 @@
           elements that satisfy the predicate, while `wrong` will contain those
           that do not.
 
+          `partition` is a special case of `group`.
+
           # Examples
 
           ```nickel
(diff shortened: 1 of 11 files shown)


core/stdlib/std.ncl
```diff
@@ -524,21 +524,60 @@
...
+          { type = "fruit", name = "pear" },
+        ]
+        # => {
+        #      fruit =
+        #        [ { name = "apple", type = "fruit", }, { name = "pear", type = "fruit", } ],
+        #      vegetable = [ { name = "leek", type = "vegetable", } ],
+        #    }
+
+        group (fun x => if x < 5 then "right" else "wrong") [ 2, 4, 5, 3, 7, 8, 6 ]
+	# => { right = [ 2, 4, 3 ], wrong = [ 5, 7, 8, 6 ] }
```
```suggestion
	    # => { right = [ 2, 4, 3 ], wrong = [ 5, 7, 8, 6 ] }
```

yannham (approved):
LGTM, thanks!

mboes:
@yannham my Emacs inserted a tab by mistake. Converted the tab to spaces.

PR #2639, opened 2026-07-17
Add std.array.chunk

mboes:
This function is the analog to Haskell's `groupBy`. The latter is a misnomer, in that grouping in relational terms is more accurately captured by nickel-lang/nickel#2638. However, `groupBy` is still useful. It computes a sequence of maximal *runs*, where runs are sequences of elements that share a same key. Where `std.array.group` mashes together all sequences of all elements identified by key, `std.array.chunk` does not: chunking `[0, 0, 1, 0]` yields `[[0, 0], [1], [0]]`.

Example use cases:
- cleaning up a list by removing duplicates or merging adjacent (e.g. firewall) rules while preserving the semantically significant input order
- grouping by key while preserving the order of keys, e.g. grouping devices for a dashboard by floor in a building without reordering floors

Thoughts on naming follows. We don't use `group` because that is already taken. Using `chunk` follows precedent from,

- [Ruby](https://ruby-doc.org/3.4.1/Enumerable.html#method-i-chunk)
- [Rust](https://doc.rust-lang.org/std/primitive.slice.html#method.chunk_by)

Alternatively, we could use `runs` as a name, but that would be breaking with the style of the rest of the library, where function names are verbs instead of nouns. And `run` would suggest something else entirely.

It should be noted that in some languages (Kotlin, C#), chunking means creating sequences of fixed size. We could consider passing the index into the user-supplied key selection function, in order to support computing chunks of fixed size as a special case. Or add `chunk_with_index`, like we have `map_with_index`. I haven't had a use so far for chunking based on element position.

Note on implementation: `std.array.group` could be implemented by composing `sort` and `chunk`. However, `partition` is implemented in terms of `group` and `sort` in terms of `partition`, so we'd have a circular dependency. Therefore, we keep separate the implementations of `group` and `chunk`.

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1111,6 +1111,65 @@
         array
         |> sort cmp
         |> dedup_sorted cmp,
+
+    chunk
+      : forall a k. (a -> k) -> Array a -> Array (Array a)
+      | Dyn -> Dyn -> Array std.array.NonEmpty
+      | doc m%"
+
+        Splits an array into maximal consecutive runs of elements that
+        share the same key, where the key of an element is computed by
+        the given function.
+
+        The order of elements is preserved, so flattening the result
+        yields the original array. For any key selection function `f`
+        and any array `xs`, we have:
+
+        `std.array.flatten (std.array.chunk f xs) == xs`
+
+        # Examples
+
+        ```nickel multiline
+        chunk (fun x => x) [1, 1, 2, 3]
+        # => [ [1, 1], [2], [3] ]
+
+        chunk (fun x => x % 2) [1, 3, 2, 4, 5]
+        # => [ [1, 3], [2, 4], [5] ]
+
+        chunk (fun x => x) []
+        # => []
+        ```
+      "%
+      = fun f xs =>
+        xs
+        |> match {
+          [] => [],
+          _ =>
+            let op = fun { chunks, current_key, start, end } x =>
+              let x_key = f x in
+              if x_key != current_key then
+                let chunks = chunks |> append (%array/slice% start end xs) in
+                let start = end in
+                let end = end + 1 in
+                let current_key = x_key in
+                { include [chunks, current_key, start, end] }
+              else
+                let end = end + 1 in
+                { include [chunks, current_key, start, end] }
+            in
+            let init = {
+              chunks = [],
+              current_key = f (first xs),
+              start = 0,
+              end = 1,
+            }
+            in
+            fold_left op init (drop_first xs)
+            |> match {
+              { chunks, start, end, .. } =>
+                chunks |> append (%array/slice% start end xs)
+            }
+        },
   },
 
   contract = {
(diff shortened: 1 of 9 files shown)


core/stdlib/std.ncl
```diff
@@ -1072,6 +1072,65 @@
...
+
+        chunk (fun x => x % 2) [1, 3, 2, 4, 5]
+        # => [ [1, 3], [2, 4], [5] ]
+
+        chunk (fun x => x) []
+        # => []
+        ```
+      "%
+      = fun f xs =>
+        xs
```
yannham: Nit: `match {...}` already acts as a lambda, so you could curry this function as `fun f => match { ... }`. Though the typing might be different, maybe that's the reason you eta-expanded on `xs` ?

core/stdlib/std.ncl
```diff
@@ -1072,6 +1072,65 @@
...
+        |> match {
+          [] => [],
```
yannham, line 1144: if it's just for testing emptyness, I would use a simple `if`. But maybe I'm being influenced by Rust here 🙂 (which promotes `if-let` for binary matches)

core/stdlib/std.ncl
```diff
@@ -1072,6 +1072,65 @@
...
+                let end = end + 1 in
+                let current_key = x_key in
+                { include [chunks, current_key, start, end] }
+              else
+                let end = end + 1 in
+                { include [chunks, current_key, start, end] }
+            in
+            let init = {
+              chunks = [],
+              current_key = f (first xs),
```
yannham: Why not use a pattern in this branch instead of a fallible function here, that is turn the case `_ => ...` into `[x, ..xs] => ...` ? That would get rid of `first xs` and `drop_first xs`.

core/stdlib/std.ncl
mboes: To take advantage of the eta contraction, per your previous comment, I'll use `match {}` over `if-then-else`. Let me know if you want it the other way round.

yannham: I didn't realize the two were incompatible 😓 but yeah let's go for the match

core/stdlib/std.ncl
mboes: OK, switched to (safer) pattern matching rather than partial functions from elsewhere in the standard library. Though the code might be a little less readable.

mboes:
@yannham PTAL.

Something I'm keen to get input on is the following design question. `chunk` as implemented here is less general than `groupBy` in Haskell and `chunk_by` in Rust itertools. However, it's more ergonomic for the common case where you're just projecting from a field and want equality to decide chunk boundaries. Also, the interface has the same structure as that of `std.array.group` that way. Is that what we want? We could add a more general variation of `chunk` at a later date if there is demand. If we did so, note that there is a variant more general still than Rust's `chunk_by`, namely the [`chunk_while` of Elixir](https://elixir.hexdocs.pm/Enum.html#chunk_while/4).

yannham (approved):
I'm fine with a specialized version for now, as long as it's not misnamed (taking real estate of a future generalization). Here it looks like a generalization would be `chunk_by` or `chunk_something`👍

PR #2643, opened 2026-07-26
Make faster implementation of `std.array.sort` using merge sort

### LINKED ISSUE #2637: std.array.sort has cubic complexity
Consider the following expression:

```nickel
let x = ... in
std.array.sort std.number.compare (std.array.range 0 x)
```

Letting `x` range from 10 to 500, for both the dev build (blue) and the release build (red), yields the following chart:

Image

**Expected:** log-linear complexity.
**Observed:** the best fit for the red data points is a cubic curve with $R²=0.9998$.

I cannot explain the cubic complexity, but I do have a theory for why it's at least quadratic: `std.array.partition` is called `O(log N)` times, but that function seems to have at least quadratic complexity, because at each step it appends an element to one of two arrays that have size `O(N)`.

L0r3m1p5um:
Fixes #2637

This fixes the cubic complexity of the `std.array.sort` implementation by reimplementing it with merge sort, which should run in O(n log n) in the worst case.

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -646,7 +646,8 @@
         "%
       = fun cmp array =>
         let rec merge = fun first second =>
-          match {
+          [first, second]
+          |> match {
             [[], ys] => ys,
             [xs, []] => xs,
             [[x, ..rest_xs],  [y, ..rest_ys]] =>
@@ -654,19 +655,20 @@
                 [y] @ merge first rest_ys
               else
                 [x] @ merge rest_xs second,
-          } [first, second]
+          }
         in
         let rec merge_sort = fun array =>
-          match {
+          array
+          |> match {
             [] => [],
             [x] => [x],
             xs =>
               let length = %array/length% array in
-              let half = (length / 2) |> std.number.floor in
+              let half = std.number.floor (length / 2) in
               let left = %array/slice% 0 half array in
               let right = %array/slice% half length array in
               merge (merge_sort left) (merge_sort right)
-          } array
+          }
         in
         merge_sort array,
 
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -644,16 +644,31 @@
           # => [ 1, 2, 4, 5 ]
           ```
         "%
-      #TODO: maybe inline partition to avoid contract checks?
       = fun cmp array =>
-        let length = %array/length% array in
-        let first = %array/at% array 0 in
-        let rest = %array/slice% 1 length array in
-        let parts = partition (fun x => (cmp x first == 'Lesser)) rest in
-        if length <= 1 then
-          array
-        else
-          (sort cmp (parts.right)) @ [first] @ (sort cmp (parts.wrong)),
+        let rec merge = fun first second =>
+          match {
+            [[], ys] => ys,
+            [xs, []] => xs,
+            [[x, ..rest_xs],  [y, ..rest_ys]] =>
+              if cmp x y == 'Greater then
+                [y] @ merge first rest_ys
+              else
+                [x] @ merge rest_xs second,
+          } [first, second]
+        in
+        let rec merge_sort = fun array =>
+          match {
+            [] => [],
+            [x] => [x],
+            xs =>
+              let length = %array/length% array in
+              let half = (length / 2) |> std.number.floor in
+              let left = %array/slice% 0 half array in
+              let right = %array/slice% half length array in
+              merge (merge_sort left) (merge_sort right)
+          } array
+        in
+        merge_sort array,
 
     flat_map
       : forall a b. (a -> Array b) -> Array a -> Array b
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -632,18 +632,33 @@
 
           # Examples
 
-          ```nickel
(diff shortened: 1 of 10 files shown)


L0r3m1p5um:
As an aside, there's something in this implementation that the tree-sitter parser fails on. I haven't identified what it is yet.

core/stdlib/std.ncl
```diff
@@ -644,16 +644,31 @@
...
+        let rec merge = fun first second =>
+          match {
+            [[], ys] => ys,
+            [xs, []] => xs,
+            [[x, ..rest_xs],  [y, ..rest_ys]] =>
+              if cmp x y == 'Greater then
+                [y] @ merge first rest_ys
+              else
+                [x] @ merge rest_xs second,
+          } [first, second]
```
```suggestion
          [first, second]
          |> match {
            [[], ys] => ys,
            [xs, []] => xs,
            [[x, ..rest_xs],  [y, ..rest_ys]] =>
              if cmp x y == 'Greater then
                [y] @ merge first rest_ys
              else
                [x] @ merge rest_xs second,
          }
```

core/stdlib/std.ncl
```diff
@@ -644,16 +644,31 @@
...
+          match {
+            [] => [],
+            [x] => [x],
+            xs =>
+              let length = %array/length% array in
+              let half = (length / 2) |> std.number.floor in
+              let left = %array/slice% 0 half array in
+              let right = %array/slice% half length array in
+              merge (merge_sort left) (merge_sort right)
+          } array
```
```suggestion
          array
          |> match {
            [] => [],
            [x] => [x],
            xs =>
              let length = %array/length% array in
              let half = (length / 2) |> std.number.floor in
              let left = %array/slice% 0 half array in
              let right = %array/slice% half length array in
              merge (merge_sort left) (merge_sort right)
          }
```

core/stdlib/std.ncl
```diff
@@ -644,16 +644,31 @@
...
+                [x] @ merge rest_xs second,
+          } [first, second]
+        in
+        let rec merge_sort = fun array =>
+          match {
+            [] => [],
+            [x] => [x],
+            xs =>
+              let length = %array/length% array in
+              let half = (length / 2) |> std.number.floor in
```
```suggestion
              let half = std.number.floor (length / 2) in
```

yannham (commented):
Do we have some vague benches to verify that:

1. O(nln(n)) asymptotically
2. Reasonably faster than the previous implementation

jneem:
On my phone so I can't easily check, but don't we have fast appends and slow prepends? (I.e `ys @ [y]` is fast but `[y] @ ys` is slow)

yannham:
> On my phone so I can't easily check, but don't we have fast appends and slow prepends? (I.e ys @ [y] is fast but [y] @ ys is slow)

I believe as well. So we should rather be merging the array from the end I guess?

L0r3m1p5um:
> Do we have some vague benches to verify that:
> 
>     1. O(nln(n)) asymptotically
> 
>     2. Reasonably faster than the previous implementation

Here's what I measured for `std.array.range 0 size |> std.array.sort std.number.compare`, where `size` ranges from 0 to 1000 for the new implementation:

sort_to_1000

For comparison, here's the current implementation (only up to 150 because the benchmark becomes very slow very quickly after that):
sort_current

> On my phone so I can't easily check, but don't we have fast appends and slow prepends? (I.e ys @ [y] is fast but [y] @ ys is slow)

I didn't realize this, I'll try it out and measure again. Does that mean that popping elements off the beginning with pattern matching like I'm doing is also slow?

jneem:
I think slicing (and therefore also pattern matching) should be fast either way

L0r3m1p5um:
I tried refactoring to use appends, but the performance seems basically identical. Here's how I tried changing the `merge` function. Let me know if you notice anything wrong that would've caused me not to see the expected improvement.

```nickel
        let rec merge = fun first second =>
          [first, second]
          |> match {
            [[], ys] => ys,
            [xs, []] => xs,
            [xs,  ys] =>
              # Split each array into its last element and the rest of the array
              let length_xs = %array/length% xs in
              let x = %array/at% xs (length_xs - 1) in
              let rest_xs = %array/slice% 0 (length_xs - 1) xs in

              let length_ys = %array/length% ys in
              let y = %array/at% ys (length_ys - 1) in
              let rest_ys = %array/slice% 0 (length_ys - 1) ys in

              if cmp x y == 'Greater then
                merge rest_xs second @ [x]
              else
                merge first rest_ys @ [y],
          }
```

Here's the performance of I measured:
sort_1000_append

Since there wasn't a difference in performance I kept the original implementation since the implementation is a lot cleaner since I can just destructure the arrays.

I also made one more change, which was to add a couple examples to the docs. Mostly I wanted to add a test for how it sorts elements that are different but equal under the comparison, like in the sorting strings by length example. I figure it's better not to change the sorting order in that case without good reason. I checked how this new implementation handles this relative to the current one and they appear to be the same.

yannham (approved):
Perfect, looks good, thanks. I don't remember the specifics of the persistent vector implementation we use; I was pretty sure appending at the beginning should be linear in the size of the array, but you measurement seems to show this is not the case, because I suppose that would make the whole algorithm at least `n^2`, which doesn't look like your curve.

Anyway this is a massive improvement already over what used to be here 👍

PR #2648, opened 2026-07-29
Fix `std.number.floor` and `std.array.range_step` functions

### LINKED ISSUE #2645: range_step sometimes incorrectly drops the last element of the range
The example in the documentation says:
```nickel
std.array.range_step (-1.5) 2 0.5
# => [ -1.5, -1, -0.5, 0, 0.5, 1, 1.5 ]
```

This is as expected because the step size divides the range evenly. However, consider the following two examples:

```nickel
std.array.range_step 0 16 8
# => [ 0, 8 ]
std.array.range_step 0 17 8
# => [ 0, 8 ]
```

The step size divides evenly as before in the first example, so the result is fine. But I would expect the result of the second expression to be `[0, 8, 16]`.

### LINKED ISSUE #2646: std.number.floor is incorrect for negative integers
I assume the following standard definitions for `floor` and `ceil` (see [Wikipedia](https://en.wikipedia.org/wiki/Floor_and_ceiling_functions)):

- `floor(x)`: biggest whole number that is less than or equal to `x`.
- `ceil(x)`: smallest whole number that is greater than or equal to `x`.

So I would expect `std.number.floor (-2) == -2`. However, currently, `std.number.floor (-2) == -3`. The other cases seem to work fine.

L0r3m1p5um:
Fixes #2646
Fixes #2645

This fixes a couple of bugs in the standard library:
- `std.number.floor` works incorrectly for negative integers
- `std.number.range_step` sometimes drops the last generated number

This also adds `std.number.ceil` to the standard library since it's needed to fix the issue with `range_step`

### DIFF
--- a/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_range_reversed_indices.ncl.snap
+++ b/cli/tests/snapshot/snapshots/snapshot__eval_stderr_array_range_reversed_indices.ncl.snap
@@ -4,9 +4,9 @@ expression: err
 ---
 error: contract broken by the caller of `range`
        invalid range
-    ┌─ <stdlib/std.ncl>:842:9
+    ┌─ <stdlib/std.ncl>:849:9
     │
-842 │       | std.contract.unstable.RangeFun Dyn
+849 │       | std.contract.unstable.RangeFun Dyn
     │         ---------------------------------- expected type
     │
     ┌─ [INPUTS_PATH]/errors/array_range_reversed_indices.ncl:3:19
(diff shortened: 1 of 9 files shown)


mboes:
For `floor` (mutatis mutandis for `ceil`), note that an alternative implementation is

```
let floor = fun x => let r = x % 1 in if r >= 0 then x - r else x - 1 - r
```

By dispatching on the remainder instead of on the argument `x`, there are only two cases to consider instead of three.

mboes:
Or perhaps better yet, implement in terms of the existing `std.number.truncate`:

```nickel
floor = fun x =>
  let t = std.number.truncate x in
  if t <= x then t else t - 1,

ceil = fun x =>
  let t = std.number.truncate x in
  if t >= x then t else t + 1,
```

yannham:
I like the alternative implementation without truncate. One (allegedly minor) issue with truncate is it currently fires the `Number` contract a second time on the argument. Given it's so simple, it can be easily inlined manually. Though you could argue that this is the problem of the interpreter which should be able to elide the input contract, which might happen in the future...

yannham (approved):
Nice, thanks!

PR #2650, opened 2026-07-30
Partially monomorphize type of std.array.chunk

mboes:
Equality tests are currently not allowed on generic variables. For the
same reason that the type of `std.array.elem` is more monomorphic than
we'd like, the key selection function in `std.array.chunk` should also
not return a key at a generic type.

### DIFF
--- a/core/stdlib/std.ncl
+++ b/core/stdlib/std.ncl
@@ -1149,8 +1149,10 @@
         |> sort cmp
         |> dedup_sorted cmp,
 
+    # **Warning**: see the note at top of `elem` for why the key type is
+    # not polymorphic.
     chunk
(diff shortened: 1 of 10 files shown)


core/stdlib/std.ncl
```diff
@@ -1149,8 +1148,10 @@
         |> sort cmp
         |> dedup_sorted cmp,
 
+    # **Warning**: see the note at top of `elem` for why the key type is
+    # not polymorphic.
     chunk
-      : forall a k. (a -> k) -> Array a -> Array (Array a)
+      : forall a. (a -> Dyn) -> Array a -> Array (Array a)
       | Dyn -> Dyn -> Array std.array.NonEmpty
```
```suggestion
```

yannham (approved):

PR #2652, opened 2026-08-02
Replace reference to std.array.length in documentation of std.string.substring

ldesgoui:


L0r3m1p5um (approved):

L0r3m1p5um:
Good catch, thanks for the contribution!

PR #2654, opened 2026-08-03
Types & eval internals for effects

twitchyliquid64:
Heya! This is a draft of the internals of effect handling, as we sketched out on the nickel discord channel.

Most code here is fairly unremarkable, I will point out the interesting bits:

 - With the eager checking of contracts, it seemed a little messy to have a contract body calling out to an effect, so I've made that illegal for now. Detection is powered by a counter in the vm, and the contract body is wrapped with an op that decrements the counter.
 - Syntactically, effects are invoked like `%effect% <ident> <n-ary args>`. Having the effect name being an ident instead of something that can be dynamically dispatched felt right, but we can always loosen.
 - In the representation, effects are just `NAryOp::EffectRaw(ident)` with the next value being the args.
 - When an effect op is encountered, its args would have already been forced. The `eval()` call stack unwinds with the details of the effect ready for handling by the caller.

A future PR can rework the `Program` layer to handle effects, in particular i was thinking we have a wrapper type around the args that lets the caller deserialize the args somewhere but not mutate them.


core/src/ast/compat.rs
```diff
@@ -1095,6 +1097,9 @@ impl FromAst<PrimOp> for TermPrimOp {
             PrimOp::RecordUnsealTail => TermPrimOp::NAry(term::NAryOp::RecordUnsealTail),
             PrimOp::LabelInsertTypeVar => TermPrimOp::NAry(term::NAryOp::LabelInsertTypeVar),
             PrimOp::ArraySlice => TermPrimOp::NAry(term::NAryOp::ArraySlice),
+            // `%effect%` desugars to a single-arg `EffectRaw` OpN — the mainloop wraps the
+            // payload in `Force` on entry so the arg is deep-evaluated before the yield.
```
yannham, line 1101: Nit: this feels a bit out of place.

```suggestion
```

core/src/eval/mod.rs
```diff
@@ -694,14 +699,38 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
+    /// loop:
+    ///
+    /// 1. Call `eval_step(initial_closure)`.
+    /// 2. If the result is `Closure`, evaluation is done.
+    /// 3. If the result is `Effect { name, args, .. }`, perform the effect and call
+    ///    `eval_step` again with the effect's result value (`result.into()`).
+    ///
+    /// Non-effectful callers should keep using [`Self::eval_closure`]; it treats an unhandled
+    /// effect as [`EvalErrorKind::UnhandledEffect`].
+    pub fn eval_step(&mut self, closure: Closure) -> Result<EvalOutcome, EvalError> {
```
yannham, line 720: Very nit, but we have so many `eval` variants, we should probably give it a more specific name that reflects the fact that it is to be used only for effects.

core/src/eval/mod.rs
```diff
@@ -855,6 +889,36 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
                     }
                 }
                 ValueContentRef::Term(Term::OpN(data)) => {
+                    // For EffectRaw, we need to (a) forbid firing effects during a contract
+                    // check, and (b) deep-evaluate each argument before yielding the effect to
+                    // the caller. We achieve (b) by wrapping each argument in `UnaryOp::Force`
+                    // — the OpN machinery walks each pending argument to WHNF, and Force's WHNF
+                    // result is the deep-evaluated payload.
```
yannham, line 896:
```suggestion
```

core/src/eval/mod.rs
```diff
@@ -855,6 +889,36 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
+                            unreachable!()
+                        };
+                        break Err(Box::new(EvalErrorKind::EffectInContract {
+                            name: *name,
+                            pos: pos_idx,
+                        }));
+                    }
+
+                    let wrap = |value: NickelValue| {
+                        if is_effect {
```
yannham, line 910: Nit style: avoid an unreachable and the let-else branch.

```suggestion
                    if let NAryOp::EffectRaw(name) = &data.op && self.contract_depth > 0 {
                        break Err(Box::new(EvalErrorKind::EffectInContract {
                            name: *name,
                            pos: pos_idx,
                        }));
                    }

                    let wrap = |value: NickelValue| {
                        if matches!(&data.op, NAryOp::EffectRaw(_)) {
```

core/src/eval/operation.rs
```diff
@@ -170,13 +172,28 @@ impl<'ctxt, R: ImportResolver, C: Cache> VirtualMachine<'ctxt, R, C> {
...
-                Ok(next)
+                Ok(EvalOutcome::Closure(next))
+            } else if let crate::term::NAryOp::EffectRaw(name) = opn_cont.op {
+                // All args were wrapped in `Force` when the effect was lowered from `Term::OpN`
+                // (see the `Term::OpN` branch in `eval_closure_impl`), so their evaluated values
+                // are the deep-forced payloads the caller expects.
+                let args = opn_cont
+                    .evaluated
+                    .into_iter()
+                    .map(|(cl, _)| cl.value)
```
yannham, line 183: I haven't thought about those things in a while, but I don't think you can discard the environment like that, even if the value has been forced. For example a record `{x = 1+1}` will still have the shape `{x = %1 } [in env x <- 2]` after `force`. If you keep the `value` you'll end up with an orphan, meaningless `{x = %1}`.

If you want it to really be environment-less, we would need to substitute all variables in it (what we do prior to serialization, for example), but I'm not sure it's necessary. The simplest approach is likely to return `Closure`s to the caller instead, letting them decide what to do (explore the value lazily with the environment at hand, or substituting everything / serializing).

yannham (commented):
Thanks for putting this together. Overall I haven't thought very deeply about the interface, but I think I like the idea of a step function. We'll probably need more glue code if we want some kind of pluggable effect extensions in the future, but for now this entirely sidesteps the questions like "what should be the exchange format for effect arguments and how should they return value" and "how should the VM call to the external effect implementation", by reversing the control flow. But this still lets implement something more pluggable in the future on top of it.

I have two remaining interrogations:

- I'm not entirely sure to understand the whole thing around contracts. It does make the code more complex (I'm always nervous to change rules around sealing, this stuff is pretty subtle), and I think I'm missing the justification. In a lazy language the order of evaluation of pretty much any non trivial expression is rather unspecified. Could you please explain the reasoning behind them being special cased?
- On a simpler note, I wonder if we should feature gate this for now. Though the changes are really minimal and likely to have no impact on normal programs, it's more as a way to mark it as experimental/unstable feature and API.

